PR #2, opened 2020-01-18
README updates after test deployment

austinbyers:
## Background
After deploying and logging in, noticed a few things that look like problems that are expected - we should point these out in the README

This is also testing the CircleCI integration
## Changes

* Update README

## Testing

* CI


austinbyers:
Doesn't look like `CODEOWNERS` has been picked up yet, it's in the right location it should be soon

PR #4, opened 2020-01-18
Fix pagination for alerts and make assigning severtity more robust

rleighton:
## Background
Pagination for alerts was not working.

## Testing
Manual testing thru UI.
mage test:unit passes


internal/log_analysis/alerts_api/api/list_alerts.go
```diff
@@ -76,9 +78,16 @@ func alertItemsToAlertSummary(items []*models.AlertItem) ([]*models.AlertSummary
...
 		if err != nil {
-			return nil, err
+			zap.L().Error("could not get rule severity",
+				zap.String("ruleId", ruleID),
+				zap.Error(err))
+
+			// a 404 means cannot find rule, treat as nil
+			if !strings.Contains(err.Error(), "getRuleNotFound") {
+				return nil, err
+			}
```
kostaspap: Might make sense for the 404 case to just log a WARN instead of an ERROR since the ListAlerts operation still succeeds. 
A 404 here is actually an "expected" state - it can be that a rule has been deleted after it created some alerts.

internal/log_analysis/alerts_api/table/list_alerts.go
```diff
@@ -52,8 +50,7 @@ func (table *AlertsTable) ListAlerts(exclusiveStartKey *string, pageSize *int) (
 			return nil, nil, err
 		}
 		scanExclusiveStartKey = map[string]*dynamodb.AttributeValue{
-			"creationTime": {S: paginationKey.CreationTime},
-			"alertId":      {S: paginationKey.AlertID},
+			"alertId": {S: paginationKey.AlertID},
```
kostaspap: Hmmm, this is actually problematic. The expected UX here should be that the alerts are listed by reverse chronologic order - users should be seeing the most recent alerts at the top of the page. 

Let's get this PR through, but we should fix this issue at some point.

internal/log_analysis/alerts_api/api/list_alerts.go
rleighton: Not sure what you mean, in the case of 404 we do not log anything we just continue. The logic is to return the err if the err does NOT have message "getRuleNotFound".

internal/log_analysis/alerts_api/table/list_alerts.go
rleighton: Agree.

internal/log_analysis/alerts_api/api/list_alerts.go
rleighton: Nevermind, I see. Moving logging into if clause...

PR #5, opened 2020-01-18
Fixed rule testing

nhakmiller:
@rleighton identified that the rule testing functionality was not working. This was a result of decoupling the `policy-engine` and `rules-engine` on the backend without updating the `TestPolicy` endpoint of the `AnalysisApi` (which was shared for policy and rule testing).

* Added logic to the `AnalysisApi` to handle policy testing and rule separately
* Modified the `rules-engine` to handle direct analysis of rules

* Deployed to dev account and tested manually
* Automated tests with `mage test:ci`


internal/log_analysis/rules_engine/src/main.py
```diff
@@ -57,6 +63,34 @@ def lambda_handler(event: Dict[str, Any], unused_context) -> None:
     else:
         logger.info("no matches found")
 
+# Evaluates a single rule against a set of events, and returns the results
+# Currently used for testing policies directly
+def direct_analysis(event: Dict[str, Any]) -> Dict[str, Any]:
+        # Since this is used for testing single rules, it should only ever have one rule
```
rleighton: I would check for multiple rules and throw a nice error(since you assume only 1). If we got a regression in the code it would crash in a way that would make debugging harder.

internal/core/analysis_api/handlers/test_policy.go
```diff
@@ -150,6 +129,114 @@ func TestPolicy(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyR
...
 
+//nolint:dupl
+func getRuleResults(input *models.TestPolicy) (*enginemodels.RulesEngineOutput, *events.APIGatewayProxyResponse) {
+	// Build the list of events to run the rule against
+	inputEvents := make([]enginemodels.Event, len(input.Tests))
+	for i, test := range input.Tests {
+		// Unmarshal event into object form
+		var attrs map[string]interface{}
+		if err := jsoniter.UnmarshalFromString(string(test.Resource), &attrs); err != nil {
+			return nil, badRequest(fmt.Errorf("tests[%d].event is not valid json: %s", i, err))
```
rleighton: Rather than use fmt.Errorf() I think we should start using errors.Wrapf() and errors.Wrap() everywhere. These functions do the same thing BUT include a stack trace to the caller which us super useful. Most of the new log processing code has been switched over to using it. It has made debugging _much_ faster. When used with zap.Error() the log has 2 fields (the result of Error() which has all the wrapped message stacks and a second key with the full call stack). Errors propagated to our UI do not see the call stack but they are in the CW logs.

api/gateway/analysis/api.yml
```diff
@@ -742,6 +742,13 @@ definitions:
     items:
       $ref: '#/definitions/UnitTest'
 
+  AnalysisType:
+    type: string
+    enum:
+      - POLICY
+      - RULE
```
kostaspap: Is there planned a PR to update Frontend and/or AppSync resolvers? We probably need those changes as well in order to fix the current Rule testing issues.

internal/core/analysis_api/handlers/test_policy.go
```diff
@@ -48,54 +48,32 @@ func TestPolicy(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyR
...
-	resources := make([]enginemodels.Resource, len(input.Tests))
-	for i, test := range input.Tests {
-		// Unmarshal resource into object form
-		var attrs map[string]interface{}
-		if err := jsoniter.UnmarshalFromString(string(test.Resource), &attrs); err != nil {
-			return badRequest(fmt.Errorf("tests[%d].resource is not valid json: %s", i, err))
+	var results *enginemodels.PolicyEngineOutput
+	// Build the policy engine request
+	if input.AnalysisType == "RULE" {
+		ruleResults, response := getRuleResults(input)
```
kostaspap: nit: Might make sense to rename `response` to `errorResponse`. It seems weird to see the code we normally use for error checking occurring in a "response" object.

api/gateway/analysis/api.yml
nhakmiller: That has already been handled by @drixta .

internal/core/analysis_api/handlers/test_policy.go
nhakmiller: Seems reasonable, updated.

internal/core/analysis_api/handlers/test_policy.go
nhakmiller: Ok, I updated this in the two places I used it, however I think updating the entire API to use this is outside the scope of this PR for now.

internal/log_analysis/rules_engine/src/main.py
nhakmiller: Ok, not sure what to return in that case so I returned nothing and logged an error. This should also throw an error in the go code when it trys to unmarshal nothing into an output struct, which I think is reasonable behavior.

internal/core/analysis_api/handlers/test_policy.go
```diff
@@ -48,54 +48,32 @@ func TestPolicy(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyR
...
+		if response != nil {
+			return response
 		}
-
-		resources[i] = enginemodels.Resource{
-			Attributes: attrs,
-			ID:         testResourceID + strconv.Itoa(i),
-			Type:       string(test.ResourceType),
+		results = &enginemodels.PolicyEngineOutput{
+			Resources: []enginemodels.Result{},
```
austinbyers: Since you know how many results there will be:

```go
Resources: make([]enginemodels.Result{}, 0, len(ruleResults.Events))
```

internal/core/analysis_api/handlers/test_policy.go
austinbyers: Note that `fmt.Errorf` also supports a new `%w` flag which seems to do something similar (https://blog.golang.org/go1.13-errors)

internal/log_analysis/rules_engine/src/main.py
```diff
@@ -18,20 +18,26 @@
...
 from io import TextIOWrapper
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Union
 
 import boto3
 
 from .engine import Engine
 from .logging import get_logger
 from .sqs import send_to_sqs
+from .rule import Rule
```
austinbyers: Nit: this should be before `from .sqs`

internal/log_analysis/rules_engine/src/main.py
austinbyers: I think returning an explicit error is also fine. E.g.

```python
count = len(event.get('rules'))
if count != 1:
    raise RuntimeError('exactly one rule expected, found {}'.format(count))
```

internal/core/analysis_api/handlers/test_policy.go
nhakmiller: Updated.

internal/log_analysis/rules_engine/src/main.py
nhakmiller: Updated.

internal/log_analysis/rules_engine/src/main.py
nhakmiller: I like this better than my previous approach, switched to this.

PR #6, opened 2020-01-18
Updating mage setup command to download tools inside repository folder

kostaspap:
Running `mage setup` command failed with the following error on my laptop: 
```
Warning: Failed to create the file /usr/local/bin/swagger: Permission denied
0 17.1M    0 16360    0     0  12347      0  0:24:16  0:00:01  0:24:15 12347
curl: (23) Failed writing body (0 != 16360)
Error: running "curl -o /usr/local/bin/swagger -fL https://github.com/go-swagger/go-swagger/releases/download/v0.21.0/swagger_darwin_amd64" failed with exit code 23
```

It seems that setup was trying to download swagger file inside `/user/local/bin/swagger`. A non-root user by default has `rx` rights but no `w` rights. I think it is more safe choice to just download swagger in a subdirectory of the current directory (the directory where panther is checked out). It is also less likely that we mess with users laptop setup by overwriting previously installed `swagger` with the version needed by Panther. 

* `golangci-lint`, `swagger`, `venv` are all located inside `./.setup` folder. 
* Upgraded golangci-lint version to 1.22.2
* Upgraded go dependencies

*  Tested all mage targets and they succeeded
`mage clean setup build:api build:lambda fmt test:lint test:unit test:cover test:ci`
* CircleCI build succeeds
* Was able to deploy panther to my account

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -22,6 +22,8 @@ github.com/aws/aws-lambda-go v1.13.3 h1:SuCy7H3NLyp+1Mrfp+m80jcbi9KYWAs9/BXwppwR
 github.com/aws/aws-lambda-go v1.13.3/go.mod h1:4UKl9IzQMoD+QF79YdCuzCwp8VbmG4VAQwij/eHl5CU=
 github.com/aws/aws-sdk-go v1.28.3 h1:FnkDp+fz4JHWUW3Ust2Wh89RpdGif077Wjis/sMrGKM=
 github.com/aws/aws-sdk-go v1.28.3/go.mod h1:KmX6BPdI08NWTb3/sm4ZGu5ShLoqVDhKgpiN924inxo=
+github.com/aws/aws-sdk-go v1.28.5 h1:yYeWPM8w5FoIj3Lo0BDZxRyDpTveKTq/qvnIEPBnev8=
+github.com/aws/aws-sdk-go v1.28.5/go.mod h1:KmX6BPdI08NWTb3/sm4ZGu5ShLoqVDhKgpiN924inxo=
 github.com/cenkalti/backoff v2.2.1+incompatible h1:tNowT99t7UNflLxfYYSlKYsBpXdEet03Pg2g16Swow4=
 github.com/cenkalti/backoff v2.2.1+incompatible/go.mod h1:90ReRw6GdpyfrHakVjL/QHaoyV4aDUVVkXQJJJ3NXXM=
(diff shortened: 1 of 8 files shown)


tools/mage/util.go
```diff
@@ -33,6 +34,11 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+var (
```
3nvi: should these perhaps be added to a config file like `panther_config.yml`?

kostaspap: I am treating this folder more like we treat the `out` folder - everything is expected to get there by convention.

I am worried of complications that come with it, e.g. right now I am adding the `.setup` folder in the gitignore file, if it is configurable we are not able to do so, or the possibility of someone just misconfiguring the folder and ovewriting local changes (e.g. if he would set as folder an already existing folder), etc

I am thinking it is actually safer and less error-prone to go with a standard location.

rleighton: I agree with Kostas. Are there any side effects that are bad if mage is run from some other dir besides the root?

kostaspap: > Are there any side effects that are bad if mage is run from some other dir besides the root?

Someone can run mage only from root -  `magefile.go` is [located](https://github.com/panther-labs/panther/blob/master/magefile.go) in root.

go.sum
```diff
@@ -252,6 +258,10 @@ golang.org/x/tools v0.0.0-20191029190741-b9c20aec41a5/go.mod h1:b+2E5dAYhXwXZwtn
 golang.org/x/tools v0.0.0-20191125144606-a911d9008d1f/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20200116225955-84cebe10344f h1:ILR8hSRYwfkRGnA5a6c/vTjIFGkLibNIFxGiosQo+LM=
 golang.org/x/tools v0.0.0-20200116225955-84cebe10344f/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
+golang.org/x/tools v0.0.0-20200117220505-0cba7a3a9ee9 h1:KOkk4e2xd5OeCDJGwacvr75ICCbCsShrHiqPEdsA9hg=
```
austinbyers: There are multiple copies of the same library, `go mod tidy` can help clean some of them up

PR #7, opened 2020-01-20
Restarting frontend on deploy

kostaspap:
Modified deploy script so that it restarts frontend (stopping Fargate tasks and waiting for ones to start running)

* Mage deploy will not stop running Frontend Fargate tasks and wait until new tasks are up and running

* Deployed to my account, seen expected behavior. Verified in ECS console that old tasks were stopped and replaced with new ones

```
deploy: panther-buckets: no changes needed
build:lambda: go build internal/*/main (21 binaries)
deploy: cloudformation package deployments/template.yml => out/deployments/package.template.yml
deploy: s3://panther-source-123456789012-eu-west-1/layers/python-analysis.zip exists and is up to date
deploy: ACM certificate already exists
deploy: panther-app: CreateChangeSet: CREATE_IN_PROGRESS
deploy: panther-app: ExecuteChangeSet: UPDATE_IN_PROGRESS
deploy: panther-app: ExecuteChangeSet: UPDATE_COMPLETE_CLEANUP_IN_PROGRESS
deploy: panther-app: ExecuteChangeSet: UPDATE_COMPLETE
Restarting frontend server...
Waiting for frontend server to become active
Finished restarting frontend

Panther URL = https://web-1820136089.eu-west-1.elb.amazonaws.com

```

### DIFF
--- a/deployments/template.yml
+++ b/deployments/template.yml
@@ -390,3 +390,6 @@ Outputs:
   WebApplicationCertificateArn:
     Description: ARN of the certificate used by the web application
     Value: !Ref WebApplicationCertificateArn
+  WebApplicationClusterName:
+    Description: ARN of the certificate used by the web application
+    Value: !Ref WebApplicationCluster
--- a/tools/mage/deploy.go
+++ b/tools/mage/deploy.go
@@ -25,15 +25,18 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cognitoidentityprovider"
+	"github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/aws/aws-sdk-go/service/lambda"
 	"github.com/aws/aws-sdk-go/service/s3"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/magefile/mage/mg"
 	"github.com/magefile/mage/sh"
+	"github.com/pkg/errors"
 
 	"github.com/panther-labs/panther/api/lambda/users/models"
 	"github.com/panther-labs/panther/pkg/shutil"
@@ -115,6 +118,10 @@ func Deploy() error {
 		return err
 	}
 
+	if err := restartFrontend(awsSession, outputs["WebApplicationClusterName"]); err != nil {
+		return err
+	}
+
 	if err := inviteFirstUser(awsSession, outputs["UserPoolId"]); err != nil {
 		return err
 	}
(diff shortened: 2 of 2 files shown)


tools/mage/deploy.go
```diff
@@ -128,6 +135,64 @@ func Deploy() error {
...
+		_, err := ecsClient.StopTask(&ecs.StopTaskInput{
+			Cluster: aws.String(clusterName),
+			Task:    taskArn},
+		)
+		if err != nil {
+			return errors.Wrap(err, "failed to stop running task")
+		}
+	}
+
+	fmt.Println("Waiting for frontend server to become active")
```
3nvi: We should try and see if we could use something better than polling. Could `WaitUntilServicesStable` work?

tools/mage/deploy.go
```diff
@@ -128,6 +135,64 @@ func Deploy() error {
 	return nil
 }
 
+func restartFrontend(awsSession *session.Session, clusterName string) error {
+	fmt.Println("Restarting frontend server...")
+	ecsClient := ecs.New(awsSession)
+
+	alreadyRunningTasks, err := ecsClient.ListTasks(&ecs.ListTasksInput{
```
3nvi: This stops all the running tasks of the cluster. We should be targeting only the tasks of the server service

tools/mage/deploy.go
```diff
@@ -128,6 +135,64 @@ func Deploy() error {
...
+		DesiredStatus: aws.String(ecs.DesiredStatusRunning),
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed to list tasks")
+	}
+	for _, taskArn := range alreadyRunningTasks.TaskArns {
+		if mg.Verbose() {
+			fmt.Printf("Stopping task %s", *taskArn)
+		}
+		_, err := ecsClient.StopTask(&ecs.StopTaskInput{
```
3nvi: Killing the task will eventually cause downtime between the time a task is killed and the time the service becomes stable. I think - unfortunately - out best bet is a rolling update (0% downtime) by forcing a new deployment.

Directly from AWS SDK docs:

```
    // ...
    // For example, you can update a service's tasks to use
    // a newer Docker image with the same image/tag combination (my_image:latest)
    // ...
    ForceNewDeployment *bool `locationName:"forceNewDeployment" type:"boolean"`
```

deployments/template.yml
```diff
@@ -390,3 +390,6 @@ Outputs:
   WebApplicationCertificateArn:
     Description: ARN of the certificate used by the web application
     Value: !Ref WebApplicationCertificateArn
+  WebApplicationClusterName:
+    Description: ARN of the certificate used by the web application
```
3nvi: typo in description

tools/mage/deploy.go
kostaspap: It's the frontend cluster, I don't think we plan to have more services running there? Eg in line 121 I am getting the `WebApplicationClusterName` - it is the cluster for the web app.

tools/mage/deploy.go
kostaspap: That one looks interesting, didn't know this method.

deployments/template.yml
austinbyers: Is it the ARN or the name?

3nvi: It's the name, but the typo is that It's not related to the certificate at all

3nvi:
> What's the high-level reason we need to do this? An updated image isn't automatically picked up by a running task during a deployment?
> 
> Is this still necessary after #11 merges?
> 
> EDIT: I see, this is necessary because the docker image isn't versioned? Would it be better to just version images instead of manually restarting the task on every deploy?

I can tell you that a restart (not a kill & deploy) is the easiest way. It's also a common and [recommended way](https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#ECS.UpdateService):

```
If your updated Docker image uses the same tag as what is in the existing task definition for your service (for example, my_image:latest), you do not need to create a new revision of your task definition. You can update the service using the forceNewDeployment option. The new tasks launched by the deployment pull the current image/tag combination from your repository when they start.
```

kostaspap:
Closing as not required. Similar steps are being performed as part of https://github.com/panther-labs/panther/pull/11

PR #8, opened 2020-01-20
[ENG-1317] Fix missing validations in rule loading

rleighton:
## Background
If the UI pushes incomplete records into ddb, then the rule engine crashes

## Changes
Added logging and validation for all keys referenced

## Testing
Testing by deploying with incomplete ddb records.

### DIFF
--- a/internal/log_analysis/rules_engine/src/engine.py
+++ b/internal/log_analysis/rules_engine/src/engine.py
@@ -44,18 +44,29 @@ class Engine:
         self.logger.info('Retrieved rules in {} seconds'.format(end - start))
         start = default_timer()
 
-        # Importing COMMON module
+        # Importing COMMON module, check all keys (do NOT trust data in ddb!), update lookup table
+        import_count = 0
         for index, raw_rule in enumerate(rules):
+            if 'id' not in raw_rule:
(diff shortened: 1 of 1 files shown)


internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -44,18 +44,29 @@ def populate_rules(self) -> None:
...
+                continue
+            if 'resourceTypes' not in raw_rule:
+                self.logger.error('Rule {} missing resourceTypes'.format(raw_rule['id']))
+                continue
             if raw_rule['id'] == COMMON_MODULE_RULE_ID:
                 Rule(raw_rule['id'], raw_rule['body'])
-                del rules[index]
-                break
-        for raw_rule in rules:
+                continue
```
kostaspap: The rule with ruleId `COMMON_MODULE_RULE_ID` needs to be imported first. The code from that Rule is going to be in a module shared by the rest of the rules.

rleighton: Awkward. Ok.

rleighton: I assume it is an error NOT to find the COMMON_MODULE_RULE_ID , correct?

PR #11, opened 2020-01-21
[ENG-1242][ENG-1272] Build Docker Image from Source

3nvi:
This PR solves the issue of having to manually re-upload the public image on every change and it instead opts for a "build from src" approach. 

Currently, we need CFN outputs in order to build the web image. At the same time, the ECS can't deploy without a specific image (deployment will be stuck in `UPDATE_IN_PROGRESS` if no image is provided). Thus, the order in which things must run is:

1. Deploy all the panther stack apart from ECS
2. Build the image and deploy it to the personal ECR of the user
3. Deploy ECS and point it to the recently uploaded image

This will only have to happen once, since from the 2nd deployment and on, an image will already exist in ECR (the one from the previous deployment), so ECS will not be stuck at all. After the deployment the this image will be overriden by the latest one, since our images are un-versioned and  always using the `:latest` tag.  ECS will then be forced to upgrade to the latest image by manually restarting the related ECS service is through a force update. 

In order to achieve all that, a `Condition` variable has been added to denote whether the stack is being deployed for the first time or not. If it is, then the ECS stack will **not** be created on the initial deployment. Afterwards, the image will be build and deployed and the stack will be redeployed once more. This time, due to the fact that the stack is no longer in its initial deployment, the ECS **will be deployed** and use image that was just built.

Future deployments will **always** deploy the entire stack (including ECS). To clarify again, the reason we don't deploy ECS on the initial deployment is because there is no image present in ECR and ECS will fail without one. At the same time though, the image requires the AppSync & Cognito stack to have been deployed before it can be built. Pointing ECS to a "dummy" image during the initial deployment and then switching it, is dirty and I wouldn't like it.

On future deployments, ECS will still use the previous image. In order to bump it to the new one, we are using the AWS SDK to force a new deployment on the ECS service which will restart the containers and allow them to use the latest image (according to official AWS docs, the image update can't happen automatically without spinning new containers)

####

Now the way front-end gets the variables it needs are the following:

1. The stack gets deployed
2. The outputs of the stack get collected
3. The outputs get written into a `.env` file which can be consumed through the famous [dotenv](https://github.com/motdotla/dotenv) package. A similar port of this package [exists in Golang](https://github.com/joho/godotenv) and we are using it for this purpose
4. The build (front-end) process that's kicked in by the `Dockerfile` adds the contents of the `.env` file as ENV vars in the shell that the front-end building will take place.

This approach has the following benefits:

1. The user can build the application locally through `npm run build` which automatically adds the ENV vars from the file, without an AWS API request.
2. The user can develop the application locally since `.env` contains all the information it would normally need, thus not needing an API request to get values like GraphQL endpoint, UserPoolId, etc.

The `.env` file gets updated **automatically** on every `mage deploy`, so it's sure that it will always be aligned with the user's stack (i.e. if the Cognito User Pool ID changes). 

* Update `.gitignore`, `.dockerignore` & `README`
* Μοve CFN `Parameters` & `Outputs`
* Update `Dockerfile` to use multi-stage builds to reduce the size of the image that's going to be uploaded
* Delete any hardcoded configuration found under `/web/config` since that's remnants of our SaaS era
* Add ECR related CFN template
* Change `mage deploy` script to:
   * Save Master template CFN outputs in a `.env` file
   * Build and deploy the docker image
   * Conditionally deploy the template two times (only once)
   * Conditionally create users and setup TOTP (only once)
   * Conditionally restart the ECS web service (every time apart from initial deployment) 

* Locally and on personal dev account

## Known Issues
 1. ~~The `isInitialDeployment` condition checks for the presence of a `panther-app` stack. If a stack is deployed, it assumes that an image was built and uploaded as well.  Unfortunately, if your token expires halfway through the process (i.e. after the initial deployment and during the image building), the system will be in stale mode. One way of solving this is to check explicitly if there is an image uploaded, but the only reason not to be uploaded is the aforementioned.~~

2. ~~When logging in the docker CLI mistakenly thinks we have added a password at the terminal. Thus it prints `WARNING! Using --password via the CLI is insecure. Use --password-stdin.` What happens is that we use the `--password` flag but through Go while using a variable. Hence there is no security issue present. Should we work around this just so it doesn't get printed or you think it's fine?~~

3. If you use `aws-vault`, you need to enable longer sessions through the `--assume-role-ttl` when deploying. For example, `aws-vault exec <ROLE> --assume-role-ttl=45m -- mage deploy`

@austinbyers @nhakmiller If possible, let me know on issues (1) & (2)  above


tools/mage/deploy.go
```diff
@@ -302,3 +342,76 @@ func inviteFirstUser(awsSession *session.Session, userPoolID string) error {
 
 	return nil
 }
+
+// Functions that build a personalized docker image from source, while pushing it to the private image repo of the user
```
nhakmiller: These functions should probably go in a separate file, such as `tools/mage/deploy_frontend` or something similar. See `tools/mage/deploy_certificate` for an example.

tools/mage/deploy.go
```diff
@@ -302,3 +342,76 @@ func inviteFirstUser(awsSession *session.Session, userPoolID string) error {
...
+func buildAndPushImageFromSource(awsSession *session.Session, imageTag string) error {
+	fmt.Println("docker: Requesting access to remote image repo...")
```
nhakmiller: In general we have not been capitalizing the first word and we have not been using trailing ellipsis for these log messages.

tools/mage/deploy.go
```diff
@@ -302,3 +342,76 @@ func inviteFirstUser(awsSession *session.Session, userPoolID string) error {
...
+		conventionalOutputs[strcase.ToScreamingSnake(k)] = v
+	}
+	if err := godotenv.Write(conventionalOutputs, filename); err != nil {
+		return err
+	}
+	return nil
+}
+
+// makes sure to force a new ECS deployment on the service server so that the latest docker image can be applied
+func restartFrontendServer(awsSession *session.Session, cluster string, service string) error {
```
nhakmiller: Does this invalidate https://github.com/panther-labs/panther/pull/7? From @kostaspap

tools/mage/deploy.go
```diff
@@ -111,20 +119,51 @@ func Deploy() error {
...
+	if err = buildAndPushImageFromSource(awsSession, outputs["WebApplicationImage"]); err != nil {
 		return err
 	}
 
-	if err := initializeAnalysisSets(awsSession, outputs["AnalysisApiEndpoint"], &config); err != nil {
-		return err
+	// If we previously deployed the app for the first time, then we didn't deploy the ECS stack due to the image
+	// not being yet present. To resolve this we re-deploy in order for the ECS stack to be created. This
+	// will only happen once, during the initial installation so that's why we have encapsulated all the "run-once"
+	// logic inside this condotional
```
nhakmiller: This comment could be more general about the following block consisting of first time deploy logic, with this comment existing directly above `deployParams["IsInitialDeployment"] = "false"`, and separate comments above `enableTOTP` and `inviteFirstUser` etc. explaining why those only run on the initial deployment.

tools/mage/deploy.go
```diff
@@ -111,20 +119,51 @@ func Deploy() error {
...
+		outputs, err = getStackOutputs(awsSession, applicationStack)
+		if err != nil {
+			return err
+		}
+
+		if err := enableTOTP(awsSession, outputs["WebApplicationUserPoolId"]); err != nil {
+			return err
+		}
+
+		if err := inviteFirstUser(awsSession, outputs["WebApplicationUserPoolId"]); err != nil {
```
nhakmiller: By putting this in this block, we lose the previous implicit functionality of re-creating the first user if it gets deleted through the UI or AWS console. Not a huge deal IMO, but it is worth noting.

tools/mage/deploy.go
austinbyers: Why not version the images instead of manually restarting the server?

tools/mage/deploy.go
3nvi: gotcha, will update

tools/mage/deploy.go
3nvi: sure thing. I'll move them

tools/mage/deploy.go
3nvi: If we version them, then we have to make sure to update the cloudformation of the ECS on every mage deploy.

Docker doesn't have the concept of versioning while using the most recent. Thus you can't say "publish a new version and just use the most recent one". You have to be explicit. Even the `:latest` tag that we are using is an explicit version. It doesn't mean "latest" as in "most recent", but "latest" as in "the keyword latest". Docker images don't work the same as dependencies do.

To avoid the need for constant manual updates (which the user can't perform), we just use the same image version.

3nvi: > Placeholder text on add SQS Queue Destination

It does. Kostas approached it differently, but as stated:

* It needs to kill only the tasks of the related ECS service (not a ll)
* It will result into downtime during the killing of a task and before the provisioning of another
* The polling might take up to 1 minute to finish and it needs to target the tasks of the correct service 

The [official docs](https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/) state the following:

```
If your updated Docker image uses the same tag as what is in the existing task definition for your service (for example, my_image:latest), you do not need to create a new revision of your task definition. You can update the service using the forceNewDeployment option. The new tasks launched by the deployment pull the current image/tag combination from your repository when they start.
```

tools/mage/deploy.go
3nvi: Will make sure to add them

tools/mage/deploy.go
austinbyers: > If we version them, then we have to make sure to update the cloudformation of the ECS on every mage deploy.

We could do that automatically, right? Publish a new (versioned) image, pass that image number into the cfn parameters for the next step. I don't fully understand how this works yet, so I trust your judgment on what works best

tools/mage/deploy.go
3nvi: You can't delete the user through the UI, only through the console. I can easily put it outside the conditional block if you feel it's beneficial. Let me know :)

README.md
```diff
@@ -32,27 +32,28 @@ _NOTE: Panther is currently in beta._
 
 To deploy Panther from source:
 
-1. Install Go 1.13+, Node 10+, and Python 3.7+
+1. Install Docker 17+ [using any of the recommended methods](https://docs.docker.com/install/)
```
austinbyers: I would list this as part of the step below, since it can also be installed with `brew`

```
1. Install Go 1.13+, Node 10+, Python 3.7+ and [Docker 17+](https://docs.docker.com/install/)
    - For mac / homebrew, `brew install docker go node python3`
```

deployments/panther_config.yml
```diff
@@ -47,6 +47,12 @@ AppParameterValues:
   # PROVIDE A CERTIFICATE FOR PRODUCTION USE
   WebApplicationCertificateArn: ''
 
+  # The version of the docker image to upload that will be used by the front-end web server.
+  #
+  # By default, Panther always builds and tags the image as `latest` in order to avoid making Cloudformation changes
+  # on every deploy
+  WebApplicationImageVersion: latest
```
austinbyers: If we are always using `latest`, would this ever change? What would users use the parameter for?

deployments/template.yml
```diff
@@ -28,6 +28,11 @@ Parameters:
     Description: Toggle debug logging for all Lambda functions
     Default: false
     AllowedValues: [true, false]
+  IsInitialDeployment:
+    Type: String
+    Description: Whether the current stack is getting deployed for the first time
+    Default: false
+    AllowedValues: [true, false]
```
austinbyers: If possible, I'd definitely prefer to break things out into separate templates instead of deploying the same template twice with different flags

deployments/template.yml
```diff
@@ -143,19 +152,23 @@ Resources:
...
       TemplateURL: web/alb.yml
 
+  WebApplicationImageRepository:
+    Type: AWS::CloudFormation::Stack
+    Properties:
+      TemplateURL: web/ecr.yml
+
   FrontendApplicationServer:
     Type: AWS::CloudFormation::Stack
+    Condition: CreateFrontendApplicationServer
```
austinbyers: Rather than listing this nested stack here, the `mage deploy` command can just deploy that frontend stack directly after the main `app` stack is finished (if I'm understanding this correctly).

So I'm proposing this resource would be deleted, and instead `mage deploy` would directly deploy web/server.yml. That way we don't have the conditional in this stack, should be a little bit faster

tools/mage/deploy.go
nhakmiller: If the user can't be deleted through the UI then this is fine.

tools/mage/deploy.go
3nvi: You would have to do it everytime you make a change though.

Thus, if users forked our project and wanted to make a change, they would have to create a new version (based on some sort of semver style) and pass it as a runtime param on mage deploy. Alternatively, they would need to change a param in panther_config.

I'm ok doing that, it's just that it involves bumping a version everytime you perform a deployment. So if I'm working on a fix and I'm testing it, I'll have to spam new versions every time I make a deploy attempt

deployments/template.yml
```diff
@@ -143,19 +152,23 @@ Resources:
         CertificateArn: !Ref WebApplicationCertificateArn
       TemplateURL: web/alb.yml
 
+  WebApplicationImageRepository:
+    Type: AWS::CloudFormation::Stack
+    Properties:
+      TemplateURL: web/ecr.yml
```
austinbyers: This nested stack only contains a single resource, should we just include that resource here?

tools/mage/deploy.go
austinbyers: Does this mean the first time users need to run the `deploy` command twice?

tools/mage/deploy.go
```diff
@@ -302,3 +342,76 @@ func inviteFirstUser(awsSession *session.Session, userPoolID string) error {
...
+	}
+	if err := godotenv.Write(conventionalOutputs, filename); err != nil {
+		return err
+	}
+	return nil
+}
+
+// makes sure to force a new ECS deployment on the service server so that the latest docker image can be applied
+func restartFrontendServer(awsSession *session.Session, cluster string, service string) error {
+	fmt.Println("cleanup: Upgrading front-end server to the latest docker image...")
```
austinbyers: Nit: the convention we're using here is the first word is the name of the `mage` command. There is no `mage cleanup`, so this should still just be `deploy`

tools/mage/util.go
```diff
@@ -139,6 +145,14 @@ func download(url string) ([]byte, error) {
 	return ioutil.ReadAll(response.Body)
 }
 
+// A wrapper around the basic `exec.Command`, running it immediately & adding streaming output to STDOUT/STDERR
+func runCommand(name string, arg ...string) error {
```
austinbyers: `sh.Run` should do pretty much the same thing: https://godoc.org/github.com/magefile/mage/sh#Run

That's what we use everywhere else, was that not sufficient for you?

web/README.md
```diff
@@ -15,18 +15,44 @@ $ npm i
 
 ### Development
 
-In order to develop locally, all you need to do is run:
+In order to develop locally, you must first have the project deployed on your AWS account. If you've successfully done that,
+then a `.env` folder will have appeared in the project root which will hold crucial ENV vars that
+is needed for local development. If this file is present, then all you need to do is run:
```
austinbyers: The `npm` commands are designed from the repo root, right? If so, can we clarify that?

tools/mage/deploy.go
nhakmiller: I don't believe so, this is sort of running the deploy command twice for users automatically. Before this block `deployTemplate` is called passing in `applicationStack`, and then it is done so again within this block after setting `IsInitialDeployment` to `false`.

tools/mage/deploy.go
austinbyers: I guess I was picturing something a little smarter than that, e.g. `mage deploy` automatically creates a new version if the hash of the `web` folder changed, for example, and passes the new parameter in only if needed.

The ultimate goal would be that if there were only backend changes, the frontend service shouldn't need to be redeployed or restarted. But that might be outside the scope of this PR.  

I'm fine with restarting the task on every deploy if that's an easy way to guarantee frontend updates

3nvi: What you are proposing with regard to the versioning is harder than it looks since we are not the only people making changes. If it was a library then you could just bump versions on every release. Here you need to bump versions on every front-end change. Those versions will have conflicts from one dev to another since if 2 devs are at v1.0.0 and they make a change, the script will have to bump versions for both of them. This will end up in two **different v1.0.1**.

In addition, you need an algorithm to correctly calculate the new version from the previous one. This means that you either have to use semantic commits (if you want proper semver) or just use integers for versioning (1,2,3...).

With regards to being smart and not redeploy, then:

1. The build process needs to be initiated no matter what. If everything is the same the digest of the image will be the same as well. Docker uses caches to make sure that it doesn't rebuild what's not needed. Thus, if everything stays the same, the docker build will take < 5 seconds
2. The deploy process currently only deploys if there are new docker layers (i.e. changes). If everything stayed the same, it will just say that everything is already deployed.
3. The front-end ECS restart is dumb at the moment. You are right that we can check if we should do that or not, based on whether the image hex digest (after the build) is the same with the existing one (i.e. API call to compare it the image digest that ECS is already running). That though I feel is an optimization that can easily go along with changes that you guys have with regards to un-necessary CFN deployments. Currently, the restart takes 5 seconds to complete and will never cause any downtime, so it literally doesn't create any issue apart from an unnecessary API call.

tools/mage/deploy.go
3nvi: You are correct, it's done automatically. I could either split them up into two different stacks, but using a condition was way cleaner (since the changes involved with splitting CFN templates were **way more**)

README.md
3nvi: sure

deployments/panther_config.yml
3nvi: The user might want to explicitly build a version that's other than `latest`. If he passes a different value here, then the next `mage deploy` is going to build, deploy and use the version specified.

Currently that's an unusual case where the user would want to "save" and never override a certain UI state, to have it as a fallback

deployments/template.yml
3nvi: Tried to do that initially. It gets more complex that you'd think since you are:

1. Changing the parameters way params are being fed to the templates through `panther_config.yml`, as well as lots of golang structs
2. Moving and grouping the outputs from two separate stacks
2. The stack deployments & packaging
3. The reading, moving and grouping of the outputs, as well as the structs they return

In addition, the **only** thing that would go on the other template is a single `ECS Service` resource. I spoke with @kostaspap about it and decided that it not worth introducing multiple changes and making a more complex workflow just for a single resource that has to **not run only once**.  Lastly, the condition is a generic one that can easily be re-used, whose value is managed by  an API call to Cloudformation.

If you feel though that this approach is too dirty for our OSS, I can split them.

deployments/template.yml
3nvi: Well that essentially means splitting them (the same comment as the one you made above). This would involve forwarding any global parameters currently passed from `template.yml` -> `web/server.yml` as well, etc, etc..

My answer is the same: https://github.com/panther-labs/panther/pull/11#discussion_r369421491

deployments/template.yml
3nvi: It doesn't make a difference  I think, I just thought it would be better if it was isolated.

If you want I can include it here directly, but I'm failing to see any major upsides

tools/mage/deploy.go
3nvi: Gotcha. Will update :)

tools/mage/util.go
3nvi: Almost. It doesn't pipe the output to stdout or stderr. Instead it's "silent". I want to make sure that during the `docker build` or `docker upload`, I can get information about how much time will it takes until the image is uploaded. 

With `sh.Run` you don't have an easy way of forwarding the output like you have with `exec.Command`. If you want to, I can change the latter to use `sh.Run` and add some additional commands to make sure the output is piped.

web/README.md
3nvi: It's clarified.

The first lines of README already mention the following:

```
> Due to multiple package management, please consider the project root as the current working directory
> for the following document sections
```

tools/mage/deploy.go
austinbyers: Sounds good! Thanks for the clarification

> That though I feel is an optimization that can easily go along with changes that you guys have with regards to un-necessary CFN deployments.

Agreed, optimization for a later date

deployments/template.yml
austinbyers: Fair enough - at that point maybe even the prerequisite "buckets" stack could go into the same master stack.

My thought was we already have two stacks, adding more to the deploy process seemed reasonable - we can then more generically think of Panther as several stages of deployment.

The conditional works fine as long as we only need it as a special case - we can leave it for now

deployments/template.yml
austinbyers: No difference in terms of performance, it's just overkill IMO to have an entire file with just a single resource in it. No strong preference

tools/mage/util.go
austinbyers: `sh.Run` will print the output when running in verbose mode (`mage -v [command]`)

Perhaps one of these will work for you?

- `sh.RunV` - https://godoc.org/github.com/magefile/mage/sh#RunV - always prints output of the command (regardless of `-v`)
- `sh.Output` - https://godoc.org/github.com/magefile/mage/sh#Output - capture output from a command
- `sh.Exec` - https://godoc.org/github.com/magefile/mage/sh#Exec - manually control stdout and stderr

Try to use something from the `sh` library if possible to be consistent with the rest of the project, but if you can't it's not critical

README.md
austinbyers: Actually, #28 moves the getting started steps out of the documentation completely - I'll make the change there

tools/mage/deploy.go
3nvi: Ended up splitting the template into two stacks. This is considered resolved now.

tools/mage/deploy.go
3nvi: Reverted the changes with regards to the conditional statement, thus this issue won't occur

deployments/template.yml
3nvi: If you think it's an overkill, wait to see what I've done in `frontend.yml`

tools/mage/util.go
3nvi: Sure thing, will do

package.json
```diff
@@ -130,11 +131,10 @@
     "yup": "^0.26.6"
   },
   "scripts": {
-    "build-assets": "webpack",
-    "serve-assets": "node web/scripts/serve-assets.js",
-    "setup-dotenv": "node web/scripts/setup-dotenv.js",
+    "build": "cross-env NODE_ENV=production node -r dotenv/config node_modules/.bin/webpack",
```
kostaspap: Do we still need the "production" ENV here? We are really running only in one env, right?

.dockerignore
```diff
@@ -15,6 +15,8 @@ venv
 
 # Generated
 out
+keys
```
kostaspap: What's that for?  What keys are these ones?

.gitignore
```diff
@@ -12,11 +12,10 @@ npm-debug.log*
 # web
 web/dist/
 web/scripts/deploy-sourcemap.js
-# Dev configurations
-web/config/*
 
 # Private keys
 keys/
+.env
```
kostaspap: What is included in this `.env`?

.gitignore
```diff
@@ -12,11 +12,10 @@ npm-debug.log*
 # web
 web/dist/
 web/scripts/deploy-sourcemap.js
-# Dev configurations
-web/config/*
```
kostaspap: I see the files still exist. Should they be stored in git?

deployments/backend.yml
```diff
@@ -374,22 +364,55 @@ Resources:
...
+    Value: !GetAtt WebApplicationNetworking.Outputs.SubnetOneId
+  WebApplicationSubnetTwoId:
+    Description: The second subnet that the web app uses
+    Value: !GetAtt WebApplicationNetworking.Outputs.SubnetTwoId
+  AWSAccountId:
+    Description: The ID of the account that this stack was deployed to
+    Value: !Ref AWS::AccountId
+  AWSRegion:
+    Description: The region that the stack was deployed to
+    Value: !Ref AWS::Region
```
kostaspap: All of the stacks are deployed in the same region, in the same account. No need to export them.

deployments/panther_config.yml
```diff
@@ -22,8 +22,8 @@ BucketsParameterValues:
   AccessLogsBucketName: ''
 
 
-# CloudFormation parameter overrides for the template.yml main application stack.
-AppParameterValues:
```
kostaspap: Let's keep it one configuration file - Panther configuration file. No need to split into frontend and backend, users in the end see only one thing.

deployments/web/Dockerfile
```diff
@@ -37,18 +30,51 @@ RUN mkdir /code
...
 ADD web web
 
-# Build the project with the ENV vars present in the system & run the node server
-CMD npm run build-assets && npm run serve-assets
+# Build argument & ENV var for versioning
+ARG PANTHER_VERSION
+ENV PANTHER_VERSION=$PANTHER_VERSION
+
+# Convert NODE_ENV to production in order to adjust webpack configuration accordingly
+ENV NODE_ENV=production
```
kostaspap: Do we still need these variables? 
Or, if I understand them correctly, maybe they should be "server" and "dev" instead of "production"? (We don't really have"production" or "staging" or "beta" or ...

deployments/web/Dockerfile
```diff
@@ -37,18 +30,51 @@ RUN mkdir /code
...
+ENV SERVER_PORT=8080
+
+# Make sure we add a folder to not pollute our root
+RUN mkdir /code
+
+# And make it the default WD
+WORKDIR /code
+
+# Copy the artifacts from the previous stage
+COPY --from=build-env /code/web/dist dist
```
kostaspap: Nice! Didn't know about the Docker stages!

deployments/web/ecr.yml
```diff
@@ -0,0 +1,35 @@
...
+
+Resources:
+  PantherWebImageRepository:
+    Type: AWS::ECR::Repository
+    Properties:
+      RepositoryName: !Ref ImageName
+
+Outputs:
+  ImageRepo:
+    Description: The URL of the repository hosting the panther web image
```
kostaspap: This doesn't seem like a URL? (it is not starting by `https` or `http`)

deployments/web/server.yml
```diff
@@ -224,3 +207,8 @@ Resources:
       NetworkMode: awsvpc
       RequiresCompatibilities:
         - FARGATE
+
+Outputs:
+  ServiceName:
+    Description: The name of the service that the hosts the front-end server
+    Value: !GetAtt WebApplicationServer.Name
```
kostaspap: This is actually provided as parameter, it doesn't make sense to return it as output.

web/src/client.ts
```diff
@@ -56,15 +56,15 @@ const createErrorLink = (history: History<LocationErrorState>) => {
...
+const httpLink = createHttpLink({ uri: process.env.WEB_APPLICATION_GRAPHQL_API_ENDPOINT });
 
 /**
  * This link is here to add the necessary headers present for AMAZON_COGNITO_USER_POOLS
  * authentication. It essentially signs the Authorization header with a JWT token
  */
 const authLink = (createAuthLink({
   region: process.env.AWS_REGION,
-  url: process.env.GRAPHQL_ENDPOINT,
+  url: process.env.WEB_APPLICATION_GRAPHQL_API_ENDPOINT,
```
kostaspap: Isn't GRAPHQL_ENDPOINT better? We only have one anyway :P

.dockerignore
3nvi: @nhakmiller has added them. They are related to self-signed certificate. We definitely don't need them in the docker context

.gitignore
3nvi: This PR deletes them all. They won't exist after this PR is merged

.gitignore
3nvi: It's detailed in the description of the current PR.

It's essentially a list of ENV variables that the front-end uses. Practically it stores the CFN outputs. For more info google `dotenv` and/or check out the [associated repo](https://github.com/motdotla/dotenv)

deployments/backend.yml
3nvi: The front-end needs to know which region is the stack deployed to. When I type `npm run dev` to kick-off the dev environment for the front-end, the front-end needs to know the region that panther is deployed to.

All the outputs are stored in `.env`, thus it makes it really convenient to have the region information there. Realistically, the alternative would be to attempt to read it through the terminal's ENV variable (`AWS_REGION` or `AWS_DEFAULT_REGION`), but that's not guaranteed that it will match the region that panther got deployed too.

All in all, the front-end needs to know where is the backend deployed. We need to provide that through some sort of variable

deployments/panther_config.yml
3nvi: I'm ok with that as long as passing extra parameters to a stack won't cause an exception (since front-end & back-end will have different parameters, but we will always "feed" everything declared in `AppParameterValues`

deployments/web/Dockerfile
3nvi: `NODE_ENV` is not an arbitrary variable, but, by convention, denotes to our build system whether to create a development build or a production build. Production mode enables minification, truncation, perf optimizations, etc.

This is documented in our [build framework's docs](https://webpack.js.org/guides/production/#specify-the-mode). Taken directly from the docs:

> NODE_ENV is a system environment variable that Node.js exposes into running scripts. It is used by convention to determine dev-vs-prod behavior by server tools, build scripts, and client-side libraries. Contrary to expectations, process.env.NODE_ENV is not set to 'production' within the build script webpack.config.js, see #2537. Thus, conditionals like process.env.NODE_ENV === 'production' ? '[name].[hash].bundle.js' : '[name].bundle.js' within webpack configurations do not work as expected.

deployments/web/ecr.yml
3nvi: Hmm it's the URL but without the protocol. I'll update the description to something better

deployments/web/server.yml
3nvi: It's not provided as a parameter per say. I just used the default value of it cause it's repeated so much that I didn't want to hardcode it.

Thus, I'm using the default parameter value as a variable here

package.json
3nvi: again, `NODE_ENV` is unrelated to us, but has to do with the build tool. The values of `NODE_ENV` are by convention either `development` or `production`.

That's unrelated to the env that you are targeting and it simply helps the build tool understand whether it should optimize its process or not

web/src/client.ts
3nvi: I figured that I should expose everything related to the web app with the prefix `WebApplication` from the CFN stack. These variable names map 1:1 to the CFN outputs. That's why you see that.

To implement what you are saying I would have to rename the output in the related stack from `WebApplicationionGraphqlEndpoint` to `GraphqlEndpoint`. I don't mind doing it, but as mentioned , the prefix denotes the scope of the output :)

.gitignore
kostaspap: Sure. It seems though that  `.env` (and actually `keys`)  could just be in the `out` directory? I'm not a fun of creating a lot of temporary folders/files in main directory. 
`out` is reserved for storing any buid/deployment related generated artifact, so we could move `.env` there.

deployments/backend.yml
austinbyers: Outputting the region is definitely not helpful, because you already need to know the AWS region to read the stack outputs in the first place. Relying on the `AWS_REGION` environment variable is pretty standard and seems reasonable to me

deployments/panther_config.yml
austinbyers: I don't think you can pass parameters that the stack doesn't need - you can just make a new `map[string]string` with whatever parameters you need for the final stack. I'm happy to help with that if needed.

deployments/panther_config.yml
```diff
@@ -50,6 +50,9 @@ AppParameterValues:
   # XRay tracing mode for API Gateway and Lambda: '', 'Active', or 'PassThrough'
   TracingMode: ''
 
+# CloudFormation parameter overrides for the frontend.yml which deploys the frontend stack.
+FrontendParameterValues:
```
austinbyers: Since all these parameters are set by us during the deploy process, I'd prefer just leaving this section out completely

tools/mage/deploy.go
```diff
@@ -42,10 +45,12 @@ import (
 
 const (
 	// CloudFormation templates + stacks
-	applicationStack    = "panther-app"
-	applicationTemplate = "deployments/template.yml"
-	bucketStack         = "panther-buckets" // prereq stack with Panther S3 buckets
-	bucketTemplate      = "deployments/core/buckets.yml"
+	backendStack     = "panther-app"
```
austinbyers: Smart to keep the stack name the same so existing deployments aren't broken

tools/mage/deploy.go
```diff
@@ -87,53 +92,84 @@ func Deploy() error {
...
 		return err
 	}
 
-	if err := setupOrganization(awsSession, outputs["UserPoolId"], outputs["RemediationLambdaArn"]); err != nil {
+	if err = buildAndPushImageFromSource(awsSession, backendOutputs["WebApplicationImage"]); err != nil {
 		return err
 	}
 
-	if err := initializeAnalysisSets(awsSession, outputs["AnalysisApiEndpoint"], &config); err != nil {
+	frontendTemplate, err := cfnPackage(frontendTemplate, bucket, frontendStack)
```
austinbyers: You actually don't need to "package" this template - that's for uploading Lambda source, embedding Swagger specs, etc. You should just be able to `deployTemplate` directly, just like the `buckets` prereq stack on line 79

If you want to leave it for now, that's fine too - I'll fix it when I remove the AWS CLI as a requirement

tools/mage/deploy_frontend.go
```diff
@@ -0,0 +1,92 @@
...
+	ecrClient := ecr.New(awsSession)
+	req, resp := ecrClient.GetAuthorizationTokenRequest(&ecr.GetAuthorizationTokenInput{})
+	if err := req.Send(); err != nil {
+		return err
+	}
+
+	ecrAuthorizationToken := *resp.AuthorizationData[0].AuthorizationToken
+	ecrServer := *resp.AuthorizationData[0].ProxyEndpoint
+
+	decodedCredentialsInBytes, _ := base64.StdEncoding.DecodeString(ecrAuthorizationToken)
```
austinbyers: Never ignore errors! This should be

```go
decodedCredentialsInBytes, err := base64.StdEncoding.DecodeString(ecrAuthorizationToken)
if err != nil {
    return err
}
```

web/src/client.ts
austinbyers: > These variable names map 1:1 to the CFN outputs

Works for me!

deployments/backend.yml
3nvi: That's not correct as well. If I run `AWS_REGION=au-east-3 mage deploy` then the `AWS_REGION` var in my system doesn't match the region that the stack got deployed to.

What I need is the region of the stack, not the region that's setup  as the default one on my pc

deployments/panther_config.yml
3nvi: Ok I will remove it for now

tools/mage/deploy.go
3nvi: Oh I thought I needed that in order to upload master templates to S3. Guess I didn't read the code well

deployments/web/Dockerfile
kostaspap: Ok, my comment was that "production" has a different connotation, connotation that we have "multiple" ENVs like "beta", "gamma", etc. I though that different nouns would be more clear... but I'm ok going with it if this is what you are used to.

tools/mage/deploy_frontend.go
3nvi: ook

tools/mage/deploy.go
3nvi: I did indeed needed that for S3. As soon as I removed it, the stack failed with the error `TemplateURL must be an Amazon S3 URL`. 

I'm reverting its removal

deployments/backend.yml
austinbyers: > What I need is the region of the stack

Right, but you also need to know the region of the stack for you to get its outputs. As far as I know, you can't lookup the stack unless you know what region you're in.

When we describe stack outputs, we are implicitly using an AWS client that has already been built for a specific region according to the `AWS_REGION` environment variable

In other words, it's not possible to programmatically determine what region their stack is deployed in unless you scanned every AWS region. In fact, they could have multiple deployments in different regions! I do that a lot when testing

tools/mage/deploy.go
austinbyers: My bad - turns out nested stack templates are first uploaded to S3! I didn't realize that

PR #12, opened 2020-01-21
Handling CloudTral validation messages

kostaspap:
When a user sets up CloudTrail SNS notifications, CloudTrail sends a message to the SNS topic to verify it has permissions to deliver notifications.
We need to be able to handle this gracefully

* If event is a CloudTrail validation message, we are now handling it gracefully
* More appropriate error handling when parsing notifications - we now try all scenarios when trying to parse a notification before returning an error 

* Unit testing (uncovered the issue with parsing error handling)
* Deployed and verified it works.

### DIFF
--- a/internal/log_analysis/log_processor/sources/s3.go
+++ b/internal/log_analysis/log_processor/sources/s3.go
@@ -38,7 +38,8 @@ import (
 )
 
 const (
-	s3TestEvent = "s3:TestEvent"
+	s3TestEvent                 = "s3:TestEvent"
+	cloudTrailValidationMessage = "CloudTrail validation message."
 )
 
 // ReadSQSMessages reads incoming messages containing SNS notifications and returns a slice of DataStream items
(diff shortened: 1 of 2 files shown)


internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -186,38 +187,33 @@ func readS3Object(s3Object *S3ObjectInfo, topicArn string) (dataStream *common.D
 
 // ParseNotification parses a message received
 func ParseNotification(message string) ([]*S3ObjectInfo, error) {
-	s3Objects, err := parseCloudTrailNotification(message)
-	if err != nil {
-		return nil, err
-	}
+	s3Objects := parseCloudTrailNotification(message)
 
 	// If the input was not a CloudTrail notification, s3Objects will be empty slice
```
nhakmiller: `// If the input was not a CloudTrail notification, s3Objects will be **an** empty slice`

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -186,38 +187,33 @@ func readS3Object(s3Object *S3ObjectInfo, topicArn string) (dataStream *common.D
...
-		}
-		return nil, errors.New("notification is not of known type: " + message)
+	s3Objects = parseS3Event(message)
+	if len(s3Objects) > 0 {
+		return s3Objects, nil
 	}
-	return s3Objects, nil
+
+	if isTestS3Event(message) || isCloudTrailValidationMessage(message) {
+		//In this case return an empty array. There are no S3 objects to process
```
nhakmiller: Nit: this is a slice, not an array. Additionally, comments should have leading spaces unless they are meant to be machine readable (e.g. by a linter).

austinbyers: Even linter comments can have leading spaces as far as I know

PR #14, opened 2020-01-21
Fixing email delivery

kostaspap:
Fixing email delivery

* During setup we ask now the user what email should be used as the source email for delivering alerts (the Mail-From header). The prompt looks like the following: 
```
 aws-vault exec dev-kostas-admin -- mage deploy
Enter token for arn:aws:iam::123456789012:mfa/kostas_papageorgiou: ******
deploy: panther-buckets: no changes needed
build:lambda: go build internal/*/main (21 binaries)
deploy: cloudformation package deployments/template.yml => out/deployments/package.template.yml
deploy: s3://panther-source-123456789012-eu-west-1/layers/python-analysis.zip exists and is up to date
deploy: ACM certificate already exists
Enter email that will be used as alert source address. Empty string if you don't want to configure one: kostas@runpanther.io
Check [kostas@runpanther.io] inbox to verify your email
deploy: panther-app: ExecuteChangeSet: UPDATE_IN_PROGRESS
deploy: panther-app: ExecuteChangeSet: UPDATE_COMPLETE_CLEANUP_IN_PROGRESS
deploy: panther-app: ExecuteChangeSet: UPDATE_COMPLETE

Panther URL = https://web-1820136089.eu-west-1.elb.amazonaws.com
```
or for empty email
```
Enter token for arn:aws:iam::948584460855:mfa/kostas_papageorgiou: 212648
deploy: panther-buckets: no changes needed
build:lambda: go build internal/*/main (21 binaries)
deploy: cloudformation package deployments/template.yml => out/deployments/package.template.yml
deploy: s3://panther-source-415773754570-eu-west-1/layers/python-analysis.zip exists and is up to date
deploy: ACM certificate already exists
Enter email that will be used as alert source address. Empty string if you don't want to configure one: 
deploy: panther-app: CreateChangeSet: CREATE_IN_PROGRESS
deploy: panther-app: ExecuteChangeSet: UPDATE_IN_PROGRESS
...
```

* If user hasn't configured an email address that will be used for sending email alerts, adding an email alert will fail with an explanatory error. 

Screen Shot 2020-01-21 at 8 01 14 PM

* Deployed locally and verified above behavior
* Still need to make the code "pretty"

### DIFF
--- a/deployments/panther_config.yml
+++ b/deployments/panther_config.yml
@@ -47,6 +47,13 @@ AppParameterValues:
   # PROVIDE A CERTIFICATE FOR PRODUCTION USE
   WebApplicationCertificateArn: ''
 
+
+  # Email address that will be used as the source for sending email alerts
+  #
+  # If not specified, the user will be prompted for the address during the first setup
+  # Specify only email addresses that you have verified through SES
+  EmailAlertsFromAddress: ''
+
   # XRay tracing mode for API Gateway and Lambda: '', 'Active', or 'PassThrough'
   TracingMode: ''
 
--- a/tools/mage/deploy_email.go
+++ b/tools/mage/deploy_email.go
@@ -1,0 +1,87 @@
+package mage
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ses"
+)
+
+const (
+	emailAlertsFromAddressOutputsKey = "EmailAlertsFromAddress"
+)
+
+func getEmailAddress(awsSession *session.Session) (string, error) {
+	// Check if certificate has already been uploaded
+	emailAddress, err := getExistingEmailAddress(awsSession)
+	if err != nil {
+		return "", err
+	}
+
+	if emailAddress != "" {
+		return emailAddress, nil
+	}
+
+
+	sesClient := ses.New(awsSession)
+
+	emailInput := promptUser("Enter email that will be used as alert source address. Empty string if you don't want to configure one: ", optionalEmailValidator)
+	if emailInput == "" {
+		return "", nil
+	}
+
+	isAlreadyVerified, err  := isAddressAlreadyVerified(sesClient, emailInput)
+
+	if !isAlreadyVerified {
+		_, err  = sesClient.VerifyEmailIdentity(&ses.VerifyEmailIdentityInput{EmailAddress: aws.String(emailInput)})
+		if err != nil {
+			return "", err
+		}
+	}
+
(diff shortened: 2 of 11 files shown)


deployments/panther_config.yml
```diff
@@ -47,6 +47,13 @@ AppParameterValues:
   # PROVIDE A CERTIFICATE FOR PRODUCTION USE
   WebApplicationCertificateArn: ''
 
+
+  # Email address that will be used as the source for sending email alerts
+  #
+  # If not specified, the user will be prompted for the address during the first setup
+  # Specify only email addresses that you have verified through SES
```
austinbyers: There basically won't be any users who have verified an email through SES, will there?

tools/mage/deploy_email.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	if emailAddress != "" {
+		return emailAddress, nil
+	}
+
+
+	sesClient := ses.New(awsSession)
+
+	emailInput := promptUser("Enter email that will be used as alert source address. Empty string if you don't want to configure one: ", optionalEmailValidator)
```
austinbyers: To keep the deploy command as simple as possible, I'd rather not have optional prompts. If it's not required to deploy Panther, I don't think we should ask for it. The config file has more advanced configuration for stuff like this

The user already has to enter an email for the initial admin user - can we just use that as the default source address?

austinbyers: If we want to keep SES email delivery for now, I still would prefer we not prompt the user at deploy time. Instead, it can just read the config file, and, if there is an email listed, verify it just like you do here

kostaspap:
> This workflow seems clunky to me. Correct me if I'm wrong, but now a fresh deploy looks like this:
> 
> 1. Prompt user email (alert source)
> 2. User must confirm email
> 3. Deploy continues
> 4. Prompt first/last name and email (first user)
> 
> I as a user would be a little annoyed I had to enter my email twice, or two different emails at two different points in the script, and email sending would still be limited by SES restrictions.
> 
> Can we drop SES entirely and do something like the following?
> 
> 1. Create a default alerts SNS topic in CloudFormation
> 2. Subscribe the initial user email to that topic (SNS will send them a confirmation email)
> 3. Set the default alerts topic to trigger for HIGH/CRITICAL rule/compliance failures

Sorry it was not clear: The deploy command doesn't block while waiting for email verification. The deploy command continues and the user can verify the address (click a link in an email that will be sent in that address) async. 
SNS was always an option, this is a PR for SES.

deployments/panther_config.yml
kostaspap: If they do it manually, maybe. Otherwise they can go through the process in deployment. Isn't the comment clear?

tools/mage/deploy_email.go
kostaspap: That's one option sure. The prompt was for simplicity reasons - it will pop up only if they haven't configured anything.

kostaspap:
Dropping in favor of: https://github.com/panther-labs/panther/pull/24

PR #23, opened 2020-01-22
Creating organization

kostaspap:
Creating an organization on first deployment. This is needed by UI but also for remediation capabilities. 
Using as organization email the email of the first user and as organization name the name of the user. This change is needed for the frontend PR https://github.com/panther-labs/panther/pull/25 that fixes a lot of UI issues. 

* On first deploy, creating an organization. 
* Modified the Organization.Get method to check for required field `email` instead of previous check that was checking optional field. 

* Deleted my user from cognito & `panther-users` DDB table. Ran deploy script again - seen it populate the Organization. Verified with @3nvi  that he was able to use the new organization field.

### DIFF
--- a/internal/core/organization_api/table/get.go
+++ b/internal/core/organization_api/table/get.go
@@ -32,7 +32,7 @@ import (
 func (table *OrganizationsTable) Get() (*models.Organization, error) {
 	zap.L().Info("retrieving organization from dynamo")
 	response, err := table.client.GetItem(&dynamodb.GetItemInput{
-		Key:       DynamoItem{"id": {S: aws.String("1")}},
+		Key:       DynamoItem{"id": {S: aws.String(orgID)}},
 		TableName: table.Name,
 	})
 	if err != nil {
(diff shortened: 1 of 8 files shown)


internal/core/organization_api/table/get.go
```diff
@@ -44,7 +44,7 @@ func (table *OrganizationsTable) Get() (*models.Organization, error) {
 		return nil, &genericapi.InternalError{
 			Message: "failed to unmarshal dynamo item to an Organization: " + err.Error()}
 	}
-	if org.AwsConfig == nil {
+	if org.Email == nil {
```
rleighton: You should add a message to the error with with id that failed or this will be hard to debug. Same with all the errors above this one in the function.

Line 35 above looks wrong (it is not in patch and I cannot comment on it), it says:
```
Key:       DynamoItem{"id": {S: aws.String("1")}},
```
Why is 1 hard coded?

kostaspap: > Why is 1 hard coded?

We can only have 1 organization. We used to have more, but I believe this is left-overs. In that sense, returning the ID is not that helpful.

kostaspap: I'll add a comment to that extend.

nhakmiller: When we made the decision to move from multi tenant to single tenant I went through and ripped out all the multi tenant code. In the majority of cases it didn't change anything functionally, but in a few places (mostly around the organization api which had to deal with multi tenant issues the most) it lead to weird things like this where organizations are required for certain things to work but there will only ever be one organization. A more thorough refactor and redesign (we certainly no longer need dynamo for this) is due in the near future I think.

PR #25, opened 2020-01-22
[ENG-1334][ENG-1305][ENG-1318][ENG-1284][ENG-1277][ENG-1275][ENG-1274][ENG-1285] Pre-Deployment Bugfixes

3nvi:
This PR resolves all the core bugfixes found when performing QA for our product in the past few days. Since changes for each bugfix were not too many, I opted for an all-in-one approach

* Resolve all bugs from all tickets named above. [ENG-1285] will only work after necessary changes from @kostaspap to ensure that that the organization DB gets populated on the initial Panther deployment 

* Locally & on personal dev account

[ENG-1285]: https://panther-labs.atlassian.net/browse/ENG-1285


web/src/pages/create-source/subcomponents/create-infra-source/subcomponents/remediation-panel.tsx
```diff
@@ -17,41 +17,30 @@
...
-region=us-west-2#/stacks/create/review?templateURL=https://s3-us-west-2.amazonaws.com/\
-panther-public-cloudformation-templates/${PANTHER_REMEDIATION_SATELLITE_ACCOUNT}/latest/template.yml&\
-stackName=${PANTHER_REMEDIATION_SATELLITE_ACCOUNT}&param_MasterAccountId=${masterAWSAccountId}`;
-};
+import { Box, Button, Heading, Text } from 'pouncejs';
+import { PANTHER_REMEDIATION_SATELLITE_ACCOUNT } from 'Source/constants';
 
 const RemediationPanel: React.FC = () => {
-  const [isStackLaunched, markStackAsLaunched] = React.useState(false);
+  const cfnLink =
```
austinbyers: Interesting - so we assume that both the template and the CFN stack are in `us-west-2`? Presumably people need to be able to deploy this in any region, we'll need to add an issue to fix this in the future

kostaspap: This is the template URL- which is in an S3 bucket in us-west-2. It works in other regions as well, I've been running panther in eu-west-1 :)

PR #31, opened 2020-01-23
Fixing flaky integration tests

kostaspap:
Fixing flaky integration tests

* Changed some `assert` to `require` - thinking it was more appropriate. 
* Replacing `"1"` magic string with `orgID` in `UpdateOrganization` API. I believe this method is no longer used, but I will postpone removing it until we have a better plan for this. 
* Fixed `UpdateSns` & `UpdateSlack` output integration tests:
   * In some parts of the code we are performing read-after-write on the DDB table `panther-default-outputs` and we need the reads to be consistent. Modified the GetDefault* methods to perform consistent queries. We make these operations upon user actions, so there is no real concern for increased costs
   * Tests were failing because we had configured `UpdateSns` & `UpdateSlack` to run in parallel. This currently can cause the two operations to overlap and cause one of them to fail. 
  * It would be a better plan of action to remove the `panther-default-outputs` table. It is not offering much of a benefit right now and it can also cause issue when two outputs are updated simultaneously. Can do in separate PR. 

* Ran 30 times the integration tests after making the changes. Didn't seen them fail once.

### DIFF
--- a/internal/core/outputs_api/api/update_output.go
+++ b/internal/core/outputs_api/api/update_output.go
@@ -72,7 +72,7 @@ func (API) UpdateOutput(input *models.UpdateOutputInput) (*models.UpdateOutputOu
 	// Removing outputId from all defaults
 	for _, defaultOutput := range defaults {
 		var removed bool
-		defaultOutput.OutputIDs, removed = removeSlice(defaultOutput.OutputIDs, input.OutputID)
+		defaultOutput.OutputIDs, removed = removeFromSlice(defaultOutput.OutputIDs, input.OutputID)
 		if removed {
 			if err := defaultsTable.PutDefaults(defaultOutput); err != nil {
 				return nil, err
@@ -93,7 +93,7 @@ func (API) UpdateOutput(input *models.UpdateOutputInput) (*models.UpdateOutputOu
 
 // Removes an item from a slice if it exists. Returns the resulting slice
 // and a boolean indicating whether an item was removed or not
-func removeSlice(slice []*string, item *string) ([]*string, bool) {
+func removeFromSlice(slice []*string, item *string) ([]*string, bool) {
 	new := make([]*string, 0, len(slice))
 	for _, element := range slice {
 		if *element != *item {
--- a/internal/core/outputs_api/api/update_output.go
+++ b/internal/core/outputs_api/api/update_output.go
@@ -71,9 +71,12 @@ func (API) UpdateOutput(input *models.UpdateOutputInput) (*models.UpdateOutputOu
 
 	// Removing outputId from all defaults
 	for _, defaultOutput := range defaults {
-		defaultOutput.OutputIDs = removeFromSlice(defaultOutput.OutputIDs, input.OutputID)
-		if err := defaultsTable.PutDefaults(defaultOutput); err != nil {
-			return nil, err
+		var removed bool
+		defaultOutput.OutputIDs, removed = removeFromSlice(defaultOutput.OutputIDs, input.OutputID)
+		if removed {
+			if err := defaultsTable.PutDefaults(defaultOutput); err != nil {
+				return nil, err
+			}
 		}
 	}
 
@@ -88,11 +91,13 @@ func (API) UpdateOutput(input *models.UpdateOutputInput) (*models.UpdateOutputOu
 	return alertOutput, nil
 }
 
-func removeFromSlice(slice []*string, item *string) []*string {
+// Removes an item from a slice if it exists. Returns a boolean indicating whether an item
+// was removed or not
+func removeFromSlice(slice []*string, item *string) ([]*string, bool) {
 	for i, element := range slice {
 		if *element == *item {
-			return append(slice[:i], slice[i+1:]...)
+			return append(slice[:i], slice[i+1:]...), true
 		}
 	}
-	return slice
+	return slice, false
 }
--- a/internal/core/outputs_api/api/update_output.go
+++ b/internal/core/outputs_api/api/update_output.go
@@ -72,7 +72,7 @@ func (API) UpdateOutput(input *models.UpdateOutputInput) (*models.UpdateOutputOu
 	// Removing outputId from all defaults
 	for _, defaultOutput := range defaults {
 		var removed bool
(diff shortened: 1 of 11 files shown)


internal/core/outputs_api/main/integration_test.go
```diff
@@ -346,24 +345,23 @@ func updateSlack(t *testing.T) {
...
 	}
 	var output models.UpdateOutputOutput
-	assert.NoError(t, genericapi.Invoke(lambdaClient, outputsAPI, &input, &output))
-	assert.Equal(t, slackOutputID, output.OutputID)
-	assert.Equal(t, aws.String("alert-channel-new"), output.DisplayName)
-	assert.Equal(t, slack, output.OutputConfig.Slack)
-	assert.Equal(t, aws.String("slack"), output.OutputType)
-	assert.Nil(t, output.OutputConfig.Sns)
+	require.NoError(t, genericapi.Invoke(lambdaClient, outputsAPI, &input, &output))
+	require.Equal(t, slackOutputID, output.OutputID)
```
rleighton: Not that it matters much, but I prefer assert as a long as it makes sense so you can see all the errors at once ... require is best if the failure implies going fwd make no sense (require.NoError()).

internal/core/outputs_api/table/get_default.go
```diff
@@ -33,6 +34,7 @@ func (table *DefaultsTable) GetDefault(severity *string) (*models.DefaultOutputs
 			Key: DynamoItem{
 				"severity": {S: severity},
 			},
+			ConsistentRead: aws.Bool(true),
```
rleighton: Can you add a comment as to why this is necessary. In particular, state what bad thing happens if you do NOT configure consistent reads.

internal/core/outputs_api/table/query_defaults.go
```diff
@@ -29,7 +30,8 @@ import (
 // GetDefaults returns the default outputs for one organization
 func (table *DefaultsTable) GetDefaults() (defaults []*models.DefaultOutputsItem, err error) {
 	var scanInput = &dynamodb.ScanInput{
-		TableName: table.Name,
+		TableName:      table.Name,
+		ConsistentRead: aws.Bool(true),
```
rleighton: Also, comment here.

internal/core/outputs_api/api/update_output.go
```diff
@@ -88,11 +91,13 @@ func (API) UpdateOutput(input *models.UpdateOutputInput) (*models.UpdateOutputOu
...
+func removeFromSlice(slice []*string, item *string) ([]*string, bool) {
 	for i, element := range slice {
 		if *element == *item {
-			return append(slice[:i], slice[i+1:]...)
+			return append(slice[:i], slice[i+1:]...), true
 		}
 	}
-	return slice
+	return slice, false
 }
```
nhakmiller: I was originally skeptical about this function, but when I tested it in the go playground it appeared to work as intended. After further digging into it (and then some discussion with Austin), I would prefer if this function were to be re-written to something like this:

```
func removeSlice(slice []*string, item string) ([]*string, bool) {
	new := make([]*string, 0, len(slice))
	for _, element := range slice {
		if *element != *item {
			new = append(new, element)
		}
	}
	return new, len(new) < len(slice)
}
```

The function you have here will actually mutate `slice` when it finds `item`, see this go playground for an example:

https://play.golang.org/p/oK3L_d_Ammz

Currently, you overwrite the input with the output so this doesn't matter. But if someone were to use this `removeFromSlice` function and not assign the output of the function back to the input, the input slice were be mutated in an (in my opinion)entirely unexpected way.

internal/core/outputs_api/main/integration_test.go
austinbyers: Agreed - `require` means don't check anything else in this test, fail now, you can't possibly check any other test output. `assert` is sort of the "default" check

internal/core/outputs_api/api/update_output.go
austinbyers: Right - the `removeFromSlice` function, as you have written now, is technically more efficient, but the unexpected behavior (the original slice no longer shows the same values) could lead to nasty bugs.

It also only removes the *first* element, the solution that nick and I worked on above is more generic and does not modify the original slice at negligible performance overhead.

Please also make sure this function is unit-tested if it isn't already

internal/core/outputs_api/main/integration_test.go
kostaspap: So, well if any of them fails, we have a problem - it means the update didn't occur. I know that if one of them fails all the other will fails as well, so it doesn't make sense to continue with the rest of the assertions. That's my thinking behind picking `require` here. 

I'd honestly prefer that we just pick one of the two and just roll with it (assert or require). I think having to debate or making a developer even think which one to pick is kinda counter-productive and honestly I think we get wrong at many of our test cases :P

internal/core/outputs_api/api/update_output.go
kostaspap: I actually wish Golang was a functional language, immutability is one of the things I am missing from my FP days...
But I get your point about making the code more clear. Looking into many of the Golang's libraries, I think it would have been more clear if the function had a signature like below: 
```
func removeFromSlice(slice []*string, item string) {
}
```
?

internal/core/outputs_api/table/get_default.go
kostaspap: Done

internal/core/outputs_api/table/query_defaults.go
kostaspap: Done

internal/core/outputs_api/api/update_output.go
kostaspap: I've added unit tests to verify that we removed/add defaults properly. 

I actually don't really like the code right now, I think it is pretty hard to understand. 
It could be greatly simplified if we would drop the `panther-default-outputs` and just keep all information in the `panther-outputs`. I am just concerned about existing customers - they would require to run a script to migrate the information in `default-outputs` to `panther-outputs`, that's why refraining from it.

PR #32, opened 2020-01-23
ENG-1321 Alert page is not sorted. 

rleighton:
## Background
Alerts page was not ordered by creationTime desc.

## Changes
*  Added new GSI to enable sorted queries by time (only)
* Consolidated alert listing code from 2 functions to 1 ListAlerts() function
  * orders alerts by creationTime desc
* Refactored logging and errors
  * logging (with some exceptions) only happens at major entry points (e.g., lambda handers)
    * oplog is used to get stats
  * errors are Wrap()'d to get context

## Testing
* mage test:unit
* generated 1000's of alerts over night and tested pagination


deployments/log_analysis/alerts.yml
```diff
@@ -116,6 +118,15 @@ Resources:
           IndexName: ruleId-creationTime-index
           Projection:
             ProjectionType: ALL
+        - # Add an index using dummy timePartition to efficiently list alerts by creationTime
```
austinbyers: Having a dummy HASH key is problematic for very large numbers of entries because it forces Dynamo to store everything in a single partition - you won't get the same performance as you would if things were split across partitions.

It's fine for now but definitely something to keep in mind. A future more sophisticated index might first partition by the day or hour, like we do for Athena

internal/log_analysis/alert_merger/merger/processor.go
```diff
@@ -65,32 +67,23 @@ const eventMergingPeriodSeconds = 3600 // One hour
...
-		return err
+		return errors.Wrap(err, "failed to get alert info")
 	}
 
-	zap.L().Info("successfully got alert id",
-		zap.String("alertId", *alertID))
-
 	if err = addEventToAlert(event, notification, alertID, creationTime); err != nil {
-		return err
+		return errors.Wrap(err, "failed to add event to alert")
```
austinbyers: :+1:

internal/log_analysis/alerts_api/main/lambda.go
```diff
@@ -33,7 +33,12 @@ var router = genericapi.NewRouter(nil, api.API{})
 
 func lambdaHandler(ctx context.Context, input *models.LambdaInput) (interface{}, error) {
 	lambdalogger.ConfigureGlobal(ctx, nil)
-	return router.Handle(input)
+	event, err := router.Handle(input)
+	if err != nil {
+		// wrap for api, InternalError the only kind of error from this lambda
```
austinbyers: Why is that? Couldn't there be `NotFound` error if you access an alert that doesn't exist?

deployments/log_analysis/alerts.yml
rleighton: Indeed. There a few issues with the ddb table design besides this one that Kostas plans to address in the rule re-design.

internal/log_analysis/alerts_api/main/lambda.go
rleighton: Good point. Original code was not  doing that . Do we want a error when you can't find  something you are looking for or just an empty result (changing that could break the UI)?

austinbyers: We should keep whatever existing behavior it had - I believe it returned a specific error when something is not found. This is an older API, when we have time to move it behind api gateway, it can return proper http status codes

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -28,24 +28,31 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/panther-labs/panther/api/lambda/alerts/models"
-	"github.com/panther-labs/panther/pkg/genericapi"
 )
 
-// ListAlertsByRule returns (a page of alerts, last evaluated key, any error)
-func (table *AlertsTable) ListAlertsByRule(ruleID *string, exclusiveStartKey *string, pageSize *int) (
+// List returns (a page of alerts ordered by creationTime, last evaluated key, any error)
+func (table *AlertsTable) List(ddbKey, ddbValue string, exclusiveStartKey *string, pageSize *int) (
```
kostaspap: The purpose of AlertsTable is to abstract the DDB layer from the code layer above, makes a nicer separation of concerns (e.g. with the alerts refactoring that we plan to do, I don't have to change other pieces of the code, just this implementation). I would remove any DDB internals from the function parameters, like the Index like ddbKey or ddbValue and have specialized method names. 
For example, we can have a 
`ListAlertsByRule(ruleID, exclusiveStartKey/cursor, pageSize)` and a
`ListAlerts(cursor, exclusiveStartKey/cursor,)`. You can rename the current `List` method to `internalListAlerts` with the current arguments and have each public method call it with appropriate arguments.

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -72,46 +76,30 @@ func (table *AlertsTable) ListAlertsByRule(ruleID *string, exclusiveStartKey *st
...
 
 	// If DDB returned a LastEvaluatedKey (the "primary key of the item where the operation stopped"),
-	// it means there are more alerts to be returned. Return populated `lastEvaluatedKey` in the response.
+	// it means there are more alerts to be returned. Return populated `lastEvaluatedKey` JSON blob in the response.
 	if len(queryOutput.LastEvaluatedKey) > 0 {
-		paginationKey := listAlertsPaginationKey{
-			AlertID: queryOutput.LastEvaluatedKey["alertId"].S,
-		}
-		marshalledKey, err := jsoniter.MarshalToString(paginationKey)
+		jsonSnippet, err := jsoniter.MarshalToString(queryOutput.LastEvaluatedKey)
```
kostaspap: Can you rename `jsonSnippet` to sth else? It is the key right? Everything marshaled with jsoniter is a JSON snippet :P

internal/log_analysis/log_processor/common/oplog.go
```diff
@@ -41,6 +41,7 @@ var (
 	OpLogSNSServiceDim       = zap.String(OpLogServiceDim, "sns")
 	OpLogProcessorServiceDim = zap.String(OpLogServiceDim, "processor")
 	OpLogGlueServiceDim      = zap.String(OpLogServiceDim, "glue")
+	OpLogAlertsAPIDim        = zap.String(OpLogServiceDim, "alertsApi")
```
kostaspap: Alerts shouldn't have any dependency to the log processor package. Please move this to alertsapi package.

internal/log_analysis/alert_merger/main/lambda.go
```diff
@@ -40,20 +42,41 @@ func main() {
...
+			errorCount++
+			logger.Error("invalid message received", zap.Error(err), zap.Any("input", input))
+			continue // skip bad data, nothing to be done
 		}
 
+		// this is where real work is done, for safety, fail lambda on any error
 		if err := merger.Handle(input); err != nil {
-			logger.Warn("encountered issue while processing event", zap.Error(err))
+			errorCount++
+			logger.Error("encountered issue while processing event", zap.Error(err), zap.Error(err), zap.Any("input", input))
```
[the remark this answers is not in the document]
rleighton: Fixed

internal/log_analysis/alert_merger/merger/processor.go
```diff
@@ -106,6 +99,7 @@ func storeMatchedEvent(notification *AlertNotification) (*MatchedEvent, error) {
 
 	marshalledItem, err := dynamodbattribute.MarshalMap(event)
 	if err != nil {
+		err = errors.WithStack(err)
```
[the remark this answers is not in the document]
rleighton: Good question. I did that because the caller for all  functions called by Handle() are "wrapping" with a message. Adding a message to these lower level calls did not seem to add much value but the stack will pin point the error in the logs to make tracing easy.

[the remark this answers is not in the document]
kostaspap: Sounds good. I agree, we can be flexible with that, stacktrace already gives a lot of context.

api/lambda/alerts/models/table.go
```diff
@@ -20,6 +20,13 @@ package models
 
 import "time"
 
+const (
```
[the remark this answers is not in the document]
rleighton: Done. This does tidy up the code.

That said, I found the organization  a bit confusing ... can't we get "alerts" from the compliance size and won't they go thru this code under "log processing"?

internal/log_analysis/alerts_api/table/list.go
rleighton: Done

internal/log_analysis/alerts_api/table/list.go
rleighton: Not sure what you mean by "It is the key right?"

rleighton: Rename to what? "asJson"? This generalizes the code ... anything that is in query.LastEvaluatedKey is serialized to json and returned ... then at top of function we just unmarshal .. no need to worry about the contents.

kostaspap: Sorry, I meant it will be better renamed to "serializedKey" or "paginationKey" or similar. Naming this `jsonSnippet` is the equivalent of naming an int value "integer". 
```
integer := 0
```
... Not that useful so usually we name variables after their content, not their type. 

It's less of a problem here since it is used just 4 lines after its declaration, just was not sure why it had to be renamed.

api/lambda/alerts/models/table.go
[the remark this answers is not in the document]
kostaspap: Alerts is a concept only for log processing currently... Compliance doesn't have this concept. When a resources is failing a specific policy, the failing resource will be visible in the Policy/Failing Resources section.

internal/log_analysis/log_processor/common/oplog.go
rleighton: I'm going to remove this and put it back in a later PR when update the logging and metrics.

internal/log_analysis/alerts_api/table/list.go
rleighton: I changed it to lastEvaluatedKeySerialized.

api/lambda/alerts/models/table.go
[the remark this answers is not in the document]
rleighton: Hmmm, which means we have no history of compliance failures, correct?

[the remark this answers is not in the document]
kostaspap: Correct... For what's worth, it is listed here: https://panther-labs.atlassian.net/wiki/spaces/EN/pages/97943585/Active+Product+Development 
"Policy and Resource History" as a P3 item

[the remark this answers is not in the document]
rleighton: Something for later ... perhaps unify these concepts add history.

internal/log_analysis/alerts_api/table/get_event.go
```diff
@@ -21,20 +21,22 @@ package table
 import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/pkg/errors"
 )
 
 // GetEvent retrieves an event from DDB
+// FIXME: the string keys should be constants in _some_ model, but which?
```
kostaspap: Is it still valid? It seems you already made the fix (making them constants)

rleighton: Yup, fixed by moving the constants ... will remove.

internal/log_analysis/alert_merger/merger/processor.go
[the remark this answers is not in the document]
austinbyers: You get the stack trace either way, I would just standardize on `errors.Wrap` - https://github.com/pkg/errors/blob/master/errors.go#L181-L196

```
// Wrap returns an error annotating err with a stack trace
// at the point Wrap is called, and the supplied message.
// If err is nil, Wrap returns nil.
func Wrap(err error, message string) error {
	if err == nil {
		return nil
	}
	err = &withMessage{
		cause: err,
		msg:   message,
	}
	return &withStack{
		err,
		callers(),
	}
}
```

PR #36, opened 2020-01-24
[ENG-1353] Add versioning to front-end images

3nvi:
This PR adds versioning to the docker images built from source in order to allow image rollbacks & automatic service re-deployments. The presence of versioning removes the need for a manual ECS restart, since the restart of the service will be handled by AWS itself.  Lastly, because of versioning, the ECR repo can get quite big (if new versions of the front-end are constantly being pushed). To combat this - and because we only use the latest image at any given time -, a lifecycle policy has been added that will automatically retain only the 10 latest images, preventing the cost increase associated with storing multiple images on ECR.  

P.S. The frontend stack will take more time in `UPDATE_IN_PROGRESS` since it handles ECS restart during that. When you receive `UPDATE_COMPLETE` the ECS is **already updated** to the latest image.

* Add versioning to docker images
* Remove code that handled ECS restart
* Add Lifecycle policy to ECR

* AWS Development account


austinbyers:
Merging this now since I'm doing a lot of work on mage and the deploy phase, but I was able to deploy this commit from scratch and get to the login page

PR #37, opened 2020-01-24
Fix warning while uploading ECR image

kostaspap:
While running `mage deploy` I got the following warning: 
```
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
```

According to docker [docs](https://docs.docker.com/engine/reference/commandline/login/#provide-a-password-using-stdin): 
```
To run the docker login command non-interactively, you can set the --password-stdin flag to provide a password through STDIN. Using STDIN prevents the password from ending up in the shell’s history, or log-files.

The following example reads a password from a file, and passes it to the docker login command using STDIN:
```

* Switched docker login to use ``--password-stdin` command
* Using a pipe to pipe the password to the docker login command

* Ran `mage deploy` and seen it succeed

### DIFF
--- a/tools/mage/deploy_frontend.go
+++ b/tools/mage/deploy_frontend.go
@@ -51,26 +51,7 @@ func buildAndPushImageFromSource(awsSession *session.Session, imageTag string) e
 	}
 	credentials := strings.Split(string(decodedCredentialsInBytes), ":")
 
-	// Piping the Docker password to stdin
-	// before calling docker login command.
-	_, pipeWriter, err := os.Pipe()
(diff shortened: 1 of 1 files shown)


tools/mage/deploy_frontend.go
```diff
@@ -77,6 +73,37 @@ func buildAndPushImageFromSource(awsSession *session.Session, imageTag string) e
 	return nil
 }
 
+func dockerLogin(ecrServer string, dockerCredentials []string) error {
```
rleighton: Nice way to handle this

PR #38, opened 2020-01-24
Fixing auto remediations

kostaspap:
* AWS Remediations are not currently working. This PR fixes it. 
* The AWS Lambda responsible for remediating resources used to be deployed in a separate account from Panther app. Now it is deployed as part of standard deployment, which means there is no reason for it to be configurable. 

* Removing Remediation configuration-related entries from backend and GraphQL schema. 
* We no longer need `panther-aws-remediations-master-account.yml` and `panther-aws-remediations-satellite-account.yml`. Master account is now where Panther App is deployed. We only have "satellite" accounts. 
* Removed SAR app from `panther-aws-remediations-master-account.yml` and putting the permissions directly in there. 
* Backend no longer needs to assume a role to invoke the remediation Lambda - it can just invoke it directly
* We now deploy only the `src` folder of the `remediation_aws` Lambda. We used to deploy also the tests in the Lambda
* Did some package re-org in the `remediation_aws` package. 
* After moving to mono-repo the Python Lambda was not able to list all remediations. It seems that some code was left out during that migration (https://github.com/panther-labs/aws-remediations/blob/master/src/app/__init__.py#L54-L60). Re-added.

* Deployed to my account, verified that I was able to retrieve remediations. 
Screen Shot 2020-01-27 at 1 32 25 PM


deployments/auxiliary/cloudformation/panther-remediations-iam.yml
```diff
@@ -16,54 +16,58 @@
...
       MaxSessionDuration: 3600
       AssumeRolePolicyDocument:
         Version: 2012-10-17
         Statement:
-          -
-            Effect: Allow
+          - Effect: Allow
             Principal:
-              AWS: !Sub arn:aws:iam::${MasterAccountId}:root
+              AWS: !Ref MasterAccountId
```
nhakmiller: CloudFormation automatically transforms `AWS: 123456789012` to `AWS: arn:aws:iam::123456789012:root` before deploying the template, but doesn't update the template. That means any stack you deploy that uses the format `AWS: 123456789012` will immediately be considered out of sync if you run stack drift detection. That is why we use the long form, I'd recommend keeping it if possible.

deployments/auxiliary/cloudformation/panther-remediations-iam.yml
```diff
@@ -16,54 +16,58 @@
...
+                  - iam:UpdateAccountPasswordPolicy
+                  - kms:EnableKeyRotation
+                  - logs:CreateLogDelivery
+                  - rds:ModifyDBInstance
+                  - rds:ModifyDBSnapshotAttribute
+                  - s3:PutBucketVersioning
+                  - s3:PutBucketLogging
+                  - s3:PutBucketAcl
+                  - s3:PutEncryptionConfiguration
+                  - s3:PutBucketPublicAccessBlock
```
nhakmiller: Just a thought on these actions, perhaps we could break them up into different statements on a per service basis, with different SID's indicating the service? That would make it much easier for users do disable certain accesses after the fact if they decide to do so (e.g. they're ok with auto remediations for cloudtrail and iam but don't want anyone touching their EC2 instances and s3 buckets).

internal/compliance/remediation_api/remediation/invoke.go
```diff
@@ -81,28 +63,24 @@ var (
...
 	}
 
 	if policy.AutoRemediationID == "" {
-		zap.L().Info("There is no remediation configured for this policy",
+		zap.L().Debug("There is no remediation configured for this policy",
 			zap.Any("policyId", remediation.PolicyID))
 		return nil
 	}
 
 	resource, err := getResource(string(remediation.ResourceID))
```
nhakmiller: What happens if the resource doesn't exist anymore? Does this function return an error or an empty struct?

deployments/auxiliary/cloudformation/panther-remediations-iam.yml
kostaspap: Sure, I like that. Updated.

deployments/auxiliary/cloudformation/panther-remediations-iam.yml
kostaspap: Fixed.

internal/compliance/remediation_api/remediation/invoke.go
kostaspap: It returns a 404 - this means the remediation will fail with an exception.

nhakmiller: That seems reasonable, no way to know if there was an error in creating the request or the resource was simply deleted in between the initiation of the auto-remediation and this step.

internal/compliance/remediation_aws/src/app/__init__.py
```diff
@@ -13,51 +13,3 @@
 #
```
nhakmiller: Is this one of those files that just needs to exist for other stuff to work properly? Seems weird to have an empty file with just a license in it, but with a name like `__init__.py` I can only assume it's a special magic file.

kostaspap: It is indeed a magic file. The __init__.py files are required to make Python treat the directories as containing packages.

deployments/compliance/remediation.yml
```diff
@@ -75,14 +75,10 @@ Parameters:
...
+  RemediationRoleName:
     Type: String
-    Description: The name of the remediation IAM role to assume in each account
-    Default: PantherLambdaInvocationRole
-  CrossAccountSessionDuration:
-    Type: Number
-    Description: The duration (in seconds) of the cross-account session
-    Default: 3600
+    Description: The name of the IAM role to assume to perform remediation actions
+    Default: AwsRemediationRole
```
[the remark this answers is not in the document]
kostaspap: Good idea. I will update this in a separate PR (we also need to update the template in the S3 bucket)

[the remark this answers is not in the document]
kostaspap: @nhakmiller  check this https://github.com/panther-labs/cfn-templates-public/pull/58 and https://github.com/panther-labs/panther/pull/54

PR #42, opened 2020-01-24
Add .gitbook yaml

drixta:
Adding .gitbook yaml will set the root folder for gitbook to be under `docs`. This will allow gitbook to compile it's content into Markdown files and sync the gitbook documentation with our github repo.

* Add .gitbook.yaml

### DIFF
--- a/.gitbook.yaml
+++ b/.gitbook.yaml
@@ -1,0 +1,1 @@
+root: ./docs/
--- a/.gitbook.yaml
+++ b/.gitbook.yaml
@@ -1,1 +1,1 @@
-root: ./docs/
+root: ./docs/gitbook


.gitbook.yaml
```diff
@@ -0,0 +1 @@
+root: ./docs/
```
austinbyers: Can we make this `./docs/gitbook/`? We already have stuff in the docs folder (README images, PR templates, license headers, etc)

drixta: sounds good

jacknagz: or `./docs/src`

.gitbook.yaml
```diff
@@ -0,0 +1 @@
+root: ./docs/gitbook
```
austinbyers: You had the trailing slash before, is it needed or no? Hopefully it doesn't matter

.gitbook.yaml
drixta: moved to `gitbook`

PR #44, opened 2020-01-24
credential report timeout handling

nhakmiller:
When doing AWS IAM User scanning, we need a credential report. Building credential reports can sometimes take a long time, but generally not too long, so we just wait for it to complete then continue the scan.

Sometimes, we get rate limited while waiting for it to complete (we're doing a polling wait with exponential backoff). In the past we've increased the wait time and interval which helped this issue, but it happened again on my dev account the other day while doing a fresh installation.

This is a more permanent fix than raising the limits, as we catch the rate limit exception explicitly and re-queue a scan for the account in 90 seconds. If we get rate limited again, we'll just re-queue the scan for another 90 seconds later. We follow a similar-ish model for waiting for CloudFormation stack drift detections to complete.

Ideally we can add some fuzzy/randomization to the re-queue logic at some point, although we sort of get that for free as we are doing exponential/fuzzy back off each time it runs while waiting on the credential report to finish.

* Added re-queue logic to the IAM User snapshot poller around rate limiting
* Added some correctness checking when performing scans based on resource type

* Testing on this is difficult, as credential reports can only be generated once every 4 hours (AWS limitation). I will attempt some further testing on Monday, perhaps trying to write an example program that imitates the poll logic and catch the error there but realistically there is not an easy way to test against a live system

### DIFF
--- a/internal/compliance/snapshot_poller/pollers/aws/iam_user.go
+++ b/internal/compliance/snapshot_poller/pollers/aws/iam_user.go
@@ -43,12 +43,14 @@ import (
 )
 
 const (
-	rootAccountNameCredReport = "<root_account>"
-	rootDeviceSerialSuffix    = ":mfa/root-account-mfa-device"
+	// Time to delay the requeue of a scan of IAM Users when the credential report times out
+	credentialReportRequeueDelaySeconds = 90
+	maxCredReportBackoff                = 1 * time.Minute
+	rootAccountNameCredReport           = "<root_account>"
+	rootDeviceSerialSuffix              = ":mfa/root-account-mfa-device"
 )
 
 var (
-	maxCredReportBackoff  = 1 * time.Minute
 	userCredentialReports map[string]*awsmodels.IAMCredentialReport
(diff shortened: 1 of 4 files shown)


internal/compliance/snapshot_poller/pollers/aws/iam_user.go
```diff
@@ -74,6 +76,15 @@ func PollIAMUser(
 	}
 	userCredentialReports, err = buildCredentialReport(client)
 	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok {
+			// Check if we got rate limited, happens sometimes when the credential report takes a long time to generate
+			if awsErr.Code() == iam.ErrCodeLimitExceededException {
+				utils.Requeue(pollermodels.ScanMsg{
+					Entries: []*pollermodels.ScanEntry{scanRequest},
+				}, credentialReportRequeueDelaySeconds)
+			}
```
kostaspap: nit: I would add some debug log message here.

nhakmiller: Sounds good, added a debug message and also below in the similar logic for polling all users.

PR #47, opened 2020-01-25
[ENG-1352] Add Panther Deployment Image 

3nvi:
This PR adds a docker image that includes everything the user needs in order to install or deploy panther. To facilitate that even more, a `init` executable has been added that will instantly initialize this environment for you (I want your opinion on that).

With this docker image, the **only pre-requisite** for a panther installation is having Docker 17+ installed. The following examples showcase this functionality:

###

Bob wants to install Panther for the first time on a fresh PC without anything installed. He:

1. Clones the Panther repo
2. Downloads Docker 17+
3. Runs `./build/deployment/init` (or if he uses aws-vault, `aws-vault exec bob -- ./build/deployment/init`)
4. Runs `mage setup`
5. Runs `mage deploy`

Bob made a change to the source code and wants to deploy his changes. He:

1. Runs `./build/deployment/init`
2. Runs `mage deploy`

Ideally, I would want to place the `init` executable on the root dir since that would make things even easier (i.e. `aws-vault exec bob -- ./init`)

###

`init` will:

* Mount your local directory as a volume to the docker container and allow cross-side reads & writes.
* Install Go, NPM, Python, AWS CLI, Mage and configure all the paths automatically for you.
* Communicate with the docker installation in your host machine to use the same daemon for the image building.
* Forward all the `AWS_` related ENV vars (regardless of whether they exist locally or are temporary through AWS VAULT).

* Add `init.go` & related executable
* Add `Dockerfile` for the deployment pack
* Build Image and push it to dockerhub as `pantherlabs/panther-deployment-pack`

* Locally & on personal dev account.


build/deployment/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    apt-get -y install apt-transport-https \
+         ca-certificates \
+         curl \
+         gnupg2 \
+         software-properties-common && \
+    curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg > /tmp/dkey; apt-key add /tmp/dkey && \
+    add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") $(lsb_release -cs) stable" && \
+    apt-get update && \
+    apt-get -y install docker-ce && \
+    pip3 install awscli --upgrade --user
```
austinbyers: Where does mage get installed?

build/deployment/init.go
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import (
+	"log"
+	"os"
+
+	"github.com/magefile/mage/sh"
```
austinbyers: Since this is a standalone binary, I'd prefer not to import `mage` here. See if you can use the [exec](https://golang.org/pkg/os/exec/) package directly (unless we use Python, see above)

build/deployment/init.go
```diff
@@ -0,0 +1,57 @@
+package main
```
austinbyers: It's generally bad practice to check compiled binaries into the repo. It pollutes the history with large frequent changes and the binary will only work on the platform it was compiled for.

I realize that if we just had this Go source, someone needs to install Go to run it, which is what we're trying to avoid. I see 2 possible alternatives:

- We can include the compiled `init` binaries in the release artifacts instead of in the repo. Then users can just download the binary and put it in the repo root or their path or wherever.
- We can write this script in Python instead of Go and we assume that installing Python is easier/more common than installing Go.

I'm leaning towards the second option: let's make this script Python (compatible with Python 2 + 3) and then most Mac and Linux systems will already have Python installed and they can just run `python image.py` or whatever we call it

docs/gitbook/quick-start.md
```diff
@@ -32,7 +32,38 @@ Precise deployment policy coming soon!
...
-### Steps
+### Getting Started
+
+In order to install Panther on your AWS account, you can choose between using our dedicated docker deployment image or manually
+installing the necessary prerequisites yourself.
+
+#### Installation & Deployment using our Docker Image
+
+1. Install [Docker](https://docs.docker.com/install/) 17+ and make sure the daemon is running (i.e. typing `docker info` doesn't cause an error)
+   - For MacOS w/ homebrew: `brew install docker`
```
austinbyers: This will *not* work (I tried) - it only installs the docker command, not the daemon. Direct users to download the app from the Docker website instead

docs/gitbook/quick-start.md
```diff
@@ -32,7 +32,38 @@ Precise deployment policy coming soon!
...
+- SSH: `git clone git@github.com:panther-labs/panther && cd panther`
+
+3. Spin-up the deployment environment by running `./build/deployment/init` [related source code](https://github.com/panther-labs/panther/tree/master/build/deployment/init.go)
+
+- If you are using `aws-vault` or similar AWS credential managers, make sure you run the executable after creating the temporary credentials
+  like so: `aws-vault exec <ROLE> -- ./build/deployment/init`.
+- We recommend that your temporary credentials stay valid for a minimum of
+  30 mins, since the initial deployment will take some time.
+
+4. From the root of the repo, run `mage setup`
```
austinbyers: Steps 4-7 are copied verbatim. Can we simplify to reduce duplication? Something like

```markdown
## Quick Start

### Development Environment
First, you need a development environment.
Either install our development docker image or run a few steps to install libraries directly...

### Deployment
Mage setup and mage deploy 
```

build/deployment/Dockerfile
3nvi: mage is pre-installed in the `pantherlabs/panther-buildpack:1.0.1`

build/deployment/init.go
3nvi: Again it poses the same issues that a go file would. What if the user is on a windows machine? 99% he's gonna have python installed, but what if he doesn't?

Do you think opting for a bash script is preferable?

build/deployment/init.go
3nvi: I can easily use `os.Exec`. Based on the answer on the comment above, I'll make sure to update

docs/gitbook/quick-start.md
3nvi: Hmmm, c/p it from the existing docs. Guess we need to update them as well :/

Will try and see if something better can be done

docs/gitbook/quick-start.md
3nvi: Yeah that looks better.

3nvi:
> Thanks for putting this together! Anything we can do to streamline the development and deployment is awesome
> 
> `init` is too vague, can we name it `image` or `dev` or something to indicate you're using a development docker image?

Sure, I just wanted something memorable. JS uses `start` by convention. Would you prefer `dev` or `start`

build/deployment/init.go
austinbyers: > What if the user is on a windows machine?

Python is not installed by default on Windows, but it's relatively easy to do so. None of our users nor developers use Windows yet, though I'm sure someday it will happen.

> Do you think opting for a bash script is preferable?

Sure! Good idea, since you're just invoking another command anyway. We can support windows in the future by adding a powershell version of the script

austinbyers:
How about `dev.sh`?

build/deployment/init.go
3nvi: > We can support windows in the future by adding a powershell version of the script
Since windows10 you don't **have to** do that. Windows now have a proper bash shell installed.

Perfect, it's decided then, I'll make the necessary updates

build/deployment/init.go
3nvi: Since we decided to opt for an `.sh` script, I'm flagging this as resolved (since it's irrelevant)

build/deployment/init.go
austinbyers: > Windows now have a proper bash shell installed.

Sure, but most scripts start with `#! /bin/sh`, which I'm not sure will work on Windows. Not a priority at the moment in any case

docs/gitbook/quick-start.md
austinbyers: > Hmmm, c/p it from the existing docs. Guess we need to update them as well 

Yeah, that was my bad, I thought it would work. You can just add a link to https://docs.docker.com/install/

3nvi:
> How about `dev.sh`?

Sure. Should I put it on the root dir (so that it requires less typing to invoke) or keep it under `/build/deployment/`.

austinbyers:
Root dir is fine

docs/gitbook/quick-start.md
3nvi: Will do

build/deployment/init.go
3nvi: `#! /bin/bash` will work just fine on windows. ([related article](https://www.howtogeek.com/261591/how-to-create-and-run-bash-shell-scripts-on-windows-10/)

docs/gitbook/quick-start.md
```diff
@@ -32,29 +32,52 @@ Precise deployment policy coming soon!
...
-2. Install the [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv1.html) and [configure](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html) your credentials
+### Setup the Development Environment
+
+Before you can deploy Panther to your AWS account, you need to setup the development environment.
+You can choose between using our dedicated docker deployment image or manually installing the
+necessary prerequisites yourself.
+
+#### Using our Docker Image
+
+1. Install [Docker](https://docs.docker.com/install/) 17+ and make sure the daemon is running (e.g. typing `docker info` doesn't cause an error)
```
3nvi: @austinbyers do you think we should move cloning the repo & installing docker outside of the "docker vs manually" sections?

These pre-requisites are common for both means of installation (the repo cloning doesn't need to happen under the $GOPATH)

austinbyers: > @austinbyers do you think we should move cloning the repo & installing docker outside of the "docker vs manually" sections?

Sure, good idea. Installing dev tools manually could even move to a separate page to be honest, we don't want to clutter the quick start with too many options, just the minimal friction to get someone up and running

dev.sh
```diff
@@ -0,0 +1,17 @@
...
+
+docker run \
+    -v $(pwd):/code \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    -e AWS_ACCESS_KEY_ID \
+    -e AWS_SECRET_ACCESS_KEY \
+    -e AWS_REGION=$(if [ -z "$AWS_REGION" ]; then echo $AWS_DEFAULT_REGION; else echo $AWS_REGION; fi) \
+    -it \
+    --rm \
+    pantherlabs/panther-development-pack:latest
```
austinbyers: So we need to remember to regularly update the development image, right?

austinbyers: I just cloned this branch (and updated with latest master) and tried `aws-vault exec dev-austin -- ./dev.sh` and got 2 errors:

1. `aws-vault: error: exec: Error execing process: exec: "./dev.sh": permission denied`
    Easy fix (which needs to be checked in): `chmod u+x dev.sh` 
2. `docker: Error response from daemon: manifest for pantherlabs/panther-development-pack:latest not found: manifest unknown: manifest unknown.`

Do we need to publish the image still? How do we do that?

3nvi: I published the image but only as `v1.0.0` not as `latest`. Unfortunately on docker `latest` is not actually "latest". It's just another conventional tag

3nvi: The `permission denied` has to do with the file itself (not its contents). I'll attempt to re-create it

3nvi: Can you check again?

austinbyers: I was able to run the docker image, but then I got

```
root@9389e3b29322:/code# mage deploy
Error: InvalidClientTokenId: The security token included in the request is invalid.
```

So I pushed a commit which adds `AWS_SESSION_TOKEN` and `AWS_SECURITY_TOKEN` environment variables as well. With that fix, it worked for me

3nvi: oh didn't know about these, good to know

austinbyers:
Tested `mage deploy` and `mage test:ci` from the development docker image, it worked for me! Merging now, I'll make more documentation updates in a subsequent PR

docs/gitbook/quick-start.md
3nvi: @austinbyers I've updated it. Can you take a look and correct the phrasing wherever it needs to? Feel free to merge as soon as you are ok

PR #48, opened 2020-01-26
Add support libraries for initial panther fields and DEEP json extraction

rleighton:
## Background
This PR adds support for "panther" fields. These include 3 required fields and multiple optional "any" fields as described in the log analysis design doc (https://panther-labs.atlassian.net/wiki/spaces/EN/pages/100401162/Log+Analysis). The "any" fields make it easy to correlate within and between datasets by using standard naming and collecting the fields from within arbitrary structures in the raw data.

This PR also includes a new pkg called 'extract' which has a single function called 'Exract()' which works with gjson to walk a parsed JSON structure to selectively extract data. **NOTE: a big shout out goes to Nick for working with me on this and suggesting the methodology we use to combine using jonsiter with gjson.** Essentially, for the fields that are polymorphic which were previously declared interface{}, we now declare as *jsoniter.RawMessage . This avoids the expense of jonsiter parsing them and outputs correct JSON when being marshaled BUT gives us the opportunity to parse them with gjson to Extract() fields from deeply nested structures. We use this in the CloudTrail and GuardDuty parsers (future PRs). The unit tests in this PR show examples how how this is done.

The 'AWSExtractor{}' is used for CloudTrail and GuardDuty (they do share some substructures). Note that that extractor will work with ANY structures and is not specific. This way when new CloudTrail events are added in the future it is very likely no code will need to be changed unless there are new kinds of data we want to extract. **For example, this extractor will extract any ARN no matter where it is found. Similarly, if a key in any sub-structure is "accountId" or ends with "AccountId" then we will extract the account id with no need for knowledge of the details of the JSON structure.**

Below is what the CloudTrail event structure looks like using *jsoniter.RawMessage and panther common fields (declared in a sub-struct). There will be future PRs for these log sources. I want to keep this PR small, as I expect there will be much discussion on how to best exploit the variety of data buried in CloudTrail and GuardDuty events.

```
/ CloudTrailRecord is an AWS CloudTrail API log.
type CloudTrail struct {
	AdditionalEventData *jsoniter.RawMessage    `json:"additionalEventData,omitempty"`
	APIVersion          *string                 `json:"apiVersion,omitempty" validate:"required"`
	AWSRegion           *string                 `json:"awsRegion,omitempty" validate:"required"`
	ErrorCode           *string                 `json:"errorCode,omitempty"`
	ErrorMessage        *string                 `json:"errorMessage,omitempty"`
	EventID             *string                 `json:"eventId,omitempty" validate:"required"`
	EventName           *string                 `json:"eventName,omitempty"`
	EventSource         *string                 `json:"eventSource,omitempty"`
	EventTime           *timestamp.RFC3339      `json:"eventTime,omitempty"`
	EventType           *string                 `json:"eventType,omitempty"`
	EventVersion        *string                 `json:"eventVersion,omitempty" validate:"required"`
	ManagementEvent     *bool                   `json:"managementEvent,omitempty"`
	ReadOnly            *bool                   `json:"readOnly,omitempty"`
	RecipientAccountID  *string                 `json:"recipientAccountId,omitempty" validate:"required,len=12,numeric"`
	RequestID           *string                 `json:"requestId,omitempty"`
	RequestParameters   *jsoniter.RawMessage    `json:"requestParameters,omitempty"`
	Resources           []CloudTrailResources   `json:"resources,omitempty"`
	ResponseElements    *jsoniter.RawMessage    `json:"responseElements,omitempty"`
	ServiceEventDetails *jsoniter.RawMessage    `json:"serviceEventDetails,omitempty"`
	SharedEventID       *string                 `json:"sharedEventId,omitempty"`
	SourceIPAddress     *string                 `json:"sourceIpAddress,omitempty"`
	UserAgent           *string                 `json:"userAgent,omitempty"`
	UserIdentity        *CloudTrailUserIdentity `json:"userIdentity,omitempty"`
	VPCEndpointID       *string                 `json:"vpcEndpointId,omitempty"`

	// NOTE: added to end of struct to allow expansion later
	common.PantherLog
}

```

## Testing
* mage test:unit
* testing in dev account with logs sources


internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,214 @@
...
+
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/parsers/timestamp"
+)
+
+// All log parsers should extend from this to get standardized fields (all prefixed with 'p_' as JSON for uniqueness)
+// NOTE: It is VERY important that fields are added to END of the structure to avoid needed to re-build existing Glue partitions.
+//       See https://github.com/awsdocs/amazon-athena-user-guide/blob/master/doc_source/updates-and-partitions.md
+type PantherLog struct {
+	//  required
+	PantherLogType   string            `json:"p_log_type,omitempty" validate:"required" description:"Panther added field with type of log"`            // nolint(lll)
```
austinbyers: Do you feel all these field names need to start with `Panther`? I know we need it in the json output (e.g. `p_log_type`), but do the go struct fields all need the Panther prefix? Seems redundant

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,214 @@
...
+	PantherRowID     string            `json:"p_row_id,omitempty" validate:"required" description:"Panther added field with unique id (within table)"` // nolint(lll)
```
austinbyers: What is the description here used for?

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,214 @@
...
+	PantherAnyAWSTags        *PantherAnyString `json:"p_any_aws_tags,omitempty" description:"Panther added field with collection of aws tags associated with the row"`                 // nolint(lll)
+}
+
+type PantherAnyString struct { // needed to declare as struct (rather than map) for CF generation
+	set map[string]struct{} // map is used for uniqueness, serializes as JSON list
+}
+
+func NewPantherAnyString() *PantherAnyString {
+	return &PantherAnyString{
+		set: make(map[string]struct{}, 4),
```
austinbyers: Any specific reason for the choice of `4`? Usually we just let the compiler use its defaults

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+}
+
+func (any *PantherAnyString) MarshalJSON() ([]byte, error) {
+	if any != nil { // copy to slice
+		values := make([]string, len(any.set))
+		i := 0
+		for k := range any.set {
+			values[i] = k
+			i++
```
austinbyers: If you don't want to maintain an index, you can also do:

```go
values := make([]string, 0, len(any.set)) // create slice with no entries, but the correct capacity
for k := range any.set {
    values = append(values, k)
}
```

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,214 @@
...
+func (any *PantherAnyString) UnmarshalJSON(jsonBytes []byte) error {
+	var values []string
+	err := jsoniter.Unmarshal(jsonBytes, &values)
+	if err != nil {
+		return err
+	}
+	any.set = make(map[string]struct{}, len(values))
+	for i := range values {
+		any.set[values[i]] = struct{}{}
+	}
```
austinbyers: This can be simplified to

```go
for _, entry := range values {
    any.set[entry] = struct{}{}
}
```

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,214 @@
...
+
+// generates uuid for row
+func (pl *PantherLog) SetRequired(logType string, eventTime timestamp.RFC3339) {
+	pl.PantherLogType = logType
+	/* from: https://godoc.org/github.com/google/uuid#New
+	   Randomly generated UUIDs have 122 random bits.  One's annual risk of being
+	   hit by a meteorite is estimated to be one chance in 17 billion, that
+	   means the probability is about 0.00000000006 (6 × 10−11),
+	   equivalent to the odds of creating a few tens of trillions of UUIDs in a
+	   year and having one duplicate.
```
austinbyers: Assuming a perfectly random number generator :P

internal/log_analysis/log_processor/common/patherlog.go
nhakmiller: The descriptions are viewable in at least Athena when viewing the fields, which is nice. Presumably other JSON parsers can show them too.

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	pl.appendAnyString(pl.PantherAnyAWSTags, values...)
+}
+
+func (pl *PantherLog) appendAnyString(any *PantherAnyString, values ...string) {
+	// add new if not present
+	for _, v := range values {
+		if _, exists := any.set[v]; exists {
+			continue
+		}
```
nhakmiller: Is the cost incurred by this conditional that much less than the cost incurred by creating an empty `struct{}{}` and assigning it? It seems you already have uniqueness guaranteed by being a map, I think you can just drop these three lines and it should still work the same.

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,214 @@
...
+				if tagKey.Exists() && tagValue.Exists() {
+					e.pl.AppendAnyAWSTags(tagKey.Str + ":" + tagValue.Str)
+				}
+				return true
+			})
+		}
+
+	case "publicIp", "ipv6Addresses", "privateIpAddress", "ipAddressV4":
+		switch key.Str {
+		case "ipv6Addresses":
```
nhakmiller: Why not just make the `ipv6Addresses` its own stand alone case? For both performance and code clarity reasons that seems preferable to me, unless you think there will be shared code between the `ipv6Addresses` case and the other cases at some point.

Note that in go, you must do an explicit case fallthrough so the default case here will never be executed during the `ipv6Addresses` case, if you intended here to fall through to the `default` case you need to explicitly use the `fallthrough` keyword. Not sure if maybe that is why you organized it like this.

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Agree. This is used as a struct to be composed into log events (see 'Background' above). I wanted to ensure uniqueness. As a practical matter, these are not directly referenced, rather the 'Append' methods are used or seen.

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Yes, these appear in the Athena tables (if you hover on a column you will see this) and in the Glue tables. We will start using this tag on all the log event fields to document.

internal/log_analysis/log_processor/common/patherlog.go
rleighton: That deserves a comment or maybe I should take it out. The default (from memory) is capacity of 1, which is what we often have for these sets but I figured making it accept a few more without an alloc would be a good idea.

internal/log_analysis/log_processor/common/patherlog.go
rleighton: True, figured this was marginally faster, although not sure it matters much.

internal/log_analysis/log_processor/common/patherlog.go
austinbyers: I've done the same in the past (pre-alloc a few spaces), @kostaspap then commented that the micro-optimization isn't worth the question it raises when reading the code. Many sets may only have a single element anyway, right?

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Yup, will tidy.

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Yeah, probably not worth the confusion unless backed by benchmarks.

rleighton: I'll remove.

internal/log_analysis/log_processor/common/patherlog.go
austinbyers: Hmm, the index does appear to be slightly faster, I would have guessed the opposite

```go
package pkg

import (
	"strconv"
	"testing"
)

type PantherAnyString struct {
	set map[string]struct{}
}

var initial = PantherAnyString{set: map[string]struct{}{}}

func init() {
	for count := 0; count < 1000; count++ {
		initial.set[strconv.Itoa(count)] = struct{}{}
	}
}

func BenchmarkSliceIndex(b *testing.B) {
	for i := 0; i < b.N; i++ {
		// Convert to string slice
		result := make([]string, len(initial.set))
		j := 0
		for k := range initial.set {
			result[j] = k
			j++
		}
	}
}

func BenchmarkSliceAppend(b *testing.B) {
	for i := 0; i < b.N; i++ {
		// Convert to string slice
		result := make([]string, 0, len(initial.set))
		for k := range initial.set {
			result = append(result, k)
		}
	}
}
```

gives

```
BenchmarkSliceIndex-12     	 9938925	       116 ns/op
BenchmarkSliceAppend-12    	10058936	       117 ns/op
```

with 3 elements and 

```
BenchmarkSliceIndex-12     	   65049	     17738 ns/op
BenchmarkSliceAppend-12    	   63792	     18255 ns/op
```

with 1000

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Not sure the performance difference ... technically that is alloc (but of zero space) ... and a write vs a read and branch. I would expect the branch to be much faster and w/no impact on gc.

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Yup, will fix ... this was an artifact of me, doing 1 entity at a time ... I got a little myopic around the ips.

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Done

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Done

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Done

internal/log_analysis/log_processor/common/patherlog.go
kostaspap: Agreeing with Austin. Also, nitpick: Maybe rename it to CommonFields or sth like that? Everything in this repo is panther-related anyway :)

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,226 @@
+package common
```
kostaspap: The `common` package was initially created for functions/structs that are shared across all many log processor packages. This file however contains logic specific to the parsers. 

An idea would be to move the PantherLog related code to a new file inside the parsers package e.g. `normalization.go` or similar. The AWSExtractor should be moved inside the parsers/awslogs package since it is meant for AWS-related logs.

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,226 @@
...
+	// key based matching (not exact)
+	if key.Str == "instanceId" || strings.HasSuffix(key.Str, "InstanceId") {
+		e.pl.AppendAnyAWSInstanceIds(value.Str)
+		return
+	}
+
+	if key.Str == "accountId" || strings.HasSuffix(key.Str, "AccountId") {
+		e.pl.AppendAnyAWSAccountIds(value.Str)
+		return
+	}
```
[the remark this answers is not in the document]
rleighton: Good idea.

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,226 @@
...
+	if strings.HasPrefix(value.Str, "arn:") {
+		e.pl.AppendAnyAWSARNs(value.Str)
+		return
+	}
+
+	// key based matching (not exact)
+	if key.Str == "instanceId" || strings.HasSuffix(key.Str, "InstanceId") {
+		e.pl.AppendAnyAWSInstanceIds(value.Str)
+		return
+	}
```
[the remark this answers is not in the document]
rleighton: +1

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,226 @@
...
+
+	// exact key based matching
+	switch key.Str {
+	case "tags":
+		if value.IsArray() {
+			value.ForEach(func(tagListKey, tagListValue gjson.Result) bool {
+				tagKey := tagListValue.Get("key")
+				tagValue := tagListValue.Get("value")
+				if tagKey.Exists() && tagValue.Exists() {
+					e.pl.AppendAnyAWSTags(tagKey.Str + ":" + tagValue.Str)
```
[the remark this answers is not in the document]
rleighton: I could but that would break the varargs pattern I'm using unless I make a struct to hold the groups which seems overkill.

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,226 @@
...
+				}
+				return true
+			})
+		}
+
+	case "ipv6Addresses":
```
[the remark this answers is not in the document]
rleighton: Sure ...

internal/log_analysis/log_processor/common/patherlog.go
```diff
@@ -0,0 +1,226 @@
...
+	   partitions close to event timestamp.
+
+	   About UUID from: https://godoc.org/github.com/google/uuid#New
+	   Randomly generated UUIDs have 122 random bits.  One's annual risk of being
+	   hit by a meteorite is estimated to be one chance in 17 billion, that
+	   means the probability is about 0.00000000006 (6 × 10−11),
+	   equivalent to the odds of creating a few tens of trillions of UUIDs in a
+	   year and having one duplicate.
+	*/
+	return logType + RowIDDelimiter + (time.Time)(eventTime).Format(RowIDTimeFormat) + RowIDDelimiter + uuid.New().String()
```
kostaspap: 1. We need to make sure that we don't use `/dev/random` but rather `/dev/urandom`. `dev/random` is generating cryptographically secure random number but it depletes the entropy pool of a system. If entropy is depleted, the uuid generation blocks until new entropy is available. We generate these UUID on every message we process in a high throughput system running on a server, so use of `/dev/random` should be avoided. 
2. This field is going to be added to every event. It contains a randomly generated field, so it is going to get very low compression ration. Size could become a problem, I wouldn't be surprised if this field takes more storage space than all other fields combined in e.g. VPC Flow logs. So every byte counts.... From the description, it is suggested this has to be used in conjunction with the table name (logtype) in order to find the event someone is looking for. Seems we can remove the logtype from here then?

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Sure. One day the AWSExtractor might get move up into /pkg if more than the log processors use it (maybe compliance).

internal/log_analysis/log_processor/common/patherlog.go
rleighton: 'Common' is just as redundant as 'Panther' (and less likely to be unique). Are you sure this is worth it? Realize that only the SetRequired() and the Any..() methods are used outside this package. Only developers working in this file will ever see these fields. In fact, I would have declared the fields private but they need to be public for JSON serialization.

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Let me muddle this ... maybe there is a way (if we remove logType) to get the hour tag and a unique id into a 63 bit int  (63 because if java readers need signed numbers ... such a PITA).

internal/log_analysis/log_processor/common/patherlog.go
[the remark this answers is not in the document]
kostaspap: I see, that would be definitely an overkill, yes. But the vargs are not really being used anyway, would make sense to just drop them?

internal/log_analysis/log_processor/common/patherlog.go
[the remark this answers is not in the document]
rleighton: Done

internal/log_analysis/log_processor/common/patherlog.go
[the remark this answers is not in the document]
rleighton: Done

internal/log_analysis/log_processor/common/patherlog.go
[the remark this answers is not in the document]
rleighton: I use them right now in tests. Let's leave it, changing it later will not break anything .

internal/log_analysis/log_processor/common/patherlog.go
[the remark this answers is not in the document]
rleighton: Done

internal/log_analysis/log_processor/common/patherlog.go
rleighton: Done

internal/log_analysis/log_processor/common/patherlog.go
rleighton: I removed use of the uuid package and added a deterministic uuid generator. This should be also much faster. I could not get it down to 63bits, sadly. It is variable but usually is around 30 bytes. I've been using Athena to check to make sure there are no duplicates and so far, so good.

internal/log_analysis/log_processor/parsers/rowid.go
```diff
@@ -0,0 +1,111 @@
...
+		zap.L().Error(err.Error(), zap.Error(err))
+		noise := make([]byte, nodeIDSize)
+		rand.Read(noise) // nolint (errcheck) , not checking error because there is noting else to do
+		copy(nodeID[:], noise)
+	} else {
+		zap.L().Debug("Found hardware interface",
+			zap.String("ifName", ifName),
+			zap.String("addr", hex.EncodeToString(addr)))
+		copy(nodeID[:], addr)
+	}
```
kostaspap: UUID Variant 1 uses hardware interface as well. You could take it from there a replace these code and the `getHardwareInterface` function with 
```
id , err := uuid.NewUUID()
id.NodeID()
...
```
It also generates random bytes in case a hardware interface couldn't be found - similar to how we do it here.

internal/log_analysis/log_processor/parsers/rowid.go
```diff
@@ -0,0 +1,111 @@
...
+func ParseRowID(hexID string) (node [nodeIDSize]byte, offset, counter uint64, err error) {
+	id, err := hex.DecodeString(hexID)
+	if err != nil {
+		return
+	}
+	copy(node[:], id[:nodeIDSize])
+	offset, timeOffsetN := binary.Uvarint(id[nodeIDSize:])
+	counter, _ = binary.Uvarint(id[nodeIDSize+timeOffsetN:])
+	return
+}
```
kostaspap: Where are we going to use this function? If we are not going to use it as part of the upcoming PRs I'd prefer to remove it.

internal/log_analysis/log_processor/parsers/rowid.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+var (
+	nodeID [nodeIDSize]byte // mac addr of lambda to use as unique id for host
+
+	// create a time basis relative to rowEpoch to decrease needed number of bits
+	rowEpoch   = time.Date(2020, 1, 1, 0, 0, 0, 0, time.UTC)                   // NEVER CHANGE THIS!
+	timeOffset = (uint64)(time.Now().UTC().Sub(rowEpoch).Nanoseconds()) / 1000 // microseconds resolution
+
+	prefix = make([]byte, nodeIDSize+timeOffsetSize) // holds precomputed prefix
```
kostaspap: nit: The code is using the `len(prefix)` multiple times, and it is a fixed length. You could declare it
`var prefixLength = nodeIDSize + timeOffsetSize`
and could be used in lines 48, 57, 59, 60.

internal/log_analysis/log_processor/parsers/rowid.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	copy(node[:], id[:nodeIDSize])
+	offset, timeOffsetN := binary.Uvarint(id[nodeIDSize:])
+	counter, _ = binary.Uvarint(id[nodeIDSize+timeOffsetN:])
+	return
+}
+
+func init() {
+	// get nodeID
+	ifName, addr := getHardwareInterface()
```
kostaspap: How does this work for Lambda functions? Does each Lambda get its own hardware interface or is it shared? Do you know how it works in case of Fargate ?

internal/log_analysis/log_processor/parsers/rowid.go
```diff
@@ -0,0 +1,111 @@
...
+		zap.L().Debug("Found hardware interface",
+			zap.String("ifName", ifName),
+			zap.String("addr", hex.EncodeToString(addr)))
+		copy(nodeID[:], addr)
+	}
+
+	// compute prefix
+	copy(prefix[:], nodeID[:])                                        // no encoding
+	timeOffsetN := binary.PutUvarint(prefix[nodeIDSize:], timeOffset) // VarInt encoding
+	prefix = prefix[:nodeIDSize+timeOffsetN]                          // clip
```
kostaspap: timeOffsetSize is declared a constant, however from this code I understand it can actually vary?

internal/log_analysis/log_processor/parsers/rowid.go
rleighton: That is where I got the code :) I'd like to know when we needed to use a random interface, this could happen if Lambda changed architecture in the future which is why I did not use uuid as the entry point.

internal/log_analysis/log_processor/parsers/rowid.go
rleighton: I can move to tests.

internal/log_analysis/log_processor/parsers/rowid.go
rleighton: Sure . Although len() is free for slices (will get inlined and slices have this value).

internal/log_analysis/log_processor/parsers/rowid.go
kostaspap: Sounds good

internal/log_analysis/log_processor/parsers/rowid.go
rleighton: Lambda currently gets 3 interfaces: lo and 2 others called "vtarget_1" and :"vinternal_1". For Fargate I expected we would get pseudo interfaces based on docker (not sure, but needs to work since this is linux).

internal/log_analysis/log_processor/parsers/rowid.go
rleighton: Size is fixed in memory but PutUvarint() does a variant compression. I documented that in the comment in the RowID() function. The big advantage of this is with the counter, which is 64 bits but will be stored much smaller. For example, if the count is < 128 then it stores 1 byte. See https://developers.google.com/protocol-buffers/docs/encoding)

internal/log_analysis/log_processor/parsers/rowid.go
rleighton: Actually, I'm going to leave them ... it is only called 2 times and it just as clear. BTW, the len() is not  nodeIDSize + timeOffsetSize because of the variant encoding, it is len(prefix). I could cache that but this is just 2 lines of code.

internal/log_analysis/log_processor/parsers/rowid.go
kostaspap: Got it. So I guess this really varies per Lambda, right? It's not that all Lambdas return the same MAC or something?

internal/log_analysis/log_processor/parsers/rowid.go
kostaspap: Got it. So, `timeOffsetSize` and `rowCounterSize` are more like `maxTimeOffsetSize` and `maxRowCounterSize`, right?

rleighton: They are the size in memory, but used that way for buffer calculation when writing.

internal/log_analysis/log_processor/parsers/rowid.go
rleighton: Mac should be unique within region I believe. By using a micro-second timestamp the uniqueness of the Mac is really only important in the case that we have 2 processes start at the same millionth of a second to distinguish the two.

internal/log_analysis/log_processor/parsers/rowid.go
rleighton: Done

PR #51, opened 2020-01-27
Removing panther-default-outputs table

kostaspap:
Removing panther-default-outputs table. This is a table that we use to keep mapping of severity -> array of outputs IDs. It was meant to allow us to do quick lookups from a severity to the outputs. We honestly don't need such quick look ups, there are little benefits provided by that table given customer will have a handful of outputs IDS. Any benefit would come if we had to deal with 1000s of outputs. 
I am changing the outputs logic so that now we store the default severity in the `panther-outputs` table. 
This has several benefits: 
* One less DDB table in customer accounts
* More robust. Previous logic had specific race condition when two users where trying to update outputs at the same time - they could overwrite the defaults
* Much much simpler code. `338 additions and 1,671 deletions.`

* Removed the GetDefaultOutputs and SetDefaultOutputs methods that are no longer used. 
* Renamed GetOrganizationOutputs method to GetOutputs
* Moved `table.go` from the `/api` folder to the outputs_api folder. It contains DDB structures, it is meant to be used only by outputs_api, not by other services. 
* Deleting `panther-default-outputs`
* Updated tests. Logic is simpler now, so some tests could be removed completely. 
* Some integration tests can now run in parallel (UpdateSns/UpdateSlack). These tests could fail with previous implementation due to race condition. 

* Integration tests
* Deployed to my account. Was able to see, edit, delete outputs. 

*Note* : if you deploy to your local env and you have already outputs configured, this means we lose the "severities" for these outputs. You will have to add them manually back.


internal/core/alert_delivery/delivery/outputs.go
```diff
@@ -22,26 +22,16 @@ import (
...
-	Output    *outputmodels.AlertOutput
-	Timestamp time.Time
-}
-
-type cachedOutputIDs struct {
-	//Map from Severity -> List of output Ids
-	Outputs   map[string][]*string
+type outputsCache struct {
+	//Map from outputID -> List of outputs
+	Outputs   []*outputmodels.AlertOutput
```
nhakmiller: Unless I'm missing something, this is not a map nor is an `outputmodels.AlertOutput` a wrapper for a map. Is this comment accurate?

internal/core/alert_delivery/delivery/outputs.go
```diff
@@ -54,62 +44,54 @@ func getRefreshInterval() time.Duration {
...
+		}
 	}
 
-	zap.L().Info("getting default outputs")
-	input := outputmodels.LambdaInput{GetDefaultOutputs: &outputmodels.GetDefaultOutputsInput{}}
-	var defaultOutputs outputmodels.GetDefaultOutputsOutput
-	if err := genericapi.Invoke(lambdaClient, outputsAPI, &input, &defaultOutputs); err != nil {
-		return nil, err
+	// If alert doesn't have outputs IDs specified, return the defaults for the severity
+	if len(alert.OutputIDs) == 0 {
```
nhakmiller: Do alerts currently specify output ID's? If so, where is that information retrieved from? We would need to update the rules/policy schemas presumably which will effect lots of code.

internal/core/alert_delivery/delivery/outputs.go
kostaspap: Good catch. It used to be a map, now it is a slice. Updating comment.

internal/core/alert_delivery/delivery/outputs.go
kostaspap: So this a leftover from the times that a Panther user was able to specify specific output for each rule/policy which would override the defaults. The Alert included these IDs in the `OutputIDs` field. 

This is a functionality that already [existed](https://github.com/panther-labs/panther/blob/3dd48fbb4f27b6c1068613191e063e2dc88cb633/internal/core/alert_delivery/delivery/outputs.go#L64-L67) in the code, I didn't add it now. 
I was tempted to remove it, but thought my PR was already big enough and didn't want to touch another functionality.

PR #53, opened 2020-01-28
[ENG-1278] Compression & Caching on Frontend files

3nvi:
Currently we are not utilizing a CDN, but instead serve the files from our express server. A CDN has a lot more to it apart from just "quicker delivery"; it adds client caching by attaching response headers to the associated file request, it performs JIT compression, etc.

These things have become de-facto for the web and not having them is causing a substantial initial loading time for the app. This PR adds just that, specifically:

* Adds `Etag`, `Last-Modified` and `Cache-control` headers, depending on the type of file
* Compresses JS and SVG assets using brotli
* Handles the serving of the compressed content, depending on whether the browser supports the algorithm or not.

This results into:

* Long term caching of immutable files (browsers cache them in local memory and don't need to issue requests to get them)
* Up to 3x faster initial loads 

* Add brotli compression as part of the build process 
* Add express middleware for serving of compressed files (i.e. redirect to appropriate compression during a vanilla file request) and addition of additional response headers
* Minor fixes here and there

* Locally, on dev server & on every major modern browser


deployments/web/Dockerfile
```diff
@@ -69,17 +69,18 @@ WORKDIR /code
 COPY --from=build-env /code/web/dist dist
 
 # Add the script that runs the express server
-ADD  web/scripts/serve.js scripts/serve.js
+ADD package.json .
```
kostaspap: Do we also need to add the `package-lock.json`? I see we are adding it on line 34

web/scripts/serve.js
```diff
@@ -18,13 +18,39 @@
...
 /* eslint-disable no-console  */
 const express = require('express');
+const expressStaticGzip = require('express-static-gzip');
 const path = require('path');
 
 // construct a mini server
 const app = express();
 
+const getCacheControlForFile = filepath => {
+  if (/favicon.*\.(png|svg|ico)/.test(filepath)) {
```
kostaspap: Interesting, why the favicon needs different handling than "normal" svg/jpg?

3nvi: because it doesn't have versioning (i.e. a random hash in its name). 

The reason is that we want a "fixed" reference to a favicon, since HTML, manifests and other external files need to reference it.

3nvi: So, we wouldn't want to cache it for a year, but rather a week or so

deployments/web/Dockerfile
3nvi: Good catch. Currently, if I add the `package-lock.json` and try to run `npm install express` it will not only install express, but every other package it sees within the lockfile.

I don't know if this is a bug or whether the lockfile is used to validate that all the packages are satisfied, but by doing that, I'm increasing the size of the image from 100MB to 1GB. The only downside to that is that we are not guaranteed to install the same minor version of the package. Due to this, I should add a fixed version to `express` in order not to depend on the lockfile.

I will do it right away

PR #59, opened 2020-01-29
[ENG-1355]Add issue type for jira destination

drixta:
The JIRA destination currently defaults to creating a `Task`. We don't have `Task`s enabled in our JIRA account so this fails. I've switched it over to create a `Bug` and it works, but theoretically another organization may have and prefer `Task`s and not even have `Bug`s enabled. Please add an input named `type` or something similar, which can accept the values `Bug`, `Story`, and `Task`. Preferably in a convenient little dropdown select thingy or something so users don't have to type it 

In addition, we'll change the user name to email address

* Add `issueType` into the jira destination form which is reflected in `schema.graphql`, list destinations component and the jira form

* Local testing with a deployed instance on personal account

### DIFF
--- a/web/src/components/forms/destination-form/jira-destination-form.tsx
+++ b/web/src/components/forms/destination-form/jira-destination-form.tsx
@@ -20,11 +20,13 @@ import React from 'react';
 import { Field } from 'formik';
 import * as Yup from 'yup';
 import FormikTextInput from 'Components/fields/text-input';
+import FormikCombobox from 'Components/fields/combobox';
 import { DestinationConfigInput } from 'Generated/schema';
 import BaseDestinationForm, {
   BaseDestinationFormValues,
   defaultValidationSchema,
 } from 'Components/forms/common/base-destination-form';
+import { JIRA_ISSUE_TYPE } from 'Source/constants';
 
 type JiraFieldValues = Pick<DestinationConfigInput, 'jira'>;
 
@@ -43,6 +45,7 @@ const jiraFieldsValidationSchema = Yup.object().shape({
       projectKey: Yup.string().required(),
       apiKey: Yup.string().required(),
       assigneeId: Yup.string(),
+      issueType: Yup.string().required(),
     }),
   }),
 });
@@ -80,7 +83,7 @@ const JiraDestinationForm: React.FC<JiraDestinationFormProps> = ({ onSubmit, ini
       <Field
         as={FormikTextInput}
         name="outputConfig.jira.userName"
-        label="User Name"
+        label="Email"
         placeholder="What's the name of the reporting user?"
         mb={6}
       />
@@ -100,6 +103,14 @@ const JiraDestinationForm: React.FC<JiraDestinationFormProps> = ({ onSubmit, ini
         label="Assignee ID"
         placeholder="Who should we assign this to?"
         mb={6}
+      />
+      <Field
+        as={FormikCombobox}
+        name="outputConfig.jira.issueType"
+        label="Issue Types"
+        mb={6}
+        aria-required
+        items={JIRA_ISSUE_TYPE}
       />
     </BaseDestinationForm>
   );
--- a/web/src/components/forms/destination-form/jira-destination-form.tsx
+++ b/web/src/components/forms/destination-form/jira-destination-form.tsx
@@ -45,7 +45,7 @@ const jiraFieldsValidationSchema = Yup.object().shape({
       apiKey: Yup.string().required(),
       assigneeId: Yup.string(),
       issueType: Yup.string().test('oneOf', 'Please select a valid value', value =>
-        Object.keys(JiraIssueTypesEnum).includes(value)
+        Object.values(JiraIssueTypesEnum).includes(value)
       ),
     }),
   }),
@@ -112,6 +112,7 @@ const JiraDestinationForm: React.FC<JiraDestinationFormProps> = ({ onSubmit, ini
         mb={6}
         aria-required
         items={Object.keys(JiraIssueTypesEnum)}
+        inputProps={{ placeholder: 'Select a type of issue' }}
       />
     </BaseDestinationForm>
   );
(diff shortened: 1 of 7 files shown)


web/src/constants.ts
```diff
@@ -79,6 +79,8 @@ export const LOG_TYPES = [
   'Osquery.Status',
 ] as const;
 
+export const JIRA_ISSUE_TYPE = ['Bug', 'Story', 'Task'] as const;
```
3nvi: I think this should be an `enum` on the `schema.graphql`, since we want the validation that an issue can only be a bug, story or task.

We should take care with casing though. I'd recommend lowercase only

3nvi: I think this should be an enum on the schema.graphql, since we want the validation that an issue can only be a bug, story or task.

We should take care with casing though. I'd recommend lowercase only

3nvi: name should be `JIRA_ISSUE_TYPES` (plural)

web/src/components/forms/destination-form/jira-destination-form.tsx
```diff
@@ -80,7 +83,7 @@ const JiraDestinationForm: React.FC<JiraDestinationFormProps> = ({ onSubmit, ini
       <Field
         as={FormikTextInput}
         name="outputConfig.jira.userName"
```
3nvi: @nhakmiller are we ok with keeping the value `userName` in the backend, while shoowing the user the label "Email"?

web/src/components/forms/destination-form/jira-destination-form.tsx
```diff
@@ -80,7 +83,7 @@ const JiraDestinationForm: React.FC<JiraDestinationFormProps> = ({ onSubmit, ini
...
-        label="User Name"
+        label="Email"
         placeholder="What's the name of the reporting user?"
```
3nvi: placeholder should mention `email`

web/src/components/forms/destination-form/jira-destination-form.tsx
```diff
@@ -43,6 +45,7 @@ const jiraFieldsValidationSchema = Yup.object().shape({
       projectKey: Yup.string().required(),
       apiKey: Yup.string().required(),
       assigneeId: Yup.string(),
+      issueType: Yup.string().required(),
```
3nvi: OPTIONAL: We can add a value validation as well through:

```js
Yup.string().test('oneOf', 'Please select a valid value', value => JIRA_ISSUE_TYPES.includes(value)
```

web/src/components/forms/destination-form/jira-destination-form.tsx
```diff
@@ -101,6 +104,14 @@ const JiraDestinationForm: React.FC<JiraDestinationFormProps> = ({ onSubmit, ini
         placeholder="Who should we assign this to?"
         mb={6}
       />
+      <Field
+        as={FormikCombobox}
+        name="outputConfig.jira.issueType"
+        label="Issue Types"
```
3nvi: should be `Issue Type` since users select one value only

web/src/pages/destinations/index.tsx
```diff
@@ -59,6 +59,7 @@ export const LIST_DESTINATIONS = gql`
           userName
           apiKey
           assigneeId
+          issueType
```
3nvi: We are missing the same thing from:

```graphql
mutation AddDestination()
```

and 

```graphql
mutation UpdateDestination
```

3nvi: In addition, we need to pass the value of `issueType` to the update form, else it's not going to have a proper initial value

3nvi:
@drixta Thanks for the help on that, we need a couple of changes since there will be some minor bugs if we leave it like that. I've also added some comments that are unrelated to the issue being merged, but reflect my personal opinions.

The issue will be fine by me as long as we make sure that all the CRUD actions work correctly

web/src/components/forms/destination-form/jira-destination-form.tsx
nhakmiller: Yes that is totally fine.

web/src/constants.ts
nhakmiller: JIRA expects exactly `Bug`, `Story`, or `Task` when we talk to it. If you want to force everything to lower case before sending it to us that's fine, we can uppercase the first letter on the backend (also I'm not sure if JIRA would accept `bug` etc., I can test that if you decide to lower case everything, but either way we can handle it on the backend).

drixta:
Addressed the comments

web/src/constants.ts
3nvi: I just saw that we can add a word with mixed casing in a GraphQL enum. We can safely disregard all this chat :)

web/src/components/forms/destination-form/jira-destination-form.tsx
```diff
@@ -43,6 +44,9 @@ const jiraFieldsValidationSchema = Yup.object().shape({
       projectKey: Yup.string().required(),
       apiKey: Yup.string().required(),
       assigneeId: Yup.string(),
+      issueType: Yup.string().test('oneOf', 'Please select a valid value', value =>
+        Object.keys(JiraIssueTypesEnum).includes(value)
```
3nvi: This should be `Object.values` since that's what the end value of the form element is

web/src/components/sidesheets/add-destination-sidesheet.tsx
```diff
@@ -72,6 +72,7 @@ const ADD_DESTINATION = gql`
           userName
           apiKey
           assigneeId
+          issueType
```
3nvi: we are also missing the initial value for this field.  Should be like this:

```tsx
<JiraDestinationForm
            initialValues={{
              ...commonInitialValues,
              outputConfig: {
                jira: {
                  // ... other initial values
                  issueType: ''
                },
              },
            }}
            // ... other props
          />
```

nhakmiller:
> @drixta Went ahead and pushed the 2 final comments that I've had.
> 
> Feel free to merge whenever you want

Since you're ok with it I'll merge it so I can start using it.

PR #60, opened 2020-01-29
[ENG-1370] Reduce CPU & RAM for ECS

3nvi:
Costs for ECS have been high for some time now, this PR will attempt to reduce them. Specifically:

* ECS Front-end server RAM costs will be reduced by $3.2 per month (-50%)
* ECS Front-end server vCPU costs will be reduced by $14.57 per month (-50%)

The estimated new monthly cost for ECS (front-end web app only) will now be:

**$14.57 (vCPU) + $3.2 (RAM) = $17.77 total cost per month**

@jacknagz @austinbyers I can lower the CPU to $7.28, but we will need to introduce Cloudfront for static assets, since the server will be slow and won't be able to handle static assets in a quick fashion

* Reduce vCPUs used
* Reduce RAM used

* Dev account and local manual testing


deployments/web/server.yml
```diff
@@ -195,15 +195,15 @@ Resources:
...
               awslogs-stream-prefix: !Ref ServiceName
-          Memory: 2048
-          MemoryReservation: 2048
+          Memory: 1024
+          MemoryReservation: 1024
           PortMappings:
             - ContainerPort: !Ref Port
           ReadonlyRootFilesystem: false
-      Cpu: '1024'
+      Cpu: '512'
```
austinbyers: Do these have to be strings? Try without, usually CloudFormation will work with numbers (like the ones you have above)

3nvi: They need to be strings. Cloudformation complains when they are not :/

3nvi: Although their value is a number, both definitions are `Type: String` in the [related cloudformation docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-cpu)

deployments/panther_config.yml
```diff
@@ -50,6 +49,19 @@ BackendParameterValues:
   # XRay tracing mode for API Gateway and Lambda: '', 'Active', or 'PassThrough'
   TracingMode: ''
 
+FrontendParameterValues:
+  # The size of the CPU allocated to the front-end web server.
+  # Allowed values: [256, 512, 1024]
+  #
+  # For example, a value of 1024 equals to 1 vCPU, while a value of 256 equals to 0.25 vCPU
+  WebApplicationServerCPU: 512
```
austinbyers: Technically, the web app is still serverless, right?

> AWS Fargate: Serverless compute for containers

So maybe let's avoid calling it the "front-end web server" just to avoid confusion. Perhaps The front-end task? web app? front-end fargate?

deployments/web/server.yml
```diff
@@ -195,15 +201,15 @@ Resources:
...
               awslogs-stream-prefix: !Ref ServiceName
-          Memory: 2048
-          MemoryReservation: 2048
+          Memory: !Ref Memory
+          MemoryReservation: !Ref Memory
           PortMappings:
             - ContainerPort: !Ref Port
           ReadonlyRootFilesystem: false
-      Cpu: '1024'
+      Cpu: !Sub ${CPU}
```
austinbyers: This should be `!Ref CPU`

deployments/web/server.yml
```diff
@@ -195,15 +201,15 @@ Resources:
...
       ExecutionRoleArn: !GetAtt WebApplicationServerExecutionRole.Arn
       Family: !Ref ServiceName
-      Memory: '2048'
+      Memory: !Sub ${Memory}
```
austinbyers: `!Ref Memory`

docs/gitbook/cost.md
```diff
@@ -0,0 +1,29 @@
+---
+description: Costs associated with Panther
```
austinbyers: We need to make it clear these are AWS costs, not Panther's enterprise pricing.

`description: AWS costs associated with running Panther`

docs/gitbook/cost.md
```diff
@@ -0,0 +1,29 @@
...
+# Costs
+
+Panther is proud to be mostly implemented on serverless technologies. This means that you will only pay for your use
+of the product and won't be charged when you don't use it.
+
+Unfortunately, some pieces of our software solution require infrastructure continuously "running", which
+means that there is going to be a minimum monthly cost associated with the deployment of our product in your
+AWS Account.
+
+The following sections analyze & associate this cost with the related technologies.
```
austinbyers: I think all of Panther is officially serverless by AWS definitions, even Fargate.

What do you think of the following replacement for this first section:

> Panther is proud to be built entirely on modern serverless technologies. A huge benefit of serverless designs is their low-cost, usage-based pricing. As a result, Panther is quite cheap to deploy and you won't have to pay for most AWS services until you onboard data!
>
> However, some infrastructure has an ongoing cost regardless of usage. Our best estimate is that Panther has a minimum AWS bill of about **$20/month** while it's deployed, depending on your AWS region and Panther configuration settings. For example, Panther creates a custom KMS key for SQS encryption, which has a fixed cost of $1/month. But the main running cost is associated with running the web application continuously in Fargate:
>
> ## Front-end web server
> ...

docs/gitbook/cost.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Front-end web server
+
+In order to serve you a web application, an ECS FARGATE service (named `panther-web`) has a single
```
austinbyers: > In order to serve you a web application, an ECS FARGATE service

=>

> In order to serve the web application, an ECS Fargate service

docs/gitbook/cost.md
```diff
@@ -0,0 +1,29 @@
...
+task running, which acts as a front-end server. By default this task gets allocated 0.5 vCPU and 1024MB
```
austinbyers: By default, (add comma)

docs/gitbook/cost.md
```diff
@@ -0,0 +1,29 @@
...
+of memory. This leads to a monthly cost of **$14.57 (vCPU) + $3.2 (RAM) = \$17.77** according to the [official ECS pricing page](https://aws.amazon.com/fargate/pricing/).
+
+This means that even if you don't actually use Panther at all, you will still be asked to pay
+**\$17.77** for the cost of running an elastic service. If you want to lower this cost
+(in exchange for a slower server and an increased web application loading time), you can
+modify the parameters found in [panther_config.yml](https://github.com/panther-labs/panther/blob/master/deployments/panther_config.yml). Specifically,
+you can lower `WebApplicationServerCPU` to `256`, lower `WebApplicationServerMemory` to `512` and deploy (or re-deploy) Panther.
+
+These values are the min allowed values that the front-end server can receive and they will drop the costs
+associated with it, down to **\$8.88** per month.
```
austinbyers: Users could also stop the task while it's not needed, right? If the task is manually stopped, will the next deploy re-enable it?

tools/mage/deploy.go
```diff
@@ -186,9 +186,11 @@ func getBackendDeployParams(awsSession *session.Session, config *PantherConfig,
 	return result, nil
 }
 
-func getFrontendDeployParams(image string, backendOutputs map[string]string) map[string]string {
-	// If there are params declared in config, we should make sure to add them as well. Currently there are none.
+func getFrontendDeployParams(config *PantherConfig, image string, backendOutputs map[string]string) map[string]string {
+	v := config.FrontendParameterValues
```
austinbyers: Let's rename the variable to `frontend`

deployments/panther_config.yml
3nvi: sure

deployments/web/server.yml
3nvi: If I do it like that, won't it take a "number" value? That's because `CPU` is a number, but I want to convert it to a string

deployments/web/server.yml
3nvi: again same comment as above,

deployments/web/server.yml
austinbyers: I don't think there should be any difference between `!Sub ${CPU}` and `!Ref CPU`. If it doesn't work, just change the parameter type to a `String` instead of a `Number`

I'd rather rely on explicit string parameter types instead of a no-op Sub to do an implicit type conversion. I as a developer wouldn't realize that's what was happening

docs/gitbook/cost.md
3nvi: Unless they "kill" the service, stopping a task will just spawn another one (cause the system always needs to be consistent).

I don't wanna recommend killing it. It's a hack that I don't like and may lead to problems upgrading down the line

deployments/web/server.yml
3nvi: The thing is that we need it **both** as a number **and** as a string. There are 2 places where we use the `Cpu` value. One field requires a Number and the other requires a String (for the exact same thing :S)

3nvi: I'll try with `!Ref` and let you know

docs/gitbook/cost.md
3nvi: yup

deployments/web/server.yml
austinbyers: I've always hated how CloudFormation handles this

3nvi: Worked fine

PR #62, opened 2020-01-29
Adding Nginx Access logs parser

kostaspap:
Adding parser for Nginx Access logs

* Added new parser. Added parser to registry. 

* Unit tests
* Deployed and seen both Athena table and Log list populated: 

Screen Shot 2020-01-29 at 4 23 06 PM
Screen Shot 2020-01-29 at 4 23 51 PM


internal/log_analysis/log_processor/parsers/nginx/access.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Parse returns the parsed events or nil if parsing failed
+func (p *AccessParser) Parse(log string) []interface{} {
+	reader := csv.NewReader(strings.NewReader(log))
+	// Separator between fields is the empty space
+	reader.Comma = ' '
+
+	records, err := reader.ReadAll()
+	if len(records) == 0 || err != nil {
+		zap.L().Debug("failed to parse the log as csv")
```
nhakmiller: Seems a bit misleading as this is not really a csv file, even if we are using a csv reader on the backend. Maybe re-word to specify the file is space separated and the csv reader failed to read it? Not sure if necessary I'll leave up to your opinion entirely just thought I'd point out it seemed weird to me.

internal/log_analysis/log_processor/parsers/nginx/access.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil
+	}
+
+	// parser should only receive 1 line at a time
+	if len(records) > 1 {
+		zap.L().Debug("failed to parse the log as csv")
```
nhakmiller: This debug message is identical to the above, if trying to do debugging on this code you would have no idea for which reason it failed. Below you provide additional information as to why it's consider failed, I think it would be beneficial to include such information here as well.

internal/log_analysis/log_processor/parsers/nginx/access.go
rleighton: I would just take off 'as csv' to message.

jacknagz:
@kostaspap this is awesome, can we please [update the docs](https://docs.runpanther.io/log-analysis/supported-logs) to include this new type?

internal/log_analysis/log_processor/parsers/nginx/access.go
kostaspap: Sure. Removed "csv" text from logging.

internal/log_analysis/log_processor/parsers/nginx/access.go
kostaspap: Done. Added different messages with more information.

kostaspap:
> update the docs

I initially merged the PR thinking that I will change the documentation by logging in to Gitbook - forgot we actually sync gitbook from the repo. 
Created a new PR to update documentation here: https://github.com/panther-labs/panther/pull/72

PR #64, opened 2020-01-29
ENG-1247 Create views database and initial view using standard fields

rleighton:
## Background
Adding a panther_views database and creating first view which searches standard fields.

## Changes
*  Added views database
* Added Athena bucket for results (will also use this in CTAS work in future)
* Added pkg/Athena 

## Testing
* mage test:unit
* deployed to dev account and tested


pkg/awsathena/query.go
```diff
@@ -0,0 +1,103 @@
...
+	return &AthenaQuery{
+		Session:       athena.New(sess),
+		SQL:           sql,
+		S3ResultsPath: s3Path,
+		Database:      database,
+	}
+}
+
+func (aq *AthenaQuery) Run() (err error) {
+	qei := athena.StartQueryExecutionInput{}
```
austinbyers: If initializing an empty struct directly, my personal preference is

```go
var qei athena.StartQueryExecutionInput
```

pkg/awsathena/query.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	if *qeo.QueryExecution.Status.State == "SUCCEEDED" {
+		ip := athena.GetQueryResultsInput{}
+		ip.SetQueryExecutionId(*aq.startResult.QueryExecutionId)
+
+		aq.QueryResult, err = aq.Session.GetQueryResults(&ip)
+		if err != nil {
+			err = errors.Wrapf(err, "athena failed running: %#v", *aq)
+			return err
```
austinbyers: Simplify to:

```go
return errors.Wrapf(...)
```

pkg/awsathena/query.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	} else {
+		err = errors.Errorf("athena failed with status %s running: %#v", *qeo.QueryExecution.Status.State, *aq)
```
austinbyers:
```go
return errors.Errorf(....)
```

Is there a difference between that and `fmt.Errorf`?

pkg/awsathena/query.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	return err
```
austinbyers: I prefer returning an error immediately when it's encountered (see above comments) so the developer doesn't have to reason here about whether this variable might be non-nil.

So with the above comments, this is now just `return nil`

tools/athenaviews/athenaviews.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/panther-labs/panther/pkg/awsathena"
+	"github.com/panther-labs/panther/pkg/awsglue"
+	"github.com/panther-labs/panther/tools/cfngen/gluecf"
+)
+
+// CreateOrReplaceViews will update Athena with all views in the Panther view database
+func CreateOrReplaceViews(athenaResultsBucket string) (err error) {
+	sess, err := session.NewSession()
+	if err != nil {
+		err = errors.Wrapf(err, "CreateOrReplaceViews() failed")
```
austinbyers: You don't have any string formatting, just use `errors.Wrap`

For example, the linter will complain if you use `fmt.Printf("string")` instead of `fmt.Print("string")`

tools/athenaviews/athenaviews.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	s3Path := "s3://" + athenaResultsBucket + "/athena/"
+	sqlStatements := GenerateViews(registry.AvailableTables())
+	for _, sql := range sqlStatements {
+		q := awsathena.NewAthenaQuery(sess, awsglue.ViewsDatabaseName, sql, &s3Path) // use default bucket
+		err = q.Run()
+		if err != nil {
+			err = errors.Wrapf(err, "CreateOrReplaceViews() failed")
+			return err
+		}
```
austinbyers: From 5 lines to 3:

```go
if err = q.Run(); err != nil {
    return errors.Wrap(err, "CreateOrReplaceViews() failed")
}
```

tools/athenaviews/athenaviews.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return err
+}
+
+// GenerateViews creates useful Athena views in the panther views database
+func GenerateViews(tables []*awsglue.GlueMetadata) (sqlStatements []string) {
+	sqlStatements = append(sqlStatements, generateViewAllLogs(tables))
+	return
+}
```
austinbyers: Isn't this just the same thing as

```go
func GenerateViews(tables []*awsglue.GlueMetadata) []string {
    return []string{generateViewAllLogs(tables)}
}
```

tools/athenaviews/athenaviews.go
```diff
@@ -0,0 +1,85 @@
...
+
+// generateViewAllLogs creates a view over all log sources using "panther" fields
+func generateViewAllLogs(tables []*awsglue.GlueMetadata) (sql string) {
+	columns := gluecf.InferJSONColumns(parsers.PantherLog{}, gluecf.GlueMappings...)
+	selectClause := ""
+	for _, col := range columns {
+		selectClause += col.Name + ","
+	}
+	selectClause += "year,month,day,hour"
```
austinbyers: Consider using a [`strings.Builder`](https://golang.org/pkg/strings/#Builder) instead of concatenation

tools/athenaviews/athenaviews.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// GenerateViews creates useful Athena views in the panther views database
+func GenerateViews(tables []*awsglue.GlueMetadata) (sqlStatements []string) {
+	sqlStatements = append(sqlStatements, generateViewAllLogs(tables))
+	return
+}
+
+// generateViewAllLogs creates a view over all log sources using "panther" fields
+func generateViewAllLogs(tables []*awsglue.GlueMetadata) (sql string) {
```
austinbyers: This function seems like an easy one to write unit tests for, can we add them?

pkg/awsathena/query.go
rleighton: Yeah, you get a stack trace added.

tools/athenaviews/athenaviews.go
rleighton: Yes, but I expect more in the future.

tools/athenaviews/athenaviews.go
rleighton: Why? I thought concatenation is Go (unlike Java) was usually more efficient due to compiler. Is that not correct?

tools/athenaviews/athenaviews.go
rleighton: Sure.

tools/athenaviews/athenaviews.go
rleighton: Fixed

tools/athenaviews/athenaviews.go
rleighton: Fixed

tools/athenaviews/athenaviews.go
rleighton: Anyway, this is not a performance critical code where a buffer is really needed. If you think it makes the code clearer I'll do it but I not sure it adds much.

pkg/awsathena/query.go
rleighton: Fixed

pkg/awsathena/query.go
rleighton: Fixed

pkg/awsathena/query.go
rleighton: Fixed

tools/athenaviews/athenaviews.go
rleighton: Fixed

tools/athenaviews/athenaviews.go
austinbyers: Even then, I'd still say you don't need the naked return:

```go
func GenerateViews(tables []*awsglue.GlueMetadata) []string {
    result = []string{generateViewAllLogs(tables)}
    result = append(result, moreThings...)
    return result
}
```

tools/athenaviews/athenaviews.go
austinbyers: > Why? I thought concatenation is Go (unlike Java) was usually more efficient due to compiler. Is that not correct?

For single strings, it's fairly fast. But string concatenation in a loop requires lots of memory copies. Did a quick benchmark, the string builder is faster but a string join is faster still:

```go
package panther

import (
	"strconv"
	"strings"
	"testing"
)

var columns = make([]string, 100)

func init() {
	for i := 0; i < len(columns); i ++ {
		columns[i] = "testcolumn-" + strconv.Itoa(i)
	}
}

func BenchmarkConcat(b *testing.B) {
	for i := 0; i < b.N; i++ {
		selectClause := ""
		for _, col := range columns {
			selectClause += col + ","
		}
		selectClause += "year,month,day,hour"

		// Use final string somehow
		if !strings.HasSuffix(selectClause, ",hour") {
			b.FailNow()
		}
	}
}

func BenchmarkBuilder(b *testing.B) {
	for i := 0; i < b.N; i++ {
		var builder strings.Builder
		for _, col := range columns {
			builder.WriteString(col + ",")
		}
		builder.WriteString("year,month,day,hour")
		result := builder.String()

		// Use final string somehow
		if !strings.HasSuffix(result, ",hour") {
			b.FailNow()
		}
	}
}

func BenchmarkJoin(b *testing.B) {
	for i := 0; i < b.N; i++ {
		var fields = make([]string, 0, len(columns) + 4)
		for _, col := range columns {
			fields = append(fields, col)
		}
		fields = append(fields, "year", "month", "day", "hour")
		result := strings.Join(fields, ",")

		// Use final string somehow
		if !strings.HasSuffix(result, ",hour") {
			b.FailNow()
		}
	}
}
```

```
BenchmarkConcat-12     	   86851	     13049 ns/op
BenchmarkBuilder-12    	  401463	      2970 ns/op
BenchmarkJoin-12       	  738624	      1543 ns/op
```

rleighton: This was the post I remember reading on the subject: https://gist.github.com/dtjm/c6ebc86abe7515c988ec

For big stuff I always use a buffer , just because.

tools/athenaviews/athenaviews.go
rleighton: True. Fixed. Is there a way to customize the linter to not allow naked returns? Old habits die hard and I used to do naked returns on short functions and the linter allows it up to ~30 lines.

tools/athenaviews/athenaviews.go
rleighton: Move to slices .

tools/athenaviews/athenaviews.go
rleighton: Fixed

austinbyers: > Is there a way to customize the linter to not allow naked returns?

It's really just a personal preference, I don't feel too strongly about it. I believe @kostaspap prefers avoiding naked returns as well?

You can edit the linter settings in [`.golangci.yml`](https://github.com/panther-labs/panther/blob/master/.golangci.yml#L52) (see [documentation](https://github.com/golangci/golangci-lint#config-file)):

```
linters-settings:
  nakedret:
    max-func-lines: 30  # default, change to whatever
```

If we decide to limit this for consistency, I'd be fine with setting the value as low as 0 or 1, since we can always disable that check for functions that require it (using `defer` to change return values)

pkg/awsathena/query.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/aws/aws-sdk-go/service/athena"
+	"github.com/pkg/errors"
+)
+
+const (
+	pollDelay = time.Second * 2
+)
+
+type AthenaQuery struct {
+	Session       *athena.Athena
```
kostaspap: I think this is the Athena client, right? (Not the session). 
Better rename and also use the `athenaiface.AthenaAPI` interface so we can mock it in our tests.

pkg/awsathena/query.go
```diff
@@ -0,0 +1,101 @@
...
+		S3ResultsPath: s3Path,
+		Database:      database,
+	}
+}
+
+func (aq *AthenaQuery) Run() (err error) {
+	var qei athena.StartQueryExecutionInput
+	qei.SetQueryString(aq.SQL)
+
+	var qec athena.QueryExecutionContext
```
kostaspap: nit: Can you rename this to  `executionContext`, `qei` to sth like `executionInput`? Makes easier to read

pkg/awsathena/query.go
```diff
@@ -0,0 +1,101 @@
...
+	qec.SetDatabase(aq.Database)
+	qei.SetQueryExecutionContext(&qec)
+
+	var rc athena.ResultConfiguration
```
kostaspap: nit: Rename to `athenaResult` or `resultConfiguration`

pkg/awsathena/query.go
```diff
@@ -0,0 +1,101 @@
...
+
+	aq.startResult, err = aq.Session.StartQueryExecution(&qei)
+	if err != nil {
+		err = errors.Wrapf(err, "athena failed running: %#v", *aq)
+	}
+	return err
+}
+
+func (aq *AthenaQuery) Wait() (err error) {
+	var qei athena.GetQueryExecutionInput
```
kostaspap: nit: similar to above. Please rename `qei`, `qeo` to something else.

tools/mage/deploy.go
```diff
@@ -119,6 +121,14 @@ func Deploy() error {
 		return err
 	}
 
+	// Athena views are created via API call because CF is not well supported. Workgroup "primary" is default.
+	if err := awsathena.WorkgroupAssociateS3(awsSession, "primary", backendOutputs["AthenaResultsBucket"]); err != nil {
```
kostaspap: Since it is the default, is this association needed?

tools/athenaviews/athenaviews.go
```diff
@@ -0,0 +1,87 @@
...
+		err = q.Wait()
+		if err != nil {
+			return errors.Wrap(err, "CreateOrReplaceViews() failed")
+		}
+	}
+	return err
+}
+
+// GenerateViews creates useful Athena views in the panther views database
+func GenerateViews(tables []*awsglue.GlueMetadata) (sqlStatements []string) {
```
kostaspap: So we use Athena to generate the views, however the rest of the tables we setup using CFN. 
It seems that CFN supports declaring views (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-glue-table-tableinput.html#cfn-glue-table-tableinput-tabletype) 
```
TableType
The type of this table (EXTERNAL_TABLE, VIRTUAL_VIEW, )
```

It would be preferable to declare the views in CFN since we do like this with the rest. Any reason why you opted for creating the views using Athena?

pkg/awsathena/query.go
rleighton: Good point. Right now I just do integration tests.

pkg/awsathena/query.go
rleighton: Sure

pkg/awsathena/query.go
rleighton: sure

tools/mage/deploy.go
rleighton: Sadly this is needed and super important.  Athena starts WITHOUT a default S3 bucket and NO queries can be run until you make this association. A very annoying boot strap problem.

tools/athenaviews/athenaviews.go
rleighton: We discussed this before .. the view has a contrived structure and I really do not want to use it. It requires declaring schema as Glue and the again as Presto (types are different) in a base64 encoded string with the SQL. A giant ugly hack. I expect them to change this ins the future also. See: https://stackoverflow.com/questions/56289272/create-aws-athena-view-programmatically

Hideous.

pkg/awsathena/query.go
rleighton: Done

pkg/awsathena/query.go
rleighton: Done

tools/athenaviews/athenaviews.go
kostaspap: So this method actually generates the SQL statements that can be used to create views (name & comment is kinda misleading). 
Could be named `generateViewsSqlStatements` or just `generateViewsSql`. Also seems it can be just package visible? Don't see some other function using it. 
Also, seems it should be only package visible.

pkg/awsathena/query.go
```diff
@@ -0,0 +1,102 @@
...
+	Client        athenaiface.AthenaAPI
+	SQL           string
+	S3ResultsPath *string // this can be nil, to use defaults
+	Database      string
+	QueryResult   *athena.GetQueryResultsOutput
+	// internal state
+	startResult *athena.StartQueryExecutionOutput
+}
+
+func NewAthenaQuery(sess *session.Session, database, sql string, s3Path *string) *AthenaQuery {
```
kostaspap: Every time we want to run a new query this creates a new client... it's not required and seems unnecessary. Alternative design could be like this (similar to how AWS's APIs are working):

```
type Wrapper struct {
client athenaiface.AthenaAPI
s3ResultsPath *string // this can be nil, to use defaults
}

type Result struct {
client athenaiface.AthenaAPI
queryResult   *athena.GetQueryResultsOutput
	// internal state
startResult *athena.StartQueryExecutionOutput
}

func (a *Wrapper) Run(database, sql string) (AthenaResult, error) {
}

func (r *Result) Wait() ( error) {
var executionInput athena.GetQueryExecutionInput
executionInput.SetQueryExecutionId(*r.startResult.QueryExecutionId)

var executionOutput *athena.GetQueryExecutionOutput

	for {
...
}

```

So people could just do sth like 
```
client := NewAthenaWrapper(session, outputPath)
result1, err := client.Run("database", "select...")
result2, err := client.Run("database2", "select...")
result1.Wait()
result2.Wait()
```
Thoughts?

tools/mage/deploy.go
kostaspap: Thought for later : Might be worth creating different workgroups to isolate queries we execute e.g. periodically vs queries that will execute as part of some user action.

tools/athenaviews/athenaviews_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/panther-labs/panther/pkg/awsglue"
+)
+
+func TestGenerateViewAllLogs(t *testing.T) {
```
kostaspap: Worth adding a test to verify that we generate proper SQL that works with multiple tables that includes the `union all` clause -since this is the main usecase.

pkg/awsathena/query.go
```diff
@@ -0,0 +1,102 @@
...
+	executionInput.SetQueryExecutionId(*aq.startResult.QueryExecutionId)
+
+	var executionOutput *athena.GetQueryExecutionOutput
+
+	for {
+		executionOutput, err = aq.Client.GetQueryExecution(&executionInput)
+		if err != nil {
+			return errors.Wrapf(err, "athena failed running: %#v", *aq)
+		}
+		if *executionOutput.QueryExecution.Status.State != "RUNNING" {
```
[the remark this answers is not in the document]
rleighton: Done

pkg/awsathena/query.go
```diff
@@ -0,0 +1,102 @@
...
+			break
+		}
+		time.Sleep(pollDelay)
+	}
+
+	if *executionOutput.QueryExecution.Status.State == "SUCCEEDED" {
```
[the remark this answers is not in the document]
rleighton: Done

tools/athenaviews/athenaviews.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// generateViewAllLogs creates a view over all log sources using "panther" fields
+func generateViewAllLogs(tables []*awsglue.GlueMetadata) (sql string) {
+	columns := gluecf.InferJSONColumns(parsers.PantherLog{}, gluecf.GlueMappings...)
+	var selectColumns []string
+	for _, col := range columns {
+		selectColumns = append(selectColumns, col.Name)
+	}
+	selectClause := strings.Join(selectColumns, ",") + ",year,month,day,hour"
```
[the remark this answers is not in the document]
rleighton: Parameterized to future proof.

pkg/awsathena/query.go
rleighton: I don't want to do that now .. this complexity which is not currently used in the PR. I suspect that even when we use this for CTAS that will not be need. If we need it we can add it but let's wait until it actually used.

tools/mage/deploy.go
rleighton: Easy enough to do later.

tools/athenaviews/athenaviews_test.go
rleighton: Done

tools/athenaviews/athenaviews.go
rleighton: Done

pkg/awsathena/query.go
kostaspap: Ok, I made that comment because this is in the `pkg` folder thus mean to be used by all of our code that wants to run queries against Athena and seemed inefficient. Agreed that is not urgent to fix.

Issue #67, opened 2020-01-29
Deploying to potentially unsupported region - 


gavinelder:
**Describe the bug**

When attempting to deploy into AWS-CN partition I believe I am running into an error creating a Simple email service subscription , I don't believe this service is available. 

Are you aware of required services or Topics required at this stage of the deploy ?

```shell 
> mage -debug deploy
-----
-----
/python-analysis.zip exists and is up to date
deploy: uploading ACM certificate
Error: SubscriptionRequiredException: The AWS Access Key Id needs a subscription for the service
	status code: 400, request id: 48c98adb-c18a-40cb-b335-e828445fac31
```` 

**To Reproduce**
1. Deploy within AWS-Cn

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - Within the docker image on mac. 

**Additional context**

gavinelder:
Actually I believe it may be Amazon certificate manager which is missing.

austinbyers:
Good find! What region exactly are you deploying to so we can reproduce it?

Thankfully, AWS has a [nice table](https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/) showing which services are supported in which regions. AWS certificate manager is supported in all regions, including gov regions and asia-pacific, except Beijing and Ningxia.

We should do a full audit of all of our AWS services and which regions they won't work in and update our documentation accordingly. Certificate manager is used to enable HTTPS on the web app, but if that is indeed the issue we could downgrade to HTTP so you can at least get a deployment up and running.

gavinelder:
Hi @austinbyers sadly the AWS-CN partition consists of Beijing and Ningxia so aws-cn support is probably reliant on allowing the customer to bring their own SSL cert or downgrading to http. 

Let me know what the outcome is.

nhakmiller:
Thanks for reporting this issue, @gavinelder . It looks the [AWS recommended workaround](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_server-certs.html) in cases where ACM is not supported is to use IAM for certificate management.

I just did a quick test on my local deployment, and was able to trivially use the same certificate and key files we use to import to ACM to import to IAM instead. Then switching the load balancer over to the IAM certificate worked with no issue.

I will see about updating the deploy tool to handle this case and fall back to IAM certificate management when ACM is not available.

nhakmiller:
[...]

PR #69, opened 2020-01-29
Add IAM fallback for ACM certificate import

nhakmiller:
## Background

@gavinelder reported that the deploy process fails when deployed in either of the `cn` regions at the upload ACM certificate step. Further investigation by @austinbyers showed that ACM is in fact not supported in those regions.

This change modifies the deploy process to fallback to uploading a certificate to IAM in cases where the ACM service is not available.

Closes: #67 

## Changes

- Added a check to see if the ACM service is available
- When the ACM service is not available, upload the self signed certificate to IAM instead of ACM

## Testing

- None yet, looking into setting up an account in `cn`

### DIFF
--- a/tools/mage/deploy_certificate.go
+++ b/tools/mage/deploy_certificate.go
@@ -128,10 +128,12 @@ func uploadLocalCertificate(awsSession *session.Session) (string, error) {
 // uploadIAMCertificate creates an IAM certificate resource and returns its ARN
 func uploadIAMCertificate(privateKeyBytes, certificateBytes []byte, session *session.Session) (string, error) {
 	iamClient := iam.New(session)
+	t := time.Now()
+	certName := "PantherCertificate-" + t.Format("2006-01-02T15-04-05")
 	input := &iam.UploadServerCertificateInput{
 		CertificateBody:       aws.String(string(certificateBytes)),
 		PrivateKey:            aws.String(string(privateKeyBytes)),
-		ServerCertificateName: aws.String("PantherCertificate"),
+		ServerCertificateName: aws.String(certName),
 	}
 	output, err := iamClient.UploadServerCertificate(input)
 	if err != nil {
--- a/tools/mage/deploy_certificate.go
+++ b/tools/mage/deploy_certificate.go
@@ -95,7 +95,7 @@ func uploadLocalCertificate(awsSession *session.Session) (string, error) {
 
 	// Check if the ACM service is supported before tossing the private key out into the ether
 	acmClient := acm.New(awsSession)
-	_, err = acmClient.ListCertificates(&acm.ListCertificatesInput{MaxItems: aws.Int64(0)})
+	_, err = acmClient.ListCertificates(&acm.ListCertificatesInput{MaxItems: aws.Int64(1)})
 	if err != nil {
 		if awsErr, ok := err.(awserr.Error); ok {
 			if awsErr.Code() == "SubscriptionRequiredException" {
--- a/tools/mage/deploy_certificate.go
+++ b/tools/mage/deploy_certificate.go
@@ -36,6 +36,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/acm"
+	"github.com/aws/aws-sdk-go/service/iam"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


tools/mage/deploy_certificate.go
```diff
@@ -90,6 +93,20 @@ func uploadLocalCertificate(awsSession *session.Session) (string, error) {
...
 	}
 
+	// Check if the ACM service is supported before tossing the private key out into the ether
+	acmClient := acm.New(awsSession)
+	_, err = acmClient.ListCertificates(&acm.ListCertificatesInput{MaxItems: aws.Int64(0)})
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok {
+			if awsErr.Code() == "SubscriptionRequiredException" {
+				// ACM is not supported in this region or for this user, fall back to IAM
+				fmt.Println("deploy: ACM not supported, falling back to IAM for certificate management")
```
austinbyers: Should we always use IAM? What's the difference?

gavinelder:
I will look at testing this, in order to get a AWS-CN account you require a Chinese residence & if it's a corporation a CN National.

tools/mage/deploy_certificate.go
nhakmiller: AWS recommends always using ACM for certificate management when it's available. That's the main reason we defaulted to it. ACM offers many more certificate management features like automatic renewal etc. but users most likely won't be taking advantage of those features for the self-signed certificates we upload for them.

The only other reason I would prefer to use ACM by default is that you cannot manage certificates imported via IAM through the AWS console. So if user's decide to setup their own certificate, they will have to use the CLI/SDK to delete the self signed certificate we uploaded. Of course they don't need to delete the certificate we uploaded if they don't want to, they can replace it with their own ACM/IAM certificate and just let the self signed IAM certificate we created sit there unused.

With that in mind, would you prefer to axe ACM altogether?

nhakmiller:
> I will look at testing this, in order to get a AWS-CN account you require a Chinese residence & if it's a corporation a CN National.

Ah well that will be difficult for me to do as I have neither a Chinese residence nor am I a CN National. In this case I will leave testing of the issue to you (or any other users active in the CN partition), and feel free to post here with any issues. We're also on [gitter](https://gitter.im/runpanther/community) for IM.

gavinelder:
@nhakmiller  This may need some more work,  


```
deploy: uploading load balancer certificate
deploy: WARNING no ACM certificate ARN provided and no certificate file provided, generating a self-signed certificate
Error: InvalidParameter: 1 validation error(s) found.
- minimum field value of 1, ListCertificatesInput.MaxItems.
```

nhakmiller:
> deploy: uploading load balancer certificate

Interesting, the aws CLI has no issues with that parameter, I suppose the go SDK has more strict requirements than the python SDK. I've just upped that to a minimum of 1, although we may just remove ACM entirely and use IAM by default.

gavinelder:
@nhakmiller it's passed now . 

```deploy: uploading load balancer certificate
deploy: ACM not supported, falling back to IAM for certificate management
deploy: panther-app: ExecuteChangeSet: CREATE_IN_PROGRESS
deploy: panther-app: ExecuteChangeSet: ROLLBACK_IN_PROGRESS```

**Comment by gavinelder** [comment]
```
deploy: ACM not supported, falling back to IAM for certificate management
Error: EntityAlreadyExists: The Server Certificate with name PantherCertificate already exists.
	status code: 409, request id: 4d1021db-6e2a-4d22-9982-5353c7ec30ae
```

nhakmiller:
> ```
> deploy: ACM not supported, falling back to IAM for certificate management
> Error: EntityAlreadyExists: The Server Certificate with name PantherCertificate already exists.
> 	status code: 409, request id: 4d1021db-6e2a-4d22-9982-5353c7ec30ae
> ```

Thanks for your help and patience. I'm going to assume that this occurred when you tried to deploy Panther in the same account in a different region, or you tried re-deploying Panther before it finished?

In either case, this latest version should fix that particular issue I believe. I apologize for sending you so much untested code, I will spend some time testing the IAM functionality separate from the CN partition tomorrow.

gavinelder:
The failure was linked to invalid service principles in the cloud formations I am currently working through the errors 1 by 1 as the documentation does not marry up 100% with implementation in CN

nhakmiller:
> The failure was linked to invalid service principles in the cloud formations I am currently working through the errors 1 by 1 as the documentation does not marry up 100% with implementation in CN

I see. Is it safe to say at this point you are experiencing no issues with certificates during deployment (granted that the deployment isn't finishing for other reasons)? If so, I will probably merge this.

Additionally, I'd recommend you run `aws iam list-server-certificates` from the aws-cli, and delete old certificates with `aws iam delete-server-certificate` after you get it to work as every failed deploy is creating a new certificate (because the deploy is failing, so it is unable to find the old one even though it exists).

nhakmiller:
I'm going to merge this now, as it sounds like any persisting deployment issues are not related to this certificate generation.

PR #71, opened 2020-01-30
Remove organization onboarding and users-api dead code

austinbyers:
## Background
We used to need a step function for onboarding new accounts in a multi-tenant model. This step function is no longer used and can be completely removed.

## Changes

- Remove `organization-onboarding` step function, API gateway, and Lambda function
    - Less stuff to deploy! 🎉 
- Remove duplicate `CognitoSNSRole` from `deployments/admin_api.yml` (it's already created in `deployments/cognito.yml`)
    - @3nvi @drixta this role is just to enable SMS 2FA, which we don't even support. Can we remove it completely?
    - EDIT: I just went ahead and deleted it and was still able to deploy and login with a new user
- Remove unnecessary permissions and dead code from `users-api`
- `go mod tidy`
- Get rid of unused `__init__.py` file

## Testing

Fresh deploy to dev account and I was able to change my password, enable 2FA, login, view policies, and update my username


austinbyers:
@kostaspap there actually isn't anything in GraphQL or Appsync for this

PR #72, opened 2020-01-30
Updated documentation to include Nginx logs

kostaspap:
## Background

Updating documentation to include Nginx logs

## Changes

- Updated documentation to include Nginx logs
- Renamed `nginx` packages to `nginxlogs` package, keeping the same format as we do with `awslogs` and `osquerylogs`. 

## Testing

- `mage test:ci`

### DIFF
--- a/api/gateway/analysis/client/operations/bulk_upload_parameters.go
+++ b/api/gateway/analysis/client/operations/bulk_upload_parameters.go
@@ -31,6 +31,7 @@ import (
 	"github.com/go-openapi/errors"
 	"github.com/go-openapi/runtime"
 	cr "github.com/go-openapi/runtime/client"
+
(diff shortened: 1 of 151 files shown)


kostaspap:
> As part of the documentation of this data source, can you add a "description" tag to each value on the Access struct . This will appear in the Athena when you hover over column and will be in the available in the schema view in the Glue catalog as documentation. For example:
> 
> ```
> RemoteAddress *string            `json:"remoteAddr,omitempty" description:"The IP address communicating with the server"`
> ```
> 
> We need to go back in a "clean up" PR and add descriptions to all data sources fields. I'll take that next sprint, it will be quick. But from now on all new data sources should have description tags added.

Good call, added. Yup, probably after making the PR you mention we can add some build-time check that the `description` tag is populated.

PR #74, opened 2020-01-30
When listing alerts and assoc rule is not found set severity to unknown

rleighton:
When listing alerts and assoc rule is not found set alert severity: Unknown (rule deleted) in order to let use know that rule is delete. Before this change it was very unclear which rules were current and severity defaulted.

## Testing
Manually in dev env using UI.


rleighton:
I spoke with Aggelos and we will leave the current behavior of null and he will handle this on the UI end.

PR #75, opened 2020-01-30
[ENG-1358][ENG-1366] Bugfixes: Alert event pagination & Alert with deleted rule

3nvi:
## Background

This PR resolves the issue where the entirety of the events of an alert are being fetched on a single roundtrip. Up until now, not only the mocked pagination that was implemented was wrong, but the API parameters have changed without the front-end adapting.

This PR makes sure that events for an alert are received in batches of 25 and that we prefetch 1 page in advance, in order to avoid lag/delay during event navigation. To achieve that, we had to split a single query into two separate ones so that we don't re-fetch un-needed data every time we request a new page of events.

Lastly, after this has been done, this PR also made sure that the related pages won't crash whenever the rule associated to an alert, gets deleted.

## Changes

* Isolate the `alert` from its related `rule` in AppSync
* Split `rule` and `alert` into two distinct queries
* Add pagination logic in the alert details component
* Bump `@apollo/client`
* Add cache normalization keys for `AlertDetails` and `AlertSummary`
* Fix crashes of alert pages when the related rule has been deleted

## Testing

Locally & on personal AWS server


web/src/pages/list-sources/log-source-columns.tsx
```diff
@@ -44,7 +44,16 @@ const columns = [
   {
     key: 's3Buckets',
     header: 'S3 Buckets',
-    flex: '1 0 200px',
+    flex: '1 0 300px',
```
rleighton: Is this related to Alert pagination?

3nvi: Not at all. I just noticed a line-break issue and thought I could put it in the same PR. I'll remove it and put it in a separate one since it might create confusion

package.json
```diff
@@ -104,7 +104,7 @@
     "yaml-cfn": "^0.2.3"
   },
   "dependencies": {
-    "@apollo/client": "^3.0.0-beta.7",
+    "@apollo/client": "^3.0.0-beta.29",
```
austinbyers: Is there a reason we have to use a `beta` release?

3nvi: Yes

PR #78, opened 2020-01-30
Additional JIRA Destination Configuration

nhakmiller:
## Background

Previously the JIRA destination assumed that the organization using JIRA had Tasks enabled. This assumption was not true for our own JIRA deployment, and likely won't be true for all users that hope to take advantage of the JIRA destination.

This PR allows users to configure whether the JIRA destination should create a Task, Story, or Bug.

Enabled by frontend change made in https://github.com/panther-labs/panther/pull/59. 

## Changes

- Make the issue type configurable
- Caught an edge case where a user enters then deletes input in an optional field

## Testing

- Deployed in my dev account, onboarded our internal JIRA as a destination


internal/core/alert_delivery/outputs/jira.go
```diff
@@ -50,12 +50,12 @@ func (client *OutputClient) Jira(
...
 		},
-		"issuetype": map[string]string{
-			"name": "Task",
+		"issuetype": map[string]*string{
+			"name": config.Type,
 		},
 	}
 
-	if config.AssigneeID != nil {
+	if config.AssigneeID != nil && *config.AssigneeID != "" {
```
austinbyers:
```go
if aws.StringValue(config.AssigneeID) != "" {
    // ...
}
```

kostaspap:
This issue is not fixed yet, right? Has the UI been updated to allow users to specify the type?

nhakmiller:
> This issue is not fixed yet, right? Has the UI been updated to allow users to specify the type?

The issue is fixed, the UI was updated in https://github.com/panther-labs/panther/pull/59.

Issue #70, opened 2020-01-30
New Destination: Asana Tasks


jacknagz:
**Describe the solution you'd like**

Asana is a work management platform. We should add the ability to send Alerts to [Asana Tasks](https://developers.asana.com/docs/#asana-tasks).

**Additional context**

Tasks are assigned within Projects and categorized under Sections (aka columns).

<img width="1090" alt="Screen Shot 2020-01-29 at 4 41 38 PM" src="https://user-images.githubusercontent.com/11466941/73410333-bcf6e680-42b6-11ea-9f37-a1f6501574f6.png">

API Ref:
* [Create Task ](https://developers.asana.com/docs/#create-a-task)
* [Task Tags](https://developers.asana.com/docs/#add-a-tag-to-a-task)
* [Link Projects to Tasks](https://developers.asana.com/docs/#add-a-project-to-a-task)

3nvi:
Starting work on the front-end

PR #79, opened 2020-01-30
Adding Asana destination

kostaspap:
## Background

Related to https://github.com/panther-labs/panther/issues/70 . Supporting Asana as a destination. When a policy fails or a rule has a match, we will create a new Asana task in an Asana project. 
When users create an Asana destination, they need to add an access token and one (or many) projects that they want the task to associated with. 

Ref: https://developers.asana.com/docs/#create-a-task
## Changes

- Adding Asana as a destination
- Fixing an issue in the alert_delivery Lambda API - marking the `createdAt` field from an alert as required. It was a field always required by the code, just the API didn't have the appropriate mark. 
- Small cleanup here and there

## Testing

- Added unit test
- Created an Asana destination and fired a test rule. Seen a new task created in a project

Screen Shot 2020-01-30 at 7 09 57 PM

### DIFF
--- a/api/lambda/outputs/models/api.go
+++ b/api/lambda/outputs/models/api.go
@@ -261,7 +261,7 @@ type SqsConfig struct {
 // AsanaConfig defines options for each Asana output
 type AsanaConfig struct {
 	PersonalAccessToken *string   `json:"personalAccessToken" validate:"required,min=1"`
-	ProjectGids         []*string `json:"projectGids" validate:"required,min=1,dive,numeric"`
+	ProjectGids         []*string `json:"projectGids" validate:"required,min=1,dive,required"`
 }
 
 // DefaultOutputs is the structure holding the information about default outputs for severity
--- a/api/lambda/outputs/models/api.go
+++ b/api/lambda/outputs/models/api.go
@@ -260,9 +260,8 @@ type SqsConfig struct {
 
 // AsanaConfig defines options for each Asana output
 type AsanaConfig struct {
-	AccessToken *string `json:"accessToken" validate:"required,min=1"`
-	WorkspaceGid *string `json:"workspaceGid" validate:"required,min=1"`
-	ProjectGids []*string `json:"projectGids,omitempty"`
+	PersonalAccessToken *string   `json:"personalAccessToken" validate:"required,min=1"`
+	ProjectGids         []*string `json:"projectGids" validate:"required,min=1,dive,numeric"`
 }
 
 // DefaultOutputs is the structure holding the information about default outputs for severity
--- a/api/lambda/outputs/models/api.go
+++ b/api/lambda/outputs/models/api.go
@@ -202,6 +202,9 @@ type OutputConfig struct {
 
 	// SqsConfig contains the configuration for SQS alert output
 	Sqs *SqsConfig `json:"sqs,omitempty"`
+
+	// AsanaConfig contains the configuration for Asana alert output
+	Asana *AsanaConfig `json:"asana,omitempty"`
 }
 
 // SlackConfig defines options for each Slack output.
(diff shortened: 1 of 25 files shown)


api/lambda/outputs/models/api.go
```diff
@@ -254,6 +257,12 @@ type SqsConfig struct {
 	QueueURL *string `json:"queueUrl" validate:"required,url"`
 }
 
+// AsanaConfig defines options for each Asana output
+type AsanaConfig struct {
+	PersonalAccessToken *string   `json:"personalAccessToken" validate:"required,min=1"`
+	ProjectGids         []*string `json:"projectGids" validate:"required,min=1,dive,numeric"`
```
austinbyers: If they're numeric, should it be a `[]*int` instead?

Also, `nil` entries are possible, you need another `required` after the `dive`

`validate:"required,min=1,dive,required,numeric"`

internal/core/alert_delivery/models/api.go
```diff
@@ -30,7 +30,7 @@ const PolicyType = "POLICY"
 type Alert struct {
 
 	// CreatedAt is the creation timestamp (seconds since epoch).
-	CreatedAt *time.Time `json:"createdAt,omitempty"`
+	CreatedAt *time.Time `json:"createdAt" validate:"required"`
```
austinbyers: This is part of the API response, right? In which case, the `validate` tag is ignored. You can keep it if you like, but it doesn't enforce anything

internal/core/alert_delivery/outputs/jira.go
```diff
@@ -67,15 +67,13 @@ func (client *OutputClient) Jira(
 
 	auth := *config.UserName + ":" + *config.APIKey
 	basicAuthToken := "Basic " + base64.StdEncoding.EncodeToString([]byte(auth))
-	accept := "application/json"
```
austinbyers: Are you 100% sure we don't need the `application/json` mime type in the request headers for all of these outputs?

api/lambda/outputs/models/api.go
nhakmiller: Unless it is documented somewhere that they are numeric, I would not be surprised if GID's include hyphens, non-numeric characters, leading 0's, etc. I think `string` is correct. The Asana REST API documentation Kostas linked seems to agree: 

`projects [string]`

internal/core/alert_delivery/outputs/jira.go
nhakmiller: Since we only accept json back, I don't see the harm in specifying that to the server when making these requests. I would prefer not to remove these everywhere simply because all testing has been done with them in, unless there is a specific reason you think we need to remove them.

austinbyers: Agreed, don't see the reason to remove it

jacknagz:
Super excited for this, can we also make sure to add it to the docs?

api/lambda/outputs/models/api.go
kostaspap: I've put numeric because all the occurrences of Gid in the documentation and also in my tests showed that it is numeric. That is different from making it an integer though - converting it to an integer would remove any trailing zeros (e.g. same with AWS Account IDs). 

On second though, I'll remove the "numeric" check since, even though that's what's I've seen till now, it might be changed in the future.

internal/core/alert_delivery/models/api.go
kostaspap: Actually not. This is the alert_delivery API model. It is being checked here: https://github.com/panther-labs/panther/blob/master/internal/core/alert_delivery/main/lambda.go#L47-L49

internal/core/alert_delivery/outputs/jira.go
kostaspap: It's not removed - I just moved it here: https://github.com/panther-labs/panther/blob/asana/internal/core/alert_delivery/outputs/post.go#L46  .
The `post.go` code was setting as "application/json" the `Content-Type` header, but not the `Accept` header - that was done in every individual destination. I just made `Accept: application/json` the default :)

kostaspap:
> Super excited for this, can we also make sure to add it to the docs?

Yup, will add once frontend has added it as well.

PR #81, opened 2020-01-30
Added post processing to CloudFormation package output

nhakmiller:
## Background

@gavinelder reported an issue when trying to deploy to a region not specified in the `~/.aws/config` file, even when specifying the desired region with flags and environment variables.

From our investigation, it appears to be a bug in the `aws cloudformation package` command which we rely on.

This PR applies a layer of post processing on all templates created by the `aws cloudformation package` command which does some string parsing to re-construct the problematic URL.

## Changes

- Perform post processing on all templates generated by `aws cloudformation package` to re-structure the `TemplateURL` fields from this format:

`https://s3.region.amazonaws.com/bucket/panther-app/1.template`

to this format:

`https://bucket.s3.amazonaws.com/panther-app/1.template`
- updated `AWSXrayWriteOnlyAccess` to `AWSXRayDaemonWriteAccess` for all lambdas

## Testing

 - Created a session with AWS credentials in the `us-west-2` region, then ran this command: `export AWS_REGION=us-west-1 && mage deploy`
    - Also confirmed that this operation failed without my fix in place


tools/mage/deploy.go
```diff
@@ -270,10 +277,55 @@ func cfnPackage(templateFile, bucket, stack string) (string, error) {
...
 
+// Post-Process all the CFN templates to fix issues the aws cloudformation package command introduces
+func cfnPackagePostProcess(templatePath string) error {
+	templateOriginal, err := ioutil.ReadFile(templatePath)
+	if err != nil {
+		return err
+	}
+
+	var result []string
+	// Someone there is a bug that is causing the environment variables specifying region to not be properly respected
```
nhakmiller: Should be `Somewhere there is a bug...` fix staged locally, will push it up with any other PR feedback.

tools/mage/deploy.go
```diff
@@ -249,7 +250,13 @@ func uploadLayer(awsSession *session.Session, libs []string, bucket, key string)
...
 // Upload resources to S3 and return the path to the modified CloudFormation template.
 // TODO - replace this with our own to avoid relying on the aws cli
-func cfnPackage(templateFile, bucket, stack string) (string, error) {
+func cfnPackage(templateFile, bucket, stack string) (out string, err error) {
+	defer func() {
+		// Apply post processing when the package command succeeds
+		if err == nil {
+			err = cfnPackagePostProcess(out)
+		}
+	}()
```
austinbyers: I would just put this at the end of the function, no need for a defer. Then you can read top-to-bottom to follow the execution

tools/mage/deploy.go
```diff
@@ -270,10 +277,55 @@ func cfnPackage(templateFile, bucket, stack string) (string, error) {
...
+			// Build the new URI
+			prefixParts[1] = prefixParts[0]
+			prefixParts[0] = uriParts[1]
+
+			// Rebuild the line
+			newURIPrefix := strings.Join(prefixParts, ".")
+			newURIParts := append([]string{newURIPrefix}, uriParts[2:]...)
+			lineParts[1] = strings.Join(newURIParts, "/")
+			line = strings.Join(lineParts, "https://")
+		}
```
austinbyers: Let's move the string replacement to its own function so we can add unit tests

```go
func fixPackageTemplateURL(line string) string {
    line = strings.TrimSpace(line)
    if !strings.HasPrefix(line, "TemplateURL: ") {
        return line // not affected
    }

    // Break the line down to the pieces we need
    lineParts := strings.Split(line, "https://")
    // .... rest of logic ....
}
```

tools/mage/deploy.go
nhakmiller: Ok, but it's way more lines of code and way less cool :'(

tools/mage/deploy.go
nhakmiller: Will do, and the implied task of adding unit tests.

tools/mage/deploy.go
austinbyers: Haha it is less cool. But boring is simpler and easier to read in this case, in my opinion lol

kostaspap:
> ## Changes
> * Perform post processing on all templates generated by `aws cloudformation package` to re-structure the `TemplateURL` fields from this format:
> 
> `https://s3.region.amazonaws.com/bucket/panther-app/1.template`
> 
> to this format:
> 
> `https://bucket.s3.region.amazonaws.com/panther-app/1.template`
> 

Have we created an AWS CLI issue? Wondering how easy it would be to fix this upstream instead of us having to hack our way around it :)

kostaspap:
Can you add a comment to https://github.com/aws/aws-cli/issues/4372 with some details on the issue we encountered? We could get the discussion going there.

nhakmiller:
> Can you add a comment to [aws/aws-cli#4372](https://github.com/aws/aws-cli/issues/4372) with some details on the issue we encountered? We could get the discussion going there.

Sure I can add a comment, I don't know if you noticed but I did also link to that issue in a comment in the code for future reference.

PR #83, opened 2020-01-31
[ENG-1375] Optimise Git Commit hooks

3nvi:
## Background

Currently, front-end related git commit hooks run on every commit regardless if the associated changes should be checked or not. 

The core issue is the Typescript compiler which stalls the commit by ~6-7 seconds, even when there are no changes in TS files. In addition, people from the backend team mentioned that the auto-formatting of tools caused issues in their work. This PR attempts to resolve just that. 

## Changes

* Make sure to only run the Typescript compiler if there are changes to TS files 
* Make sure to auto-format only YAML and JSON files found:
    * on the project root dir
    * under the `/web` directory
* Make sure to **always** auto-format Markdown files regardless of where they are

Also:

* Fix linting issues in project root dir files
* Rename `.gitbooks.yaml` to `.gitbooks.yml` for consistency with the rest oof the project

## Testing

Locally


kostaspap:
> * Rename `.gitbooks.yaml` to `.gitbooks.yml` for consistency with the rest oof the project

Is it possible this will break our Github repo -> gitbooks syncing? Do we have to make any change there?

3nvi:
> Is it possible this will break our Github repo -> gitbooks syncing? Do we have to make any change there?

We don't, but it's the only `.yaml` file which forces me to add `.yaml` to the inspection tools. I figure that since lots of repos already have it as `gitbooks.yml`, then it would be ok.

Is there a way to verify it?

kostaspap:
> > Is it possible this will break our Github repo -> gitbooks syncing? Do we have to make any change there?
> 
> We don't, but it's the only `.yaml` file which forces me to add `.yaml` to the inspection tools. I figure that since lots of repos already have it as `gitbooks.yml`, then it would be ok.
> 
> Is there a way to verify it?

I actually checked the configuration, there doesn't seem to be any specific configuration option that requires the name of the file to be `gitbooks.yaml`. I think it is fine to merge.

3nvi:
@austinbyers Do you want me to include `deployments` folder in the git commit hooks?

austinbyers:
Not yet, I think we need more discussion as a team first

PR #85, opened 2020-01-31
[Docs] Update web docs

3nvi:
## Background

Fix small inconsistencies in the web's README 

## Changes

* Update README.md

## Testing

None required

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -16,8 +16,8 @@ $ npm i
 ### Development
 
 In order to develop locally, you must first have the project deployed on your AWS account. If you've successfully done that,
-then a `.env` file will have appeared in the `/out` directory, which will hold crucial ENV vars that
-are needed for local development. With this file present in `/out/.env`, all you need to do is run:
+then a `.env` file will be present in the `/out` directory, which holds the ENV vars that
+are needed for local development. With this file present in `/out/.env`, all you need to do is run (from the panther root directory):
 
 ```
 npm run start
--- a/web/README.md
+++ b/web/README.md
@@ -16,8 +16,8 @@ $ npm i
 ### Development
 
 In order to develop locally, you must first have the project deployed on your AWS account. If you've successfully done that,
-then a `.env` folder will have appeared in the project root which will hold crucial ENV vars that
-is needed for local development. If this file is present, then all you need to do is run:
+then a `.env` file will have appeared in the `/out` directory, which will hold crucial ENV vars that
+are needed for local development. With this file present in `/out/.env`, all you need to do is run:
 
 ```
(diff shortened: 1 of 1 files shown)


web/README.md
```diff
@@ -16,8 +16,8 @@ $ npm i
 ### Development
 
 In order to develop locally, you must first have the project deployed on your AWS account. If you've successfully done that,
-then a `.env` folder will have appeared in the project root which will hold crucial ENV vars that
-is needed for local development. If this file is present, then all you need to do is run:
+then a `.env` file will have appeared in the `/out` directory, which will hold crucial ENV vars that
+are needed for local development. With this file present in `/out/.env`, all you need to do is run:
```
rleighton: maybe say "all you need to do is run from the panther root directory" (I assume it matters what dir you are in, right?)

web/README.md
```diff
@@ -32,7 +32,7 @@ This package shouldn't have to be deployed individually, since its deployment is
 of `mage deploy` which exposes a publicly available URL for you to access the UI.
 
 There are situations though that may force you to have a local deployment (e.g. firewall restrictions)
-of the interface. To do that, all you need to do is build the web project and then serve its HTML through
+of the Panther interface. To do that, all you need to do is build the web project and then serve its HTML through
 a lightweight web server. Run:
```
rleighton: Again, perhaps make the directory clear if it matters.

web/README.md
3nvi: Higher up in the same readme, there is the following text:

> Due to multiple package management, please consider the project root as the current working directory for the following document sections

web/README.md
3nvi: same answer :)

web/README.md
3nvi: Do you feel that we should be more explicit?

austinbyers: Doesn't hurt to mention it again occasionally, especially right before listing an example command. People tend to jump straight to the code examples, so having a reminder about where to run it could be helpful

web/README.md
```diff
@@ -16,8 +16,8 @@ $ npm i
 ### Development
 
 In order to develop locally, you must first have the project deployed on your AWS account. If you've successfully done that,
-then a `.env` folder will have appeared in the project root which will hold crucial ENV vars that
-is needed for local development. If this file is present, then all you need to do is run:
+then a `.env` file will be present in the `/out` directory, which holds the ENV vars that
+are needed for local development. With this file present in `/out/.env`, all you need to do is run (from the panther root directory):
```
austinbyers: perfect! love it

PR #88, opened 2020-01-31
Std fields

rleighton:
## Changes
* panther fields added to all log parsers
* added ability to extract account ids and instance ids from ARNs to aws extractor

## Testing
* mage test:unit
* mage test:integration
* manual testing in dev account processing log files


internal/log_analysis/log_processor/parsers/awslogs/extractor.go
```diff
@@ -42,6 +43,30 @@ func (e *AWSExtractor) Extract(key, value gjson.Result) {
 	// value based matching
 	if strings.HasPrefix(value.Str, "arn:") {
 		e.pl.AppendAnyAWSARNs(value.Str)
```
kostaspap: 1. In line 54 we verify that the string is indeed an ARN. We should thus invoke the `e.pl.AppendAnyAWSARNs` only after we have verified that this is indeed an ARN.
2. Optionally, the `AppendAnyAWSArns` method should itself invoke the `AppendAnyAWSAccountIds` - if we add an ARN and that ARN has an AccountID in it, we need to add it. It will make the code less error-prone.

internal/log_analysis/log_processor/parsers/awslogs/alb_test.go
```diff
@@ -204,11 +224,32 @@ func TestHTTPSNoTarget(t *testing.T) {
...
+
+func checkALBLog(t *testing.T, log string, expectedEvent *ALB) {
+	parser := &ALBParser{}
+	events := parser.Parse(log)
+	require.Equal(t, 1, len(events))
+	event := events[0].(*ALB)
+
+	// rowid changes each time
+	require.Greater(t, len(event.PantherRowID), 0)                      // ensure something is there.
+	require.NotEqual(t, event.PantherRowID, expectedEvent.PantherRowID) // ensure they are not same
```
kostaspap: It seems more appropriate to have a check like: 
```
require.NotNil(t, event.PantherRowID)
```

Line 230 can be removed - there is no real reason from the ALBParser's perspective why if the generated UUID cannot be "1234".

internal/log_analysis/log_processor/parsers/awslogs/alb_test.go
```diff
@@ -204,11 +224,32 @@ func TestHTTPSNoTarget(t *testing.T) {
...
 }
+
+func checkALBLog(t *testing.T, log string, expectedEvent *ALB) {
+	parser := &ALBParser{}
+	events := parser.Parse(log)
+	require.Equal(t, 1, len(events))
+	event := events[0].(*ALB)
+
+	// rowid changes each time
+	require.Greater(t, len(event.PantherRowID), 0)                      // ensure something is there.
```
kostaspap: If the `len(event.PantherRowID)` was 0, then parsing should have failed already. We should change the parser code so that setting the RowID is a requirement.

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
```diff
@@ -40,17 +40,20 @@ type VPCFlow struct {
 	Version     *int               `json:"version,omitempty" validate:"required"`
 	Account     *string            `json:"account,omitempty" validate:"omitempty,len=12,numeric"`
 	InterfaceID *string            `json:"interfaceId,omitempty"`
-	SourceAddr  *string            `json:"sourceAddr,omitempty"`
-	Dstaddr     *string            `json:"dstAddr,omitempty"`
+	SrcAddr     *string            `json:"srcAddr,omitempty"`
+	DstAddr     *string            `json:"dstAddr,omitempty"`
 	SrcPort     *int               `json:"srcPort,omitempty" validate:"omitempty,min=0,max=65535"`
-	DstPort     *int               `json:"destPort,omitempty" validate:"omitempty,min=0,max=65535"`
+	DstPort     *int               `json:"dstPort,omitempty" validate:"omitempty,min=0,max=65535"`
```
kostaspap: Thanks for fixing this!

internal/log_analysis/log_processor/parsers/osquerylogs/batch.go
```diff
@@ -56,15 +56,12 @@ func (p *BatchParser) Parse(log string) []interface{} {
 	event := &Batch{}
 	err := jsoniter.UnmarshalFromString(log, event)
 	if err != nil {
-		zap.L().Debug("failed to unmarshal log", zap.Error(err))
 		return nil
 	}
 
-	tsa, _ := jsoniter.MarshalToString(event)
-	fmt.Println(tsa)
+	event.updatePantherFields(p)
```
kostaspap: Looking into this, I am thinking it makes little sense to have fields like "any AWS ARNs" in such logs that are non-AWS. 
The "AWS" standard fields should really be only inside the AWS log types and not in these logs. While some columns should be common for all of our log types (IPs, Domains, etc), AWS-specific columns makes more sense to be only in the Glue tables that contain AWS data. 
I think it makes sense to have a "PantherLog" but also potentially "PantherAWSLog" that inherits from `PantherLog` and adds the AWS-specific fields. Thoughts?

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
```diff
@@ -136,3 +147,73 @@ func (p *CloudTrailParser) Parse(log string) []interface{} {
...
+				}
+				if event.UserIdentity.SessionContext.SessionIssuer.Arn != nil {
+					event.AppendAnyAWSARNs(*event.UserIdentity.SessionContext.SessionIssuer.Arn)
+				}
+			}
+		}
+	}
+
+	// polymorphic (unparsed) fields
+	awsExtractor := NewAWSExtractor(&(event.PantherLog))
```
kostaspap: It actually seems that the AWSExtractor should be changed to 
```
type AWSExtractor struct {
	pl parsers.PantherLog
}

func NewAWSExtractor(pl parsers.PantherLog) *AWSExtractor {
	return &AWSExtractor{pl: pl}
}
```

Then here you would do
```
awsExtractor := NewAWSExtractor(event.PantherLog)
```

There extractor shouldn't get a 'nil' log anyway.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
```diff
@@ -136,3 +147,73 @@ func (p *CloudTrailParser) Parse(log string) []interface{} {
...
+		extract.Extract(result, awsExtractor)
+	}
+
+	if event.ResponseElements != nil {
+		result := gjson.Parse(string(*event.ResponseElements))
+		extract.Extract(result, awsExtractor)
+	}
+
+	if event.ServiceEventDetails != nil {
+		result := gjson.Parse(string(*event.ServiceEventDetails))
```
kostaspap: 1. You could replace the `Parse` with `ParseBytes` method - then you don't need casting
```
result := gjson.ParseBytes(*event.ServiceEventDetails)
```
2. It seems this is a repeated pattern across all codebase. 
You could just modify the Extract function to sth like: 
```
// Extract walks parsed JSON extracting tasty things
func Extract(rawMessage jsoniter.RawMessage, extractors ...Extractor) {
	result := gjson.ParseBytes(rawMessage)
	extractInternal(result, extractors...)
}

func extractInternal(result gjson.Result, extractors ...Extractor) {
	result.ForEach(func(key, value gjson.Result) bool {
		for _, extractor := range extractors {
			extractor.Extract(key, value)
		}
		if value.IsArray() || value.IsObject() {
			extractInternal(value, extractors...)
		}
		return true // keep iterating
	})
}

```
That way you avoid duplication in the code.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
```diff
@@ -105,8 +112,8 @@ type CloudTrailSessionContextSessionIssuer struct {
 
 // CloudTrailSessionContextWebIDFederationData contains Web ID federation data
 type CloudTrailSessionContextWebIDFederationData struct {
-	FederatedProvider *string     `json:"federatedProvider,omitempty"`
-	Attributes        interface{} `json:"attributes,omitempty"`
+	FederatedProvider *string              `json:"federatedProvider,omitempty"`
+	Attributes        *jsoniter.RawMessage `json:"attributes,omitempty"`
```
kostaspap: So jsoniter.RawMessage is actually a byte array. Does it get properly serialized when we store it to S3? Is it easy to interact with this data in Athena if the type is `BINARY`?

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -66,6 +66,9 @@ type ALB struct {
 	ActionsExecuted        []string           `json:"actionsExecuted,omitempty"`
 	RedirectURL            *string            `json:"redirectUrl,omitempty"`
 	ErrorReason            *string            `json:"errorReason,omitempty"`
+
+	// NOTE: added to end of struct to allow expansion later
```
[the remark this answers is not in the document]
rleighton: Then the schema column order would change and we would need to update all existing partitions (otherwise Athena will complain about a schema incompatibility due to the fact Athena compares using positional indexed for JSON , but not Parquet, BTW.  There is a comment in the panther_log.go file with the reference:  https://github.com/awsdocs/amazon-athena-user-guide/blob/master/doc_source/updates-and-partitions.md). I think we will need to build a utility to do that, because it will happen.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -151,6 +154,8 @@ func (p *ALBParser) Parse(log string) []interface{} {
 		ErrorReason:            parsers.CsvStringToPointer(record[24]),
 	}
 
+	event.updatePantherFields(p)
```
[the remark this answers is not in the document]
rleighton: No, because the "required" fields are tagged as "required" and we want that checked to avoid mistakes. I don't understand this comment "if we tried to parse a VPC FL event with the CloudTrail parser, it would fail at that step", can you explain?

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -163,3 +168,24 @@ func (p *ALBParser) Parse(log string) []interface{} {
...
 	return "AWS.ALB"
 }
+
+func (event *ALB) updatePantherFields(p *ALBParser) {
+	if event.Timestamp != nil {
+		event.SetRequired(p.LogType(), *event.Timestamp)
+	}
+	if event.ClientIP != nil {
+		event.AppendAnyIPAddresses(*event.ClientIP)
+	}
```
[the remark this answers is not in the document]
rleighton: True , but some place we do not have ptrs, an I always hate to do the cast there. Would object to having 2 methods for each "any", for example:

```
func (pl *PantherLog) AppendAnyIPAddresses(values ...string) { ...

func (pl *PantherLog) AppendAnyIPAddressPtrs(values ...*string) { ...
```

I think would go a long way making the calling code prettier.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -163,3 +168,24 @@ func (p *ALBParser) Parse(log string) []interface{} {
 func (p *ALBParser) LogType() string {
 	return "AWS.ALB"
 }
+
+func (event *ALB) updatePantherFields(p *ALBParser) {
+	if event.Timestamp != nil {
+		event.SetRequired(p.LogType(), *event.Timestamp)
+	}
```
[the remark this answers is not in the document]
rleighton: The validation is done after this and the validate:"required" tag catches this.

internal/log_analysis/log_processor/parsers/awslogs/extractor.go
rleighton: I disagree, I'd like to catch any malformed ARNS ... from a security POV this would be important and why I did this in this manner.  For 2) , isn't it possible to have an ARN w/out an account id? Can it be '*' ? Not sure.

internal/log_analysis/log_processor/parsers/osquerylogs/batch.go
rleighton: Hmmm... good point. Easy to factor the in the Go code. I do like this idea. The view generation code will need to be smarter tho, we would need it to generate all possible fields in the view (adding nulls for fields there where the panther fields i not present).  Let me muddle this suggestion.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
rleighton: That forces a copy of the struct each time, right? What's the benefit?

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
rleighton: I like.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
rleighton: Nope, see: https://github.com/panther-labs/panther/blob/master/tools/cfngen/gluecf/schema.go#L108

... and : https://github.com/panther-labs/panther/blob/master/tools/cfngen/gluecf/gluecf.go#L40

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
kostaspap: I thought we had only pointers, which fields are not pointers?
We should just stick to pointers for fields of outputs of parsers. The output gets serialized and we never want to have the "default" value for such fields - if a field is not present it should be 'nil'. 

I think the code for the parsers would be just simpler if we went with the the convention we have until now. 
I think it is good practice not to use pointers in all other places in the code, but honestly, the code for parsers is just simpler if we stick to that convention. I think in this case trying to support both pointers and plain values will make the code more complex instead of simpler.

internal/log_analysis/log_processor/parsers/awslogs/alb_test.go
rleighton: Actually, it does already because of the tags.

internal/log_analysis/log_processor/parsers/awslogs/alb_test.go
rleighton: But PantherRowID is a string and can't be nil (tagged as required), so that check does not matter. That's why I did length, because we want _something_.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
rleighton: The deeper structures (after first level) are usually not pointers.

internal/log_analysis/log_processor/parsers/awslogs/extractor.go
kostaspap: 2) Yup, e.g. S3 bucket arns don't have AccountID
1) I think the "any" fields is better that they have the contents they advertise rather than malformed ones. Is "arn:" a malformed ARN? What about the string "ar"? 
Similarly, is "192.100.0" a malformed IP? What about "192"? Is "0" a malformed AWS Account ID?

I expect that we can have malformed data coming from the log source, but in this case we are the ones doing the extraction and making assumptions about fields. It is really hard to reason if an ARN was really malformed from the source or we just truncated by mistake as part of the extraction process.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
kostaspap: VPC FL was a wrong example. But let's say, we receive OsQuery Status event. The osQuery.Batch parser runs against the event. jsoniter will succeed parsing the event (they are both JSON). But it is the validation process that will determine if the event is Batch or Status type because the output the jsoniter parsing will not have the required fields.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
kostaspap: It seems wasteful to do all this logic though. If event.Timestamp is nil, we already no there is no reason to proceed. 
Moreover, given we do the validation first (see other comment) we could just do `event.SetRequired(p.LogType(), *event.Timestamp)` - the timestamp will always be populated unless sth went really bad in the code and we didn't perform the validation first - a scenario that a unit test would easily capture. 
Code is faster and also more compact if we go that way.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
rleighton: Done.

internal/log_analysis/log_processor/parsers/awslogs/extractor.go
rleighton: For normal sw this is an easy call (only take clean data), for security software it gives me pause.  This code is the first time we actually interpret the data (previously it was all pass-thru). In most ETL systems there is a side channel to collect things that look odd or wrong. I'm wondering if we need that but I'm uncertain how to design that here .

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
kostaspap: Uff, I don't realize that. In that case, the way you do it currently is fine.
Another option: Have the "AppendAnyIPAddresses" method check for "" and skip such entries. Invoke it from code like `event.AppendAnyIPAddresses(aws.StringValue(event.ClientIP)`. 
Or, modify `AppendAnyIPAddresses` to accept pointers. Use `&` or `aws.String()` when invoking it. 
Not sure which I like more, your call.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
rleighton: But then the validity check will fail on the panther required fields. I think your proposal  is much less clear actually ... the code currently reads: parse + append/more-parsing + validate. You are suggesting: parse + validate + append/more-parsing + re-validate. I don't see the advantage. Also, I suspect we will move away from autodetecting file types as the dominant mode of processing as we discussed ... pretty soon we will need to force users to tell us the type of data to disambiguate the data so can deal with different sources using common log file formats.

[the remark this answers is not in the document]
kostaspap: My suggestion was actually:  parse + validate + append/more-parsing. No need to re-validate since any fields we have added in the "append" process is under our control.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
rleighton: All of this is due to the "auto-detection" mode ... (see above comment), that is going away. If In light of that , I think this is much easier to read and micro optimizations to worry about a few if tests when we are parsing JSON.

internal/log_analysis/log_processor/parsers/osquerylogs/batch.go
rleighton: This was a good idea. I did the refactor and fixed the view generation. The view generation now has a place for us to handle the fields that are "view only" (I've not implemented but I left a comment). When we are ready, we can define some fields for just the view and use tags to describe the mapping and then the view generation needs only a small change to support that.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
rleighton: Sorry I'm not being clear and I'm now confused :) ... let's start over.

The required Panther fields are declared:

```
//  required
PantherLogType   string            `json:"p_log_type,omitempty" validate:"required" description:"Panther added field with type of log"`
PantherRowID     string            `json:"p_row_id,omitempty" validate:"required" description:"Panther added field with unique id (within table)"`
PantherEventTime timestamp.RFC3339 `json:"p_event_time,omitempty" validate:"required" description:"Panther added standardize event time (UTC)"`
```

These are use via composition, so if you do not set them and run validation the validation will fail.

If we take out the "required" tags so you can skip validation we lose an important constraint. 

This why the code does: parse + append + validate.

Make sense? Now what are you proposing and how do we ensure required fields are set?

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
rleighton: Let's leave it for now. Can be changed later with little penalty.

internal/log_analysis/log_processor/parsers/awslogs/extractor.go
rleighton: Since this is such a thorny issue, I'll just pull in the appending the ARN inside the parse (so it will always well formed). The issue of sketchy data is a big one and deserves more thought. We need to decide our semantics ... for example are we extracting "what is there" or "validated data that is there", right now it is a mix, which I think is ok for the short term.

rleighton: Updated.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
kostaspap: The issue here is not an issue of micro-optimization. Is a logical issue. The above line is the equivalent of 
```
someValue, err := myMethod()
if err == nil {
... // do sth
}
Do more stuff
```
The fact that the "SetRequired" is set when a condition is met, makes me wonder if they are required at all.

internal/log_analysis/log_processor/parsers/awslogs/alb_test.go
kostaspap: Got it. Yeah, Makes sense to keep the `require.Greater()` check then. The `require.notEqual` check doesn't make much sense though and could be removed along with the PantherRowID field initialization for the `expectedEvent`

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
kostaspap: Well, not necessarily... The parsers will always be doing validation of the incoming log either explicit or implicit (e.g. type conversions). Even if we take out auto-detection, it is quite possible that due to some misconfiguration we receive different logs than we expect and we will have to do some basic validation.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
kostaspap: What worries me is the increased complexity of the method (branches increase complexity) and difficulty to test. 
The code used to have 100% branch coverage but not any more. And it will be painful to actually achieve 100% branch coverage for all of our parsers if we follow this logic. 
From the options above, using the `aws.StringValue` method reduces complexity and restructures the code in a more testable way. I'd suggest we go for that change now and also for all upcoming parsers.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
kostaspap: Got it. Question: Why do we need the custom mappings? Couldn't we achieve the same just by putting any custom mapping in this switch statement? The benefit of the latter is that a developer can easily see all mappings in a single place.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
kostaspap: I am proposing: 
1. We parse the incoming data.
1. We validate the incoming data. This is a validation that the incoming data are in the proper format. 
1. We set the required common panther fields through the "SetRequired" method. If this fails, fail the parsing. 

My point is that the current logic is equivalent to something like below: 
```
....
value, err := doSth()
if err == nil {
  item.setValue(value)
}
....
if item.getValue() != value {
// fail
}
```

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
```diff
@@ -35,7 +38,7 @@ type CloudTrailRecords struct {
 
 // CloudTrailRecord is an AWS CloudTrail API log.
 type CloudTrail struct {
-	AdditionalEventData interface{}             `json:"additionalEventData,omitempty"`
+	AdditionalEventData *jsoniter.RawMessage    `json:"additionalEventData,omitempty"`
```
kostaspap: I know it's not part of your changes, but could you update line 36 to 
```
Records []*CloudTrail `json:"Records" validate:"required,dive"`
```
I did some tests and realized this is required in order to instruct the validator to validate the nested items.

pkg/extract/extractor.go
```diff
@@ -16,20 +16,32 @@ package extract
...
+func Extract(rawMessage *jsoniter.RawMessage, extractors ...Extractor) {
+	if rawMessage == nil {
+		return
+	}
+	result := gjson.ParseBytes(*rawMessage)
+	Parsed(result, extractors...)
+}
+
+// Parsed walks parsed JSON extracting tasty things (use if you already parsed the JSON)
+func Parsed(result gjson.Result, extractors ...Extractor) {
```
kostaspap: nit: Should be `parsed`? I'm guessing it is only `Extract` that is meant to be used outside of this package.

internal/log_analysis/log_processor/parsers/awslogs/pantherlog.go
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import "github.com/panther-labs/panther/internal/log_analysis/log_processor/parsers"
+
+// nolint(lll)
+type AWSPantherLog struct {
+	parsers.PantherLog
```
kostaspap: Isn't this going to cause problems with column indexing if we add new fields to `PantherLog`?

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
rleighton: Done. Also fixed validations in CT event struct ... they did not match docs.

pkg/extract/extractor.go
rleighton: No, Nick may want to use this in compliance and often the structures are parsed, so 2 entry points.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
rleighton: Sure, that's easy.

internal/log_analysis/log_processor/parsers/patherlog.go
```diff
@@ -26,26 +26,27 @@ import (
...
 // All log parsers should extend from this to get standardized fields (all prefixed with 'p_' as JSON for uniqueness)
 // NOTE: It is VERY important that fields are added to END of the structure to avoid needed to re-build existing Glue partitions.
 //       See https://github.com/awsdocs/amazon-athena-user-guide/blob/master/doc_source/updates-and-partitions.md
+// nolint(lll)
 type PantherLog struct {
 	//  required
-	PantherLogType   string            `json:"p_log_type,omitempty" validate:"required" description:"Panther added field with type of log"`            // nolint(lll)
-	PantherRowID     string            `json:"p_row_id,omitempty" validate:"required" description:"Panther added field with unique id (within table)"` // nolint(lll)
-	PantherEventTime timestamp.RFC3339 `json:"p_event_time,omitempty" validate:"required" description:"Panther added standardize event time (UTC)"`    // nolint(lll)
+	PantherLogType   string            `json:"p_log_type,omitempty" validate:"required" description:"Panther added field with type of log"`
```
kostaspap: Had another though... This is probably not needed? We can make sure that the views populate the "p_log_type" column with appropriate value and avoid storing this field in the actual logs where it seems redundant ^^  It would be a good cost saver and should be easy to update View generation code to deal with it

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
rleighton: Turns out it was not easy. Using aws.StringValue() made things less readable. I went ahead and made a Ptrs() methods for all the common fields methods. I think this has the desired effect. The conditionals are gone (usually) for the callers AND the varargs are useful. For example, now the alb code has no conditionals and takes advantage of the varargs for concise calls:

```
func (event *ALB) updatePantherFields(p *ALBParser) {
	event.SetRequiredPtr(p.LogType(), event.Timestamp)
	event.AppendAnyIPAddressPtrs(event.ClientIP, event.TargetIP)
	event.AppendAnyDomainNamePtrs(event.DomainName)
	event.AppendAnyAWSARNPtrs(event.ChosenCertARN, event.TargetGroupARN)
}
```

internal/log_analysis/log_processor/parsers/patherlog.go
rleighton: We can and will map other columns into the view but log type is special because it is actually not in the data and not being re-mapped. I really want this in the files to future proof the data.   This will be important with S3 select as we _really_ do not want to rely on S3 paths for this data as it could change in the future and the code would be ugly. NOTE: this is basically no cost when the files get turned to Parquet.

internal/log_analysis/log_processor/parsers/awslogs/alb_test.go
rleighton: Fixed.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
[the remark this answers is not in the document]
rleighton: I built a target in mage called glue:sync that will update all partitions. User can select specific tables or do them. This can be used anytime we change schemas. PR next week.

internal/log_analysis/log_processor/parsers/awslogs/pantherlog.go
rleighton: See first comment above, we can discuss there. Same issue and solution.

PR #90, opened 2020-01-31
[Feat] Add User Management Page (Front-end)

3nvi:
## Background

This PR adds user management back to the app

## Changes

- Add user management page
- Allow `admin` invitatinos
- Allow user listing

## Testing

Not proper testing until related lambdas are implemented


3nvi:
@austinbyers Let me know when you have added the related lambdas. Specifically we need:

* `listUsers`
* `inviteUser`

austinbyers:
@3nvi we already have `InviteUser` (see https://github.com/panther-labs/panther/blob/master/api/lambda/users/models/api.go)

So is list users all that is missing? Presumably also deleting users?

3nvi:
> So is list users all that is missing? Presumably also deleting users?

Correct

3nvi:
@austinbyers Apparently, the delete user functionality [was never exposed](https://github.com/panther-labs/users-api/tree/master/api). That's why I didn't/forgot-to add it

3nvi:
> So is list users all that is missing? Presumably also deleting users?

Forgot to add the delete user resolver. Added it

api/graphql/schema.graphql
```diff
@@ -650,6 +653,23 @@ type User {
...
+type InviteUserResponse {
+  id: ID!
+}
+
+input InviteUserInput {
+  givenName: String
+  familyName: String
+  email: AWSEmail
+  phoneNumber: AWSPhone
+  role: RoleNameEnum
```
austinbyers: FYI I removed this field in #100

3nvi: It's safely ignored by the lambda. Helps to have it here.

PR #61, opened 2020-01-29
Support AWS Partition deployment within Cloudformation.

gavinelder:
The following change moves to using AWS Supported [Pseudo paramaters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html) where possible enabling deployment in AWS partitions `aws-cn` & `aws-govcloud`

* Allow passing of `AWS_PARTITION` into the dev docker 
* Change to using AWS Pseudo variables as part of the SDK & Cloud formation templating. 

* Deploy to `AWS-CN`
* Deploy to `AWS (West)`

PR #98, opened 2020-02-03
Generic URL suffix + AWS partition

austinbyers:
## Background

Following up from #61 to use a cfn psuedoparameter for all URL suffixes and AWS partitions

## Changes

- Replace `.amazonaws.com` with `.${AWS::URLSuffix}`
- Replace `:arn:aws:` with `:arn:${AWS::Partition}:`
- Remove the `Xray` policy we conditionally attach - SAM does this automatically (much like the `BasicExecutionRole` policy it attaches to all Lambda functions so they can generate logs)
    - Add missing TracingConfig from alert processor

## Testing

- Deployed to my dev account

### DIFF
--- a/deployments/compliance/alert_processing.yml
+++ b/deployments/compliance/alert_processing.yml
@@ -114,11 +114,11 @@ Resources:
       Environment:
         Variables:
           DEBUG: !Ref Debug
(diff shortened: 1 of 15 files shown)


austinbyers:
> but as long as you did a quick grep/find all for amazon.aws.com and arn:aws: looks good to me.

I did! The remaining `arn:aws:` references are just examples in comments. I also verified the integration tests in my dev account

PR #100, opened 2020-02-04
Add ability to list and remove users

austinbyers:
## Background

Add ListUsers and RemoveUser endpoints for #90 

## Changes

- Add `ListUsers` and `RemoveUser` endpoints, as we had them originally
- Remove `Role` from `InviteUser` input struct - only one role is actually supported

I left the "userPoolID" as part of the request. Since we only have one user pool, the `users-api` could look it up, but for simplicity I just left it how we had it before. The users-api and organization-api could use a more thorough redesign for the open-source single-tenant model to address tech debt like that.

## Testing

- `mage test:ci`

Haven't had a chance to deploy it and invoke it in my test account, but wanted to get the PR up for @3nvi to work with

### DIFF
--- a/internal/core/users_api/api/invite_user.go
+++ b/internal/core/users_api/api/invite_user.go
@@ -19,7 +19,6 @@ package api
  */
 
 import (
-	"github.com/aws/aws-sdk-go/aws"
 	"go.uber.org/zap"
 
 	"github.com/panther-labs/panther/api/lambda/users/models"
@@ -47,7 +46,7 @@ func (API) InviteUser(input *models.InviteUserInput) (*models.InviteUserOutput, 
 		return nil, err
 	}
 
-	if err = userGateway.AddUserToGroup(id, aws.String("Admin"), input.UserPoolID); err != nil {
+	if err = userGateway.AddUserToGroup(id, input.UserPoolID); err != nil {
 		return nil, err
 	}
 
--- a/internal/core/users_api/api/remove_user.go
+++ b/internal/core/users_api/api/remove_user.go
@@ -19,37 +19,33 @@ package api
  */
 
 import (
-	"github.com/aws/aws-sdk-go/aws"
 	"go.uber.org/zap"
 
 	"github.com/panther-labs/panther/api/lambda/users/models"
-	"github.com/panther-labs/panther/internal/core/users_api/gateway"
 )
 
-// InviteUser adds a new user to the Cognito user pool.
-func (API) InviteUser(input *models.InviteUserInput) (*models.InviteUserOutput, error) {
-	// Add user to org mapping in dynamo
-	if err := addUser(input.Email); err != nil {
-		return nil, err
-	}
-
-	// Create user in Cognito
-	id, err := userGateway.CreateUser(&gateway.CreateUserInput{
-		GivenName:  input.GivenName,
-		FamilyName: input.FamilyName,
-		Email:      input.Email,
-		UserPoolID: input.UserPoolID,
-	})
+// RemoveUser deletes a user from cognito.
+func (API) RemoveUser(input *models.RemoveUserInput) error {
+	// Get user sub from Cognito
+	user, err := userGateway.GetUser(input.ID, input.UserPoolID)
 	if err != nil {
-		if deleteErr := userTable.Delete(input.Email); deleteErr != nil {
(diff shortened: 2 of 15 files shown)


austinbyers:
Hey, PR 💯 ! Achievement unlocked

3nvi:
@austinbyers Thanks for that. I'll test the backend and report back

internal/core/users_api/api/remove_user.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/panther-labs/panther/api/lambda/users/models"
+)
+
+// RemoveUser deletes a user from cognito.
+func (API) RemoveUser(input *models.RemoveUserInput) error {
+	// Get user sub from Cognito
+	user, err := userGateway.GetUser(input.ID, input.UserPoolID)
+	if err != nil {
+		zap.L().Error("error getting user from user pool", zap.Error(err))
```
kostaspap: we should be logging the stacktrace, so you can use `errors.Wrap` method. Same of all error logging

internal/core/users_api/api/invite_user.go
```diff
@@ -46,7 +47,7 @@ func (API) InviteUser(input *models.InviteUserInput) (*models.InviteUserOutput,
 		return nil, err
 	}
 
-	if err = userGateway.AddUserToGroup(id, input.Role, input.UserPoolID); err != nil {
+	if err = userGateway.AddUserToGroup(id, aws.String("Admin"), input.UserPoolID); err != nil {
```
kostaspap: Might worth making `Admin` a constant and put a comment on it or just put a comment here, an "all users are administrators" kind of thing

internal/core/users_api/api/remove_user.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		zap.L().Error("error deleting user from user pool", zap.Error(err))
+		return err
+	}
+
+	// Delete user from Dynamo
+	err = userTable.Delete(user.Email)
+	if err != nil {
+		zap.L().Error("error deleting user from dynamo", zap.Error(err))
+		return err
```
rleighton: If this fails we are in an inconsistent state ... if you try again, then I think the cognito DeleteUser will fail and so we are stuck never able to delete the user (right?). Ideally, on this failure you would "undo" the cognito delete. Kostas wrote some similar code for an api to avoid these kinds of deadlocks where a defer() is used to undo.

austinbyers:
Great suggestions about error handling, but I'm just restoring the identical functionality we had before for now. Once it works with the frontend, I'll merge and then can make those changes in a subsequent PR.

This entire service is pretty much 100% tech debt - the `users-api` and `organization-api` don't make sense in the single-tenant world the way they are currently setup. They reference userPoolIDs, of which there is only one, the org database is now just a single row, etc. If time allows, I may be able to refactor these services before v1, in which case it's not worth spending too long here just yet

internal/core/users_api/api/remove_user.go
austinbyers: I just restored the code as it existed before. I'll make that change in a later PR, hopefully one that refactors this service

internal/core/users_api/api/invite_user.go
austinbyers: Sure, I can make it a constant, that works

internal/core/users_api/api/remove_user.go
austinbyers: It would be strange to "undo" a Cognito delete - you have to then re-invite the user by sending them an email with temporary login credentials.

I'll take this into account in a future PR

austinbyers: Actually, at this point, the Dynamo table is pretty much just a copy of what's in Cognito. We only had it before to map users to organizations

kostaspap: Alright, then seems we need one DDB table less :) 
Do we have a story to clean this up? Seems it could be an easy win.

austinbyers: @3nvi actually feels that even though it's redundant for now, it's better to have a separate users database and that we shouldn't treat the auth layer like a database.

Regardless, some refactoring of this service would be nice to have, it certainly warrants discussion

PR #102, opened 2020-02-04
[Perf] Minor performance increase on forms

3nvi:
## Background

> Why are you making this change? Please provide a reference to any related issues and PRs

Currently the forms on policy/rule creation/update are a bit complex and laggy. Although a proper fix would involve component isolation and refactoring, adding memoization (i.e. making react understand that it doesn't need to update a certain compoonent) helps by increasing performance by ~40%

## Changes

- Add component instance memoization on all formik-related elements
- Make sure to pass referentially stable props to core form components

## Testing

> How did you test your change?

Locally, along with a panther deployment


web/src/components/fields/textarea.tsx
```diff
@@ -22,7 +22,7 @@ import { FieldConfig, useField } from 'formik';
 
 const FormikTextArea: React.FC<TextAreaProps & Required<Pick<FieldConfig, 'name'>>> = props => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const [__, meta] = useField(props.name);
+  const [field, meta] = useField(props.name);
```
rleighton: Just curious (new to typescript), why did you add 'field'?

3nvi: purely for people to see what the variable was. Unfortunately `__` doesn't mean "ignore variable"  as it happens in Go. Thus, I thought I should expose the name of the un-used variable for people to know.

All in all, it's kind of an irrelevant change :P

PR #104, opened 2020-02-04
[Docs] Add segment for EC2 deployment

### LINKED ISSUE #96: Can't deploy on EC2 `t2.micro` on linux
**Describe the bug**

Running `mage deploy` with a `t2.micro` EC2 instance as the deployment's host machine, causes the following exception:

```
Step 19/24 : COPY --from=build-env /code/web/dist web/dist COPY failed: stat /var/lib/docker/overlay2/1413660babed1db7f8483d604df7e67273a4f6843b3dbc9e3727979b0923bd54/merged/code/web/dist: no such file or directory
```

**To Reproduce**

1. Create an EC2 instance with `Amazon Linux 2 AMI (HVM), SSD Volume Type` as the image
2. Download Docker and run the docker daemon
3. Run `./dev.sh`
4. Run `mage deploy` (a quicker shortcut to just test bug would also be `docker build -f deployments/web/Dockerfile .`)

**Expected behavior**

There should be no errors

**Environment**

- Panther version or commit: `v0.1.1`
- OS: `Linux`
- Infra: `EC2 t2.micro`

**Additional context**

There seems to be an issue with multi stage build on the dockerfile

3nvi:
## Background

> Why are you making this change? Please provide a reference to any related issues and PRs

This PR adds a section in the docs that has to do with using an EC2 instance as the deployment host (instead of a local machine). It also resolves #96 

## Changes

- Add section in `Get started`

## Testing

Markdown checker

### DIFF
--- a/docs/gitbook/quick-start.md
+++ b/docs/gitbook/quick-start.md
@@ -84,6 +84,15 @@ Your AWS credentials _must_ be exported as environment variables for the docker 
 {% endhint %}
 
 {% hint style="info" %}
+Rather than deploying from your local machine, you can instead use an EC2 instance with Docker and
+git installed. This saves you the need of exporting your AWS credentials as environment variables, while benefiting
+from a potentially more performant machine for your deployment. Remember to make sure that the IAM role that the EC2
+instance assumes, has enough permissions for the creation of all Panther resources.
+
+The minimum requirements for an EC2 are 1 vCPU and 2GB of Memory. The lowest-cost instance that satisfies those requirements, is an EC2 `t2.small`.
+{% endhint %}
+
+{% hint style="info" %}
 Rather than deploying from within a docker container, you can instead configure your [development environment](development.md#manual-installation) locally. This will take more time initially but will lead to faster deployments.
 {% endhint %}
 
--- a/docs/gitbook/quick-start.md
+++ b/docs/gitbook/quick-start.md
@@ -85,7 +85,7 @@ Your AWS credentials _must_ be exported as environment variables for the docker 
 
 {% hint style="info" %}
 Rather than deploying from your local machine, you can opt to use an EC2 instance with Docker and
-git installed. Instead of exporting your AWS credentials as environment variables, you will now need an IAM role (that your EC2 instance will assume) with enough permissions for the creation of all Panther resources.
+git installed. Instead of exporting your AWS credentials as environment variables, you will need to attach an IAM role to your EC2 instance profile, with enough permissions for the creation of all Panther resources.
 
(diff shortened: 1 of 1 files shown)


docs/gitbook/quick-start.md
```diff
@@ -83,6 +83,15 @@ Your AWS credentials _must_ be exported as environment variables for the docker
 `aws-vault exec <profile> -- ./dev.sh`
 {% endhint %}
 
+{% hint style="info" %}
+Rather than deploying from your local machine, you can instead use an EC2 instance with Docker and
+git installed. This saves you the need of exporting your AWS credentials as environment variables, while benefiting
+from a potentially more performant machine for your deployment. Remember to make sure that the IAM role that the EC2
+instance assumes, has enough permissions for the creation of all Panther resources.
```
kostaspap: I'm not sure we should put an opinion on why people would use an EC2 instance over local deployment. E.g.I don't see as exporting the AWS credentials as an issue. 
I think it would be enough to say that 2GB of memory would be the minimum requirement.

3nvi: You are absolutely right. Will rephrase

docs/gitbook/quick-start.md
```diff
@@ -83,6 +83,13 @@ Your AWS credentials _must_ be exported as environment variables for the docker
 `aws-vault exec <profile> -- ./dev.sh`
 {% endhint %}
 
+{% hint style="info" %}
```
3nvi: @kostaspap Is that better?

kostaspap: >  you will now need an IAM role (that your EC2 instance will assume)

nitpick: I think the terminology used generally is "adding role to Instance profile" (see [sample](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html)). So I would slightly rephrase to `you will need to add a role to your EC2 instance profile with enough permissions for the creation....`. 
Or similar. 
Looks good overall, glad we tried it.

docs/gitbook/quick-start.md
```diff
@@ -83,6 +83,13 @@ Your AWS credentials _must_ be exported as environment variables for the docker
...
+Rather than deploying from your local machine, you can opt to use an EC2 instance with Docker and
+git installed. Instead of exporting your AWS credentials as environment variables, you will need to attach an IAM role to your EC2 instance profile, with enough permissions for the creation of all Panther resources.
+
+The minimum requirements for an EC2 machine are 1 vCPU and 2GB of Memory. The lowest-cost instance that satisfies those requirements, is an EC2 `t2.small`.
```
austinbyers: Nit: don't need to capitalize `Memory` and don't need the comma:

> The minimum requirements for an EC2 machine to deploy Panther are 1 vCPU and 2GB of memory. The cheapest instance which satisfies those requirements is a `t2.small`.

PR #108, opened 2020-02-04
Reduce compliance caching

### LINKED ISSUE #97: Policy Status Updates Much More Slowly Than Resource Status
**Describe the bug**

After updating a Policy that switches all resources from passing to failing or vice versa, the policy overview page updates all the resources' correct status very quickly (within 2-3 seconds). But it the policy status updates much more slowly (within ~ 1 minute). This leads to an inconsistent view presented to the users after a policy is updated.

**To Reproduce**

1. Go to a policy that has failing resources
2. Click on edit policy
3. Scroll down to policy body
4. Add a `return True` line at the beginning of the policy body
5. Update the policy and return to the policy overview page
6. All the resources will be Passing, but the policy status will still be failing
7. This state persists for ~1 minute

**Expected behavior**

I would expect as soon as the resources' status are updated, the policy status is updated as well.

**Screenshots**

Screen Shot 2020-02-03 at 9 51 52 AM

Screenshot shows that all policies are failing but the policy is passing.

**Environment**

- Current master 8ddce61

**Additional context**

@austinbyers tracked it to this line https://github.com/panther-labs/panther/blob/master/internal/core/analysis_api/handlers/compliance.go#L31

austinbyers:
## Background
Fixes #97 

The policy and resource overview pages include a summary of their compliance status - is this policy/resource PASS or FAIL (or ERROR) in aggregate?

That information is not explicitly stored anywhere - to determine overall pass/fail status, we have to scan the `compliance-api` to find all applicable (policy, resource) entries and compute the aggregate status.

Since this is a relatively expensive operation, the original idea was to cache the overall pass/fail status for every policy and resource to avoid constantly scanning the compliance table. However, this may have been a premature optimization and creates an unexpected user experience (#97)

I was going to remove the caching completely, but it turns out we do still need it in one place - `ListPolicies` and `ListResources` both allow you to sort and filter by compliance status, which is only possible if the `policy-api` and `resource-api`, respectively, remember at least for a short time their set of compliance information.

The fix, then, is to reduce the cache time to be as small as possible so that the list operations don't hammer `compliance-api` with queries but the overall status still updates very quickly for the user.

Better designs are possible that eliminate the caching completely, but require a much larger refactor.

## Changes

- Reduce compliance cache time from 60 seconds to 3

## Testing

- `mage test:ci`

### DIFF
--- a/internal/compliance/resources_api/handlers/compliance.go
+++ b/internal/compliance/resources_api/handlers/compliance.go
@@ -27,7 +27,9 @@ import (
 	"github.com/panther-labs/panther/api/gateway/resources/models"
 )
 
-const complianceCacheDuration = time.Minute
+// Cache pass/fail status for each policy for a few seconds so that ListResources can filter and
+// sort by compliance status without having to query the compliance-api for every resource.
+const complianceCacheDuration = 3 * time.Second
 
 type complianceCacheEntry struct {
 	ExpiresAt time.Time
--- a/internal/core/analysis_api/handlers/compliance.go
+++ b/internal/core/analysis_api/handlers/compliance.go
@@ -28,7 +28,9 @@ import (
 	compliancemodels "github.com/panther-labs/panther/api/gateway/compliance/models"
 )
 
-const complianceCacheDuration = time.Minute
+// Cache pass/fail status for each policy for a few seconds so that ListPolicies can filter and
+// sort by compliance status without having to query the compliance-api for every policy.
+const complianceCacheDuration = 3 * time.Second
 
 type complianceCacheEntry struct {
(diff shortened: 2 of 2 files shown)


kostaspap:
What are the response times after removing this caching? 
I am thinking whether we should actually prioritize this refactoring over other work since it affects the way we store data - and it would just make it more painful to migrate users.

austinbyers:
> What are the response times after removing this caching?

Haha it should be exactly 3 seconds to update in the page, since the caching was just reduced, not removed

> I am thinking whether we should actually prioritize this refactoring over other work

I would certainly like to, not sure if we'll have time. Happy to discuss

PR #109, opened 2020-02-04
Make CFN post-processing more resilient

nhakmiller:
## Background

@rleighton reported an issue where deployments in `us-east-1` would fail. This was corroborated by another user is `us-east-2`. I was not able to reproduce in `us-west-2`.

It turns out the s3 global endpoint format we were converting to is not viable in certain regions, and also the regional endpoint we were trying to convert from was actually a global endpoint in some regions. It appears the `aws cli` is region aware when producing these template URLs, so we need to be more resilient in handling them.

## Changes

- Changed format of transformed template URLs to match globally supported format suggested by @kostaspap 
- Added checks to not change the URL if it is already a global endpoint

## Testing

- deployed in `us-west-2`, `us-east-1`, `us-east-2`, and `eu-west-1`

### DIFF
--- a/tools/mage/deploy_test.go
+++ b/tools/mage/deploy_test.go
@@ -27,16 +27,21 @@ import (
 func TestFixTemplateURL(t *testing.T) {
 	// Basic example
 	originalTemplate := "TemplateURL: https://s3.region.amazonaws.com/bucket/panther-app/1.template"
-	expectedTemplate := "TemplateURL: https://bucket.s3.amazonaws.com/panther-app/1.template"
+	expectedTemplate := "TemplateURL: https://s3.amazonaws.com/bucket/panther-app/1.template"
 	assert.Equal(t, expectedTemplate, fixPackageTemplateURL(originalTemplate))
 
 	// Maintains leading whitespace
 	originalTemplate = "   TemplateURL: https://s3.region.amazonaws.com/bucket/panther-app/1.template"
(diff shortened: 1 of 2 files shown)


nhakmiller:
While I say I tested in `eu-west-1`, the deployment there never actually completed (it was stuck in the `REVIEW_IN_PROGRESS` state), but the deployment did get past the part where this issue would have manifested itself so I'm still considering it a successful test.

tools/mage/deploy_test.go
```diff
@@ -27,16 +27,21 @@ import (
 func TestFixTemplateURL(t *testing.T) {
 	// Basic example
 	originalTemplate := "TemplateURL: https://s3.region.amazonaws.com/bucket/panther-app/1.template"
-	expectedTemplate := "TemplateURL: https://bucket.s3.amazonaws.com/panther-app/1.template"
+	expectedTemplate := "TemplateURL: https://s3.amazonaws.com/bucket/panther-app/1.template"
```
kostaspap: How does these URLs work look in China? Is it still `amazonaws.com`?

nhakmiller: I imagine it would be `amazonaws.cn` instead, but we don't actually change that line. We just re-arrange the existing URL to be global. It is worth checking whether S3 buckets in CN default to being globally name spaced or regionally name spaced by the `aws cloudformation package` command. Unfortunately I am unable to test this as I whenever I run that command with a china region specified, it just throws an error. If the china regions follow the same pattern that the us-west and eu regions follow, then it will still work fine.

Issue #113, opened 2020-02-05
Add support for "custom" fields in VPC Flow Logs


rleighton:
Currently Panther supports only "standard" (should be called legacy) fields in VPC Flow Logs. Missing are important fields like instance id and tcp flags. 
Reference: https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs.html#flow-log-records

rleighton:
This is mostly done now. PR tomorrow or next day after testing in dev account.

rleighton:
![image](https://user-images.githubusercontent.com/18419762/73893285-39ee0700-4847-11ea-9cbf-c4a154fee6ae.png)

rleighton:
Done

PR #139, opened 2020-02-07
Support Custom VPC Flow Logs

rleighton:
## Background
See  https://github.com/panther-labs/panther/issues/113

## Testing
mage test:ci
ran in dev account, enabled custom VPC Flow Logs  and used Athena to spot check results.


internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
```diff
@@ -53,75 +50,190 @@ type VPCFlow struct {
...
-	// parser should only receive 1 line at a time
-	record := records[0]
+	return event
+}
 
-	if len(record) < vpcFlowMinNumberOfColumns {
-		zap.L().Debug("failed to parse the log as csv (wrong number of columns)")
+// Parse returns the parsed events or nil if parsing failed
+func (p *VPCFlowParser) Parse(log string) []interface{} {
+	if p.columnMap == nil { // have not read header!
```
nhakmiller: Does this mean that this log gets thrown away? When would this happen in practice?

internal/log_analysis/log_processor/classification/classifier.go
```diff
@@ -92,7 +94,14 @@ func safeLogParse(parser parsers.LogParser, log string) (parsedEvents []interfac
...
 	}()
-	parsedEvents = parser.Parse(log)
+	if c.headerParsed {
+		parsedEvents = parser.Parse(log)
+	} else {
+		parsedEvents = parser.ParseHeader(log)
+		if parsedEvents != nil { // set on first SUCCESSFUL parse
+			c.headerParsed = true
+		}
+	}
```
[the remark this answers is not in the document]
nhakmiller: Re-reading this code a few times, I'm a bit confused.

Does `headerParsed` get set back to `False` before parsing a future log? Or are we expecting the headers to never change? I feel that is bad assumption if we're processing VPC Flow Logs from two separate accounts. Or am I missing something that makes this not necessary?

Also, what happens if we support another log type in the future that also has headers. Would parsing the header for one of those log types set this to true, and then it doesn't get set back to false for the next one, thus incorrectly believing that the header is already parsed?

Does a new classifier struct get instantiated each time a new log file is processed? That was not my original understanding, as then I don't the weighting of the parsers would be relevant. Perhaps I'm just fundamentally misunderstanding some things here.

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
rleighton: Only if there is bug ... I cannot run w/out a header.

internal/log_analysis/log_processor/classification/classifier.go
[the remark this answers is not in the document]
rleighton: Yes, there is a new classifier per file (https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/processor/processor.go#L74). So we can record if we have processed the header in the classifier. This change adding the ParseHeader() method will be useful for other CSV data sources we might get in the future.

BTW, I suspect the "classifier" concept will get less and less useful as we get more log sources and we will not have good ways to disambiguate (e.g., 2 sources using Common Log Format).

Right now we always have the info to know what the kind of log is being processed but we have no way of passing it thru the system. One day we should let users tell us the type during onboarding and pass that info thru to log processing.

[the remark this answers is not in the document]
nhakmiller: If each individual log file has its own classifier, then what is the point of the parser priority queue? I guess this is outside the scope of this PR so feel free to move forward, but the priority queue seems like a lot of overhead and a very complicated way to search a list one time and remember what you found.

Will the classifier not just search the (new & randomly ordered) parser priority queue to find what it's looking for as it parses the first log line, put that parser that at the front of the queue, and then just peek the front of the queue for every single subsequent parse of that classifiers lifetime?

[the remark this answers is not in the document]
rleighton: Yup! Peek() https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/classification/classifier.go#L133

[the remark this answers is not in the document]
rleighton: Ah, but this is the first stateful parser ... I just realized we are using the same parser struct (ptr) for all files. I have to fix that. I've not seen any issues because load is not high and data is uniform. Patch coming... sigh.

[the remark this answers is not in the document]
nhakmiller: If they're sharing the same parser this is even more of an issue right? Because then the header structure could be different for two files being parsed in parallel. There may be thread safety issues here even if you reset the `headerParsed` variable in between files because two files with different formats could be getting parsed at the same time.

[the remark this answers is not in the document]
rleighton: Sort of. The issue would be processing 2 vpc flow files concurrently, each has a separate classifier (with independent headerParsed flags) BUT they actually share a parser where the headers are stored. If the headers were different then it would all be garbage (patch coming after burn-in for an hour or so processing logs). Parsers are now stateful (possibly) so we need a New() method to act like a factory pattern. A classifier creates a q of parsers to apply for a file. Before this change we just use a ptr to the registered parser structs. After this change New() is called on each registered parser to get a new instance. It is a 1 line change in the classifier but I have to define a New() method for each parser.

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
rleighton: Added a log line if this happens.

internal/log_analysis/log_processor/classification/classifier.go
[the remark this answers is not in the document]
rleighton: Nick, I'm very glad you started this discussion. The log processing code has gone through a number of refactorings and the now that classifiers are per-datastream (which are really just files) the notion of needing a priority q no longer exists. So I just took it out :)

Here is how the classifier works now:
On the first log line it loops over all registered parsers, if one works, it sets a local ptr to the parser. On subsequent log lines it just uses the parser found processing the first log line.

The code is much simpler now and slightly more performant (no need to push/pop/peak into a heap).

We will see what Kostas thinks :)

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
rleighton: Changed the code this does not happen anymore.

internal/log_analysis/log_processor/classification/classifier.go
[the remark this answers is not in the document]
kostaspap: I don't think we can take out the queue. There is no guarantee that each file containing JSONL data will contain only one type of logs. e.g. in a local installation we are consuming OsQuery logs from laptops. OsQuery sends all the logs through the same Firehose stream and we are reading them from there.  Each S3 object created by OsQuery has any type of OsQuery logs in it. 

In the case of CSV files though (e.g. VPC Flow logs) I believe we can safely assume that there will be only one parser per file because ...csv . 

Until before this PR, parsers were stateless so sharing them between classifiers was safe and intentional. This PR introduces state to the parsers so by definition they cannot be shared across classifier instances.

internal/log_analysis/log_processor/parsers/awslogs/guardduty.go
```diff
@@ -67,6 +67,10 @@ type GuardDutyService struct {
 // VPCFlowParser parses AWS VPC Flow Parser logs
 type GuardDutyParser struct{}
 
+func (p *GuardDutyParser) New() parsers.LogParser {
+	return &GuardDutyParser{}
+}
```
kostaspap: This method seems redundant. It's perfectly aligned with Golang's style (and preferred) to create new instances using `&structname{}` this is what we do in most parts of the code, apart from places that require a special initialization process which doesn't seem to be the case here.

internal/log_analysis/log_processor/classification/classifier.go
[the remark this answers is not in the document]
rleighton: The New() method takes care of the state issue (even with a Queue).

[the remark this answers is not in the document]
rleighton: Now to the need for a priority queue. Given the state of Osquery, I guess we need it for now but I question the premise. For example, in the osquery example, we know the source is all osquery, we should take advantage of that knowledge and not throw it away. Imagine the day we get 1000 parsers , then the first time a lambda runs it has to look at on average at 500 parsers. For a CloudTrail logs this is very expensive since a single log line is so large. I have to look into why, but we have p100 parse times of over a few seconds (only a few times an hour) in my dev env which just seems wrong (I can see that with the new dashboard). My suspicion is that this happens when we guess wrong at first and big CT files are parsed as CSV over and over again (failing) until we finally get to the CT parser. As we discussed before, also there is the case when 2 different sources use the same logging format and then we have ambiguity (when likely we knew all along at the source the correct log type when configuring the source).

The future is passing the info we know down ... the classifier constructor can take a set of hints and we would try those first (and would be  correct almost all the time). This will fix the scaling issue with the design. This does not fix the fact that osquery can have multiple log types in a stream. Syslog may be similar also.  One way to fix this is to make osquery a single "LogParser", with the current parsers as "subparsers". Since for osquery I would think (but do not know) that the logs come in random order, so having a priority q is not of much help between the osquery types.  Would having a a "wrapper" log parser be acceptable to handle cases when a source has multiplexed data types?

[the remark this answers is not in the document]
kostaspap: We have already agreed that we want to be able to provide hints so that the classification will be more efficient. I was not arguing this. My comment was the for the current PR. If we remove the priority queue in the current PR we will run into issue for the OsQuery usecase. We also don't have right now a way for users to pass hints to the classifier. 

I think it's probably better to move the parser/subparser discussion in a separate thread, it is definitely another way to achieve what we currently do with priority queue with its pros and cons.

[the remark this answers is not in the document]
rleighton: Agree, I just wanted to socialize this here. This something to deal with sooner than later. I'd like the contractors that implement the parsers not be forced to re-architect the framework when they hit a case that does not fit. I'll put the q back.

internal/log_analysis/log_processor/parsers/awslogs/guardduty.go
rleighton: I don't understand. This is required for stateful parsers and the classifier does not know the type. 

The old code did (shares a parser for all processing, not what we want):
q.items = append(q.items, &ParserQueueItem{		
 			parser:  parserMetadata.Parser,		
 			penalty: 1,		
 		})
The new code (when I put the q back) will need to do:

q.items = append(q.items, &ParserQueueItem{		
 			parser:  parserMetadata.Parser.New(),		
 			penalty: 1,		
 		})

Is there another way?

kostaspap: You are very right, I misread the code. Seems this is required indeed.

internal/log_analysis/log_processor/classification/classifier.go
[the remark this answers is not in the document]
rleighton: Priority queue restored. New() is used when creating it providing state for parsers that need it.

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
```diff
@@ -29,17 +30,13 @@ import (
...
-)
-
 var VPCFlowDesc = `VPCFlow is a VPC NetFlow log, which is a layer 3 representation of network traffic in EC2.
 Log format & samples can be seen here: https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-records-examples.html`
 
 // nolint:lll
-type VPCFlow struct {
-	Version     *int               `json:"version,omitempty" validate:"required" description:"The VPC Flow Logs version. If you use the default format, the version is 2. If you specify a custom format, the version is 3."`
-	Account     *string            `json:"account,omitempty" validate:"omitempty,len=12,numeric" description:"The AWS account ID for the flow log."`
+type VPCFlow struct { // NOTE: since fields are customizable by users, the only "required" fields are the Start/End times since those are critical and data is useless w/out those
```
kostaspap: Is this accurate? From the documentation I see that users can put additional fields, but there is a core number of fields that is required and will actually arrive in specific order ( so we don't really need to try to map those since we already know the order and will save some lines of code)
https://aws.amazon.com/blogs/aws/learn-from-your-vpc-flow-logs-with-additional-meta-data/

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
```diff
@@ -53,75 +50,196 @@ type VPCFlow struct {
...
-// Parse returns the parsed events or nil if parsing failed
-func (p *VPCFlowParser) Parse(log string) []interface{} {
-	// Flow log files usually (always?) have a header (might have more columns):
-	//    version account-id interface-id srcaddr dstaddr srcport dstport protocol packets bytes start end action log-status
-	// If this is a header, return success but no events
-	if strings.HasPrefix(log, vpcFlowHeader) {
-		return []interface{}{} // empty list
-	}
+const (
+	vpcFlowHeaderThreshold = 5 // the number of headers that have to match to detect as VPCFlow
```
kostaspap: See my comment above: According to the AWS blog post, we logs can have additional fields, but there is a core number of fields that will always be there (and in specific order)

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
```diff
@@ -53,75 +50,196 @@ type VPCFlow struct {
...
-		return nil
+func (p *VPCFlowParser) isVpcFlowHeader(log string) bool {
+	// CloudTrail can be detected as VPCFlow due to lucky token matching, skip JSON looking things here!
+	if len(log) > 0 && log[0] == '{' {
+		return false
+	}
+	headers := strings.Split(log, " ")
+	matchCount := 0
+	for _, header := range headers {
+		header = strings.TrimSpace(header) // just in case
```
kostaspap: We do split on empty space. This is probably not needed?

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
```diff
@@ -53,75 +50,196 @@ type VPCFlow struct {
...
+	// require a minimal number of matching fields
+	isHeader := matchCount >= vpcFlowHeaderThreshold
+	if isHeader {
+		p.columnMap = make(map[int]string, len(headers))
+		for i, header := range headers {
+			header = strings.TrimSpace(header) // just in case
+			p.columnMap[i] = header
+		}
 	}
+	return isHeader
```
kostaspap: nit: You don't need the isHeader variable. 
```
if matchCount < vpcFlowHeaderThreshold {
  return false
}
p.columnMap = make(map[int]string, len(headers))
... 
return true
```

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
```diff
@@ -53,75 +50,196 @@ type VPCFlow struct {
...
+		vpcFlowPktSrcAddr: {},
+		vpcFlowPktDstAddr: {},
+	}
+)
 
-	records, err := reader.ReadAll()
-	if len(records) == 0 || err != nil {
-		zap.L().Debug("failed to parse the log as csv")
-		return nil
+func (p *VPCFlowParser) isVpcFlowHeader(log string) bool {
```
kostaspap: We have been following the code style that package private methods come after public methods...
`isVpcFlowHeader` and `populateEvent` should go to the end of the file.

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
rleighton: Let me test it. I added all the fields in my dev account. I think you can add what you want.

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
rleighton: I was worried about an Amazon bug where there is an extra space, then the lookup fails. I don't trust them :)

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
rleighton: True.

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
rleighton: kk

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
rleighton: Looks like you can pick anything in any order. I picked 6 , out off order, it let me. Here is an output file:
```
account-id dstaddr srcport action start end srcaddr
050603629990 10.0.0.56 80 ACCEPT 1581357144 1581357145 10.0.0.71
050603629990 10.0.0.71 17418 ACCEPT 1581357144 1581357145 10.0.0.56
050603629990 10.0.0.56 80 ACCEPT 1581357168 1581357173 10.0.0.71
050603629990 10.0.0.71 17434 ACCEPT 1581357168 1581357173 10.0.0.56
```

kostaspap: ^^ Interesting! Thanks for verifying this

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow.go
```diff
@@ -53,23 +50,89 @@ type VPCFlow struct {
...
+type VPCFlowParser struct {
+	columnMap map[int]string // column position to header name
+}
+
+func (p *VPCFlowParser) New() parsers.LogParser {
+	return &VPCFlowParser{}
+}
+
+const (
+	vpcFlowHeaderThreshold = 5 // the number of headers that have to match to detect as VPCFlow
```
kostaspap: Just for clarification: This is our own requirement? Does VPC FL have a minimum number of columns?

rleighton: Our own ... otherwise we might get a false positive classification on some of our CSV files. It actually happened for CloudTrail but I put a test to reject if the line starts with '{' to avoid that. If we could set the log type I would take this check out.

PR #149, opened 2020-02-09
Fix regression dealing with empty sets in panther log

rleighton:
## Background
Previously I changed the "set" serialization to emit nothing when set is empty. This was wrong because the set is "non-nil" and the omit empty tag will still emit the  field which results in incorrect json.

This only happens when we parse something like: "arn": "" . We skip blank strings populating the set BUT the set gets made (non-nil) but empty. A corner case.

## Changes
Emit a [] when set is empty.

## Testing
mage test:ci
ran in dev account


jacknagz:
@rleighton not sure why the CLA bot is saying this, you are whitelisted, ignore for now

PR #150, opened 2020-02-09
Handle bad json data in rules engine, log matches and execution time

rleighton:
## Background
During development of a CloudWatch dashboard for the rules engine I needed to make some improvements to logging and error handing.

## Changes
* put try/except around json event parsing in case of bad data (found some, source of another PR)
* log number of matches and match time (will be graphed in dashboard)
* log name of rule if a rule exception was thrown (I was testing a crashme rule and found this)

## Testing
In my dev account using new dashboard.

### DIFF
--- a/internal/log_analysis/rules_engine/src/main.py
+++ b/internal/log_analysis/rules_engine/src/main.py
@@ -56,11 +56,11 @@ def lambda_handler(event: Dict[str, Any], unused_context) -> Union[None, Dict[st
     for log_type, data_streams in log_type_to_data.items():
         for data_stream in data_streams:
             for data in data_stream:
-                try: # bad json data can cause exceptions to be thrown
+                try: # Bad json data can cause exceptions to be thrown. Best effort: log and continue
                     for matched_rule in rules_engine.analyze(log_type, json.loads(data)):
                         matched.append((matched_rule, data))
                 except Exception as err:  # pylint: disable=broad-except
-                    logger.error("Error matching {}: {}".format(data, err))
+                    logger.error("Error during matching: {}".format(err)) # do not log data!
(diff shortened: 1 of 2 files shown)


internal/log_analysis/rules_engine/src/main.py
```diff
@@ -54,14 +56,17 @@ def lambda_handler(event: Dict[str, Any], unused_context) -> Union[None, Dict[st
     for log_type, data_streams in log_type_to_data.items():
         for data_stream in data_streams:
             for data in data_stream:
-                for matched_rule in rules_engine.analyze(log_type, json.loads(data)):
-                    matched.append((matched_rule, data))
+                try: # bad json data can cause exceptions to be thrown
```
kostaspap: This is concerning... did we actually hit this issue? Or it is more a protective measure just in case...? 
Rules engine input is generated by either log processor or (in the future) rules engine itself, so this should never happen. Just interested to know if you hit the issue, it's good to have protection against it in any case. 

Might make sense to do the try/catch around the `json.loads` command if that's what you are trying to protect against and put the appropriate descriptive error message? The current error message is different from the comment, so it is not so clear what we are trying to secure ourselves from.

rleighton: Yes, originally I  wrapped just the `json.loads` but in python you cannot use a continue in an except block, so instead I just wrapped the whole match loop.

rleighton: The comment was just meant to show what the major concern was, we could still get an exception for other reasons. I just noticed I log the bad data but we don't want to do that by policy (despite how handy it is). Let me change that .

Issue #120, opened 2020-02-06
panther-created S3 buckets have AWS Account IDs in their name


kbroughton:
**Describe the bug**
This is a minor security bug but might be considered more of an enhancement. 

I know AWS uses account IDs in created resources all the time, but there are two good reasons not to do so.

1.  Pacu allows unauthenticated enumeration attacks given only a victim AWS account ID. This can be very useful. It's best to leak the account ID in as few places as possible.

2. Sometimes an attacker gains a degree of privilege to read or write to a bucket but cannot list buckets. Having random strings in bucket names adds a little protection. So using a random GUID instead of the AWS account ID means any panther-created buckets can't be brute-force scanned by someone with a big database of AWS Account Ids.

**To Reproduce**
1. deploy panther
2. note the S3 bucket names created

**Expected behavior**
Create all S3 resources with random short GUIDs instead of Account IDs

**Screenshots**

**Environment**
N/A

**Additional context**
As a pen tester, we love finding account IDs and knowing S3 bucket names.

nhakmiller:
Thanks for the suggestion. 

As far as I'm aware, Panther creates the following 5 buckets:

- panther-analysis-versions
- panther-processed-data
- panther-s3-logs
- panther-source
- panther-athena-results

panther-s3-logs and panther-source could both be easily renamed with a random ID, as those are only referenced during the deployment stage.

panther-analysis-versions would be pretty easy to rename as it is already stored as an environment variable in the lambda that uses it, we would just need to pass in the random string to CloudFormation during the initial deployment.

panther-athena-results and panther-processed-data are part of the Log Analysis/Historical Search flow which I am less familiar with. @rleighton @kostaspap , would you imagine it is difficult to name these two buckets with a random ID generated at first deployment? I think it would just get passed through CloudFormation as a parameter, and stored in an environment variable in lambdas that need it.

--

As to whether or not we should do this, I think it is a good idea from a defense in depth perspective. Obviously the bucket names are not secret, but if it is easy for us to implement and makes enumeration that much harder than I'm all for it.

It also can leak information in that if I know a company has account ID 11111, I can try to create a bucket with a name in the format `panther-source-1111-region` (or whatever the right format is) and if I can't create it then I can guess that they are running Panther, which again isn't the end of the world but if we can avoid leaking that info we should.

kostaspap:
[...]

PR #161, opened 2020-02-10
Autogenerate S3 bucket names

austinbyers:
## Background
Fixes: #120

Panther currently creates five S3 buckets, which have names of the form "panther-(name)-(accountId)-(region)".  We chose this format because this is how AWS names default buckets: the accountId + region generally guarantee global uniqueness. However, there are security problems associated with a predictable name, especially one that includes an accountID (#120)

So what does an ideal bucket name look like?

- Must be globally unique (S3 is a global service)
- Must persist across deployments
- Should not leak any information about the account
- Should resist enumeration attacks (i.e. include randomness)

Fortunately, CloudFormation can autogenerate bucket names which satisfy all of the above requirements: `(stack)-(resource)-(random suffix)`

**This is a breaking change**, but we might as well do it now before v1. Existing deployments will have to start over or copy the data from the old to the new buckets.

## Changes

- Remove all `BucketName` properties - let CloudFormation name the buckets
    - Simplify resource names since they are part of the bucket name now
    - S3 access logs are still prefixed with accountID and region. This means access logs won't have the exact bucket name as a prefix, but it's still clear which bucket generated the logs.
- Fix `analysis-api` integration test (which has hardcoding bucket name)
    - Unrelated fix: always show output from integration tests

### Alternatives Considered
I considered using the [StackId pseudo parameter](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-stackid), which is also random and persists across deployments. We could name buckets "(stackName)-(resourceName)-(stackId)", which would allow us to construct the bucket name for S3 access log prefixes as well, but then we would have to do string splitting to pull out just the UUID from the stack id/arn.

## Testing

- Deployed to dev account
- Signed in with a new user
- `mage test:integration`

Generated buckets:

Screen Shot 2020-02-10 at 2 56 35 PM

### DIFF
--- a/deployments/backend.yml
+++ b/deployments/backend.yml
@@ -384,6 +384,9 @@ Resources:
         Status: Enabled
 
 Outputs:
+  AnalysisVersionsBucket:
+    Description: Bucket where Python version histories are stored
+    Value: !GetAtt AnalysisAPI.Outputs.BucketName
   AthenaResultsBucket:
     Description: Bucket where Athena writes query results
     Value: !Ref AthenaResults
--- a/deployments/backend.yml
+++ b/deployments/backend.yml
@@ -295,10 +295,10 @@ Resources:
   ProcessedDataBucket:
     Type: AWS::S3::Bucket
     Properties:
-      BucketName: !Sub panther-processed-data-${AWS::AccountId}-${AWS::Region}
+      BucketName: !Sub panther-processed-data-${AWS::StackId}
       LoggingConfiguration:
         DestinationBucketName: !ImportValue Panther-LogBucket
-        LogFilePrefix: !Sub panther-classified-data-${AWS::AccountId}-${AWS::Region}/
+        LogFilePrefix: !Sub panther-processed-data-${AWS::StackId}/
       BucketEncryption:
         ServerSideEncryptionConfiguration:
           - ServerSideEncryptionByDefault:
@@ -364,14 +364,14 @@ Resources:
   AthenaResultsBucket:
     Type: AWS::S3::Bucket
     Properties:
-      BucketName: !Sub panther-athena-results-${AWS::AccountId}-${AWS::Region}
+      BucketName: !Sub panther-athena-results-${AWS::StackId}
       LifecycleConfiguration:
         Rules:
           - ExpirationInDays: 30
             Status: Enabled
       LoggingConfiguration:
         DestinationBucketName: !ImportValue Panther-LogBucket
-        LogFilePrefix: !Sub panther-athena-results-${AWS::AccountId}-${AWS::Region}/
+        LogFilePrefix: !Sub panther-athena-results-${AWS::StackId}/
       BucketEncryption:
         ServerSideEncryptionConfiguration:
           - ServerSideEncryptionByDefault:
--- a/deployments/backend.yml
+++ b/deployments/backend.yml
@@ -292,13 +292,12 @@ Resources:
       TopicName: panther-processed-data-notifications
       KmsMasterKeyId: !Ref QueueEncryptionKey
 
-  ProcessedDataBucket:
+  ProcessedData:
     Type: AWS::S3::Bucket
     Properties:
-      BucketName: !Sub panther-processed-data-${AWS::StackId}
       LoggingConfiguration:
         DestinationBucketName: !ImportValue Panther-LogBucket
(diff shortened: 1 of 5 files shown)


rleighton:
Can you list some example names from your test deployments?

austinbyers:
@nhakmiller one of the integration tests was hardcoding the bucket name (just pushed a fix for that), but I searched every other bucket prefix and couldn't find anything. Feel free to double check

@rleighton I updated the PR description with a screenshot of the generated buckets

rleighton:
Do we want to do:
  (stack)-(region)-(resource)-(random suffix)
So they nicely group by region when listed? 

I'd like to avoid clicking around to find the bucket I care about. Folks will likely deploy this to multiple regions.

Thoughts?

austinbyers:
> (stack)-(region)-(resource)-(random suffix)

The only difference between that and the default version is the addition of the region, right? You can sort buckets by region in the S3 console, I'm not sure that putting the region in the name adds much value. Uniqueness is still guaranteed across regions (I deployed to a separate region to verify)

To implement this suggestion, we would have to build the random suffix ourselves by string-splitting the stackID, something like:

```yaml
BucketName: !Sub
  - ${AWS::StackName}-mybucket-${AWS::Region}-${Suffix}
  - {Suffix: !Select [!Split [!Ref AWS::StackId, "/"], 2]}
```

This would generate names like `panther-buckets-source-us-east-1-51af3dc0-da77-11e4-872e-1234567db123`. That name is clearly too long, we could split it even further to get just some of the UUID:

```yaml
BucketName: !Sub
  - ${AWS::StackName}-mybucket-${AWS::Region}-${Suffix}
  - {Suffix: !Select [!Split [!Select [!Split [!Ref AWS::StackId, "/"], 2]], ":"], 3]}
```

to get `panther-buckets-source-us-east-1-1234567db123`, but this seems unnecessarily complex compared to just letting CloudFormation generate the name

austinbyers:
> Folks will likely deploy this to multiple regions.

We as developers certainly will for testing, but why would the Panther application itself be in multiple regions?

rleighton:
> > (stack)-(region)-(resource)-(random suffix)
> 
> The only difference between that and the default version is the addition of the region, right? You can sort buckets by region in the S3 console, I'm not sure that putting the region in the name adds much value. Uniqueness is still guaranteed across regions (I deployed to a separate region to verify)
> 
> To implement this suggestion, we would have to build the random suffix ourselves by string-splitting the stackID, something like:
> 
> ```yaml
> BucketName: !Sub
>   - ${AWS::StackName}-mybucket-${AWS::Region}-${Suffix}
>   - {Suffix: !Select [!Split [!Ref AWS::StackId, "/"], 2]}
> ```
> 
> This would generate names like `panther-buckets-source-us-east-1-51af3dc0-da77-11e4-872e-1234567db123`. That name is clearly too long, we could split it even further to get just some of the UUID:
> 
> ```yaml
> BucketName: !Sub
>   - ${AWS::StackName}-mybucket-${AWS::Region}-${Suffix}
>   - {Suffix: !Select [!Split [!Select [!Split [!Ref AWS::StackId, "/"], 2], ":"], 3]}
> ```
> 
> to get `panther-buckets-source-us-east-1-1234567db123`, but this seems unnecessarily complex compared to just letting CloudFormation generate the name

Yuk!  Nevermind!

kostaspap:
> > Folks will likely deploy this to multiple regions.
> 
> We as developers certainly will for testing, but why would the Panther application itself be in multiple regions?

It quite possible that we will eventually offer multi-region or even purely regional setups. This will be really needed for the log-processing where cross-region data transport is pretty $$$ but also for redundancy.

rleighton:
> > > Folks will likely deploy this to multiple regions.
> > 
> > 
> > We as developers certainly will for testing, but why would the Panther application itself be in multiple regions?
> 
> It quite possible that we will eventually offer multi-region or even purely regional setups. This will be really needed for the log-processing where cross-region data transport is pretty $$$ but also for redundancy. We don't do that now, but something to keep in mind nevertheless.

I fully expect most customers running in multiple regions with large log processing needs to do:
1. Deploy Panther in their closest region (compliance + log processing)
2. For every account/region NOT in 1 deploy Panther , but just enable log processing.

PR #163, opened 2020-02-11
mage glue:sync

rleighton:
## Background
If a log parser schema changes to add more columns, since we are using JSON and/or Parquet, the core data is backward compatible since access is not positional (good). Unfortunately, Athena does a POSITIONAL check of a JSON table's schema against the schema of the partitions in a query and if they do not align it fails. This is a bug IMO (Athena has no such a requirement for Parquet tables) but we cannot change it.

This PR adds a mage target 'glue:sync' which is an interactive program that asks the user for tables to update (default to all) and a date range to update. The program itself just deletes and recreates Glue table partitions so that the latest table schema is applied to all partitions. After a release where schemas changed the release notes will instruct user to run this command for data processed before the upgrade for the specific affected tables. If we have enough state in a release (version and version update date) we might in the future have the deployment automatically execute the 'sync'.  The tool is also useful during development of log parsers as schemas change.

There are 2 more changes in this PR related to partitions:
1. Refactor CreateJSONPartition() to NOT take a bucket as an argument. This is not needed since the method looks up the table and the table has the bucket in the meta data.

2. Check for existence of the partition first in CreateJSONPartition(). I was watching the CloudTrail during log processing I noticed when a lambda starts we get 2 Glue API calls, 1 to fetch the table and 1 to create a partition (which usually fails because the partition usually exists since actual partition creation is done exactly 1 time per hour). So most of the time we see 2 API calls and one has a failure (partition exists) which is expensive (mutating APIs are slow) and the failure is concerning to anyone monitoring CloudTrail. This change reduces the calls to 1 most of the time (checking for the existence of the partition, which is usually true)  and there are no spurious errors in the logs.

## Changes
1. mage glue:sync target
2. Change signature for CreateJSONPartition() 
3. Add existence check for partition in CreateJSONPartition() 

## Testing
mage test:ci
mage test:integration
mage deploy and run in dev account
mage glue:sync followed by Athena queries


pkg/awsglue/partition.go
```diff
@@ -17,33 +17,53 @@ package awsglue
...
 
-func (gm *GlueMetadata) CreateJSONPartition(client glueiface.GlueAPI, s3Bucket string, t time.Time) (err error) {
+func (gm *GlueMetadata) CreateJSONPartition(client glueiface.GlueAPI, t time.Time) error {
+	// check first if present, avoids extra API calls and "AlreadyExistsException" errors in CloudTrail logs
+	if _, getPartitionErr := gm.GetPartition(client, t); getPartitionErr != nil {
+		if awsErr, ok := getPartitionErr.(awserr.Error); ok {
+			if awsErr.Code() != "EntityNotFoundException" {
+				return getPartitionErr
+			}
+		}
```
nhakmiller: I feel like this `if` statement should have an `else` block that returns `getPartitionErr`, to handle non-AWS errors that may get returned, right? The current code will continue in two cases:

1. `getPartitionErr` is an AWS error, and it had an `EntityNotFoundException` error code
2. `getPartitionErr` is a non-AWS error

I think we only want to continue in case 1, correct?

pkg/awsglue/partition.go
```diff
@@ -75,3 +95,20 @@ func (gm *GlueMetadata) DeletePartition(client glueiface.GlueAPI, t time.Time) (
...
+// SyncPartition deletes and re-creates a partition using the latest table schema. Used when schemas change.
+func (gm *GlueMetadata) SyncPartition(client glueiface.GlueAPI, t time.Time) (err error) {
+	_, err = gm.DeletePartition(client, t)
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok {
+			if awsErr.Code() != "EntityNotFoundException" { // don't fail if partition is not there
+				return errors.Wrapf(err, "delete partition for %s.%s at %v failed", gm.DatabaseName(), gm.TableName(), t)
+			}
+		}
+	}
```
austinbyers: The if statements can be simplified:

```go
if _, err = gm.DeletePartition(client, t); err != nil {
    if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() != "EntityNotFoundException" {
        // don't fail if partition is not there
       ...
    }
}
```

But what if there is an error that can't be cast to an `awserr`? Then the non-nil `err `variable will be ignored and overwritten on line 109. Maybe that's not actually possible, but for safety, I would rewrite this as

```go
if _, err = gm.DeletePartition(client, t); err != nil {
    if awsErr, ok := err.(awserr.Error); !ok || awsErr.Code() != "EntityNotFoundException" {
        // error is either not from aws or has an error code other than not found
        return errors.Wrapf...
    }
}
```

I also would not use the named return here, since you're never returning the `err` variable anyway, you're always returning a wrapper.

pkg/awsglue/partition.go
```diff
@@ -75,3 +95,20 @@ func (gm *GlueMetadata) DeletePartition(client glueiface.GlueAPI, t time.Time) (
...
+
+// SyncPartition deletes and re-creates a partition using the latest table schema. Used when schemas change.
+func (gm *GlueMetadata) SyncPartition(client glueiface.GlueAPI, t time.Time) (err error) {
+	_, err = gm.DeletePartition(client, t)
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok {
+			if awsErr.Code() != "EntityNotFoundException" { // don't fail if partition is not there
+				return errors.Wrapf(err, "delete partition for %s.%s at %v failed", gm.DatabaseName(), gm.TableName(), t)
+			}
+		}
```
nhakmiller: Same question as above, this will just continue if a non-aws error was returned. Not sure if that is the desired behavior.

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,127 @@
...
+
+	fmt.Print("Please input a regular expression to select a subset of tables (or <enter> for all tables): ")
+	enteredText, _ = reader.ReadString('\n')
+	enteredText = strings.TrimSpace(enteredText)
+	matchTableName, err := regexp.Compile(enteredText)
+	if err != nil {
+		return err
+	}
+
+	fmt.Print("Please input start day (YYYY-MM-DD): ")
```
nhakmiller: Does it make sense to default no date entered to all time? I don't have enough context to know if that makes sense totally.

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/magefile/mage/mg"
+
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/registry"
+	"github.com/panther-labs/panther/pkg/awsglue"
+)
+
+// targets for managing Glue tables
+type Glue mg.Namespace
+
+// Sync's all glue table partitions with current table schema (used when table schema changes)
```
austinbyers: The description is too long in my opinion

```
$ mage

Targets:
  build:api           Generate Go client/models from Swagger specs in api/
  build:lambda        Compile Go Lambda function source
  clean               Remove auto-generated build artifacts
  deploy              Deploy application infrastructure
  fmt                 Format source files
  glue:sync           Sync's all glue table partitions with current table schema (used when table schema changes)
  setup:all           Install all development dependencies
  setup:go            Install goimports, go-swagger, and golangci-lint
  setup:python        Install the Python virtual env
```

We also don't need the apostrophe. So change to something like

```go
// Sync Sync glue table partitions after schema change
```

(The first "Sync" is ignored by mage since it matches the function name)

Please also update the mage output in the documentation: https://docs.runpanther.io/development#mage

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,127 @@
...
+func (t Glue) Sync() error {
+	const dateFormat = "2006-01-02"
+	var enteredText string
+
+	awsSession := session.Must(session.NewSession())
```
austinbyers: There is a `getSession` function in `util.go`, please use that instead:

https://github.com/panther-labs/panther/blob/master/tools/mage/util.go#L69-L93

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,127 @@
...
+	glueClient := glue.New(awsSession)
+
+	reader := bufio.NewReader(os.Stdin)
+
+	fmt.Print("Please input a regular expression to select a subset of tables (or <enter> for all tables): ")
```
austinbyers: For consistency, consider using the `promptUser` method in `util.go` (feel free to adapt to your needs) - all user prompts are currently done this way (integration tests and deploy setup). It handles space trimming and re-prompting the user if there was an error.

For example,

```go
// util.go
func regexValidator(text string) error {
	if err := regexp.Compile(text); err != nil {
            return fmt.Errorf("invalid regex: %v", err)
        }
        return nil
}

// glue_namespace.go
enteredText := promptUser("Enter regex to select a subset of tables (or <enter> for all tables): ", regexValidator)
matchTableName := regexp.MustCompile(enteredText)
```

^ This setup compiles the regex twice, which isn't really a big deal, but if you want to modify the validator style to also return a value, you could

pkg/awsglue/partition.go
rleighton: Yup, good catch. This package just has integration tests, let me add some mocks to exercise all the error paths.

pkg/awsglue/partition.go
rleighton: Yup, good catch.

tools/mage/glue_namespace.go
rleighton: No, that will take a long time and we need a way to figure out the bounds. That could be an Athena query but most users will know the time period. It could be done with a lot of complexity but I would rather not now.

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	fmt.Print("Please input end day (YYYY-MM-DD): ")
+	enteredText, _ = reader.ReadString('\n')
+	enteredText = strings.TrimSpace(enteredText)
+	endDay, err := time.Parse(dateFormat, enteredText)
+	if err != nil {
+		return fmt.Errorf("cannot parse %s as YYYY-MM-DD", enteredText)
+	}
+	endDay = endDay.Add(time.Hour * 23) // move to last hour of the day
```
austinbyers: Similarly, you could have a `dateValidator` function for `promptUser`

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/registry"
+	"github.com/panther-labs/panther/pkg/awsglue"
+)
+
+// targets for managing Glue tables
+type Glue mg.Namespace
+
+// Sync's all glue table partitions with current table schema (used when table schema changes)
+func (t Glue) Sync() error {
```
austinbyers: We don't return errors from top-level mage functions anymore - `logger.Fatal` them instead.

Otherwise, `mage` will print the final error instead of our logging library

pkg/awsglue/partition.go
rleighton: Yup , good catch. This package has only integration tests, let me add some mocks to exercise the error branches.

pkg/awsglue/partition.go
rleighton: Yup, will fix.

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,127 @@
...
+	enteredText, _ = reader.ReadString('\n')
+	enteredText = strings.TrimSpace(enteredText)
+	endDay, err := time.Parse(dateFormat, enteredText)
+	if err != nil {
+		return fmt.Errorf("cannot parse %s as YYYY-MM-DD", enteredText)
+	}
+	endDay = endDay.Add(time.Hour * 23) // move to last hour of the day
+
+	if startDay.After(endDay) {
+		return fmt.Errorf("start day (%s) cannot be after end day (%s)", startDay, endDay)
```
austinbyers:
```go
logger.Fatalf("...")
```

tools/mage/glue_namespace.go
rleighton: That is hard to do because we do not know the bounds. Users will likely know them. We could add that in the future but it adds a lot of complexity.

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// delete and re-create concurrently cuz the Glue API is very slow
+	var wg sync.WaitGroup
+	for i := 0; i < concurrency; i++ {
+		wg.Add(1)
+		go func() {
+			for update := range updateChan {
+				err := update.table.SyncPartition(glueClient, update.at)
+				if err != nil {
+					fmt.Println(err) // best effort, let users know there are failures (this can be re-run)
```
austinbyers:
```go
logger.Errorf("failed to sync partition: %v", err)
```

(zap loggers are concurrency-safe)

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,127 @@
...
+		}()
+	}
+
+	// for each table, for each time partition, delete and re-create
+	for _, table := range registry.AvailableTables() {
+		name := fmt.Sprintf("%s.%s", table.DatabaseName(), table.TableName())
+		if !matchTableName.MatchString(name) {
+			continue
+		}
+		fmt.Printf("sync'ing %s\n", name)
```
austinbyers: We never `fmt.Print` anywhere in mage anymore:

```go
logger.Infof("syncing %s", name)
```

pkg/awsglue/partition.go
rleighton: Done, much tidier.

tools/mage/glue_namespace.go
rleighton: Done

tools/mage/glue_namespace.go
rleighton: Done

pkg/awsglue/partition.go
rleighton: Done

tools/mage/glue_namespace.go
rleighton: Done

tools/mage/glue_namespace.go
rleighton: Done

tools/mage/glue_namespace.go
rleighton: Done

tools/mage/glue_namespace.go
rleighton: Done

tools/mage/glue_namespace.go
rleighton: Done

pkg/awsglue/partition.go
rleighton: Done

tools/mage/glue_namespace.go
rleighton: Done

tools/mage/glue_namespace.go
austinbyers: You still have the `error` return type haha don't need it anymore

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,130 @@
...
+	gluePartitionDateFormat = "2006-01-02"
+)
+
+// Sync Sync glue table partitions after schema change
+func (t Glue) Sync() error {
+	var enteredText string
+
+	awsSession, err := getSession()
+	if err != nil {
+		logger.Fatalf("Error getting session: %s", err)
```
austinbyers: here you can just `logger.Fatal(err)` (the `getSession` function already fills the error with the right context)

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,130 @@
...
+		}()
+	}
+
+	// for each table, for each time partition, delete and re-create
+	for _, table := range registry.AvailableTables() {
+		name := fmt.Sprintf("%s.%s", table.DatabaseName(), table.TableName())
+		if !matchTableName.MatchString(name) {
+			continue
+		}
+		logger.Infof("sync'ing %s", name)
```
austinbyers: Don't need the apostrophe `"syncing"`

tools/mage/glue_namespace.go
rleighton: Done

tools/mage/glue_namespace.go
rleighton: Done

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		logger.Fatal(err)
+	}
+	glueClient := glue.New(awsSession)
+
+	enteredText = promptUser("Enter regex to select a subset of tables (or <enter> for all tables): ", regexValidator)
+	matchTableName, err := regexp.Compile(enteredText)
+	if err != nil {
+		logger.Fatal(err)
+	}
```
kostaspap: You have already validated it, right? Maybe you can skip it (like you do in lines 58 & 61). Or, if you keep it, you should add the check back at lines 58 & 61.

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,128 @@
...
+
+	enteredText = promptUser("Please input start day (YYYY-MM-DD): ", dateValidator)
+	startDay, _ := time.Parse(gluePartitionDateFormat, enteredText) // no error check already validated
+
+	enteredText = promptUser("Please input end day (YYYY-MM-DD): ", dateValidator)
+	endDay, _ := time.Parse(gluePartitionDateFormat, enteredText) // no error check already validated
```
kostaspap: How are people meant to choose the start and end day? If the choose the "wrong" start day, could it be that their tables are unusable? There is a possibility for failure due to wrong user input. Couldn't we be doing the checks ourselves (which columns need to be changed and which not?)
Same for tables. Shouldn't we be fixing all the tables anyway? What if someone forgets to add a table? I was imagining `glue:sync` checking & repairing where needed.

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,128 @@
...
+	syncPartitions(glueClient, matchTableName, startDay, endDay)
+}
+
+func syncPartitions(glueClient *glue.Glue, matchTableName *regexp.Regexp, startDay, endDay time.Time) {
+	const concurrency = 10
+	updateChan := make(chan *gluePartitionUpdate, concurrency)
+
+	// delete and re-create concurrently cuz the Glue API is very slow
+	var wg sync.WaitGroup
+	for i := 0; i < concurrency; i++ {
```
kostaspap: Glue API offers Batch Create and Batch delete partition. You don't need to use concurrency to achieve the same performance, and operation will actually be faster even because Glue can optimize better for batch operations.

tools/mage/glue_namespace.go
```diff
@@ -0,0 +1,128 @@
...
+		for timeBin := startDay; !timeBin.After(endDay); timeBin = table.Timebin().Next(timeBin) {
+			updateChan <- &gluePartitionUpdate{
+				table: table,
+				at:    timeBin,
+			}
+		}
+	}
+
+	close(updateChan)
+	wg.Wait()
```
kostaspap: An informative message would be useful at the end, to give people summary of what happened (e.g. updated X partitions in Y table from T1 to T2). Could be useful feedback, especially given users can provide a regex as input (which might be wrong regex)

tools/mage/glue_namespace.go
rleighton: True, I missed that one when I did the others. Maybe for another PR when we are in this code.

tools/mage/glue_namespace.go
rleighton: Let's talk on Slack. This is super safe and some things you said don't make sense to me.

tools/mage/glue_namespace.go
rleighton: Actually, that would add a lot of complexity to the code (batching) and this is not an often run utility. I don't think it is worth it. Also, I would actually measure batch vs concurrent performance. If I had to bet, the Glue team runs a batch serially per call. I would bet that Golang running individual API calls, which get done in parallel would be faster. We would need to measure and this case it is just not worth it. Runs fast and code is simple.

tools/mage/glue_namespace.go
rleighton: The code logs each table as it goes so user gets feedback. Adding stats is a nice to have but does not add much value. If it was successful, then user knows all partitions are updated in the time interval.

PR #165, opened 2020-02-11
Update AWS IAM Service principals

gavinelder:
## Background

The following is a fix for an earlier PR #61 however it turns out AWS IAM service principals do not use the regional suffix. 

## Changes

- Update IAM Service principals to use hardcoded values.

## Testing

- Confirmed IAM SP within AWS CN.

### DIFF
--- a/deployments/compliance/compliance_api.yml
+++ b/deployments/compliance/compliance_api.yml
@@ -108,7 +108,7 @@ Resources:
     Properties:
       Action: lambda:InvokeFunction
       FunctionName: !Ref Function
-      Principal: !Sub apigateway..amazonaws.com
+      Principal: apigateway.amazonaws.com
       SourceArn: !Sub arn:${AWS::Partition}:execute-api:${AWS::Region}:${AWS::AccountId}:${GatewayApi}/*
 
   ComplianceTable:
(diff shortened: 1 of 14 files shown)


deployments/compliance/compliance_api.yml
```diff
@@ -108,7 +108,7 @@ Resources:
     Properties:
       Action: lambda:InvokeFunction
       FunctionName: !Ref Function
-      Principal: !Sub apigateway.${AWS::URLSuffix}
+      Principal: !Sub apigateway..amazonaws.com
```
nhakmiller: Extra `.` here I think?

deployments/compliance/snapshot.yml
```diff
@@ -189,7 +189,7 @@ Resources:
         Variables:
           DEBUG: !Ref Debug
           SNAPSHOT_POLLERS_QUEUE_URL: !Ref Queue
-          LOG_PROCESSOR_QUEUE_URL: !Sub https://sqs.${AWS::Region}.${AWS::URLSuffix}/${AWS::AccountId}/${LogAnalysisQueueName}
+          LOG_PROCESSOR_QUEUE_URL: !Sub https://sqs.${AWS::Region}..amazonaws.com/${AWS::AccountId}/${LogAnalysisQueueName}
```
nhakmiller: Another extra `.`

PR #167, opened 2020-02-11
Support errorReportingConsent

### LINKED ISSUE #152: [Feature] Store crash log reporting preference
## Introduction
In the front-end, whenever a runtime exception occurs, it would be beneficial for the Panther team to be able to receive an alert for it.

As a user, I want to be able to enable or disable crash reporting. I should be able to opt-in or out of this behavior through a switch in the general settings.

This task involves the creation of the necessary storage entity, in order to store the user's preference.

## Implementation

The proposal is as follows:

A single boolean-like value in the `organization-api` which would have a name similar to `forwardErrorLogs`, `submitErrorLogs`, `allowErrorLogSubmission`, etc.

This value will accept the following values:

`True` - If the user (org) chose to allow crash reporting
`False` - If the user (org) chose to not partake in the reporting
`nil` - If the user hasn't yet submitted yet with a decision (this is the initial value).

This single value will be:

* Returned as part of the `GetOrganization` API endpoint:

```go
GetOrganization    *GetOrganizationInput    `json:"getOrganization"`
```

* Updated as part of the `UpdateOrganization` API endpoint:

```go
UpdateOrganization *UpdateOrganizationInput `json:"updateOrganization"`
```

austinbyers:
## Background
Provides backend support for users to opt into error reporting.

Closes #152 

## Changes

- Support `errorReportingConsent` in the org database
    - I wanted to put "consent" in the name so it was explicitly clear that the field is populated by user consent
    - This is just for frontend errors logs, right? We could in theory also send backend error logs, but that would take more work. Should the same consent flag apply to both? I don't see why not
- Hide `docker push` output by default (per @3nvi suggestion), which is more consistent with the rest of the deploy process

## Testing

- Unit tests
- Deploy + integration test

### DIFF
--- a/api/lambda/organization/models/api.go
+++ b/api/lambda/organization/models/api.go
@@ -38,11 +38,12 @@ type CompleteActionOutput struct {
 
 // CreateOrganizationInput creates a new Panther customer account.
 type CreateOrganizationInput struct {
-	AlertReportFrequency *string    `json:"alertReportFrequency" validate:"omitempty,oneof=P1D P1W"`
-	AwsConfig            *AwsConfig `json:"awsConfig"`
-	DisplayName          *string    `json:"displayName" validate:"required,min=1"`
-	Email                *string    `genericapi:"redact" json:"email" validate:"required,email"`
-	Phone                *string    `genericapi:"redact" json:"phone"`
+	AlertReportFrequency  *string    `json:"alertReportFrequency" validate:"omitempty,oneof=P1D P1W"`
+	AwsConfig             *AwsConfig `json:"awsConfig"`
+	DisplayName           *string    `json:"displayName" validate:"required,min=1"`
+	Email                 *string    `genericapi:"redact" json:"email" validate:"required,email"`
+	ErrorReportingConsent *bool      `json:"errorReportingConsent"`
+	Phone                 *string    `genericapi:"redact" json:"phone"`
 }
 
 // CreateOrganizationOutput returns the newly created organization.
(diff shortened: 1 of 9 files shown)


api/lambda/organization/models/api.go
```diff
@@ -38,11 +38,12 @@ type CompleteActionOutput struct {
...
-	AlertReportFrequency *string    `json:"alertReportFrequency" validate:"omitempty,oneof=P1D P1W"`
-	AwsConfig            *AwsConfig `json:"awsConfig"`
-	DisplayName          *string    `json:"displayName" validate:"required,min=1"`
-	Email                *string    `genericapi:"redact" json:"email" validate:"required,email"`
-	Phone                *string    `genericapi:"redact" json:"phone"`
+	AlertReportFrequency  *string    `json:"alertReportFrequency" validate:"omitempty,oneof=P1D P1W"`
+	AwsConfig             *AwsConfig `json:"awsConfig"`
+	DisplayName           *string    `json:"displayName" validate:"required,min=1"`
+	Email                 *string    `genericapi:"redact" json:"email" validate:"required,email"`
+	ErrorReportingConsent *bool      `json:"errorReportingConsent"`
```
3nvi: I'm not sure whether we should require it as input during the organization creation since it should be `nil` by default. That's your choice :)

tools/mage/deploy_frontend.go
```diff
@@ -108,7 +108,7 @@ func buildAndPushImageFromSource(awsSession *session.Session, imageRegistry stri
 	}
 
 	logger.Info("deploy: pushing docker image to remote repo")
-	if err := sh.RunV("docker", "push", remoteImage); err != nil {
+	if err := sh.Run("docker", "push", remoteImage); err != nil {
```
3nvi: Saw what you did there :P

api/lambda/organization/models/api.go
austinbyers: It's not required - you can omit the field and it will be `nil` by default

PR #168, opened 2020-02-11
S3accesslogs do not require httpstatus

rleighton:
## Background
Monitoring detected a failed log parse for s3 access logs. Looking into it a 'S3.EXPIRE.OBJECT' event was not able to be parsed because we require an http status code be set. This seems to be a backend triggered transaction that is logged that has no http associated attributes.

## Changes
Removed required validations for HttpStatus field.

## Testing
mage test:ci
deployed to dev account

### DIFF
--- a/internal/log_analysis/log_processor/parsers/awslogs/s3_server_access.go
+++ b/internal/log_analysis/log_processor/parsers/awslogs/s3_server_access.go
@@ -46,7 +46,7 @@ type S3ServerAccess struct {
 	Operation          *string            `json:"operation,omitempty" description:"The operation listed here is declared as SOAP.operation, REST.HTTP_method.resource_type, WEBSITE.HTTP_method.resource_type, or BATCH.DELETE.OBJECT."`
 	Key                *string            `json:"key,omitempty" description:"The key part of the request, URL encoded, or NULL if the operation does not take a key parameter."`
 	RequestURI         *string            `json:"requesturi,omitempty" description:"The Request-URI part of the HTTP request message."`
-	HTTPStatus         *int               `json:"httpstatus,omitempty" validate:"required,max=600,min=100" description:"The numeric HTTP status code of the response."`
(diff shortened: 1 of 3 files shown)


internal/log_analysis/log_processor/parsers/awslogs/s3_server_access.go
```diff
@@ -46,7 +46,7 @@ type S3ServerAccess struct {
 	Operation          *string            `json:"operation,omitempty" description:"The operation listed here is declared as SOAP.operation, REST.HTTP_method.resource_type, WEBSITE.HTTP_method.resource_type, or BATCH.DELETE.OBJECT."`
 	Key                *string            `json:"key,omitempty" description:"The key part of the request, URL encoded, or NULL if the operation does not take a key parameter."`
 	RequestURI         *string            `json:"requesturi,omitempty" description:"The Request-URI part of the HTTP request message."`
-	HTTPStatus         *int               `json:"httpstatus,omitempty" validate:"required,max=600,min=100" description:"The numeric HTTP status code of the response."`
+	HTTPStatus         *int               `json:"httpstatus,omitempty" description:"The numeric HTTP status code of the response."`
```
kostaspap: I think you should leave the number check in there (it's HTTP status after all) but just put a `omitempty` instead of `required`.

rleighton:
I tried that, does not work when value is missing, unless there is special
syntax

internal/log_analysis/log_processor/parsers/awslogs/s3_server_access.go
kostaspap: `omitempty` has to be before the rest of the checks. So it should be 
`validate:"omitempty,max=600,min=100"`. If the `omitempty` is placed after the max/min, in case the value is empty the validation will fail.

rleighton:
Ah, that would be ideal. Let my try it. I just took off the required

PR #169, opened 2020-02-11
ECS Cluster Scanning Support

### LINKED ISSUE #166: Add Scanning Support for ECS
Is your feature request related to a problem? Please describe.

Add support for ECS Cluster scanning to Cloud Security. ECS (an AWS managed container service) is a fairly popular service. We should support ECS for compliance scanning.

Describe the solution you'd like

I would like the Cloud Security suite to scan ECS Clusters, and include some additional information about tasks and services associated with that cluster.

Describe alternatives you've considered

N/A

Additional context

None

nhakmiller:
## Background

This change adds support for ECS Cluster scanning to the Cloud Security suite. Closes #166.

## Changes

- Updated the snapshot poller to be able to scan ECS Clusters
- Updated the aws event processor to be able to process ECS API calls
- Updated the frontend constants list to include `AWS.ECS.Cluster`
- A few minor tweaks of various inconsistencies on some unrelated aws event processor and snapshot poller code
- Removed the `ServicePollersOrdered` slice, it was just place we were duplicating work for a feature that has been deprecated (feature is still supported, it's just less efficient now but again its deprecated)

## Testing

- Manually tested in my dev account the following activities:
  - Tested with both long and short names (full ARNs and just the name)
  - Account wide ECS scans
  - Updating tags on an ECS Cluster
  - Updating the service associated to an ECS Cluster
  - Creating a new service for an ECS Cluster
  - Deleting a service for an ECS Cluster
  - Creating a cluster
- Unit testing of the snapshot poller

### DIFF
--- a/internal/compliance/aws_event_processor/processor/ec2.go
+++ b/internal/compliance/aws_event_processor/processor/ec2.go
@@ -316,7 +316,7 @@ func classifyEC2(detail gjson.Result, accountID string) []*resourceChange {
 		// security group scan.
 		return []*resourceChange{{
 			AwsAccountID: ec2ARN.AccountID,
-			Delete:       deleteResource,
+			Delete:       false,
 			EventName:    eventName,
 			Region:       region,
 			ResourceType: ec2Type,
@@ -332,7 +332,7 @@ func classifyEC2(detail gjson.Result, accountID string) []*resourceChange {
 		// VPC Sub resources
 		return []*resourceChange{{
 			AwsAccountID: ec2ARN.AccountID,
-			Delete:       deleteResource,
+			Delete:       false,
 			EventName:    eventName,
 			Region:       region,
 			ResourceType: aws.Ec2VpcSchema,
@@ -341,7 +341,7 @@ func classifyEC2(detail gjson.Result, accountID string) []*resourceChange {
 		// Volume sub resources
 		return []*resourceChange{{
 			AwsAccountID: ec2ARN.AccountID,
-			Delete:       deleteResource,
+			Delete:       false,
 			EventName:    eventName,
 			Region:       region,
 			ResourceType: aws.Ec2VolumeSchema,
@@ -350,7 +350,7 @@ func classifyEC2(detail gjson.Result, accountID string) []*resourceChange {
 		// Instance sub resources
 		return []*resourceChange{{
 			AwsAccountID: ec2ARN.AccountID,
-			Delete:       deleteResource,
+			Delete:       false,
 			EventName:    eventName,
 			Region:       region,
 			ResourceType: aws.Ec2InstanceSchema,
@@ -366,7 +366,7 @@ func classifyEC2(detail gjson.Result, accountID string) []*resourceChange {
 			// currently so we do a region wide EC2 VPC scan.
 			return []*resourceChange{{
 				AwsAccountID: ec2ARN.AccountID,
-				Delete:       deleteResource,
+				Delete:       false,
 				EventName:    eventName,
 				Region:       region,
 				ResourceType: aws.Ec2VpcSchema,
@@ -379,7 +379,7 @@ func classifyEC2(detail gjson.Result, accountID string) []*resourceChange {
 		// EC2 VPC scan.
 		return []*resourceChange{{
 			AwsAccountID: ec2ARN.AccountID,
-			Delete:       deleteResource,
+			Delete:       false,
 			EventName:    eventName,
 			ResourceType: aws.Ec2VpcSchema,
 		}}
--- a/internal/compliance/aws_event_processor/processor/ecs.go
+++ b/internal/compliance/aws_event_processor/processor/ecs.go
@@ -88,8 +88,17 @@ func classifyECS(detail gjson.Result, accountID string) []*resourceChange {
 		return nil
 	}
 
-	// If the short name was provided, convert to full ARN
+	// If clusterARN is empty, we failed to parse the ARN out at some point despite trying
+	if clusterARN == "" {
+		zap.L().Error("ecs: known event name, but still failed to parse clusterARN", zap.String("eventName", eventName))
+		return nil
+	}
+
+	// All ECS Cluster API calls can be made with the full ARN or just the cluster name as the 'cluster' parameter.
+	// If we received just the cluster name we can construct the full ARN, which we do in order to reduce
+	// complexity for the snapshot poller.
 	if _, err := arn.Parse(clusterARN); err != nil {
+		// A short cluster name was provided, construct the full ARN
 		clusterARN = arn.ARN{
 			Partition: "aws",
 			Service:   "ecs",
--- a/internal/compliance/aws_event_processor/processor/ecs.go
+++ b/internal/compliance/aws_event_processor/processor/ecs.go
@@ -19,10 +19,10 @@ package processor
  */
 
 import (
-	"github.com/pkg/errors"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/pkg/errors"
 	"github.com/tidwall/gjson"
 	"go.uber.org/zap"
 
--- a/internal/compliance/aws_event_processor/processor/ecs.go
+++ b/internal/compliance/aws_event_processor/processor/ecs.go
@@ -19,6 +19,7 @@ package processor
  */
 
 import (
+	"github.com/pkg/errors"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
@@ -62,6 +63,7 @@ func classifyECS(detail gjson.Result, accountID string) []*resourceChange {
 				"ecs: unable to parse resource ARN",
 				zap.String("eventName", eventName),
 				zap.String("resource ARN", clusterARN),
+				zap.Error(errors.WithStack(err)),
 			)
 			return nil
 		}
--- a/internal/compliance/aws_event_processor/processor/ecs.go
+++ b/internal/compliance/aws_event_processor/processor/ecs.go
@@ -32,11 +32,6 @@ func classifyECS(detail gjson.Result, accountID string) []*resourceChange {
 	eventName := detail.Get("eventName").Str
 
 	// https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazonelasticcontainerservice.html
-	if eventName == "READONLY" {
-		zap.L().Debug("ecs: ignoring event", zap.String("eventName", eventName))
-		return nil
-	}
-
 	var clusterARN string
 	switch eventName {
 	case "CreateTaskSet", "DeleteCluster", "DeleteTaskSet", "UpdateServicePrimaryTaskSet", "UpdateTaskSet":
@@ -44,10 +39,11 @@ func classifyECS(detail gjson.Result, accountID string) []*resourceChange {
 	case "CreateService", "DeleteAttributes", "DeleteService", "DeregisterContainerInstance", "PutAttributes",
 		"RegisterContainerInstance", "RunTask", "StartTask", "StopTask", "SubmitAttachmentStateChanges", "SubmitContainerStateChange",
 		"SubmitTaskStateChange", "UpdateContainerInstancesState", "UpdateService":
-		// These API calls understand an absent cluster value to mean the default cluster.
+		// These API calls interpret a missing cluster value to mean the default cluster.
 		//
 		// Sadly we can't differentiate between a failed extraction (rare), and a request to modify the
-		// default cluster. We will just default to scanning the default cluster.
+		// default cluster. We will just default to scanning the default cluster, which is the cluster
+		// with the name "default".
 		clusterARN = detail.Get("requestParameters.cluster").Str
 		if clusterARN == "" {
 			clusterARN = "default"
@@ -73,7 +69,7 @@ func classifyECS(detail gjson.Result, accountID string) []*resourceChange {
 			break
 		}
 
-		// It wasn't a cluster, so we have to scan the whole region.
+		// If it wasn't a cluster, we have to scan the whole region.
 		return []*resourceChange{{
 			AwsAccountID: accountID,
 			Delete:       false,
--- a/internal/compliance/aws_event_processor/processor/ecs.go
+++ b/internal/compliance/aws_event_processor/processor/ecs.go
@@ -31,7 +31,7 @@ import (
 func classifyECS(detail gjson.Result, accountID string) []*resourceChange {
 	eventName := detail.Get("eventName").Str
 
-	// https://docs.aws.amazon.com/IAM/latest/UserGuide/list_awscertificatemanager.html
+	// https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazonelasticcontainerservice.html
 	if eventName == "READONLY" {
 		zap.L().Debug("ecs: ignoring event", zap.String("eventName", eventName))
 		return nil
@@ -69,7 +69,7 @@ func classifyECS(detail gjson.Result, accountID string) []*resourceChange {
 			)
 			return nil
 		}
-		if strings.HasPrefix("cluster", parsed.Resource) {
+		if strings.HasPrefix(parsed.Resource, "cluster") {
 			break
 		}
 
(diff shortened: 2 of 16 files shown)


internal/compliance/aws_event_processor/processor/ec2.go
```diff
@@ -316,7 +316,7 @@ func classifyEC2(detail gjson.Result, accountID string) []*resourceChange {
 		// security group scan.
 		return []*resourceChange{{
 			AwsAccountID: ec2ARN.AccountID,
-			Delete:       deleteResource,
+			Delete:       false,
```
nhakmiller: `deleteResource` will always be false here anyways (same in the below changes).

internal/compliance/aws_event_processor/processor/ecs.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+func classifyECS(detail gjson.Result, accountID string) []*resourceChange {
+	eventName := detail.Get("eventName").Str
+
+	// https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazonelasticcontainerservice.html
+	if eventName == "READONLY" {
+		zap.L().Debug("ecs: ignoring event", zap.String("eventName", eventName))
+		return nil
+	}
```
nhakmiller: Boilerplate, will remove this shortly.

internal/compliance/aws_event_processor/processor/ecs.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil
+	default:
+		zap.L().Warn("ecs: encountered unknown event name", zap.String("eventName", eventName))
+		return nil
+	}
+
+	// If the short name was provided, convert to full ARN
+	if _, err := arn.Parse(clusterARN); err != nil {
+		clusterARN = arn.ARN{
+			Partition: "aws",
```
nhakmiller: This pattern is somewhat frequent in the aws event processor, but I just realized while implementing this that it will not work in the cn or gov partitions. For now I am ok with that because we do not officially support those partitions anyways, but I think I will add a story to make the aws event processor partition aware.

internal/compliance/snapshot_poller/models/aws/ecs_cluster.go
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+const (
+	EcsClusterSchema = "AWS.ECS.Cluster"
+)
+
+// EcsCluster contains all the information about an ACM certificate
```
nhakmiller: Typo.

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
```diff
@@ -0,0 +1,339 @@
...
+			},
+			CapacityProviderStrategy:      service.CapacityProviderStrategy,
+			TimeCreated:                   utils.DateTimeFormat(aws.TimeValue(service.CreatedAt)),
+			CreatedBy:                     service.CreatedBy,
+			DeploymentConfiguration:       service.DeploymentConfiguration,
+			DeploymentController:          service.DeploymentController,
+			Deployments:                   service.Deployments,
+			DesiredCount:                  service.DesiredCount,
+			EnableECSManagedTags:          service.EnableECSManagedTags,
+			Events:                        service.Events,
```
nhakmiller: This field can contain at least 100 events. In my testing I only ever saw it containing exactly 100 events so I would suspect that is the limit. 100 events is a lot, but not an unreasonable amount so I left them all in. We might consider reducing this to only the five most recent events.

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
```diff
@@ -0,0 +1,339 @@
...
+	return ecsCluster
+}
+
+// PollEcsCluster gathers information on each ECS Cluster for an AWS account.
+func PollEcsClusters(pollerInput *awsmodels.ResourcePollerInput) ([]*apimodels.AddResourceEntry, error) {
+	zap.L().Debug("starting ECS Cluster resource poller")
+	ecsClusterSnapshots := make(map[string]*awsmodels.EcsCluster)
+
+	for _, regionID := range utils.GetServiceRegions(pollerInput.Regions, "ecs") {
+		sess := session.Must(session.NewSession(&aws.Config{Region: regionID}))
```
rleighton: session.Must() panics on error. Are you ok with that?

internal/compliance/aws_event_processor/processor/ecs.go
```diff
@@ -0,0 +1,113 @@
...
+		// If we add an ECS account wide resource or an ECS TaskDefinition resource we can use these
+		return nil
+	default:
+		zap.L().Warn("ecs: encountered unknown event name", zap.String("eventName", eventName))
+		return nil
+	}
+
+	// If the short name was provided, convert to full ARN
+	if _, err := arn.Parse(clusterARN); err != nil {
+		clusterARN = arn.ARN{
```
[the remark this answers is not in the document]
nhakmiller: Use the parsed values where? The inside of the `if` block only executes if the parsing failed (in which case we can't use the parsed values), and the return struct is expecting a string (the return struct is shared amongst all processors) so saving the parsed values just to convert back into a string seems unnecessary.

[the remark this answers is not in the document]
rleighton: Sorry, mis-read it.

internal/compliance/aws_event_processor/processor/ecs.go
```diff
@@ -0,0 +1,113 @@
...
+		// Nothing to do here
+		//
+		// If we add an ECS account wide resource or an ECS TaskDefinition resource we can use these
+		return nil
+	default:
+		zap.L().Warn("ecs: encountered unknown event name", zap.String("eventName", eventName))
+		return nil
+	}
+
+	// If the short name was provided, convert to full ARN
```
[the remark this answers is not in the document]
nhakmiller: Good question, there are actually three cases once we get to this piece of the code.

1. `clusterARN` is an empty string. This case will only happen if we failed to parse out the `cluster` field somewhere, but we actually don't handle that case currently and if it does occur it will cause an error in the snapshot poller. Thanks for asking this question and making me realize this, patch incoming for this case!
2. `clusterARN` is a full ARN of an ECS Cluster. In this case, it gets parsed properly by `arn.Parse` and we skip past the rest of this `if` block and just return it
3. `clusterARN` is an ECS Cluster name. In this case, it fails to get properly parsed and so we construct the ARN since we know all the pieces. `clusterARN` being `default` is actually just a special case of this, the default ECS cluster is always named `default` (that is in fact the definition of the default ECS Cluster).

Cases 2 and 3 we are handling correctly I believe, it is just case 1 where I'm realizing we are not handling it as cleanly as we could be. Do you think this deserves a bigger comment explaining all that? Since I'm going back in to handle case 1 I might as well add it.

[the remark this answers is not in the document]
nhakmiller: Additionally, since the API's all accept both short names and full ARNs, we have no way of differentiating between case 2 and case 3 by looking at anything besides whether or not it is parsable as an ARN (we cannot infer from e.g. the `eventName` or any other field).

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
nhakmiller: I've just always used `session.Must()` when processing cannot continue without the session. My understanding is that lambda will just retry the function. Do you think it's better to use a call that doesn't panic on error, and manually log an error? I don't really want to continue scanning other regions necessarily because I want the lambda function to re-run.

rleighton: Can it happen that 1 region keeps failing on the session but the others _would_ work? Then a panic would block all processing if you hit the bad region before the good regions. I personally reserve panics for "the sw is broken" cases and try to handle the errors explicitly. In this case I think checking the error, if not nil , log and continue is more robust. That said, I'm not sure how the session can really fail ... I thought it was lazy and you normally hit errors on the first API you use so it may not matter as practical matter but from a best practice IMO, panics are reserved for very special cases.

internal/compliance/snapshot_poller/pollers/aws/ec2_volume.go
```diff
@@ -156,7 +156,7 @@ func buildEc2VolumeSnapshot(ec2Svc ec2iface.EC2API, volume *ec2.Volume) *awsmode
 
 	snapshots := describeSnapshots(ec2Svc, volume.VolumeId)
 	if snapshots != nil {
-		ec2Volume.Snapshots = make([]*awsmodels.Ec2Snapshot, len(snapshots))
+		ec2Volume.Snapshots = make([]*awsmodels.Ec2Snapshot, 0, len(snapshots))
```
austinbyers: 👍

internal/compliance/aws_event_processor/processor/ecs.go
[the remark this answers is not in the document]
rleighton: Yeah, if you just copied all the knowledge from the above into a comment that would he super helpful to the next person in.

internal/compliance/aws_event_processor/processor/ecs.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazonelasticcontainerservice.html
+	if eventName == "READONLY" {
+		zap.L().Debug("ecs: ignoring event", zap.String("eventName", eventName))
+		return nil
+	}
+
+	var clusterARN string
+	switch eventName {
+	case "CreateTaskSet", "DeleteCluster", "DeleteTaskSet", "UpdateServicePrimaryTaskSet", "UpdateTaskSet":
```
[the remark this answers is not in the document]
nhakmiller: They're defined as constants in the SDK, but not exported so we cannot reference them.

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
nhakmiller: Ok, I will update this here but everywhere else in the snapshot poller we use `session.Must()`. I've had a task in my backlog for months to unify how the snapshot poller handles errors, so I will include this as part of that.

internal/compliance/aws_event_processor/processor/ecs.go
```diff
@@ -0,0 +1,118 @@
...
+		// Since we don't know who the parent resource is that changed, we have to scan all resources
+
+		// In the case of clusters at least we can continue
+		clusterARN = detail.Get("requestParameters.resourceArn").Str
+		parsed, err := arn.Parse(clusterARN)
+		if err != nil {
+			zap.L().Error(
+				"ecs: unable to parse resource ARN",
+				zap.String("eventName", eventName),
+				zap.String("resource ARN", clusterARN),
```
rleighton: zap.Error(err)

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
```diff
@@ -0,0 +1,365 @@
...
+		PendingTasksCount:                 details.PendingTasksCount,
+		RegisteredContainerInstancesCount: details.RegisteredContainerInstancesCount,
+		RunningTasksCount:                 details.RunningTasksCount,
+		Settings:                          details.Settings,
+		Statistics:                        details.Statistics,
+		Status:                            details.Status,
+	}
+
+	ecsCluster.Tasks, err = getClusterTasks(ecsSvc, details.ClusterArn)
+	if err != nil {
```
rleighton: Needs logging

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
```diff
@@ -0,0 +1,365 @@
...
+		return nil
+	}
+
+	ecsCluster.Services, err = getClusterServices(ecsSvc, details.ClusterArn)
+	if err != nil {
```
rleighton: Needs logging

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
```diff
@@ -0,0 +1,365 @@
...
+		}
+
+		creds, err := AssumeRoleFunc(pollerInput, sess)
+		if err != nil {
+			// The client failed to create, log an error and continue to the next region
+			zap.L().Error(
+				"unable to create aws client",
+				zap.String("region", *regionID),
+				zap.String("service", "ecs"),
+				zap.Error(err),
```
rleighton: errors.Wrap(err)

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
```diff
@@ -0,0 +1,365 @@
...
+
+	for _, regionID := range utils.GetServiceRegions(pollerInput.Regions, "ecs") {
+		sess, err := session.NewSession(&aws.Config{Region: regionID})
+		if err != nil {
+			// The session failed to create, log an error and continue to the next region
+			zap.L().Error(
+				"unable to create aws session",
+				zap.String("region", *regionID),
+				zap.String("service", "ecs"),
+				zap.Error(err),
```
rleighton: errors.Wrap(err)

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
```diff
@@ -0,0 +1,365 @@
...
+	rawServices, err := ecsSvc.DescribeServices(&ecs.DescribeServicesInput{
+		Cluster: clusterArn,
+		// This only accepts one argument, which is the string TAGS
+		// Indicates that we want to included the task tags
+		Include:  []*string{aws.String("TAGS")},
+		Services: serviceArns,
+	})
+
+	if err != nil {
+		utils.LogAWSError("ECS.DescribeServices", err)
```
rleighton: This is different from other logging, why? 
Also errors.Wrap(err)

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
```diff
@@ -0,0 +1,365 @@
...
+	// Enumerate services
+	var serviceArns []*string
+	err := ecsSvc.ListServicesPages(&ecs.ListServicesInput{Cluster: clusterArn},
+		func(page *ecs.ListServicesOutput, lastPage bool) bool {
+			serviceArns = append(serviceArns, page.ServiceArns...)
+			return true
+		})
+
+	if err != nil {
+		utils.LogAWSError("ECS.ListServicesPages", err)
```
rleighton: This is different from other logging, why?
Also errors.Wrap(err)

nhakmiller: This is a utility function that automatically extracts the error code and error message from AWS API errors. We use this throughout the snapshot poller to try to have more standardized error logging. I'm thinking of refactoring it a bit, at the very least I can log `errors.Wrap(err)` as well in the utility function which should help this.

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
nhakmiller: Logged in the called function, I will update how error handling is done across the snapshot poller in a future PR per our Slack discussion.

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
nhakmiller: I added `errors.WithStack()` because I think the message portion is sufficiently covered by the other zap fields.

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
nhakmiller: I added errors.WithStack(err) because I think the message portion is sufficiently covered by the other zap fields.

austinbyers: `errors.Wrap` includes the stack, essentially the same thing

internal/compliance/snapshot_poller/pollers/utils/logging.go
```diff
@@ -30,6 +31,7 @@ func LogAWSError(apiCall string, err error) {
 			apiCall,
 			zap.String("errorCode", awsErr.Code()),
 			zap.String("errorMessage", awsErr.Message()),
+			zap.Error(errors.Wrap(err, "AWS API call failed")),
```
nhakmiller: @rleighton this should update the majority of the snapshot pollers to include stack traces when logging AWS specific errors.

internal/compliance/aws_event_processor/processor/ecs.go
austinbyers: We could use the AWS_REGION [environment variable](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime) in the Lambda function to figure out what the partition should be

nhakmiller: Good implementation suggestion, I will add it to the issue I've opened.

PR #170, opened 2020-02-12
[Feat] Error logging mechanism

### LINKED ISSUE #122: Create mechanism for front-end error reporting
## Introduction

As a developer, I want to be notified of crashes or front-end runtime exceptions from Panther users, if they have opted-in to this behavior. This is gonna help us catch bugs quicker and improve the product

## Implementation

We should:

1. Upload each version's source maps on sentry
2. Allow users to opt-in or out of error logs (through a Modal).
3. Allow users to modify their preference through their settings
4. Build a mechanism that is going to send the logs

3nvi:
## Background

This PR adds error logging to our front-end web platform, by making sure to report JS crashes to Sentry. Up until now, reporting has been disabled due to missing ENV variables that this PR makes sure to add.
Error reporting is now done for every user, but a following PR will make sure to restrict it to users that opt-in to that. 

The real struggle with this ticket had to do with the uploading of the source files, since all of our JS is minified and thus all of the reported errors are minified as well. Sentry required the original source files in order to translate the minified error and point it to the actual line in the source files.
This proved to be difficult, since each deployment has different source files (since the ENV variables that are contained within the source files are different for each organization). This meant that each deployment had different hashed file names and Sentry was not able to understand that all of these deployments have the same code with just some variables modified.
This issue would have been easily fixed if we asked or forced users to upload the sourcemaps after their deployment, but that's not feasible. Again, we couldn't upload the sourcemaps ourselves since they are **not the same for each organization** from Sentry's perspective.

What I ended up doing is allowed sentry to query the source file mappings from the URL that the error was thrown in. That means that the server, apart from the normal JS files, will now also serve the minified-to-source-file mappings. Normally, this can be an issue for some companies, since they expose the un-minified version of their front-end source code, but since we are lucky enough to be open source, this is not an issue for us. The upside to this approach is **much less code** since I don't have to do any uploading of releases to sentry, but the downside is that if the user deploys the app to their localhost, then Sentry won't be able to query to get the proper source files mappings. This shouldn't matter that much though, since we don't expect a localhost deployment to be a norm. 

## Changes

- Add project versioning through `git`
- Add sourcemap generation & serving
- Remove redundant variables from Dockerfile
- Add optimization to express on remote deployments
- Split `build` and `start` commands into corresponding NodeJS files
- Remove user data logging during an error
- Closes #122 

## Testing

Both locally & remotely, I manually threw a JS error and sentry was able to catch it both times:
 - On the remote deployment, sentry showed the source code that the error was located in
 - On the local deployment, sentry showed the type of error but couldn't get the proper mappings to show us where the error is located in (due to the fact that it can't reach my localhost)


.github/ISSUE_TEMPLATE/epic.md
```diff
@@ -1,4 +1,5 @@
 ---
+
```
3nvi: changes from `prettier` tool

deployments/web/Dockerfile
```diff
@@ -22,7 +22,7 @@ FROM node:13.2-alpine AS build-env
 LABEL description="The image that builds the web code and serves it through an express NodeJS server"
 
 # Update debian distro
-RUN apk update && apk add --no-cache --update nasm libtool alpine-sdk autoconf automake
+RUN apk update && apk add --no-cache --update git nasm libtool alpine-sdk autoconf automake
```
3nvi: we need `git` since `npm run build` is going to get  the project version by utilizing `git`

deployments/web/Dockerfile
```diff
@@ -38,16 +38,10 @@ RUN npm i
 
 # Mount the necessary source code
 ADD .babelrc tsconfig.json webpack.config.js ./
-ADD out/.env out/.env
+ADD .git .git
```
3nvi: we need to add the package's git history, to properly calculate the version of the project

deployments/web/Dockerfile
```diff
@@ -38,16 +38,10 @@ RUN npm i
...
+ADD out/.env.aws out/.env.aws
```
3nvi: small rename  from `.env` to  `.env.aws`

deployments/web/Dockerfile
```diff
@@ -38,16 +38,10 @@ RUN npm i
...
 ADD web web
 
-# Build argument & ENV var for versioning
-ARG PANTHER_VERSION
```
3nvi: `PANTHER_VERSION` was not used as a build argument, but was plan to be used like that. This is no longer truee

deployments/web/Dockerfile
```diff
@@ -38,16 +38,10 @@ RUN npm i
...
-ENV PANTHER_VERSION=$PANTHER_VERSION
-
-# Convert NODE_ENV to production in order to adjust webpack configuration accordingly
-ENV NODE_ENV=production
```
3nvi: un-necessary ENV since `npm run build` automatically adds that for us

deployments/web/Dockerfile
```diff
@@ -70,7 +64,7 @@ COPY --from=build-env /code/web/dist web/dist
 
 # Add the script that runs the express server
 ADD package.json .
-ADD  web/scripts/serve.js web/scripts/serve.js
+ADD  web/scripts web/scripts
```
3nvi: now adding **all the scripts** present under `web/scripts`, since we need them all now

web/scripts/utils.js
```diff
@@ -0,0 +1,39 @@
...
+
+function loadDotEnvVars(path) {
+  const dotenvResult = dotenv.config({ path });
+  if (dotenvResult.error) {
+    throw new Error(chalk.red(dotenvResult.error));
+  }
+}
+
+function getPantherDeploymentVersion() {
+  try {
```
3nvi: I calculate the version of the project using `git describe --tags`. This:

* Returns `v0.2.0` if the user is on the latest release and hasn't done any manual commits 
* Returns `v0.2.0-X-YYYYY` where `X` is the number of commits **past** the version `v0.2.0` and `YYYYY` is the latest commit hash that the user is on. This is helpful, since we know whether the user is on a clean deployment when he encountered an error or whether he tampered with the code

web/webpack.config.js
```diff
@@ -39,11 +38,7 @@ module.exports = {
   // Stop compilation early in production, saving time
   bail: isEnvProduction,
   // add a proper source map in order to debug the code easier through the sources tab.
-  devtool: isEnvProduction // eslint-disable-line no-nested-ternary
-    ? shouldUseSourceMap
-      ? 'source-map'
-      : false
-    : isEnvDevelopment && 'cheap-module-source-map',
+  devtool: isEnvProduction ? 'source-map' : isEnvDevelopment && 'cheap-module-source-map',
```
3nvi: always create source maps in production from now on

web/src/helpers/loggers.ts
```diff
@@ -48,12 +45,6 @@ export const logError = (error: Error | ErrorResponse, { operation, extras }: Er
     // As soon as sentry is init, we add a scope to the error. Adding the scope here makes sure that
     // we don't have to manage the scopes on login/logout events
     Sentry.withScope(scope => {
-      // Set the organization data and the email of the user
```
3nvi: don't send any private information

PR #172, opened 2020-02-12
[Feature] Add x-ray to appsync

3nvi:
## Background

Appsync released an X-Ray integration a couple of days ago. This PR allows Appsync actions (queries/mutations) to be analyzed using X-ray

## Changes

- Enable X-Ray for Appsync

## Testing

Successfully deployed to dev account

### DIFF
--- a/deployments/core/appsync.yml
+++ b/deployments/core/appsync.yml
@@ -47,6 +47,7 @@ Resources:
       LogConfig:
         CloudWatchLogsRoleArn: !GetAtt AppsyncServiceRole.Arn
         FieldLogLevel: ALL
+      XrayEnabled: True
 
   GraphQLSchema:
     Type: AWS::AppSync::GraphQLSchema
--- a/deployments/core/appsync.yml
+++ b/deployments/core/appsync.yml
@@ -33,6 +33,13 @@ Parameters:
   UserPoolId:
     Type: String
     Description: Cognito user pool ID
+  TracingMode:
+    Type: String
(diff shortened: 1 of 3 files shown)


nhakmiller:
I think perhaps the labels will better serve you than putting [Feature] in the title.

deployments/core/appsync.yml
```diff
@@ -47,6 +47,7 @@ Resources:
       LogConfig:
         CloudWatchLogsRoleArn: !GetAtt AppsyncServiceRole.Arn
         FieldLogLevel: ALL
+      XrayEnabled: True
```
nhakmiller: In the other stacks this is configurable. We add a parameter:

```
  TracingMode:
    Type: String
    Description: Enable XRay tracing on Lambda and API Gateway
    Default: ''
```

With a condition:

```
TracingEnabled: !Not [!Equals ['', !Ref TracingMode]]
```

And then we set in the lambda with:

```
      Tracing: !If [TracingEnabled, !Ref TracingMode, !Ref 'AWS::NoValue']
```

Would this be applicable here?

austinbyers: Agreed - let's not enable XRay by default (comes with a higher cost) - use the option from the config file

3nvi: Got it. I opted into it by default since the free tier was really big. Forgot to account for the fact that other services might use this and - in turn - consume part of the free tier

PR #177, opened 2020-02-12
Oplog for genericapi

rleighton:
## Changes

Switch to  oplog instead of be-spoke zap log line in Handle() function to standardize the logging fields and provide performance metrics for dashboards.

## Testing
mage test:ci
Made some dashboards using stats in dev account.


internal/core/users_api/main/lambda.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/panther-labs/panther/pkg/lambdalogger"
 )
 
-var router = genericapi.NewRouter(models.Validator(), &api.API{})
+var router = genericapi.NewRouter("core", "usersApi", models.Validator(), &api.API{})
```
austinbyers: The `AWS_LAMBDA_FUNCTION_NAME` [environment variable](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime) already has the name of the Lambda function, I don't think you need to pass it through.

Why do you need to log the function name in every log field, doesn't the log group already tell you the function name?

pkg/genericapi/router.go
```diff
@@ -54,12 +59,22 @@ func NewRouter(validate *validator.Validate, routes interface{}) *Router {
...
+		operation.Stop()
+		operation.Log(err)
 		return nil, err
 	}
 
+	operation := oplog.NewManager(r.namespace, r.component).Start(req.route)
+	defer func() {
+		operation.Stop()
+		operation.Log(err, zap.Any("input", redactedInput(req.input)))
+	}()
```
austinbyers: Isn't this essentially the same as the Lambda duration? Why log this as a separate metric instead of just using Lambda duration?

rleighton:
> `genericapi` is deprecated - it's how we used to implement Lambda functions which serve as API handlers.
> 
> At least two of these (`users-api` and `organizations-api`) will probably be refactored before v1, with the rest migrated sometime later when we have time to unify all the APIs.
> 
> We can still make this change now, but why do it with just the API handlers and not all Lambda functions? I may need to understand better what we're trying to accomplish with the logging

I plan to work though all the lambdas to standardize the output of at least the top level logging. This makes the ops work much easier.

internal/core/users_api/main/lambda.go
rleighton: The namespace and component allow us to aggregate the logs (even over log groups).  These are not the best examples of using the capability for these simple lamndas  but consistency the important thing. For example we the log process lambda has 3 main functions (read from S3, process, write to s3), they all share namespace but they can be grouped or separated by the tags. Once we get this everywhere I want to spend time to revise the categorization.

pkg/genericapi/router.go
rleighton: Almost. We already have the lambda metric. This one normalizes out any cold start time for lambdas. It also make it easier to read operationally when all the info in 1 line. If you have an error, you can see it all at once and not have to search to group by requestId to get everything. In the new dashboards that show the last few errors you can just click on one to see the details and you get everything (assuming it is all propagate up) without a lot of digging around.

internal/compliance/snapshot_api/main/handler.go
```diff
@@ -36,7 +36,7 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
-	router = genericapi.NewRouter(validator, api.API{})
+	router = genericapi.NewRouter("compliance", "snapshotApi", validator, api.API{})
```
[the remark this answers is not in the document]
rleighton: I like `cloudsec`, will change

PR #182, opened 2020-02-13
Engine outputs to S3

### LINKED ISSUE #128: Rules engine should output matched events to S3
Make sure to link this Story to a related Epic!

## Introduction

Currently we store events that match rules in DDB. By storing these events in S3 we get significant benefits: 
*  Rules engine can handle spikes of matching events more efficient. In testing we have found that the current mechanism of storing events has scalability limitations and is possible to create an backlog in internal queue if some rule has big number of matches
* Lower storage cost for storing matched events. S3 is much cheaper than DDB and we potentially make further enhancements to limit further this cost e.g. transition older matches to S3 IA.

## Implementation

Rules engine outputing matched events to S3. The events need to have (additionally to the original fields)
1. `p_alert_id` field - the alert that this event is associated with 
2. `p_creation_time` field - the time the event was matched with the rule
3. `p_rule_id` field - the rule id of the rule that matched the event

kostaspap:
## Background

Closes #128 
Modifying Rules Engine to output to S3. The Rules Engine does now the following things: 
1. Buffer the events that match rules
1. When processing has finished or if we have buffered a bit number of messages, write matched events to S3. The events are written in the S3 bucket that contains the processed logs but under a different top-level folder: `rules/<log_type>/year=<year>/month=<month>/day=<day>/hour=<hour>/YYYYmmDDHHMMSS-uuid.gz`
  1. Once we have written events to S3, we send a notification to SNS topic with details
1. Rules engine can now run a `dedup` method. This method allows users to have a custom deduplication mechanism for their logs. 

## Changes

- Added SNS Message Attributes when sending a notification to SNS topic. The message attributes are "type" (rule output, log output) and "id" (rule ID or log type)
- Changed SNS->SQS subscription of Rules Engine to SNS topic to filter for LogData. We don't want (yet) the Rules Engine to process output from itself (rule chaining)
- Added DDB table that will be used for grouping of events to an alert. Grouping will be done based on a dedup string and time period. 

## Testing

- Running on my dev account
- unit tests


internal/log_analysis/rules_engine/src/alert_merger.py
```diff
@@ -0,0 +1,124 @@
...
+        try:
+            alert_info = _update_get_alert_info_conditional(match_time, num_matches, rule_id, dedup)
+            return alert_info
+        except _ddb_client.exceptions.ConditionalCheckFailedException as e:
+            # If conditional update failed on Condition, the event needs to be merged
+
+            return _update_get_alert_info(match_time, num_matches, rule_id, dedup)
+
+
+def _update_get_alert_info_conditional(match_time: datetime, num_matches: int, rule_id: str, dedup: str) -> AlertInfo:
```
rleighton: Should this be in the class above?

internal/log_analysis/rules_engine/src/alert_merger.py
```diff
@@ -0,0 +1,124 @@
...
+    )
+    alert_count = response['Attributes'][_ALERT_COUNT_ATTR_NAME]['N']
+    return AlertInfo(
+        alert_id=rule_id + '-' + alert_count,
+        alert_creation_time=match_time,
+        alert_update_time=match_time
+    )
+
+
+def _update_get_alert_info(match_time: datetime, num_matches: int, rule_id: str, dedup: str) -> AlertInfo:
```
rleighton: Should this be in the class above?

internal/log_analysis/rules_engine/src/alert_merger.py
```diff
@@ -0,0 +1,124 @@
...
+
+# TODO Once rules store alert merge period, retrieve it from there
+_ALERT_MERGE_PERIOD_SECONDS = 3600
+
+
+def _generate_key(rule_id: str, dedup: str) -> str:
+    return rule_id + '-' + dedup
+
+
+class Merger:
```
rleighton: I assume you have this class just for namespacing ... it currently has no state.

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -34,81 +35,75 @@ def __init__(self) -> None:
...
-                self.logger.error('Rule {} missing resourceTypes'.format(raw_rule['id']))
-                continue
             if raw_rule['id'] == COMMON_MODULE_RULE_ID:
                 # skip, should be already loaded above if present
                 continue
             # update lookup table from log type to rule
             import_count = import_count + 1
-            rule = Rule(raw_rule['id'], raw_rule['body'])
+            rule = Rule(
+                rule_id=raw_rule['id'],
```
rleighton: I strongly believe that raw rule keys need to be validated, you cannot trust what is in ddb. One bug in the UI/Api putting in a record with a missing key and you take out this process. That's what happened before and why I put the checks in. Now, if we had (somehow) constraints in the db to prevent this case, them we do not need the checks.  I'm not sure that is doable.

rleighton: Optionally, you could use .get() I'm not sure that buys anything . We want to skip the bad data and log.

internal/log_analysis/rules_engine/src/main.py
```diff
@@ -52,53 +87,28 @@ def lambda_handler(event: Dict[str, Any], unused_context) -> Union[None, Dict[st
...
             for data in data_stream:
-                try: # Bad json data can cause exceptions to be thrown. Best effort: log and continue
-                    for matched_rule in rules_engine.analyze(log_type, json.loads(data)):
-                        matched.append((matched_rule, data))
+                try:  # Bad json data can cause exceptions to be thrown. Best effort: log and continue
+                    json_data = json.loads(data)
                 except Exception as err:  # pylint: disable=broad-except
-                    logger.error("Error during matching: {}".format(err)) # do not log data!
+                    logger.error("Error during matching: {}".format(err))  # do not log data!
+
```
rleighton: This needs to 'continue' the loop, not execute the loop below with the bad data. In fact, I think it would die because json_data is not set.

internal/log_analysis/rules_engine/src/main.py
```diff
@@ -52,53 +87,28 @@ def lambda_handler(event: Dict[str, Any], unused_context) -> Union[None, Dict[st
...
+                for analysis_result in rules_engine.analyze( log_type,json_data ):
+                    output_buffer.add_event(analysis_result)
```
rleighton: I think we need to have a threshold on the number of elements and if greater, then flush/send sns. Imagine a big VPC flow file and a lot of matching rules ... we will OOM.

internal/log_analysis/rules_engine/src/output.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class MatchedEventsBuffer:
+    """Buffer containing the matched events"""
+
+    def __init__(self):
+        self.merger = Merger()
+        self.data: Dict[BufferDictKey, List[EventMatch]] = collections.defaultdict(list)
+
+    def add_event(self, match: EventMatch) -> None:
```
rleighton: Why -> None?

internal/log_analysis/rules_engine/src/alert_merger.py
kostaspap: Code changed - no class any more (was not needed and pylint was complaining)

internal/log_analysis/rules_engine/src/alert_merger.py
kostaspap: Code changed - no class any more (was not needed and pylint was complaining)

internal/log_analysis/rules_engine/src/alert_merger.py
kostaspap: True. Removed - pylint was complaining. No real state was in there.

internal/log_analysis/rules_engine/src/engine.py
kostaspap: Fair point. I am actually thinking to call .get(), then do the validation inside the Rule class, maybe it's more clean?

internal/log_analysis/rules_engine/src/output.py
kostaspap: It doesn't return anything. Linter complains (it wants return types everywhere)

internal/log_analysis/rules_engine/src/main.py
kostaspap: True. This is something I had realized as well. I was considering to do this as a subsequent fix... but you are right, I might as well do now. 
I'll go ahead and add a threshold on how many events we can store in memory.

internal/log_analysis/rules_engine/src/main.py
kostaspap: You are right, I should do the continue. Fixing now

internal/log_analysis/rules_engine/src/main.py
kostaspap: Fixed!

internal/log_analysis/rules_engine/src/engine.py
kostaspap: I've updated the code. Let me know what you think.

internal/log_analysis/rules_engine/src/output.py
```diff
@@ -0,0 +1,175 @@
...
+        value = self.data.get(key)
+        if value:
+            value.matches.append(match)
+            value.size_in_bytes += size
+        else:
+            value = BufferValue([match], size)
+            self.data[key] = value
+
+        self.total_bytes += size
+        # Check the total size of data in memory. If we exceed threshold, flush data from the biggest "offender"
```
rleighton: Why not flush until you reach a nice S3 file size (say at least 50% of _MAX_BYTES_IN_MEMORY) ? This method could create many small files, right?

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -40,8 +40,15 @@ import (
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/registry"
 )
 
-// s3ObjectKeyFormat represents the format of the S3 object key
-const s3ObjectKeyFormat = "%s%s-%s.gz"
+const (
+	// s3ObjectKeyFormat represents the format of the S3 object key
```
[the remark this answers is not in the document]
kostaspap: Good point. Will do.

internal/log_analysis/rules_engine/src/alert_merger.py
```diff
@@ -0,0 +1,124 @@
...
+
+def _update_get_alert_info_conditional(match_time: datetime, num_matches: int, rule_id: str, dedup: str) -> AlertInfo:
+    """Performs a conditional update to DDB to verify whether we need to create a new alert"""
+    response = _DDB_CLIENT.update_item(
+        TableName=_DDB_TABLE_NAME,
+        Key={_PARTITION_KEY_NAME: {
+            'S': _generate_key(rule_id, dedup)
+        }},
+        # Setting proper values for alertCreationTie, alertUpdateTime,
+        UpdateExpression='SET #1=:1, #2=:2, #3=:3\nADD #4 :4',
```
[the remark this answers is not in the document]
kostaspap: So, I want to increment the field by one. 
I can do SET #field = #field + :value (:value = 1), but ADD method comes handy here. 

Nope, this will commit if the creatiion time was older than our time threshold.

internal/log_analysis/rules_engine/src/alert_merger.py
```diff
@@ -0,0 +1,124 @@
...
+    try:
+        alert_info = _update_get_alert_info_conditional(match_time, num_matches, rule_id, dedup)
+        return alert_info
+    except _DDB_CLIENT.exceptions.ConditionalCheckFailedException:
+        # If conditional update failed on Condition, the event needs to be merged
+        return _update_get_alert_info(match_time, num_matches, rule_id, dedup)
+
+
+def _update_get_alert_info_conditional(match_time: datetime, num_matches: int, rule_id: str, dedup: str) -> AlertInfo:
+    """Performs a conditional update to DDB to verify whether we need to create a new alert"""
```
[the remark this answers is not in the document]
kostaspap: Good point, will do.

internal/log_analysis/rules_engine/src/main.py
```diff
@@ -15,94 +15,106 @@
...
+                    # Appends the events to queue of events that will be sent through SQS
+                    matched.append(data)
 
     if len(matched) > 0:
+        _LOGGER.info("sending %d matches", len(matched))
         send_to_sqs(matched)
+    else:
+        _LOGGER.info("no matches found")
+    output_buffer.flush()
```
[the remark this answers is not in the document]
kostaspap: Actually, the `matched` is used only for SQS delivery - the new mechanism uses the buffer to keep matches. 

The `matched` list will be removed eventually

internal/log_analysis/rules_engine/src/output.py
kostaspap: It would create some potentially smaller files .The current logic checks the threshold every time a new event is added, not periodically so we are guaranteed to not exceed the memory threshold.

internal/log_analysis/rules_engine/src/alert_merger.py
```diff
@@ -0,0 +1,124 @@
...
+_ALERT_MERGE_PERIOD_SECONDS = 3600
+
+
+def _generate_key(rule_id: str, dedup: str) -> str:
+    return rule_id + '-' + dedup
+
+
+def update_get_alert_info(match_time: datetime, num_matches: int, rule_id: str, dedup: str) -> AlertInfo:
+    """The method will return the alert information after evaluating if a new alert needs to be created
+        or if we can re-use an existing alert."""
```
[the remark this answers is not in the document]
kostaspap: Done

internal/log_analysis/rules_engine/src/output.py
```diff
@@ -0,0 +1,175 @@
...
+
+import boto3
+
+from . import EventMatch, OutputNotification, AlertInfo
+from .alert_merger import update_get_alert_info
+from .logging import get_logger
+
+_KEY_FORMAT = 'rules/{}/year={}/month={}/day={}/hour={}/rule_id={}/{}-{}.gz'
+# Maximum number of events in an S3 object
+_MAX_BYTES_IN_MEMORY = 100000000
```
rleighton: I think you should have this as a default but override from LAMBDA env, which would get passed down via CF. This would allow us to make this a tuneable number (one day) from the panther config. Right now I would just say allow env to override if present.

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -34,81 +35,76 @@ def __init__(self) -> None:
...
-            if 'body' not in raw_rule:
-                continue
-            if raw_rule['id'] == COMMON_MODULE_RULE_ID:
-                Rule(raw_rule['id'], raw_rule['body'])
+            if raw_rule.get('id') == COMMON_MODULE_RULE_ID:
+                try:
+                    Rule(rule_id=raw_rule.get('id'), rule_body=raw_rule.get('body'), rule_version=raw_rule.get('versionId'))
+                except Exception as err:  # pylint: disable=broad-except
+                    self.logger.error('Failed to import rule %s', err)
+                rules.remove(raw_rule)
```
rleighton: Is this safe to do (mutate) while iterating in the loop?

internal/log_analysis/rules_engine/src/main.py
```diff
@@ -15,94 +15,106 @@
...
+
+    # Dictionary containing mapping from log type to list of TextIOWrapper's
+    log_type_to_data: Dict[str, List[TextIOWrapper]] = collections.defaultdict(list)
     for record in event['Records']:
         record_body = json.loads(record['body'])
         bucket = record_body['s3Bucket']
         object_key = record_body['s3ObjectKey']
-        logger.info("loading object from S3, bucket [{}], key [{}]".format(bucket, object_key))
-        log_type_to_data[record_body['id']].append(load_contents(bucket, object_key))
+        _LOGGER.info("loading object from S3, bucket [%s], key [%s]", bucket, object_key)
```
rleighton: debug level? This will generate huge numbers of logs that we mostly do not care about unless doing dev/debug.

internal/log_analysis/rules_engine/src/main.py
```diff
@@ -15,94 +15,106 @@
...
 
     if len(matched) > 0:
+        _LOGGER.info("sending %d matches", len(matched))
         send_to_sqs(matched)
+    else:
+        _LOGGER.info("no matches found")
+    output_buffer.flush()
 
-    end = default_timer()
-    logger.info("Matched {} events in {} seconds".format(len(matched), end - start))
```
rleighton: I need this log line for dashboards.

internal/log_analysis/rules_engine/src/output.py
rleighton: It looks like it works to avoid OOM but it seems to have undesirable behavior under load. If we are triggering this, we likely have lots of data, which means we don't want lots of little files but this could create that, right? Why not just change so that when triggered, instead of picking 1, pick N where N reduces memory pressure by 50%?

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -29,74 +30,105 @@
...
+
+        if not rule_version:
+            self.rule_version = 'default'
+        else:
+            self.rule_version = rule_version
+
+        if not hasattr(self._module, 'rule'):
+            raise AssertionError("rule needs to have a method named 'rule'")
+
+        if hasattr(self._module, 'dedup'):
```
rleighton: Nice, and we can add methods in the future.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -29,74 +30,105 @@
...
+            self._has_dedup = True
+        else:
+            self._has_dedup = False
 
-        self._import_error = None
+    def run(self, event: Dict[str, Any]) -> RuleResult:
+        """Analyze a log line with this rule and return True, False, or an error."""
+
+        dedup_result: Optional[str] = None
```
rleighton: 'dedup_result' seems a little unclear to me. Would 'dedup_key' or 'dedup_string' be better?

internal/log_analysis/rules_engine/src/main.py
kostaspap: Good point. Fixed.

internal/log_analysis/rules_engine/src/main.py
kostaspap: I see. I can make sure to log this information. Question: 
1. I can log the number of events matched. Do you want this to be logged once and include the number of matches in the whole Lambda invocation or is it also fine to log multiple times per invocation (e.g. one log statement per S3 object processed)
1. What is the time duration you are interested in? The Lambda invocation duration or some other metric?

internal/log_analysis/rules_engine/src/rule.py
kostaspap: Right. Yeah, will change to `dedup_string`

internal/log_analysis/rules_engine/src/engine.py
kostaspap: In Python apparently this is not a problem - that's a recommended way to remove an item while iterating a list according to documentation I found online. 

I actually added a test that checks this branch and seen it pass.

kostaspap:
@rleighton  ( I can't seem to be able to reply to your comment earlier so putting it here)
> It looks like it works to avoid OOM but it seems to have undesirable behavior under load. If we are triggering this, we likely have lots of data, which means we don't want lots of little files but this could create that, right? Why not just change so that when triggered, instead of picking 1, pick N where N reduces memory pressure by 50%?

With the current mechanism we keep in memory a Dict that contains events grouped together in the same way that they would be written to S3. One such "group" contains potentially multiple events. 

The moment a new event is added that would cause the overall size of the Dict to exceed a given threshold, we write to S3 the "group" of events that occupy the most space in memory. So this avoids the problem of "small files" in S3.

internal/log_analysis/rules_engine/src/main.py
rleighton: I'd like it to be over the execution of the lambda hander (total time and total matches)

internal/log_analysis/rules_engine/src/output.py
rleighton: Saw reply. Resolving.

deployments/log_analysis/rules_engine.yml
```diff
@@ -120,6 +124,22 @@ Resources:
       MessageRetentionPeriod: 1209600 # Max duration - 14 days
       VisibilityTimeout: !Ref TimeoutSec
 
+
+  ##### Dynamo recent alerts table #####
```
austinbyers: Is this comment still accurate? There's still a separate `recent-alerts` table, isn't there?

internal/log_analysis/rules_engine/src/__init__.py
```diff
@@ -13,4 +13,36 @@
...
+    """Information about an alert"""
+    alert_id: str
+    alert_creation_time: datetime
+    alert_update_time: datetime
+
+
+# pylint: disable=invalid-name
+@dataclass
+class OutputNotification:
+    """Output notification. This class will be serialized to JSON, thus following camelCase rather than snake_case"""
```
austinbyers: Python docstrings should always have a short sentence first (can we have something more descriptive than just the class name again?)

```python
class OutputNotification:
    """A quick summary of what this is for.
    
    This class will be serialized to JSON...
    """

File: internal/log_analysis/rules_engine/src/alert_merger.py [hunk 22]
```diff
@@ -0,0 +1,140 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import os
+from datetime import datetime
```
austinbyers: Nit: alphabetical imports (`from datetime` would happen before `import os`)

internal/log_analysis/rules_engine/src/alert_merger.py
```diff
@@ -0,0 +1,140 @@
...
+_ALERT_MERGE_PERIOD_SECONDS = 3600
+
+
+def _generate_key(rule_id: str, dedup: str) -> str:
+    return rule_id + '-' + dedup
+
+
+def update_get_alert_info(match_time: datetime, num_matches: int, rule_id: str, dedup: str) -> AlertInfo:
+    """The method will update the alertCreationTime, eventCount of an alert. If a new alert will have to be created,
+    it will also create a new alertId with the appropriate alertCreationTime. """
```
austinbyers: Another example of the docstring break:

```python
def update_get_alert_info(...):
    """Update the alertCreationTime and eventCount of an alert.

    If a new alert will have to be created, it will also create a new alertId ...
    """
```

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -17,98 +17,94 @@
 import collections
 from datetime import datetime, timedelta
 from timeit import default_timer
-from typing import Any, Dict, List
+from typing import Any, List, Dict
```
austinbyers: The order you had before was right (`Dict` < `List` alphabetically)

internal/log_analysis/rules_engine/src/output.py
```diff
@@ -0,0 +1,177 @@
...
+import collections
+import gzip
+import json
+import os
+import sys
+import uuid
+from dataclasses import dataclass, asdict
+from datetime import datetime
+from io import BytesIO
+from typing import Dict, List, Optional
```
austinbyers: Nit: alphabetical imports by group (not sure why `pylint` doesn't catch it)

internal/log_analysis/rules_engine/tests/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+from unittest import mock
+
+S3_MOCK = mock.MagicMock()
+DDB_MOCK = mock.MagicMock()
+SNS_MOCK = mock.MagicMock()
+
+
+def mock_to_return(value: str) -> mock.MagicMock:
+    if value == 'sns':
+        return SNS_MOCK
```
austinbyers: Usually it's better to create the mock every time you need it instead of sharing it across test methods, since mocks remember their state

internal/log_analysis/rules_engine/src/alert_merger.py
```diff
@@ -0,0 +1,140 @@
...
+_ALERT_COUNT_ATTR_NAME = 'alertCount'
+_ALERT_EVENT_COUNT = 'eventCount'
+
+# TODO Once rules store alert merge period, retrieve it from there
+# Currently grouping in 1hr periods
+_ALERT_MERGE_PERIOD_SECONDS = 3600
+
+
+def _generate_key(rule_id: str, dedup: str) -> str:
+    return rule_id + '-' + dedup
```
nhakmiller: I suspect this is not going to guarantee correctness, as there is the chance for overlaps. Imagine two alerts, one for rule_id `Check-CloudTrail` with dedup string `Correct`, and one for rule_id `Check`, with dedup string `CloudTrail-Correct`. These would overlap, causing whichever alert occurred second to be suppressed incorrectly.

In your `_allowed_char` function, you allow the characters ` `, `-`, and `_` in rule_ids (in addition to alpha numeric characters). I think if the separator you use here is not defined as an "allowed character" then this should be safe. Maybe using `?` or `/` or something as the separator.

internal/log_analysis/rules_engine/src/output.py
```diff
@@ -0,0 +1,177 @@
...
+
+    # MessageAttributes are required so that subscribers to SNS topic
+    # can filter events in the subscription
+    _SNS_CLIENT.publish(
+        TopicArn=_SNS_TOPIC_ARN,
+        Message=json.dumps(asdict(notification)),
+        MessageAttributes={
+            "type": {
+                "DataType": "String",
+                'StringValue': notification.type
```
nhakmiller: Is there significance to alternating between `"` and `'` for string literals?

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -29,74 +30,105 @@
...
-        path = os.path.join(_RULE_FOLDER, safe_id + '.py')
-        return path
 
-    def allowed_char(self, char: str) -> bool:
-        """Return true if the character is part of a valid rule ID."""
-        return char.isalnum() or char in {' ', '-', '.'}
+def _run_command(function: Callable, event: Dict[str, Any], expected_type: Any) -> Any:
+    result = function(event)
+    if not isinstance(result, expected_type):
+        raise Exception('rule returned {}, expected {}'.format(type(result).__name__, expected_type.__name__))
```
nhakmiller: Perhaps logging the name of the failing function would be useful as well, as this function is used for both the `dedup` and `rule` functions and potentially more functions in the future if we expand this functionality.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -29,74 +30,105 @@
...
-            return self._import_error
 
-        try:
-            # Python source should have a method called "rule"
-            matched = self._module.rule(event)
-        except Exception as err:  # pylint: disable=broad-except
-            return err
+def _rule_id_to_path(rule_id: str) -> str:
+    """Method returns the file path where the rule will be stored"""
+    safe_id = ''.join(x if _allowed_char(x) else '_' for x in rule_id)
```
nhakmiller: Also unprotected from collisions, but I think this is fair as long as we have documented what are allowed characters in rule/policy IDs (we don't currently, but I'll add an issue for myself to update the writing rules/policy docs page with this information and other details).

internal/log_analysis/rules_engine/src/sqs.py
```diff
@@ -26,35 +26,33 @@
 # Max size of an SQS batch request
 _MAX_MESSAGE_SIZE = 256 * 1000
 
-sqs_resource = boto3.resource('sqs')
-queue = sqs_resource.get_queue_by_name(QueueName=os.environ['ALERTS_QUEUE'])
+_SQS_RESOURCE = boto3.resource('sqs')
+_QUEUE_NAME = _SQS_RESOURCE.get_queue_by_name(QueueName=os.environ['ALERTS_QUEUE'])  # pylint: disable=no-member
```
nhakmiller: This seems like a misleading variable name, as this isn't actually the Queue Name but a reference to the Queue directly, that was retrieved based on the queues name. I'd suggest `_QUEUE` or `_QUEUE_HANDLE` or something.

internal/log_analysis/rules_engine/tests/test_engine.py
```diff
@@ -0,0 +1,103 @@
...
+        analysis_api = mock.MagicMock()
+        analysis_api.get_enabled_rules.return_value = [
+            {
+                'id': 'aws_globals',
+                'resourceTypes': ['log'],
+                'body': 'def is_true():\n\treturn True'
+            }, {
+                'id': 'rule_id',
+                'resourceTypes': ['log'],
+                'body': 'def rule(event):\n\treturn True'
```
nhakmiller: I think this test would benefit if `rule_id` actually imported `aws_globals`, something like:

```
                'id': 'rule_id',
                'resourceTypes': ['log'],
                'body': 'from aws_globals import is_true\ndef rule(event):\n\tis_true()'
```

or even just doing the import without using what's imported.

deployments/log_analysis/rules_engine.yml
kostaspap: Good catch. Yup, will rename.

internal/log_analysis/rules_engine/src/__init__.py
kostaspap: Done

internal/log_analysis/rules_engine/src/alert_merger.py
kostaspap: Done

internal/log_analysis/rules_engine/src/alert_merger.py
kostaspap: Done

internal/log_analysis/rules_engine/src/engine.py
kostaspap: Done. It's pretty hard to check for this though, pretty disappointing that pylint doesn't check for that.

internal/log_analysis/rules_engine/src/output.py
kostaspap: Updated.

internal/log_analysis/rules_engine/tests/__init__.py
kostaspap: I know, I absolutely hate that I have to "reset" the mocks. I couldn't find an elegant way to avoid this though, I am open to suggestions :)

internal/log_analysis/log_processor/common/common.go
```diff
@@ -62,18 +62,18 @@ type S3DataStreamHints struct {
...
 	// Bytes is the uncompressed size in bytes of the S3 object
-	Bytes *int `json:"bytes"`
+	Bytes *int `json:"bytes,omitempty"`
 	// Type is the type of data available in the S3 object (LogData,RuleOutput)
-	Type *string `json:"type"`
+	Type *string `json:"type,omitempty"`
 	// ID is an identified for the data in the S3 object. In case of LogData this will be
 	// the Log Type, in case of RuleOutput data this will be the RuleID
-	ID *string `json:"id"`
+	ID *string `json:"id,omitempty"`
```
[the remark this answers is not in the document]
kostaspap: Actually not - just followed general practice. Can remove.

internal/log_analysis/rules_engine/src/output.py
kostaspap: Not really. Just mistake from my part. Fixed!

internal/log_analysis/rules_engine/src/rule.py
kostaspap: Good catch. Updated.

internal/log_analysis/rules_engine/tests/test_engine.py
kostaspap: Great idea. Added.

internal/log_analysis/rules_engine/src/sqs.py
kostaspap: Good catch. Renaming.

internal/log_analysis/rules_engine/src/alert_merger.py
kostaspap: Great catch. Yes, will update.

internal/log_analysis/rules_engine/src/rule.py
kostaspap: Good point. I actually think the whitelisting of characters should be done on the analysis-api level (and also add documentation for it). It's fair to restrict RuleID/PolicyID to only a handfull of characters. Makes our life easier without compromising user experience (users are going to use the separators we allow them to use).

internal/log_analysis/rules_engine/src/main.py
kostaspap: Added back.

PR #186, opened 2020-02-14
Cloud Security CloudTrail Processing

### LINKED ISSUE #175: Allow Cloud Security to Use Log Analysis Data
## Introduction

Both Cloud Security and Log Analysis can make use of CloudTrail data, but currently it needs to be setup separately for each component. There is no reason that Cloud Security could not hook into the existing CloudTrail data coming in for Log Analysis and save the user some setup steps.

## Implementation

- Subscribe the `panther-aws-events` queue to the same SNS topics as the `panther-input-data-notifications` queue
- Modify the `panther-aws-event-processor` to handle SNS events that are either raw CloudTrail or CloudTrail wrapped in SNS (current behavior), or SNS events that are S3 object created notifications.
  - In the second case, we will need to imitate or re-use the S3 downloading capabilities of the `panther-log-processor`

nhakmiller:
## Background

Closes #175, full context is there.

This change is to allow the `aws-event-processor` to process the same CloudTrail data that the `rules-engine` is processing in order to provide the real time Cloud Security functionality.

This is supplementing the existing implementation, which requires CloudTrail logs to be sent via SNS to the `panther-aws-events` queue.

## Changes

- Allow the `panther-aws-events` queue to be subscribed to the same topics as the `panther-rules-engine` queue
- Make the `aws-event-processor` differentiate between an SNS message that is just wrapping CloudTrail, and an SNS message that is an S3Notification from the `panther-log-processor`
- Allow the `aws-event-processor` to download CloudTrail logs from S3

## Testing

- Unit testing
- Deployed in dev account and tested


deployments/compliance/aws_event_processor.yml
```diff
@@ -167,3 +174,12 @@ Resources:
             - Effect: Allow
               Action: lambda:InvokeFunction
               Resource: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-snapshot-api
+        - Id: AssumePantherLogProcessingRole
+          Version: 2012-10-17
+          Statement:
+            - Effect: Allow
+              Action: sts:AssumeRole
+              Resource: !Sub arn:${AWS::Partition}:iam::*:role/PantherLogProcessingRole
```
kostaspap: You shouldn't need to do this, since you should be just reading from `panther-processed-data-...` bucket

deployments/auxiliary/cloudformation/panther-log-processing-infra.yml
```diff
@@ -148,15 +148,25 @@ Resources:
...
 
+  # This subscription subscribes the Panther Cloud Security SQS queue to the topic created above
+  CloudSecuritySubscription:
+    Type: AWS::SNS::Subscription
+    Properties:
+      Endpoint: !Sub arn:${AWS::Partition}:sqs:us-west-2:${MasterAccountId}:panther-aws-events
+      Protocol: sqs
+      RawMessageDelivery: false
+      TopicArn: !Ref LogProcessingNotificationsTopic
+
```
kostaspap: No need to subscribe to this topic. You should subscribe to Panther internal topic, the same the rules-engine is [subscribing](https://github.com/panther-labs/panther/blob/engine/deployments/log_analysis/rules_engine.yml#L79-L90) to.

nhakmiller:
> You don't need all this. You should actually just be reading the output of the LogProcessor. You should subscribe to [this](https://github.com/panther-labs/panther/blob/engine/deployments/backend.yml#L289-L293) SNS topic, not the one from the customer. The data pipeline should be the following:
> `Customer S3 bucket -> Log processor -> internal S3 bucket -> aws-event-processor . `
> Rules engine is doing something similar:
> `Customer S3 bucket -> Log processor -> internal S3 bucket -> rules-engine `
> the only difference being that aws-event-processor only cares for CloudTrail data (rules engine cares for all data).

Oh, I thought the design I did here was the one you suggested me to use. I guess I should have double checked with you, but I had some time yesterday and really wanted to get this done. Measure twice and cut once though I guess. I will take a look at re-working this then.

nhakmiller:
@kostaspap I was actually able to maintain the majority of the code, since I'm just reading from a different S3 bucket (and with a different input) than I was expecting. Take a look now and see if this is more in line with what you were thinking.

deployments/compliance/aws_event_processor.yml
```diff
@@ -89,6 +104,13 @@ Resources:
             Condition:
               ArnLike:
                 aws:SourceArn: !Sub arn:${AWS::Partition}:sns:*:*:*-PantherEventsTopic-*
+          - Effect: Allow
+            Principal: '*'
+            Action: sqs:SendMessage
+            Resource: '*'
+            Condition:
+              ArnLike:
+                aws:SourceArn: !Sub arn:${AWS::Partition}:sns:${AWS::Region}:${AWS::AccountId}:panther-processed-data-notifications
```
kostaspap: Since both SNS topic and SQS queue are in the same account, you don't need explicit policy to allow the subscription.

deployments/auxiliary/cloudformation/panther-log-processing-notifications.yml
```diff
@@ -94,8 +94,8 @@ Resources:
       Topics:
         - !Ref Topic
 
-  # SNS Topic subscription to Panther
-  Subscription:
+  # SNS Topic subscription to Panther for Log Analysis
+  LogAnalysisSubscription:
```
kostaspap: nit: It is implicit that this is for log-analysis due to the name of the yml file.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -53,45 +59,76 @@ func Handle(batch *events.SQSEvent) error {
 		return err
 	}
 
+	// Using gjson to get only the fields we need is > 10x faster than running json.Unmarshal multiple times
+	//
+	// Since we don't want one bad notification to lose us the rest, we log failures and continue.
```
kostaspap: Why would we lose any data? If the Lambda fails, the messages will not be removed from the SQS queue and will be retried.

deployments/compliance/aws_event_processor.yml
```diff
@@ -76,6 +82,15 @@ Resources:
         maxReceiveCount: 10
       VisibilityTimeout: !Ref TimeoutSec
 
+  SnsSubscription:
```
kostaspap: Once [this](https://github.com/panther-labs/panther/pull/182) PR is merged, you will be able to use SNS subscription filters to receive notifications only for CloudTrail events. 
I'd recommend adding a TODO and maybe a Github issue for that. It will be just 3 lines change.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -53,45 +59,76 @@ func Handle(batch *events.SQSEvent) error {
...
+			zap.L().Debug("processing raw CloudTrail")
+			err := processCloudTrail(detail, changes)
+			if err != nil {
+				zap.L().Error("error processing raw CloudTrail", zap.Error(errors.WithStack(err)))
+			}
+			continue
+		}
+
+		// This case is checking for a notification from the log processor that there is newly processed CloudTrail logs
+		bucket := gjson.Get(record.Body, "s3Bucket")
```
kostaspap: You should filter for CloudTrail events. 
`type` will be `LogData`, `id` will be `AWS.CloudTrail`. Eventually this filtering can be moved to the SNS subscription (see comment above).

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -100,6 +137,43 @@ func handleCloudtrail(body string, changes map[string]*resourceChange) {
 			changes[key] = summary // the newest event for this resource we've seen so far
 		}
 	}
+	return nil
+}
+
+func processS3Download(object *sources.S3ObjectInfo, changes map[string]*resourceChange) error {
+	s3Svc := s3.New(sess)
```
kostaspap: No need to create a new S3 Client on every event. You should just re-use the same client.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -100,6 +137,43 @@ func handleCloudtrail(body string, changes map[string]*resourceChange) {
...
+		if err != nil {
+			if err == io.EOF { // we are done
+				err = nil
+			}
+			break
+		}
+		// Since we don't wont to lose an entire batch of logs to one bad message, we just log failures and continue
+		err = processCloudTrail(gjson.Parse(line), changes)
+		if err != nil {
+			zap.L().Error("error processing CloudTrail from S3", zap.Error(err))
```
kostaspap: You should log the full stacktrace using `errors.Wrap` or equivalent.

deployments/compliance/aws_event_processor.yml
kostaspap: Replace value with `!Ref SnsTopicArn`

deployments/auxiliary/cloudformation/panther-log-processing-notifications.yml
nhakmiller: Sure this was just leftover from the previous setup, I didn't think it was hurting anything but I can remove it.

deployments/compliance/aws_event_processor.yml
nhakmiller: Updated.

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: Two of the three data processing cases are fairly deterministic as they make zero network calls, so re-trying will most likely mean they just run into the same failure again. So it will just keep failing until it is moved to the dead letter queue, and then any events grouped with the bad event will be lost.

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: Added these checks.

deployments/compliance/aws_event_processor.yml
nhakmiller: Added.

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: Added.

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: Updated.

internal/compliance/aws_event_processor/processor/handler.go
austinbyers: I see, because we do things in batches, you don't want to drop all 10 events if just one of them was malformatted. I agree that's a good idea

In that case, should we deliver the one bad event directly to the dead letter queue?

nhakmiller: I'd be open to this, would it involve a direct call to `sqs.SendMessage` and perhaps passing in the dead letter queue name as an environment variable?

kostaspap: My point is, we should make sure the code is resilient to failures. With the current logic, if the S3 partition we are hitting on line 90 has encountered some issue for 1', the message will be dropped. And S3 fails has issues like that pretty frequently. There are few ways to deal with this: 
1. Use application retries. This comes with cost (processing & $$ cost) and has some time limitations. 
1. Use messaging system retry mechanism. 

I don't think sending messages directly to the DLQ makes sense - now you have to deal with the scenario where sending the message to the DLQ fails... and there is no other way to deal with that than retry. 

> Two of the three data processing cases are fairly deterministic as they make zero network calls, so re-trying will most likely mean they just run into the same failure again. So it will just keep failing until it is moved to the dead letter queue, and then any events grouped with the bad event will be lost.

I don't think we should worry for such a scenario. In a high-TPS system the chance that few bad messages will cause a big chunk of good messages to fail is pretty unlikely. It is unlikely that all the messages will be grouped always in the same way together. It's an issue in theory but the chances of it happening are very low. 
In a low TPS system messages will be in small groups or not grouped at all in which case there is little or chance for `collateral` damage. 

Messages ending up in the DLQ doesn't mean they are lost. By having a DLQ we can very easily script forwarding messages from the DLQ back to the main queue.

PR #194, opened 2020-02-14
Connect Cognito custom message trigger to Lambda function

### LINKED ISSUE #190: Forgot password flow is incomplete
### Describe the bug

When I click forget password, I'm prompted to add my email and then I'm told that I've received an email with instructions on how to reset my password.

What I end up receiving is an email saying `Your temporary token is XXXXX`. I don't know how to use it or what to do with it. 

### Steps to reproduce

1. Go to login page
2. Click on forget password
3. Enter your email
4. Check your inbox

### Expected behavior

The email should have instructions on how to reset it. 

### Environment

- Panther version or commit: v0.2.0
- OS: Mac OOS
- Browser: Chrome

### Additional context
Up until recently, it included a single URL where it would redirect users to `panther-domain.com/password-reset` with the params `email` & `token` pre-filled. 

The `email` would be the email of the related user and the `token` would be the temporary token that I'm currently receiving through the email.

An example of the final URL is the following:

```
https://whatever.runpanther.com/password-reset?token=119495&email=whatever@whatever.io
```
(params should be URL encoded)

### Screenshots

The final email should look like this:
Screen Shot 2020-02-14 at 7 51 55 PM

austinbyers:
## Background
We have a custom format for password reset emails, but Cognito wasn't using it. Turns out we had the custom message Lambda function, it was just never connected to Cognito (we used to build the Congito user pool with an API call instead of CloudFormation)

Closes #190 

## Changes

- Connect `panther-custom-message-trigger` Lambda function to Cognito
- Remove some dead code in `users-api` related to how we used to set up Cognito pools

## Testing

- Deployed to dev account, triggered password reset flow:

Screen Shot 2020-02-14 at 3 25 33 PM

Clicking the link in the email worked and I was able to reset the password and login with the new password

## Future
We are redesigning the `users-api` before v1; the custom message trigger can probably go there as well. It's a bit strange having an entire Lambda function for just this 1 use case (forgot password)

### DIFF
--- a/internal/core/custom_message/email/clients.go
+++ b/internal/core/custom_message/email/clients.go
@@ -34,7 +34,7 @@ var (
 		Product: hermes.Product{
 			// Appears in header & footer of e-mails
 			Name:      "Panther",
-			Link:      "runpanther.io",
+			Link:      "https://runpanther.io",
 			Copyright: "Copyright © " + strconv.Itoa(time.Now().Year()) + " Panther Labs Inc. All rights reserved.",
 			Logo:      pantherEmailLogo,
 		},
--- a/internal/core/custom_message/email/clients.go
+++ b/internal/core/custom_message/email/clients.go
@@ -19,7 +19,6 @@ package email
  */
 
 import (
-	"os"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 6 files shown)


internal/core/custom_message/email/clients.go
```diff
@@ -29,14 +28,13 @@ import (
...
 	pantherEmailLogo = "https://d14d54mfia7r7w.cloudfront.net/panther-email-logo-white.png"
-	appDomainURL     = os.Getenv("APP_DOMAIN_URL")
 	// PantherEmailTemplate is used as a boilerplate for Panther themed email
 	PantherEmailTemplate = hermes.Hermes{
 		Theme: new(hermes.Flat),
 		Product: hermes.Product{
 			// Appears in header & footer of e-mails
 			Name:      "Panther",
-			Link:      appDomainURL,
+			Link:      "runpanther.io",
```
kostaspap: Since this is meant to be a link, I think it should be a full URL e.g. `https://runpanther.io`

austinbyers:
> We should perhaps remove Hermes in the future since it's very opinionated

Sure, I imagine at some point we will have a more sophisticated email template with the Panther logo and a backdrop and so on. The initial invite email template (hardcoded in Cognito CloudFormation) is actually just copied from the html output of this hermes template

PR #196, opened 2020-02-16
Updated initial analysis sets to always point to latest

nhakmiller:
## Background

The initial policies/rules set was set to a specific tagged release of the `panther-analysis`, this switches it to latest so it doesn't have to be updated all the time.

## Changes

- Pointed initial analysis set at latest

## Testing

- Copied the URL into my browser and the proper pack was downloaded

### DIFF
--- a/deployments/panther_config.yml
+++ b/deployments/panther_config.yml
@@ -79,4 +79,4 @@ PipLayer:
 # If the analysis-api is non-empty, this setting is ignored and you can instead
 # use the BulkUpload functionality from the web app to upload new or modified rule sets.
 InitialAnalysisSets:
-  - https://github.com/panther-labs/panther-analysis/releases/download/v1.0.1/panther-analysis-all.zip
+  - https://github.com/panther-labs/panther-analysis/releases/latest/download/panther-analysis-all.zip


jacknagz:
@nhakmiller part of me actually likes the clarity of version pinning. Don't you think there's value in users knowing which version of the packs they have installed?

rleighton:
> @nhakmiller part of me actually likes the clarity of version pinning. Don't you think there's value in users knowing which version of the packs they have installed?

Is there a way to choose? For production I want it pinned, for dev I want latest :)

kostaspap:
> @nhakmiller part of me actually likes the clarity of version pinning. Don't you think there's value in users knowing which version of the packs they have installed?

On the same topic, is it visible somewhere what `analysis-version` the customers have?  I agree that it would be useful to have a version pinned, even though if we ever have to debug a customer usecase, we have to keep in mind that the rules might be modified by the users.

nhakmiller:
@jacknagz @rleighton : Users can still choose to pin to a specific version of policies/rules by overwriting this parameter, I'm just setting this as the new default. It also only changes when doing a fresh deployment, this parameter is ignored during normal updates and day to day operations.

@kostaspap We currently do not display the version of policies/rules anywhere. Perhaps we can store it in the same location we store the version of Panther that's deployed? I'm not sure where that is though.

My thought process for this change was the same as @austinbyers , that since we're already controlling versioning in the `panther-analysis` repo we may as well point to latest here. I see now it's not as cut and dry as that and there are some differing opinions, I'll hold off on merging this PR until we have some more internal discussion on it.

nhakmiller:
Chatted with Austin, it's an easily reversible and easily overwriteable configuration so we will deploy it. If we get negative feedback we can revert.

PR #199, opened 2020-02-16
Wrap all lambdas with execution logging

rleighton:
## Background
This PR wraps all lambda with execution logging. The benefits will be described below. This touches many files in compliance (cloudsec), so those familiar with this code please give this a good hard look to make sure I did not break some desired behavior (it passed all tests). Log analysis has already been changed to implement this pattern.

The basic pattern is the all lambda handlers begin with a code fragment similar to:
```
func reporterHandler(ctx context.Context, event events.DynamoDBEvent) (err error) {
	lc, _ := lambdalogger.ConfigureGlobal(ctx, nil)
	operation := oplog.NewManager("cloudsec", "alert_forwarder").Start(lc.InvokedFunctionArn).WithMemUsed(lambdacontext.MemoryLimitInMB)
	defer func() {
		operation.Stop().Log(err, zap.Int("numEvents", len(event.Records)))
	}()
```
Notice:
* WithMemUsed(lambdacontext.MemoryLimitInMB), this will log the percent memory used by the Go process as well a other memory and garbage collection stats
* The defer closes over the error returned by the lambda, which will give the error, a stack trace and all the execution stats reported in a single log message

Now that we log errors at the top level caller, this should encourage LESS logging down the stack, particularly for errors. The pattern encouraged is to log as little as possible and then errors.Wrap() the errors down the stack. This makes the code simpler and operationally eases tracing errors because errors will always have useful context messages and a stack trace.

### Tagging and Audit
This PR adds to the lambda logger an "application" tag which maps to "panther". This allows us to find all "panther" application logs over all log groups.

The oplog package tags all logs with a 'namespace' and 'component' along with the name of the operation being wrapped. This PR creates a tagging hierarchy of:
* namespace: cloudsec
* namespace: log_analysis
* namespace: admin 

Each namespace has component tags to group related sub-system activity.  NOTE: "admin" is really for all the apis that seem to be used by the UI to manage the application and resources. If you can think of a better label, that would be good.

By grouping logs by these tags we can do aggregate searches (e.g., show me all errors in log processing). 

My current plan is to use this tagging to allow us to extract the panther application logs in a standard format into S3 where we will put a Glue table over the data (which is now a mostly standard format) allowing searches to facilitate easy audit.

### Grouping
Previous to this change there were many log lines for errors and other status and when doing a log dive one would have to search and group by "requestId". With this change, this is far less likely to be necessary. A log line with an error now contains most of the information needed to root cause.

### Memory and GC Stats
The logs will also have memory and garbage collection stats which can be _very_ useful for troubleshooting productions issues.

## Changes
* Wrap all lambda execution with oplog (that were not already wrapped)
* Add "application" -> "panther" to lambda logger
* Add WithMemUsed() option to oplog

## Testing
mage test:ci
mage test:integration
ran in dev account


internal/compliance/aws_event_processor/processor/handler_test.go
```diff
@@ -140,26 +142,8 @@ func TestHandleInvalid(t *testing.T) {
...
-			},
-		},
-		{
-			Entry:   zapcore.Entry{Level: zapcore.WarnLevel, Message: "dropping unknown notification type"},
-			Context: []zapcore.Field{zap.String("body", batch.Records[0].Body)},
-		},
-	}
-	assert.Equal(t, expected, logs.AllUntimed())
+	require.Nil(t, Handle(testContext, batch))
+	assert.Equal(t, 1, len(logs.FilterMessage("dropping unknown notification type").AllUntimed()))
```
nhakmiller: This seems like a much weaker unit test, and while these tests are annoying to update every time something changes I have found them to be quite helpful in catching unexpected behavior. Is it no longer possible to do tests like these with this new setup?

internal/compliance/remediation_processor/main/lambda.go
```diff
@@ -37,15 +41,21 @@ func main() {
...
-			return err
+		if err = jsoniter.UnmarshalFromString(record.Body, &input); err != nil {
+			operation.LogError(errors.Wrap(err, "Failed to unmarshal item"))
+			continue
 		}
-		if err := invoker.Remediate(&input); err != nil {
+		if err = invoker.Remediate(&input); err != nil {
+			err = errors.Wrap(err, "encountered issue while processing event")
 			return err
 		}
```
nhakmiller: Why continue on an un-marshalling failure and return on a remediation failure? I would think they should both return an error or both log and continue.

internal/compliance/resource_processor/processor/handler.go
```diff
@@ -63,7 +73,7 @@ func Handle(batch *events.SQSEvent) error {
 
 		if policy != nil {
 			// Policy updated - analyze applicable resources now
-			if err := results.analyzeUpdatedPolicy(policy); err != nil {
+			if err = results.analyzeUpdatedPolicy(policy); err != nil {
 				return err
```
nhakmiller: Should this be `errors.WithStack(err)`?

internal/compliance/snapshot_poller/pollers/lambda.go
```diff
@@ -58,43 +61,37 @@ func batchResources(resources []*api.AddResourceEntry) (batches [][]*api.AddReso
...
 
 		for _, entry := range scanRequest.Entries {
-			_, logger := lambdalogger.ConfigureGlobal(ctx, map[string]interface{}{
-				"sqsEntry": entry,
-			})
-			logger.Info(
-				"starting poller",
-				zap.Int("messageNumber", indx),
-			)
+			operation.LogSuccess(zap.Any("sqsEntry", entry),
```
nhakmiller: What is the `LogSuccess` for?

internal/core/custom_message/api/event_router.go
```diff
@@ -19,14 +19,29 @@ package custommessage
...
+	"github.com/panther-labs/panther/pkg/oplog"
 )
 
 // HandleEvent routes Custom Message event based on the triggerSource
-func HandleEvent(event *events.CognitoEventUserPoolsCustomMessage) (*events.CognitoEventUserPoolsCustomMessage, error) {
+func HandleEvent(ctx context.Context, event *events.CognitoEventUserPoolsCustomMessage) (
+	returnedEvent *events.CognitoEventUserPoolsCustomMessage, err error) {
+
+	lc, _ := lambdalogger.ConfigureGlobal(ctx, nil)
+	operation := oplog.NewManager("admin", "customer_message").Start(lc.InvokedFunctionArn).WithMemUsed(lambdacontext.MemoryLimitInMB)
```
nhakmiller: I think this is probably a typo, should be `custom_message` for the namespace.

internal/compliance/alert_forwarder/main/lambda.go
```diff
@@ -40,26 +43,31 @@ func main() {
...
-			logger.Warn("Skipping records")
+			operation.LogWarn(errors.New("Skipping records"))
 			continue
 		}
 		var alert models.Alert
-		if err := jsoniter.Unmarshal(record.Change.NewImage[alertConfigKey].Binary(), &alert); err != nil {
-			logger.Warn("Failed to unmarshall ddb stream item", zap.Error(err))
-			return err
+		if err = jsoniter.Unmarshal(record.Change.NewImage[alertConfigKey].Binary(), &alert); err != nil {
+			operation.LogError(errors.Wrap(err, "Failed to unmarshal item"))
```
austinbyers: Does this mean all instances of `zap.L().Error` should be replaced with `operation.LogError`? If so, you've only updated the main handlers and not all of the error logging. If not, when do we use each one?

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -146,25 +153,23 @@ func submitChanges(changes map[string]*resourceChange) error {
 
 	// Send deletes to resources-api
 	if len(deleteRequest.Resources) > 0 {
-		zap.L().Info("deleting resources", zap.Any("deleteRequest", &deleteRequest))
+		zap.L().Debug("deleting resources", zap.Any("deleteRequest", &deleteRequest))
```
austinbyers: DEBUG logging is off by default, what sorts of operations do you see as INFO worthy? I think this should be info - we're about to communicate with another service, which is often when things fail.

internal/compliance/remediation_processor/main/lambda.go
austinbyers: They should both return to keep the same behavior we had before

internal/compliance/resource_processor/processor/handler.go
```diff
@@ -73,11 +83,11 @@ func Handle(batch *events.SQSEvent) error {
...
 
 	// Analyze updated resources with applicable policies
-	if err := results.analyze(resources, nil); err != nil {
+	if err = results.analyze(resources, nil); err != nil {
 		return err
 	}
-
-	return results.deliver()
+	err = results.deliver()
+	return err
```
austinbyers: No need to break apart, just `return results.deliver()`

internal/compliance/snapshot_scheduler/main/handler.go
```diff
@@ -23,14 +23,21 @@ import (
...
-func lambdaHandler(ctx context.Context, request events.CloudWatchEvent) error {
-	lambdalogger.ConfigureGlobal(ctx, nil)
-	return scheduler.PollAndIssueNewScans()
+func lambdaHandler(ctx context.Context, request events.CloudWatchEvent) (err error) {
+	lc, _ := lambdalogger.ConfigureGlobal(ctx, nil)
+	operation := oplog.NewManager("cloudsec", "snapshot").Start(lc.InvokedFunctionArn).WithMemUsed(lambdacontext.MemoryLimitInMB)
+	defer func() {
+		operation.Stop().Log(err)
+	}()
+	err = scheduler.PollAndIssueNewScans()
```
austinbyers: `return scheduler.PollAndIssueNewScans()`

pkg/gatewayapi/router.go
```diff
@@ -52,35 +55,36 @@ func LambdaProxy(methodHandlers map[string]RequestHandler) func(
...
 			"requestMethod":          input.HTTPMethod, // e.g. GET
 			"requestPathParameters":  input.PathParameters,
 			"requestQueryParameters": input.QueryStringParameters,
 			"requestResource":        input.Resource, // e.g. /orgs/{orgId}/accounts/{accountType}
 		})
 
-		handler, ok := methodHandlers[input.HTTPMethod+" "+input.Resource]
+		methodKey := input.HTTPMethod + " " + input.Resource
+
+		operation := oplog.NewManager("admin", lc.InvokedFunctionArn).Start(methodKey).WithMemUsed(lambdacontext.MemoryLimitInMB)
```
austinbyers: Maybe "api" instead of "admin"

pkg/gatewayapi/router.go
```diff
@@ -52,35 +55,36 @@ func LambdaProxy(methodHandlers map[string]RequestHandler) func(
...
+
+		handler, ok := methodHandlers[methodKey]
 		if !ok {
-			logger.Warn("unexpected method/resource")
+			operation.Stop().LogError(errors.New("unexpected method/resource"))
```
austinbyers: It is probably just a warning, not an error - the API service is not erroring if the request is invalid; that's the client's fault

pkg/gatewayapi/router.go
```diff
@@ -52,35 +55,36 @@ func LambdaProxy(methodHandlers map[string]RequestHandler) func(
...
-				zap.Int("statusCode", result.StatusCode), zap.Int("bodyLength", len(result.Body)))
+			operation.Stop().LogSuccess(zap.Int("statusCode", result.StatusCode), zap.Int("bodyLength", len(result.Body)))
 		case result.StatusCode < 500:
-			logger.Warn("client error",
-				zap.Int("statusCode", result.StatusCode), zap.String("responseBody", result.Body))
+			operation.Stop().LogError(errors.New("client error"), zap.Int("statusCode", result.StatusCode), zap.String("responseBody", result.Body))
 		default:
-			logger.Warn("server error",
-				zap.Int("statusCode", result.StatusCode), zap.String("responseBody", result.Body))
+			operation.Stop().LogError(errors.New("server error"), zap.Int("statusCode", result.StatusCode), zap.String("responseBody", result.Body))
```
austinbyers: Again, client errors should not trigger error logs in the API service, just warnings.

Error logs to me indicate something wrong with the runtime of the service itself - a client error means the API service is operating exactly as expected.

internal/compliance/alert_forwarder/main/lambda.go
```diff
@@ -40,26 +43,31 @@ func main() {
...
+	lc, _ := lambdalogger.ConfigureGlobal(ctx, nil)
+	operation := oplog.NewManager("cloudsec", "alert_forwarder").Start(lc.InvokedFunctionArn).WithMemUsed(lambdacontext.MemoryLimitInMB)
+	defer func() {
+		operation.Stop().Log(err, zap.Int("numEvents", len(event.Records)))
+	}()
 
 	for _, record := range event.Records {
 		if record.Change.NewImage == nil {
-			logger.Warn("Skipping records")
+			operation.LogWarn(errors.New("Skipping records"))
```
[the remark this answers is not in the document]
rleighton: I'm using pkg/errors, we get a stack here.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -22,25 +22,33 @@ import (
...
+	"github.com/panther-labs/panther/pkg/oplog"
 )
 
 const maxBackoffSeconds = 30
 
 // Handle is the entry point for the event stream analysis.
 //
 // Do not make any assumptions about the correctness of the incoming data.
-func Handle(batch *events.SQSEvent) error {
+func Handle(lc *lambdacontext.LambdaContext, batch *events.SQSEvent) (err error) {
```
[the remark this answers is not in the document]
rleighton: You merge you change first and deal with the conflicts.

internal/compliance/aws_event_processor/processor/handler_test.go
rleighton: The issue is the new logs have dynamic state so you can't just compare static data. I don't agree tho that is really any weaker as longs as you use 'FilterMessage' get the logs you want to check. Thoughts?

internal/compliance/remediation_processor/main/lambda.go
rleighton: I'll revert.  I suggest when the api/compliance gets re-worked that look closely at the behavior.

internal/compliance/resource_processor/processor/handler.go
rleighton: I was going to leave the Wrap decisions to you when you re-do the lower level error handling you said you planned to do.

internal/compliance/snapshot_poller/pollers/lambda.go
rleighton: Does an Info() level log. Normally you wrap 1 execution (operation) but here we have a loop and the operation wraps it all and we want to log progress. Perhaps this should just be zap.L().Debug(), I do not see much value in the log line.

internal/core/custom_message/api/event_router.go
rleighton: Yup, thx.

internal/compliance/alert_forwarder/main/lambda.go
rleighton: I intentionally did just the main handlers. Lower level errors likely should just Wrap() error and let the caller log. This needs a bigger audit and I want this in place before that work. I know Nick was planning to do some of that soon.

internal/compliance/aws_event_processor/processor/handler.go
rleighton: The logs were very noisy before (oriented to developer's tracing not a stable operational system). It is a judgement call tho. If you assume things are working, then this kind of logging is not really needed.

internal/compliance/resource_processor/processor/handler.go
rleighton: No, need the err because we close over it.

internal/compliance/snapshot_scheduler/main/handler.go
rleighton: Defer closes over err, so need it bound.

pkg/gatewayapi/router.go
rleighton: Good idea, will change everywhere.

pkg/gatewayapi/router.go
rleighton: Will change.

pkg/gatewayapi/router.go
rleighton: That's a "server" error tho.

internal/compliance/aws_event_processor/processor/handler.go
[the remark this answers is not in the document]
nhakmiller: deal_with_it

pkg/gatewayapi/router.go
austinbyers: Not necessarily - if a user in the Panther UI asks for policy XYZ, which doesn't exist, the API will return 404 - not found. That shouldn't log an error in the API service and page someone as if the infrastructure is broken

internal/compliance/snapshot_poller/pollers/lambda.go
rleighton: Will change to Debug log.

pkg/gatewayapi/router.go
rleighton: Will change the result.StatusCode < 500: to Warn(). I think that is what you want.

Issue #107, opened 2020-02-04
[Feature] Allow User Updates 


3nvi:
**Is your feature request related to a problem? Please describe.**
Platform should allow to update the details of a user. Currently, the system only allows you to invite & delete users, but not to update them.

An example of that would be a typo in their name or email.

**Describe the solution you'd like**
Platform should allow user updates

3nvi:
@austinbyers Is there any work related to that that needs to be executed in the backend? If the user has just been invited, perhaps we need to re-send an email if his email is changed

austinbyers:
A user can already update their own first/last name - I just tried with a fresh deployment. Is this issue referring to updating names for other users as well?

We do not allow updating the email address, which was a conscious decision at the time because (a) the `panther-users` Dynamo table is keyed by email address and (b) many security products do not allow you to change email addresses once a user is created (you just delete the user and add a new one).

I believe the frontend sends Cognito userIDs in API calls, so as far as I know updating email is technically feasible. A typo in the email would be fixed before a user could even sign in (since they wouldn't get credentials otherwise), do we feel it's important to be able to update the email?

austinbyers:
As part of the users-api redesign, I'll make sure we allow:

- Updating user name (already possible)
- Updating user email (I think already possible in the backend, but not tested)
- Change role (enterprise edition)
- Reset password

austinbyers:
I've confirmed that as of v0.3.0, you can

- Change first/last name
- Change email
- Reset password

For any user. So there's nothing blocking this in the backend

alexmylonas:
@austinbyers What's the process for changing emails? Do users get a confirmation email on last email or just a prompt on new email?

PR #213, opened 2020-02-19
More cleanup for users-api and organizations-api

### LINKED ISSUE #224: Deploy fails with AWS CLI v2: Template format error
### Describe the bug
When you try to deploy current master, it fails with 

```
ComplianceAPI | Template format error: YAML not well-formed. (line 786, column 24)
```

### Steps to reproduce

1. `git clone git@github.com:panther-labs/panther`
2. `mage deploy`
3. `panther-app` stack rolls back almost immediately with the error above

### Expected behavior

Deploy does not fail

### Environment

- Panther version or commit: master (`84931a`)

### Additional Context
- The actual `compliance-api` YAML template is generated by `mage` with embedded Swagger definition
- Git history shows the `compliance-api` stack hasn't been modified in over a week, so the error may not be related to that service at all

austinbyers:
## Background
We will refactor the `users-api` service before v1 to enable role-based access control for enterprise users. To help me with that process, I'm removing everything we don't need from `users-api` and `organizations-api` so that I can reference just the code that we actually use.

This is also the beginning of the migration - it permanently removes some things we will not need anymore.

Closes #224 

## Changes

- `users-api`
   - Remove `GetUserOrganizationAccess` endpoint - this used to map users to their associated organization in a multi-tenant model. It was used in only one place - the `outputs-api` was using it to see if a user email existed (it didn't actually use the output), but we don't even support email outputs anymore.
    - Remove `userPoolId`, `phone`, and `role`
    - Remove `panther-users` Dynamo table, which literally only had a user email in each row
- `organizations-api`
   - Refactor endpoints to just `GetSettings` and `UpdateSettings`
   - Remove `AwsConfig` and `Phone` attributes - we don't use either of these anymore
   - Add missing IAM permission to fix password reset
- Cognito
   - Remove Admin group. This was not used in the backend, but the frontend was checking Cognito groups before rendering some elements of the page
   - Changed the `deploy` command to just list users from `users-api` instead of asking Cognito to list users in the (now deleted) Admin group
- `web`
    - Remove the `RoleRestrictedAccess` tag
    - Hardcode the user role as `Admin` in the table (since it's no longer reported from the API)
    - `mage fmt` now formats the generated graphql typescript as well
- `outputs-api`
    - Remove unused email output and verification
- CFN changes for compatibility with AWS CLI v2
- Fresh deployments can now specify the company name (which defaults to the same value as before: first-last), just a small thing to make the settings page make more sense when they see it

## Testing

- Deployed from `master`, then deployed this branch - migration worked without any extra steps
    - The org settings (name, email, error reporting) will be blank since it's now stored in Dynamo under a different hash key. You can just edit them again to save the values.
- Deployed fresh from this branch (to ensure Cognito user pool could attach a Lambda function that didn't exist yet)

In both cases, I was able to invite myself, login, update my name, list users, and reset my password


deployments/backend.yml
```diff
@@ -184,13 +184,13 @@ Resources:
...
         LayerVersionArns: !Join [',', !Ref LayerVersionArns]
         TracingMode: !Ref TracingMode
+        UserPoolId: !GetAtt Cognito.Outputs.UserPoolId
       TemplateURL: core/admin_api.yml
 
   Cognito:
     Type: AWS::CloudFormation::Stack
     Properties:
       Parameters:
-        CustomMessageTriggerFunctionArn: !GetAtt AdminAPI.Outputs.CustomMessageTriggerFunctionArn
```
austinbyers: Removed to avoid a circular dependency (AdminAPI => Cognito => AdminAPI)

kostaspap:
> * Remove `GetUserOrganizationAccess` endpoint - this used to map users to their associated organization in a multi-tenant model. It was used in only one place - the `outputs-api` was using it to see if a user email existed (it didn't actually use the output). I'll make sure the new `users-api` includes a way to lookup users by their email address, but for now I've just changed this to list users until it finds the email it's looking for (@kostaspap )
> * Remove `userPoolId` from the API input (which I know will make @3nvi happy) - there's only one pool which is now passed as an environment variable

This was actually used by `outputs-api`  only when an `email` destination was added. This was used so that if an email of a registered user was added as a destination we wouldn't require email confirmation. 
Given that we no longer use SES, you could actually remove that logic - or we can do in subsequent PR.

api/lambda/users/models/validator.go
```diff
@@ -29,7 +29,7 @@ func Validator() *validator.Validate {
 
 func atLeastOneUpdate(sl validator.StructLevel) {
```
kostaspap: It's not clearly visible from this PR, where is this used?

internal/core/outputs_api/verification/get_verification_status.go
```diff
@@ -70,18 +68,31 @@ func (verification *OutputVerification) GetVerificationStatus(input *models.Aler
 	}
 }
 
+// TODO: get user by email directly after users-api refactor
 func (verification *OutputVerification) isVerifiedUserEmail(email *string) (*bool, error) {
```
kostaspap: Actually we currently don't support email as destination - we instead recommend users to use an SNS topic and subscribe their emails to it. 
I believe this function could be removed.

austinbyers:
> This was actually used by outputs-api only when an email destination was added. This was used so that if an email of a registered user was added as a destination we wouldn't require email confirmation. Given that we no longer use SES, you could actually remove that logic - or we can do in subsequent PR.

Awesome, will do!

> This PR changes a lot of APIs (thanks for that!) but it doesn't modify the GraphQL schema. I think it should be relatively easy to do that in this PR (should be just a few more changes) - or do you plan it to be done as part of another issue (issue link? )

Yep, I realized that last night - I'll update AppSync as well, and search for web app references to userPoolId that can be removed now

api/lambda/users/models/validator.go
austinbyers: It's a custom validator for the `UpdateUser` struct:

```go
// Validator builds a custom struct validator.
func Validator() *validator.Validate {
	result := validator.New()
	result.RegisterStructValidation(atLeastOneUpdate, &UpdateUserInput{})
	return result
}
```

Custom validators defined on the struct level allow you to check more complex things than the standard `validate` tags which only apply to specific fields

internal/core/users_api/gateway/update_user_test.go
```diff
@@ -56,45 +55,31 @@ func TestUpdateUserFamilyName(t *testing.T) {
...
 	assert.NoError(t, gw.UpdateUser(&UpdateUserInput{
-		GivenName:   aws.String("Richie"),
-		FamilyName:  aws.String("Homie"),
-		PhoneNumber: aws.String("+1234567890"),
-		Email:       aws.String("rich@homie.com"),
-		ID:          aws.String("user123"),
-		UserPoolID:  aws.String("userPoolId"),
+		GivenName:  aws.String("Richie"),
+		FamilyName: aws.String("Homie"),
+		Email:      aws.String("rich@homie.com"),
```
nhakmiller: One crie for the rich homie :'(

austinbyers:
@kostaspap I went ahead and deleted everything in the backend related to SES email outputs since they're not supported anymore

@3nvi I removed the unused fields from the graphql schema and the AppSync resolvers, but I noticed nowhere does the frontend use the `GetUser` endpoint, only `ListUsers`. Is that because all of the user context (first/last name, email, userID) is part of their Cognito auth token and not pulled from the API? Do you think I should just delete the `GetUser` endpoint if we're not using it? (the backend isn't using it either)

austinbyers:
Tested a fresh deploy again, was able to login, list users, and change my name and password

austinbyers:
Ah, I can no longer see the users / general settings from the navbar:

Screen Shot 2020-02-19 at 5 15 16 PM

I'm guessing these features are now hidden due to the `allowedRoles={ADMIN_ROLES_ARRAY}` gate we have here:

https://github.com/panther-labs/panther/blob/master/web/src/components/navigation.tsx#L180

(Interestingly, I can navigate directly to `/settings/users` but not `/settings/general`, is that a bug?)

@3nvi is the frontend pulling the groups directly from the Cognito auth token? Now that the admin group is deleted, I'm guessing you don't see the user as an "admin" (even though if you were to call `GetUser`, it would return `Role: Admin`)

What do you see as the best way forward? Options as best I can understand them:

-  I can leave the admin group in place, but it seems awkward to have a Cognito group just to let users see things in the frontend when the backend completely ignores it (then we have to always add new users to that group, for example)
- The web app could remove all of these gates (ideal, but I know it may be hard when we sync with enterprise code)
- The web app could call `users-api.GetUser` to see what their role is instead of relying on Cognito (inefficient because extra API call)

I would say can we just remove the gates for now and then re-add them once we've finished redesigning the `users-api`?

api/lambda/organization/models/api.go
```diff
@@ -20,30 +20,17 @@ package models
...
-
 // CreateOrganizationInput creates a new Panther customer account.
 type CreateOrganizationInput struct {
-	AlertReportFrequency  *string    `json:"alertReportFrequency" validate:"omitempty,oneof=P1D P1W"`
-	AwsConfig             *AwsConfig `json:"awsConfig"`
-	DisplayName           *string    `json:"displayName" validate:"required,min=1"`
-	Email                 *string    `genericapi:"redact" json:"email" validate:"required,email"`
-	ErrorReportingConsent *bool      `json:"errorReportingConsent"`
-	Phone                 *string    `genericapi:"redact" json:"phone"`
+	AlertReportFrequency  *string `json:"alertReportFrequency" validate:"omitempty,oneof=P1D P1W"`
```
3nvi: currently we don't use `AlertReportFrequency` anywhere

api/lambda/users/models/api.go
```diff
@@ -22,42 +22,27 @@ package models
...
-	InviteUser                *InviteUserInput                `json:"inviteUser"`
-	ListUsers                 *ListUsersInput                 `json:"listUsers"`
-	RemoveUser                *RemoveUserInput                `json:"removeUser"`
-	ResetUserPassword         *ResetUserPasswordInput         `json:"resetUserPassword"`
-	UpdateUser                *UpdateUserInput                `json:"updateUser"`
+	GetUser           *GetUserInput           `json:"getUser"`
+	InviteUser        *InviteUserInput        `json:"inviteUser"`
+	ListUsers         *ListUsersInput         `json:"listUsers"`
+	RemoveUser        *RemoveUserInput        `json:"removeUser"`
+	ResetUserPassword *ResetUserPasswordInput `json:"resetUserPassword"`
```
3nvi: `resetUserPassword` is not used because it doesn't work. I brought it up in a meeting that we need to allow admins to force a password reset on users.

We can either remove it and re-add it when we implement it or fix it to allow admins to force a cognito password reset

api/lambda/users/models/api.go
```diff
@@ -67,7 +52,6 @@ type InviteUserOutput struct {
 
 // ListUsersInput lists all users in Panther.
 type ListUsersInput struct {
-	UserPoolID      *string `json:"userPoolId" validate:"required,min=1"`
```
3nvi: I don't think we will ever paginate the users. I don't expect users to be more than 50-100 at any organization level.

The UI doesn't use it, but we can keep it as is if you want to (although it's miss-aligned with how the rest of the microservices implement pagination)

deployments/core/appsync.yml
```diff
@@ -239,7 +238,6 @@ Resources:
       DataSourceName: !GetAtt UsersAPILambdaDataSource.Name
       RequestMappingTemplate: |
         #set ($input = $util.defaultIfNull($ctx.args.input, {}))
```
3nvi: we can delete L240

(i..e `#set ($input = $util.defaultIfNull($ctx.args.input, {}))`)

and replace

```yaml
"payload": $util.toJson({
       "updateUser": $input
  })
```

with

```yaml
"payload": $util.toJson({
       "updateUser": $ctx.args.input
  })
```

deployments/core/cognito.yml
```diff
@@ -44,13 +41,13 @@ Resources:
...
             <br />Yours truly,
-            <br />Panther - runpanther.io
+            <br />Panther - https://runpanther.io
             <br />
             <br /><small>Copyright © 2020 Panther Labs Inc. All rights reserved.</small>
       AutoVerifiedAttributes: # Attributes you want the user to verify (poor naming choice by AWS)
         - email
       LambdaConfig:
-        CustomMessage: !Ref CustomMessageTriggerFunctionArn
+        CustomMessage: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-cognito-custom-message-trigger
```
3nvi: should we pass `panther-cognito-custom-message-trigger` as a parameter since it's referenced in two separate nested stacks?

deployments/core/outputs.yml
```diff
@@ -201,21 +186,6 @@ Resources:
...
-        - Id: VerifyEmail
-          Version: 2012-10-17
-          Statement:
-            - Effect: Allow
-              Action:
-                - ses:GetIdentityVerificationAttributes
-                - ses:SendCustomVerificationEmail
-                - ses:VerifyEmailAddress
-              Resource: '*'
-        - Id: UsersAPI
```
3nvi: Is it not needed anymore because we don't support email destinations anymore?

internal/core/organization_api/api/update_organization.go
```diff
@@ -26,11 +26,9 @@ func (API) UpdateOrganization(
```
3nvi: One problem I have with this, is that whenever I want to update just `errorReportingConsent` I have to sent the **entire** organization details.

Since you are refactoring related code, can we make it so that I can update **only** the organization fields that I'm sending? 🥇

internal/core/users_api/api/remove_user.go
```diff
@@ -26,26 +26,11 @@ import (
...
-	user, err := userGateway.GetUser(input.ID, input.UserPoolID)
-	if err != nil {
-		zap.L().Error("error getting user from user pool", zap.Error(err))
-		return err
-	}
-
 	// Delete user from Cognito user pool
-	err = userGateway.DeleteUser(input.ID, input.UserPoolID)
-	if err != nil {
+	if err := userGateway.DeleteUser(input.ID); err != nil {
```
3nvi: @austinbyers I'm guessing this should go here https://github.com/panther-labs/panther/issues/217#issuecomment-588408709, unless you meant that you'll address it in another PR.

internal/core/users_api/api/reset_user_password.go
```diff
@@ -24,5 +24,5 @@ import (
 
 // ResetUserPassword resets a user password.
 func (API) ResetUserPassword(input *models.ResetUserPasswordInput) error {
-	return userGateway.ResetUserPassword(input.ID, input.UserPoolID)
+	return userGateway.ResetUserPassword(input.ID)
```
3nvi: have we tested that this actually works? :P

I think that the role that the `users-api` assumes needs some more roles in order to achieve that. Might be mistaken though

3nvi:
@austinbyers I pushed a commit to change the dropdown from what we had to this

Screen Shot 2020-02-20 at 6 52 21 PM

Feel free to revert it if we prefer the dropdown :)

api/lambda/organization/models/api.go
austinbyers: Good point, and probably won't be relevant for the open-source project. I'll remove

api/lambda/users/models/api.go
austinbyers: I'll see if I can quickly fix it, otherwise I'll just remove it for now and re-add it later

api/lambda/users/models/api.go
austinbyers: I'll leave it as is for now, but in the redesign I'll just handle any paging implicitly in the backend (you're right, it would be very rare) instead of exposing the paging tokens to the frontend that you won't use

deployments/core/cognito.yml
austinbyers: No, it's really a constant, but CloudFormation doesn't really have a notion of constants (a "mapping" is the closest thing).

We can't `!Ref` the value because of a circular dependency, and I'd rather not make parameters we don't need.

We actually hardcode names like this fairly often in CloudFormation. For example, a Lambda function's name is used for the function, its log group, and in the IAM roles of any service that talks to it - all hard-coded to the same value. It's easy enough to find+replace if the name changes

deployments/core/outputs.yml
austinbyers: Correct

internal/core/organization_api/api/update_organization.go
austinbyers: Given time constraints, I might actually just leave the `organizations-api` pretty much as is after this PR, it's really only the `users-api` that needs a complete redesign.

If time allows, I can make that change, sure

internal/core/users_api/api/remove_user.go
austinbyers: Yeah, this PR doesn't change any existing functionality, it just removes a bunch of stuff we didn't need. I'll address #217 and other user-related features in future PRs

internal/core/users_api/api/reset_user_password.go
austinbyers: I was testing by resetting my password on the login page, which I didn't realize was going straight to Cognito. I'll test this today and make necessary fixes

api/lambda/users/models/api.go
austinbyers: Fixed! It was just missing a permission

internal/core/organization_api/api/update_organization.go
austinbyers: Updated! You can now update the general settings, specifying any combination of display name, email, and error reporting

internal/core/users_api/api/reset_user_password.go
austinbyers: Fixed

austinbyers:
@3nvi please review again when you get the chance, I've now:

- refactored `organization-api`
    - you can send only the fields you need
- fixed password reset
- completely removed all references to roles, including all of the `RoleRestrictedAccess` tags from the web components

(the PR description is updated to reflect the full set of changes)

I've tested everything again, it's looking good:

- upgrading an existing deployment
- deploying everything fresh
- integration tests
- login, reset password (both through web UI and directly invoking users-api), invite users, delete users, modify org settings, modify username

**In the next PR**, I'll:

- Verify it's possible to change user email (#107)
- Protect deletion when there's only one user remaining (#217)
- Remove paging from the exposed API and handle it implicitly in the backend

api/lambda/users/models/api.go
3nvi: Great

deployments/core/cognito.yml
3nvi: Gotcha.

In my mind constants are translated to parameters in CFN (just because there is no notion of constants). I guess this is not correct.

internal/core/organization_api/api/update_organization.go
3nvi: Perfect. This saves me additional queries :)

internal/core/users_api/api/remove_user.go
3nvi: cool

internal/core/organization_api/table/update.go
```diff
@@ -31,70 +30,76 @@ import (
...
-			Message: "failed to unmarshal dynamo item to an Organization: " + err.Error()}
+			Message: "failed to unmarshal dynamo item to GeneralSettings: " + err.Error()}
 	}
 
-	return &newOrg, nil
+	return &newSettings, nil
+}
+
+// Update only the fields listed in the request.
+func buildUpdateExpression(settings *models.GeneralSettings) (expression.Expression, error) {
```
3nvi: It's so funny that it's so hard to update only that's sent in Golang :P

deployments/core/admin_api.yml
```diff
@@ -118,20 +104,16 @@ Resources:
...
               Action:
-                - cognito-idp:AdminAddUserToGroup
                 - cognito-idp:AdminCreateUser
                 - cognito-idp:AdminDeleteUser
                 - cognito-idp:AdminDisableUser
                 - cognito-idp:AdminEnableUser
                 - cognito-idp:AdminGetUser
-                - cognito-idp:AdminListGroupsForUser
-                - cognito-idp:AdminRemoveUserFromGroup
+                - cognito-idp:AdminResetUserPassword
```
3nvi: yup, that's the permission I was talking about

tools/mage/deploy.go
```diff
@@ -232,56 +232,56 @@ func postDeploySetup(awsSession *session.Session, backendOutputs map[string]stri
...
 
-	// Prompt the user for email + first/last name
+	// Prompt the user for basic information.
 	logger.Info("setting up initial Panther admin user...")
 	fmt.Println()
 	firstName := promptUser("First name: ", nonemptyValidator)
 	lastName := promptUser("Last name: ", nonemptyValidator)
 	email := promptUser("Email: ", emailValidator)
+	defaultOrgName := firstName + "-" + lastName
+	orgName := promptUser("Company/Team name ("+defaultOrgName+"): ", nil)
```
3nvi: nice

tools/mage/deploy.go
```diff
@@ -232,56 +232,56 @@ func postDeploySetup(awsSession *session.Session, backendOutputs map[string]stri
...
 	}
 	logger.Infof("invite sent to %s: check your email! (it may be in spam)", email)
 
-	// Hit organization-api.CreateOrganization to create organization entry
-	createOrgInput := &orgmodels.LambdaInput{
-		CreateOrganization: &orgmodels.CreateOrganizationInput{
-			Email:       &email,
-			DisplayName: aws.String(firstName + "-" + lastName),
-		},
+	// Hit organization-api.UpdateSettings to create organization entry
```
3nvi: comment should be re-written to something that makes more sense (you are not creating an organization entry)

internal/core/organization_api/table/update.go
3nvi: @austinbyers This does validate that the user can't update the display name or the email of the organization to an empty string, right?

3nvi:
> @3nvi please review again when you get the chance, I've now:
> 
> * refactored `organization-api`
>   
>   * you can send only the fields you need
> * fixed password reset
> * completely removed all references to roles, including all of the `RoleRestrictedAccess` tags from the web components
> 
> (the PR description is updated to reflect the full set of changes)
> 
> I've tested everything again, it's looking good:
> 
> * upgrading an existing deployment
> * deploying everything fresh
> * integration tests
> * login, reset password (both through web UI and directly invoking users-api), invite users, delete users, modify org settings, modify username
> 
> **In the next PR**, I'll:
> 
> * Verify it's possible to change user email (#107)
> * Protect deletion when there's only one user remaining (#217)
> * Remove paging from the exposed API and handle it implicitly in the backend

Reviewed it. Looks fine. Also tested the organization updates and it looks fine as well!

Fixed a couple of minor bugs that TS was not able to pick up (we need more automation for GraphQL <-> TS, which I'm gonna add soon).

Also I altered some variable names and did some minor text changes. 

Please review my commits and if you are good, then feel free to merge

internal/core/organization_api/table/update.go
austinbyers: > This does validate that the user can't update the display name or the email of the organization to an empty string, right?

Yes, that validation happens in the input struct definition (the `validate` tags):

```go
type GeneralSettings struct {
	DisplayName  *string `json:"displayName" validate:"omitempty,min=1"`
	Email        *string `genericapi:"redact" json:"email" validate:"omitempty,email"`
}
```

`validate:"omitempty,min=1"` means IF the field is defined, it must be have length at least 1

austinbyers:
Thanks for your help @3nvi! I fixed the comment, pulled in master, and I'll merge this as soon as CI finishes

Issue #200, opened 2020-02-16
New Log Parsers


jacknagz:
### Description

Add parser support for hosts, networks, containers, and network-based logs to support additional data analysis and historical search capabilities.

The log types in question:

Host:
- [x] [OSSEC](https://www.ossec.net/docs/formats/alerts.html)
- [x] Syslog (RFC 5424 and default formats)
- [ ] [Go-Audit](https://github.com/slackhq/go-audit)
- [ ] [Sysmon](https://eqllib.readthedocs.io/en/latest/guides/sysmon.html) (Windows)
- [ ] [Santa](https://github.com/google/santa)

Application:
- [x] [Apache](https://httpd.apache.org/docs/2.4/logs.html)
- [ ] [EKS Control Plane](https://docs.aws.amazon.com/eks/latest/userguide/control-plane-logs.html)
- [x] Kolide Server
- [ ] [Kubernetes Audit](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/)
- [x] [Teleport](https://github.com/gravitational/teleport) Audit

Network:
- [ ] [Zeek](https://docs.zeek.org/en/current/frameworks/logging.html) (JSON format, [Ref](https://docs.zeek.org/en/current/script-reference/log-files.html))
- [ ] [Suricata](https://suricata.readthedocs.io/en/suricata-4.1.0/output/eve/eve-json-format.html) (JSON format)

### Acceptance Criteria

- New parser(s) for each log type listed above
- New data is properly classified, analyzed, and alerted upon (tested end-to-end)
- 2-3 example rules per log type
- Data is searchable in Athena
- Documentation is updated

stoggi:
Started looking into OSSEC log format https://www.ossec.net/docs/docs/formats/alerts.html

Source code for ossec 3.3.0 **src/analysisd/format/to_json.c** has methods to convert to JSON for the different log types. Including an extended format defined in **src/analysisd/format/json_extended.c**

An initial attempt at a go struct for the EventInfo type is:

```go
type EventInfo struct {
	Rule           *Rule     `json:"rule" validate:"required"`
	ID             *string   `json:"id,omitempty"`
	Timestamp      *int      `json:"Timestamp,omitempty"`
	Decoder        *string   `json:"decoder" validate:"required"`
	DecoderParent  *string   `json:"decoder_parent,omitempty"`
	Action         *string   `json:"action,omitempty"`
	Protocol       *string   `json:"protocol,omitempty"`
[...]

PR #215, opened 2020-02-19
OSSEC Parser for EventInfo logs

stoggi:
## Background

This PR adds an OSSEC Parser for the EventInfo log types as listed in #200 

Created an issue https://github.com/ossec/ossec-hids/issues/1838 to track a possible regression in the OSSEC JSON log formatter.

## Changes

- Created a new parser `OSSEC.EventInfo` to parse the [EventInfo format](https://www.ossec.net/docs/docs/formats/alerts.html) 
- Used the [Eventinfo_to_jsonstr()](https://github.com/ossec/ossec-hids/blob/bf7d2670fe089be891caededb44b208c8f49189a/src/analysisd/format/to_json.c#L21-L211) function to create the JSON EventInfo struct.
- Required fields deduced from behaviour of [OS_Log()](https://github.com/ossec/ossec-hids/blob/6134b14ba703111cb5906616a7c57db8979c0aa3/src/analysisd/alerts/log.c#L168-L251) in github.com/ossec/ossec-hids
- Added a new `timestamp.UnixMillisecond` convenience type to parse the timestamp and marshal Unix timestamps in milliseconds (int or string)
- Registered the parser as `OSSEC.EventInfo`
- Added it as a selectable option in `LOG_TYPES`

## Testing

- Added unit tests for the parser and new timestamp type
- Deployed into a test account, to add a new rule and tested via the UI
- Added sample inputs into an s3 bucket and verified an alert is generated

## Handy Resources
https://www.ossec.net/
https://github.com/ossec/ossec-hids
https://www.ossec.net/docs/docs/formats/alerts.html


internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
```diff
@@ -0,0 +1,152 @@
...
+	return &EventInfoParser{}
+}
+
+// Parse returns the parsed events or nil if parsing failed
+func (p *EventInfoParser) Parse(log string) []interface{} {
+	eventInfo := &EventInfo{}
+
+	err := jsoniter.UnmarshalFromString(log, eventInfo)
+	if err != nil {
+		fmt.Println(err)
```
jacknagz: This looks leftover 😛

stoggi: Ah good spotting! Thanks

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
```diff
@@ -0,0 +1,149 @@
+package ossec
```
kostaspap: We tend to name the packages with `logs` suffix, see `awslogs`,`nginxlogs`, etc  In this case it should be `osseclog`.

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
```diff
@@ -0,0 +1,149 @@
...
+	// Required
+	ID        *string                    `json:"id" validate:"required" description:"Event ID (timestamp.streamposition)"`
+	Rule      *Rule                      `json:"rule" validate:"required" description:"Rule object with information about what rule created the alert"`
+	Timestamp *timestamp.UnixMillisecond `json:"TimeStamp" validate:"required" description:"Timestamp in Unix Epoch (Milliseconds)"`
+	Location  *string                    `json:"location" validate:"required" description:"Source of the alert (filename or command)"`
+	Hostname  *string                    `json:"hostname" validate:"required" description:"Hostname of the host that generated the alert"`
+	FullLog   *string                    `json:"full_log" validate:"required" description:"The full captured log of the alert"`
+
+	// Optional
+	Action             *string   `json:"action,omitempty" description:"Action"`
```
kostaspap: Descriptions should have something more meaningful .

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/parsers/timestamp"
+)
+
+var EventInfoDesc = `OSSEC EventInfo alert parser. Currently only JSON output is supported.
+Reference: https://www.ossec.net/docs/docs/formats/alerts.html`
+
+// nolint:lll
+type EventInfo struct {
+	// Required
+	ID        *string                    `json:"id" validate:"required" description:"Event ID (timestamp.streamposition)"`
```
kostaspap: I am not sure I understand the description for this field. Can you elaborate on it (not sure what the `timestamp.streamposition`) means. Keep in mind that this description will be displayed to people querying the logs

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
```diff
@@ -0,0 +1,149 @@
...
+	Rule      *Rule                      `json:"rule" validate:"required" description:"Rule object with information about what rule created the alert"`
```
kostaspap: validation tag should include the `dive` directive, so that validation will be performed in the `Rule` structure.

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
```diff
@@ -0,0 +1,149 @@
...
+	FullLog   *string                    `json:"full_log" validate:"required" description:"The full captured log of the alert"`
+
+	// Optional
+	Action             *string   `json:"action,omitempty" description:"Action"`
+	AgentIP            *string   `json:"agentip,omitempty" description:"AgentIP"`
+	AgentName          *string   `json:"agent_name,omitempty" description:"AgentName"`
+	Command            *string   `json:"command,omitempty" description:"Command"`
+	Data               *string   `json:"data,omitempty" description:"Data"`
+	Decoder            *string   `json:"decoder,omitempty" description:"Decoder"`
+	DecoderDescription *Decoder  `json:"decoder_desc,omitempty" description:"DecoderDescription"`
```
kostaspap: See comment above. Add the `dive` validation directive for this field.

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
```diff
@@ -0,0 +1,149 @@
...
+	Logfile            *string   `json:"logfile,omitempty" description:"Logfile"`
+	PreviousOutput     *string   `json:"previous_output,omitempty" description:"PreviousOutput"`
+	ProgramName        *string   `json:"program_name,omitempty" description:"ProgramName"`
+	Protocol           *string   `json:"protocol,omitempty" description:"Protocol"`
+	SrcGeoIP           *string   `json:"srcgeoip,omitempty" description:"SrcGeoIP"`
+	SrcIP              *string   `json:"srcip,omitempty" description:"SrcIP"`
+	SrcPort            *string   `json:"srcport,omitempty" description:"SrcPort"`
+	SrcUser            *string   `json:"srcuser,omitempty" description:"SrcUser"`
+	Status             *string   `json:"status,omitempty" description:"Status"`
+	Syscheckfile       *FileDiff `json:"SyscheckFile,omitempty" description:"Syscheckfile"`
```
kostaspap: `dive`

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
```diff
@@ -0,0 +1,149 @@
...
+
+type Rule struct {
+	// Required
+	Comment *string `json:"comment" validate:"required"`
+	Group   *string `json:"group" validate:"required"`
+	Level   *int    `json:"level" validate:"required"`
+	SIDID   *int    `json:"sidid" validate:"required"`
+
+	// Optional
+	CIS        *[]string `json:"CIS,omitempty"`
```
kostaspap: Slices we use to define as `[]` (without the `*`). Same applies to every similar field.

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
```diff
@@ -0,0 +1,149 @@
...
+	return &EventInfoParser{}
+}
+
+// Parse returns the parsed events or nil if parsing failed
+func (p *EventInfoParser) Parse(log string) []interface{} {
+	eventInfo := &EventInfo{}
+
+	err := jsoniter.UnmarshalFromString(log, eventInfo)
+	if err != nil {
+		zap.L().Error("failed to parse log", zap.Error(err))
```
kostaspap: Should be Debug level. 

When we process events, we try all parsers to check the type of the event. If we fail to parse it doesn't mean there was an issue with the application - it means that the event was not of the same type of the parser.

internal/log_analysis/log_processor/parsers/ossec/eventinfo_test.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/parsers/timestamp"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEventInfo(t *testing.T) {
```
kostaspap: Could you add a couple more events here? 
There are some structs that this test is not verifying - FileDiff, Decoder

internal/log_analysis/log_processor/parsers/timestamp/timestamp.go
```diff
@@ -78,3 +80,25 @@ func (ts *ANSICwithTZ) UnmarshalJSON(text []byte) (err error) {
...
+func (ts *UnixMillisecond) String() string {
+	return (*time.Time)(ts).UTC().String() // ensure UTC
+}
+
+func (ts *UnixMillisecond) MarshalJSON() ([]byte, error) {
+	return []byte((*time.Time)(ts).UTC().Format(jsonMarshalLayout)), nil // ensure UTC
+}
+
+func (ts *UnixMillisecond) UnmarshalJSON(jsonBytes []byte) (err error) {
+	text := strings.Trim(string(jsonBytes), "\"")
```
kostaspap: I see why this might be useful but is this needed here? In your test the value is `"TimeStamp":1510376401000`

kostaspap:
Thanks for the PR! I have put some comments but code looks very good overall!

internal/log_analysis/log_processor/parsers/timestamp/timestamp.go
rleighton: Instead of "\\"" , you could write 
```
`"` 
```
and avoid the escape (more clear)

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
```diff
@@ -0,0 +1,149 @@
...
+
+var EventInfoDesc = `OSSEC EventInfo alert parser. Currently only JSON output is supported.
+Reference: https://www.ossec.net/docs/docs/formats/alerts.html`
+
+// nolint:lll
+type EventInfo struct {
+	// Required
+	ID        *string                    `json:"id" validate:"required" description:"Event ID (timestamp.streamposition)"`
+	Rule      *Rule                      `json:"rule" validate:"required" description:"Rule object with information about what rule created the alert"`
+	Timestamp *timestamp.UnixMillisecond `json:"TimeStamp" validate:"required" description:"Timestamp in Unix Epoch (Milliseconds)"`
```
rleighton: The description is not about the input, but about the output (the string will be in the Athena table, visible to users). So the description should be "Timestamp UTC".

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
```diff
@@ -0,0 +1,149 @@
...
+	GroupOwnerAfter  *string `json:"gowner_after,omitempty"`
+	GroupOwnerBefore *string `json:"gowner_before,omitempty"`
+	Md5After         *string `json:"md5_after,omitempty"`
+	Md5Before        *string `json:"md5_before,omitempty"`
+	OwnerAfter       *string `json:"owner_after,omitempty"`
+	OwnerBefore      *string `json:"owner_before,omitempty"`
+	Path             *string `json:"path,omitempty"`
+	PermAfter        *int    `json:"perm_after,omitempty"`
+	PermBefore       *int    `json:"perm_before,omitempty"`
+	SHA1After        *string `json:"sha1_after,omitempty"`
```
rleighton: This is the first dataset with file hashes. Can you add a "standard" field for SHA1, here:

https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/parsers/patherlog.go

On future parsers, if you find other hashes (e.g,  md5, sha256, etc), please add as standard fields.

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
jacknagz: @kostaspap wouldn't we then want `osseclogs` (plural)?

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
jacknagz: I would have thought it should be `[]*string`

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
stoggi: > The description is not about the input, but about the output 
Ah this make more sense thanks.

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
stoggi: Should I still use the `dive` directive if no fields in the struct are required?

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
stoggi: Ah yes, good spotting. I think this was left over from trying to output errors in unit tests. Will change back to `Debug`

internal/log_analysis/log_processor/parsers/ossec/eventinfo_test.go
stoggi: Good point. I don't have a huge number of input samples, but I'll spin up an OSSEC instance and try to get a log of each type.

internal/log_analysis/log_processor/parsers/timestamp/timestamp.go
stoggi: For OSSEC logs I think they will always be Integers. But since this is a more generic type, the next parser that uses the UnixMIllisecond type it's not immediately obvious that it only supports JSON integers. Should it throw an error instead?

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
stoggi: I can add validate:"omitempty,dive"` and it works as expected.

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
stoggi: `[]string` is used in alb and s3_server_access logs, so I'll follow suit with `[]string`

internal/log_analysis/log_processor/parsers/ossec/eventinfo.go
stoggi: This is a good point. I'm not familiar enough with OSSEC to describe these fields well. I'll ask in their mailing list if there is any relevant documentation.

stoggi: Found these links to help.
https://www.ossec.net/docs/docs/syntax/head_rules.html
https://www.ossec.net/docs/docs/syntax/head_decoders.html#rules-decoders

internal/log_analysis/log_processor/parsers/osseclogs/eventinfo.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// LogType returns the log type supported by this parser
+func (p *EventInfoParser) LogType() string {
+	return "OSSEC.EventInfo"
+}
+
+func (event *EventInfo) updatePantherFields(p *EventInfoParser) {
+	event.SetCoreFieldsPtr(p.LogType(), (*timestamp.RFC3339)(event.Timestamp))
+	event.AppendAnyIPAddressPtrs(event.SrcIP, event.DstIP)
```
kostaspap: You should probably add here the `AgentIP` field.

stoggi: I thought about that, but I'm not sure it's actually an IP address. I don't have a log example using it, but I think it gets extracted from the hostname.

stoggi: This is the code that adds it:
```c
    if(lf->hostname[0] == '(') {
        char* search;
        char string[MAX_STRING];
        strncpy(string, lf->hostname, MAX_STRING - 1);
        int index;
        search = strchr(string, ')');
        if(search) {
            index = (int)(search - string);
            str_cut(string, 0, index);
            str_cut(string, 0, 2);
            search = strchr(string, '-');
            index = (int)(search - string);
            str_cut(string, index, -1);
            cJSON_AddStringToObject(root, "agentip", string);
        }

    }
```

internal/log_analysis/log_processor/parsers/timestamp/timestamp.go
kostaspap: I think the method will throw an error later on (when it tries to parse the it) so you don't need explicit error. It's fine just to leave this line out for the moment. This is currently used only by OSSEC parser, if any parsers needs to use it in the future it can be updated. 

It would make sense to leave this piece of code in if this code was some "generic" library that we were shipping. This case is however only used by Panther app - and the app doesn't need this functionality now (it just adds some extra line and some extra tests - and it's basically unused code).

internal/log_analysis/log_processor/parsers/osseclogs/eventinfo.go
```diff
@@ -0,0 +1,156 @@
...
+	if event.SyscheckFile != nil {
+		event.AppendAnyMD5HashPtrs(event.SyscheckFile.MD5Before, event.SyscheckFile.MD5After)
```
rleighton: Love it, we have hashes now!

rleighton:
Can do I more thing before pushing to master and update the docs as you add parsers here: https://github.com/panther-labs/panther/blob/master/docs/gitbook/log-analysis/supported-logs/README.md

PR #218, opened 2020-02-19
Cloud Security & Log Analysis - Navbar split

3nvi:
## Background

This PR makes sure to make the necessary changes so that Panther can showcase `Cloud Security` and `Log Analysis` as two separate sub-products of the same app.

It essentially is the first step to isolating the two apps. 

The second step would be to split source pages (into a log source & compliance source page) and to add "coming soon" tags to WIP pages (like the log analysis overview). 

**This PR is complimented by #239. It's just here to ease reviewing of the entire task**

## Changes

- Re-organize navigation
- Re-organize URL structure
- Fix issues with lengthy breadcrumbs (due to addition of an additional breadcrumb "step")
- Configure proper primary/secondary menu UX
- Update URLs all across the app

## Testing

- Locally

## Screenshots

Screen Shot 2020-02-20 at 9 14 57 AM
Screen Shot 2020-02-20 at 9 15 16 AM


web/src/components/breadcrumbs.tsx
```diff
@@ -56,8 +56,12 @@ const Breadcrumbs: React.FC = () => {
     // is before-and-including our key). The key is essentially the URL path itself just prettified
     // for displat
     return pathKeys.map(key => ({
-      text: capitalize(decodeURIComponent(key).replace(/-_/g, ' ')),
       href: `${pathname.substr(0, pathname.indexOf(`/${key}/`))}/${key}/`,
+      text: decodeURIComponent(key)
+        .replace(/([-_])+/g, ' ')
```
3nvi: fix a bad regex & make sure to capitalize all words (not just the first in each phrase)

web/src/components/header.tsx
```diff
@@ -29,16 +29,25 @@ const Header = () => {
...
-        <Flex alignItems="center">
-          <Icon type="user" size="small" mr={2} borderRadius="circle" bg="grey200" color="white" />
-          {userInfo && (
-            <Text size="medium">
-              {userInfo.given_name} {userInfo.family_name[0]}.
-            </Text>
-          )}
-        </Flex>
-      </Button>
+      <Box flex="0 0 auto">
```
3nvi: add `flex="0 0 auto"` so that the name of the user is never split into two lines

web/src/components/navigation/compliance-navigation.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
3nvi: Navigation for cloud security

web/src/components/navigation/log-analysis-navigation.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
3nvi: Navigation for log analysis sub-product

web/src/components/navigation/navigation.tsx
```diff
@@ -0,0 +1,132 @@
+/**
```
3nvi: top level product navigation & logic for knowing which sub-product navigation to show

web/src/components/navigation/settings-navigation.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
3nvi: navigation for the settings sub-pages

jacknagz:
@3nvi Awesome, I am really hyped for this! 

Also, I think you meant to say:

> so that Panther can showcase Cloud Security and Log Analysis as two separate sub-products of the same app.

And would you mind adding a screenshot?

3nvi:
> @3nvi Awesome, I am really hyped for this!
> 
> Also, I think you meant to say:
> 
> > so that Panther can showcase Cloud Security and Log Analysis as two separate sub-products of the same app.
> 
> And would you mind adding a screenshot?

Yeah I meant to, but got carried away. Updated the description.

kostaspap:
Deployed the branch to my dev account some issues I see: 
1. The "sources" page is blank for both log analysis and cloud security
2. I am not sure Log Analysis "Overview" makes sense (we don't have such thing). I think the equivalent Cloud Security "Overview" would the Alerts page. So I could create a list of "Alerts", "Rules", "Sources" in that order. 
3. My personal opinion is that the Panther icon no longer needs to be clickable since we don't have a "general" overview page. Users can select any of the rest. 
4. Would it make sense to bring the "Documentation" and "Settings" icons on the sidebar higher, right below the log analysis?

3nvi:
> Deployed the branch to my dev account some issues I see:
> 
> 1. The "sources" page is blank for both log analysis and cloud security

From the PR description, this PR:

>It essentially is the first step to isolating the two apps.

> The second step would be to split source pages (into a log source & compliance source page) and to > add "coming soon" tags to WIP pages (like the log analysis overview).

> This PR is complimented by #239. It's just here to ease reviewing of the entire task

> 2. I am not sure Log Analysis "Overview" makes sense (we don't have such thing). I think the equivalent Cloud Security "Overview" would the Alerts page. So I could create a list of "Alerts", "Rules", "Sources" in that order.

The finalized designs for v2 have just that: https://projects.invisionapp.com/share/S3VS7B3GFN4#/405349154_runpanther_-_Log_Analysis_-_Alerts_-_Dark

> 3. My personal opinion is that the Panther icon no longer needs to be clickable since we don't have a "general" overview page. Users can select any of the rest.

Every Icon in every app is clickable. It's what the user expects. I agree that we shouldn't point it to compliance, but it should be a link. This is outside the scope of the PR, but I can point it in an "introductory" page where you can select which product do you want to focus on (another PR before v1)

> 4. Would it make sense to bring the "Documentation" and "Settings" icons on the sidebar higher, right below the log analysis?

Same answer as (2) :)

web/src/components/breadcrumbs.tsx
kostaspap: Why is the replace needed here?

kostaspap:
> > 1. I am not sure Log Analysis "Overview" makes sense (we don't have such thing). I think the equivalent Cloud Security "Overview" would the Alerts page. So I could create a list of "Alerts", "Rules", "Sources" in that order.
> 
> The finalized designs for v2 have just that: https://projects.invisionapp.com/share/S3VS7B3GFN4#/405349154_runpanther_-_Log_Analysis_-_Alerts_-_Dark
> 

Actually when I deployed this branch, I saw that the "Overview" for Log Analysis was just showing a blank page. I didn't know if it was a bug or it was intentional (comes out it was intentional). 
I see that this issue is getting fixed in #239 and a proper message is display ("Coming soon"...) . I think it's fine to proceed like this.

kostaspap:
Ok, a UX-related comment. I find the navigation between the  pages a bit un-natural.... 
Let's say, I am viewing the "Cloud Security"/"Policies" page and I want to to see the "Log Analysis"/"Rules" page. The process is the following: 
1. I navigate the mouse to the "Log Analysis" button on the left panel. 
2. I click on it. This pops up the different options for "Log Analysis". The main panel contains the previous screen even though I clicked on the "Log Analysis"
3. I click on "Rules" from the extended left panel. This changes the main panel to display me the rules. 

I think a different flow would make the experience better. My suggestion is that the extended panel gets shown when a user hovers the mouse of the "Log Analysis" or "Cloud Security" or "Settings". Then, clicking on the appropriate field ("Overview", "Rules", etc ) updates the main panel. Basically this reduces the mouse clicks required by one. The buttons on the left are anyway small, so I don't think it's going to be frequent that people hover over these buttons "by mistake".

Another idea is that clicking on the "Cloud Security", "Log Analysis", "Settings" immediately changes the main panel to display "Cloud Security"/"Overview", "Log Analysis"/"Overview", "Settings"/"General". 

That's my personal opinion though, if majority has different opinion I'll accept this.

web/src/components/breadcrumbs.tsx
3nvi: because I need to convert `cloud-security` (which is part of the URL) to `Cloud Security`

kostaspap: Interesting. Should we be doing sth similar in web/src/components/navigation/compliance-navigation.tsx , line 28  and web/src/components/navigation/log-analysis-navigation.tsx line 28? There is seems to be hardcoded. (I'm not a UI expert so it might be a stupid question)

3nvi: Breadcrumbs are globally shared, so this is the only component that handles breadcrumbs generally.

The navigation component has to do with the left sidebars, not with the breadcrumbs. I just made this change cause we never has a "dashed" url before, so I didn't notice this bug earlier

3nvi:
This PR has been open for 5 days now. Just because it's becoming hard to keep it in sync (since there are multiple conflicts every day), please help me either close it or merge it

3nvi:
> Ok, a UX-related comment. I find the navigation between the pages a bit un-natural....
> Let's say, I am viewing the "Cloud Security"/"Policies" page and I want to to see the "Log Analysis"/"Rules" page. The process is the following:
> 
> 1. I navigate the mouse to the "Log Analysis" button on the left panel.
> 2. I click on it. This pops up the different options for "Log Analysis". The main panel contains the previous screen even though I clicked on the "Log Analysis"
> 3. I click on "Rules" from the extended left panel. This changes the main panel to display me the rules.
> 
> I think a different flow would make the experience better. My suggestion is that the extended panel gets shown when a user hovers the mouse of the "Log Analysis" or "Cloud Security" or "Settings". Then, clicking on the appropriate field ("Overview", "Rules", etc ) updates the main panel. Basically this reduces the mouse clicks required by one. The buttons on the left are anyway small, so I don't think it's going to be frequent that people hover over these buttons "by mistake".
> 
With regards to that I spoke to Jim on how he envisioned it and he said that he had it as a click as well. Our unified reasoning is:

1. Tablet users can't hover
2. Accidentally hovering over it can cause random sidebars opening and closing
3. You need to be precise in your cursor movement since if you hover out for a tiny bit, the submenu will close

Implementation wise, this is hard, because the entity you want to hover is a circle that doesn't "touch" the sidebar. The way hover-navigation works, is that the two elements are placed in such a way, so that the cursor is always hovering on either one or the other. That's why you will see that most hover-navigations, have adjacent blocks touching one another (no pun intended). 

Ours has a gap between the hover-able icon and the sidebar which I will have to handle with a "hover-out" grace period. This means a JS timer, which will add complexity.

All in all I feel that this approach is not currently well thought out by the entire team and its quick implementation will lead into unforeseen issues. I don't have an issues implementing a hover navigation despite its advantages & disadvantages, but only after we fully plan it with the  designer

> Another idea is that clicking on the "Cloud Security", "Log Analysis", "Settings" immediately changes the main panel to display "Cloud Security"/"Overview", "Log Analysis"/"Overview", "Settings"/"General".

That can only be implemented if navigation changes to a hover, since now it can't make sense. If it does change, I'm not opposed to this

PR #221, opened 2020-02-19
 Move the cloudformation for the glue tables to a dedicated stack 

rleighton:
## Background
Currently if the Glue tables or the CF (resource names) for the Glue tables change significantly it is very difficult to easily update the deployment because the table CF is under a nested stack. 

Developers sometimes will want to drop the stack and just allow it to be re-created.
Also, in the future we will likely have data migrations and having this dynamically  created configuration as a nested stack will require complex migrations if we cannot just drop+recreate.
Since this is just meta data, with no hard dependencies, dropping and re-creating is not a problem for this CF.

This change moves the CF for the glue tables to an independent stack.

## Testing
mage test:ci
mage deploy


deployments/backend.yml
```diff
@@ -42,6 +42,12 @@ Parameters:
     AllowedValues: ['', Active, PassThrough]
     Default: ''
 
+  # Set automatically by "mage deploy"
+  PantherLogProcessingDatabase:
+    Type: String
+    Description: The name of the database for the output of log processing
+    Default: ''
```
austinbyers: If it's set automatically and is required for the deployment to work, don't set a default value

tools/cfngen/cf.go
```diff
@@ -87,3 +88,10 @@ func NewTemplate(description string, parameters map[string]interface{}, resource
 	}
 	return
 }
+
+// Re-map characters not allowed in CF names consistently (CF resources must be alphanum)
+var cleanResourceName = regexp.MustCompile(`([^[:alpha:]])`)
+
+func ResourceClean(name string) string {
```
austinbyers: I would name this something like `SanitizeCfnResourceName`

`ResourceClean` makes me think the resource itself is being modified

tools/mage/deploy.go
```diff
@@ -114,6 +113,9 @@ func Deploy() {
 		logger.Fatal(err)
 	}
 
+	// Creates Glue/Athena related resources
+	deployDatabases(awsSession, bucket, backendOutputs)
```
austinbyers: This could actually run in parallel with the frontend stack, but I plan to (hopefully) make that change later

austinbyers:
@rleighton looks like this PR is missing the `deployments/databases.yml` template:

```
exec: aws cloudformation package --output-template-file out/deployments/package.databases.yml --s3-bucket panther-buckets-source-omntmnyznm6n --s3-prefix panther-app-databases --template-file deployments/databases.yml
Invalid template path deployments/databases.yml
```

rleighton:
Oops, that’s bad. Will fix when I get home

PR #222, opened 2020-02-19
Adding Alert forwarder component

### LINKED ISSUE #183: Add alert_forwarder component
Make sure to link this Story to a related Epic!

## Introduction

Alert forwarder component will update the alerts DB when new alert is created or existing alert is updated. 

It feeds of the alert-dedup DB that is updated by the rules engine

kostaspap:
## Background

This PR introduces the Alert forwarder Lambda. This Lambda listens for changes in the `panther-alert-dedup` table. `panther-alert-dedup` table is used for alert deduplication 

The Lambda takes the DDB update and stores the associated information on the `panther-log-alerts-info` table. This table will eventually replace `panther-log-alerts` table. 

Closes #183 

## Changes

- Added new Lambda and DDB table. 

## Testing

- Unit tests
- Deployed in my account

### DIFF
--- a/deployments/log_analysis/alerts.yml
+++ b/deployments/log_analysis/alerts.yml
@@ -330,7 +330,7 @@ Resources:
     Type: AWS::Serverless::Function
     Properties:
       CodeUri: ../../out/bin/internal/log_analysis/alert_forwarder/main
-      Description: Lambda for storing events from
+      Description: Lambda that creates and updates alerts in alerts-info table
       Environment:
         Variables:
           DEBUG: !Ref Debug
--- a/deployments/log_analysis/alerts.yml
+++ b/deployments/log_analysis/alerts.yml
@@ -350,17 +350,6 @@ Resources:
       Timeout: 30
       Tracing: !If [TracingEnabled, !Ref TracingMode, !Ref 'AWS::NoValue']
       Policies:
-        - Id: ReadTableStreams
-          Version: 2012-10-17
-          Statement:
-            - Effect: Allow
-              Action:
-                - dynamodb:DescribeStream
-                - dynamodb:GetRecords
-                - dynamodb:GetShardIterator
-                - dynamodb:ListStreams
-              Resource: !Sub ${LogAlertsTable.Arn}/stream/*
-
         - Id: ManageAlerts
           Version: 2012-10-17
           Statement:
--- a/deployments/log_analysis/alerts.yml
+++ b/deployments/log_analysis/alerts.yml
@@ -358,5 +358,3 @@ Resources:
                 - dynamodb:PutItem
               Resource:
                 - !GetAtt LogAlertsTable.Arn
-
-
--- a/deployments/log_analysis/alerts.yml
+++ b/deployments/log_analysis/alerts.yml
@@ -44,6 +44,9 @@ Parameters:
   SQSKeyId:
     Type: String
     Description: KMS key ID for SQS encryption
+  AlertDedupTableStreamArn:
+    Type: String
+    Description: The stream arn of the alerts dedup table
 
 Conditions:
   AttachLayers: !Not [!Equals [!Join ['', !Ref LayerVersionArns], '']]
@@ -336,7 +339,7 @@ Resources:
         DynamoDBEvent:
           Type: DynamoDB
           Properties:
-            Stream: !GetAtt LogAlertsTable.StreamArn
+            Stream: !Ref AlertDedupTableStreamArn
             StartingPosition: TRIM_HORIZON
             BatchSize: 10
       FunctionName: panther-log-alert-forwarder
--- a/deployments/log_analysis/alerts.yml
+++ b/deployments/log_analysis/alerts.yml
@@ -273,3 +273,98 @@ Resources:
                 - !GetAtt AlertsTable.Arn
                 - !GetAtt EventsTable.Arn
                 - !GetAtt RecentAlertsTable.Arn
+
+  ##### Dynamo table that stores alert information #####
+  LogAlertsTable:
+    Type: AWS::DynamoDB::Table
+    Properties:
+      TableName: panther-log-alerts-info
+      AttributeDefinitions:
+        - AttributeName: id
+          AttributeType: S
+        - AttributeName: creationTime
+          AttributeType: S
+        - AttributeName: ruleId
+          AttributeType: S
+        - AttributeName: timePartition
+          AttributeType: S
+      BillingMode: PAY_PER_REQUEST
+      GlobalSecondaryIndexes:
+        - # Add an index ruleId to efficiently list alerts for a specific rule
+          KeySchema:
+            - AttributeName: ruleId
+              KeyType: HASH
+            - AttributeName: creationTime
+              KeyType: RANGE
+          IndexName: ruleId-creationTime-index
+          Projection:
+            ProjectionType: ALL
+        - # Add an index using timePartition to efficiently list alerts by creationTime
+          KeySchema:
+            - AttributeName: timePartition
+              KeyType: HASH
+            - AttributeName: creationTime
+              KeyType: RANGE
+          IndexName: timePartition-creationTime-index
+          Projection:
+            ProjectionType: ALL
+      KeySchema:
+        - AttributeName: id
+          KeyType: HASH
+      PointInTimeRecoverySpecification:
+        PointInTimeRecoveryEnabled: True
+      SSESpecification:
+        SSEEnabled: True
+
+  ##### Alert forwarder Lambda
+  AlertForwarderLogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: /aws/lambda/panther-log-alert-forwarder
+      RetentionInDays: !Ref CloudWatchLogRetentionDays
+
+  AlertsForwarderFunction:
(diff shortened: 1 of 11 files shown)


deployments/log_analysis/alerts.yml
```diff
@@ -299,3 +301,98 @@ Resources:
...
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: /aws/lambda/panther-log-alert-forwarder
+      RetentionInDays: !Ref CloudWatchLogRetentionDays
+
+  AlertsForwarderFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: ../../out/bin/internal/log_analysis/alert_forwarder/main
+      Description: Lambda for storing events from
```
austinbyers: Incomplete description

deployments/log_analysis/alerts.yml
```diff
@@ -299,3 +301,98 @@ Resources:
...
+        - Id: ReadTableStreams
+          Version: 2012-10-17
+          Statement:
+            - Effect: Allow
+              Action:
+                - dynamodb:DescribeStream
+                - dynamodb:GetRecords
+                - dynamodb:GetShardIterator
+                - dynamodb:ListStreams
+              Resource: !Sub ${LogAlertsTable.Arn}/stream/*
```
austinbyers: I'm not sure if you need this - SAM might automatically add the policy because the stream is the event source

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -0,0 +1,64 @@
...
+func Process(event *AlertDedupEvent) error {
+	alert := &Alert{
+		ID:              generateAlertID(event),
+		TimePartition:   defaultTimePartition,
+		AlertDedupEvent: *event,
+	}
+
+	marshalledAlert, err := dynamodbattribute.MarshalMap(alert)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshall alert")
```
austinbyers: here and elsewhere, `marshal` should have only one "l"

internal/log_analysis/alert_forwarder/forwarder/models.go
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+import (
+	"time"
+
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/pkg/errors"
+)
+
+// AlertDedupEvent represents the event sent to the AlertProcessor by the compliance engine.
```
austinbyers: Is this comment still accurate?

internal/log_analysis/alert_forwarder/forwarder/models.go
```diff
@@ -0,0 +1,74 @@
...
+type AlertDedupEvent struct {
```
austinbyers: If this model is meant to be imported by another backend service, consider moving it into a separate `models` package so that the entire `forwarder` logic doesn't need to be imported by the caller

internal/log_analysis/alert_forwarder/forwarder/models.go
```diff
@@ -0,0 +1,74 @@
...
+	RuleID              string    `dynamodbav:"ruleId,string"`
+	DeduplicationString string    `dynamodbav:"dedup,string"`
+	AlertCount          int64     `dynamodbav:"-"` // Not storing this field in DDB
+	CreationTime        time.Time `dynamodbav:"creationTime,string"`
+	UpdateTime          time.Time `dynamodbav:"updateTime,string"`
+	EventCount          int64     `dynamodbav:"eventCount,number"`
```
austinbyers: I don't think you need the `string` nor the `number` tags - Dynamo will automatically store them that way. You can leave them if you like, but they don't add any value

internal/log_analysis/alert_forwarder/main/lambda.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/panther-labs/panther/internal/log_analysis/alert_forwarder/forwarder"
+	"github.com/panther-labs/panther/pkg/lambdalogger"
+)
+
+func main() {
+	lambda.Start(reporterHandler)
+}
+
+func reporterHandler(ctx context.Context, event events.DynamoDBEvent) error {
+	lambdalogger.ConfigureGlobal(ctx, nil)
```
austinbyers: #199 changes the logging setup for all Lambda functions, we should incorporate that now

internal/log_analysis/alert_forwarder/main/lambda.go
```diff
@@ -0,0 +1,50 @@
...
+
+	for _, record := range event.Records {
+		event, err := forwarder.FromDynamodDBAttribute(record.Change.NewImage)
+		if err != nil {
+			return errors.Wrap(err, "failed to unmarshall new image")
+		}
+
+		if err := forwarder.Process(event); err != nil {
+			return errors.Wrap(err, "encountered issue while handling event")
```
austinbyers: You're ok with failing an entire batch of events if just one of them fails to send?

What if you forward the first 5 alerts successfully, then the 6th one fails. The first 5 will be sent again, right?

internal/log_analysis/alert_forwarder/main/lambda.go
rleighton: Simple example here: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/main/lambda.go#L46

deployments/log_analysis/alerts.yml
kostaspap: Updated

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
kostaspap: Thanks, fixed all occurrences.

internal/log_analysis/alert_forwarder/forwarder/models.go
kostaspap: Nope. Updated

internal/log_analysis/alert_forwarder/forwarder/models.go
kostaspap: It's not meant to be shared - it represents the item stored in DDB.

internal/log_analysis/alert_forwarder/main/lambda.go
kostaspap: Yup.  Note that all the alerts are send to the same endpoint (in this case we are just storing them in DDB) .  Going for "at-least-once" approach.

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -0,0 +1,64 @@
+package forwarder
+
+import (
+	"os"
+	"strconv"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
```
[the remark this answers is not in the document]
kostaspap: Done

internal/log_analysis/alert_forwarder/main/lambda.go
kostaspap: Interesting. I see you are putting the Lambda ARN in there. What happens when we start running some of these stuff in e.g. Fargate? Will we have to adjust the metrics?

kostaspap: Added!

internal/log_analysis/alert_forwarder/main/lambda.go
austinbyers: Sounds good, maybe a comment to that effect somewhere so we remember

internal/log_analysis/alert_forwarder/forwarder/models.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// Alert contains all the fields associated to the alert stored in DDB
+type Alert struct {
+	ID            string `dynamodbav:"id,string"`
+	TimePartition string `dynamodbav:"timePartition,string"`
+	AlertDedupEvent
+}
+
+func FromDynamodDBAttribute(input map[string]events.DynamoDBAttributeValue) (*AlertDedupEvent, error) {
```
rleighton: I think you should check all the map lookups before you dereference the DynamoDBAttributeValue. We have had multiple issues with upstream writers not honoring the implicit contract and downstream code crashing hard.

rleighton: Unless that is impossible here ... looking at the code I'm unsure.

deployments/log_analysis/alerts.yml
kostaspap: You are right! Removed them and still everything works fine.

internal/log_analysis/alert_forwarder/main/lambda.go
kostaspap: Added.

internal/log_analysis/alert_forwarder/forwarder/models.go
kostaspap: It "shouldn't" but it "might". You are right, i'll add checks.

kostaspap: Added + tests.

PR #226, opened 2020-02-20
Add missing databases.yml

rleighton:
## Background

Add missing databases.yml that was missed in previous commit.

## Testing
mage test:ci
pushed to my dev account


deployments/databases.yml
```diff
@@ -0,0 +1,33 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+AWSTemplateFormatVersion: 2010-09-09
+Description: Template for Panther database related resources
+
+Parameters:
+  ProcessedDataBucket:
+    Type: String
+    Description: Bucket where log processing writes data
+    Default: ''
```
austinbyers: This can't be an empty value, can it? A default value means the parameter can be omitted and the stack will work correctly.

If it's required for the stack to deploy, let's not set an empty default value

rleighton: kk

rleighton: done

PR #233, opened 2020-02-20
Whitelist Decrypt as readonly

rleighton:
## Background
KMS Decrypt ops were logging warnings. This is a read only api and since we do not trust the readOnly attribute we must whitelist.

## Changes
- Add "Decrypt" to whitelist
- Add logging for when requestId is empty (was true in this case, helps debugging)

## Testing
mage test:ct
in my dev account


rleighton:
Nick is going to do this as part of another larger PR today.

PR #239, opened 2020-02-20
Cloud Security & Log Analysis - Source pages split

3nvi:
## Background

This PR compliments #218 by making sure to split the single source page into two separate pages, each with their own dedicated URL

## Changes

- Break down source page listing into two distinct ones (for `aws-scan` and `aws-s3`)
- Break down source page creation into two distinct ones (for `aws-scan` and `aws-s3`)
- Rename Wizard-related components
- Abstract common components
- Add new dedicated routes for each page

## Testing

- Locally

## Screenshots

Screen Shot 2020-02-20 at 8 28 56 PM
Screen Shot 2020-02-20 at 8 29 21 PM
Screen Shot 2020-02-20 at 8 29 26 PM


nhakmiller:
> Where will the `BulkUpload` functionality be after the split? Since it applies to both cloud security and log analysis

It applies to both, but on the rules/policy list pages not on the source page. Might be a good question on #218?

PR #241, opened 2020-02-20
Complete FOSSA Configuration

jacknagz:
## Background

An updated FOSSA configuration to include all project libraries

## Changes

- A newly generated FOSSA config (via `fossa init`) to cover Python, Node, and Golang

## Testing

- Validated by running `fossa analyze --output`, which does not send results to the API

### DIFF
--- a/.fossa.yml
+++ b/.fossa.yml
@@ -8,18 +8,153 @@ cli:
   project: git@github.com:panther-labs/panther.git
 analyze:
   modules:
-    - name: backend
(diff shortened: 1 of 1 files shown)


jacknagz:
This might generate new license issues, in which case I'll resolve

kostaspap:
@jacknagz  Does this mean that we can run FOSSA analysis from our laptops (before submitting a PR?)

PR #242, opened 2020-02-20
AWS Event Processor doesn't process duplicate S3/CWE data

### LINKED ISSUE #164: Missing Dead Letter Queue for panther-snapshot-queue
Missing standard infra dlq for panther-snapshot-queue

### LINKED ISSUE #212: Only initiate re-scans from aws event processor on CloudWatch Events or S3 Downloads, not both
### Description

When #186 merges,  we will be able to process both s3 downloads and also direct CloudTrail via SNS within the AWS event processor. If users have both these flows setup (for example, if they want the low latency of cloudwatch events and also log analysis), then every change will initiate two scans. One when it happens, and one ~15 minutes later when CloudTrail catches up to CWE.

the aws event processor should remember if it has seen CWE for a given account & region (keying off both in case they have high priority regions with CWE turned on), and only process S3 downloads for regions/accounts it is not receiving CWE for. It is ok if there is some slight overlap around lambda start times, as there is no correctness penalty for over-scanning. We are just trying to save lambda compute time and SQS/Dynamo write costs by reducing the amount of scans to the minimum needed.

### Acceptance Criteria

- Users can have CWE and Log Analysis CloudTrail setup in the same account, and not duplicate all scans
- The aws event processor can differentiate between different regions in the same account (so if CWE is setup in just two regions, those regions will use CWE and the others will use CloudTrail via s3)
- The aws event processor will prefer CWE over CloudTrail via S3 (due to the lower latency)

nhakmiller:
## Background

The main purpose of this PR is to update the aws event processor to not re-process data when we receive the same data via CWE and the log analysis pipeline.

This is accomplished by remembering a combination of accountID + region for every event that was processed from CloudTrail delivered via CloudWatch Events, and checking if this key exists whenever processing data from S3. If it does exist, we ignore the data in S3 because we already know we received it previously via CWE. The key is expired after 15 minutes. This closes #212.

In support of this caching work, we needed to do some pre-processing. In order to not duplicate effort, we now send the pre-processed `cloudTrailMetaData` struct all the way down to the classifiers. Now we should never ever ever have to call `gjson.Get('someField')` twice for the same `someField` in the course of processing one event.

In addition, I took some time to streamline some parts of the classifying flow that were offending our sensibilities :) Completely re-worked how ignored event handling is done, it is now solely a function of the pre-processor. This means that all events we are going to eventually drop based on their `eventName` are dropped immediately with the minimal amount of extraneous processing done. It also means that the entire definition for ignored events is contained in one place, which should make managing and updating it much more straightforward.

A dead letter queue was added to the snapshot poller to close #164.

Finally, some minor cleanup of the aws event processor was performed, most notably standardizing the format of the unit tests and re-doing some of the function/file names to be more inline with their current responsibilities.

## Changes

- Re-organized and renamed a few things in the aws event processor
- Standardized unit tests
- Added some meta data pre-processing to aws event processing (only for data that was being extracted multiple times)
  - Passed this pre-processed metadata down through the classifying flow
  - Moved all handling of ignored events to the pre-processor
- Added caching of accountID/regions that we received CWE from, and skipped S3 notifications for data from that same accountID/region.
- Fixed an issue introduced by my previous PR wherein WARN messages where continuously logged when log analysis was setup

## Testing

- Deployed in my dev account, setup both CWE in some regions and log analysis in all regions, and observed that the S3 notifications for log analysis were processed in regions without CWE and not dropped in regions with CWE
  - Tested s3 notifications are working
  - Tested raw CloudTrail via SNS is working
  - Tested CloudTrail via SNS is working
- Unit testing


internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -154,20 +175,38 @@ func processS3Download(object *sources.S3ObjectInfo, changes map[string]*resourc
...
+			return errors.WithMessage(err, "error extracting metadata from CloudTrail in s3")
 		}
+		if _, ok := cweAccounts[metadata.generateSourceKey()]; ok {
+			// If we're currently seeing CloudTrail via CWE, we don't process the duplicate data in S3
+			zap.L().Debug(
+				"skipping s3 notification in favor of CloudTrail via CWE",
+				zap.String("region", metadata.Region),
+				zap.String("accountID", metadata.AccountID),
+			)
+			continue
```
nhakmiller: @kostaspap @rleighton , I wasn't sure if this should be a `continue` or a `return`. For a single object in the processed data s3 bucket, can that object contain CloudTrail data from multiple accounts? Multiple regions? If the answer to either of those is yes I think we need to continue here.

deployments/compliance/aws_event_processor.yml
```diff
@@ -124,7 +124,7 @@ Resources:
     Properties:
       MessageRetentionPeriod: 1209600 # Max duration - 14 days
       QueueName: panther-aws-events-dlq
-      VisibilityTimeout: 60
+      VisibilityTimeout: !Ref TimeoutSec
```
rleighton: Why does a dlq need a visibility timeout?

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -67,7 +67,13 @@ func Handle(batch *events.SQSEvent) error {
 		detail := gjson.Get(record.Body, "detail")
 		if detail.Exists() {
 			zap.L().Debug("processing raw CloudTrail")
-			err := processCloudTrail(detail, changes)
+			metadata, err := preprocessCloudTrailLog(detail)
+			if err != nil {
+				zap.L().Error("error extracting metadata from raw CloudTrail", zap.Error(err))
```
rleighton: return/continue?

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -83,30 +120,36 @@ func classifyCloudTrailLog(detail gjson.Result) []*resourceChange {
...
 	}
 
 	// Process the body
-	changes := classifier(detail, accountID)
+	changes := classifier(detail, metadata.AccountID)
 	eventTime := detail.Get("eventTime").Str
+	if len(changes) > 0 {
+		readOnly := detail.Get("readonly").Bool()
+		if readOnly {
+			zap.L().Warn("processing changes from event marked readOnly")
```
nhakmiller: @rleighton this is the WARN message we discussed on slack for baselining how untrustworthy the `readOnly` field is.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -67,7 +67,13 @@ func Handle(batch *events.SQSEvent) error {
...
+			}
+			cweAccounts[metadata.generateSourceKey()] = struct{}{}
+
+			err = handleCloudTrail(detail, metadata, changes)
 			if err != nil {
 				zap.L().Error("error processing raw CloudTrail", zap.Error(errors.WithStack(err)))
```
rleighton: return/continue?

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -63,8 +65,43 @@ var classifiers = map[string]func(gjson.Result, string) []*resourceChange{
...
+}
+
+// generateSourceKey creates the key used for the cweAccounts cache for a given CloudTrail metadata struct
+func (metadata *CloudTrailMetaData) generateSourceKey() string {
+	return fmt.Sprintf("%s/%s", metadata.AccountID, metadata.Region)
+}
+
+// preprocessCloudTrailLog extracts some meta data that is used repeatedly for a CloudTrail log
+func preprocessCloudTrailLog(detail gjson.Result) (*CloudTrailMetaData, error) {
+	accountID := detail.Get("userIdentity.accountId")
```
nhakmiller: @rleighton this method of extracting the accountID should fix both the issues you encountered, both `recipientAccountId` not always being present and also actually the issue you were having with KMS I think.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -77,29 +83,38 @@ func Handle(batch *events.SQSEvent) error {
...
+		case "Notification":
 			// This case is checking for CloudTrail logs directly wrapped in SNS Events
 			zap.L().Debug("processing SNS notification")
 			message := gjson.Get(record.Body, "Message").Str
 			detail := gjson.Get(message, "detail")
-			err := processCloudTrail(detail, changes)
+
+			metadata, err := preprocessCloudTrailLog(detail)
+			if err != nil {
+				zap.L().Error("error extracting metadata from raw CloudTrail", zap.Error(err))
```
rleighton: return/continue?

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -77,29 +83,38 @@ func Handle(batch *events.SQSEvent) error {
...
+			}
+			cweAccounts[metadata.generateSourceKey()] = struct{}{}
+
+			err = handleCloudTrail(detail, metadata, changes)
 			if err != nil {
 				zap.L().Error("error processing SNS wrapped CloudTrail", zap.Error(errors.WithStack(err)))
```
rleighton: return/continue?

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -118,13 +133,15 @@ func Handle(batch *events.SQSEvent) error {
 	return submitChanges(changes)
 }
 
-func processCloudTrail(cloudtrail gjson.Result, changes map[string]*resourceChange) error {
+// handleCloudTrail takes a single CloudTrail log line and determines what scans if any need to be made as a result
+// of the log.
+func handleCloudTrail(cloudtrail gjson.Result, metadata *CloudTrailMetaData, changes map[string]*resourceChange) error {
 	if !cloudtrail.Exists() {
 		return errors.WithStack(errors.New("dropping bad event"))
```
rleighton: No need for WithStack() if errrors is pkg/errors

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -63,8 +65,43 @@ var classifiers = map[string]func(gjson.Result, string) []*resourceChange{
...
+// extracting the same information
+type CloudTrailMetaData struct {
+	Region    string
+	AccountID string
+	eventName string
+}
+
+// generateSourceKey creates the key used for the cweAccounts cache for a given CloudTrail metadata struct
+func (metadata *CloudTrailMetaData) generateSourceKey() string {
+	return fmt.Sprintf("%s/%s", metadata.AccountID, metadata.Region)
```
rleighton: Why not use '+' ? fmt.Sprintf() is very slow.

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -63,8 +65,43 @@ var classifiers = map[string]func(gjson.Result, string) []*resourceChange{
...
+}
+
+// preprocessCloudTrailLog extracts some meta data that is used repeatedly for a CloudTrail log
+func preprocessCloudTrailLog(detail gjson.Result) (*CloudTrailMetaData, error) {
+	accountID := detail.Get("userIdentity.accountId")
+	if !accountID.Exists() {
+		return nil, errors.WithStack(errors.New("unable to extract CloudTrail accountId field"))
```
rleighton: Don't need any of these WithStack() in this function

internal/compliance/aws_event_processor/processor/process.go
rleighton: Add the name of the event to log message

deployments/compliance/aws_event_processor.yml
nhakmiller: Not sure, this was just the pattern that the other dead letter queues in cloud security and log analysis follow.

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: Good catch 👍

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: The next line of code after this one is a `continue`

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: This switch statement is at the end of the for loop, no code is executed in between this statement and the next loop. Do you think there should be an explicit `continue` anyways just in case code is added after this switch statement at some point in the future?

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: Didn't know that, thanks for the heads up. Removed.

internal/compliance/aws_event_processor/processor/process.go
nhakmiller: Not sure, updated.

deployments/compliance/aws_event_processor.yml
austinbyers: This is only relevant if someone is pulling messages from the queue to look at them, doesn't particularly matter its value in the case of a DLQ. You can omit the value to use the default

For queues that feed Lambda functions, the visibility needs to be at least the max Lambda duration

nhakmiller: Happy to remove it, but just as a heads up all the dead letter queues in Panther that I looked at had this set.

internal/compliance/aws_event_processor/processor/accounts.go
```diff
@@ -38,9 +38,14 @@ const (
...
 
 var (
-	// Keyed on accountID
+	// Valid accounts, keyed on accountID
 	accounts            = make(map[string]*models.SourceIntegration)
 	accountsLastUpdated time.Time
+
+	// Accounts where CloudWatch Events are enabled, and should be preferred over S3 notifications
+	// Keyed by accountID + region
+	cweAccounts = make(map[string]struct{})
```
austinbyers: This map is internal to each Lambda container, so multiple event processors running at the same time may have slightly different maps here. Are we ok with that? It seems to be fine as best I can tell

nhakmiller: Correct, we are ok with that. We're not relying on actually processing the same exact CloudTrail log from S3 and CWE in the same lambda invocation. Just CloudTrail from the same account/region. And if the logs are so quite that we only have one or two events per account/region per lambda lifetime (~15 minutes), then we can afford the duplicate scanning.

internal/compliance/aws_event_processor/processor/handler.go
rleighton: Actually, since this is from the log processor and it could process 2 files and then mix the results, so yes.

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -0,0 +1,177 @@
...
+		"s3.amazonaws.com":                   classifyS3,
+		"waf.amazonaws.com":                  classifyWAF,
+		"waf-regional.amazonaws.com":         classifyWAFRegional,
+	}
+
+	// These events may have blank/empty userIdentity fields, so we cannot process them.
+	// Luckily, we can just skip them since they are all related to authentication and so won't be changing resources.
+	eventsWithoutAccountID = map[string]struct{}{
+		"SetUserMFAPreference":   {},
+		"GetUser":                {},
```
kostaspap: The events in lines 72-75 are readonly from our perspective, in the sense they don't modify a resource that would require rescanning, right? Where do we filter the events we care about (modifying state), is it lines 139-143? I would expect that read-only events are dropped quite in the beginning of the processing logic and here we don't have to deal with such events at all here. 
E.g. the filtering of "modifying" events could already happen in the preprocessing stage, before the check of line 101.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -63,43 +63,66 @@ func Handle(batch *events.SQSEvent) error {
 	//
 	// Since we don't want one bad notification to lose us the rest, we log failures and continue.
 	for _, record := range batch.Records {
-		// SNS raw message delivery of CloudTrail
+		// Check for SNS raw message delivery of CloudTrail
 		detail := gjson.Get(record.Body, "detail")
 		if detail.Exists() {
```
kostaspap: Personally I would put lines 69-83 in a method called `handleCloudWatchEvent` and lines 88-99 to a method called `handleCloudTrailLog` or similar. Maybe create a small function `isCloudWatchEvent` that returns boolean. Code could then look like:
```
if isCloudWatchEvent(record) {
  handleCloudWatchEvent(record)
} else {
 handleCloudTrailLog(record)
}
```
The `handleCloudTrailLog` logic could be merged with the logic inside the existing `handleCloudTrail` method (line 146)
Then you don't even need necessarily the comments. 
Anyway, an idea I had as I was reading the code, take it or leave it.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -63,43 +63,66 @@ func Handle(batch *events.SQSEvent) error {
...
-			err := processCloudTrail(detail, changes)
+			metadata, err := preprocessCloudTrailLog(detail)
+			if err != nil {
+				zap.L().Error("error extracting metadata from raw CloudTrail", zap.Error(err))
+				continue
+			}
+			if metadata == nil {
+				continue
+			}
+			cweAccounts[metadata.generateSourceKey()] = struct{}{}
```
kostaspap: The `generateSourceKey` looks like it could better be a method of this file (`func generateSourceKey(metadata.*CloudTrailMetadata`) rather than a method of the Metadata (given that the `SourceKey` is more related to the `Handle` method context than to the actual Metadata. )

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -63,43 +63,66 @@ func Handle(batch *events.SQSEvent) error {
...
+
+			err = handleCloudTrail(detail, metadata, changes)
 			if err != nil {
-				zap.L().Error("error processing SNS wrapped CloudTrail", zap.Error(errors.WithStack(err)))
+				zap.L().Error("error processing SNS wrapped CloudTrail", zap.Error(err))
 			}
```
kostaspap: 1. Can this ever happen? We provide the customers with the CFN templates, the templates should either deliver the raw message or not. 
2. If you think this is needed, then If I am not mistaken this repeats the logic from lines 69-84. I made a comment above about refactoring the logic slightly, you could then just re-use it here and call `handleCloudWatchEvent` method

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -0,0 +1,177 @@
...
+	eventName string
+}
+
+// generateSourceKey creates the key used for the cweAccounts cache for a given CloudTrail metadata struct
+func (metadata *CloudTrailMetaData) generateSourceKey() string {
+	return metadata.AccountID + "/" + metadata.Region
+}
+
+// preprocessCloudTrailLog extracts some meta data that is used repeatedly for a CloudTrail log
+func preprocessCloudTrailLog(detail gjson.Result) (*CloudTrailMetaData, error) {
```
kostaspap: 1. Could you add to the description what's the conceptual difference between returning nil metadata and returning an error? 
2. nit: Should be `CloudTrailMetadata` (metadata is an actual noun)

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -0,0 +1,177 @@
...
+	eventName := detail.Get("eventName")
+	if !eventName.Exists() {
+		return nil, errors.New("unable to extract CloudTrail eventName field")
+	}
+	accountID := detail.Get("userIdentity.accountId")
```
kostaspap: Would it make sense to check the `sessionIssuer.accountId` field as fallback? https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-user-identity.html

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -0,0 +1,177 @@
...
+		strings.HasPrefix(metadata.eventName, "List") {
+
+		zap.L().Debug(source+": ignoring read-only event", zap.String("eventName", metadata.eventName))
+		return nil
+	}
+
+	// Check if this log is from a supported account
+	integration, ok := accounts[metadata.AccountID]
+	if !ok {
+		zap.L().Warn("dropping event from unauthorized account",
```
kostaspap: Maybe 'unauthorized' is not valid here? Maybe `dropping event from account that hasn't been onboarded` or similar.

internal/compliance/aws_event_processor/processor/accounts.go
kostaspap: This map will be here for the lifecycle of the container in which Lambda is running (several hours). It's not getting re-initialized on every invocation. So this map will keep getting populated for ... long. I am wondering if we should put a time limit on this. Given that CT from S3 has avg delay of 15' and might be some other small delays here and there, we could e.g. keep the cache for double that time. So e.g. cleanup the cache every 30' or sth like that.

internal/compliance/aws_event_processor/processor/handler.go
rleighton: Yes, if there is a large visual "distance", you could get bugs later. I would treat this like naked returns (fine for short bits of code, but dangerous over long distances).

internal/compliance/aws_event_processor/processor/accounts.go
rleighton: Excellent point. We should definitely make some expiration. Simplest is what Kostas suggested. Another approach is switch to an LRU cache (there are many opens source packages) and set the size at something reasonable.

deployments/compliance/snapshot.yml
```diff
@@ -100,7 +100,18 @@ Resources:
     Properties:
       QueueName: panther-snapshot-queue
       KmsMasterKeyId: !Ref SQSKeyId
+      # Reference on KeyReuse: https://amzn.to/2ngIsFB
+      KmsDataKeyReusePeriodSeconds: 3600 # 1 hour
```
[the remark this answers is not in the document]
nhakmiller: I copied this configuration from the `deployments/log_analysis/log_processor.yml` configuration for the panther-input-data-notifications queue. It seems like a reasonable practice in general.

deployments/compliance/aws_event_processor.yml
nhakmiller: Removed.

internal/compliance/aws_event_processor/processor/accounts.go
nhakmiller: Not sure I understand the desire to cleanup the cache, wouldn't a longer cache lifetime be better? Are you concerned with the case where users have setup CWE for an account and then later disable it?

internal/compliance/aws_event_processor/processor/process.go
nhakmiller: Unfortunately this is what we have to work with in these situations:

```
    "userIdentity": {
        "type": "Unknown",
        "principalId": "",
        "arn": "",
        "accessKeyId": "",
        "userName": ""
    },
```

So `sessionContext.sessionIssuer` isn't event present. These events literally have no indication anywhere of an `accountId` field.

internal/compliance/aws_event_processor/processor/accounts.go
kostaspap: Yup. If someone decides to make the switch, it could take up to 8hr to do it.

internal/compliance/aws_event_processor/processor/process.go
nhakmiller: From my perspective, an account that has not been onboarded is not authorized (i.e. granted permission) to send logs so I think the warning makes sense. Because it could be an account that was onboarded and then offboarded but the infra hasn't been cleaned up, or an account that has no intention of being onboarded but somehow is sending logs anyways (perhaps maliciously). `dropping event from account that hasn't been onboarded` makes it sound like users should just onboard the account to make the warning go away, but really they should be taking a good hard look at that source account and determine if it should even be sending logs in the first place.

rleighton: Given that this could be a major mis-config, do you want log level error for this?

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: 1. There is nothing stopping a user from setting these topics up by hand, via Terraform/some other cloud infra as code tool, or any number of other things. Especially if someone is doing some more advanced setups where they don't necessarily plan to use our templates because they are only enabling CWE for a few regions for example, or if they want to re-use an existing SNS topic and so they are creating the subscriptions by hand or porting over the subscription CFN etc. Seems like a low overhead to me to support this case.
2. It is repeated code from 69-84, I will factor these 11 lines of duplicated code out into a helper function

internal/compliance/aws_event_processor/processor/process.go
kostaspap: If we have reached this step, it means that we have already confirmed the SNS->SQS subscription, so the user had added the account at some point. So the concern is what happens after a user have off-boarded the account but we still keep receiving messages from it, right? Isn't this more the case of a previously onboarded account still sending messages? Couldn't the message just indicate that? Would help the people debug the issue easier, right?

internal/compliance/aws_event_processor/processor/process.go
nhakmiller: To answer your first question directly, I believe `SetUserMFAPreference` should not require a re-scan because it is a part of the authentication negotiation, sort of like setting a preferred cipher suite during a TLS negotiation. I could be mistaken, but there is zero information in the event so we can't really use it for anything anyways.

To answer your question about checking for read only events, I was actually just having a conversation with @austinbyers about this yesterday.

With this change we are checking for read only events in three places:
 1. Here, where we check for events that don't specify an `accountId` (which also happen to all be read only events) and drop them
2. In lines 134-143 as you point out, which happens very early in the processing pipeline but crucially AFTER the preprocessing occurs, which is what check number 1 is protecting since we need the accountId to generate the key in between preprocessing and processing
3. In each individual service classifier, where we put in more in depth checks for various read only events that are service specific

The general idea was that we try to be efficient by not doing 200+ sequential string comparison and `strings.HasPrefix(...)` checks at the beginning of processing every event, because since we can determine the service fairly easily we can narrow that down to like 10-20 checks and save some time. So the checks occur:

1. When they MUST occur, because we will fail without them (the first check)
2. When they are likely to catch an extremely large number of read only events and save processing time (the second group of checks)
3. When we have enough additional context to quickly determine what they are likely to be (the third group of checks)

I think that this methodology has perhaps outgrown its usefulness, and it is certainly very confusing. As you say, if the very first thing we do is efficiently determine whether or not this is a read only event, we can save all other processing. And now that we have a this pre-processing function, that seem like an ideal place to do so.

I proposed a single giant map of all read only events and prefixes, and we just check if the event is in that map and return nil here if it is. That makes it just one hash + lookup, and then we can immediately throw away the event if it's read only. What do you think?

deployments/compliance/snapshot.yml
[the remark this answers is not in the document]
austinbyers: SQS already has a default key re-use period of 5 minutes, I don't think setting a longer value here provides any value. This is the period of time between SQS refreshing it's data key (`kms:GenerateDataKey`) to avoid slamming KMS with calls for every message in the queue.

My preference would be to omit this field everywhere unless we have a very good reason we need to keep the encryption keys around for an entire hour

internal/compliance/aws_event_processor/processor/accounts.go
nhakmiller: Ah ok, I guess I didn't realize it would live for up to 8 hours. Is this documented anywhere? Or just hard earned personal experience 😄 ? I will update this with a 15 minute lifetime per cached event. If we are getting exactly one event per fifteen minutes and so the cache expires before CloudTrail -> S3 gets there, I think we can afford the duplicate scanning.

austinbyers: As far as I know, there is no guarantee about how long a Lambda container can live. In practice, most probably don't live more than a day, but code should assume indefinite container retention. The Lambda service can recycle containers or create new ones at any time

nhakmiller: Updated.

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: Sounds good, updated.

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: Updated.

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: Ok, I took a crack at this. Let me know what you think.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -163,25 +169,51 @@ func processS3Download(object *sources.S3ObjectInfo, changes map[string]*resourc
...
-		// Since we don't wont to lose an entire batch of logs to one bad message, we just log failures and continue
-		err = processCloudTrail(gjson.Parse(line), changes)
+
+		// Parse the line and determine if we should continue
+		detail := gjson.Parse(line)
+		var metadata *CloudTrailMetadata
+		metadata, err = preprocessCloudTrailLog(detail)
 		if err != nil {
-			zap.L().Error("error processing CloudTrail from S3", zap.Error(errors.WithStack(err)))
+			return errors.WithMessage(err, "error extracting metadata from CloudTrail in s3")
```
rleighton: Why not Wrap()? You do not want the stack here?

internal/compliance/aws_event_processor/processor/process.go
rleighton: Makes sense.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -127,28 +134,27 @@ func Handle(lc *lambdacontext.LambdaContext, batch *events.SQSEvent) (err error)
...
-func processCloudTrail(cloudtrail gjson.Result, changes map[string]*resourceChange) error {
-	if !cloudtrail.Exists() {
-		return errors.WithStack(errors.New("dropping bad event"))
+// handleCloudTrail takes a single CloudTrail log line and determines what scans if any need to be made as a result
+// of the log.
+func handleCloudTrail(cloudtrail gjson.Result, changes map[string]*resourceChange) error {
+	metadata, err := preprocessCloudTrailLog(cloudtrail)
+	if err != nil {
+		zap.L().Error("error extracting metadata from CloudTrail", zap.Error(err))
+		return err
```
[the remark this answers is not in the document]
nhakmiller: You're right. Sorry I keep switching up which parts are logging and which parts are returning errors and confusing myself :(

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: Since `preprocecssCloudTrailLog` is a function we define, it is already returning an `errors.New()` or `errors.Wrap()` error so I think it should already have the stack trace from that right?

rleighton: Makes sense. Unless that message is important maybe just return err.

internal/compliance/aws_event_processor/processor/process.go
rleighton: Are you planning to do this in this PR or a subsequent PR?

nhakmiller:
@rleighton @kostaspap this PR has changed fairly substantially since I initially opened it. I have updated the PR description to reflect this, please take a look at your convenience and let me know if this looks good now.

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -0,0 +1,362 @@
...
+	// check to save developer time from having to maintain an even more massive list
+	ignoredPrefixes = []string{
+		// general
+		"Get",
+		"Describe",
+		"List",
+	}
+)
+
+// CloudTrailMetaData is a data struct that contains re-used fields of CloudTrail logs so that we don't have to keep
```
rleighton: Big improvement!

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -0,0 +1,362 @@
...
+	ResourceID    string `json:"resourceId"`    // e.g. "arn:aws:s3:::my-bucket"
+	ResourceType  string `json:"resourceType"`  // e.g. "AWS.S3.Bucket"
+}
+
+// Map each event source to the appropriate classifier function.
+//
+// The "classifier" takes a cloudtrail log and summarizes the required change.
+// integrationID does not need to be set by the individual classifiers.
+var (
+	classifiers = map[string]func(gjson.Result, *CloudTrailMetadata) []*resourceChange{
```
rleighton: I like this consolidation.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -70,43 +71,50 @@ func Handle(lc *lambdacontext.LambdaContext, batch *events.SQSEvent) (err error)
 	//
 	// Since we don't want one bad notification to lose us the rest, we log failures and continue.
 	for _, record := range batch.Records {
-		// SNS raw message delivery of CloudTrail
+		// Check for SNS raw message delivery of CloudTrail
 		detail := gjson.Get(record.Body, "detail")
 		if detail.Exists() {
 			zap.L().Debug("processing raw CloudTrail")
-			err := processCloudTrail(detail, changes)
+			err := handleCloudTrail(detail, changes)
```
kostaspap: Here we are passing state (the 'changes' map) in each method call, and each time the `handleCloudTrail` or `handleS3Download` is invoked they get the state as argument and modify that. Going "stateless" in these methods could actually make the code slightly easier to read and easier to reasonate about. 

You could change the `handleCloudTrail` to have the following signature: 
```
handleCloudTrail(*gjson.Result) ([]*resourceChange, error)
```
Same for `handleS3Download`. 
Each of this methods now accept an "event" (cloudtrail event or batch of events) and return a list of resource changes. You can just collect the list of resource changes. 

Your "submitChanges" method can just get the list and do the deduplication in there. 

Let me know if you think it makes sense, I think should be a small change.

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -0,0 +1,362 @@
...
+		// Prevents the following from being de-duped mistakenly:
+		//
+		// Resources with the same ID in different regions (different regions)
+		// Service scans in the same region (different resource types)
+		// Resources with the same type in the same region (different resource IDs)
+		key := change.ResourceID + change.ResourceType + change.Region
+		if entry, ok := changes[key]; !ok || change.EventTime > entry.EventTime {
+			changes[key] = change // the newest event for this resource we've seen so far
+		}
+	}
```
kostaspap: See comment above. Ideally this logic should be present in the same file/class that the "changes" object is declared.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -127,28 +135,26 @@ func Handle(lc *lambdacontext.LambdaContext, batch *events.SQSEvent) (err error)
 	return err
 }
 
-func processCloudTrail(cloudtrail gjson.Result, changes map[string]*resourceChange) error {
-	if !cloudtrail.Exists() {
-		return errors.WithStack(errors.New("dropping bad event"))
+// handleCloudTrail takes a single CloudTrail log line and determines what scans if any need to be made as a result
+// of the log.
+func handleCloudTrail(cloudtrail gjson.Result, changes map[string]*resourceChange) error {
+	metadata, err := preprocessCloudTrailLog(cloudtrail)
```
kostaspap: In this type of error you actually don't want to retry (there is something wrong with the event and it is not retryable). 
You could actually modify the `preprocessCloudTrailLog` method to just return a value or nil. If `nil` is returned, just skip it. 
The method could log the appropriate errors/warnings.

nhakmiller: In discussing with @rleighton the other day, we sort of settled on a pattern of always returning errors as high up as possible until you decide to accept them, and then logging them. Since all errors in this function are errors we don't want to retry, I let the main `Handle` function call this one, then log any error it returns and continue.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -70,43 +71,50 @@ func Handle(lc *lambdacontext.LambdaContext, batch *events.SQSEvent) (err error)
...
 			message := gjson.Get(record.Body, "Message").Str
 			detail := gjson.Get(message, "detail")
-			err := processCloudTrail(detail, changes)
+			if !detail.Exists() {
+				zap.L().Error("error extracting detail from SNS wrapped CloudTrail")
+				continue
+			}
+			err := handleCloudTrail(detail, changes)
 			if err != nil {
 				operation.LogError(errors.Wrap(err, "error processing SNS wrapped CloudTrail"))
```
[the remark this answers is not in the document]
nhakmiller: Any error returned by `processCloudTrail` is an error that we don't want to retry, so we just log it and continue. That is because `processCloudTrail` is deterministic in the sense that it is only doing string processing.

PR #244, opened 2020-02-20
Update lambda cf to remove mem and timeout params

rleighton:
## Background
We will be generating alarms from CF using CW metric filters to alarm when lambdas reach 90% of their memory capacity for a sustained period. Since the yaml parser does not understand !Ref from CF, we are moving to hardcoded values for memory so we can compute the % (and timeout because Austin asked for that).

## Testing
mage test:ci
deployed to dev account, tested memory alarms by setting threshold very low.


deployments/compliance/alert_processing.yml
```diff
@@ -88,7 +63,7 @@ Resources:
       QueueName: panther-alert-processor-queue
       MessageRetentionPeriod: 1209600 # Max duration - 14 days
       KmsMasterKeyId: !Ref SQSKeyId
-      VisibilityTimeout: !Ref ProcessorTimeoutSec
+      VisibilityTimeout: 60 # Should match lambda timeout
```
austinbyers: Ah, I forgot there are times we use the timeout in more than one place.

In these cases, if you want, you can put the timeout in a mapping so it can be referenced in both places. But leaving it as-is also works, since that would allow us to generate alarms if the max duration is approaching its limit as well

rleighton: I'll leave it for now. Easy to change later.

PR #245, opened 2020-02-21
Fix athena query for queued responses

rleighton:
## Before
Athena queries would fail if the execution response was "QUEUED".

## After
Code will continue to wait for a terminal status.

## Testing
mage test:ci
mage deploy (which uses the code to create views)


pkg/awsathena/query.go
```diff
@@ -64,26 +64,15 @@ func (aq *AthenaQuery) Run() (err error) {
 
 	aq.startResult, err = aq.Client.StartQueryExecution(&startInput)
 	if err != nil {
-		err = errors.Wrapf(err, "athena failed running: %#v", *aq)
+		err = errors.Wrapf(err, "athena failed starting: %#v", *aq)
```
kostaspap: nit: "athena failed to start" or "failed to start athena query" might read better?

pkg/awsathena/query.go
```diff
@@ -92,11 +81,31 @@ func (aq *AthenaQuery) Wait() (err error) {
 
 		aq.QueryResult, err = aq.Client.GetQueryResults(&ip)
 		if err != nil {
-			return errors.Wrapf(err, "athena failed running: %#v", *aq)
+			return errors.Wrapf(err, "athena failed getting results: %#v", *aq)
```
kostaspap: nit: "failed to get results from athena"?

pkg/awsathena/query.go
```diff
@@ -92,11 +81,31 @@ func (aq *AthenaQuery) Wait() (err error) {
...
 		}
 	} else {
 		return errors.Errorf("athena failed with status %s running: %#v", *executionOutput.QueryExecution.Status.State, *aq)
 	}
 
 	return nil
 }
+
+func (aq *AthenaQuery) poll() (executionOutput *athena.GetQueryExecutionOutput, err error) {
```
kostaspap: 1. We need a timeout for this operation. Maybe something that can be passed in through the `AthenaQuery` struct (so that it's configurable). 
2. In case of timeout, it would be good to do a best effort attempt to cancel the query.

pkg/awsathena/query.go
rleighton: Sure.

pkg/awsathena/query.go
rleighton: Sure

pkg/awsathena/query.go
rleighton: Athena has built-in time out (from memory, 30min default) and you get a canceled message.  Later (likely when I do the CTAS work next week) we can elaborate on this if needed , such has passing in user specified time outs (which have to be less than the global to work).

kostaspap: That means a max of 30' waiting time during deployment... But not sure if there is really a "right" time there. Alright, I think it's fine to just leave it for now.

PR #253, opened 2020-02-22
users-api: prevent deleting last user and move pagination to backend

### LINKED ISSUE #217: Admins can delete themselves
### Describe the bug

As the only admin user on the Panther UI, I have the option to delete myself which may lead to unexpected behaviours and a hard recovery process for the account.

### Steps to reproduce

Steps to reproduce the behaviour:

1. Deploy Panther
2. Log in as an admin
3. Delete all other users
4. Delete your user

### Expected behavior

This is pretty open ended imo. Some solutions that we can implement:
1. Only admin cant delete himself either on frontend and/or backend
2. Admins in general can't delete themselves.
3. Last admin must promote someone else as admin before deleting
4. If we allow only admin to be removed, shouldn't a cleanup process triggered?

### Environment

- Panther version or commit: v0.3
- OS: MacOS 10.15.3
- Browser: All

austinbyers:
## Background
Some small changes to the `users-api`, following up from #213 

Closes #217 

## Changes

- The `users-api` service will refuse a user deletion request if that user is the last one remaining
- The `limit` and `paging` info in the `ListUsers` endpoint has been removed
    - There will almost certainly only ever be 1 page of users, the frontend does not and will not add paging here. In the rare event we do need to page over the results, the backend handles this before returning

## Testing

- Deployed to dev account, was able to list users but not able to delete myself (because I was the only user)

Screen Shot 2020-02-21 at 5 17 05 PM

(we may want a more helpful error message)

### DIFF
--- a/internal/core/users_api/api/remove_user.go
+++ b/internal/core/users_api/api/remove_user.go
@@ -33,7 +33,7 @@ func (API) RemoveUser(input *models.RemoveUserInput) error {
 		return err
 	}
 
-	if len(users) == 0 {
+	if len(users) == 1 {
 		return &genericapi.InUseError{Message: "can't delete the last user"}
 	}
 
--- a/internal/core/users_api/api/remove_user.go
+++ b/internal/core/users_api/api/remove_user.go
@@ -19,7 +19,6 @@ package api
  */
 
 import (
-	"github.com/aws/aws-sdk-go/aws"
 	"go.uber.org/zap"
 
 	"github.com/panther-labs/panther/api/lambda/users/models"
@@ -28,13 +27,13 @@ import (
 
 // RemoveUser deletes a user from cognito.
 func (API) RemoveUser(input *models.RemoveUserInput) error {
-	list, err := userGateway.ListUsers(aws.Int64(1), nil)
+	users, err := userGateway.ListUsers()
 	if err != nil {
 		zap.L().Error("error listing users", zap.Error(err))
 		return err
 	}
 
-	if len(list.Users) == 0 {
+	if len(users) == 0 {
 		return &genericapi.InUseError{Message: "can't delete the last user"}
 	}
 
--- a/internal/core/users_api/api/remove_user.go
+++ b/internal/core/users_api/api/remove_user.go
@@ -19,13 +19,25 @@ package api
  */
 
 import (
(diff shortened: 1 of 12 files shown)


internal/core/users_api/api/remove_user.go
```diff
@@ -22,10 +22,21 @@ import (
...
 
 // RemoveUser deletes a user from cognito.
 func (API) RemoveUser(input *models.RemoveUserInput) error {
+	users, err := userGateway.ListUsers()
+	if err != nil {
+		zap.L().Error("error listing users", zap.Error(err))
+		return err
+	}
+
+	if len(users) == 1 {
```
3nvi: In OSS that should suffice. In Enterprise, the check should test whether there is at least one `Admin` not just "one user"

internal/core/users_api/api/remove_user.go
```diff
@@ -22,10 +22,21 @@ import (
...
 	"github.com/panther-labs/panther/api/lambda/users/models"
+	"github.com/panther-labs/panther/pkg/genericapi"
 )
 
 // RemoveUser deletes a user from cognito.
 func (API) RemoveUser(input *models.RemoveUserInput) error {
+	users, err := userGateway.ListUsers()
+	if err != nil {
+		zap.L().Error("error listing users", zap.Error(err))
+		return err
```
rleighton: Can we also do:
return errors.Wrap(err, "error listing users")

 ... also surrounding code.

internal/core/users_api/api/remove_user.go
austinbyers: Correct, this check (and this entire service) would be different in enterprise

internal/core/users_api/api/remove_user.go
austinbyers: This is not currently possible with the `genericapi` (Lambda-style) APIs, because we actually rely on structs for error types. The error struct is marshalled when returned to the caller, and wrapping the error breaks the json marshaling: https://play.golang.org/p/QIAJyEvI1eH

For example, we have structs for `DoesNotExistError` and `AlreadyExistsError`. The `genericapi` handler injects the route into the error struct before returning: https://github.com/panther-labs/panther/blob/master/pkg/genericapi/router.go#L145-L157

Having consistent error handling would be part of the effort to unify our APIs

PR #254, opened 2020-02-23
Add rule match Athena/Glue tables

rleighton:
## Background
The new rules engine outputs matches to s3. This change generates glue tables so this data can be queried in Athena.

## Changes
* added Clone() option to glue metadata tables since rule matches have a parallel structure to the log data
* change the name of the log database from panther_tables to panther_logs
* added database panther_rule_matches
* updated mage glue:sync

## Testing
mage test:ci
ran in my dev account

image


pkg/awsglue/glue.go
```diff
@@ -186,20 +198,23 @@ func NewGlueMetadata(databaseName, tableName, description string, timebin GlueTa
...
 	}
 
 	return
 }
+
+func standardizeS3Prefix(s3Prefix, tableName string) string {
+	s3Prefix += "/" + tableName
+	if s3Prefix[len(s3Prefix)-1] != '/' { // ensure last char is '/'
+		s3Prefix += "/"
+	}
```
kostaspap: The s3Prefix is constructed in line 215 and will not have a `/` (table names don't have `/`, if they had we'd be in bigger trouble. You can remove this condition.

tools/athenaviews/athenaviews.go
```diff
@@ -58,21 +58,40 @@ func CreateOrReplaceViews(athenaResultsBucket string) (err error) {
...
+// generateViewAllLogs creates a view over all log sources in log db using "panther" fields
 func generateViewAllLogs(tables []*awsglue.GlueMetadata) (sql string, err error) {
+	return generateViewAllHelper("all_logs", tables, []gluecf.Column{})
+}
+
+// generateViewAllRuleMatches creates a view over all log sources in rule match db the using "panther" fields
+func generateViewAllRuleMatches(tables []*awsglue.GlueMetadata) (sql string, err error) {
+	// the rule match tables share the same structure as the logs with some extra columns
+	var ruleTables []*awsglue.GlueMetadata
+	for _, t := range tables {
```
kostaspap: nit: t-> table. We tend to avoid single letter variable names

tools/athenaviews/athenaviews.go
```diff
@@ -58,21 +58,40 @@ func CreateOrReplaceViews(athenaResultsBucket string) (err error) {
...
+		ruleTables = append(ruleTables, t.Clone(awsglue.RuleMatchS3Prefix, awsglue.RuleMatchDatabaseName))
+	}
+	return generateViewAllHelper("all_rule_matches", ruleTables, gluecf.RuleMatchColumns)
+}
+
+func generateViewAllHelper(viewName string, tables []*awsglue.GlueMetadata, extraColumns []gluecf.Column) (sql string, err error) {
 	if len(tables) == 0 {
-		return "", errors.New("no tables specified for generateViewAllLogs()")
+		return "", errors.New("no tables specified for generateViewAllHelper()")
```
kostaspap: Our private method shouldn't be calling another private method of ours with invalid arguments. Argument check should be performed at the entry point of the public method, so better to move this check to right above line 43.

tools/cfngen/gluecf/gluecf.go
```diff
@@ -68,44 +92,54 @@ func GenerateTables(tables []*awsglue.GlueMetadata) (cf []byte, err error) {
...
 		},
 		"PantherViewsDatabase": &cfngen.Output{
 			Description: awsglue.ViewsDatabaseDescription,
 			Value:       cfngen.Ref{Ref: cfngen.SanitizeResourceName(awsglue.ViewsDatabaseName)},
 		},
 	}
 
-	// add tables for all parsers
-	for _, t := range tables {
+	addTable := func(t *awsglue.GlueMetadata, extraColumns ...Column) {
```
kostaspap: nit: We tend to avoid declaring nested functions like this to keep a consistent code style. Is this needed here? Maybe better move out to own function.

tools/cfngen/gluecf/gluecf_test.go
```diff
@@ -40,20 +40,22 @@ type commonFields struct {
...
 	require.NoError(t, err)
 	tables := []*awsglue.GlueMetadata{table}
 
 	cf, err := GenerateTables(tables)
 	require.NoError(t, err)
 
-	// un-comment to see output
-	// os.Stdout.Write(cf)
+	// uncomment to make a new expected file
+	// writeTestFile(cf, expectedFile)
```
kostaspap: Not sure, how is this meant to be used?

tools/mage/glue_namespace.go
```diff
@@ -79,12 +79,7 @@ func syncPartitions(glueClient *glue.Glue, matchTableName *regexp.Regexp) {
 		}()
 	}
 
-	// for each table, for each time partition, delete and re-create
-	for _, table := range registry.AvailableTables() {
-		name := fmt.Sprintf("%s.%s", table.DatabaseName(), table.TableName())
-		if !matchTableName.MatchString(name) {
-			continue
-		}
+	listPartitions := func(name string, table *awsglue.GlueMetadata) {
```
kostaspap: Same comment as above, we should avoid declaring functions like this if possible.

pkg/awsglue/glue.go
rleighton: Ah, was left over. Fixed.

tools/athenaviews/athenaviews.go
rleighton: Fixed.

tools/athenaviews/athenaviews.go
rleighton: Done

tools/cfngen/gluecf/gluecf_test.go
rleighton: When we change the schema we need to change the JSON which is a PITA to do by hand. Easier to uncomment, inspect, comment. That way we are protected from regression but can update quickly.

tools/mage/glue_namespace.go
rleighton: Why, what is the downside? In both of the cases, the only caller is below and it is done for clarity. Moving moving it outside makes you scan down and lose context. In this case it makes it very clear there are no other entry points.

PR #261, opened 2020-02-25
Syslog Parser for RFC5424 formetted logs

stoggi:
## Background

This PR introduces a new Parser for Syslog RFC5424 formatted logs as planned in #200.

## Changes

New Parser:
* Created new parser `Syslog.RFC5424` utilising https://github.com/influxdata/go-syslog
* Added `Syslog.RFC5424` to frontend available log types.
* Added `sysloglogs.RFC5424` to the parser registry.

New Common Parser Field:
* Added a new global panther boolean `PantherEventTimeWhenParsed` which should be set to true if `PantherEventTime` was set to `time.Now().UTC()` during parsing.
* Added `p_event_time_when_parsed` to the athena generated view tests.

Support mapping type map[string]map[string]string in Glue:
* Allow `inferMap` to be recursively called when the key type of a Map is also a Map.
* Added test cases for MapStringToMap

## Testing

* Unit tests for sysloglogs.RFC5424.
* Unit tests for new Glue mapping.
* Deployed into test account, created rule and verified parsed logs are alerted.


internal/log_analysis/log_processor/parsers/patherlog.go
```diff
@@ -45,10 +45,11 @@ type PantherLog struct {
...
-	PantherAnyMD5Hashes   *PantherAnyString `json:"p_any_md5_hashes,omitempty" description:"Panther added field with collection of MD5 hashes associated with the row"`
+	PantherAnyIPAddresses      *PantherAnyString `json:"p_any_ip_addresses,omitempty" description:"Panther added field with collection of ip addresses associated with the row"`
+	PantherAnyDomainNames      *PantherAnyString `json:"p_any_ip_domain_names,omitempty" description:"Panther added field with collection of domain names associated with the row"`
+	PantherAnySHA1Hashes       *PantherAnyString `json:"p_any_sha1_hashes,omitempty" description:"Panther added field with collection of SHA1 hashes associated with the row"`
+	PantherAnyMD5Hashes        *PantherAnyString `json:"p_any_md5_hashes,omitempty" description:"Panther added field with collection of MD5 hashes associated with the row"`
+	PantherEventTimeWhenParsed *bool             `json:"p_event_time_when_parsed,omitempty" description:"Panther could not determine the event time from the log, so set to the current time when parsed"`
```
stoggi: Example event being triggered in in Panther without a timestamp:
```
{
    "appname": "e",
    "facility": 20,
    "hostname": "mymach.it",
    "message": "An application event log entry...",
    "msgid": "1",
    "p_event_time": "2020-02-25 03:36:54.405312484",
    "p_event_time_when_parsed": true,
    "p_log_type": "Syslog.RFC5424",
    "p_row_id": "caf9ad4cc9d1b1948e87d78a0102",
    "priority": 165,
    "severity": 5,
    "structured_data": {
        "ex@32473": {
            "iut": "3"
        }
    },
    "version": 4
}
```

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
```diff
@@ -0,0 +1,112 @@
...
+		zap.L().Debug("failed to parse log", zap.Error(err))
+		return nil
+	}
+	internalRFC5424 := msg.(*rfc5424.SyslogMessage)
+
+	externalRFC5424 := &RFC5424{
+		Priority:       aws.Int16(int16(*internalRFC5424.Priority)),
+		Facility:       aws.Int16(int16(*internalRFC5424.Facility)),
+		Severity:       aws.Int16(int16(*internalRFC5424.Severity)),
+		Version:        aws.Int16(int16(internalRFC5424.Version)),
```
stoggi: These had to be `int16` because as far as I can tell Glue and Athena don't support unsigned ints.

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424_test.go
```diff
@@ -0,0 +1,241 @@
...
+	// rowid changes each time
+	require.Greater(t, len(*event.PantherRowID), 0) // ensure something is there.
+	expectedEvent.PantherRowID = event.PantherRowID
+
+	if expectedEvent.PantherEventTimeWhenParsed != nil && *expectedEvent.PantherEventTimeWhenParsed == true {
+		// PantherEventTime will be set to time.Now().UTC(), require it to be within one second of expectedTime
+		delta := (*time.Time)(event.PantherEventTime).Sub(*(*time.Time)(expectedEvent.PantherEventTime)).Nanoseconds()
+		require.Less(t, delta, 1*time.Second.Nanoseconds())
+		require.Greater(t, delta, -1*time.Second.Nanoseconds())
+		expectedEvent.PantherEventTime = event.PantherEventTime
```
stoggi: I'm a bit mindful of including a test that depends on the current time. I want to verify that the event was added at `time.Now(). UTC ()` is there a better way? We could dependency inject a time function into the parser?

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (event *RFC5424) updatePantherFields(p *RFC5424Parser) {
+	if event.Timestamp != nil {
+		event.SetCoreFieldsPtr(p.LogType(), event.Timestamp)
+	} else {
+		// A null timestamp is valid in RFC5424 (https://tools.ietf.org/html/rfc5424#section-6.2.3).
+		// Record the current time instead, and set PantherEventTimeWhenParsed to indicate the reconstructed time.
+		event.SetCoreFields(p.LogType(), (timestamp.RFC3339)(time.Now().UTC()))
+		event.PantherEventTimeWhenParsed = aws.Bool(true)
```
stoggi: @rleighton From the RFC (https://tools.ietf.org/html/rfc5424#section-6.2.3):
>  A syslog application MUST use the NILVALUE as TIMESTAMP if the syslog
>  application is incapable of obtaining system time.

For this corer case, setting the PantherEventTime to `time.Now().UTC()`.

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
```diff
@@ -0,0 +1,109 @@
...
+// RFC5424Parser parses Syslog logs in the RFC5424 format
+type RFC5424Parser struct{}
+
+func (p *RFC5424Parser) New() parsers.LogParser {
+	return &RFC5424Parser{}
+}
+
+// Parse returns the parsed events or nil if parsing failed
+func (p *RFC5424Parser) Parse(log string) []interface{} {
+	parser := rfc5424.NewParser(rfc5424.WithBestEffort())
```
kostaspap: Can the parser be used concurrently? If yes, I'd suggest declaring it as a var once and reuse it on every `Parse` method invocation. No need to create it each time.

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
kostaspap: I'd say just keep them uint8 in the code and just add a mapping in our Glue table generation logic to convert the `uint8` to `SMALLINT`. In practice the result will be the same as now, just you avoid the casting in code

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
kostaspap: 1. We should make sure that when the value of `PantherEventTimeWhenParsed` is not true, it should always be false, never nil. Maybe the setCoreFields could automatically set the value to 'false' and each individual parser can modify it to true when needed. 
2. I am not sure I like the name. I am thinking sth like `p_inferred_timestamp` or `p_timestamp_in_logs` or ... @rleighton do you have some thoughts?

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
```diff
@@ -0,0 +1,109 @@
...
+var RFC5424Desc = `Syslog parser for the RFC5424 format.
+Reference: https://tools.ietf.org/html/rfc5424`
+
+// nolint:lll
+type RFC5424 struct {
+	Priority       *int16                        `json:"priority" validate:"required" description:"Priority is calculated by (Facility * 8 + Severity). The lower this value, the higher importance of the log message."`
+	Facility       *int16                        `json:"facility" validate:"required" description:"Facility value helps determine which process created the message. Eg: 0 = kernel messages, 3 = system daemons."`
+	Severity       *int16                        `json:"severity" validate:"required" description:"Severity indicates how severe the message is. Eg: 0=Emergency to 7=Debug."`
+	Version        *int16                        `json:"version" validate:"required" description:"Version of the syslog message protocol. RFC5424 mandates that version cannot be 0, so a 0 value signals no version."`
+	Timestamp      *timestamp.RFC3339            `json:"timestamp,omitempty" description:"Timestamp of the syslog message."`
```
rleighton: Add "in UTC" to description

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
rleighton: Naming is hard. The should convey that the time now was "processing time". I don't have a better suggestion that what is there.

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424_test.go
rleighton: Yeah, I hear you. I think is fine vs making the more complex with a time function for testing. If this was a more complex issue we would want some kind of mocking to drive it. What some people do is just have a global var bound to time.Now().UTC() and then switch it in the tests.

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
rleighton: Be very careful about it holding any state if you do this!

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
rleighton: Can you update this doc with `PantherEventTimeWhenParsed` : https://github.com/panther-labs/panther/blob/master/docs/gitbook/historical-search/panther-fields.md

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	internalRFC5424 := msg.(*rfc5424.SyslogMessage)
+
+	externalRFC5424 := &RFC5424{
+		Priority:       aws.Int16(int16(*internalRFC5424.Priority)),
+		Facility:       aws.Int16(int16(*internalRFC5424.Facility)),
+		Severity:       aws.Int16(int16(*internalRFC5424.Severity)),
+		Version:        aws.Int16(int16(internalRFC5424.Version)),
+		Timestamp:      (*timestamp.RFC3339)(internalRFC5424.Timestamp),
+		Hostname:       internalRFC5424.Hostname,
```
rleighton: Should updatePantherFields() add Hostname to PantherAnyDomainNames ?

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
rleighton: Can you update this doc with PantherEventTimeWhenParsed : https://github.com/panther-labs/panther/blob/master/docs/gitbook/historical-search/panther-fields.md

Also, can you add documentation for the hash fields you added earlier (I forgot to ask you to do that).

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
nhakmiller: In general are we using that field for any plain english network identifiers? I think my machine would report this field as `Nicholass-MacBook-Pro`, which is not exactly what I would be expecting in a field called `PantherAnyDomainNames`. But if it's generally a field for any network identifier, local or fully qualified, then that would be reasonable.

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424_test.go
nhakmiller: We also do this in a few places in the cloud security side of things, for integration tests where we are expecting results with random IDs or timestamps. I think it is reasonable.

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
stoggi: Hm, what about adding `time.Now().UTC()` as `PantherParseTime` to all events? And then making the EventTime optionally nil, or equal to ParseTime in these cases where there is no event time. This feels like something that will occur with other log types too, and it may be useful to know what the lag between event time and parse time is as a metric :thinking:

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
stoggi: From the RFC (https://tools.ietf.org/html/rfc5424#section-6.2.4)

> 
>  The HOSTNAME field SHOULD contain the hostname and the domain name of
>    the originator in the format specified in STD 13 [RFC1034].  This
>    format is called a Fully Qualified Domain Name (FQDN) in this
>    document.
> 
>    In practice, not all syslog applications are able to provide an FQDN.
>    As such, other values MAY also be present in HOSTNAME.  This document
>    makes provisions for using other values in such situations.  A syslog
>    application SHOULD provide the most specific available value first.
>    The order of preference for the contents of the HOSTNAME field is as
>    follows:
> 
>    1.  FQDN
> 
>    2.  Static IP address
> 
>    3.  hostname
> 
>    4.  Dynamic IP address
> 
>    5.  the NILVALUE

So yeah it could be many things. We could attempt to parse the hostname and add the field to AnyIPAddresses or AnyHostNames if successful?

rleighton: It is a pretty important key. Suggestion:  Add to domains UNLESS it looks like and IP address (pre-compile regex for IP and put in panther common fields so we can use this in other parsers) and then put in any ips.

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
rleighton: Actually, that is a common ETL pattern and very useful (e.g., when you get late data).
Unless Kostas has issues, I think we should to this. `PantherParseTime` is a required field. I would finish this PR with this change and then do one that updates the other parsers and documentation.

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
stoggi: Sounds good, I will check for IP.

`net.ParseIP` is surprisingly quite performant, I found this blog post here comparing different methods of parsing IPs:

https://medium.com/@sergio.anguita/detecting-a-valid-ipv4-in-go-like-a-boss-32eda9bf422f

The code looks quite reasonable too:
https://golang.org/src/net/ip.go?s=15675:15700#L538

Given I need to parse ipv4 and ipv6, I will add `if net.ParseIP(string) == nil` in the parser.

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
stoggi: @rleighton I ended up creating an issue explaining `PantherParseTime` (#282) and will submit a separate PR for it. Then I can merge master into this branch and submit it as only the Syslog Parser.

kostaspap: Agree with the `PantherParseTime`. It is very common to have such a field processes like that, it could help down the road with generating metrics etc

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
rleighton: I mentioned this in https://github.com/panther-labs/panther/pull/320 , I think you should
move the rfc5424.NewParser(rfc5424.WithBestEffort()) into the New() method above which is called once per file.

stoggi: I like this suggestion. One thing, is the tests don't call `New()` on the parser. I suggest updating (at least the rfc5421_test.go tests) to use this pattern:

```go
var parser parsers.LogParser

func TestRFC5424(t *testing.T) {
	zap.ReplaceGlobals(zaptest.Logger(t))
	syslogRFC5424 := &RFC5424Parser{}
	parser = syslogRFC5424.New()

	t.Run("Version4", testRFC5424Version4)
	t.Run("NoTimestmap", testRFC5424NoTimestmap)
	t.Run("NoStructuredDataNoProcID", testRFC5424NoStructuredDataNoProcID)
	t.Run("NoStructuredDataNoMsgID", testRFC5424NoStructuredDataNoMsgID)
	t.Run("WithStructuredData", testRFC5424WithStructuredData)
	t.Run("StructuredDataOnly", testRFC5424StructuredDataOnly)
}
```

That way we can verify all tests pass while reusing the same global `parser`. I also want to add [zaptest](https://github.com/fgrosse/zaptest) to surface debug log messages in failed tests. This helps debug tests. A failed test would look like this:
```
23:06:38	INFO	test:go: unit tests
--- FAIL: TestRFC5424 (0.00s)
    logger.go:81: DEBUG	failed to validate log	{"error": "Key: 'RFC5424.PantherLog.PantherLogType' Error:Field validation for 'PantherLogType' failed on the 'required' tag\nKey: 'RFC5424.PantherLog.PantherRowID' Error:Field validation for 'PantherRowID' failed on the 'required' tag\nKey: 'RFC5424.PantherLog.PantherEventTime' Error:Field validation for 'PantherEventTime' failed on the 'required' tag"}
    --- FAIL: TestRFC5424/NoTimestmap (0.00s)
        rfc5424_test.go:253:
            	Error Trace:	rfc5424_test.go:253
            	            				rfc5424_test.go:112
            	Error:      	Not equal:
            	            	expected: 1
            	            	actual  : 0
            	Test:       	TestRFC5424/NoTimestmap
```

internal/log_analysis/log_processor/parsers/sysloglogs/rfc5424.go
```diff
@@ -0,0 +1,121 @@
...
+	Timestamp      *timestamp.RFC3339            `json:"timestamp,omitempty" description:"Timestamp of the syslog message in UTC."`
+	Hostname       *string                       `json:"hostname,omitempty" description:"Hostname identifies the machine that originally sent the syslog message."`
+	Appname        *string                       `json:"appname,omitempty" description:"Appname identifies the device or application that originated the syslog message."`
+	ProcID         *string                       `json:"procid,omitempty" description:"ProcID is often the process ID, but can be any value used to enable log analyzers to detect discontinuities in syslog reporting."`
+	MsgID          *string                       `json:"msgid,omitempty" description:"MsgID identifies the type of message. For example, a firewall might use the MsgID 'TCPIN' for incoming TCP traffic."`
+	StructuredData *map[string]map[string]string `json:"structured_data,omitempty" description:"StructuredData provides a mechanism to express information in a well defined and easily parsable format."`
```
kostaspap: Are you able do query this data in Athena using the `json_extract` method? I am thinking this should be of type `*jsoniter.RawMessage`, what do you think @rleighton  ?

kostaspap: Had discussion with @rleighton offline. There is definitely some benefit of defining it as `map<string,map<string,string>>` in Athena - users don't have to use `json_extract` to query the data, can just do "SELECT structured_data['key1']['key2'] ...."

It's fine to keep like that.

PR #285, opened 2020-02-26
Update example-queries.md

sugandhalahoti:
## Background

updating documentation

## Changes

-minor language changes

## Testing

### DIFF
--- a/docs/gitbook/historical-search/example-queries.md
+++ b/docs/gitbook/historical-search/example-queries.md
@@ -5,7 +5,7 @@ Please note that all queries should be qualified with partition columns (year, m
 ## Did this IP address have any activity in my network (and in what logs)?
 
 This is often one of the first questions asked in an investigation. Given there is some known bad indicator such
-as an IP address, then if there is related activity in your network/systems, a detailed investigation will be needed to proceed.
+as an IP address, then if there is related activity in your network/systems, a detailed investigation will be needed.
 Panther makes asking such questions easy using the 'all_logs' Athena view which will search all data for the
 indicator of interest. Since most often the answers to these question are negative, making this a fast and efficient
 operation reduces investigation time. In this example the Panther field `p_any_ip_addresses` is used. Panther extracts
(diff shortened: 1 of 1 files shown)


docs/gitbook/historical-search/example-queries.md
```diff
@@ -136,7 +136,7 @@ ORDER BY p_event_time ASC
 
 ## Find all of the sourceIPAddresses for console logins in CloudTrail and rank
 
-This query is similar to the above query but instead ranking the IP addresses for all console logins to understand both which
+This query is similar to the above query but instead we are ranking the IP addresses for all console logins to understand which
```
3nvi: english is weird in both the original and the modified version. Generally, in official documents you speak in passive form without adding a person (i.e. instead of `we are ranking` you would say something like `is being ranked`)

For this particular case, I would change to:


>This query is similar to the query above, with the IP addresses ranked for all console logins. This helps identifying the IP addresses that are signing into the console, while ranking the relative activity. ....blah blah blah

docs/gitbook/historical-search/example-queries.md
```diff
@@ -4,12 +4,12 @@ Please note that all queries should be qualified with partition columns (year, m
 
 ## Did this IP address have any activity in my network (and in what logs)?
 
-This is often one of the first questions asked in an investigation. Given there is some known-bad indicator such
-as an IP address, then if there is related activity in your network/systems a detailed investigation will need to proceed.
+This is often one of the first questions asked in an investigation. Given there is some known bad indicator such
+as an IP address, then if there is related activity in your network/systems, a detailed investigation will be needed to proceed.
```
3nvi: again the original version has a weird use of the verb `proceed`. Generally, you say `proceed **with** something`. 

I would remove `to proceed` and just have `a detailed investigation will be needed.`

docs/gitbook/historical-search/example-queries.md
sugandhalahoti: okay, will remember that for future corrections. Have made edits to this as suggested.

docs/gitbook/historical-search/example-queries.md
sugandhalahoti: agreed, made changes

PR #298, opened 2020-02-26
Glue updater lambda

kostaspap:
## Background

Introducing Data Catalog Updater Lambda. The lambda listers to SNS notifications for new objects in S3 in Panther "ProcessedData" bucket and create new partitions in tables as needed. 

## Changes

- Moved the "S3Notification" - the notification sent to the SNS topic when new data is available to our "api" folder 
- Fixed file extension for data we store to S3 to more appropriate `.json.gz`
- Fixed rule-engine S3 objects to pad integers. Added unit test to verify appropriate format
- Fixed issue that integration tests in the `pkg` folder were not running
- Did some refactoring in the glue package: 
  - Moved GlueTableTimebin to its own file
  - Renamed "GlueMetadata" to "GlueTableMetadata"
  - Moved all methods of "GlueMetadata" struct and tests to the same file. Previously the methods of the struct were split across files.
  - Introducing GluePartition struct that can be used to create partitions in Glue. Added a method that infers GluePartition information from S3 bucket and S3 object
  - Changed the constructor of "GlueMetadata". The constructor now gets as argument the log type and the type of the data (rule output or logs) along with some other information (time partitions granularity) and returns the GlueMetadata. This change was required so that the logic of how the type of the data(rules, logs) maps to databases is contained within the same place - so the same logic can be used (in reverse) to infer the database + data type from the s3 object key. 
  - I have kept the pre-existing CreateJSONPartition method since it was used by the `glue:sync` method and removing it would require much further refactoring and make this PR bigger than it should. However I have added a note that this is deprecated - will fix later on
     - I fix some of the tests that were used to test this method (adding `mock.AssertExpectations(t)` showed some failures) and split one big test to several smaller (so failures are easier to debug)

## Testing

- Deployed and seen expected behavior (logs and rules tables updated)
- Unit tests


internal/log_analysis/rules_engine/src/output.py
```diff
@@ -30,7 +30,7 @@
 from .alert_merger import update_get_alert_info
 from .logging import get_logger
 
-_KEY_FORMAT = 'rules/{}/year={}/month={}/day={}/hour={}/rule_id={}/{}-{}.gz'
+_KEY_FORMAT = 'rules/{}/year={:02d}/month={:02d}/day={:02d}/hour={:02d}/rule_id={}/{}-{}.json.gz'
```
rleighton: 4 digits

pkg/awsglue/glue.go
```diff
@@ -36,181 +43,202 @@ const (
...
-	if gm.timeUnpadded {
-		intFormat = "%d"
-	} else {
-		intFormat = "%02d"
+// Based on Timebin(), return an S3 prefix for objects of this table
+func (gm *GlueTableMetadata) GetPartitionPrefix(t time.Time) (prefix string) {
+	prefix = gm.Prefix()
+	switch gm.timebin {
+	case GlueTableHourly:
+		prefix += fmt.Sprintf("year=%02d/month=%02d/day=%02d/hour=%02d/", t.Year(), t.Month(), t.Day(), t.Hour())
```
rleighton: 4 digits for year in all cases

pkg/awsglue/glue.go
```diff
@@ -36,181 +43,202 @@ const (
...
+		PartitionValues: gm.partitionValues(t),
+	}
+	return client.DeletePartition(input)
 }
 
-func standardizeS3Prefix(s3Prefix, tableName string) string {
-	return s3Prefix + "/" + tableName + "/" // ensure last char is '/'
+// Based on Timebin(), return an []*string values (used for GlueTableMetadata APIs)
+func (gm *GlueTableMetadata) partitionValues(t time.Time) (values []*string) {
+	values = []*string{aws.String(fmt.Sprintf("%d", t.Year()))} // always unpadded
```
rleighton: comment not needed anymore

pkg/awsglue/glue.go
kostaspap: Good catch! Previous code was just using "%d" (since year doesn't need padding), I'll switch back to that.

pkg/awsglue/glue.go
kostaspap: Done

go.mod
```diff
@@ -3,23 +3,56 @@ module github.com/panther-labs/panther
...
+	github.com/huandu/xstrings v1.3.0 // indirect
+	github.com/imdario/mergo v0.3.8 // indirect
+	github.com/jaytaylor/html2text v0.0.0-20200220170450-61d9dc4d7195 // indirect
 	github.com/joho/godotenv v1.3.0
 	github.com/json-iterator/go v1.1.9
+	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/leodido/go-urn v1.2.0 // indirect
 	github.com/magefile/mage v1.9.0
+	github.com/mailru/easyjson v0.7.1 // indirect
+	github.com/matcornic/hermes v1.2.0
```
austinbyers: I still don't fully understand exactly how this file is supposed to work - why did these all get added now and before?

You can run `go mod tidy` to cleanup unused dependencies

internal/log_analysis/datacatalog_updater/main/lambda.go
```diff
@@ -0,0 +1,86 @@
...
+		gluePartition, err := awsglue.GetPartitionFromS3(*notification.S3Bucket, *notification.S3ObjectKey)
+		if err != nil {
+			zap.L().Error("failed to get partition information from notification",
+				zap.Any("notification", notification), zap.Error(errors.WithStack(err)))
+			continue
+		}
+		err = gluePartition.CreatePartition(glueClient)
+		if err != nil {
+			zap.L().Error("failed to create partition", zap.Any("notification", notification), zap.Error(errors.WithStack(err)))
+			return err
```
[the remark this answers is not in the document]
kostaspap: I've actually put that line in mostly to log the `notification` struct - does it make sense? The error will be logged as you mention.

go.mod
kostaspap: Good suggestion! 
I did a `go get -u ./...` to make sure I have the latest version and then `go mod tidy`.

deployments/log_analysis/log_processor.yml
```diff
@@ -167,3 +158,94 @@ Resources:
...
+            Action: sqs:SendMessage
+            Resource: '*'
+            Condition:
+              ArnLike:
+                aws:SourceArn: !Ref SnsTopicArn
+
+  UpdaterQueue:
+    Type: AWS::SQS::Queue
+    Properties:
+      QueueName: panther-datacatalog-updater
```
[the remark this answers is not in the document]
rleighton: Can you add comment below (see sqs q's in my PR) like:
```
# <cfndoc>
# describe dependencies
# Failure Impacts
# * what ever
# * explain about dlq and requeue (see PR)
# </cfndoc>
```

deployments/log_analysis/log_processor.yml
```diff
@@ -167,3 +158,94 @@ Resources:
...
+      KmsMasterKeyId: !Ref SQSKeyId
+      VisibilityTimeout: 30 # Should match lambda
+      RedrivePolicy:
+        deadLetterTargetArn: !GetAtt UpdaterDeadLetterQueue.Arn
+        maxReceiveCount: 10
+
+  UpdaterDeadLetterQueue:
+    Type: AWS::SQS::Queue
+    Properties:
+      QueueName: panther-datacatalog-updater-dlq
```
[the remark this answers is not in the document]
rleighton: Can you add comment below (see sqs q's in my PR) like:
```
# <cfndoc>
# describe dependencies
# explain about dlq and requeue (see PR)
# </cfndoc>

```

deployments/log_analysis/log_processor.yml
```diff
@@ -167,3 +158,94 @@ Resources:
...
+  UpdaterFunctionLogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: /aws/lambda/panther-datacatalog-updater
+      RetentionInDays: !Ref CloudWatchLogRetentionDays
+
+  UpdaterFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      FunctionName: panther-datacatalog-updater
```
rleighton: Can you add comment below (see sqs q's in my PR) like:
```
# <cfndoc>
# describe dependencies
# Failure Impacts
# * explain
# </cfndoc>
```

internal/log_analysis/datacatalog_updater/main/lambda.go
```diff
@@ -0,0 +1,96 @@
...
+			zap.L().Debug("partition has already been created")
+			continue
+		}
+
+		err = gluePartition.CreatePartition(glueClient)
+		if err != nil {
+			zap.L().Error("failed to create partition", zap.Any("notification", notification), zap.Error(errors.WithStack(err)))
+			return err
+		}
+		partitionPrefixCache[gluePartition.GetPartitionPrefix()] = struct{}{}
```
rleighton: Do we have to worry about 2 lambda invocations concurrently updating? I _think_ we do. If so, I would wrap the stack in a struct with a mutex and methods to add/check.

internal/log_analysis/datacatalog_updater/main/lambda.go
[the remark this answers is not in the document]
kostaspap: Sure, changing.

deployments/log_analysis/log_processor.yml
```diff
@@ -116,15 +116,6 @@ Resources:
               # This account will be whitelisted and SNS topic from it can subscribe to the SQS queue.
               Action: sns:ConfirmSubscription
               Resource: '*'
-        - Id: ReceiveFromInputSqsQueue
```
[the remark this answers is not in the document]
kostaspap: Yes! Austin made similar comment and realized it was not needed. The CFN tranformation adds the appropriate permissions automatically.

tools/cfngen/gluecf/gluecf.go
```diff
@@ -136,17 +140,19 @@ func GenerateTables(tables []*awsglue.GlueMetadata) (cf []byte, err error) {
...
 	// add tables for all parsers, and matching tables for rule matches
-	for _, t := range tables {
-		addTable(t)
+	for _, table := range tables {
+		addTable(table)
+		ruleTable := awsglue.NewGlueTableMetadata(
+			models.RuleData, table.LogType(), table.Description(), awsglue.GlueTableHourly, table.EventStruct())
 		// add a matching table for rule matches, add the columns that the rules engine appends
-		addTable(t.Clone(awsglue.RuleMatchS3Prefix, awsglue.RuleMatchDatabaseName), RuleMatchColumns...)
+		addTable(ruleTable, RuleMatchColumns...)
```
rleighton: The reason I added Clone() , which you took out was for here... the S3 prefix paths will not be correct, the match tables will be referring to the rule s3 paths , right? You want the match tables to be the same EXCEPT the s3 path and the extra columns.

internal/log_analysis/datacatalog_updater/main/lambda.go
kostaspap: Two lambda's can be invoked concurrently but each one of them will have its own map (each Lambda runs in its own container). The problem would be if I was be accessing the map from multiple goroutines within the same Lambda invocation. So it's not really required.

deployments/log_analysis/log_processor.yml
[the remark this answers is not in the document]
kostaspap: > Change name to 'panther-datacatalog-updater-queue-dlq'

`dlq` stands for DeadLetterQueue. Not sure the `queue` suffix is needed here. There is no naming conflict (e.g. with Lambdas) in any case.

internal/log_analysis/datacatalog_updater/main/lambda.go
rleighton: Ok, I was just checking that they do not call the function against the container concurrently since they let the container live for hours.

deployments/log_analysis/log_processor.yml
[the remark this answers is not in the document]
kostaspap: Done

deployments/log_analysis/log_processor.yml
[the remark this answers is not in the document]
kostaspap: Added documentation.

deployments/log_analysis/log_processor.yml
kostaspap: Added.

tools/cfngen/gluecf/gluecf.go
kostaspap: This table contains the rule matches. This table will be the same as the "normal" table, the only difference with the normal table is that it has a different database and a different prefix (`rules/table_name` vs `logs/table_name` for the log data). I think that's what we wanted in any case, right?

rleighton: Right nevermind, misread. I thought you passed the table to the second addTable(). Need more coffee. Clone() is more concise (but the same thing as making a new one). I don't really care.

kostaspap: Cool thanks. My motivation behind this change was to "hide" the table prefix / database names from packages outside awsglue so it's not possible to create a GlueTableMetadata object with e.g. rule prefix but log database.

PR #303, opened 2020-02-27
Policy/Alert URLs + Cloud Sec Overview Wording

jacknagz:
## Background

I found a bug in the alerting workflow after we split the pages. The policy/alert links were missing the new path prefixes.

Made some additional wording edits too.

Request - @3nvi can we make the "Resource Platforms" breakdown by type?

## Changes

- Fix alert/policy urls
- Reword the auth page text again (can we also increase the padding around the text?)
- Rename the chart headers on the Cloud Security overview.

## Testing

- Deployed into my test account


3nvi:
Thanks, didn't even know we were using the URLs in cloudformation

Issue #282, opened 2020-02-26
The Log Parser should record the time each event is parsed.


stoggi:
### Describe the ideal solution

When the Parser successfully parses an event, it should record the current time in UTC and store it with the event.

### Describe your use cases

Some log entries may not have a timestamp, for example Syslog RFC5424 (#261). However the `PantherEventTime` field is required to correctly store and find historical events. In this case, it would be good to know that the EventTime was set to a "best guess" of the time it was parsed.

Recording the time an event is parsed is useful when evaluating the "lag" of a logging and alerting system. Knowing how long it takes from log creation to successful log parsing can be a useful metric in overall logging infrastructure health.

In the event when the logs are severely behind, `parsed time - event time` be monitored to ensure the system is catching up.

### Describe alternatives you have considered

In #261, an alternative was considered: A boolean flag that indicated when the `PantherEventTime` was set to the current time. This only accounted for the first use case, where an event has no timestamp, not for monitoring lag.

One consideration is the overhead of calling `time.Now()` for every parsed log. [This blog post](https://blog.cloudflare.com/its-go-time-on-linux/) has some interesting information about `time.Now()` in Golang. On a 64bit Linux system this might add somewhere in the order of 20 nano seconds for each call. Would we want to store `time.Now()` once for the lambda invocation, and pass it to each classifier?

rleighton:
This is needed for V1.

rleighton:
Good point about calling time.Now() for event.

I suggest you change the interface for a parser from:
```
	// Parse attempts to parse the provided log line
	// If the provided log is not of the supported type the method returns nil
	Parse(log string) []interface{}
```
to:
```
	// Parse attempts to parse the provided log line
	// If the provided log is not of the supported type the method returns nil
	Parse(parseTime time.Time, log string) []interface{}
```
See: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/parsers/parsers.go#L30

The classifier (which calls the parsers) already keeps a start time:
https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/classification/classifier.go#L135

So, pass that down to where it calls Parse().

kostaspap:
[...]

PR #304, opened 2020-02-27
Add PantherParseTime to the Parser

stoggi:
## Background

Adding a new field `PantherParseTime` to each parsed log. #282 

## Changes

* Created new required field `PantherParseTime` in the common `PantherLog` struct. Note: This will require glue partitions to be rebuilt.
* Handle the case where the eventTime is nil, and set it to the parseTime.
* Handle the case where parseTime is nil, fall back to `timestamp.Now()`.
* Updated the classifer to pass it's startParseTime to the parser.
* Updated each parser to accept parseTime and update PantherParseTime.
* Refactored unit tests to expect a parse time.
* Added field p_parse_time to the expected view generated for Athena.
* Added a convinience method to timestamp to get a UTC time.Now() in RFC3339.

## Testing
- Unit tests
- Deployed Panther into a test account, and verified a log is generated end-to-end

Example event captured by Panther with the `p_parse_time` field:
```
{
	"id": "1540845340.16991",
	"rule": {
		"comment": "Integrity checksum changed.",
		"group": "ossec,syscheck,",
		"level": 7,
		"sidid": 550
	},
	"TimeStamp": "2018-10-29 20:35:40.000000000",
	"location": "syscheck",
	"hostname": "ip-172-16-2-16",
	"full_log": "Integrity checksum changed for: '/usr/bin/ssm-cli'\nOld md5sum was: '22271cce0732d887e3980e5a6868e459'\nNew md5sum is : '220a8f105af5e711f99e52583209a871'\nOld sha1sum was: '4df65340f366c18f85be228c26817e20391f32c4'\nNew sha1sum is : 'c7414fd048c81361720e2d9c8d2f82faf33748b6'\n",
	"decoder": "syscheck_integrity_changed",
	"SyscheckFile": {
		"md5_after": "220a8f105af5e711f99e52583209a871",
		"md5_before": "22271cce0732d887e3980e5a6868e459",
		"path": "/usr/bin/ssm-cli"
	},
	"p_log_type": "OSSEC.EventInfo",
	"p_row_id": "226e61b7417f90a7dae5ce8f018601",
	"p_event_time": "2018-10-29 20:35:40.000000000",
	"p_parse_time": "2020-02-27 03:08:16.796940520",
	"p_any_md5_hashes": [
		"220a8f105af5e711f99e52583209a871",
		"22271cce0732d887e3980e5a6868e459"
	]
}
```

Verified View in Athena has `p_parse_time`:
```
CREATE OR REPLACE VIEW all_logs AS 
SELECT
  "day"
, "hour"
, "month"
, null "p_any_aws_account_ids"
, null "p_any_aws_arns"
, null "p_any_aws_instance_ids"
, null "p_any_aws_tags"
, "p_any_ip_addresses"
, "p_any_ip_domain_names"
, "p_any_md5_hashes"
, "p_any_sha1_hashes"
, "p_event_time"
, "p_log_type"
, "p_parse_time"
, "p_row_id"
, "year"
FROM
  panther_logs.osquery_snapshot
```


stoggi:
Follow up question for my comment above:
> Created new required field PantherParseTime in the common PantherLog struct. Note: This will require glue partitions to be rebuilt.

I noticed in the code there was this comment:
https://github.com/panther-labs/panther/blob/a81e7b6d6f683b36d19bcced0dd9f3169a826411/internal/log_analysis/log_processor/parsers/patherlog.go#L37-L40

But it seems that the fields in the Athena view are ordered alphabetically. So wouldn't adding at the end still require the partitions to be reloaded?

stoggi:
> But it seems that the fields in the Athena view are ordered alphabetically. So wouldn't adding at the end still require the partitions to be reloaded?

Ah, I think I have found my own answer. The view is alphabetical, but the table is not:
```
p_log_type (string)
p_row_id (string)
p_event_time (timestamp)
p_parse_time (timestamp)
p_any_ip_addresses (array<string>)
p_any_ip_domain_names (array<string>)
p_any_sha1_hashes (array<string>)
p_any_md5_hashes (array<string>)
```

I'm not sure what the implications are of adding `p_parse_time` in the middle are. Do I just need to run `mage glue:sync` or is there something else to do?

kostaspap:
See my latest comment: https://github.com/panther-labs/panther/issues/282#issuecomment-591845273 
Happy to discuss further if needed.

stoggi:
> See my latest comment: [#282 (comment)](https://github.com/panther-labs/panther/issues/282#issuecomment-591845273)
> Happy to discuss further if needed.

Simplified the pull request with this suggestion. In order to keep the API consistent (between `SetCoreFields` and `SetCoreFieldsPtr`), `SetCoreFieldsPtr` still expects a parseTime pointer, but if nil will set to the current time. This way it can be overridden if necessary, but most parsers can just let `SetCoreFieldsPtr` set it for them.

Ran another end-to-end test and verified the `p_parse_time` was in the captured event.

internal/log_analysis/log_processor/parsers/osquerylogs/batch.go
```diff
@@ -78,8 +78,6 @@ func (p *BatchParser) LogType() string {
 }
 
 func (event *Batch) updatePantherFields(p *BatchParser) {
-	if event.CalendarTime != nil {
-		event.SetCoreFields(p.LogType(), timestamp.RFC3339(*event.CalendarTime))
-	}
+	event.SetCoreFieldsPtr(p.LogType(), (*timestamp.RFC3339)(event.CalendarTime), nil)
```
stoggi: This was updated (and in the following events for the osquerylogs parser) to use `SetCoreFieldsPtr` like the other parsers. This means that a `event.CalendarTime` of nil will create a valid log and set the `PantherEventTime` to the parse time.

internal/log_analysis/log_processor/parsers/patherlog.go
```diff
@@ -88,17 +89,23 @@ func (any *PantherAnyString) UnmarshalJSON(jsonBytes []byte) error {
...
-		pl.SetCoreFields(logType, *eventTime)
+func (pl *PantherLog) SetCoreFieldsPtr(logType string, eventTime *timestamp.RFC3339, parseTime *timestamp.RFC3339) {
+	if parseTime == nil {
+		pt := timestamp.Now()
+		parseTime = &pt
 	}
+	if eventTime == nil {
+		eventTime = parseTime
+	}
+	pl.SetCoreFields(logType, *eventTime, *parseTime)
```
rleighton: Should we reverse the call order and have SetCoreFields() call SetCoreFieldsPtr() to avoid some copies?
You would move the NewRowID() into SetCoreFieldsPtr(). I think it is the way it is now due to how it was developed , where SetCoreFieldsPtr() was added later I but I think it should be reversed.

stoggi: Hm, it's a good point. In that case should we reverse the other ones too?

1. AppendAnyIPAddresses/AppendAnyIPAddressPtrs
2. AppendAnyDomainNames/AppendAnyDomainNamePtrs
3. ...

If the core fields might be refactored soon maybe it's not worth changing?

rleighton:
Might as well do it now for those too.  The refactoring I was referring to
earlier is the interface between classifier and parsers.

internal/log_analysis/log_processor/parsers/patherlog.go
```diff
@@ -88,17 +89,23 @@ func (any *PantherAnyString) UnmarshalJSON(jsonBytes []byte) error {
...
-	pl.PantherLogType = &logType
 	rowID := rowCounter.NewRowID()
 	pl.PantherRowID = &rowID
-	pl.PantherEventTime = &eventTime
+	pl.PantherLogType = &logType
+	pl.PantherEventTime = eventTime
+	pl.PantherParseTime = parseTime
+}
+
+func (pl *PantherLog) SetCoreFields(logType string, eventTime timestamp.RFC3339, parseTime timestamp.RFC3339) {
```
kostaspap: This is called only internally I think? So probably you can just lowercase it?

internal/log_analysis/log_processor/parsers/patherlog.go
```diff
@@ -88,17 +89,23 @@ func (any *PantherAnyString) UnmarshalJSON(jsonBytes []byte) error {
 	return nil
 }
 
-func (pl *PantherLog) SetCoreFieldsPtr(logType string, eventTime *timestamp.RFC3339) {
-	if eventTime != nil {
-		pl.SetCoreFields(logType, *eventTime)
+func (pl *PantherLog) SetCoreFieldsPtr(logType string, eventTime *timestamp.RFC3339, parseTime *timestamp.RFC3339) {
```
kostaspap: `parseTime` is always be `nil` throughout the code - at it is intended to be like that since it's now the responsibility of the `SetCoreFieldsPtr` method to set the appropriate value. 

You can just change the method signature to 
```
SetCoreFieldsPtr(logType string, eventTime *timestamp.RFC3339)
```

stoggi: Yep, I mainly wanted to keep the two methods consistent with others using the same pattern, but if I lowercase setCoreFields, then I agree I can change the method signature.

internal/log_analysis/log_processor/parsers/patherlog.go
stoggi: Actually it's not used at all. Should I remove it, and rename `SetCoreFieldsPtr` to just `SetCoreFields`?

internal/log_analysis/log_processor/parsers/awslogs/aurora_mysql_audit_test.go
```diff
@@ -73,5 +73,13 @@ func checkAuroraMysqlAuditLogLog(t *testing.T, log string, expectedEvent *Aurora
 	require.Greater(t, len(*event.PantherRowID), 0) // ensure something is there.
 	expectedEvent.PantherRowID = event.PantherRowID
 
+	// For a nil timestamp, expect the event time to be the parse time
+	if expectedEvent.PantherEventTime == nil {
+		expectedEvent.PantherEventTime = event.PantherParseTime
+	}
```
kostaspap: This is not really true? The Aurora MySQL logs always have a timestamp

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_test.go
```diff
@@ -154,6 +154,14 @@ func checkCloudTrailLog(t *testing.T, log string, expectedEvents []*CloudTrail)
 		event := events[i].(*CloudTrail)
 		require.Greater(t, len(*event.PantherRowID), 0) // ensure something is there.
 		event.PantherRowID = expectedEvent.PantherRowID
+
+		// For a nil timestamp, expect the event time to be the parse time
+		if expectedEvent.PantherEventTime == nil {
+			expectedEvent.PantherEventTime = event.PantherParseTime
+		}
```
kostaspap: This check doesn't seem valid. CloudTrail events always have timestamp.

internal/log_analysis/log_processor/parsers/awslogs/guardduty_test.go
```diff
@@ -189,5 +189,13 @@ func checkGuardDutyLog(t *testing.T, log string, expectedEvent *GuardDuty) {
 	require.Greater(t, len(*event.PantherRowID), 0) // ensure something is there.
 	expectedEvent.PantherRowID = event.PantherRowID
 
+	// For a nil timestamp, expect the event time to be the parse time
+	if expectedEvent.PantherEventTime == nil {
+		expectedEvent.PantherEventTime = event.PantherParseTime
+	}
```
kostaspap: Same comment as above.

internal/log_analysis/log_processor/parsers/awslogs/s3_server_access_test.go
```diff
@@ -211,5 +211,13 @@ func checkS3AccessLog(t *testing.T, log string, expectedEvent *S3ServerAccess) {
 	require.Greater(t, len(*event.PantherRowID), 0) // ensure something is there.
 	expectedEvent.PantherRowID = event.PantherRowID
 
+	// For a nil timestamp, expect the event time to be the parse time
+	if expectedEvent.PantherEventTime == nil {
+		expectedEvent.PantherEventTime = event.PantherParseTime
+	}
```
kostaspap: Same comment as above.

internal/log_analysis/log_processor/parsers/awslogs/alb_test.go
```diff
@@ -246,5 +246,13 @@ func checkALBLog(t *testing.T, log string, expectedEvent *ALB) {
 	require.Greater(t, len(*event.PantherRowID), 0) // ensure something is there.
 	expectedEvent.PantherRowID = event.PantherRowID
 
+	// For a nil timestamp, expect the event time to be the parse time
+	if expectedEvent.PantherEventTime == nil {
+		expectedEvent.PantherEventTime = event.PantherParseTime
+	}
```
kostaspap: This doesn't seem required? All the tests have timestamp.

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow_test.go
```diff
@@ -145,5 +145,13 @@ func checkVPCFlowLog(t *testing.T, header, log string, expectedEvent *VPCFlow) {
 	require.Greater(t, len(*event.PantherRowID), 0) // ensure something is there.
 	expectedEvent.PantherRowID = event.PantherRowID
 
+	// For a nil timestamp, expect the event time to be the parse time
+	if expectedEvent.PantherEventTime == nil {
+		expectedEvent.PantherEventTime = event.PantherParseTime
+	}
```
kostaspap: Same

internal/log_analysis/log_processor/parsers/nginxlogs/access_test.go
```diff
@@ -90,5 +90,13 @@ func checkAccessLog(t *testing.T, log string, expectedEvent *Access) {
 	require.Greater(t, len(*event.PantherRowID), 0) // ensure something is there.
 	expectedEvent.PantherRowID = event.PantherRowID
 
+	// For a nil timestamp, expect the event time to be the parse time
+	if expectedEvent.PantherEventTime == nil {
+		expectedEvent.PantherEventTime = event.PantherParseTime
+	}
```
kostaspap: Same

internal/log_analysis/log_processor/parsers/osquerylogs/batch_test.go
```diff
@@ -81,5 +81,13 @@ func checkOsQueryBatcLog(t *testing.T, log string, expectedEvent *Batch) {
 	require.Greater(t, len(*event.PantherRowID), 0) // ensure something is there.
 	expectedEvent.PantherRowID = event.PantherRowID
 
+	// For a nil timestamp, expect the event time to be the parse time
+	if expectedEvent.PantherEventTime == nil {
+		expectedEvent.PantherEventTime = event.PantherParseTime
+	}
```
kostaspap: Same

PR #312, opened 2020-02-27
Opstool: requeue

rleighton:
## Background
When lambdas fail the messages end up in a dlq (often). When the issue is resolved oncall may want to push the messages back into the main queue.  The new documentation references this utility to help in these situations.

## Changes
- Added 'requeue' under /cmd
- Updated mage

## Testing
ran integration test
used in dev account to move failed messages in dlq back to processing q


cmd/opstools/requeue/integration_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// make 2 queues to move things between
+	fromq := queuePrefix + "-from"
+	toq := queuePrefix + "-toq"
+	// delete first in case these were left from previous failed test (best effort)
+	deletedQueue := false
+	err = deleteQueue(fromq)
+	if err == nil {
+		deletedQueue = true
+	}
```
kostaspap: if there is an error deleting the queue, shouldn't we just fail the test? If we continue and the we use the previous queue, the test might returns wrong results.

cmd/opstools/requeue/integration_test.go
```diff
@@ -0,0 +1,163 @@
...
+		// you have to wait 60+ seconds to use a q that has been deleted
+		time.Sleep(time.Second * 61)
+	}
+	err = createQueue(fromq)
+	require.NoError(t, err)
+	err = createQueue(toq)
+	require.NoError(t, err)
+
+	// add some messages to fromq
+	err = addQueue(fromq, numberTestBatches)
```
kostaspap: You can rename `addQueue` -> `addMessagesToQueue` , then the comment is not even needed.

cmd/opstools/requeue/integration_test.go
```diff
@@ -0,0 +1,163 @@
...
+	require.NoError(t, err)
+
+	// move them to toq
+	err = Requeue(sqsClient, fromq, toq)
+	require.NoError(t, err)
+
+	// check
+	numberMovedMessages, err := countQueue(toq)
```
kostaspap: `countQueue` -> `countQueueMessages`

cmd/opstools/requeue/requeue/main.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/panther-labs/panther/cmd/opstools/requeue"
+)
+
+const (
+	banner = "moves messages from one sqs queue to another"
+)
+
+var (
+	ACCOUNTID = flag.String("accountid", "", "The AWS account id where the queues exists")
```
kostaspap: I'm thinking this is probably not needed? The users are going to run this command from CLI and they need to use credentials of some account. We can just use the caller account (retrievable using https://docs.aws.amazon.com/sdk-for-go/api/service/sts/#STS.GetCallerIdentity )

cmd/opstools/requeue/requeue/main.go
```diff
@@ -0,0 +1,83 @@
...
+	flag.Usage = usage
+}
+
+func validateFlags() {
+	var err error
+	defer func() {
+		if err != nil {
+			fmt.Printf("%s\n", err)
+			flag.Usage()
+			os.Exit(-2)
```
kostaspap: We already  call "log.fatal" in case of an error that causes the program to terminate. I'm not sure this is needed?

cmd/opstools/requeue/requeue/main.go
```diff
@@ -0,0 +1,83 @@
...
+
+var (
+	ACCOUNTID = flag.String("accountid", "", "The AWS account id where the queues exists")
+	REGION    = flag.String("region", "", "The AWS region where the queues exists")
+	FROMQ     = flag.String("from.q", "", "The name of the queue to copy from")
+	TOQ       = flag.String("to.q", "", "The name of the queue to copy to")
+)
+
+func usage() {
+	fmt.Fprintf(flag.CommandLine.Output(),
```
kostaspap: In mage we use logger instead of fmt and that has proven popular. I'm thinking we should do the same here?

cmd/opstools/requeue/requeue/main.go
```diff
@@ -0,0 +1,83 @@
...
+	flag.Parse()
+
+	if *ACCOUNTID == "" {
+		err = errors.New("-accountid not set")
+		return
+	}
+	if *REGION == "" {
+		err = errors.New("-region not set")
+		return
+	}
```
kostaspap: I'm thinking probably we can use the default region if not specified.

tools/mage/deploy.go
```diff
@@ -122,6 +122,9 @@ func Deploy() {
 	// Deploy monitoring (must be last due to possible references to frontend/backend resources)
 	deployMonitoring(awsSession, bucket, backendOutputs, &config)
 
+	// Build opstools
+	Build.Opstools(Build{})
```
kostaspap: Seems weird to build the opstools after we have deployed the monitoring. Let's say that build fails, why should the deployment fail? 
it seems more appropriate to build the ops tools as part of the `test:go` and `test:ci` commands.

cmd/opstools/requeue/requeue.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/pkg/errors"
+)
+
+const (
+	waitTimeSeconds   = 20
+	messageBatchSize  = 10
+	visibilityTimeout = 20
+)
+
+func Requeue(sqsClient sqsiface.SQSAPI, fromq, toq string) error {
```
nhakmiller: With the go camelCase convention, would these be better named `fromQ` and `toQ`?

cmd/opstools/requeue/integration_test.go
rleighton: No, this is corner case . This is deleting _before_ the test incase a previous test failed and left them. Deleting here _should_ fail under normal circumstances. If it succeeds, you need to wait 60 seconds. At the end of the test the queues are deleted.

tools/mage/deploy.go
nhakmiller: But users don't need to run `test:go` or `test:ci` to deploy, and in fact may not be able to do so without doing a more complicated environment setup. I would think you would want ops tools even if you are just deploying and not developing.

cmd/opstools/requeue/requeue/main.go
rleighton: This prints out the nice help message:
```
Russells-MBP:panther russ$ ./out/bin/opstools/requeue 
-accountid not set
requeue moves messages from one sqs queue to another
Usage:
  -accountid string
    	The AWS account id where the queues exists
  -from.q string
    	The name of the queue to copy from
  -region string
    	The AWS region where the queues exists
  -to.q string
    	The name of the queue to copy to
```

cmd/opstools/requeue/requeue/main.go
rleighton: No, we do not want timestamps, this is the help message (see above).

tools/mage/deploy.go
rleighton: I'm not sure building tools as part of "test" is intuitive. The `mage build:optools` works but how do ensure it is already built for customers with making them run commands? That is why I added it to deploy.

cmd/opstools/requeue/requeue.go
austinbyers: I would change the name to be more informative - is it a queue name or an ARN?

cmd/opstools/requeue/requeue/main.go
austinbyers: Agreed - the account and region should be specified with environment variables or however else they load AWS credentials, we don't need to ask for them explicitly.

If you need the region in your code, you can get it from the AWS session. If you need the account ID, you can get it from `sts get-caller-identity`

rleighton:
> Instead of building a separate binary, why not just add a new mage target? `mage ops:requeue`. Then you don't have to worry about when to build the binary - it's built into the `mage` binary. You can't use command-line flags, but you can use environment variables as flags instead, which we already do in some places. So `FROM_Q=... TO_Q=... mage ops:requeue`. That's my preference
> 
> I agree that building an ops binary as part the deploy command is strange, and again, we want to avoid bloating the `deploy` command - it should only be minimal steps for actually getting Panther deployed and running in your account.
> 
> If we feel it's really important to have as a separate utility, consider writing it as a simple Python script (in `/scripts`) so that it doesn't have to be compiled at all.

I'd like the opstools to be outside of mage so they can be copied to other environments to run easily. This is a common need. I don't like Python for this because it has a run time that needs to exist. The nice thing about default Go compilation is you can copy static binaries where you need them. I've experienced problems in the past with pyhton, ruby AND perl ops tools that required a bunch of setup just to get things to run (or worse, the install env is not compatible with the scrupt). The last thing you need in an emergency. Copying a static binary to where it needs to be is so much nicer.

cmd/opstools/requeue/requeue.go
```diff
@@ -0,0 +1,90 @@
...
+	"strconv"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"github.com/aws/aws-sdk-go/service/sqs/sqsiface"
+	"github.com/pkg/errors"
+)
+
+const (
+	waitTimeSeconds   = 20
```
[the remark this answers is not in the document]
rleighton: Can't hurt. Done

cmd/opstools/requeue/integration_test.go
rleighton: Done

cmd/opstools/requeue/requeue/main.go
rleighton: Removed. Not sure I why I even added it.

cmd/opstools/requeue/requeue/main.go
rleighton: Done

cmd/opstools/requeue/requeue.go
rleighton: Done

tools/mage/deploy.go
austinbyers: > how do ensure it is already built for customers without making them run commands

Well, they already ran the `deploy` command, running `mage build:optools && ./out/bin/optools/requeue` seems reasonable. I would prefer to keep it out of deploy.

It can be built in the test suite to make sure it's correct, but I agree that shouldn't be the main mechanism.

In the Panther Github releases, we can distribute precompiled binaries if you think that would be useful.

nhakmiller: I like the `mage build:optools` option, and I especially like it combined with pre-compiled binaries. We could have a single zip file of binaries per platform, that way you can get the tools downloadable without even having to have git setup somewhere to clone the repo.

cmd/opstools/requeue/integration_test.go
rleighton: Done

tools/mage/deploy.go
rleighton: I took the build out of the deploy step and left `mage build:opstools` and updated the operations docs so folks know what to do.

I chatted with Austin and we think ultimately we should  put it all into mage and cross compile for majors archs and distribute. I think that would be ideal in the long run , but a another PR.

internal/compliance/aws_event_processor/processor/lambda.go
```diff
@@ -20,6 +20,7 @@ package processor
 
 import (
 	"regexp"
```
rleighton: This diff got in accidentally ... removing

cmd/opstools/requeue/requeue/main.go
kostaspap: Ah got it. But still, the `os.Exit(-2)` should not be needed.

cmd/opstools/requeue/requeue/main.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func validateFlags() {
+	var err error
+	defer func() {
+		if err != nil {
+			fmt.Printf("%s\n", err)
```
kostaspap: Should you use `log.Printf` to be aligned to how it's done in the rest of this file?

cmd/opstools/requeue/requeue/main.go
rleighton: When you write a cli you need to return a non-zero exit code for scripts to know there is an error.

kostaspap: Yup, `log.Fatal` already does it

cmd/opstools/requeue/requeue/main.go
rleighton: No, this is part of argument parsing, you do not want time stamps in the output help message.
This is how it looks now:
```
-from.q not set
requeue moves messages from one sqs queue to another
Usage:
  -from.q string
    	The name of the queue to copy from
  -region string
    	The AWS region where the queues exists (optional, defaults to session env vars)
  -to.q string
    	The name of the queue to copy to

```
this is how it would look using log:
```
2020/02/28 05:59:54 -from.q not set
requeue moves messages from one sqs queue to another
Usage:
  -from.q string
    	The name of the queue to copy from
  -region string
    	The AWS region where the queues exists (optional, defaults to session env vars)
  -to.q string
    	The name of the queue to copy to

```

kostaspap: True, thanks!

tools/mage/deploy.go
austinbyers: We don't have to worry about the cross-compilation (that's part of the release process), but I would prefer if this PR adds a `mage ops:requeue` instead of `mage build:opstools`

rleighton: So, for me an ops tool has 2 very important characteristics: 
1. quick to deploy to where you need it
2. safe (minimal functionality)

If I just add a mage target I get neither of these characteristics. In particular, we  have mage targets like deploy and tear-down make it very powerful. These are not capabilities you want in an ops tool.

If we build an executable we get both of these characteristics (static compiled binary, can move to what ever env needs it / has permissions) and minimal functionality (safe).

austinbyers: Then I misunderstood our previous offline conversation, carry on

Issue #313, opened 2020-02-27
Verification email when user email is changed


alexmylonas:
### Describe the bug

After #107 is implemented we 'll allow users to change emails without confirmation/verification
This could potentially lead to account deadlock.

### Steps to reproduce

1. Sign in to platform 
2. Change your email to a one you don't have access or misspell the new email
3. If you forget your password, you can't reset your password

### Expected behavior

A verification email to the new and/or old email before persist the new email

3nvi:
@austinbyers what's the status on this? Is it still valid based on our recent work on reset-password & verifications?

austinbyers:
It doesn't look like you can change your own email from the profile icon, but admins can still change user emails to anything they want, yes. So it's possible for an admin to change their own email, mistype it, and be locked out. A similar problem happens if you mistype the email when deploying.

I just don't know what we can reasonably do about it. We mark emails as "auto-verified" in Cognito, I have no idea how the user flow would change if we require emails to be verified or how much work it would be to implement.

We can leave this open as something to research someday, or we can close it as low priority and  not something we'll realistically have time to implement. Other admins can fix your email for you, or you can fix it from the users-api / Cognito directly

3nvi:
I think that we should keep it open. Meanwhile perhaps asking for the user's email 2 times (both during `mage deploy` and within the UI) can perhaps temporarily help reduce the cases of mistypings.

I know for sure that I mistyped my email once during `mage deploy` and I had to go through Cognito and lose some time

austinbyers:
This isn't really a bug, it's a feature - require email verification when updating user emails. As such, I've moved this to product board: https://runpanther.productboard.com/feature-board/1567477-feature-organization/features/5467897/detail

Until then, if you lock yourself out with the wrong email, an admin can reset it, or we have an `invite` opstool which can manually invite a new user for you (which will eventually be superseded by the real API)

PR #314, opened 2020-02-27
[Feature] Allow User Updates

### LINKED ISSUE #107: [Feature] Allow User Updates

alexmylonas:
## Background

New feature in Platform that allows admin users to edit user info (first name, last name, email).

It's implemented as a side-sheet that pops from the relevant option on user list.

Closes #107 
Follow-up task: #313 

## Changes

- Added new edit user form that displays only relevant fields
- Added new edit user sidesheet
- Added option on user list "Edit Profile"

## Testing

- Locally

### DIFF
--- a/web/src/components/forms/common/base-user-form.tsx
+++ b/web/src/components/forms/common/base-user-form.tsx
@@ -1,0 +1,99 @@
+/**
+ * Panther is a scalable, powerful, cloud-native SIEM written in Golang/React.
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import { Field, Formik } from 'formik';
+import SubmitButton from 'Components/submit-button';
+import { Flex } from 'pouncejs';
+import FormikTextInput from 'Components/fields/text-input';
+
+interface IdValue {
+  id: string;
+}
+
+export interface BaseUserFormProps<BaseUserFormValues extends IdValue> {
+  /** The initial values of the form */
+  initialValues: BaseUserFormValues;
+
+  /** callback for the submission of the form */
+  onSubmit: (values: BaseUserFormValues) => void;
+}
+
+function BaseUserForm<BaseUserFormValues extends IdValue>({
+  initialValues,
+  onSubmit,
+}: React.PropsWithChildren<BaseUserFormProps<BaseUserFormValues>>): React.ReactElement<
+  BaseUserFormProps<BaseUserFormValues>
+> {
+  return (
+    <Formik<BaseUserFormValues>
+      initialValues={initialValues}
+      onSubmit={onSubmit}
+      enableReinitialize
+    >
+      {({ handleSubmit, isSubmitting, isValid, dirty }) => {
+        return (
+          <form onSubmit={handleSubmit}>
+            <Field
+              as={FormikTextInput}
+              label="Email address"
+              placeholder="john@doe.com"
+              name="email"
+              aria-required
+              mb={3}
+            />
+            <Flex mb={6} justifyContent="space-between">
+              <Field
+                as={FormikTextInput}
+                label="First Name"
+                placeholder="John"
+                name="givenName"
+                aria-required
+              />
+              <Field
+                as={FormikTextInput}
+                label="Last Name"
+                placeholder="Doe"
+                name="familyName"
+                aria-required
+              />
+            </Flex>
+            <Flex
+              borderTop="1px solid"
+              borderColor="grey100"
+              pt={6}
+              mt={10}
+              justifyContent="flex-end"
+            >
+              <Flex>
+                <SubmitButton
+                  submitting={isSubmitting}
+                  disabled={!dirty || !isValid || isSubmitting}
+                >
+                  {initialValues.id ? 'Update' : 'Invite'}
+                </SubmitButton>
+              </Flex>
+            </Flex>
+          </form>
+        );
+      }}
+    </Formik>
+  );
+}
+
+export default BaseUserForm;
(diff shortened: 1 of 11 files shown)


web/src/components/forms/edit-user-form/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+const EditProfileForm: React.FC<EditProfileFormProps> = ({ onSuccess, user }) => {
+  const [editUser, { error: editUserError, data }] = useMutation<boolean, ApolloMutationInput>(
+    EDIT_USER
+  );
+  const { pushSnackbar } = useSnackbar();
+
+  React.useEffect(() => {
+    if (data) {
+      pushSnackbar({ variant: 'success', title: `Successfully edited user` });
+      // TODO: Refetch user if updating self
```
3nvi: this is a tricky one.
As you said we should make sure that we update the user stored in `auth-context` if you somehow update yourself. Can we add a method to do that (or even better, can we use an existing method?)

web/src/components/sidesheets/edit-user-sidesheet.tsx
```diff
@@ -0,0 +1,52 @@
...
+import EditUserForm from 'Components/forms/edit-user-form';
+import React from 'react';
+import useSidesheet from 'Hooks/useSidesheet';
+import { User } from 'Generated/schema';
+
+// const RESET_USER_PASS = gql`
+//   mutation ResetUserPassword($id: ID!) {
+//     resetUserPassword(id: $id)
+//   }
+// `;
```
kostaspap: Why are these lines commented out (also line 20)? If they are not used they can just be removed.

web/src/components/forms/edit-user-form/index.tsx
alexmylonas: Latest update, implements that and state seem updated properly -- local storage is not updated you can check if you refresh page after editing yourself, you will show up with the name before editing

web/src/components/forms/common/base-user-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+ */
+
+import React from 'react';
+import { Field, Formik } from 'formik';
+import SubmitButton from 'Components/submit-button';
+import { Flex } from 'pouncejs';
+import FormikTextInput from 'Components/fields/text-input';
+import * as Yup from 'yup';
+
+interface IdValue {
```
3nvi: `BaseUserFormValues` should include all the values that the form can accept:

```ts
interface BaseUserFormValues {
  id?: string // optional value
  email: string;
  familyName: string;
  givenName;
}
```

No need to extend `IdValue`. Is there a reason that I'm missing as to why you were doing that?

web/src/components/forms/common/base-user-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+const validationSchema = Yup.object().shape({
+  email: Yup.string()
+    .email('Must be a valid email')
+    .required('Email is required'),
+  familyName: Yup.string().required('Last name is required'),
+  givenName: Yup.string().required('First name is required'),
+});
+
+function BaseUserForm<BaseUserFormValues extends IdValue>({
```
3nvi: `BaseUserForm` can be written like:

```tsx
const BaseUserForm: React.FC<BaseUserFormProps> = ({ initialValues, onSubmit }) => ....
```

web/src/components/forms/common/base-user-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+import SubmitButton from 'Components/submit-button';
+import { Flex } from 'pouncejs';
+import FormikTextInput from 'Components/fields/text-input';
+import * as Yup from 'yup';
+
+interface IdValue {
+  id: string;
+}
+
+export interface BaseUserFormProps<BaseUserFormValues extends IdValue> {
```
3nvi: You don't need to add a generic here at all. `BaseUserFormProps` should use `BaseUserFormValues` without accepting them as a generic prop (since it's always gonna be the same thing)

web/src/components/forms/edit-user-form/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+interface ApolloMutationInput {
+  input: UpdateUserInput;
+}
+
+interface EditProfileFormProps {
+  onSuccess: () => void;
+  user: User;
+}
+
+interface EditProfileFormValues {
```
3nvi: This is not needed anywhere

web/src/components/forms/user-invitation-form/index.tsx
```diff
@@ -40,30 +37,17 @@ interface ApolloMutationInput {
...
-  familyName?: string;
-  givenName?: string;
-}
-
 interface UserInvitationFormProps {
   onSuccess: () => void;
 }
 
 const initialValues = {
+  id: null,
```
3nvi: don't add this at all. Let it be `undefined` :)

web/src/components/forms/edit-user-form/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const [editUser, { error: editUserError, data }] = useMutation<boolean, ApolloMutationInput>(
+    EDIT_USER
+  );
+  const { pushSnackbar } = useSnackbar();
+
+  React.useEffect(() => {
+    if (data) {
+      pushSnackbar({ variant: 'success', title: `Successfully edited user` });
+      // Refetch user info if editing self
+      if (user.id === userInfo.sub) getCurrentUserInfo({});
```
3nvi: we add brackets in `if` statements in this project :D

web/src/components/forms/edit-user-form/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+      onSuccess();
+    }
+  }, [data]);
+
+  const initialValues = {
+    id: user.id,
+    email: user.email || '',
```
3nvi: email is required so it will never be empty or undefined :) (you can remove the `||`)

web/src/components/forms/edit-user-form/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+}
+
+interface EditProfileFormValues {
+  id: string;
+  givenName: string;
+  familyName: string;
+  email: string;
+}
+
+const EditProfileForm: React.FC<EditProfileFormProps> = ({ onSuccess, user }) => {
```
3nvi: This component's code should be added to the `edit-user-sidesheet`. What you are defining here is not actually a form (since the form was defined previously). You are defining the callbacks, queries & updates which exist outside the context of the form itself.

If you feel strongly about this being a form we can keep everything as is though.

web/src/components/forms/common/base-user-form.tsx
```diff
@@ -0,0 +1,109 @@
+/**
```
3nvi: rename to `user-form` and all `BaseUserXXX` to `UserXXX`

web/src/components/forms/user-invitation-form/index.tsx
```diff
@@ -17,16 +17,13 @@
  */
```
3nvi: This code should now be moved to the `user-invitation-sidesheet`

Issue #272, opened 2020-02-25
Update log onboarding flow - backend


kostaspap:
### Description

This story contains all the features that have to be implemented in the backend to support the new log onboarding flow. 

### Acceptance Criteria
- Storing an integration will now prompt the backend to assume the appropriate role (LogProcessing) and try to verify that it has the appropriate S3 permissions (e.g. perform an S3 Object List operation)
- API method that accepts the parameters provided by the user and returns the CFN template that the users have to deploy to their account. Check with @3nvi to validate how this API should look like.

nhakmiller:
Didn't realize the full scope of this issue when I closed it, re-opening.

PR #318, opened 2020-02-28
Source Health Checks

### LINKED ISSUE #271: Update cloud security account onboarding flow - backend
### Description

This story contains all the features that have to be implemented in the backend to support the new security account onboarding flow. 

### Acceptance Criteria
- Storing an integration will now prompt the backend to assume the appropriate roles (Audit, AutoRemediation) and verify that the real-time event infrastructure is in place (CWE are setup and subscriptions are in place)
- API method that accepts the parameters provided by the user and returns the CFN template that the users have to deploy to their account. Check with @3nvi to validate how this API should look like.

nhakmiller:
## Background

When adding a new source, there is currently no feedback for the user about whether the source is correctly integrated or not. Even after adding the source, the only way to know if it is working is to try to observe data coming in. This PR adds a source health check endpoint to the backend to be made use of by the frontend in a future PR.

There are a few minor issues in the PR still, but I wanted to get it out in front of the frontend team so they could take a look and give feedback on the proposed format.

Closes #271 and #272.

## Changes

- Rename the `panther-snapshot-api` to the `panther-source-api`, as `panther-snapshot-api` was really not indicative of its function
  - **THIS IS A BREAKING CHANGE, ALL SOURCES WILL NEED TO BE RE-ONBOARDED**
  - Updated the documentation and metrics to reflect this change **THE MONITORING STACK WILL NEED TO BE DELETED THEN REDEPLOYED**
- Moved the `panther-source-api` from the compliance tree to the core tree, as it is a service that manages both policies and rules.
- Add a `CheckIntegration` endpoint for the new `panther-source-api` that does a few basic checks to determine if a new source is in a good state or not. Should be easily extensible in the future for more advanced checks

## Testing

- Deployed to dev account and manually invoked lambda function
  - Also tested that the existing functionality of the `source-api` was preserved after being re-named
- Unit & integration tests

Example usage:

Checking the status of a potential `aws-scan` source integration.

Input:
```
{
  "integrationHealthCheck": {
    "AWSAccountId": "123456789012",
    "IntegrationType": "aws-scan",
    "EnableCWESetup": true,
    "EnableRemediation": true
  }
}
```

Output (CWE role is NOT setup properly in this example account):
```
{
  "awsAccountId": "123456789012",
  "auditRoleStatus": {
    "healthy": true,
    "errorMessage": null
  },
  "cweRoleStatus": {
    "healthy": false,
    "errorMessage": "AccessDenied: User: arn:aws:sts::123456789012:assumed-role/panther-app-SourceApi-1/panther-source-api is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::123456789012:role/PantherCloudFormationStackSetExecutionRole\n\tstatus code: 403, request id: 111"
  },
  "remediationRoleStatus": {
    "healthy": true,
    "errorMessage": null
  },
  "processingRoleStatus": null,
  "s3BucketsStatus": {},
  "kmsKeysStatus":{}
}
```

Checking the status of a potential `aws-s3` source integration.

Input:
```
{
  "integrationHealthCheck": {
    "AWSAccountId": "857418155548",
    "IntegrationType": "aws-s3",
    "S3Buckets": [
      "bucket-1",
      "bucket-2"
    ],
    "KMSKeys": [
    "arn:aws:kms:us-west-2:123456789012:key/1",
    "arn:aws:kms:us-west-2:123456789012:key/2"
    ]
  }
}
```

Output (`bucket-2` does not have the proper permissions granted):
```
{
  "awsAccountId": "123456789012",
  "integrationType": "aws-s3",
  "auditRoleStatus": null,
  "cweRoleStatus": null,
  "remediationRoleStatus": null,
  "processingRoleStatus": {
    "healthy": true,
    "errorMessage": null
  },
  "s3BucketsStatus": {
    "bucket-2": {
      "healthy": false,
      "errorMessage": "AccessDenied: Access Denied\n\tstatus code: 403, request id: 111, host id: 111/222="
    },
    "bucket-1": {
      "healthy": true,
      "errorMessage": null
    }
  },
  "kmsKeysStatus": {
    "arn:aws:kms:us-west-2:123456789012:key/1": {
      "healthy": true,
      "errorMessage": null
    },
    "arn:aws:kms:us-west-2:123456789012:key/2": {
      "healthy": false,
      "errorMessage": "NotFoundException: Key 'arn:aws:kms:us-west-2:123456789012:key/2' does not exist"
    }
  }
}
```


api/lambda/source/models/api.go
```diff
@@ -33,6 +35,24 @@ type LambdaInput struct {
...
+	IntegrationType *string `json:"integrationType" validate:"required,oneof=aws-scan aws-s3"`
+
+	// Checks for cloudsec integrations
+	EnableCWESetup    *bool `json:"enableCWESetup"`
+	EnableRemediation *bool `json:"enableRemediation"`
+
+	// Checks for log analysis integrations
+	S3Buckets []*string `json:"s3Buckets"`
+	KmsKeys   []*string `json:"kmsKeys"`
+}
```
nhakmiller: @3nvi @alexmylonas , this is the input struct you'll need to construct to send to the backend when making this request. Does this look ok to you? It is in the same pattern of the other endpoints in this API, one function/input shared between both `aws-scan` and `aws-s3` resources, and you specify the `integrationType` field to show which is which.

api/lambda/source/models/api.go
```diff
@@ -33,6 +35,24 @@ type LambdaInput struct {
...
+	AWSAccountID    *string `genericapi:"redact" json:"awsAccountId" validate:"required,len=12,numeric"`
+	IntegrationType *string `json:"integrationType" validate:"required,oneof=aws-scan aws-s3"`
+
+	// Checks for cloudsec integrations
+	EnableCWESetup    *bool `json:"enableCWESetup"`
+	EnableRemediation *bool `json:"enableRemediation"`
+
+	// Checks for log analysis integrations
+	S3Buckets []*string `json:"s3Buckets"`
+	KmsKeys   []*string `json:"kmsKeys"`
```
nhakmiller: @kostaspap @rleighton we currently don't actually have a way to test these on the backend, the only permission granted by the log processing role is `Decrypt` and we don't even know if there will be data to decrypt when testing this. One alternative suggested by @jacknagz is to update the log processing role to also have `kms:DescribeKey` or something similar so that we can try to describe the keys and see if they exist that way.

deployments/core/source_api.yml
```diff
@@ -0,0 +1,167 @@
...
+        - Id: AssumePantherAuditRoles
+          Version: 2012-10-17
+          Statement:
+            - Effect: Allow
+              Action: sts:AssumeRole
+              Resource:
+                - !Sub arn:${AWS::Partition}:iam::*:role/${AuditRoleName}
+                - !Sub arn:${AWS::Partition}:iam::*:role/${RemediationRoleName}
+                - !Sub arn:${AWS::Partition}:iam::*:role/${CWERoleName}
+                - !Sub arn:${AWS::Partition}:iam::*:role/${LogProcessingRoleName}
```
nhakmiller: I was slightly concerned about concentrating so many permissions into one lambda function. An alternative approach I considered was only granting access to the `AuditRole` and the `LogProcessingRole`, and using the `AuditRole` to describe the `RemediateRole` and the `CWERole` to see if they exist. That mean the lambda only has read access in the other accounts, not write access.

I decided to go with the assuming role method for now because it is more direct and I can't think of any way this lambda could be exploited right now, but I am open to other suggestions.

api/lambda/source/models/api.go
kostaspap: I like the idea of adding permissions for `kms:DescribeKey` and using that - I don't think it's too intrusive.

deployments/core/source_api.yml
```diff
@@ -0,0 +1,167 @@
...
+      # <cfndoc>
+      # The `panther-source-api` lambda manages Cloud Security and Log Analysis sources. This includes
+      # creating, testing, updating, listing, and deleting sources.
+      #
+      # Failure Impact
+      # * Failure of this lambda will prevent sources from being manageable, and will interrupt daily scans.
+      # </cfndoc>
+      Handler: main
+      Layers: !If [AttachLayers, !Ref LayerVersionArns, !Ref 'AWS::NoValue']
+      MemorySize: 512
```
kostaspap: This Lambda is dealing with small amount of data and doing small amount of processing, I think you could give it the minimum memory (128). Same for timeout, most operations seems trivial, could be lower (e.g. 60). 
I'm just nitpicking here, just sharing thoughts I had going through this code. Take it or leave it 😄

deployments/core/source_api.yml
```diff
@@ -0,0 +1,167 @@
...
+  LogProcessingRoleName:
+    Type: String
+    Default: PantherLogProcessingRole
+    Description: The name of the IAM role to assume to perform AWS account/service scans
+  LogAnalysisQueueName:
+    Type: String
+    Default: panther-input-data-notifications-queue
+  SnapshotPollerQueueName:
+    Type: String
+    Default: panther-snapshot-queue
```
kostaspap: I think you could save some space in the template and just remove this ones? The role names are anyway hardcoded in the templates customers use for onboarding (and also in the code). Queue names are also not going to change that frequently.

nhakmiller: Sounds good, I was mostly just extending the pattern that was used previously for the `LogAnalysisQueueName` but it's a lot of overhead for something that doesn't really add any value.

deployments/core/source_api.yml
nhakmiller: Fair enough, I'll drop these down.

deployments/compliance/snapshot.yml
```diff
@@ -261,7 +171,7 @@ Resources:
             Schedule: rate(24 hours)
       FunctionName: panther-snapshot-scheduler
       # <cfndoc>
-      # The `panther-snapshot-scheduler` lambda calls the `panther-snapshot-api` for infrastructure scans triggered
+      # The `panther-snapshot-scheduler` lambda calls the `panther-source-api` for infrastructure scans triggered
```
rleighton: Is this comment correct? Maybe elaborate why the panther-snapshot-scheduler would call the panther-source-api

internal/compliance/aws_event_processor/processor/accounts.go
```diff
@@ -28,14 +28,14 @@ import (
...
-	"github.com/panther-labs/panther/api/lambda/snapshot/models"
+	"github.com/panther-labs/panther/api/lambda/source/models"
 	"github.com/panther-labs/panther/pkg/genericapi"
 )
 
 const (
 	cweAccountTimeout       = 15 * time.Minute
 	refreshInterval         = 2 * time.Minute
-	snapshotAPIFunctionName = "panther-snapshot-api"
+	snapshotAPIFunctionName = "panther-source-api"
```
rleighton: s/snapshotAPIFunctionName /sourceAPIFunctionName

internal/compliance/snapshot_scheduler/scheduler/schedule.go
```diff
@@ -27,12 +27,12 @@ import (
...
 
-	"github.com/panther-labs/panther/api/lambda/snapshot/models"
-	snapshotapi "github.com/panther-labs/panther/internal/compliance/snapshot_api/api"
+	"github.com/panther-labs/panther/api/lambda/source/models"
+	snapshotapi "github.com/panther-labs/panther/internal/core/source_api/api"
 	"github.com/panther-labs/panther/pkg/genericapi"
 )
 
-const snapshotAPIFunctionName = "panther-snapshot-api"
+const snapshotAPIFunctionName = "panther-source-api"
```
rleighton: s/snapshotAPIFunctionName/sourceAPIFunctionName

internal/core/source_api/api/check_integration.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return bucketStatuses
+}
+
+func getCredentialsWithStatus(
+	roleARN *string,
+) (*credentials.Credentials, *models.SourceIntegrationItemStatus) {
+
+	zap.L().Info("checking role", zap.String("roleArn", *roleARN))
```
rleighton: I would do Debug()

deployments/compliance/snapshot.yml
nhakmiller: Good call, updated.

internal/compliance/aws_event_processor/processor/accounts.go
nhakmiller: updated.

internal/compliance/snapshot_scheduler/scheduler/schedule.go
nhakmiller: updated.

internal/core/source_api/api/check_integration.go
nhakmiller: Updated.

PR #320, opened 2020-02-28
Syslog parser for RFC3164 formatted logs

stoggi:
## Background

This PR introduces a new Parser for Syslog RFC3164 formatted logs as planned in #200.

## Changes

* Created new parser for RFC3164 format (BSD syslog messages)
* Updated parser documentation for syslog.
* Added `Syslog.RFC3164` as a whitelisted log type for the frontend UI.
* Added `sysloglogs.RFC3164Parser` to the log registry.
* Added examples from https://tools.ietf.org/html/rfc3164#section-5.4 as unit tests.

## Testing

- Unit tests

Waiting on #304 and #261 to merge before merging from master and running an end-to-end test.

### DIFF
--- a/internal/log_analysis/log_processor/parsers/sysloglogs/rfc3164.go
+++ b/internal/log_analysis/log_processor/parsers/sysloglogs/rfc3164.go
@@ -19,9 +19,11 @@ package sysloglogs
  */
 
 import (
+	"errors"
 	"net"
 	"time"
 
+	"github.com/influxdata/go-syslog/v3"
 	"github.com/influxdata/go-syslog/v3/rfc3164"
 	"go.uber.org/zap"
 
@@ -49,22 +51,29 @@ type RFC3164 struct {
 }
 
 // RFC3164Parser parses Syslog logs in the RFC3164 format
-type RFC3164Parser struct{}
+type RFC3164Parser struct {
+	parser syslog.Machine
+}
 
+// New returns an initialized LogParser for Syslog RFC3164 logs
 func (p *RFC3164Parser) New() parsers.LogParser {
-	return &RFC3164Parser{}
+	return &RFC3164Parser{
+		parser: rfc3164.NewParser(
+			rfc3164.WithBestEffort(),
+			rfc3164.WithTimezone(time.UTC),
+			rfc3164.WithYear(rfc3164.CurrentYear{}),
+			rfc3164.WithRFC3339(),
+		),
+	}
 }
 
 // Parse returns the parsed events or nil if parsing failed
 func (p *RFC3164Parser) Parse(log string) []interface{} {
-	parser := rfc3164.NewParser(
-		rfc3164.WithBestEffort(),
-		rfc3164.WithTimezone(time.UTC),
-		rfc3164.WithYear(rfc3164.CurrentYear{}),
-		rfc3164.WithRFC3339(),
-	)
-
-	msg, err := parser.Parse([]byte(log))
+	if p.parser == nil {
+		zap.L().Debug("failed to parse log", zap.Error(errors.New("parser can not be nil")))
+		return nil
+	}
+	msg, err := p.parser.Parse([]byte(log))
 	if err != nil {
 		zap.L().Debug("failed to parse log", zap.Error(err))
 		return nil
--- a/internal/log_analysis/log_processor/parsers/sysloglogs/rfc3164.go
+++ b/internal/log_analysis/log_processor/parsers/sysloglogs/rfc3164.go
@@ -1,0 +1,114 @@
+package sysloglogs
+
+/**
+ * Panther is a scalable, powerful, cloud-native SIEM written in Golang/React.
+ * Copyright (C) 2020 Panther Labs Inc
+ *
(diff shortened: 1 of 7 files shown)


internal/log_analysis/log_processor/parsers/sysloglogs/rfc3164.go
```diff
@@ -0,0 +1,114 @@
...
+// RFC3164Parser parses Syslog logs in the RFC3164 format
+type RFC3164Parser struct{}
+
+func (p *RFC3164Parser) New() parsers.LogParser {
+	return &RFC3164Parser{}
+}
+
+// Parse returns the parsed events or nil if parsing failed
+func (p *RFC3164Parser) Parse(log string) []interface{} {
+	parser := rfc3164.NewParser(
```
rleighton: If this can be re-used for each event, then make this is the New() method  which is called once per file by the classifier and add to struct (probably want to do for the other syslog parser you did):
```
func (p *RFC3164Parser) New() parsers.LogParser {
	return &RFC3164Parser{
          parser: parser := rfc3164.NewParser(
rfc3164.WithBestEffort(),
		rfc3164.WithTimezone(time.UTC),
		rfc3164.WithYear(rfc3164.CurrentYear{}),
		rfc3164.WithRFC3339(),
	)
        }
}
```

rleighton:
> This looks almost the same as #261 for `RFC5424`. The only difference I see is that the other one has a required Version and optional StructuredData
> This looks like a really small difference. I am thinking that both of them could actually be the same parser - thus the data from each will end up in the same table in Athena. Is there more that I might be missing?
> 
> @nhakmiller @jacknagz and @rleighton thoughts?

Since we are using a 3rd party parser and that makes a distinction, it might complicate things.  That said, if it does not make a big difference to the code then I would be okay with "unioning" these and updating the doc accordingly. If it make things messy , then I would leave it.

kostaspap:
> > This looks almost the same as #261 for `RFC5424`. The only difference I see is that the other one has a required Version and optional StructuredData
> > This looks like a really small difference. I am thinking that both of them could actually be the same parser - thus the data from each will end up in the same table in Athena. Is there more that I might be missing?
> > @nhakmiller @jacknagz and @rleighton thoughts?
> 
> Since we are using a 3rd party parser and that makes a distinction, it might complicate things. That said, if it does not make a big difference to the code then I would be okay with "unioning" these and updating the doc accordingly. If it make things messy , then I would leave it.

My main worry is that Glue Tables are a generally "expensive" resource - not in terms of $$ or maintenance but rather when it comes to querying. Do we really want `RFC3164` to be stored separately from `RFC5424` and people to query each table separately to get their data? In my eyes they are very similar and it might make sense to be stored in the same table.

stoggi:
> > > This looks almost the same as #261 for `RFC5424`. The only difference I see is that the other one has a required Version and optional StructuredData
> > > This looks like a really small difference. I am thinking that both of them could actually be the same parser - thus the data from each will end up in the same table in Athena. Is there more that I might be missing?
> > > @nhakmiller @jacknagz and @rleighton thoughts?
> > 
> > 
> > Since we are using a 3rd party parser and that makes a distinction, it might complicate things. That said, if it does not make a big difference to the code then I would be okay with "unioning" these and updating the doc accordingly. If it make things messy , then I would leave it.
> 
> My main worry is that Glue Tables are a generally "expensive" resource - not in terms of $$ or maintenance but rather when it comes to querying. Do we really want `RFC3164` to be stored separately from `RFC5424` and people to query each table separately to get their data? In my eyes they are very similar and it might make sense to be stored in the same table.

The outputs are only similar because we are using the go-syslog parser, and it is making the output consistent. We could try to parse as RFC5424, and if that fails try RFC3164, but then we are just duplicating classifier functionality in a parser. If you keep them separate then at least the user has the choice of which syslog format they are expecting for their rule (or both if they want).

rleighton:
> > > > This looks almost the same as #261 for `RFC5424`. The only difference I see is that the other one has a required Version and optional StructuredData
> > > > This looks like a really small difference. I am thinking that both of them could actually be the same parser - thus the data from each will end up in the same table in Athena. Is there more that I might be missing?
> > > > @nhakmiller @jacknagz and @rleighton thoughts?
> > > 
> > > 
> > > Since we are using a 3rd party parser and that makes a distinction, it might complicate things. That said, if it does not make a big difference to the code then I would be okay with "unioning" these and updating the doc accordingly. If it make things messy , then I would leave it.
> > 
> > 
> > My main worry is that Glue Tables are a generally "expensive" resource - not in terms of $$ or maintenance but rather when it comes to querying. Do we really want `RFC3164` to be stored separately from `RFC5424` and people to query each table separately to get their data? In my eyes they are very similar and it might make sense to be stored in the same table.
> 
> The outputs are only similar because we are using the go-syslog parser, and it is making the output consistent. We could try to parse as RFC5424, and if that fails try RFC3164, but then we are just duplicating classifier functionality in a parser. If you keep them separate then at least the user has the choice of which syslog format they are expecting for their rule (or both if they want).

I would say keep them separate right now.

PR #321, opened 2020-02-28
Update lint-staged scripts

3nvi:
## Background

Since git hooks are not opt-in (disabled by default), it's time to cleanup the code of the git hook to something simpler and quicker

## Changes

- Update tasks on git hooks

## Testing

- Through commits

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,17 @@
+# General
+.setup
+keys
+
+# Deps
+venv
+node_modules
+
+# Service artifacts
+out
+
+# Web artifacts
+web/dist
+
+# Docs
+docs/gitbook
+
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,6 @@
 # General
+.git/
+.idea
 .setup
 keys
 
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,5 @@
 # General
-.git/
+.git
 .idea
 .setup
 keys
--- a/.prettierignore
+++ b/.prettierignore
@@ -19,5 +19,4 @@ docs/gitbook
 
 
 # Direct file content assertion
-tools/cfngen/cloudwatchcf/testdata
-tools/cfngen/gluecf/testdata
+tools/**/testdata
--- a/.prettierignore
+++ b/.prettierignore
@@ -17,3 +17,7 @@ web/dist
 # Docs
 docs/gitbook
 
+
+# Direct file content assertion
+tools/cfngen/cloudwatchcf/testdata
+tools/cfngen/gluecf/testdata
--- a/lint-staged.config.js
+++ b/lint-staged.config.js
@@ -23,14 +23,9 @@ module.exports = {
    */
 
   /*
-   * Run prettier on:
-   * 1. All JSON/YAML inside the /web dir
-   * 2. All JSON/YAML found in the root project dir
-   * 3. All TS, JS and Markdown files found anywhere in the project
+   * Run prettier TS, JS, JSON, YAML and Markdown files found anywhere in the project
    */
-  'web/**/*.{json,yml}': ['prettier --write', 'git add'],
-  './*.{json,yml}': ['prettier --write', 'git add'],
-  '*.{ts,tsx,js,md}': ['prettier --write', 'git add'],
+  '*.{ts,tsx,js,md,yaml,yml,json}': ['prettier --write', 'git add'],
 
   /*
    * Run ESLint checks for all TS & JS files found anywhere in hte project
--- a/lint-staged.config.js
+++ b/lint-staged.config.js
@@ -35,5 +35,5 @@ module.exports = {
   /*
    * only run the TS compiler when there are changes inTS files
    */
-  '**/*.ts?(x)': () => 'tsc -p .',
+  '*.ts?(x)': () => 'tsc -p .',
 };
--- a/tools/mage/fmt.go
+++ b/tools/mage/fmt.go
@@ -58,12 +58,12 @@ func Fmt() {
 	}
 
 	// prettier (web)
-	args = []string{"--write", "{web/src/**,web/__generated__,.}/*.{ts,js,tsx,md,json,yml}"}
+	args = []string{"--write", "**/*.{ts,js,tsx,md,json,yaml,yml}"}
 	if !mg.Verbose() {
 		args = append(args, "--loglevel", "error")
 	}
 	if err := sh.Run(nodePath("prettier"), args...); err != nil {
-		logger.Fatalf("failed to format {web/src/**,.}: %v", err)
+		logger.Fatalf("failed to format: %v", err)
 	}
 
 	// Generate documentation
--- a/tools/mage/fmt.go
+++ b/tools/mage/fmt.go
@@ -47,23 +47,14 @@ func Fmt() {
 		logger.Fatalf("failed to format python: %v", err)
 	}
 
-	// prettier (cloudformation)
+	// prettier
 	logger.Info("fmt: prettier")
-	args = []string{"--write", "deployments/**.yml"}
-	if !mg.Verbose() {
-		args = append(args, "--loglevel", "error")
-	}
-	if err := sh.Run(nodePath("prettier"), args...); err != nil {
-		logger.Fatalf("failed to format deployments/**.yml: %v", err)
-	}
-
-	// prettier (web)
 	args = []string{"--write", "**/*.{ts,js,tsx,md,json,yaml,yml}"}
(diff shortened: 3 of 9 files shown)


lint-staged.config.js
```diff
@@ -23,14 +23,9 @@ module.exports = {
    */
 
   /*
-   * Run prettier on:
-   * 1. All JSON/YAML inside the /web dir
-   * 2. All JSON/YAML found in the root project dir
-   * 3. All TS, JS and Markdown files found anywhere in the project
+   * Run prettier TS, JS, JSON, YAML and Markdown files found anywhere in the project
```
nhakmiller: Seems reasonable, but be aware you may break CI if you check in changes to JSON/YAML files used as part of unit tests!

lint-staged.config.js
```diff
@@ -23,14 +23,9 @@ module.exports = {
...
    */
-  'web/**/*.{json,yml}': ['prettier --write', 'git add'],
-  './*.{json,yml}': ['prettier --write', 'git add'],
-  '*.{ts,tsx,js,md}': ['prettier --write', 'git add'],
+  '*.{ts,tsx,js,md,yaml,yml,json}': ['prettier --write', 'git add'],
```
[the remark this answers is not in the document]
austinbyers: Actually, it looks like it was the github issues it didn't like. You can try it, see if it works

But either way, this should also be consistent with `mage fmt`: https://github.com/panther-labs/panther/blob/master/tools/mage/fmt.go#L60-L67

So we should do one of the following:

- Change this to call `mage fmt`
    - If that's too slow for you, we can add a `mage fmtweb` command to do this
- Or update `mage fmt` source to make sure it does the same thing you're doing in your hook

lint-staged.config.js
3nvi: Right now, these hooks are only run for people who opt in by the way

lint-staged.config.js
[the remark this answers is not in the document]
3nvi: As mentioned in the email, those rules run for checked out files and **not** the entire codebase. That's why they run the prettier changes and then they `git add`, so that the file changes from prettier can be checked out.

In short, what it does is append the list of checked out filenames after the command. For example, if we are trying to commit the files `/web/src/index.tsx` and `/web/src/codegen.yml`, this lib will run the following:

```
prettier --write web/src/index.tsx web/src/codegen.yml
git add web/src/index.tsx web/src/codegen.yml
```

@austinbyers can we make it so that if `mage fmt` accepts a list of files, then it runs its code only for those files?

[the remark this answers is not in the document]
austinbyers: I just want to make sure there aren't different formatting standards for those that have the githook and those who do not. Whether the hook applies to the entire codebase or just checked out files doesn't matter - the fact remains that two engineers who apply the same change may format code differently depending on their environment.

We're already calling `prettier` as part of `mage fmt`, I'm just suggesting we change that prettier command to match the one you propose in the githook. The easiest way to do that is to just copy your setup into mage as well, I can do that in a subsequent PR if you prefer.

However, are you sure `*.{ts,tsx,js,md,yaml,yml,json}` is the right syntax you want? It seems like that would match only top-level files:

```
% node_modules/.bin/prettier --write *.{ts,tsx,js,md,yaml,yml,json}
zsh: no matches found: *.ts
% node_modules/.bin/prettier --write **/*.{ts,tsx,js,md,yaml,yml,json}
zsh: no matches found: **/*.yaml
% node_modules/.bin/prettier --write **/*.{ts,tsx,js,md,yml,json}
zsh: argument list too long: node_modules/.bin/prettier
```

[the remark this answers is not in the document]
3nvi: Funnily enough no. Lint-staged uses [micromatch](https://www.npmjs.com/package/micromatch) and has a [different version of filtering](https://github.com/okonet/lint-staged#filtering-files) where `*.js` matches **every** javascript file.

I understand your concern, so I'm proposing the following:

* Have prettier  run against `'*.{ts,tsx,js,md,yaml,yml,json}'` in both `lint-staged` and `mage fmt`
* Create a [.prettierignore](https://prettier.io/docs/en/ignore.html#ignoring-files) file to force prettier to **not run** in certain files or certain directories (e.g. tests)
* Instruct engineers to use  ignore comments (i.e. `// prettier-ignore` or `# prettier-ignore`) to instruct prettier to not run for a particular block or JSON blob, when the related file is not included in the `.prettierignore` file.

I feel that this would:

* Solve all of our issues
* Allow prettier to automatically format files in all crucial directories without complicated blobs 
* Reduce manual effort by automatically formatting new files in new directories without us having to manually include those directories in both `lint-staged` and `mage fmt`.

Thoughts?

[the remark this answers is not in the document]
austinbyers: I'm fine with running prettier on everything, even the tests (like you say, we can just fix the tests if something fails). I didn't realize it had great ignore controls, works for me.

It turns out it's hard to make `mage fmt` run prettier on everything because there are too many files (see the output from my previous comment). So we can't run exactly the same command in `mage` you do here (also, different syntax, as you point out)

So I can just change `mage fmt` later to cover all of the repo, probably by invoking `prettier` multiple times. You can do it now if you like, but not needed. Your PR looks good, already approved

3nvi:
@austinbyers Updated the PR to add ignore file for key locations. Let me know if that solves your issue of running prettier against the whole repo. If so, I can go ahead and merge it

.prettierignore
```diff
@@ -0,0 +1,19 @@
+# General
+.git/
```
austinbyers: Any reason this has a trailing slash but the other directories do not? I don't think we need it, let's be consistent

tools/mage/fmt.go
```diff
@@ -58,12 +58,12 @@ func Fmt() {
 	}
 
 	// prettier (web)
-	args = []string{"--write", "{web/src/**,web/__generated__,.}/*.{ts,js,tsx,md,json,yml}"}
+	args = []string{"--write", "**/*.{ts,js,tsx,md,json,yaml,yml}"}
```
austinbyers: Thanks! And you tried `mage fmt` and it worked? Super awesome

Now you can delete the block above this (where we were running prettier separately on CloudFormation)

tools/mage/fmt.go
```diff
@@ -58,12 +58,12 @@ func Fmt() {
...
 	if !mg.Verbose() {
 		args = append(args, "--loglevel", "error")
 	}
 	if err := sh.Run(nodePath("prettier"), args...); err != nil {
-		logger.Fatalf("failed to format {web/src/**,.}: %v", err)
+		logger.Fatalf("failed to format: %v", err)
```
austinbyers: there's a lot of formatting here in this command, let's say `failed to format with prettier: %v`

.prettierignore
3nvi: hmmm, initially I thought that if I didn't add it it would also ignore `.gitignore` and `.github/*` files. Now that I think about it, I think that I would need to write `.git*` for that to occur

3nvi:
@austinbyers @rleighton Now that `mage fmt` runs `prettier` against the entire repo, tests started failing. Specifically, that has to do with the files present in:

* `/tools/mage/testdata/
* `/tools/cfngen/gluecf/testdata/`
* `/tools/cfngen/cloudwatchcf/testdata/`

The reason is that the `.json` and `.yml` files were reformatted with the correct indentations (as well as removal of empty intermediate lines), which caused tests to fail, since the tests were making assertions against the `bytes[]` of the read file, instead of comparing the actual JSON/YAML content.

Personally I feel that this is not correct, since a test shouldn't fail if in a JSON file the indentation of a line is  3 spaces instead of 4. Nor if the YAML has two empty lines instead of one. What should matter is the content. If the JSON blob is the same, then the test should pass. Thus, I feel that the way the test assertions are being made right now are not optimal. Instead of using `assert.Equal` we should opt for something like `assert.JSONEq` and no longer parse & compare a JSON or YAML file as bytes. This is something that involves changes in the utils that you guys are using so it's not something that was a quick change for me to fix.

 Currently I'm forcing `prettier` to ignore all `/tools/**/testdata/` content, but the underlying issue is frailty in the tests. We can either make a ticket or choose to ignore all of this together.

Let me know your thoughts.

P.S. I'm merging this current PR as is, since any potential test change is outside the scope of this PR

austinbyers:
> Personally I feel that this is not correct, since a test shouldn't fail if in a JSON file the indentation of a line is 3 spaces instead of 4.

Agreed, we can just adjust the tests. Either by changing the `[]byte` comparison to the new format or comparing the actual structure. I'll make an issue for it, add it the list of `mage` things I work on :)

PR #323, opened 2020-02-28
Handle function versions in aws-event-processor lambda dispatch

rleighton:
## Before
Lambda has versioned function names. Before this change they were enumerated (incompletely) and keeping the list up to date was difficult, with associated warnings appearing in monitoring.

## After 
Versions are kept in a list, and event names pre-processed to strip off the versions before dispatch.

## Changes
- Added getLambdaBaseEventName() + test

## Testing
mage test:ci
ran in dev env


internal/compliance/aws_event_processor/processor/lambda.go
```diff
@@ -79,9 +84,32 @@ func classifyLambda(detail gjson.Result, metadata *CloudTrailMetadata) []*resour
...
+
+// lambda has a number of "sets" of versioned event names. We do not care about the specific versions so strip off.
+var lambdaVersions = []string{
+	"20170331",
+	"20170331v2",
+	"20150331",
+	"20150331v2",
+}
+
+func init() {
```
nhakmiller: Sorry if I'm being dumb, does this function get called anywhere? Is this one of those magic functions that gets called automatically or something?

rleighton: Magic. Called on program load and you can have as many as you want in a package:https://golang.org/doc/effective_go.html#init

PR #327, opened 2020-02-28
Resource Lookup

### LINKED ISSUE #278: Resource Lookups
### Description

Policies currently run against a single resource, and with only the context of that resource. Some configurations are intrinsically related to multiple resources, which makes evaluating them difficult. If a policy being executed against one resource was able to lookup another resource for context, this would enable a lot of more powerful use cases.

Example: We want to write a policy that checks we are using CMKs for all S3 buckets that are storing CloudTrail logs.

Current way: Currently, this could be implemented by writing a policy that evaluates all S3 buckets, with an in code whitelist of just the buckets that are being used by CloudTrail. This has two major drawbacks. First of all, this list of buckets would have to be managed manually and would quickly get out of sync in large organizations. Secondly, the configuration being checked is fundamentally about how CloudTrail data is stored, S3 is just the medium. Intuitively we would expect to write such a policy to evaluate CloudTrails, not S3 buckets.

With resource lookups: We would write a policy that evaluates CloudTrail resources, and performs a resource lookup based on the `S3BucketName` field, then evaluates the returned resource. This is more intuitive, and requires no manual management of lists.

### Acceptance Criteria

- While a policy is being run against a given resource, it is able to query Panther to get the contents of another resource
- This lookup should be easy to do in code
- Ideally, this lookup should consider both cost and time implications as it could be made potentially millions of times a day

### Future considerations

It should be noted that in the case above, when the S3 bucket is updated the corresponding CloudTrail will NOT be re-evaluated until the next daily scan (potentially up to 24 hours later). In general this is true, changes to linked resources will not necessarily be immediately reflected in the root resource and so will not be considered.

A future improvement (which I may make another story for, but will NOT be in scope for v1) could create these backlinks dynamically whenever one resource looks up another resource. The back links could expire after 24 hours, and get hooked into the resources-api in some way so that whenever a resource is evaluated, all of it's backlinked resources are evaluated as well.

nhakmiller:
## Background

This PR adds support for the resource lookup functionality described in and closes #278.

This functionality is accomplished by allowing the `panther-policy-engine` to make dynamodb GetItem API calls to the `panther-resources` table, and adding a helper function the the `panther-policy-engine` to do the lookup for users. This helper function is injected into `sys.modules` with the name `panther` so that the user written policies can import it. Alternatively, users can choose to import `boto3` directly and create a dynamodb client to talk to the `panther-resources` table.

We decided to talk to the `panther-resources` table directly instead of the API gateway or `panther-resources-api` lambda because we wanted to minimize time spent waiting for a response.

## Changes

- Gave `dynamodb:GetItem` permissions to the `panther-policy-engine` to talk to the `panther-resources` table
- Created the `helpers` file in the `panther-policy-engine`. For now I'm thinking of putting all helper functions in this one file, in the future we can do a proper package with different sub packages if we end up adding lots of helper functions

## Testing

- Deployed to my dev account and tested the following policy:

```
from panther import resource_lookup

def policy(resource):
    bucketArn = 'arn:aws:s3:::' + resource['S3BucketName']
    bucket = resource_lookup(bucketArn)
    for encryption_rule in bucket['EncryptionRules'] or []:
        if encryption_rule.get('ApplyServerSideEncryptionByDefault', False):
            return encryption_rule['ApplyServerSideEncryptionByDefault']['SSEAlgorithm'] is not None
    return False
```


internal/compliance/policy_engine/src/helpers.py
```diff
@@ -0,0 +1,52 @@
...
+import boto3
+
+
+class BadLookup(Exception):
+    """Error returned when a resource lookup fails."""
+
+
+def resource_lookup(resource_id: str) -> Dict[str, Any]:
+    """This function is used to get a resource from the resources-api based on its resourceID."""
+    lambda_client = boto3.client('lambda')
```
kostaspap: You could declare the `lambda_client` once and just re-use it, it should be more efficient.

internal/compliance/policy_engine/src/helpers.py
```diff
@@ -0,0 +1,52 @@
...
+        },
+    }
+
+    # Invoke the resources-api
+    response = lambda_client.invoke(
+        FunctionName='panther-resources-api', InvocationType='RequestResponse', LogType='None', Payload=json.dumps(request_payload)
+    )
+
+    # These responses are small, so we just load the whole thing into memory
+    response_payload = json.loads(response['Payload'].read())
```
kostaspap: According to Boto3 documentation, the response is a dict of the form: 
```
{
    'StatusCode': 123,
    'FunctionError': 'string',
    'LogResult': 'string',
    'Payload': StreamingBody(),
    'ExecutedVersion': 'string'
}
```

You should check the `StatusCode` or `FunctionError` to make sure there was no issue invoking the Lambda before loading the Lambda's response - `Payload`

internal/compliance/policy_engine/src/helpers.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+"""Utility functions provided to policies during execution."""
+import json
+from typing import Any, Dict
+import boto3
+
+
+class BadLookup(Exception):
```
kostaspap: Unit tests would be useful here.

internal/compliance/policy_engine/src/helpers.py
nhakmiller: I was never able to get `StatusCode` to be anything besides 200 no matter what input I sent, I think we set it up to always return ok for the top level response. I guess if lambda itself is down? I guess I can check and just raise a generic exception.

nhakmiller:
> 3\. I'm trying to think if it makes sense to display a different interface to the users. Do you have an example of a policy that will use the resource lookup? Looking in example in the description, I don't think that's a real-world example - the lookup in this case is not needed since the resource is already provided in the `resource` method argument?

Just to answer point 3 here, this policy is an example of the used case described in detail in the linked issue (#278), where the resource being evaluated is a CloudTrail resource. So it is very much a real world example.

kostaspap:
> > 3. I'm trying to think if it makes sense to display a different interface to the users. Do you have an example of a policy that will use the resource lookup? Looking in example in the description, I don't think that's a real-world example - the lookup in this case is not needed since the resource is already provided in the `resource` method argument?
> 
> Just to answer point 3 here, this policy is an example of the used case described in detail in the linked issue (#278), where the resource being evaluated is a CloudTrail resource. So it is very much a real world example.

Ah ok, you are right. Not sure why I misinterpreted the code. Thanks for the clarification!
So in general the suggestion to the users would be to pass the ARN of the resource as a parameter correct? Just double-checking that the resource_id is actually equivalent the resource ARN (and "custom" arn for resources that don't include ARNs)

nhakmiller:
> Ah ok, you are right. Not sure why I misinterpreted the code. Thanks for the clarification!
> So in general the suggestion to the users would be to pass the ARN of the resource as a parameter correct? Just double-checking that the resource_id is actually equivalent the resource ARN (and "custom" arn for resources that don't include ARNs)

Right, they need to pass in the Panther defined `resourceId`, which is documented for each resource type in the docs. In general, it is either the resource's ARN or an ARN-like string for resources that don't have ARNs.

I was thinking of also providing some helper functions that would help users construct these ARNs, but I wasn't sure what that would look like beyond just having them provide us with an account ID, service,  region, resource type, and resource name, at which point they've practically constructed the ARN themselves already.

For a few resources it would be simpler, especially s3. I suppose we could do a `get_linked_resource` where the user passes in the full resource they received, then specifies a resource type and field with the resource name? Then we can lookup the region and account ID from the passed in resource directly. That might cover a lot of use cases.

internal/compliance/policy_engine/src/helpers.py
austinbyers: If I remember right, this is actually a bug with the `boto3` library (where the status is always the same). Worth checking, but you're right that it will probably never be useful unless they fix it

deployments/compliance/resources_api.yml
```diff
@@ -315,6 +315,14 @@ Resources:
...
       Timeout: 120
       Tracing: !If [TracingEnabled, !Ref TracingMode, !Ref 'AWS::NoValue']
+      Policies:
+        - Id: ResourceLookup
+          Version: 2012-10-17
+          Statement:
+            - Effect: Allow
+              Action:
+                - lambda:InvokeFunction
+              Resource: !GetAtt HandlerFunction.Arn
```
austinbyers: I agree with the decision to go straight to Lambda. We don't need the overhead and complexity of talking to API gateway.

As for whether to talk directly to Dynamo, I could go either way. I think your approach (Lambda) strikes a good balance of easy implementation and performance, we can always change it to talk to Dynamo in the future if performance becomes an issue

internal/compliance/policy_engine/src/policy.py
```diff
@@ -76,6 +78,12 @@ def __init__(self, policies: List[Dict[str, Any]]) -> None:
         self._policies_by_type: Dict[str, List[Policy]] = collections.defaultdict(list)
         self._global_policies: List[Policy] = []  # List of policies that apply to all log types
 
+        # Setup panther helper functions
+        helpers_spec = import_util.spec_from_file_location(helpers.__name__, helpers.__file__)
+        helpers_mod = import_util.module_from_spec(helpers_spec)
+        sys.modules['panther_helpers'] = helpers_mod
+        helpers_spec.loader.exec_module(helpers_mod)  # type: ignore
```
austinbyers: We have a static method for this very purpose:

```suggestion
sys.modules['panther_helpers'] = Policy.import_module(helpers.__name__, helpers.__file__)
```

Also consider naming this library just `panther`. It reads a little cleaner in the policy:

```python
import panther

def policy(resource):
    panther.resource_lookup(...)
```

deployments/compliance/resources_api.yml
nhakmiller: I decided to go with dynamo because it did seem easier after all with the thing that Kostas suggested.

internal/compliance/policy_engine/src/helpers.py
nhakmiller: Added.

PR #329, opened 2020-03-02
Update endpoints return modified settings

### LINKED ISSUE #324: Integrations should return new values when updated
### Describe the bug

Currently, when an integration is updated (through the `snapshot-api`), nothing is returned back (apart from a boolean denoting whether there was an error).

When I update an integration, it's really helpful to get the new values of this field.

### Steps to reproduce

Call `updateIntegrationSettings` with some new values and see its return value

### Expected behavior

Call `updateIntegrationSettings` should return an object with the updated information

### Additional Context

We need to do add:

```golang
ReturnValues:     aws.String("ALL_NEW"),
```
to dynamo's `updateItem` method and propagate the return value down the chain. Lastly we need to update the tests.

### LINKED ISSUE #325: updateUser should return the updated user's values
Describe the bug
Currently, when a user is updated (through the `users-api`), nothing is returned back (apart from a boolean denoting whether there was an error).

When I update a user, it's really helpful to get the new values in order to correctly update the table without a full re-fetch.

Call `updateUser` with some new values and see its return value

Expected behavior
Calling `updateUser` should return an object with the updated information

Additional Context

Unless `AdminUpdateUserAttributesOutput` returns all the necessary information, this may involve re-fetching the updated user from Cognito once the update is applied

### LINKED ISSUE #326: Improve lambda methods' return values
### Description

This epic contains issues that are going to act as a necessary pre-requisite for snappier UI, faster visual updates, correct addition of optimistic responses and better API schema.

Resolving these tickets will allow the UI to alter its code in order to create a better UX for the end user

austinbyers:
## Background
Fixes `UpdateUser` and `UpdateIntegration` endpoints to return the modified settings.

Closes #324 
Closes #325 
Closes #326

## Changes

- `users-api`
    - `UpdateUser` returns the new user settings. The Cognito update API does not have this information, so we just get it in the backend before returning.
    - Removes some redundant structs
- `source-api`
    - All of the update endpoints return the full integration model in their response
    - bugfix: `ScanEnabled` in `PutIntegration` is optional, but if omitted the API panics with a nil pointer dereference
    - bugfix: `UpdateIntegrationSettingsInput` includes an `AWSAccountId` field, but that field is completely ignored in the backend. It doesn't make sense to update the account ID of an existing source, so this field has been removed.
    - bugfix: The integration label is required when updating an integration but not when creating one. Changed to optional in both cases
- `mage fmt`
    - Remove unused references to `enterprise` submodule (which we will no longer be using)

## Testing

- Added integration test verification of settings update
- Deployed to dev account, invoked Lambdas directly to verify output.

Update user:

```json
{
  "createdAt": 1583174295,
  "email": "@runpanther.io",
  "familyName": "Byers",
  "givenName": "Austin",
  "id": "fce79740-d805-4121-97a5-8c31f2dc72b1",
  "status": "CONFIRMED"
}
```

Update integration:

```json
{
  "awsAccountId": "101802775469",
  "createdAtTime": "2020-03-02T20:52:51.051204342Z",
  "createdBy": "fce79740-d805-4121-97a5-8c31f2dc72b1",
  "integrationId": "ef168c2c-38ae-4141-8c6e-ff7a1a362dc7",
  "integrationLabel": "DisabledTest",
  "scanEnabled": false,
  "scanIntervalMins": null,
  "s3Buckets": [],
  "kmsKeys": []
}
```

### DIFF
--- a/api/lambda/source/models/api.go
+++ b/api/lambda/source/models/api.go
@@ -116,9 +116,8 @@ type UpdateIntegrationLastScanEndInput struct {
 
 // UpdateIntegrationSettingsInput is used to update integration settings.
 type UpdateIntegrationSettingsInput struct {
-	AWSAccountID     *string   `genericapi:"redact" json:"awsAccountId,omitempty" validate:"omitempty,len=12,numeric"`
 	IntegrationID    *string   `json:"integrationId" validate:"required,uuid4"`
-	IntegrationLabel *string   `json:"integrationLabel" validate:"min=1"`
+	IntegrationLabel *string   `json:"integrationLabel,omitempty" validate:"omitempty,min=1"`
 	ScanEnabled      *bool     `json:"scanEnabled"`
 	ScanIntervalMins *int      `json:"scanIntervalMins" validate:"omitempty,oneof=60 180 360 720 1440"`
 	S3Buckets        []*string `json:"s3Buckets"`
--- a/api/lambda/users/models/api.go
+++ b/api/lambda/users/models/api.go
@@ -77,3 +77,6 @@ type UpdateUserInput struct {
 	FamilyName *string `json:"familyName" validate:"omitempty,min=1"`
 	Email      *string `json:"email" validate:"omitempty,min=1"`
 }
+
+// UpdateUserOutput returns the new Panther user details.
+type UpdateUserOutput = User
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,7 @@ require (
 	golang.org/x/lint v0.0.0-20200130185559-910be7a94367 // indirect
 	golang.org/x/net v0.0.0-20200226121028-0de0cce0169b // indirect
 	golang.org/x/sys v0.0.0-20200223170610-d5e6a3e2c0ae // indirect
-	golang.org/x/tools v0.0.0-20200226205201-eb7c56241bdb // indirect
+	golang.org/x/tools v0.0.0-20200302155637-b1e4e04173e0 // indirect
 	gopkg.in/go-playground/assert.v1 v1.2.1 // indirect
 	gopkg.in/go-playground/validator.v9 v9.31.0
 	gopkg.in/russross/blackfriday.v2 v2.0.0 // indirect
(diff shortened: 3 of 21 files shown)


api/lambda/source/models/api.go
```diff
@@ -116,9 +116,8 @@ type UpdateIntegrationLastScanEndInput struct {
 
 // UpdateIntegrationSettingsInput is used to update integration settings.
 type UpdateIntegrationSettingsInput struct {
-	AWSAccountID     *string   `genericapi:"redact" json:"awsAccountId,omitempty" validate:"omitempty,len=12,numeric"`
 	IntegrationID    *string   `json:"integrationId" validate:"required,uuid4"`
-	IntegrationLabel *string   `json:"integrationLabel" validate:"min=1"`
+	IntegrationLabel *string   `json:"integrationLabel,omitempty" validate:"omitempty,min=1"`
 	ScanEnabled      *bool     `json:"scanEnabled"`
 	ScanIntervalMins *int      `json:"scanIntervalMins" validate:"omitempty,oneof=60 180 360 720 1440"`
 	S3Buckets        []*string `json:"s3Buckets"`
```
nhakmiller: Perhaps I'm not fully understanding the purpose of the `omitempty` tag, but S3Buckets (and KmsKeys) will not always be present for a source integration, could those be omit empty as well?

austinbyers: `json:omitempty` just determines whether the field will be included in the response.

My guess is the frontend always prefers an empty list to null if the type is supposed to be a list. So leaving it out forces the response to be `S3Buckets: []` instead of omitting the field completely. Not sure if @kostaspap had some other reason for it

kostaspap: > My guess is the frontend always prefers an empty list to null if the type is supposed to be a list. So leaving it out forces the response to be S3Buckets: [] instead of omitting the field completely. Not sure if @kostaspap had some other reason for it

That's pretty much it. I remember @3nvi  has mentioned that it was preferred for him to return empty lists instead of nil.

Issue #135, opened 2020-02-07
mage teardown


austinbyers:
**Is your feature request related to a problem? Please describe.**
Tearing down a Panther deployment requires several manual steps before the CloudFormation stacks can be deleted:

- Empty all panther S3 buckets
- Teardown the web cluster
- Delete panther-app and panther-app-frontend stacks
- Delete panther-buckets stack

Sometimes, CloudWatch log groups are also not deleted even when the stack is deleted, because leftover Lambda invocations are still writing logs. In this case, the next deployment will fail because the log group already exists.

Also, self-signed ACM certificates are created manually and will remain even after deleting the CloudFormation stacks.

**Describe the solution you'd like**
Add a `mage teardown` command that will, after user confirmation, orchestrate the deletion of all Panther resources

**Describe alternatives you've considered**
Doing it by hand is annoying and wastes developer cycles when we inevitably forget one of the steps

kostaspap:
Is this really a problem? Should we really spend developer cycle for this? 

In my personal experience, the problem was never the tearing down the infrastructure was slow. The biggest problem that comes with tearing down the infrastructure is that we then need to setup everything once again. Deploying the stack could take several minutes, then it takes long time to setup compliance for accounts, log processing for log sources and destinations.

austinbyers:
> Is this really a problem? Should we really spend developer cycle for this?

It's not high priority at the moment (I didn't mean to assign it to the v0.3 milestone), but absolutely worth doing. It will only take maybe a day of work. @3nvi and @nhakmiller have both asked for such a feature, and I spend ~30 minutes a day tearing down deployments. (I'm constantly tearing down and redeploying)

Plus, it guarantees that a user can completely remove a Panther deployment. There are artifacts right now that are left behind even if the CloudFormation stacks are deleted (ACM certificates, log groups)

nhakmiller:
There some artifacts that are left because they are not controlled by CloudFormation (ACM Certificates), and some that are left because the are controlled by CFN but are protected against deletion (S3 buckets), and some that left because they are controlled by CFN and can be deleted but after CFN deletes them they get referenced and then re-created which CFN doesn't know about (log groups). Due to the complexity here, I think it is worth a dedicated command. I agree this is not a high priority (not needed before V1), but I think it is needed eventually because it:

1. Provides developers with an easier way to test fresh deployments
2. Provides customers with an easier way to "uninstall" panther if they decide they don't want to use it

[...]

Issue #143, opened 2020-02-07
Better error handling for failed deployments


austinbyers:
**Is your feature request related to a problem? Please describe.**
The user experience is pretty bad when a deployment fails:

- The returned error message doesn't have any information other than ROLLBACK_COMPLETE
- You have to login to the AWS console, find the deleted stack, look through many error messages to find the relevant one
- Trying to deploy Panther again will fail because the stack has to be deleted first

**Describe the solution you'd like**
When a deployment fails:

- Automatically delete the stack if it's ROLLBACK_COMPLETE (you can still see deleted stack events in the console)
- Find the relevant error and print it for the user

**Additional context**
We've considered refactoring the CloudFormation into a few top-level stacks instead of ~1 huge stack with lots of nesting. One advantage of having a few top-level stacks is that an error late in the deploy process doesn't force CloudFormation to revert the entire monolithic stack. The rollback is faster and also the retry is faster

austinbyers:
See #118 as an example of trying to troubleshoot a failed deployment

PR #330, opened 2020-03-03
Mage Teardown

austinbyers:
## Background
Provides a `mage teardown` command to completely remove deployed Panther infrastructure

I considered calling it `mage destroy`, but "destroy" is too similar to "deploy" in my opinion

Closes: #135 

## Changes

- Add AppSync log group to CloudFormation, which will give it the same retention and teardown as our other log groups
- Add `mage teardown` command
    1. Prompts the user for confirmation
    1. Removes ECR repo + ECS cluster
    1. Removes CFN stacks in parallel
    1. Removes S3 buckets + self-signed ACM certs + leftover CloudWatch log groups
        - A 1-day expiration policy is applied to large S3 buckets instead of deleting objects manually
- When creating an IAM server cert, set the path "/panther/" to have a second attribute we can verify when deleting
- Add AWS region to destructive confirmations
- Add `DeletionPolicy` (and `UpdateReplacePolicy`) to all S3 buckets - this way the CloudFormation stacks can be deleted even if the buckets can't be deleted yet
- Remove the exports from the first `panther-buckets` stack - exports force sequential CFN operations. Removing these makes it possible to teardown all stacks in parallel and is more consistent with the rest of our infra - we lookup stack outputs and pass them around instead of relying on exports.

## Testing

- Deploy + teardown (teardown took ~15 minutes)

```
15:01:33	WARN	Teardown will destroy all Panther infrastructure in account 101802775469 (us-east-2)
Are you sure you want to continue? (yes|no) yes
15:01:35	INFO	scanning Panther CloudFormation stacks
15:01:50	INFO	removing ECR repository panther-web
15:01:50	INFO	deleting s3://panther-app-analysisapi-rml7o59a-analysisversions-1w21vb3ubqv0e
15:01:52	INFO	deleting s3://panther-app-athenaresults-1bahaa96bnd73
15:01:53	INFO	deleting s3://panther-app-processeddata-17dfopusc37u8
15:01:54	INFO	deleting s3://panther-buckets-accesslogs-1bd49lpu7q0yx
15:01:54	INFO	deleting s3://panther-buckets-source-1hhpst0qucp6i
15:01:55	INFO	deleting CloudFormation stack: panther-app-frontend
15:03:57	INFO	    √ panther-app-frontend successfully deleted
15:03:57	INFO	deleting CloudFormation stacks: panther-app, panther-app-monitoring, panther-app-databases, panther-buckets, panther-app-onboard
15:04:28	INFO	    √ panther-app-onboard successfully deleted
15:04:28	INFO	    √ panther-app-databases successfully deleted
15:04:28	INFO	    √ panther-buckets successfully deleted
15:07:59	INFO	    √ panther-app-monitoring successfully deleted
15:16:04	INFO	    √ panther-app successfully deleted
15:16:04	INFO	deleting CloudWatch log group /aws/lambda/panther-remediation-processor
15:16:04	INFO	deleting CloudWatch log group /aws/lambda/panther-users-api
15:16:04	INFO	deleting ACM cert arn:aws:acm:us-east-2:101802775469:certificate/202f8b11-6aad-4f75-83c0-4520b4574495
15:16:05	INFO	successfully removed Panther infrastructure
```

- Teardown again (with no infrastructure) - ensures that there are no errors when removing resources which have already been deleted
- Tested with max bucket size = 1 to verify the lifecycle policy is added

### Migration Notes
Existing deployments will need to make some small adjustments before they can apply this change:

1. Delete `/aws/appsync/apis/` CloudWatch log groups manually
    - Otherwise, CFN will try to create a log group which already exists and fail
1. Temporarily restore the `panther-backend` stack exports, but keep the other stacks updated with these changes
1. Deploy
    - This will ensure that references to the export are removed before the export itself
1. Remove the `panther-backend` stack exports


tools/mage/teardown.go
```diff
@@ -0,0 +1,312 @@
...
+	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/sts"
+
+	"github.com/panther-labs/panther/pkg/testutils"
+)
+
+const (
+	ecrRepoName    = "panther-web"
```
3nvi: Although they are most likely to remain the same, I just want to point out that all these are AWS template exports. We already have a way of storing those in a file in `.env.aws` and we can do something similar for these as well (i.e. store the cluster, repo & service names in a `.env.XXX` file).

Then you can simply read them using the `dotenv` package instead of hardcoding their names. Just a suggestion, although I'm cool eitherway.

tools/mage/teardown.go
```diff
@@ -0,0 +1,312 @@
...
+	identity, err := sts.New(awsSession).GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	if err != nil {
+		logger.Fatalf("failed to get caller identity: %v", err)
+	}
+
+	logger.Warnf("THIS WILL DESTROY ALL PANTHER INFRASTRUCTURE IN AWS ACCOUNT %s (%s)",
+		*identity.Account, *awsSession.Config.Region)
+	result := promptUser("Are you sure you want to continue? (yes|no) ", nonemptyValidator)
+	if strings.ToLower(result) != "yes" {
+		logger.Fatal("permission denied: teardown canceled")
```
3nvi: I feel that `permission denied` isn't the correct phrase to use here, since it implies that "we tried and failed since we didn't have permissions to delete your Panther installation".

Should we change it to something more explicit like "Teardown aborted" ?

tools/mage/teardown.go
```diff
@@ -0,0 +1,312 @@
...
+// Remove the ECR repo and all of its images
+func destroyEcrRepo(awsSession *session.Session) {
+	logger.Infof("removing ECR repository %s", ecrRepoName)
+	client := ecr.New(awsSession)
+	_, err := client.DeleteRepository(&ecr.DeleteRepositoryInput{
+		Force:          aws.Bool(true), // remove images as well
+		RepositoryName: aws.String(ecrRepoName),
+	})
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "RepositoryNotFoundException" {
```
3nvi: when can this occur? Did you see any weird behaviors? If so, we may need to first delete all the images and then delete the repo

tools/mage/teardown.go
```diff
@@ -0,0 +1,312 @@
...
+	client := ecs.New(awsSession)
+
+	logger.Infof("stopping ECS service %s in cluster %s", ecsServiceName, ecsClusterName)
+	_, err := client.DeleteService(&ecs.DeleteServiceInput{
+		Cluster: aws.String(ecsClusterName),
+		Force:   aws.Bool(true), // stop running tasks as well
+		Service: aws.String(ecsServiceName),
+	})
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ClusterNotFoundException" {
```
3nvi: again, why do we have this check? Should we perhaps stop the tasks first, take down the service and then delete it?

tools/mage/teardown.go
```diff
@@ -0,0 +1,312 @@
...
+		Services: []*string{aws.String(ecsServiceName)},
+	}
+	if err = client.WaitUntilServicesInactive(waitInput); err != nil {
+		logger.Fatalf("ECS service %s did not stop successfully: %v", ecsServiceName, err)
+	}
+
+	// Now the cluster itself can be deleted
+	logger.Infof("deleting ECS cluster %s", ecsClusterName)
+	if _, err = client.DeleteCluster(&ecs.DeleteClusterInput{Cluster: aws.String(ecsClusterName)}); err != nil {
+		logger.Fatalf("failed to delete ECS cluster %s: %v", ecsClusterName, err)
```
3nvi: Is the deletion of the Load balancer & Networking stacks, part of `destroyCfnStacks`

tools/mage/test_namespace.go
```diff
@@ -302,7 +302,8 @@ func (t Test) Integration() {
 		logger.Fatalf("failed to get caller identity: %v", err)
 	}
 
-	logger.Warnf("INTEGRATION TESTS WILL ERASE ALL PANTHER DATA IN AWS ACCOUNT %s", *identity.Account)
+	logger.Warnf("INTEGRATION TESTS WILL ERASE ALL PANTHER DATA IN AWS ACCOUNT %s (%s)",
```
3nvi: Instead of adding caps, why don't we use our nice library that changes the color of the text?

tools/mage/teardown.go
```diff
@@ -0,0 +1,312 @@
...
+	awsSession, err := getSession()
+	if err != nil {
+		logger.Fatal(err)
+	}
+	identity, err := sts.New(awsSession).GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	if err != nil {
+		logger.Fatalf("failed to get caller identity: %v", err)
+	}
+
+	logger.Warnf("THIS WILL DESTROY ALL PANTHER INFRASTRUCTURE IN AWS ACCOUNT %s (%s)",
```
3nvi: Instead of adding caps, why don't we use our nice library that changes the color of the text? A bright red would be good (or white text with a red background)

tools/mage/teardown.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+
+		if !foundTag {
+			logger.Warnf("skipping s3 bucket %s: no 'Application=Panther' tag found", *bucket.Name)
+		}
+	}
+}
+
+// Empty, then delete the given S3 bucket
+func removeBucket(awsSession *session.Session, client *s3.S3, bucketName *string) {
```
rleighton: This is very likely to timeout or never finish on large buckets. I have had to add a life cycle rule and wait a day to delete big buckets in the past.

Could you give people the option to leave the buckets and tear down the stack anyway? I know CF in the console allows this.

kostaspap: Yup, also expensive. It's probably better to just change  the S3 buckets that contains the processed data to "DeletePolicy: Retain" and maybe during teardown just update the expiration policy. 
Also deleting objects costs $$, letting them expire is for free.

tools/mage/teardown.go
austinbyers: We hardcode resource names like this all over the place in `mage`, it's not a big deal. It's simpler and easier to understand than trying to read them from a file somewhere

tools/mage/teardown.go
austinbyers: Sounds good, I'll update

tools/mage/teardown.go
austinbyers: This happens if you run `mage teardown` on a deployment which has already been partially or completely deleted. Suppose, for example, your first `teardown` attempt succeeds at removing S3 + ECR + ECS, but fails deleting one of the CFN stacks. Then if you run `teardown` again, it should skip the already deleted stuff.

That's what this check is - if we're trying to delete an ECR repo that was already deleted, it's not an error and we can just skip it. CloudFormation itself works the same way

tools/mage/teardown.go
austinbyers: Same comment as above - this is to ensure that resources which already successfully deleted are skipped.

When we take down the service, we do so "forcefully", which stops the tasks as well. The cluster cannot be deleted until the service is inactive.

We have to stop the service manually (and wait for it to go inactive), but I can let CFN delete the cluster itself instead of doing it here

tools/mage/teardown.go
austinbyers: Yes, all of the infrastructure surrounding the ECS cluster is defined in one or more CloudFormation templates which are then destroyed when the stacks are destroyed

tools/mage/test_namespace.go
austinbyers: The [color](https://github.com/fatih/color) library you added, while popular, is deprecated:

> This project is not maintained anymore and is archived

Moreover, the logging library already adds color to the log level (in this case, the WARN will be printed in yellow), so I think just using the logging library is sufficient. You're right, I don't think the caps are necessary, I'll remove those

tools/mage/teardown.go
austinbyers: See comment above - the WARN is already printed in yellow, and I agree we don't need the all caps

tools/mage/teardown.go
austinbyers: > It's probably better to just change the S3 bucket that contains the processed data to "DeletePolicy: Retain" and maybe during teardown just update the expiration policy. Or just let the buckets be.

The teardown command has essentially 2 usecases:
- Developers regularly deploying + tearing down
- Users trying it out for a little while, then deleting it

In both cases, there will not be much data and we want all of the infrastructure completely destroyed - this is the "uninstall" command. If we just set a bucket lifecycle, the bucket itself will persist. I deploy 5-15 times a week, I don't want dozens of unused buckets cluttering my account. This is by far the most common scenario and I think we should leave this as the default behavior - production deployments with massive amounts of data aren't likely to teardown their infrastructure.

But I see the problem - there could be large buckets, even if someone does a trial run. How about this: we start deleting the items, like normal, but when we reach some threshold (10k or so), we stop, log a warning for the user, add an expiration policy, and then skip that bucket when deleting the stacks (there is a `RetainResources` field in `DeleteStacksInput`). This way, the most common scenario is handled ideally - all infra is deleted - but if there's lots of data we stop early and let S3 delete the objects later (and the bucket stays until the user manually deletes it).

austinbyers:
@3nvi 

> I only reviewed my part of the infrastructure. If I understand it correctly, we split deletion into two phases:
> 
> Delete manually stuff that are blocking the deletion of CFN stacks
> Delete CFN stacks

There's actually a third "phase" at the end - some resources have to be deleted *after* the CloudFormation stacks are destroyed (ACM certificates and leftover CloudWatch log groups). ACM certificates are not managed in CFN and log groups are sometimes re-created from buffered Lambda logs even after the stacks are destroyed.

> I feel that we can limit the manual work even more. For example, the ECR Repo can be automatically deleted by the destroyCfnStacks call if we just empty it aforehand (instead of force deleting it with an API call).

My goal is to minimize the API calls we have to do in `mage`. Here, we forcefully delete the repo with a single `Delete` call (which empties it as well). If we instead emptied it manually (there is no `EmptyRepo` API call), we would have to list all of the images and then delete each image individually, which is far more API calls. It's faster and simpler to let ECR delete the images for us.

> Same goes for some of the ECS related stuff, we don't need to manually destroy them (unless there is some sort of conflict with the emptying of the ECR that I'm not currently seeing).

You're right that the ECS cluster itself can be deleted with CFN, but I thought we had to manually stop the service first? I'll try leaving that out and see if it works.

> I feel that I've had to delete our installations multiple times and the main steps were to:
> 
> Empty S3
> Empty ECR
> The rest was handled automatically by CFN. Can I get some more context on the issues faced? I feel that I might be missing something...

The only extra steps we have besides S3 + ECR + CFN are:
* ECS (which I'll see if I can leave out)
* ACM certs - if you go to ACM, I bet you will see lots of "example.com" certs that were never deleted :) They're not managed in CloudFormation
* log groups - sometimes these are written after the stacks are deleted. It's very annoying when it happens, but you may have never encountered it

tools/mage/test_namespace.go
austinbyers: We also can't do both the color and the logging, it's one or the other

austinbyers:
Interestingly, it looks like CloudFormation can indeed delete the ECS cluster without any manual intervention, but the logs do show an error temporarily:

Screen Shot 2020-03-03 at 1 30 11 PM

I guess CFN tries again and the error resolves itself automatically? Works for me!

tools/mage/teardown.go
austinbyers: > we start deleting the items, like normal, but when we reach some threshold (10k or so), we stop, log a warning for the user, add an expiration policy

This is what I wound up implementing. I tried to delete the stacks excluding just the large S3 buckets, but that proved to be unnecessarily complicated. Instead, I just added `DeletionPolicy: Retain` to all S3 buckets so CFN wasn't blocked by them

austinbyers:
@3nvi @kostaspap comments addressed, ready for re-review

tools/mage/teardown.go
```diff
@@ -0,0 +1,367 @@
...
+
+// Empty, then delete the given S3 bucket.
+//
+// Or, if there are too many objects to delete directly, set an expiration lifecycle policy instead.
+func removeBucket(client *s3.S3, bucketName *string) {
+	input := &s3.ListObjectVersionsInput{Bucket: bucketName}
+	var objectVersions []*s3.ObjectIdentifier
+
+	// List all object versions (including delete markers)
+	err := client.ListObjectVersionsPages(input, func(page *s3.ListObjectVersionsOutput, lastPage bool) bool {
```
kostaspap: Interesting, so you don't need to list objects - just to list object versions? And this will return all objects, even the ones that don't have multiple versions? 

I was checking the [AWS example](https://docs.aws.amazon.com/AmazonS3/latest/dev/delete-or-empty-bucket.html#delete-bucket-sdk-java) on how to empty a bucket and they were listing objects AND listing object versions. If you tested this and it works, all good :)

tools/mage/teardown.go
```diff
@@ -0,0 +1,367 @@
...
+			}
+			return
+		}
+	}
+
+	// This cert is not tagged with Panther, ignore it
+}
+
+// Destroy any leftover /aws/lambda/panther- log groups
+func destroyLogGroups(awsSession *session.Session) {
```
kostaspap: It is possible that some customer will have other resources with prefix "panther" running in that account. If they decide to expand Panther functionality with a Lambda and they happen to prefix it with "panther" this will delete it. Is that likely? I wouldn't say it's very likely... nevertheless it is something that bugs me :P

I was thinking that a full-proof method would be to just get the list of resources we want to delete by  by listing the resources in the the CFN stacks (we know their names - they are actually hardcoded in our mage). See : https://docs.aws.amazon.com/cli/latest/reference/cloudformation/list-stack-resources.html

Then you can get e.g. the list of buckets by filtering for "AWS::S3::Bucket" ResourceType etc.
I know this requires some refactoring of the current process but maybe it's not such a big change in the end? Not sure how you think of it.

tools/mage/teardown.go
3nvi: My question is why don't we let CFN just delete the entire ECS stack. After ECR is emptied, then ECS services,tasks & clusters can be taken down through CFN alone. That's how I do it when I manually teardown a deployment.

Did it present any issues?

tools/mage/teardown.go
```diff
@@ -0,0 +1,367 @@
...
+
+	if _, err = client.DeleteBucket(&s3.DeleteBucketInput{Bucket: bucketName}); err != nil {
+		logger.Fatalf("failed to delete bucket %s: %v", *bucketName, err)
+	}
+}
+
+// Destroy self-signed Panther ACM certificates.
+//
+// Only certs for "example.com" tagged with "Application:Panther" and not currently in use will be deleted.
+func destroyAcmCerts(awsSession *session.Session) {
```
nhakmiller: When we fail to create ACM certificates, we fall back to creating IAM certificates: https://github.com/panther-labs/panther/blob/master/tools/mage/deploy_certificate.go#L176

Unfortunately, IAM server certificate resources cannot be tagged, but I think it would be safe to just list them with `list-server-certificates`, then delete all the ones with `PantherCertificate` as a prefix.

tools/mage/test_namespace.go
```diff
@@ -302,10 +302,11 @@ func (t Test) Integration() {
 		logger.Fatalf("failed to get caller identity: %v", err)
 	}
 
-	logger.Warnf("INTEGRATION TESTS WILL ERASE ALL PANTHER DATA IN AWS ACCOUNT %s", *identity.Account)
+	logger.Warnf("Integration tests will erase all Panther data in account %s (%s)",
```
rleighton: I hope we fix this one day.

tools/mage/teardown.go
austinbyers: > Interesting, so you don't need to list objects - just to list object versions? And this will return all objects, even the ones that don't have multiple versions?

Correct, because in a versioned bucket all objects have at least one version - their current one.

> I was checking the AWS example on how to empty a bucket and they were listing objects AND listing object versions.

That AWS example shows how to empty a nonversioned and a versioned bucket separately. The first code block is for nonversioned buckets only, the comment right above it reads:

> // To delete objects from versioned buckets, delete all of the object versions before deleting
> // the bucket (see below for an example).

> If you tested this and it works, all good :)
Yes, this is nearly the same code as `testutils.ClearS3Bucket` that we've been using in integration tests for many months. It works :)

tools/mage/teardown.go
austinbyers: > It is possible that some customer will have other resources with prefix "panther" running in that account. If they decide to expand Panther functionality with a Lambda and they happen to prefix it with "panther" this will delete it. Is that likely? I wouldn't say it's very likely... nevertheless it is something that bugs me :P

It is possible and I agree we should be very careful when deleting resources - in every other case (S3 buckets, certificates), I'm explicitly verifying the`Application = Panther` tag to double-check. Unfortunately, log groups are not tagged. If they were, what we have now would probably be sufficient.

> I was thinking that a full-proof method would be to just get the list of resources we want to delete by by listing the resources in the the CFN stacks https://docs.aws.amazon.com/cli/latest/reference/cloudformation/list-stack-resources.html

Agreed: that is the safest way to proceed, but its implementation is more complex than you suggest - we have to find and enumerate all of the nested stacks as well. And some resources (ACM certificates) are not managed with CloudFormation at all so their deletion will have to remain as is.

> I know this requires some refactoring of the current process but maybe it's not such a big change in the end? 

It is a big change - please try to suggest such large refactors in the _first_ round of reviews instead of the second. But I will implement it - I was planning on doing a similar resource traversal for #143 when troubleshooting a failed deployment, so this work may help with that also.

tools/mage/test_namespace.go
3nvi: Sounds good

tools/mage/teardown.go
```diff
@@ -0,0 +1,439 @@
...
+				// repo doesn't exist - that's fine, nothing to do here
+				continue
+			}
+			logger.Fatalf("failed to delete ECR repository: %v", err)
+		}
+	}
+}
+
+// Stop ECS services so CloudFormation can delete the ECS clusters
+func stopEcsServices(awsSession *session.Session, clusterNames []*string) {
```
3nvi: this deletes the services it doesn't stop them (since you call `client.DeleteService`)

3nvi:
> EDIT: You do have to manually stop the ECS service, otherwise CloudFormation fails to delete the main `panther-app` stack:
> 
> Screen Shot 2020-03-03 at 1 30 11 PM

I might be fully mistaken, but I feel that the reason you see this is the order of CFN deletions. You try to delete the cluster before deleting the associated services. If you attempt to call `destroyCfn` on the service, tasks, albs, etc. then you will be able to successfully delete the cluster.

I want to understand why can't we rely on CFN deleting the entire ECS related stuff. I've never had to mess with ECS while tearing down a panther installation manually. The only thing I did was to empty ECR. Nothing else. Everything else was handled automatically for me (I didn't even stop the service). What's the limitation here? Is it that we can't properly create a deletion order or is it something else that I'm missing?

tools/mage/teardown.go
3nvi: I understand and I'm ok with that. I'm still not getting one thing though; why would we ever be in the position where we are trying to delete something that was already deleted?

Is it because we are trying to re-run the same command multiple times, because - for example - of timeouts?

tools/mage/teardown.go
3nvi: cool :)

3nvi:
@rleighton @kostaspap just because this PR is having a lot of open review comments, let's try and resolve the stale ones :)

tools/mage/teardown.go
austinbyers: > why would we ever be in the position where we are trying to delete something that was already deleted?

The command might fail halfway through for any number of reasons - timeouts, permission denied, network problems. In that case, we would just run the command again. If we delete some of the infra but not all of it, we want to be able to skip the stuff that was already deleted

tools/mage/teardown.go
austinbyers: > Did it present any issues?

Yes, see my screenshot in another comment below. If we don't do any manual changes to ECS, CloudFormation destroy fails with "the Cluster cannot be deleted while Services are active"

tools/mage/teardown.go
austinbyers: Ah I see, yes I can rename the function

austinbyers:
> @austinbyers I might be fully mistaken, but I feel that the reason you see this is the order of CFN deletions. You try to delete the cluster before deleting the associated services. If you attempt to call destroyCfn on the service, tasks, albs, etc. then you will be able to successfully delete the cluster.

All of the stacks are being deleted in parallel, we don't have much control over the order of operations. Is there a specific order you were doing the deletion manually? I see that the service is defined in the `frontend` stack but the cluster is defined in the `backend` stack, I'll try deleting the frontend stack first to see if it works.

One thing I noticed is that it takes nearly a minute after deleting an ECS service before it's actually marked "inactive" and the cluster can be deleted. It's possible the deletion was working for you manually just because you weren't deleting the stacks as fast as we are here.

austinbyers:
> Do a re-review and add new comments if necessary for approval criteria

Thanks @3nvi, but this is no longer ready for review, I'll ping you all when it is, lots of changes still in progress

tools/mage/teardown.go
3nvi: ok 👍

3nvi:
> @3nvi I was able to delete the frontend stack first (which stopped the ECS service completely) and then the rest of the stacks in parallel. That definitely simplifies the code, thanks!

That's great news :)

tools/mage/util_cfn.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	// List stack resources
+	input := &cfn.ListStackResourcesInput{StackName: stackID}
+	var nestedErr error
+	err = client.ListStackResourcesPages(input, func(page *cfn.ListStackResourcesOutput, isLast bool) bool {
+		for _, summary := range page.StackResourceSummaries {
+			handler(cfnResource{Resource: summary, Stack: stack})
+			if aws.StringValue(summary.ResourceType) == "AWS::CloudFormation::Stack" &&
+				aws.StringValue(summary.ResourceStatus) != "DELETE_COMPLETE" {
```
kostaspap: Replace with `cfn.ResourceStatusDeleteComplete`

tools/mage/util_cfn.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// Recursively list resources for a single stack.
+//
+// The stackID can be the stack name or arn.
+func walkStack(client *cfn.CloudFormation, stackID *string, handler func(cfnResource)) error {
+	logger.Debugf("enumerating stack %s", *stackID)
+	detail, err := client.DescribeStacks(&cfn.DescribeStacksInput{StackName: stackID})
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ValidationError" &&
```
kostaspap: Damn, that's bad API from AWS over there. They could have a "not exists" error...

tools/mage/util_cfn.go
```diff
@@ -0,0 +1,130 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// Recursively list resources for a single stack.
+//
+// The stackID can be the stack name or arn.
+func walkStack(client *cfn.CloudFormation, stackID *string, handler func(cfnResource)) error {
```
kostaspap: nit: You are actually walking a Panther stack not any CFN stack since you are checking for the Panther tag in it. This method cannot be used for "any" stack, so it might make sense to rename to `walkPantherStack` and the method above `walkPantherStacks`

tools/mage/teardown.go
```diff
@@ -0,0 +1,396 @@
...
+			}
+
+			logger.Infof("deleting CloudWatch log group %s", *group.LogGroupName)
+			_, err := client.DeleteLogGroup(&cloudwatchlogs.DeleteLogGroupInput{LogGroupName: group.LogGroupName})
+			if err != nil {
+				logger.Fatalf("failed to delete log group %s: %v", *group.LogGroupName, err)
+			}
+		}
+		return true // keep paging
+	})
```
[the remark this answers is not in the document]
austinbyers: Since nearly all of the log groups have been deleted by CloudFormation already, it should be faster to list the `/aws/lambda/panther-` log groups and then delete the 1 or 2 that remain instead of issuing 30+ delete requests for all of the log groups.

But if we just delete all of the log groups defined in the CFN stacks, it is more general, so I'm fine with switching to that. Speed isn't a big deal here

PR #331, opened 2020-03-03
Deploy CloudSecurity into the Panther Account

rleighton:
## Background
Panther should monitor its own account. This PR adds a step to enable CloudSec in the deployed account. A follow-on PR will enable log processing

## Changes
- Updated PutIntegrations to be insensitive to adding same integration
- Added deploy "onboard" step that implements what is currently done by the UI wizard
- Parallelized some deployment steps

## Testing
mage test:ci
deployed into my dev account


internal/core/source_api/api/put_integration.go
```diff
@@ -94,6 +102,30 @@ func (API) PutIntegration(input *models.PutIntegrationInput) ([]*models.SourceIn
...
+	if err != nil {
+		err = errors.Wrap(err, "Error listing integrations")
+		return nil, err
+	}
+	currentIntegrationsMap := make(map[string]struct{})
+	for _, integration := range currentIntegrations {
+		currentIntegrationsMap[*integration.AWSAccountID+*integration.IntegrationType] = struct{}{}
+	}
+	for _, integration := range inputIntegrations {
+		if _, found := currentIntegrationsMap[*integration.AWSAccountID+*integration.IntegrationType]; found {
```
austinbyers: Instead of scanning the table and filtering it in the caller, we can just add a new column to the table (accountID + integrationType) and let Dynamo filter it for us. Then this code is a bit simpler and more efficient - you just look for any elements with that keypair. We use a similar trick when filtering policies, for example.

This endpoint is not used very often, so I don't think it's terribly important which way around we do it, but I like the simpler solution

tools/mage/deploy_onboard.go
```diff
@@ -0,0 +1,126 @@
...
+	}{
+		&models.PutIntegrationInput{
+			Integrations: []*models.PutIntegrationSettings{
+				{
+					AWSAccountID:     aws.String(pantherAccountID),
+					IntegrationLabel: aws.String("Panther Account"),
+					IntegrationType:  aws.String(models.IntegrationTypeAWSScan),
+					ScanEnabled:      aws.Bool(true),
+					ScanIntervalMins: aws.Int(1440),
+					UserID:           aws.String(uuid.New().String()), // generate a uuid4
```
austinbyers: We use all zeros for userIDs ("00000000-0000-0000....") when hitting APIs from mage, not a random one. That way the frontend can know it was a deployment script

tools/mage/deploy_onboard.go
```diff
@@ -0,0 +1,126 @@
...
+				},
+			},
+		},
+	}
+	if err := invokeLambda(awsSession, "panther-source-api", apiInput, nil); err != nil {
+		logger.Errorf("error calling lambda to register account: %v", err)
```
austinbyers: Why not make this (and all the similar log messages) `logger.Fatalf`? These errors are not recoverable - there's no point in continuing the deploy script. CloudFormation will still be running in the background and a subsequent `mage deploy` should see the stacks are in progress and wait for them.

As it is now, if all the parallel deployments fail, the main routine will continue and exit normally.

If you don't want to fatally log in the goroutines, an alternative is to communicate with channels instead of using a wait group, which is in my opinion a little cleaner anyway. [Example](https://github.com/panther-labs/panther/pull/330/files#diff-0e9b2b6b880362c0bfeb65cf38f52004R205):

```go
type deleteStackResult struct {
	stackName string
	err       error
}

// Log the results of a delete stack request
func (r deleteStackResult) log() {
	if r.err != nil {
		logger.Fatalf("stack %s failed to delete: %v", r.stackName, r.err)
	}
	logger.Infof("%s successfully deleted", r.stackName)
}

// Destroy CloudFormation stacks
func destroyCfnStacks(awsSession *session.Session) {
	results := make(chan deleteStackResult)
	client := cloudformation.New(awsSession)

	// Trigger the deletion of each stack
	logger.Infof("deleting CloudFormation stacks: %s", strings.Join(deleteStacksParallel, ", "))
	for _, stack := range deleteStacksParallel {
		go deleteStack(client, aws.String(stack), results)
	}

	// Wait for all of the deletions to finish
	for range deleteStacksParallel {
		(<-results).log()
	}
```

austinbyers:
Great idea! I love that a Panther deployment can onboard itself. Also, thanks for parallelizing some of the deployment

tools/mage/deploy_onboard.go
rleighton: If we log.Fatal() the process exits, the main will not keep going AFAIK. 

I did not log.Fatal() on the "onboarding" because I thought it less important. I can do that and everything will stop.

internal/core/source_api/api/put_integration.go
rleighton: Quite true. I was trying avoid schema changes (those tend to create regressions).

internal/core/source_api/api/put_integration.go
```diff
@@ -94,6 +102,30 @@ func (API) PutIntegration(input *models.PutIntegrationInput) ([]*models.SourceIn
 	return newIntegrations, err
 }
 
+func (api API) filterExistingIntegrations(inputIntegrations []*models.PutIntegrationSettings) (
+	existingIntegrations []*models.PutIntegrationSettings, err error) {
+
+	// avoid inserting if already done
+	currentIntegrations, err := api.ListIntegrations(&models.ListIntegrationsInput{})
+	if err != nil {
+		err = errors.Wrap(err, "Error listing integrations")
```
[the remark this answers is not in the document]
rleighton: Ah, I forgot that. Will make it match the others.

tools/mage/deploy_onboard.go
austinbyers: > If we log.Fatal() the process exits, the main will not keep going AFAIK.

Right, that's what I want. Why keep going? You can wait for all the goroutines to finish and then log an error if you want, but this is easier

austinbyers: > I did not log.Fatal() on the "onboarding" because I thought it less important. I can do that and everything will stop.

It's easy for someone to miss an error message and think the deploy succeeded. If you want to allow just the onboarding stack to fail, I suppose we can, but I'd prefer to just be consistent

internal/core/source_api/api/put_integration.go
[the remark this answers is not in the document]
austinbyers: We were originally planning on deprecating `genericapi` completely, but if we plan to keep using it, I'm sure there's a much better way to do the error handling when we have time!

internal/core/source_api/api/put_integration.go
austinbyers: The nice thing about Dynamo/NoSQL is that this doesn't have to be a schema change (no CloudFormation change required), unless you want to make it an index key (which is probably overkill).

Again, this is a rare endpoint, I'm happy with this implementation as-is if you want to keep it

internal/core/source_api/api/put_integration.go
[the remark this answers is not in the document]
rleighton: Done

tools/mage/deploy_onboard.go
rleighton: Done

tools/mage/deploy.go
```diff
@@ -352,7 +354,7 @@ func initializeAnalysisSets(awsSession *session.Session, endpoint string, config
 		response, err := apiClient.Operations.BulkUpload(&operations.BulkUploadParams{
 			Body: &analysismodels.BulkUpload{
 				Data:   analysismodels.Base64zipfile(encoded),
-				UserID: "00000000-0000-0000-0000-000000000000",
+				UserID: mageUserID,
```
austinbyers: 👍 thanks

Issue #192, opened 2020-02-14
Show Python revision history, allow reverting


austinbyers:
## Description
We store every revision to Python policies and rules, but we don't do anything with them. The idea was to give users the power to revert a bad rule change.

For example, deleted rules are removed from Dynamo but still exist in S3 and could be recovered.

## Acceptance Criteria

- Users can see the version history of any rule or policy (version number, timestamp, author)
- Users can revert rules and policies to any previous version
- Users can optionally show deleted rules/policies and restore them

OR

- Remove rule revision history if we don't ever plan to use it to greatly simplify the core analysis api

The Dynamo table storing policies and rules does have backups enabled, so it's not unreasonable to say that catastrophic failures could be reverted this way. But I think it would be better to allow users to revert individual rules + policies

### Backend
We already store old versions, and the `GetPolicy` / `GetRule` endpoint already allow you to retrieve older versions (even deleted ones). Changes we need to `analysis-api`:

- Add `ListVersions` endpoint which returns a list of version IDs, timestamps, and author (for a given policy/rule ID)
- Add some way to list deleted rules

### Frontend

- On rule/policy details, add dropdown to show older versions
- On rule/policy edit page, allow reverting to old version
- On rule/policy list, add option to show deleted rules

3nvi:
For me this is not super hard. Will we add this for v1?

austinbyers:
I don't think it's a priority, no, but should be an easy add after

3nvi:
@austinbyers 

> * Add ListVersions endpoint which returns a list of version IDs, timestamps, and author (for a given policy/rule ID)
> * Add some way to list deleted rules

Is any of the above implemented? If not we can turn that into an epic, else I can just assign myself in this one and remove the backend related work from the description

austinbyers:
Some backend exists (rule/policy versions are stored and you can GetPolicy or GetRule by a specific version), but we probably need other endpoints as well (listing, mainly)

I converted to an epic

austinbyers:
Reassigning this (very old epic) to @s0l0ist , who will tentatively be working on it this quarter!

The main requirement now is that alerts link back to the specific version of the policy/rule that triggered the alert, so users can see the version of the Python at the time. The rule/policy page should support loading old versions, show if the rule is the current version, and if not, how long ago it was.

Being able to list old/deleted versions and revert/restore can come later

PR #334, opened 2020-03-03
Switching to new mechanism to store and retrieve events that matched an alert

kostaspap:
## Background

This change make the switch to the new mechanism for processing, storing and querying events that match rules. This is the final step in a series of PRs to move to the new logic. 
The bulk of the work in this PR is to get the Alerts API to retrieve the matched events using S3 Select. In some tests I did in my dev account, Lambda was taking between 50-250ms to return results. 
I also fixed some small problems here and there that i discovered while doing end-to-end testing. 
With this PR you can also now make use of custom alert dedup. 

*ATTENTION* 
After merging this PR, you are going to lose all your previous alerts and any matched events!

## Changes

- Removing DDB tables `panther-recent-alerts`, `panther-events`. 
- Removing AWS Lambda `panther-alert-merger`
- Changed Alert API Alert struct field from `lastEventMatched` to (alert) `updateTime` 
- The `panther-log-alerts-info` table now contains also the Severity of an alert (taken from the Rule severity). That way the `panther-alerts-api` no longer needs to query the `analysis-api` to retrieve the severity of the rule - it is already present in the DDB table. 
- Handling panic in the `panther-alert-forwarder` Lambda. While we were unmarshaling the DDB item, if an attribute is of different type than the one provided, DDB APi panics. Added test case for this. 
- Updated the key format for log processor output. The timestamp in the key format is now slightly more compacted in the format `20060102150405 ` (removed the 'T' and 'Z'). Now it matched the format of rules-engine output
- Updated rules-engine to now store in DDB the log types that have matched an alert. 
- Updated `GetAlert` method from alerts-api. The method now queries S3 to fetch the events that matched an alert. It works like the following: 
 - User provides the alertId and a pagination token. 
   - The pagination token is a Base64 encoded JSON map. The map is a mapping from logtype -> (last s3 object processed, index of last event in that S3 object)
  - Alerts-API load from DDB information about the alert including the logtypes of events, the time the alert was created, the time the alert was last updated. 
  - If there was a pagination token provided, alerts-api will query that S3 object using S3 select. It will skip all the events prior to the eventIndex provided in the pagination token. 
  - Alerts-Api queries data from S3 using S3 Select for each of the log types:
     - It will perform list operations for each of the partition prefixes (partition prefixes depend on time). Time partitions are all the partitions between alert create and alert update time
     - It lists S3 objects under each partition. The S3 object's name contains the time the events in the S3 object were matched with second precision. We discard any S3 objects that are not within the alert.createTime / alert.update time interval. This allows us to filter even better the S3 objects we want to query, since partitions have only hour granularity
   - While we list the objects, we perform S3 Select queries on the S3 objects that might contain events. When we reach the `eventPageSize` number of events, we stop processing. 

## Testing

- Deployed in my dev account. Verified the following: 
  - Dedup is working!
  - I am able to query matched events in Athena
  - I can see events that matched rules from the UI
- Unit tests


docs/gitbook/operations/runbooks.md
```diff
@@ -211,30 +198,20 @@ This is the dead letter queue for the `panther-input-data-notifications-queue`.
  When the system has recovered they should be re-queued to the `panther-input-data-notifications-queue` using
  the Panther tool `requeue`.
 
+## panther-log-alert-dedup
+The `panther-rules-engine` lambda manages this table and it is used to
+ deduplicate of alerts. The `panther-log-alert-forwarder` read the ddb stream from this table.
```
rleighton: s/read/reads

internal/log_analysis/alert_forwarder/forwarder/models.go
```diff
@@ -42,7 +44,16 @@ type Alert struct {
 	AlertDedupEvent
 }
 
-func FromDynamodDBAttribute(input map[string]events.DynamoDBAttributeValue) (*AlertDedupEvent, error) {
+func FromDynamodDBAttribute(input map[string]events.DynamoDBAttributeValue) (event *AlertDedupEvent, err error) {
+	defer func() {
```
rleighton: Not sure this is needed/wanted. The function only calls getAttribute(), looking at the code I do not see how it could panic.

docs/gitbook/operations/runbooks.md
```diff
@@ -13,37 +13,27 @@ Resource names below refer to resources in the Cloud Formation templates in Pant
...
 
-## panther-alert-merger
-will be removed before v1
+## panther-alert-forwarder
+The `panther-alert-forwarder` lambda reads from the ddb stream for the table `panther-alert-forwarder`
+ and sends them to the `panther-alerts-queue` sqs queue.
+
+ Failure Impact
+ * Failure of this lambda will stop delivery of alerts to destinations.
+ * There will be no data loss until events are purged from the ddb stream (24 hours).
```
rleighton: Talk about duplicate alerts possible on failure since "batch" gets retried.

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -41,67 +56,212 @@ func (API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOutput,
...
 
+		// We only need to retrieve as many returns as to fit the EventsPageSize given by the user
+		eventsToReturn := *input.EventsPageSize - len(events)
+		eventsReturned, resultToken, err := getEventsForLogType(logType, token.LogTypeToToken[logType], alertItem, eventsToReturn)
+		if err != nil {
+			return nil, err
+		}
+		token.LogTypeToToken[logType] = resultToken
+		events = append(events, eventsReturned...)
+		if len(events) == *input.EventsPageSize {
```
rleighton: ```>=```

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -41,67 +56,212 @@ func (API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOutput,
...
+	return result, nil
+}
+
+// This method returns events from a specific log type that are associated to a given alert.
+// It will only return up to `maxResults` events
+func getEventsForLogType(
+	logType string,
+	token *LogTypeToken,
+	alert *table.AlertItem,
+	maxResults int) (result []string, resultToken *LogTypeToken, err error) {
```
rleighton: uint since it should never be negative?

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -41,67 +56,212 @@ func (API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOutput,
...
+	if token != nil {
+		events, index, err := queryS3Object(token.S3ObjectKey, alert.AlertID, token.EventIndex, maxResults)
+		if err != nil {
+			return nil, resultToken, err
+		}
+		result = append(result, events...)
+		// updating index in token with index of last event returned
+		resultToken.S3ObjectKey = token.S3ObjectKey
+		resultToken.EventIndex = index
+		if len(result) == maxResults {
```
rleighton: ```>=```

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -41,67 +56,212 @@ func (API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOutput,
...
+	}
+
+	var partitionLocations []string
+	for nextTime := alert.CreationTime; !nextTime.After(alert.UpdateTime); nextTime = awsglue.GlueTableHourly.Next(nextTime) {
+		partitionLocation := awsglue.GetPartitionPrefix(logprocessormodels.RuleData, logType, awsglue.GlueTableHourly, nextTime)
+		partitionLocations = append(partitionLocations, partitionLocation)
+	}
+
+	for _, partition := range partitionLocations {
+		if len(result) == maxResults {
```
rleighton: Wasn't this tested above? Why is this here?

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -41,67 +56,212 @@ func (API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOutput,
...
+	timeInString := strings.Split(keyParts[len(keyParts)-1], "-")[0]
+	return time.ParseInLocation(destinations.S3ObjectTimestampFormat, timeInString, time.UTC)
+}
+
+// Queries a specific S3 object events associated to `alertID`.
+// Returns :
+// 1. The events that are associated to the given alertID that are present in that S3 oject. It will return maximum `maxResults` events
+// 2. The index of the last event returned. This will be used as a pagination token - future queries to the same S3 object can start listing
+// after that.
+func queryS3Object(key, alertID string, exclusiveStartIndex, maxResults int) ([]string, int, error) {
```
rleighton: maxResults uint?

austinbyers:
> The panther-log-alerts-info table now contains also the Severity of an alert (taken from the Rule severity). That way the panther-alerts-api no longer needs to query the analysis-api to retrieve the severity of the rule - it is already present in the DDB table.

What if the severity of a rule changes? I agree it makes the most sense to store the severity at the time the rule was triggered, but we should also store the rule versionId that triggered the alert so that when we expose versioning (#192) a user can click through to see the version of the rule at the time the alert was actually triggered.

internal/log_analysis/alerts_api/api/get_alert.go
austinbyers: I don't think we use `uint`s anywhere just because then you would be constantly forced to typecast when doing comparisons or arithmetic. I personally think `int` is fine, but don't feel strongly

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -41,67 +56,212 @@ func (API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOutput,
...
+	// nolint:gosec
+	query := fmt.Sprintf("SELECT * FROM S3Object o WHERE o.p_alert_id='%s'", alertID)
```
austinbyers: Is the `alertID` actually verified to be a uuid? Because `gosec` is right, this could be susceptible to a SQL injection attack

Please add a comment about why you are 100% sure the alertID could not be `' or '1' = '1`, which would lead to:

```sql
SELECT * FROM S3Object o WHERE o.p_alert_id='' or '1'='1'
```

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -42,9 +42,12 @@ import (
 const (
 	// s3ObjectKeyFormat represents the format of the S3 object key
 	// It has 3 parts:
-	// 1. The key prefix 2. Timestamp in format `20060102T150405Z` 3. UUID4
+	// 1. The key prefix 2. Timestamp in format `s3ObjectTimestampFormat` 3. UUID4
 	s3ObjectKeyFormat = "%s%s-%s.json.gz"
 
+	// The timestamp format in the S3 objects
+	S3ObjectTimestampFormat = "20060102150405"
```
austinbyers: There's no `T` in this string like there used to be - is that intentional?

If we can, let's leave the `T` and `Z` so it's a little easier to understand. If we can't, maybe a brief comment about what that format means

internal/log_analysis/rules_engine/src/alert_merger.py
```diff
@@ -63,27 +69,29 @@ def _update_get_alert_info_conditional(match_time: datetime, num_matches: int, r
...
             '#4': _ALERT_UPDATE_TIME_ATTR_NAME,
             '#5': _ALERT_EVENT_COUNT,
-            '#6': _ALERT_COUNT_ATTR_NAME,
-            '#7': _ALERT_CREATION_TIME_ATTR_NAME,
-            '#8': _PARTITION_KEY_NAME,
+            '#6': _ALERT_SEVERITY_ATTR_NAME,
+            '#7': _ALERT_LOG_TYPES,
+            '#8': _ALERT_COUNT_ATTR_NAME,
+            '#9': _ALERT_CREATION_TIME_ATTR_NAME,
+            '#10': _PARTITION_KEY_NAME,
```
austinbyers: Weird that the Python SDK doesn't have an expression building library like Go does

docs/gitbook/operations/runbooks.md
kostaspap: Done

internal/log_analysis/alert_forwarder/forwarder/models.go
kostaspap: It actually does unfortunately. Very bad design from AWS in this case. If I try to get an attribute e.g. as a String and it is not, the `String()` method will panic. 
```
// String provides access to an attribute of type String.
// Method panics if the attribute is not of type String.
func (av DynamoDBAttributeValue) String() string {
...
}
```

internal/log_analysis/alerts_api/api/get_alert.go
kostaspap: True, better safe than sorry :P

internal/log_analysis/alerts_api/api/get_alert.go
kostaspap: No, it can only be positive. Going for `int` just for simplicity reasons tbh

internal/log_analysis/alerts_api/api/get_alert.go
kostaspap: Fixed

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -41,67 +56,212 @@ func (API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOutput,
 	if err != nil {
 		return nil, err
 	}
+	var token *EventPaginationToken
+	if input.EventsExclusiveStartKey != nil {
+		token, err = decodePaginationToken(*input.EventsExclusiveStartKey)
+		if err != nil {
+			return nil, err
+		}
+	} else {
```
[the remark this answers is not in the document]
kostaspap: True! Updated.

internal/log_analysis/log_processor/destinations/s3.go
kostaspap: I was on the fence with this one. Fair point, I'll add back the `T` and `Z`, I agree it makes it more readable.

internal/log_analysis/rules_engine/src/alert_merger.py
kostaspap: I know, it is actually pretty annoying but couldn't find such.

internal/log_analysis/alert_forwarder/forwarder/forwarder_test.go
```diff
@@ -74,14 +116,73 @@ func TestProcess(t *testing.T) {
...
+func TestStoreDDBError(t *testing.T) {
 	ddbMock := &mockDynamoDB{}
 	ddbClient = ddbMock
 
 	ddbMock.On("PutItem", mock.Anything).Return(&dynamodb.PutItemOutput{}, errors.New("error"))
-	assert.Error(t, Process(testAlertDedupEvent))
+	assert.Error(t, Store(testAlertDedupEvent))
+}
+
+// The handler signatures must match those in the LambdaInput struct.
```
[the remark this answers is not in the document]
kostaspap: It doesn't. Removed!

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -27,18 +27,36 @@ import (
...
 )
 
 var (
-	alertsTable                           = os.Getenv("ALERTS_TABLE")
-	awsSession                            = session.Must(session.NewSession())
-	ddbClient   dynamodbiface.DynamoDBAPI = dynamodb.New(awsSession)
+	alertsTable                                = os.Getenv("ALERTS_TABLE")
+	alertingQueueURL                           = os.Getenv("ALERTING_QUEUE_URL")
+	analysisAPIHost                            = os.Getenv("ANALYSIS_API_HOST")
+	analysisAPIPath                            = os.Getenv("ANALYSIS_API_PATH")
```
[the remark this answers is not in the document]
kostaspap: True, it's more clean. Updated!

docs/gitbook/operations/runbooks.md
kostaspap: Updated documentation.

internal/log_analysis/alerts_api/api/get_alert.go
kostaspap: So with this check, if we have already received the number of events we wanted we don't proceed to process more partitions. I have similar checks:
1. Line 82 : If we have already received the requested number of events, don't iterate over more log types
2. Line 137: This one. Don't iterate over more partitions
3. Line 176: Don't list more S3 objects.
4. Line 258: Don't get more events from the S3 object.

api/lambda/alerts/models/api.go
```diff
@@ -94,20 +83,20 @@ type ListAlertsOutput struct {
...
 // AlertSummary contains summary information for an alert
 type AlertSummary struct {
-	AlertID          *string    `json:"alertId"`
-	RuleID           *string    `json:"ruleId"`
-	CreationTime     *time.Time `json:"creationTime"`
-	LastEventMatched *time.Time `json:"lastEventMatched"`
-	EventsMatched    *int       `json:"eventsMatched"`
-	Severity         *string    `json:"severity"`
+	AlertID       *string    `json:"alertId"`
+	RuleID        *string    `json:"ruleId"`
```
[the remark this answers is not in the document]
kostaspap: Added. Currently we don't take rule version into consideration when doing the deduplication - if a rule changes, the new events that matched the rule will be merged with the previous events of the same rule. So I am storing the rule but this is more of a groundwork for the versioning feature.

internal/log_analysis/alerts_api/api/get_alert.go
kostaspap: This is a very good comment. 
Currently the alertID is actually ruleID:dedupstring:counter so it's not easy to do validation on the alertID. In fact, I realized that if the dedup string has a `'` character, this could cause the S3 Select to fail, so I have updated the code to escape such occurrences.

Regarding the impact of SQL injection: 
We run this query using S3 Select. This query cannot be manipulated to alter data (S3 Select is read only - thus the name) . It could be manipulated to return all data (the example you give above) but this would be all the data of a given S3 Object. Each S3 Object contains events that have matched a single rule (but maybe they have different dedup strings). So currently, it is not possible to be manipulated to return data that it shouldn't. 

Nevertheless, I don't like it and you are right to bring this issue up. We should have proper sanitization. I'm proposing the following path to move forward: 
1. Change the code to use as alertID a well defined pattern.
2. Update AlertAPI to only accept alertID of the defined pattern. Anything else is rejected .
3. Put restrictions around dedup string. This is something I realized that can be problematic, currently dedup string can be arbitrary in size. I am thinking just a size restriction for the time being. Not sure I want to be really restrictive here. 
4. UI needs to be updated to hide the AlertID and instead display the RuleID and Dedup string as separate columns. Currently AlertID is displayed in the Alerts page so it is used as a way to identify the rule that triggered the alert. 

I am thinking to do this in a follow-up issue. I'm just worried that this PR is already pretty big and also there is no real security concern given how our software currently works. 
Let me know your thoughts.

internal/log_analysis/alerts_api/api/get_alert.go
kostaspap: See above.

internal/log_analysis/alerts_api/api/get_alert.go
rleighton: I see, the above check and return only happens when token !=  nil

internal/log_analysis/alerts_api/api/get_alert.go
austinbyers: > I am thinking to do this in a follow-up issue. I'm just worried that this PR is already pretty big and also there is no real security concern given how our software currently works.

No real security issue, I'd agree, escaping quotes is I think a good first step. Can we escape all non-alphanumeric characters? E.g. not sure if the comment ('#') is recognized by S3 select and could also be in the dedup string

kostaspap: It doesn't have problem with `#` or `"` or empty spaces. Just ran some tests. 
Given that dedup string is currently user defined (in some tests I have locally I am using AWS regions `us-east-1`, `us-west-2`, etc)  that can contain nonalphanumeric. I'll leave it as is for now - we'll soon switch to some more well defined structure (maybe UUIDs) in which case this will no longer be an issue.

PR #336, opened 2020-03-04
Bug scanning buckets in us-east-1

nhakmiller:
## Background

The snapshot poller was failing to scan buckets in `us-east-1`. That is because the `get-bucket-location` API call does not work for buckets in `us-east-1`.

## Changes

- If the `get-bucket-location` API call returns `null`, then assume the bucket is in `us-east-1`

## Testing

- Deployed to development account and manually verified
- Added new unit test to handle this case

### DIFF
--- a/internal/compliance/snapshot_poller/pollers/aws/s3_bucket.go
+++ b/internal/compliance/snapshot_poller/pollers/aws/s3_bucket.go
@@ -192,8 +192,9 @@ func getBucketLocation(s3Svc s3iface.S3API, bucketName *string) *string {
 		return nil
 	}
 
+	// The get-bucket-location API call returns null for buckets in us-east-1. This behavior is not documented by AWS.
 	if out.LocationConstraint == nil {
-		return nil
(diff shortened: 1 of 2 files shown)


internal/compliance/snapshot_poller/pollers/aws/s3_bucket.go
```diff
@@ -192,8 +192,9 @@ func getBucketLocation(s3Svc s3iface.S3API, bucketName *string) *string {
 		return nil
 	}
 
+	// The get-bucket-location API call returns null for buckets in us-east-1. This behavior is not documented by AWS.
 	if out.LocationConstraint == nil {
-		return nil
```
nhakmiller: Previously (before open sourcing) this function returned `(*string, error)`, and this check was used to return an error. Then when we re-factored some things it changed to just return `*string`, but for some reason we left this check in anyways. Now we're making use of it to fix this bug, so we have found a new loving home for this poor little `if` statement :)

PR #337, opened 2020-03-04
Terraform translation of compliance template

glerb:
Didn't make an issue first, but have discussed extensively with Panther personnel.


glerb:
Updates made.

deployments/auxiliary/terraform/panther_compliance_terraform_template.tf
```diff
@@ -0,0 +1,277 @@
+locals {
+  MasterAccountId = "<insert accountID for account running Panther>"
```
austinbyers: This should instead be a [variable](https://www.terraform.io/docs/configuration/variables.html) so that terraform prompts the user when they're trying to apply the change. We want this to be completely standalone, someone can just `terraform apply` the change.

Something we haven't thought about before - since we already have the [deployment role](https://github.com/panther-labs/panther/blob/master/deployments/auxiliary/terraform/panther-deployment-role.tf) in this same folder, a `terraform apply` from this directory will deploy all of the `aux` templates.  Is that what we want?

I would say no, in which case we need to move each terraform template to its own folder. This can move to `deployments/auxiliary/terraform/compliance/main.tf`, for example

deployments/auxiliary/terraform/panther_compliance_terraform_template.tf
```diff
@@ -0,0 +1,277 @@
...
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "arn:aws:iam::${local.MasterAccountId}:root"
+      },
+      "Action": "sts:AssumeRole",
+      "Condition": {
+	"Bool": {"aws:SecureTransport": "true"}
```
austinbyers: Nit: fix the spacing here (bool should be indented)

If it's easier, you can instead use [jsonencode](https://www.terraform.io/docs/configuration/functions/jsonencode.html) to represent this as a Terraform object, then `terraform fmt` will format it correctly in the source.

```hcl
assume_role_policy = jsonencode({
    "Version"   = "2012-10-17",
    "Statement" = [
//...
})
```

This also helps ensure the final JSON is syntactically correct. But we can also make that change later, raw JSON is fine for now

deployments/auxiliary/terraform/panther_compliance_terraform_template.tf
```diff
@@ -0,0 +1,277 @@
...
+	}    
+    }
+  ]
+}
+EOF
+
+  tags = {
+    Name        = "Panther Audit Role"
+    Description = "The Panther master account assumes this role for read-only security scanning"
```
austinbyers: The role itself already has a description (and name), tags are fixed values used for automation or billing drilldown.

Let's keep just the `Application` tag to be consistent with the rest of the infrastructure

deployments/auxiliary/terraform/panther_compliance_terraform_template.tf
```diff
@@ -0,0 +1,277 @@
...
+    Application = "Panther"
+  }
+}
+
+resource "aws_iam_role_policy_attachment" "SecurityAudit" {
+  role       = "${aws_iam_role.PantherAudit.name}"
```
austinbyers: In the latest tf version, you don't need the quotes and brace for simple interpolation:

```hcl
resource "aws_iam_role_policy_attachment" "security_audit" {
  role       = aws_iam_role.panther_audit.name
  policy_arn = "arn:aws:iam::aws:policy/SecurityAudit"
}
```

deployments/auxiliary/terraform/panther_compliance_terraform_template.tf
```diff
@@ -0,0 +1,277 @@
...
+EOF
+
+  tags = {
+    Name        = "Panther Audit Role"
+    Description = "The Panther master account assumes this role for read-only security scanning"
+    Application = "Panther"
+  }
+}
+
+resource "aws_iam_role_policy_attachment" "SecurityAudit" {
```
austinbyers: It is more common to use `snake_case` for Terraform identifiers, which is consistent with Terraform resource names and makes it easier to do command-line operations:

```
terraform import resource.aws_iam_role_policy_attachement.security_audit
```

austinbyers:
We also need to add the AGPL license header, `mage fmt` doesn't yet support tf files. I can make that change in this PR

deployments/auxiliary/terraform/panther_compliance_terraform_template.tf
glerb: changed MasterAccountId to a variable and updated refs.

deployments/auxiliary/terraform/panther_compliance_terraform_template.tf
glerb: Neat trick. You'd think Hashi would put this function in their doc examples....

Done.

deployments/auxiliary/terraform/panther_compliance_terraform_template.tf
glerb: done

deployments/auxiliary/terraform/panther_compliance_terraform_template.tf
glerb: done.

austinbyers:
You should be whitelisted in CLA, not sure why that's not going through. @jacknagz to take a look later this week

austinbyers:
Temporarily disabling the CLA required check so we can get this in because we have the necessary paperwork

PR #341, opened 2020-03-05
Require integrations to be healthy

nhakmiller:
## Background

Since we've added the ability to check an integrations status, we will now require all create and update operations on integrations to only work if the proposed integration state would pass the health checks.

## Changes

- Required a new integration to be healthy before proceeding in the `PutIntegration` API
- Required an updated integration to be healthy before proceeding in the `UpdateIntegration` API
- Store the configuration of `RemediationEnabled` and `CWEEnabled` with the integration, and return them in read operations

## Testing

- Unit testing
- Deployed to dev account

Example of PutIntegration failure:
Input (in this case the CWE role is not deployed):
```
{
  "putIntegration": {
      "integrations": [{
    "AWSAccountId": "123456789012",
    "IntegrationType": "aws-scan",
    "CWEEnabled": true,
    "RemediationEnabled": true,
    "UserId": "94c3e8e2-b239-4a62-b25d-413ecb5077f7",
    "IntegrationLabel": "test-integration"
  }]
  },
  "skipScanQueue": true
}
```

Response:
```
{
  "errorMessage": "PutIntegration failed: invalid input: integration 123456789012 did not pass health check",
  "errorType": "InvalidInputError"
}
```

Example of ListIntegrations:
Input:
```
{
  "getEnabledIntegrations": {
    "integrationType": "aws-scan"
  }
}
```

Response:
```
[
  {
    "awsAccountId": "123456789012",
    "createdAtTime": "2020-03-05T00:06:39.440309891Z",
    "createdBy": "111",
    "integrationId": "222",
    "integrationLabel": "test-integration",
    "scanEnabled": true,
    "remediationEnabled": true,
    "cweEnabled": false,
    "scanIntervalMins": null,
    "s3Buckets": [],
    "kmsKeys": []
  }
]
```


internal/core/source_api/api/check_integration.go
```diff
@@ -137,13 +139,41 @@ func getCredentialsWithStatus(
...
+
+	// One of these will be nil, one of these will not. We only care about the value of the not nil one.
+	passing := aws.BoolValue(status.AuditRoleStatus.Healthy) || aws.BoolValue(status.ProcessingRoleStatus.Healthy)
+
+	// For these two, we are ok if they are not enabled or if they are passing
+	passing = passing && (!aws.BoolValue(integration.EnableRemediation) || aws.BoolValue(status.RemediationRoleStatus.Healthy))
+	passing = passing && (!aws.BoolValue(integration.EnableCWESetup) || aws.BoolValue(status.CWERoleStatus.Healthy))
+
+	// For these two, we are ok if none are set or all are passing
+	if len(integration.S3Buckets) > 0 {
```
kostaspap: nit: You don't really need the if statement. The `range` keyword can deal with nils: https://play.golang.org/p/xhCNFiDxOB9

internal/core/source_api/api/check_integration.go
```diff
@@ -137,13 +139,41 @@ func getCredentialsWithStatus(
...
+		for _, bucket := range status.S3BucketsStatus {
+			passing = passing && aws.BoolValue(bucket.Healthy)
+		}
+	}
+	if len(integration.KmsKeys) > 0 {
```
kostaspap: nit: Same as before.

internal/core/source_api/api/put_integration.go
```diff
@@ -38,8 +38,40 @@ import (
...
+		}
+		if !passing {
+			return nil, &genericapi.InvalidInputError{
+				Message: fmt.Sprintf("integration %s did not pass health check", *integration.AWSAccountID),
+			}
+		}
+	}
+
+	// Filter out existing integrations
+	integrations, err := api.filterExistingIntegrations(input.Integrations)
```
kostaspap: nit: I would rename the `filterExistingIntegrations` to `filterOutExistingIntegrations` or `getNewIntegrations` or similar. 
Filter is a bit ambiguous ("filter for" or "filter out")

PR #342, opened 2020-03-05
Provide CFN Templates via API

nhakmiller:
## Background

This change is to create an endpoint in the sources api that will return a configured CFN template, same as if the user had pressed the launch stack button. This is to support more automated/orchestrated workflows where users of Panther may not necessarily have the permissions to launch stacks in the satellite accounts to be monitored.

This is more of a draft PR, I just wanted to get it up tonight so that @3nvi and @alexmylonas could review the API format, provide feedback, and possibly get unblocked on the source onboarding rework.

Currently I am still testing/refining this change, but I suspect the way it interacts with the frontend won't change unless there are any specific requests from reviewers.

## Changes

- Added a `GetIntegrationTemplate` endpoint to the sources-api

## Testing

- Unit testing
- Manual testing still in progress


api/lambda/source/models/api.go
```diff
@@ -85,6 +87,20 @@ type ListIntegrationsInput struct {
...
+// GetIntegrationTemplateInput allows specification of what resources should be enabled/disabled in the template
+type GetIntegrationTemplateInput struct {
+	AWSAccountID       *string   `genericapi:"redact" json:"awsAccountId" validate:"required,len=12,numeric"`
+	IntegrationType    *string   `json:"integrationType" validate:"oneof=aws-scan aws-s3"`
+	RemediationEnabled *bool     `json:"remediationEnabled"`
+	CWEEnabled         *bool     `json:"cweEnabled"`
+	S3Buckets          []*string `json:"s3Buckets"`
+	KmsKeys            []*string `json:"kmsKeys"`
+}
+
```
nhakmiller: This is the struct you would send to the backend when making the request.

api/lambda/source/models/integration.go
```diff
@@ -73,3 +73,7 @@ type SourceIntegrationItemStatus struct {
 	Healthy      *bool   `json:"healthy"`
 	ErrorMessage *string `json:"errorMessage"`
 }
+
+type SourceIntegrationTemplate struct {
+	Body []byte `json:"body"`
+}
```
nhakmiller: This is the response. It is a byte string representation of a YAML file. Preferably this would be provided to the user as a download, not sure if this is the desired format for that.

deployments/auxiliary/cloudformation/panther-compliance-iam.yml
```diff
@@ -25,32 +25,26 @@ Parameters:
   MasterAccountId:
     Type: String
     Description: AWS account ID of the account running the Panther backend
+    Default: '' # MasterAccountId
```
nhakmiller: I added these comments to make string find/replace easier so we don't actually have to do YAML parsing or anything.

internal/core/source_api/api/get_integration_template.go
```diff
@@ -0,0 +1,142 @@
...
+	kmsKeyFind      = []byte("Default: '' # EncryptionKeys")
+	kmsKeyReplace   = "Default: %s # EncryptionKeys"
+)
+
+type templateCacheItem struct {
+	Timestamp time.Time
+	Body      []byte
+}
+
+// CheckIntegration adds a set of new integrations in a batch.
```
nhakmiller: Need to replace this comment.

internal/core/source_api/api/get_integration_template.go
```diff
@@ -0,0 +1,142 @@
...
+
+func getTemplate(integrationType *string) ([]byte, error) {
+	// First check the cache
+	if item, ok := templateCache[*integrationType]; ok && time.Since(item.Timestamp) < cacheTimout {
+		zap.L().Debug("using cached template")
+		return item.Body, nil
+	}
+
+	// Get the template from Panther's public S3 bucket
+	s3Svc := s3.New(sess)
```
nhakmiller: I need to specify region `us-west-2` here.

api/lambda/source/models/integration.go
kostaspap: @alexmylonas @3nvi  should have a saying on this.

api/lambda/source/models/api.go
nhakmiller: @3nvi this part is for the request.

api/lambda/source/models/integration.go
nhakmiller: @3nvi this is the part I'm most interested in your feedback on. We want to allow the users to download a YAML file. How should I send the data back to you to make that easiest? @austinbyers advised just sending back the raw string so I will do that for now, but please let me know what's best. I've never done something like this before so I'm not sure.

internal/core/source_api/api/get_integration_template.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"go.uber.org/zap"
+
+	"github.com/panther-labs/panther/api/lambda/source/models"
+)
+
+const (
+	TemplateBucket           = "panther-public-cloudformation-templates"
```
austinbyers: You're passing the bucket as a stack parameter to the `source-api` for IAM permissions. This means in theory the parameter in the stack could have a different value than this constant.

Either load the bucket name here from an environment variable or hardcode the name in both the CloudFormation and here

Since you're hardcoding the bucket region when creating the S3 client, I would hardcode the name in the stack instead of having it as a parameter

deployments/core/source_api.yml
```diff
@@ -40,6 +40,10 @@ Parameters:
   SQSKeyId:
     Type: String
     Description: KMS key ID for SQS encryption
+  TemplateSourceBucket:
```
austinbyers: This shouldn't be a parameter unless you're also passing it through to the function (see next comment)

I think this is fine to just hardcode in the IAM permission block

nhakmiller:
> I don't fully understand how we plan to use the dynamically generated templates, but the code LGTM

The main reason is that many organizations will not want to just launch a stack directly from the Panther UI for many reasons:

1. They may have a CI/CD pipeline that all CloudFormation resources are supposed to go through.
2. The team managing the Panther deployment may not have admin access in the account where panther is being deployed
3. There may be dozens to hundreds of accounts being onboarded.

In any of these cases, it would be preferable to just provide the template directly (via a template instead of a CloudFormation link) so they can feed it into whatever other systems they're using to manage AWS infrastructure. The frontend will use this API to provide these dynamically generated templates to users so they can still go through the web UI to configure everything.

PR #343, opened 2020-03-05
Output log events in event time based partitions

rleighton:
## Before
The log processor wrote events to S3 partitioned by _processing_ time.  This means that typically events  that occurred in the previous hour would be in a partition for the current hour. When users did SQL queries specifying the time partitions (needed for performance) they would miss events.

## After
The log processor now uses event time to create the S3 partitions.  

image

## Changes
* changed parser interface to return []*parsers.PantherLog rather than []interface{}, the PantherLog has both logType and event time needed
* PantherLog now has a back ptr field called Event that is of type interface{} to hold the raw event
* refactored tests

## Testing
mage test:ci
ran in dev account


internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -142,35 +139,38 @@ func (destination *S3Destination) SendEvents(parsedEventChannel chan *common.Par
...
 }
 
-func (destination *S3Destination) sendExpiredData(logTypeToEvents map[string]*s3EventBuffer) error {
-	currentTime := time.Now().UTC()
-	for logType, buffer := range logTypeToEvents {
-		if currentTime.Sub(buffer.firstEventProcessedTime) > maxDuration {
-			err := destination.sendData(logType, buffer)
+func (destination *S3Destination) sendExpiredData(bufferSet s3EventBufferSet) error {
+	return bufferSet.apply(func(buffer *s3EventBuffer) error {
+		if time.Since(buffer.createTime) > maxDuration {
```
kostaspap: `time.Since` is a shorthand for `time.Now().Sub(t).`. This means that we call `time.Now` (and make the associated syscalls) multiple times per `sendExpiredData` invocation - the previous code would do it once per  `sendExpiredData` invocation. You might want to call time.Now() once and pass the time to the function.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -175,7 +175,7 @@ func (p *ALBParser) LogType() string {
 }
 
 func (event *ALB) updatePantherFields(p *ALBParser) {
-	event.SetCoreFields(p.LogType(), event.Timestamp)
+	event.SetCoreFields(p.LogType(), event.Timestamp, event)
```
kostaspap: This confuses me a bit. The `event` object is passing itself as a parameter to its own method? I'm pretty sure this is not needed.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Good point.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
rleighton: It is AFAIK. Sets the back ptr as type interface{}.

kostaspap: Hmm, interesting. If you don't set it, is the result of the parser not populated? Worth testing it to see how Golang handles this case? Without knowing a lot about the internals, it seems weird because the parser has already set the appropriate fields to the event ^^

If we do have to do this, do we have to pass the `event` parameter to the method? I would imagine that `event` has already a reference to itself, it doesn't need to take it from the parameter?

rleighton: I would love to not pass it but cannot think of any other way so that we have the structure available when we need to serialize to S3. If we serialize a "PantherLog" then we just get those fields, not the full event.

rleighton:
Added tests.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -143,47 +144,44 @@ func (destination *S3Destination) SendEvents(parsedEventChannel chan *common.Par
...
 	zap.L().Debug("finished sending messages", zap.Int("events", eventsProcessed))
 }
 
-func (destination *S3Destination) sendExpiredData(logTypeToEvents map[string]*s3EventBuffer) error {
-	currentTime := time.Now().UTC()
-	for logType, buffer := range logTypeToEvents {
-		if currentTime.Sub(buffer.firstEventProcessedTime) > maxDuration {
-			err := destination.sendData(logType, buffer)
+func (destination *S3Destination) sendExpiredData(bufferSet s3EventBufferSet) error {
+	now := time.Now()
```
kostaspap: `time.Now().UTC()`

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -72,15 +75,17 @@ type S3Destination struct {
 	// snsTopic is the SNS Topic ARN where we will send the notification
 	// when we store new data in S3
 	snsTopicArn string
+	maxFileSize int
+	maxDuration time.Duration
```
kostaspap: Hmmm, not sure it makes sense to declare them as variables while the constants are declared just a few lines above. Why would anyone use the struct variables while we have a constant declared for this exact purpose? 

If you think they are useful to have, I'd say we should move the constants to the place they are used (in the `destinations.go`) and rename them to sth like "s3MaxFileSize", "s3BufferDuration".  Or just keep the code as it was.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -72,15 +75,17 @@ type S3Destination struct {
...
 // and stores them in the appropriate S3 path. If the method encounters an error
 // it writes an error to the errorChannel and continues until channel is closed (skipping events).
-func (destination *S3Destination) SendEvents(parsedEventChannel chan *common.ParsedEvent, errChan chan error) {
+func (destination *S3Destination) SendEvents(parsedEventChannel chan *parsers.PantherLog, errChan chan error) {
 	failed := false // set to true on error and loop will drain channel
-	logTypeToBuffer := make(map[string]*s3EventBuffer)
+	bufferSet := newS3EventBufferSet()
 	eventsProcessed := 0
 	zap.L().Debug("starting to read events from channel")
 	for event := range parsedEventChannel {
```
kostaspap: We don't need to drain the channel at line 92? We close the channel at the end of the execution so there is no memory-leak.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -124,14 +125,14 @@ func (destination *S3Destination) SendEvents(parsedEventChannel chan *common.Par
 			if !canAdd {
```
kostaspap: So this is a non-recoverable case. This can happen only if an event doesn't fit in the buffer (if the event is bigger than maxMemorySize). We should just log the error and drop the event since there is nothing we can do. We shouldn't return error. Better to add a comment to that extend.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -268,29 +265,71 @@ func getS3ObjectKey(logType string, timestamp time.Time) string {
...
+
 // s3EventBuffer is a group of events of the same type
 // that will be stored in the same S3 object
 type s3EventBuffer struct {
-	buffer                  *bytes.Buffer
-	writer                  *gzip.Writer
-	bytes                   int
-	events                  int
-	firstEventProcessedTime time.Time
+	destination *S3Destination
```
kostaspap: No need to pass the full destination here. You can just have a "maxBufferSize" or "maxSize" or similar - something more "reasonable" for a buffer struct.  "S3Destination" is a complex object with a lot of logic, not just a struct holding configuration so it is a bit weird that the buffer would need it - it is more reasonable an "S3Destination" to contain a bufferSet than the other way around.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -268,29 +265,71 @@ func getS3ObjectKey(logType string, timestamp time.Time) string {
 		uuid.New().String())
 }
 
+// s3BufferSet is a group of buffers associated with hour time bins, pointing to maps logtype->s3EventBuffer
+type s3EventBufferSet map[time.Time]map[string]*s3EventBuffer
```
kostaspap: You can change the bufferset to contain a "maxBufferSize" variable to avoid passing the destination as parameter to the s3EventBufferSet.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -268,29 +265,71 @@ func getS3ObjectKey(logType string, timestamp time.Time) string {
...
+
+func newS3EventBufferSet() s3EventBufferSet {
+	return make(map[time.Time]map[string]*s3EventBuffer)
+}
+
+func (bs s3EventBufferSet) getBuffer(destination *S3Destination, event *parsers.PantherLog) *s3EventBuffer {
```
kostaspap: See my comments above: you can probably change this to `func (bs s3EventBufferSet) getBuffer(event *parsers.PantherLog) *s3EventBuffer`

internal/log_analysis/log_processor/parsers/patherlog.go
```diff
@@ -39,6 +39,9 @@ var (
 //       See https://github.com/awsdocs/amazon-athena-user-guide/blob/master/doc_source/updates-and-partitions.md
 // nolint(lll)
 type PantherLog struct {
+	// back pointer to event
+	Event interface{} `json:"-" validate:"-"` // points to event that encapsulates this, do NOT JSON serialize or validate!
```
kostaspap: We probably want to put this one at the end of the struct? It is more likely that the "event" will change structure and get some extra fields - and when this happens we have to re-sync the glue data catalog. If we move it at the end of the struct, we will have to do this operation only when we add extra common fields - which will become eventually a more rare process.

internal/log_analysis/log_processor/parsers/awslogs/aurora_mysql_audit.go
```diff
@@ -117,7 +117,7 @@ func (p *AuroraMySQLAuditParser) LogType() string {
 }
 
 func (event *AuroraMySQLAudit) updatePantherFields(p *AuroraMySQLAuditParser) {
```
kostaspap: Something that just realized is that we have circular dependency between events and parsers in our all parsers. Parser depends on event, event depends on parser. It's huge deal (they are both part of the same package) but it would better to avoid it and have sth like this: 
```
func (p *AuroraMySQLAuditParser) Parse(log string) []*parsers.PantherLog{
...
p.updatePantherFields(event)
}

func (p *AuroraMySQLAuditParser) updatePantherFields(event * AuroraMySQLAudit) {
event.SetCoreFields(p.LogType(), event.Timestamp, event)
...
} 
```

No need to do this as part of this PR but I wanted to bring this up. I can take a stab at it once this is merged.

internal/log_analysis/log_processor/parsers/awslogs/aurora_mysql_audit_test.go
```diff
@@ -67,15 +68,6 @@ func checkAuroraMysqlAuditLogLog(t *testing.T, log string, expectedEvent *Aurora
...
-	require.Greater(t, len(*event.PantherRowID), 0) // ensure something is there.
-	expectedEvent.PantherRowID = event.PantherRowID
-
-	// PantherParseTime is set to time.Now().UTC(). Require not nil
-	require.NotNil(t, event.PantherParseTime)
-	expectedEvent.PantherParseTime = event.PantherParseTime
-
-	require.Equal(t, expectedEvent, event)
+	expectedEvent.Event = expectedEvent // set back ptr
+	testutil.EqualPantherLog(t, expectedEvent.Log(), events[0])
```
kostaspap: I am thinking it would be enough to have
```
testutil.EqualPantherLogs(t, expectedEvent.Logs(), events)
```

It is the most common scenario (all parsers return slice) and we can avoid having to do size check on every unit test (line 70 here)

internal/log_analysis/log_processor/parsers/patherlog.go
kostaspap: For the same reason, I believe we should do the same in the `AWSPantherLog` (not present in this PR)

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -80,7 +80,7 @@ func (p *ALBParser) New() parsers.LogParser {
 }
 
 // Parse returns the parsed events or nil if parsing failed
-func (p *ALBParser) Parse(log string) []interface{} {
+func (p *ALBParser) Parse(log string) []*parsers.PantherLog {
```
kostaspap: Comment for line 72: It seems line 72 is not needed anymore? The AWSPantherLog now contains the "Event" field - that should be enough I would imagine since we have all the references we need. 

We can have a method like this to simplify things. 
```
func new(event interface{}) *PantherLog{
	return &PantherLog{Event: event}
}
```
method in the `pantherlog.go`.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: It is locally used, does not need timezone, right?

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Good catch, I think this was left over when I moved these local to the object. I think they can be removed.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Agree. That is the old code, I did not change that but makes more sense.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Good idea.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
rleighton: I considered this but it adds complexity to creation we do not have now for very little gain and if someone forgets to do this it will crash. It is nice to just declare the struct . Essentially, you are moving the "set" from 1 new param in SetCoreFields() to here with a lot more code per parser.

internal/log_analysis/log_processor/parsers/awslogs/aurora_mysql_audit_test.go
rleighton: Good idea.

internal/log_analysis/log_processor/parsers/patherlog.go
rleighton: Actually, this does not matter as it is not serialized.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Ah, no these are  not variables, they are fields. We need this so testing works, we do not want global config for these values.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: I've simplified the whole thing, have a look.

internal/log_analysis/log_processor/parsers/awslogs/aurora_mysql_audit.go
rleighton: Seems unnecessary.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: I simplified the loop, have a look.

internal/log_analysis/log_processor/parsers/awslogs/aurora_mysql_audit_test.go
rleighton: Done

internal/log_analysis/log_processor/parsers/awslogs/aurora_mysql_audit.go
rleighton: Seems unnecessary.

internal/log_analysis/log_processor/destinations/s3.go
kostaspap: Might be right. But it's more for consistency and ease of reviewing - devs shouldn't have to resonate about this.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -72,118 +75,103 @@ type S3Destination struct {
...
 
 	operation := common.OpLogManager.Start("sendData", common.OpLogS3ServiceDim)
 	defer func() {
-		// if no error reset buffer
-		if err == nil {
-			if err = buffer.reset(); err != nil {
-				err = errors.Wrap(err, "failed to reset buffer")
-			}
+		if resetErr := buffer.reset(); resetErr != nil {
+			operation.LogError(err)
```
kostaspap: Shouldn't this stay the same as before? We are calling operation.LogError again just a few lines below, at line 176.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -268,53 +255,91 @@ func getS3ObjectKey(logType string, timestamp time.Time) string {
...
+		bs[hour] = logTypeToBuffer
+	}
+
+	logType := *event.PantherLogType
+	buffer, ok := logTypeToBuffer[logType]
+	if !ok {
+		buffer = &s3EventBuffer{
+			logType:    logType,
+			hour:       hour,
+			createTime: time.Now(),
```
kostaspap: See comment above. I think time.Now() would work here but I think for future developers better just have UTC everywhere - `time.Now().UTC()`

internal/log_analysis/log_processor/destinations/s3.go
kostaspap: Looks nice!

internal/log_analysis/log_processor/parsers/awslogs/alb.go
kostaspap: What if we did it like this: 

```
func NewPantherLog(logtype string, eventTime *time.Time, event interface{}) *PantherLog{
	result := &PantherLog{}
	result.Event = event
        result.PantherLogType = &logType
        result.PantherEventTime = eventTime
        ....      
	return result
}
```

and remove the SetCoreFields? 
Now it is more clear in the code that to get a PantherLog, there are some fields that are required. 

Some other benefits that I see: 
1. We make sure that the "Event" is always serialized after the standard Panther fields, since the 
`Event` fields are the last in the PantherLog. Now we have to make sure this is present in every Parser. 
```
type PantherLog struct {
        //  required
	PantherLogType   *string            `json:"p_log_type,omitempty" validate:"required" description:"Panther added field with type of log"`
       ......
	// back pointer to event
	Event 
}
``` 
2. Not need to call SetCoreFieldsPtr every time - the logic is embedded in the constructor. This makes the requirement for core fields more clear. 

*Note:* : I know it is not that critical and current solution is actually going to work just fine. I just think it is an interesting debate to have since we are now designing this solution.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Actually, I think the opposite. I know you are worried about someone missing a UTC() during log processing (very valid concern) but that is not the general case. When you see UTC() you should think "timezone matters here", which is misleading here because it does not.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: No, 2 different errors, this case is a failure in the reset. BUT, you did find a bug, this should be:
`operation.LogError(restErr)`

internal/log_analysis/log_processor/parsers/awslogs/alb.go
rleighton: I know why you want to do this but this is a little challenging (but worth talking thru!).

First, the Event field is not serialized at all so the order of it does not matter (see the tags on it). Since this has come up multiple times I'm going to make it private so this is very clear. If you are confused then another developer will really be confused.

The reason we can't do this in a normal "constructor" pattern is we do not event time available when we would create the event. We could make a function as you suggest but we would have to call it not at the start but where we currently call SetCoreFields(). So all you are doing is renaming SetCoreFields() to a function that does the same thing (but does an unnecessary alloc). Does that make sense?

I'll update the code shortly with Event as private and this may become more clear.

internal/log_analysis/log_processor/destinations/s3.go
kostaspap: I think it is better to have a "general" case though. It is difficult to understand when omitting the timezone is accidental and when it is intentional. It is really hard for a code reviewer if they have to resonate every time if UTC() or Local time zone is required. 
From personal experience on this PR: I saw "time.Now()" and immediately flagged it as suspicious. To check if it introduced some problems or not, I had to checkout the file locally to search using my IDE  where this value was used (it could have been used in code that was not modified as part of this PR and thus not visible) to make sure we were not using it in an "appropriate" place. 

I think it is much safer to just have UTC everywhere in the code. If some piece of the code requires for some reason local time, then we can do that there - and likely put a comment why local time was chosen over UTC. That should be the exceptional case though, since comments many times get outdated. 

Makes sense?

internal/log_analysis/log_processor/parsers/awslogs/alb.go
kostaspap: > So all you are doing is renaming SetCoreFields() to a function that does the same thing (but does an unnecessary alloc). Does that make sense?

Exactly! The idea is to call the constructor instead of calling the SetCoreFields. 
I think it is more clear as it avoids :
1. `Event` that is not validated, nor serializable
2. The `event.SetCoreFields(logType, time, event)` is somewhat confusing

I think the code will have "less comments" and thus be more clear if we go that way? Are there reasons to pick the approach currently in this PR over what we are discussing here? 
There's been already a lot of work in this PR so it would be fine to merge it as is. Just thinking whether we should keep in mind this alternative approach to do in subsequent PR.

rleighton: No, I don't think you understand. I need to walk you through the code.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: No, I disagree. Having UTC() tells the program this code needs a timezone. Very misleading to use it when not needed. We have been careful to wrap all the data processing code, I think we are safe.

That said, this was a good discussion. I looked closely at the expiration code, and while correct is _very_ inefficient.  It was checking on each event. I changed it to be timer driven, have a look and tell me what you think.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
rleighton: I need to walk you thru the code. I don't think you understand. See latest update

PR #358, opened 2020-03-06
Graceful handling of Python errors

### LINKED ISSUE #335: Graceful Handling of Python Errors in Analysis Testing
### Describe the bug

When writing a rule in the UI editor, any runtime error generated from a test throws 500 errors and deletes all local state. This is reproducible if you are importing a library that does not exist, or any other error occurs once the Python rule/policy function executes.

### Steps to reproduce

Create a fake test event, and run the tests:

### Expected behavior

We should capture the output of the rules engine, and present it to the user without crashing the page.

### Environment

- Panther version or commit: v0.3.0-50-g7a636e2
- Browser: Chrome Version 80.0.3987.122 (Official Build) (64-bit)

### Additional context

None

### Screenshots

Attached above

kostaspap:
## Background

Resolving the issue described in #335 

Closes #335 

## Changes

- Small improvement in analysis api. I am now reusing the AWS Lambda client and avoid creating it on every "Test Policy" invocation. 
- Scaling down policy engine. 512 MB seemed a bit too much. The engine has have maximum 4MB of input and similar amount of output. My metrics show that it uses minimum amount of memory and i can't see how we can end up using more than a few MB of RAM. @nhakmiller @austinbyers let me know if you are ok with this. Similar for timeout, it seems 1 minute should be enough for a single policy engine invocation? 
- Handling rule errors in direct analysis

## Testing

- Added unit tests for direct analysis! (we had none prior)
- Tested the scenario described in the issue in my dev account. Behavior seems correct: 
Screen Shot 2020-03-06 at 5 29 11 PM


internal/log_analysis/rules_engine/tests/test_main.py
```diff
@@ -0,0 +1,96 @@
...
+    'ANALYSIS_API_FQDN': 'analysis_fqdn',
+    'S3_BUCKET': 's3_bucket',
+    'NOTIFICATIONS_TOPIC': 'sns_topic',
+    'ANALYSIS_API_PATH': 'path'
+}
+with mock.patch.dict(os.environ, _ENV_VARIABLES_MOCK):
+    with mock.patch.object(boto3, 'client', side_effect=mock_to_return):
+        with mock.patch.object(SigV4Auth, 'add_auth'):
+            with mock.patch.object(requests, 'get', return_value=_RESPONSE_MOCK):
+                from ..src.main import lambda_handler
```
austinbyers: You can add multiple patches in the same `with` statement:

```python
with mock.patch.dict(os.environ, _ENV_VARIABLES_MOCK),
    mock.patch.object(),
    mock.patch.object(),
    mock.patch.object():
```

internal/core/analysis_api/handlers/clients.go
```diff
@@ -65,6 +68,7 @@ func Setup() {
 	dynamoClient = dynamodb.New(awsSession)
 	s3Client = s3.New(awsSession)
 	sqsClient = sqs.New(awsSession)
+	lambdaClient = lambda.New(awsSession)
```
austinbyers: The reason I created it separately was simply that the majority of invocations are not coming from the web app and don't need this client.

The overhead is probably trivial, I think this is fine.

deployments/compliance/resources_api.yml
```diff
@@ -302,7 +302,7 @@ Resources:
       # * Failure of this lambda will impact evaluating policies.
       # </cfndoc>
       Handler: src.main.lambda_handler
-      MemorySize: 512
+      MemorySize: 128
```
austinbyers: Remember, memory is proportional to CPU. The idea was to keep the core functions running faster to reduce latency. I'm fine with lowering it for now, but it wasn't about the memory per se

internal/log_analysis/rules_engine/tests/test_main.py
kostaspap: Ah, right! Yeah, much better like this. Updated.

internal/core/analysis_api/handlers/clients.go
kostaspap: Ah yeah, I don't think it was such a big issue. I just updated mostly to make the code follow the same style (clients initialized once per container) and change was much simpler. WIth this change unit testing will eventually be easier as well

PR #362, opened 2020-03-06
Fixing a misnamed function

nhakmiller:
## Background

A function was renamed in some places but not others it appears.

## Changes

- Renamed the function
- `mage fmt` caught some things

## Testing

- Unit tests

### DIFF
--- a/docs/gitbook/operations/runbooks.md
+++ b/docs/gitbook/operations/runbooks.md
@@ -206,7 +206,7 @@ The `panther-rules-engine` lambda manages this table and it is used to
  * Processing of rules could be slowed or stopped if there are errors/throttles.
 
 ## panther-log-alert-forwarder
-This lambda reads from a DDB stream for the `panther-alert-dedup` table and writes alerts to the `panther-log-alerts-info` ddb table.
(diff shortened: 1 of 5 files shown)


docs/gitbook/operations/runbooks.md
```diff
@@ -206,7 +206,7 @@ The `panther-rules-engine` lambda manages this table and it is used to
  * Processing of rules could be slowed or stopped if there are errors/throttles.
 
 ## panther-log-alert-forwarder
-This lambda reads from a DDB stream for the `panther-alert-dedup` table and writes alerts to the `panther-log-alerts-info` ddb table.
```
nhakmiller: Not actually sure about this change and the below one, but it was done automatically be either `mage fmt` or `mage setup:all` or `mage test:ci`, I didn't see which one was the one that made the change. Does this seem right?

austinbyers: Yes, it is supposed to be `panther-log-alert-info` - this documentation is autogenerated by `mage fmt` (which runs `mage doc:cfn`)

https://github.com/panther-labs/panther/blob/master/deployments/log_analysis/alerts.yml#L119-L126

PR #366, opened 2020-03-08
Tidies for aws event processor 

rleighton:
NOTE: I did not add any unit tests in the interest of time but I think we should go over all the code and add tests (big task).

## Changes
* Logging tidy in ec2
* Whitelist ReEncrypt
* Add lambda event and. version: PublishLayerVersion20181031
* add DeleteKey event for kms
* fix key extraction for kms

## Testing
* mage test:ci
* deployed to dev account


internal/compliance/aws_event_processor/processor/kms.go
```diff
@@ -38,27 +38,33 @@ func classifyKMS(detail gjson.Result, metadata *CloudTrailMetadata) []*resourceC
...
-			resourceARN, err := arn.Parse(resource.Get("ARN").Str)
-			if err != nil {
-				zap.L().Error("kms: unable to extract ARN", zap.String("eventName", metadata.eventName))
-				return nil
-			}
-			if strings.HasPrefix(resourceARN.Resource, "key/") {
-				keyARN = resourceARN.String()
-			}
-		}
+	case "CancelKeyDeletion",
```
nhakmiller: In general I actually prefer having it all on one line, especially on big services it makes it hard to read all split out like this.

PR #372, opened 2020-03-09
Add max retries to datacatalog_updater lambda

rleighton:
## Background
During load testing of log processing the datacatalog_updater was being throttled by the Glue API. Increase retries to 20.

## Changes
* Add MaxRetries(20) to session

## Testing
mage test:ci


internal/log_analysis/datacatalog_updater/main/lambda.go
```diff
@@ -38,9 +39,13 @@ import (
...
 )
 
+const (
+	maxRetries = 20 // setting Max Retries to a higher number - we'd like to retry VERY hard before failing.
+)
+
 var (
 	validation                   = validator.New()
-	glueClient glueiface.GlueAPI = glue.New(session.Must(session.NewSession()))
+	glueClient glueiface.GlueAPI = glue.New(session.Must(session.NewSession(aws.NewConfig().WithMaxRetries(maxRetries))))
```
kostaspap: Are they exponential retries or retries at constant time period? If they are exponential, 20 might not make much sense.

rleighton: Depends on the step size (exponent), I'm not sure what the sdk uses. We use 20 for the log processor and it seems good. I'll be playing with this more as I load test.

I wrote an ops tool (PR soon) that will list a bucket and send notifications to a queue. I've been "replaying" large amounts of CloudTrail. Customers can use this tool to backfill data after they onboard.

PR #374, opened 2020-03-09
Fixing deployment issue

kostaspap:
## Background

Currently it is not possible to deploy from master due to an issue in the CFN templates. The `onboard.yml` is providing property `DeployAuditRole` to the `panther-compliance-iam.yml` template, but such property doesn't exist. 

## Changes

- Removed `DeployAuditRole` property from `onboard.yml`
- Removed default `MasterAccountId`. Such a default doesn't really make sense, the template cannot be deployed without it. 

## Testing

- Successfully deployed to my dev account.

### DIFF
--- a/deployments/auxiliary/cloudformation/panther-compliance-iam.yml
+++ b/deployments/auxiliary/cloudformation/panther-compliance-iam.yml
@@ -25,6 +25,8 @@ Parameters:
   MasterAccountId:
     Type: String
     Description: AWS account ID of the account running the Panther backend
+    Default: '' # DO NOT REMOTE. We need an empty value here (even though the parameter is required). The empty string will be
+    # replaced with appropriate value by Golang application.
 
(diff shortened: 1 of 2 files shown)


deployments/auxiliary/cloudformation/panther-compliance-iam.yml
```diff
@@ -25,7 +25,8 @@ Parameters:
   MasterAccountId:
     Type: String
     Description: AWS account ID of the account running the Panther backend
-    Default: '' # MasterAccountId
+    Default: '' # DO NOT REMOTE. We need an empty value here (even though the parameter is required). The empty string will be
```
rleighton: s/REMOTE/REMOVE

kostaspap: Fixed

PR #381, opened 2020-03-09
Conditionally Deploy CloudSec Roles and Improve Teardown

rleighton:
## Before
The roles for CloudSec were unconditionally created on each deploy. Since these roles are global, this created errors when deploying to more than 1 region in an account. During `mage teardown` the stack was unconditionally removed (which may not be what you want if you have more than 1 region). Also, teardown did not remove the real time event stack.

## After
The deploy target now checks if the expected roles exist, and if so, skips deploying the stack for the CloudSec roles.

Teardown has been updated to ask if you want to delete the stack with the CloudSec roles (if it exists in that region). Also, teardown removes the real time events stack set and stack instance.

## Changes
* Factored the the CloudSec roles in the CF to be passed down from the backend stack and supplied by mage
* Ask user during teardown if they want to remove stack with CloudSec roles
* Remove real time event stack set
* Handled deleted buckets in teardown
* added MaxRetries = 20 to session to avoid throttling errors
* added log line in deploy showing region
* Added the Go version on the lambda build line

## Testing
mage test:ci
mage deploy
mage teardown


tools/mage/teardown.go
```diff
@@ -197,6 +204,129 @@ func destroyCfnStacks(awsSession *session.Session) error {
...
+			RetainStacks: aws.Bool(false),
+		})
+		if err != nil {
+			// need to return 2 errors
+			results <- deleteStackResult{stackName: stackSetInstanceName, err: err}
+			results <- deleteStackResult{stackName: *stackSet,
+				err: errors.Errorf("stack set instance not empty")}
+			return
+		}
+	}
```
nhakmiller: I'm worried this is not sufficient to safely delete this stack set. I think that it would be very likely that users may decide to expand this initial stackset to all regions in the account, as opposed to deploying a second stackset with all the other regions except the default one enabled. In that case, you would have to first list instances, get the regions of each instance, and delete those as well. Double unfortunately, users may expand this stackset to manage OTHER accounts as well. Do you think its reasonable to expand this to list instances first, then delete each instance?

tools/mage/deploy.go
```diff
@@ -102,6 +106,8 @@ func Deploy() {
 	Build.Lambda(Build{})
 	preprocessTemplates()
 
+	logger.Infof("deploy: %s", *awsSession.Config.Region)
```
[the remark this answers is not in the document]
rleighton: Good idea.

tools/mage/deploy_onboard.go
```diff
@@ -31,20 +32,42 @@ const (
...
+		logger.Info("deploy: creating iam roles for CloudSecurity")
+		params := map[string]string{} // currently none
+		deployTemplate(awsSession, onboardTemplate, bucket, onboardStack, params)
+	} else {
+		logger.Info("deploy: iam roles for CloudSecurity exist (not creating)")
+	}
 }
 
 func registerPantherAccount(awsSession *session.Session, pantherAccountID string) {
+	logger.Info("deploy: registering account with Panther")
```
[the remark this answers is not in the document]
rleighton: Good idea.

tools/mage/teardown.go
```diff
@@ -197,6 +204,129 @@ func destroyCfnStacks(awsSession *session.Session) error {
...
+// The onboard stack has global IAM roles, ask user if they want to delete since this could affect other regions
+func deleteOnboardStack(awsSession *session.Session, results chan deleteStackResult) {
+	logger.Infof("deleting CloudFormation stack: %s", onboardStack)
+	cfClient := cloudformation.New(awsSession)
+	onboardStackExists, err := stackExists(cfClient, onboardStack)
+	if err != nil {
+		logger.Fatalf("error checking stack %s: %v", onboardStack, err)
+	}
+	if onboardStackExists {
+		auditRoleExists, err := roleExists(iam.New(awsSession), auditRole)
```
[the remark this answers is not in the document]
rleighton: That case cannot happen since it is this code that deploys it.

tools/mage/teardown.go
rleighton: Right now users can't do anything with this. This stack set is deployed per region with 1 stack instance in that region. Users should not be changing this stack set since it is created by `mage deploy`

deployments/compliance/snapshot.yml
```diff
@@ -40,7 +40,6 @@ Parameters:
 
   AuditRoleName:
     Type: String
-    Default: PantherAuditRole
```
austinbyers: 👍

tools/mage/build_namespace.go
```diff
@@ -120,7 +121,8 @@ func (b Build) Lambda() {
 		}
 	})
 
-	logger.Infof("build:lambda: compiling %d Go Lambda functions (internal/.../main)", len(packages))
+	logger.Infof("build:lambda: compiling %d Go Lambda functions using %s (internal/.../main)",
```
austinbyers: Nit: the paths relate to the compilation, not the Go version:

```suggestion
	logger.Infof("build:lambda: compiling %d Go Lambda functions (internal/.../main) using %s",
```

tools/mage/deploy_onboard.go
```diff
@@ -31,20 +32,42 @@ const (
...
+	if err != nil {
+		logger.Fatalf("error checking audit role name %s: %v", auditRole, err)
+	}
+	remediationRoleExists, err := roleExists(iamClient, remediationRole)
+	if err != nil {
+		logger.Fatalf("error checking remediation role name %s: %v", remediationRole, err)
+	}
 
-	deployRealTimeStackSet(awsSession, backendOutputs["AWSAccountId"])
+	if !auditRoleExists || !remediationRoleExists {
```
austinbyers: What if one role exists but not the other? Here you'll create the entire template again, which will still cause the error, right?

Maybe

```go
if !auditRoleExists && !remediationRoleExists
```

tools/mage/teardown.go
```diff
@@ -197,6 +204,129 @@ func destroyCfnStacks(awsSession *session.Session) error {
...
+			logger.Fatalf("error checking audit role name %s: %v", auditRole, err)
+		}
+		if auditRoleExists {
+			answer := promptUser("Do you want to delete CloudSecurity roles (this can affect deployments in other regions)? (yes|no) ",
+				nonemptyValidator)
+			if strings.ToLower(answer) == "yes" {
+				logger.Infof("deleting stack %s", onboardStack)
+				go deleteStack(cfClient, aws.String(onboardStack), results) // can be done in background
+			} else {
+				results <- deleteStackResult{stackName: onboardStack, err: errors.Errorf("skipping dropping stack %s", onboardStack)}
```
austinbyers: Returning an error here will cause the whole teardown to fail immediately (the main routine will log the result, which uses `logger.Fatalf` for errors).

I would just log a warning and then return a nil error

```suggestion
				logger.Warnf("skipping stack %s", onboardStack)
				results <- deleteStackResult{stackName: onboardStack, err: nil}
```

tools/mage/teardown.go
```diff
@@ -197,6 +204,129 @@ func destroyCfnStacks(awsSession *session.Session) error {
...
+			StackSetName: stackSet,
+			Accounts:     []*string{identity.Account},
+			Regions:      []*string{client.Config.Region},
+			RetainStacks: aws.Bool(false),
+		})
+		if err != nil {
+			// need to return 2 errors
+			results <- deleteStackResult{stackName: stackSetInstanceName, err: err}
+			results <- deleteStackResult{stackName: *stackSet,
+				err: errors.Errorf("stack set instance not empty")}
```
austinbyers: Use `fmt.Errorf` to avoid adding a new import we don't need

tools/mage/util.go
```diff
@@ -92,10 +98,67 @@ func getSession() (*session.Session, error) {
...
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NoSuchEntity" {
+			err = nil
+		}
+		return false, err
+	}
+	return true, nil
+}
 
+// Return true if CF stack exits
```
austinbyers: typo: `exists`

tools/mage/util.go
```diff
@@ -92,10 +98,67 @@ func getSession() (*session.Session, error) {
...
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ValidationError" {
+			err = nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+// Return true if CF stack set exits
```
austinbyers: typo: `exists`

tools/mage/util.go
```diff
@@ -92,10 +98,67 @@ func getSession() (*session.Session, error) {
...
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "StackSetNotFoundException" {
+			err = nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+// Return true if CF stack set exits
```
austinbyers: One more! `exists`

tools/mage/build_namespace.go
rleighton: done

tools/mage/teardown.go
rleighton: done

tools/mage/deploy_onboard.go
rleighton: Technically that can't happen, but sure.

rleighton: done

tools/mage/teardown.go
rleighton: sure, done

tools/mage/util.go
rleighton: fixed

tools/mage/util.go
rleighton: fixed

tools/mage/util.go
rleighton: fixed

Issue #360, opened 2020-03-06
`mage test:ci` does not compile Lambda functions


austinbyers:
### Description

`mage test:ci` used to compile the Lambda functions and now it does not. Even though `go test` is doing most of the work, we need to explicitly test compilation for two reasons:

- If changes are made to a Swagger file, we need to prove the generated Go SDK still works with the code
- If changes are made to the compilation flags, we need to prove it still works

### Acceptance Criteria

- Compile all Go Lambda functions as part of `mage test:go`

PR #382, opened 2020-03-10
Fix swagger generation

austinbyers:
## Background

`mage build:api` does not work for new API changes because of a strange new bug in `swagger`:

1. `rm -r api/gateway/compliance/client api/gateway/compliance/models`
1. `.setup/swagger generate client -q -t api/gateway/compliance -f api/gateway/compliance/models`
1. `mage build:lambda` now fails because `complianceClient.Operations.GetStatus` does not exist. 
1. The autogenerated `api/gateway/compliance/client/panther_compliance_client.go` now has the wrong import!! It imports `github.com/panther-labs/panther/api/gateway/ANALYSIS/client/operations` (without the caps) instead of `gateway/COMPLIANCE/client/operations`

The version of swagger we use hasn't changed, but most likely one of the libraries it depends on has. When troubleshooting this, I also realized the `swagger` generated SDK has changed slightly.

Closes: #360

## Changes

- Regenerate all swagger clients and models
- Fix `build:api` to work around swagger import bug - run from the same directory
- Format only `api` files when running `build:api` (a bug was causing the entire project to be reformatted)
- Add `build:lambda` to `test:go` (#360)

## Testing

- `mage build:api fmt test:ci`

### DIFF
--- a/api/gateway/resources/client/operations/operations_client.go
+++ b/api/gateway/resources/client/operations/operations_client.go
@@ -27,11 +27,11 @@ import (
 	"fmt"
 
 	"github.com/go-openapi/runtime"
-	strfmt "github.com/go-openapi/strfmt"
+	"github.com/go-openapi/strfmt"
 )
 
 // New creates a new operations API client.
-func New(transport runtime.ClientTransport, formats strfmt.Registry) *Client {
+func New(transport runtime.ClientTransport, formats strfmt.Registry) ClientService {
 	return &Client{transport: transport, formats: formats}
 }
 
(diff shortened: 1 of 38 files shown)


api/gateway/resources/client/operations/operations_client.go
```diff
@@ -43,8 +43,25 @@ type Client struct {
 	formats   strfmt.Registry
 }
 
+// ClientService is the interface for Client methods
+type ClientService interface {
```
austinbyers: This is new! Swagger clients now automatically provide interfaces, which may make testing easier (we were mocking at the HTTP layer previously)

internal/compliance/remediation_api/remediation/invoke.go
```diff
@@ -60,7 +60,7 @@ var (
 			WithHost(resourcesServiceHostname)
 	resourcesClient = resourcesclient.NewHTTPClientWithConfig(nil, resourcesConfig)
 
-	RemediationNotFound = &remediationmodels.Error{Message: aws.String("Remediation not associated with policy")}
+	ErrNotFound = errors.New("Remediation not associated with policy")
```
austinbyers: The autogenerated `Error` model no longer actually implements the `error` interface, so compilation broke until I made this an actual error

austinbyers:
Hmmm, if we regenerate the swagger files on every build, then you need `swagger` installed in order to deploy. I'll revert that part

PR #385, opened 2020-03-10
Improving robustness of alert generation mechanism

### LINKED ISSUE #346: Restrict Dedup string size
### Description

Currently dedup string is a string of arbitrary length. Given this is stored in DDB, it can actually be problematic since max item size for DDB is 400kb. 

### Acceptance Criteria

- Put a restriction on dedup string size
- Document appropriately

### LINKED ISSUE #350: AlertID should not depend on user input
### Description

For details see issue discussed here: https://github.com/panther-labs/panther/pull/334#discussion_r388340855

### Acceptance Criteria

- AlertID is not depending on dedup string
- AlertsAPI performs appropriate validation of the alertID given by the frontend.

kostaspap:
## Background

- DedupString is a string provided by the users. We were putting no restrictions around its size which could lead to performance and stability problems. Setting a limit of 1000 characters, should be enough to cover most usecases. Anything above 1000 characters is truncated. 
- Alert IDs are now strings of 32 hex digits. This addresses the issue discussed [here](https://github.com/panther-labs/panther/pull/334#discussion_r388340855). 

Closes #350 
Closes #346 
## Changes

- Dedupstring has max lenght 1000. Anything more than that is concatenated
- Added validations for alertID in the alerts-api. This helps avoid SQL injection attacks
- AlertID is now a fixed 32 hex digit string (MD5 hash). Previously it was a concatenation of (among other things) user provided input. 

## Testing

- Unit tests
- Deployed and verified I was able to retrieve alert with the new mechanism. 

### Backwards compabitility

This PR changes the alerts-api to validate the alertID (it now has to be a hex). This means that users will not be able to see alerts generated prior to this - alertst-api/frontend will throw an error as the old alertIDs will fail validation.


internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -81,12 +87,14 @@ def run(self, event: Dict[str, Any]) -> RuleResult:
             rule_result = _run_command(self._module.rule, event, bool)
             if rule_result and self._has_dedup:
                 dedup_string = _run_command(self._module.dedup, event, str)
+                if len(dedup_string) > MAX_DEDUP_STRING_SIZE:  # type: ignore
+                    dedup_string = dedup_string[:MAX_DEDUP_STRING_SIZE]  # type: ignore
```
rleighton: Any reason not to fail hard? This silently changes behavior. I would prefer a fail here.

kostaspap: If we fail, this event will not be considered as failing the rule (we run the dedup function only for events that matched a rule). It will never pop up in an alert. 

I think it is more safe behavior to just truncate it, so the event will be grouped with different set of events. The limit is pretty high in any case 1000 characters should be enough for most dedup scenarios. 

I plan to make sure it is clear the behavior to the users as well:
- Explain the behavior in the documentation
- When someone creates a new rule, we show a "sample" rule in UI. I plan to have a comment in the dedup function explaining the behavior.

austinbyers: > Any reason not to fail hard? This silently changes behavior. I would prefer a fail here.

I think the API should just reject dedup strings that are too long, you said it's provided by the user, right?

nhakmiller: My understanding is that the dedup string can be dynamically generated based on different properties of the policy/resource, so it is hard to know how long it will be before it runs. I'd be okay with truncating and maybe a warning message or something? Truncating is better than just dropping it or not doing any deduping.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -81,12 +87,14 @@ def run(self, event: Dict[str, Any]) -> RuleResult:
...
         except Exception as err:  # pylint: disable=broad-except
             return RuleResult(exception=err)
 
         # If users haven't specified a dedup function return a default value
         if rule_result and not dedup_string:
-            dedup_string = "default"
+            dedup_string = DEFAULT_DEDUP_STRING
```
nhakmiller: I thought the default dedup string would be the rule ID?

internal/log_analysis/rules_engine/src/rule.py
rleighton: I'm very worried about strange things happening here, ask Nick pointed out the dedupe can be dynamic, otherwise we could have API check it (and fail hard here as a precaution, but it should never happen). Given it is dynamic, we are opening up to strange things. The worse that happens is more deduping is done than is desired but that will be super hard to debug. Perhaps we add an output field that records the fact that this event had a truncated dedupe string. Thoughts?

austinbyers: After understanding it better, truncating seems reasonable, but any reason we can't go with 5k or 10k as the limit?

kostaspap: > The worse that happens is more deduping is done than is desired but that will be super hard to debug. Perhaps we add an output field that records the fact that this event had a truncated dedupe string. Thoughts?

Sure, I will add some logging

kostaspap: > After understanding it better, truncating seems reasonable, but any reason we can't go with 5k or 10k as the limit?

1K is a mostly arbitrary limit. Some thoughts I had about this: 
- 400kb is the maximum size of a DDB item. We definitely have to be smaller than that. 
- 1 DDB write request unit is 1KB. Performing 1 write in DDB with 2KB payload will consume (and cost) 2 "writes". If some customer ends up really using 1k characters for deduping, given how we write to DDB this is going to cost him 3 DDB write units. 
- I initially had a limit of 256 to make sure a request to DDB is never more than 1 write request unit for cost and performance reasons. This would cover a lot of usecases e.g. max file name size in linux is 255 characters, maxi domain name size is 253, etc. I thought however to extend it more since maybe some customers might want to concatenate such fields. Most systems that I've seen doing deduping had much smaller limit so I thought that would be enough.
- I consider 1k characters to be enough for deduping for the vast majority of the use cases. Deduping string would be a specific log field (hostname, ip, username, region, ....) or maybe a combination of those. 1k characters seems already too large and probably it is a misuse of the field.
 - If a customer wants to dedup on something larger, it will actually be more efficient for them to just use as dedup a hash (MD5, SHA1, ...) of the payload than putting the full dedup payload. It is going to be cheaper on their bill as well. The limit is there also for the protection of the customer in case of potential misuse.  

These are the thoughts that went through my mind more or less. 
If you think it's not sufficient for some usecases, I could raise it further.

internal/log_analysis/rules_engine/src/rule.py
kostaspap: I never though about it. I wanted to pick something that would make it very clear that no dedup string was defined and minimize conflicts with potential "real" dedups. I like the idea a lot though! Changed :)

PR #388, opened 2020-03-10
Fixed Setup and Teardown of Cloud Watch Events Stack Set

nhakmiller:
## Background

Currently, the `panther-onboard` stack fails to create any stack set instances due to the fact that there is no stack set admin role present. So if you deploy right now the stack set is created, but no instances of it are deployed. This PR fixes that.

Additionally,  the `mage teardown` command deletes the compliance IAM roles stack before deleting the stack set, which causes the stack set deletion to fail as the stackset admin and execution roles no longer exist. This is actually fairly annoying to fix manually, as you have to re-deploy both those stacks, manually delete the instance, then delete the stackset, then delete the stack set. This change also fixes this out of order teardown issue.

Finally, I added a toggle to the `panther_config` file to determine whether or not to onboard the Panther account as a source. It defaults to true because normally its ideal, but in some cases where more advanced management is desired it will not be desired because it will just have to be torn out anyways.

## Changes

- Re-ordered the mage teardown command to delete the stackset before deleting the IAM roles necessary to delete the stack set
- Added the `panther-stackset-iam-admin` stack to the `panther-onboard` stack
- Added some configuration options for toggling on/off the onboarding of the account running Panther to Panther
- Updated the auxiliary panther-stack-set template to remove unnecessary rules

## Testing

- Deployed/tore down several times in my dev account to first verify the incorrect behavior, then verify this code fixes it


nhakmiller:
> Since the panther-cloudwatch-events.yml changed, do you need to update user documentation at all?

I will need to publish the new one to s3, but the documentation is actually more in line with this new template than the old template. The things that were removed were never officially supported and just put there for future use with log analysis back when we had intended log analysis and cloud security to share a data ingestion pipeline. The documentation actually never mentioned the other configuration options for that reason.

tools/mage/config.go
```diff
@@ -42,12 +42,17 @@ type monitoringParameters struct {
 	AlarmSNSTopicARN string `yaml:"AlarmSNSTopicARN"` // where to send alarms (optional)
 }
 
+type onboardingParameters struct {
+	OnboardSelf bool `yaml:"OnboardSelf"` // where to send alarms (optional)
```
austinbyers: comment copy-pasta is not accurate pasta

tools/mage/teardown.go
```diff
@@ -184,14 +184,19 @@ func destroyCfnStacks(awsSession *session.Session, identity *sts.GetCallerIdenti
...
 	handleResult(<-results)
 
-	// this one may ask user
+	// The stackset must be deleted before the StackSetExecutionRole
+	go deleteRealTimeEventStack(awsSession, identity, results)
+	// deleteRealTimeEventStack sends two results, one for the stack set instance and one for the stack set itself.
+	// We only need to block on the deletion of the stack set instance (the first result), we can delete the stack
+	// set itself in parallel with the rest of the delete operations.
+	handleResult(<-results)
+	nStacks := 1
```
austinbyers: This `nStacks` is confusing (I know you didn't add it)

Let's just block to finish these two sequential deletes and get rid of `nStacks` completely. Some stacks are sequential, some are parallel, let's not mix-and-match

nhakmiller: Sure, I will refactor so that first all the serial stacks are deleted, then all the parallel stacks are deleted, no mix & matching.

PR #390, opened 2020-03-10
 Re-implement glue:sync to not delete/add but to update just the schema in a partition

rleighton:
## Before
mage glue:sync would delete all the partitions for a table and recreate them inheriting the table's new schema (used when schemas change). 

This was fine,  but we will soon have both JSON and Parquet partitions in a table and we do not want to replace the whole partition since the type and S3 path will NOT be inherited from the table.

## After
Now each partition is fetched , the schema changed (but no other elements) and the partition updated.

## Changes
* moved most of the logic into SyncPartitions()
* does an update of just schema, leave other elements unchanged.

## Testing
mage test:ci
mage test:integration
mage glue:sync


pkg/awsglue/glue.go
```diff
@@ -151,19 +152,81 @@ func GetTableName(logType string) string {
...
 
-// SyncPartition deletes and re-creates a partition using the latest table schema. Used when schemas change.
-func (gm *GlueTableMetadata) SyncPartition(client glueiface.GlueAPI, t time.Time) error {
-	_, err := gm.deletePartition(client, t)
-	if err != nil {
-		if awsErr, ok := err.(awserr.Error); !ok || awsErr.Code() != "EntityNotFoundException" {
-			return errors.Wrapf(err, "delete partition for %s.%s at %v failed", gm.DatabaseName(), gm.TableName(), t)
-		}
+// SyncPartitions updates a table's partitions using the latest table schema. Used when schemas change.
+func (gm *GlueTableMetadata) SyncPartitions(client glueiface.GlueAPI) (failed error) {
```
austinbyers: Any reason we can't call this variable `err` like all the others?

pkg/awsglue/glue.go
```diff
@@ -151,19 +152,81 @@ func GetTableName(logType string) string {
...
+				}
+
+				getPartitionInput := &glue.GetPartitionInput{
+					DatabaseName:    aws.String(gm.databaseName),
+					TableName:       aws.String(gm.tableName),
+					PartitionValues: gm.partitionValues(update),
+				}
+				getPartitionOutput, err := client.GetPartition(getPartitionInput)
+				if err != nil { // skip partitions not there
+					if awsErr, ok := err.(awserr.Error); !ok || awsErr.Code() != "EntityNotFoundException" {
```
kostaspap: Replace string with `glue.EntityNotFoundException`

rleighton: Done

pkg/awsglue/glue.go
rleighton: Shadowing can cause bugs, so I name the "outer" error failed and inner locally scoped errors err.

PR #396, opened 2020-03-11
Fixed generated schema for dedup

alexmylonas:
## Background

Needed to update generated schema for changes in dedup 
## Changes

- Regenerated web schema
- Changed codegen.json to codegen.yml
- Added package for codegen dependency

## Testing

- Locally


austinbyers:
We can add a check to CI to fail if the generated web schema is out of date (I'll make a note)

kostaspap:
I did some changes on this few days ago. 
What is the recommended way for me going forward to update the GraphQL schema when I modify an existing API? @alexmylonas

alexmylonas:
@kostaspap It seems like `mage test:ci` doesn't catch this issue right now. We 'll need to add this to the CI probably to check if files changing with `npm run generate-gql` and either fail or commit them. @3nvi thoughts?

austinbyers:
> We 'll need to add this to the CI probably to check if files changing with npm run generate-gql and either fail 

Agreed - that's my plan. Fail CI if you try to update the graphQL schema without also re-generating the web source. I'll also add `npm run generate-gql` to the `mage build:api` target

austinbyers:
I went ahead and merged so we could get this fix into our deployments

PR #401, opened 2020-03-12
Update glue:sync to take start date and add glue:update

rleighton:
## Before
Glue sync would only update partitions in a table _after_ the create time of the table. Also, if a partition for a time period did not exist, it skipped  updating that time period.

## The Need
We want to be able "restore" reference to data in S3 before the create time of the table. This can happen if we drop the cf template that created the tables.

## After
Glue:sync now prompts the user for date to start creating partitions, if none is supplied it will default to the create time of the table. Also, if a partition for a time period does not exist, S3 is checked for at least 1 object of size > 0 , and if that exists, a partition is created. This last change is needed to backfill partitions after the database cloudformation  is dropped.

A new target was added to mage called glue:update which will update/create the database cloudformation. This is useful to avoid a full deployment if only the database schemas change or if you drop the database cloudformation template.

## Changes
* Added prompt for data in glue:sync
* Added target glue:update to mage
* Added mage util describeStack()
* Added S3 check to glue partition code
* Updated all tests

## Testing
mage test:ci
mage test:integration
ran mage glue:update glue:sync in dev account and security account


nhakmiller:
Looks like your master branch has a whole bunch of stuff on it, you may want to do a `git reset --hard origin/master` on your master branch after this.

pkg/awsglue/glue.go
```diff
@@ -282,6 +299,33 @@ func (gm *GlueTableMetadata) deletePartition(client glueiface.GlueAPI, t time.Ti
...
+	}
+
+	// list files w/pagination
+	inputParams := &s3.ListObjectsV2Input{
+		Bucket:  aws.String(location.Host),
+		Prefix:  aws.String(gm.GetPartitionPrefix(t)),
+		MaxKeys: aws.Int64(1), // look for at least 1
+	}
+	var hasData bool
+	err = client.ListObjectsV2Pages(inputParams, func(page *s3.ListObjectsV2Output, morePages bool) bool {
```
austinbyers: I know we're not using the `morePages` variable, but it has the opposite meaning - name it `lastPage` or `isLast` - the value is true if this is the last page of processing

tools/mage/glue_namespace.go
```diff
@@ -33,31 +36,56 @@ import (
 // targets for managing Glue tables
 type Glue mg.Namespace
 
+// Update Updates the panther-app-databases cloudformation template (used for schema migrations)
+func (t Glue) Update() {
```
austinbyers: Should this be part of the `deploy` command? It seems like its implementation is a direct subset of what `deploy` would do. E.g. `mage deploy:databases` or `mage deploy:glue` or something

This is fine too, just throwing an idea out there

tools/mage/glue_namespace.go
```diff
@@ -33,31 +36,56 @@ import (
...
+	_, bucketStackOutput, err := describeStack(cfClient, bucketStack)
+	if err != nil {
+		logger.Fatal(err)
+	}
+
+	status, backendStackOutput, err := describeStack(cfClient, backendStack)
+	if err != nil {
+		logger.Fatal(err)
+	}
+	if status != "CREATE_COMPLETE" && status != "UPDATE_COMPLETE" {
```
austinbyers: Use `cloudformation.StackStatusCreateComplete` and `cloudformation.StackStatusUpdateComplete` constants (from https://docs.aws.amazon.com/sdk-for-go/api/service/cloudformation/#pkg-constants)

tools/mage/util.go
```diff
@@ -157,6 +159,20 @@ func stackSetInstanceExists(cfClient *cloudformation.CloudFormation, stackSetNam
...
 
+func describeStack(cfClient *cloudformation.CloudFormation, stackName string) (status string, output map[string]string, err error) {
+	input := &cloudformation.DescribeStacksInput{StackName: &stackName}
+	response, err := cfClient.DescribeStacks(input)
+	if err != nil {
+		return status, output, err
+	}
+
+	status = *response.Stacks[0].StackStatus
+	if status == "CREATE_COMPLETE" || status == "UPDATE_COMPLETE" {
```
austinbyers: Same here (use the cfn constants)

tools/mage/glue_namespace.go
rleighton: I tried that but if I change deploy to a name space then all the docs need to change. Perhaps when you overhaul mage you can reorganize.

tools/mage/glue_namespace.go
rleighton: done

tools/mage/util.go
rleighton: done

PR #418, opened 2020-03-13
osquery:status - make logType field optional

rleighton:
## Changes
* make LogType optional
* add test for events currently failing

## Testing
mage test:ci


internal/log_analysis/log_processor/parsers/osquerylogs/status.go
```diff
@@ -36,7 +36,7 @@ type Status struct { // FIXME: field descriptions need updating!
 	Filename          *string                `json:"filename,omitempty" validate:"required" description:"Filename"`
 	HostIdentifier    *string                `json:"hostIdentifier,omitempty" validate:"required" description:"HostIdentifier"`
 	Line              *int                   `json:"line,omitempty,string" validate:"required" description:"Line"`
-	LogType           *string                `json:"logType,omitempty" validate:"required,eq=status" description:"LogType"`
+	LogType           *string                `json:"logType,omitempty"  description:"LogType"`
```
kostaspap: `validate:"omitempty,eq=status"` might be more appropriate?

rleighton: I thought about that but Filename keeps this unique within the osquery types and if we get this in another log type we just have to do another PR. Did not seem worth it right now.

PR #420, opened 2020-03-13
Allowing users to customize alert title using Python

### LINKED ISSUE #238: Allow users to create custom alert title
### Description

This feature allows users to customize the Alert title in Python. 

### Acceptance Criteria

- Rules engine is supporting execution of a new method
- Alert delivery mechanism is delivering the new message

kostaspap:
## Background

Allowing users to customize alert title using Python

Closes #238 

## Changes

- Rule's can now have an additional `title` method. When a rule matches an event, we run the "title" method to specify the title of the alert. 
- Added unit tests
- web was failing on `npm audit` command. Did `mage setup:web` and that modified the `package-lock.json` and now it passes

## Testing

- Unit tests
- Deployed to my account. A rule with the following body: 
Screen Shot 2020-03-13 at 6 24 02 PM

Generated the following alert: 

Screen Shot 2020-03-13 at 6 24 21 PM

### DIFF
--- a/internal/core/alert_delivery/models/api.go
+++ b/internal/core/alert_delivery/models/api.go
@@ -61,4 +61,7 @@ type Alert struct {
 
 	// Type specifies if an alert is for a policy or a rule
 	Type *string `json:"type,omitempty" validate:"omitempty,oneof=RULE POLICY"`
+
+	// Title is the optional title for the alert
+	Title *string `json:"title,omitempty"`
 }
--- a/internal/log_analysis/alert_forwarder/forwarder/forwarder_test.go
+++ b/internal/log_analysis/alert_forwarder/forwarder/forwarder_test.go
@@ -83,6 +83,7 @@ var (
 		Severity:            "INFO",
 		EventCount:          100,
 		LogTypes:            []string{"Log.Type.1", "Log.Type.2"},
+		Title:               aws.String("test title"),
 	}
 
 	testRuleResponse = &models.Rule{
@@ -129,7 +130,6 @@ func TestStoreDDBError(t *testing.T) {
 	assert.Error(t, Store(testAlertDedupEvent))
 }
 
-// The handler signatures must match those in the LambdaInput struct.
 func TestSendAlert(t *testing.T) {
 	sqsMock := &mockSqs{}
 	sqsClient = sqsMock
@@ -152,6 +152,7 @@ func TestSendAlert(t *testing.T) {
 		Tags:              aws.StringSlice([]string{"Tag"}),
 		Type:              aws.String(alertModel.RuleType),
 		AlertID:           aws.String("8c1b7f1a597d0480354e66c3a6266ccc"),
+		Title:             aws.String("test title"),
 	}
 	expectedMarshaledEvent, err := jsoniter.MarshalToString(expectedAlert)
 	require.NoError(t, err)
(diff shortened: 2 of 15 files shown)


internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -76,27 +80,59 @@ def __init__(self, rule_id: Optional[str], rule_body: Optional[str], rule_severi
...
-            dedup_string = self.rule_id
-        return RuleResult(matched=rule_result, dedup_string=dedup_string)
+        return RuleResult(matched=rule_result, dedup_string=dedup_string, title=title)
+
+    def _get_dedup(self, event: Dict[str, Any]) -> str:
+        if not self._has_dedup:
+            # If no dedup function defined, return rule id
+            return self.rule_id
+        dedup_string = _run_command(self._module.dedup, event, str)
+        if dedup_string and dedup_string != '':
```
austinbyers: These mean essentially the same thing: `if dedup_string` would be False if the string were empty. So you can simplify to:

```python
if dedup_string:
```

internal/log_analysis/alert_forwarder/forwarder/forwarder_test.go
```diff
@@ -165,6 +166,54 @@ func TestSendAlert(t *testing.T) {
...
+func TestSendAlertWithoutTitle(t *testing.T) {
+	sqsMock := &mockSqs{}
+	sqsClient = sqsMock
+
+	mockRoundTripper := &mockRoundTripper{}
+	httpClient = &http.Client{Transport: mockRoundTripper}
+	policyConfig = policiesclient.DefaultTransportConfig().
+		WithHost("host").
+		WithBasePath("path")
+	policyClient = policiesclient.NewHTTPClientWithConfig(nil, policyConfig)
```
rleighton: I find the conflation of policies and rules confusing. After V1 I would like to suggest we merge the alarming infrastructure/code which will give us history and make the code easier to understand.

internal/log_analysis/rules_engine/src/alert_merger.py
```diff
@@ -35,13 +36,28 @@
...
+_ALERT_LOG_TYPES = 'logTypes'
+_ALERT_TITLE = 'title'
 
 # TODO Once rules store alert merge period, retrieve it from there
 # Currently grouping in 1hr periods
 _ALERT_MERGE_PERIOD_SECONDS = 3600
 
 
+# pylint: disable=too-many-instance-attributes
+@dataclass
```
rleighton: Nice clean up

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -76,27 +80,59 @@ def __init__(self, rule_id: Optional[str], rule_body: Optional[str], rule_severi
...
+            return dedup_string
+        # If dedup string was the empty string, put the default value (rule_id)
+        return self.rule_id
+
+    def _get_title(self, event: Dict[str, Any]) -> Optional[str]:
+        if not self._has_title:
+            return None
+
+        title_string = _run_command(self._module.title, event, str)
+        if title_string and title_string != '':
```
rleighton: Same as Austin's comment above

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -76,27 +80,59 @@ def __init__(self, rule_id: Optional[str], rule_body: Optional[str], rule_severi
...
+            if len(title_string) > MAX_TITLE_SIZE:
```
rleighton: What I have done before for this kind of use  case is rather than a hard truncate (which displays poorly), you truncate MAX_TITLE_SIZE - len("... (truncated)"), then add "...(truncated)" to the title after clipping so it is clear there was more and it was just clipped by the system.

kostaspap:
> Can you also update the documentation related to writing rules?

Yup yup. I will handle it as part of #430

internal/core/alert_delivery/models/api.go
```diff
@@ -61,4 +61,7 @@ type Alert struct {
 
 	// Type specifies if an alert is for a policy or a rule
 	Type *string `json:"type,omitempty" validate:"omitempty,oneof=RULE POLICY"`
+
+	// Title is the optional title for the alert
+	Title *string `json:"title,omitempty"`
```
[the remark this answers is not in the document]
kostaspap: Not yet but you are right we should consider. We have been discussing about merging rules-engine with policy-engine after V1 (analysis-engine?), I think this can be done in a way that it provides feature parity for both engines but also allows us to handle resource history and resource/log pivoting. 
If that merging takes long, we can just update policy-engine to have the same functionality. Should be a small change on the policy-engine side.

internal/log_analysis/rules_engine/src/rule.py
kostaspap: Huh, nice! Fixed

internal/log_analysis/rules_engine/src/rule.py
kostaspap: I like it. Updated and added test.

PR #422, opened 2020-03-14
Update osquery parsers to allow ints as json strings or ints

rleighton:
## Background
There seems to have been changes in osquery since we developed the original parsers.
Things that should be ints used to be strings, they fixed that and are now ints. 
LogType seems to not be used.

## Changes
* Add a new type to parse json that is either string or int gracefully so we are backwards compatible with older versions
* Make log type optional

## Testing
* mage test:ci
* deployed to security account and it _seems_ to be working. Will know by Monday. I'll leave this PR open until then


internal/log_analysis/log_processor/parsers/numerics/numerics.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	return strconv.Itoa((int)(*i))
+}
+
+func (i *Integer) MarshalJSON() ([]byte, error) {
+	return ([]byte)(i.String()), nil
+}
+
+func (i *Integer) UnmarshalJSON(jsonBytes []byte) (err error) {
+	cleanBytes := strings.Replace((string)(jsonBytes), `"`, "", -1) // remove quotes
```
austinbyers: We just need to remove quotes from either side, right?

```suggestion
	trimmed := strings.Trim(string(jsonBytes), `"`))
```

rleighton: Done

PR #424, opened 2020-03-14
Whitelist ConfirmForgotPassword in aws event processor

rleighton:
## Changes
* white list ConfirmForgotPassword  from cognito
* improve error logging

## Testing
mage test:ci


internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -270,24 +273,26 @@ type CloudTrailMetadata struct {
...
 	// If this is an ignored event, immediately halt processing
 	if isIgnoredEvent(eventName.Str) {
 		zap.L().Debug("ignoring read only event",
-			zap.String("evenSource", eventSource.Str),
+			zap.String("evenSource", detail.Get("evenSource").Str), // best effort to add context
 			zap.String("eventName", eventName.Str))
 		return nil, nil
 	}
 
+	eventSource := detail.Get("eventSource")
```
austinbyers: You have to fetch this variable in all of the cases above, not sure the value of moving the var down here

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -95,6 +95,9 @@ var (
 		"CreateLogStream":      {},
 		"FilterLogEvents":      {},
 
+		// cognitio
```
nhakmiller: *cognito

internal/compliance/aws_event_processor/processor/process.go
nhakmiller: It's a very minor thing, but I think that it's actually very unlikely for the `eventSource` to not exist, but it's very common for the `eventName` to be read only. So we're technically saving ourselves an `eventSource.Exists()` call in the most common case, at no real cost.

austinbyers: got it

rleighton: It also save an annoying warning in the logs if the source is whitelisted. That is the main reason I did it.

internal/compliance/aws_event_processor/processor/process.go
rleighton: Ooph! Good catch.

PR #427, opened 2020-03-15
Optimize log processor

rleighton:
## Background
During stress tests csv parsing and s3 writing were identified as opportunities for optimization.
Performance was improved 22% on a csv benchmark with these changes.

## Changes
1) run sendData() as go routine to unblock processing and hide network latency
2) create csvstream object and use for all delimited parsers to improve performance
3) Allow users to configure larger memory sizes for log processor lambda
4) raise timeout on panther-datacatalog-updater

## Testing
mage test:ci
stress tests in my dev account


internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -84,11 +85,14 @@ type S3Destination struct {
 // It continuously reads events from outputChannel, groups them in batches per log type
 // and stores them in the appropriate S3 path. If the method encounters an error
 // it writes an error to the errorChannel and continues until channel is closed (skipping events).
+// The sendData() method is called as go routine to allow processing to continue and hide network latency.
 func (destination *S3Destination) SendEvents(parsedEventChannel chan *parsers.PantherLog, errChan chan error) {
 	// used to flush expired buffers
 	flushExpired := time.NewTicker(destination.maxDuration)
 	defer flushExpired.Stop()
 
+	var sendWaitGroup sync.WaitGroup
```
kostaspap: The "SendEvents" is already running as a separate process. So we have: 
* One process per input S3 file. Each processor reads from S3 and parses logs. I'll name these processes Readers. 
* One process for writing to S3 and sending SNS notifications (this one). I'll name this process Writer. 

The Readers communicate to the Writer through a buffered channel. The channel is buffered so that if the Writer is falling behind the Readers can continue reading but also to make sure that if the Writer falls much behind backpressure is applied and we don't run into memory issues. 

My problem with this change is that it essentially removes the backpressure. If Writer I/O is taking long, we will just keep reading from S3 and parsing logs and loading the memory. So in case of some increased PUT latencies from S3 side we are just going to crash. This is especially a problem because Writer I/O behavior seems to be bursty (on every tick we do a lot of I/O)

Some thoughts:
* The Writer is likely working in small bursts so it might block reading for that period. We can increase the channel buffer to e.g. 10,000 or 100,000 (if log is e.g. 1KB, this would mean 10MB or 100MB buffer). That could help with performance while allowing us to still keep the benefits of backpressure. 
* If we want to keep the async approach , we should limit the max number of "sendData" routines that can be active at the same time as a backpressure mechanism. If "sendData" is slow, Writer will stop trying to "sendData", it will stop reading data from the Readers, Readers will pause. 
* Backpressure is no required for the sends we do in line 160, there we can just go as fast as possible. Is it possible that in your tests that is were you got issues? If that's the case, maybe you can just perform async I/O only in that part and leave the rest of the code as it was (maybe with greater channel buffer). 
* Something else we have to fix at some point: We actually do "byte size" checks on the buffer level. We should actually be doing it in bufferSet level.. The reason we do the bufferSize checks it to avoid memory issues, but in case of one log source with high volume, this could result in non optimal performance while in cases of many log types with lots of data, we could run into memory issues. That we can fix in another PR or this one if you want (should be an easy fix). This can probably increase performance as it will limit I/O operations to the absolutely necessary.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -152,29 +156,33 @@ func (destination *S3Destination) SendEvents(parsedEventChannel chan *parsers.Pa
...
 
-	zap.L().Debug("finished sending messages", zap.Int("events", eventsProcessed))
+	sendWaitGroup.Wait() // wait until all writes to s3 are done
+
+	zap.L().Debug("finished sending s3 files", zap.Int("events", eventsProcessed))
 }
 
 // sendData puts data in S3 and sends notification to SNS
-func (destination *S3Destination) sendData(buffer *s3EventBuffer) (err error) {
+func (destination *S3Destination) sendData(buffer *s3EventBuffer, errChan chan error) {
```
kostaspap: An idea on how to write this differently: 

* Make sendWaitGroup a variable of the destination struct (it actually kinda is already)
* Modify "sendData" method to start like this: 
```
func (destination *S3Destination) sendData(buffer *s3EventBuffer, errChan chan error) {
    destination.sendWaitGroup.Add(1)
    defer sendWaitGroup.Done()
    ..... send logic....
}
```
Then your apply methods could look like this: 

```
_ = bufferSet.apply(func(buffer *s3EventBuffer) error {
    bufferSet.unlinkBuffer(buffer)
    go destination.SendData(buffer, errChan)
}
```

internal/log_analysis/log_processor/parsers/csvstream/csvstream.go
```diff
@@ -0,0 +1,55 @@
...
+func (scr *StreamingCSVReader) Read(b []byte) (n int, err error) {
+	n = copy(b, scr.logLine)
+	if n < len(scr.logLine) { // partial copy
+		scr.logLine = scr.logLine[n:] // the rest for next call
+	} else { // full copy
+		scr.logLine = ""
+		if n < len(b) { // terminate line if there is room
+			b[n] = '\n'
+			n++
+		}
```
kostaspap: If I understand correctly we need the `\n` so that the CSVReader knows when the record ends, correct? 
If that's the case, we could just change line 53 to 
```
scr.logLine = log + "\n"
```

internal/log_analysis/log_processor/parsers/csvstream/csvstream.go
```diff
@@ -0,0 +1,55 @@
+package csvstream
```
kostaspap: Nice!

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -84,11 +85,14 @@ type S3Destination struct {
 // It continuously reads events from outputChannel, groups them in batches per log type
 // and stores them in the appropriate S3 path. If the method encounters an error
 // it writes an error to the errorChannel and continues until channel is closed (skipping events).
+// The sendData() method is called as go routine to allow processing to continue and hide network latency.
```
kostaspap: Previous changes caused regression. 
In line 58 we have : 
```
maxFileSize = 100 * 1000 * 1000 // 100MB uncompressed file size, should result in ~10MB output file size
	// it should always be greater than the maximum expected event (log line) size
```

We have 100MB as "uncompressed" file size. But now we are actually comparing this against the "compressed" data. We should change that line to 10MB. Writing one 100MB file to S3 is highly inefficient as we do it now and not recommended - we should be doing it using multi-part uploads and not with PutObject operation.

deployments/log_analysis/log_processor.yml
```diff
@@ -32,6 +32,11 @@ Parameters:
     Type: CommaDelimitedList
     Description: List of base LayerVersion ARNs to attach to every Lambda function
     Default: ''
+  LogProcessorLambdaMemorySize:
+    Type: Number
+    Description: Size of lambda for log processor
+    Default: 512
+    AllowedValues: [512, 1024, 1536, 2048, 3008]
```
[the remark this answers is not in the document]
rleighton: You are right.  I'm going to change the threshold used for ejection to be dynamic based on this choice. I want to see how small I can get down to.

deployments/log_analysis/log_processor.yml
```diff
@@ -270,7 +275,7 @@ Resources:
       Layers: !If [AttachLayers, !Ref LayerVersionArns, !Ref 'AWS::NoValue']
       MemorySize: 128
       Runtime: go1.x
-      Timeout: 30
+      Timeout: 120
```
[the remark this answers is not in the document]
rleighton: This Glue being a PITA. Once every few weeks (my guess is during a deployment) their API's get really, really slow. Changing the threshold here is just to avoid annoying alarms on the lambda timing out.

deployments/panther_config.yml
```diff
@@ -34,6 +34,23 @@ OnboardingParameterValues:
...
+  # a small memory size, however, for extremely large log volumes
+  # the larger sizes may be required. The relative speed of each
+  # size is below (as measured during benchmarks against
+  # the smallest size we allow).
+  #
+  LogProcessorLambdaMemorySize: 512 # 1.0
+  # LogProcessorLambdaMemorySize: 1024   # 2.2 X faster
+  # LogProcessorLambdaMemorySize: 1536   # 3.5 X faster
+  # LogProcessorLambdaMemorySize: 2048   # 4.3 X faster
+  # LogProcessorLambdaMemorySize: 3008   # 6.2 X faster
```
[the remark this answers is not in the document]
rleighton: Yup, I'm making the s3 ejection threshold dynamically calculated based in the lambda size. PR coming.

internal/log_analysis/log_processor/parsers/csvstream/csvstream.go
rleighton: I started with it like that but that costs 1 alloc per log line. Using an if statement is much cheaper. Tests have 100% code coverage, so I like this better.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Getting rid of this constant and dynamically calculating. PR coming after burn-in.

deployments/log_analysis/log_processor.yml
```diff
@@ -73,7 +78,7 @@ Resources:
       KmsMasterKeyId: !Ref SQSKeyId
       # Reference on KeyReuse: https://amzn.to/2ngIsFB
       KmsDataKeyReusePeriodSeconds: 3600 # 1 hour
-      VisibilityTimeout: 180 # Should match lambda
+      VisibilityTimeout: 360 # Should match lambda
```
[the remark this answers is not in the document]
rleighton: Can' read !Refs in CFN generation  :(

internal/log_analysis/log_processor/destinations/s3.go
rleighton: I've read-done all of this in the latest. Can you look/comment w..r.t. to the latest (sorry). Resolving this one.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Please look at latest. This has changed.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -279,13 +289,20 @@ func (bs s3EventBufferSet) getBuffer(event *parsers.PantherLog) *s3EventBuffer {
...
 	if !ok {
 		buffer = newS3EventBuffer(logType, hour)
-		_ = buffer.reset() // no need to check error
 		logTypeToBuffer[logType] = buffer
 	}
 
 	return buffer
 }
 
+func (bs s3EventBufferSet) unlinkBuffer(buffer *s3EventBuffer) {
```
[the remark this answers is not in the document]
rleighton: Done

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -54,18 +56,54 @@ const (
 
 	messageAttributeDataType = "String"
 
-	maxFileSize = 100 * 1000 * 1000 // 100MB uncompressed file size, should result in ~10MB output file size
-	// it should always be greater than the maximum expected event (log line) size
-
 	maxDuration = 1 * time.Minute //hHolding events for maximum 1 minute in memory
```
kostaspap: nit: Extra `h` in the beginning of the comment:
`// Holding ...`

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -54,18 +56,54 @@ const (
...
+		*/
+		largestAllInMemFileMB     = 45
+		processingExpansionFactor = 4
+		memoryFootprint           = largestAllInMemFileMB * processingExpansionFactor
+		minimumScratchMemMB       = 5 // how much overhead is needed to process a file
+	)
+	maxFileSizeMB := lambdaSizeMB - memUsedAtStartupMB - memoryFootprint - minimumScratchMemMB
+	if maxFileSizeMB < 5 {
+		panic(fmt.Sprintf("available memory too small for log processing, increase lambda size from %dMB", lambdaSizeMB))
+	}
```
kostaspap: * This seems problematic... The people will successfully deploy the software, onboard logs. They will see an alarm and will have to troubleshoot why the software is not working. We should just block them from deploying with anything less than the minimum. It seems that most of this is actually static values anyway. We can just define the minimum in the configuration and put this formula there in the comments. 
The only thing not known is `memusedAtStartupMB` - but this should be similar for all deployments I believe. We are probably talking about this being few MBs? We can just add some head room and define it statically rounded up to the closest 64MB (Lambda allows 64MB increments in memory)

* I think we can still run out of memory. We don't keep in memory just one file or `s3EventBuffer`. This used to be a lesser issue earlier because the only dimension of the s3BufferSet was `(log type)`. Now it is `(log type, hour bucket)`.

If we really want to be resilient, we actually need a different logic: 
- Receive event
- Write event to S3 buffer
- Updater the number of bytes in the *Buffer set*
- Check if number of bytes in the bufferSet has exceeded the limit. Write to S3 the s3 buffer with the biggest size
(I mentioned it in number 4 [here](https://github.com/panther-labs/panther/pull/427/files#r392907597))
Makes sense?

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -54,18 +56,54 @@ const (
 
 	messageAttributeDataType = "String"
 
-	maxFileSize = 100 * 1000 * 1000 // 100MB uncompressed file size, should result in ~10MB output file size
-	// it should always be greater than the maximum expected event (log line) size
-
 	maxDuration = 1 * time.Minute //hHolding events for maximum 1 minute in memory
+
+	bytesPerMB          = 1024 * 1024
+	maximumS3FileSizeMB = 100 // upper (not lower) bound on the size of the s3 files written (avoids really big files)
```
kostaspap: I've made similar [comment](https://github.com/panther-labs/panther/pull/427/files#r392940467) earlier. We probably want this to be 10MB (since it is gzipped). 100MB is a really big file to store in a single S3 PUT operation - we would need multi part upload. S3 doesn't like big file puts in a single PUT operations, latencies can vary significantly.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -84,11 +122,24 @@ type S3Destination struct {
...
+	// use a single go routine for safety/back pressure when writing to s3 concurrently with buffer accumulation
+	var sendWaitGroup sync.WaitGroup
+	sendChan := make(chan *s3EventBuffer) // unbuffered for back pressure (we want only sendData() in flight)
+	sendWaitGroup.Add(1)
+	go func() {
+		for buffer := range sendChan {
+			destination.sendData(buffer, errChan)
+		}
+		sendWaitGroup.Done()
+	}()
```
kostaspap: This is really cool solution! I really like the simplicity and efficiency of it. Love this change.

tools/cfngen/cloudwatchcf/alarms_lambda.go
```diff
@@ -81,7 +82,18 @@ func generateLambdaAlarms(resource map[interface{}]interface{}, config *Config)
...
+	const pantherLogProcessorLambda = "panther-log-processor"
+	if getResourceProperty("FunctionName", resource) == pantherLogProcessorLambda {
+		logProcessorMemory, err := strconv.ParseFloat(config.stackOutputs["LogProcessorLambdaMemorySize"], 32)
+		if err != nil {
+			panic("Cannot find MemeorySize for " + pantherLogProcessorLambda)
+		}
+		lambdaMem = (float32)(logProcessorMemory)
+	} else {
+		lambdaMem = getResourceFloat32Property(memorySizeKey, resource)
+	}
```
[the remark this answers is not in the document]
rleighton: added issue: https://github.com/panther-labs/panther/issues/435

internal/log_analysis/log_processor/destinations/s3.go
rleighton: This is true. I we really want to be robust, we need to track at the bufferset level. The above really only has any effect on small lambdas. On big ones we can afford a lot more as the buffers are gzip'd. I was hoping by doing some modeling we could get to using 128M lambdas but until we fix the CT JSON issue, that is too dangerous.

There are 3 places of large memory consumption (let's assume large CT files):

1) the in-memory file and parsed representation (until we do streaming JSON, we can't do anything about this and this is why I change "processor" to go 1 file at a time). This means we can ALWAYS OOM, no matter what else we do.

2) The accumulating s3buffers (this is gzip'd in a fragmented bytes.Buffer). There can be more than one in flight if there are more than one logtype or hour. These should be small, but we cannot control how many. I would guess most of the time each buffer stays below the max size and time thresholds, so they stay until the end of the processing. Switching to thresholding on total bytes in the BufferSet will help here.

3) The sendData() buffer , there will be only 1 at a time and it will be the gzip'd bytes from 1 s3Buffer. It will exist until S3 finishes writes (usually subsecond, at most a few seconds for a 10M file).

I'll tally at the BufferSet and make that change.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Hmmm., good point. I think we should switch to a file uploader api which does this. Let me look into that, I've done this before. You can wrap the bytes in the right interface (rand access reader) and it works. We can still adjust the size BTW but I think we should use the right api for this.

deployments/log_analysis/log_processor.yml
```diff
@@ -32,6 +32,12 @@ Parameters:
     Type: CommaDelimitedList
     Description: List of base LayerVersion ARNs to attach to every Lambda function
     Default: ''
+  LogProcessorLambdaMemorySize:
+    Type: Number
+    Description: Size of lambda for log processor
+    Default: 1024 # this size can handle the largest possible CloudTrail file (smaller will likely not work)
```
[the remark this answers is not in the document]
rleighton: Done

deployments/backend.yml
```diff
@@ -39,6 +39,10 @@ Parameters:
     Type: CommaDelimitedList
     Description: List of LayerVersion ARNs to attach to every function
     Default: ''
+  LogProcessorLambdaMemorySize:
+    Type: Number
+    Description: Size of lambda for log processor
+    Default: 512
```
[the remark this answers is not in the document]
rleighton: Removed

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Fixed

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Done

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Done

PR #431, opened 2020-03-16
Removed unused pictures

kostaspap:
## Background

I noticed that we have a lot of pictures that are not used. Did a cleanup removing few MBs of pictures. 

## Changes

- Removed pictures that are not referenced

## Testing

- Deleted pictures that were not referenced in the documents. Made use of IDE's "Safe Delete" functionality that verifies that the file is not referenced anywhere in the code/docs.


3nvi:
https://www.youtube.com/watch?v=ryCIAJRTwJg

PR #434, opened 2020-03-16
Optimistic response for Deletions

### LINKED ISSUE #406: Improve perceived speed by not refetching on Deletions
### Description

Currently, whenever we delete an entity, we ask Apollo to refetch all the queries related to this entity, so the UI is up to date. Instead of doing that, we can just remove it from the associated cache & while also adding an `optimisticResponse` so that this update becomes instant.

The problem is that data to delete can exist in multiple queries. In addition, if the same query has been asked with different parameters, then Apollo will store different cache entries for each different query+params combination. Of course, you can [override this behavior](https://www.apollographql.com/docs/react/v3.0-beta/caching/cache-field-behavior/#reducing-cache-duplicates-by-specifying-key-arguments), but that's not always what you want. Thus, you may need to remove an entry from:

* multiple cache entries from **different** queries 
* multiple cache entries that come from variations of the **same** query with different parameters 

To make that easy, Apollo recently released [cache eviction & garbage collection](https://www.apollographql.com/docs/react/v3.0-beta/caching/cache-interaction/#garbage-collection-and-cache-eviction). What this can do is remove fields from the cache and then garbage collect any un-needed pointers. For example,

```tsx
// deletes all stored variations of the graphQL query "users". After doing that all the User entries might
// not be needed to be in the cache anymore, so we run `cache.gc()` to remove any dangling data
cache.evict('ROOT_QUERY', 'users');

// removes the field `name` from a single User. Helpful in some situations, but not in ours 
cache.evict('User:1234', 'name'); 
```

Now those solutions help invalidate queries, but doesn't help when we don't want to remove the **whole** query, but instead modify its contents in order to remove the item that got deleted. That's where `cache.modify` comes into place. This allows you to modify [specific entries in the cache](https://github.com/apollographql/apollo-client/pull/5909) and is a generic cache utility function. With this, you can do the following:

```tsx
cache.modify('ROOT_QUERY`, {
  users: (data, helpers) => {
    const objToDelete = helpers.toReference({ __typename: 'User', id: deletedUser.id });
    return data.filter(d => d.__ref !== objectToDelete.__ref); 
  }
});

cache.gc();
``` 
This will make sure that **all variations** of the `users` query (if there are more than one) remove the deleted user from their cache. The UI update is done automatically for you. After that, the Deleted User object is not part oof any query, but its data are still in the cache without reason. We run `cache.gc()` to clean them up.

Sometimes, we want to only remove data from cache **specific variations** of a particular query. The `helpers` param contains a key `storeFieldName` which shows you the particular variation name. Thus you can do something like the following:

```tsx
cache.modify('ROOT_QUERY`, {
  users: (data, helpers) => {
    const objToDelete = helpers.toReference({ __typename: 'User', id: deletedUser.id });
    const shouldDeleteFromVariation = helpers.storeFieldName.includes('a particular param');
    
    if (shouldDeleteFromVariation) {  
      return data.filter(d => d.__ref !== objectToDelete.__ref); 
    }
   return data
  }
});

cache.gc();
``` 

We should make sure to try all of them out, dig into the docs and understand how caching works in Apollo; then attempt to optimize deletions. You can find related code [here](https://github.com/panther-labs/panther-enterprise/pull/16/files#diff-0a116a05dd43f67a869af7beeec878deR27-R34)

### Acceptance Criteria

**Every** deletion that happens in the system should update the cache entries through code instead of relying on `refetchQueries`. Thus:

- Replace the `refetchQueries` with manual code to modify the associated queries that were going to be refetched.
- No deletion should ever have a `refetchQueries` anymore
- Each deletion should never trigger any network request
- Each deletion should be instant (reflected in the UI)

alexmylonas:
## Background

Currently, whenever we delete an entity, we ask Apollo to refetch all the queries related to this entity, so the UI is up to date. Instead of doing that, we can just remove it from the associated cache & while also adding an `optimisticResponse` so that this update becomes instant.

Closes #406

## Changes

- Schema response change for Users

Optimistic response for:
- DeleteUserModal
- DeletePolicyModal
- DeleteRuleModal
- DeleteDestinationModal

## Testing

- Locally


jacknagz:
Hey, @alexmylonas would you mind adding a higher-level context on the change?

api/graphql/schema.graphql
```diff
@@ -48,7 +48,7 @@ type Query {
   organizationStats(input: OrganizationStatsInput): OrganizationStatsResponse
   rule(input: GetRuleInput!): RuleDetails
   rules(input: ListRulesInput): ListRulesResponse
-  users(limit: Int, paginationToken: String): ListUsersResponse
+  users: [User]
```
3nvi: we can add a `!` both at `User!` and at `[User!]!` since:

* we always expect a `User` object back (i.e. we will never have  an array containing a `null` valuee)
* we always expect an array to be reeturned (i.e. we will never have `null` instead of an array)

web/src/components/modals/DeleteDestinationModal/DeleteDestinationModal.tsx
```diff
@@ -32,7 +31,21 @@ const DeleteDestinationModal: React.FC<DeleteDestinationModalProps> = ({ destina
...
     },
-    refetchQueries: [{ query: ListDestinationsAndDefaultsDocument }],
+    optimisticResponse: {
+      deleteDestination: true,
+    },
+    update: async cache => {
+      cache.modify('ROOT_QUERY', {
+        destinations: (destinations, helpers) => {
+          const { __ref: destinationRef } = helpers.toReference({
+            __typename: 'Destination',
```
3nvi: I know that I showcased it differently in the related issue, but can we instead of `__typename: 'Destination'` write it as`__typename: destination.__typename` so that we don't have to add hardcoded strings?

3nvi: That applies to all places in this PR

web/src/components/modals/DeletePolicyModal/DeletePolicyModal.tsx
```diff
@@ -44,6 +40,21 @@ const DeletePolicyModal: React.FC<DeletePolicyModalProps> = ({ policy }) => {
         ],
       },
     },
+    optimisticResponse: {
+      deletePolicy: true,
+    },
+    update: async cache => {
```
3nvi: I don't think this works since `data` is of type `ListPoliciesResponse`. I think you should modify the `policies` from `ListPoliciesResponse` and not the `ROOT_QUERY`.

Can you validate that and test it?

3nvi: I also feel you should delete the related `PolicyDetails` from the cache as well, since this policy will not exist anymore both as a `PolicyDetails` and as a `PolicySummary`.

In general, check where those 2 types are used and make sure to update the queries in the proper place

web/src/components/modals/DeleteRuleModal/DeleteRuleModal.tsx
```diff
@@ -45,6 +41,21 @@ const DeleteRuleModal: React.FC<DeleteRuleModalProps> = ({ rule }) => {
         ],
       },
     },
+    optimisticResponse: {
+      deletePolicy: true,
```
3nvi: how can this work? you are deleting a rule here not a policy. Could it be a wrong hook import since the hook should bee `useRulePolicy`?

web/src/components/modals/DeleteRuleModal/DeleteRuleModal.tsx
```diff
@@ -45,6 +41,21 @@ const DeleteRuleModal: React.FC<DeleteRuleModalProps> = ({ rule }) => {
...
+    },
+    update: async cache => {
+      cache.modify('ROOT_QUERY', {
+        rules: (data, helpers) => {
+          const { __ref: ruleRef } = helpers.toReference({
```
3nvi: same thing. `ROOT_QUERY` doesn't have a field names `rules`. It only has a `ListRulesResponse` field (since `ROOT_QUERY` has as fields only the things that are present in the `type Query { ... }` in our schema).

web/src/components/modals/DeleteDestinationModal/DeleteDestinationModal.tsx
```diff
@@ -32,7 +31,21 @@ const DeleteDestinationModal: React.FC<DeleteDestinationModalProps> = ({ destina
...
       id: destination.outputId,
     },
-    refetchQueries: [{ query: ListDestinationsAndDefaultsDocument }],
+    optimisticResponse: {
+      deleteDestination: true,
+    },
+    update: async cache => {
+      cache.modify('ROOT_QUERY', {
+        destinations: (destinations, helpers) => {
+          const destinationRef = helpers.toReference({
```
3nvi: **Optional**: If you want to make it more concise, you can do `const destinationRef = helpers.toReference(destination)`.

This way you don't have to remember if the field is named `id` or `outputId` or `destinationId`, etc.

web/src/components/modals/DeleteRuleModal/DeleteRuleModal.tsx
3nvi: Update: Turns out  we use the same endpoint for deletions of rules & policies.  #464 should split them to be more clear

web/src/components/modals/DeleteUserModal/DeleteUserModal.tsx
```diff
@@ -38,8 +36,21 @@ const DeleteUserModal: React.FC<DeleteUserModalProps> = ({ user }) => {
...
     },
-    awaitRefetchQueries: true,
-    refetchQueries: [getOperationName(ListUsersDocument)],
+    optimisticResponse: {
+      deleteUser: true,
+    },
+    update: async cache => {
+      cache.modify('ROOT_QUERY', {
+        users: (data, helpers) => {
+          const userRef = helpers.toReference({
```
3nvi: Again, optional: `const userRef = helpers.toReference(user)`.

PR #439, opened 2020-03-17
New opstool: s3queue

rleighton:
## Background
There is a need operationally for customers to backfill data into log processing. This tool will list an S3 bucket and push processing notifications into log processing.

This tool is also useful for testing log processing as specific files can be pushed through the processing.

## Changes
* Added ops tool s3queue
* Refactored test code
* Added ability to cross compile so customers can run both on desktop and on ec2 instances.

## Testing
mage test:ci
I've been using this for load testing log processing.

### DIFF
--- a/cmd/opstools/s3queue/s3queue.go
+++ b/cmd/opstools/s3queue/s3queue.go
@@ -1,0 +1,176 @@
+package s3queue
+
+import (
+	"fmt"
+	"math"
+	"net/url"
+	"sync"
+
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3iface"
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"github.com/aws/aws-sdk-go/service/sqs/sqsiface"
+	jsoniter "github.com/json-iterator/go"
+	"github.com/pkg/errors"
+)
+
+const (
+	concurrency          = 10
+	pageSize             = 1000
+	fakeTopicArnTemplate = "arn:aws:sns:us-east-1:%s:panther-fake-s3queue-topic" // account is added for sqs messages
+)
+
+type Stats struct {
+	NumFiles uint64
+	NumBytes uint64
+}
+
+type cloudTrailNotification struct {
+	S3Bucket    *string   `json:"s3Bucket"`
+	S3ObjectKey []*string `json:"s3ObjectKey"`
+}
+
+func S3Queue(sess *session.Session, account, s3path, queueName string, limit uint64, stats *Stats) (err error) {
+	return s3Queue(s3.New(sess), sqs.New(sess), account, s3path, queueName, limit, stats)
+}
+
+func s3Queue(s3Client s3iface.S3API, sqsClient sqsiface.SQSAPI, account, s3path, queueName string,
+	limit uint64, stats *Stats) (failed error) {
+
+	queueURL, err := sqsClient.GetQueueUrl(&sqs.GetQueueUrlInput{
+		QueueName: &queueName,
+	})
+	if err != nil {
+		return errors.Wrapf(err, "could not get queue url for %s", queueName)
+	}
+
+	// the account id is taken from this  arn to assume role for reading in the log processor
+	topicARN := fmt.Sprintf(fakeTopicArnTemplate, account)
+
+	errChan := make(chan error)
+	notifyChan := make(chan *cloudTrailNotification, 1000)
+
+	var wg sync.WaitGroup
+	for i := 0; i < concurrency; i++ {
+		wg.Add(1)
+		go func() {
+			queueNotifications(sqsClient, topicARN, queueURL.QueueUrl, notifyChan, errChan)
+			wg.Done()
+		}()
+	}
+
+	wg.Add(1)
+	go func() {
+		listPath(s3Client, s3path, limit, notifyChan, errChan, stats)
+		wg.Done()
+	}()
+
+	go func() {
+		for err := range errChan { // return last error
+			failed = err
+		}
+	}()
+
+	wg.Wait()
+	close(errChan)
+
+	return failed
+}
+
+// Given an s3path (e.g., s3://mybucket/myprefix) list files and send to notifyChan
+func listPath(s3Client s3iface.S3API, s3path string, limit uint64,
+	notifyChan chan *cloudTrailNotification, errChan chan error, stats *Stats) {
+
+	if limit == 0 {
+		limit = math.MaxUint64
+	}
+
+	defer func() {
+		close(notifyChan) // signal to reader that we are done
+	}()
+
+	parsedPath, err := url.Parse(s3path)
+	if err != nil {
+		errChan <- errors.Errorf("bad s3 url: %s,", err)
+		return
+	}
+
+	if parsedPath.Scheme != "s3" {
+		errChan <- errors.Errorf("not s3 protocol (expecting s3://): %s,", s3path)
+		return
+	}
+
+	bucket := parsedPath.Host
+	if bucket == "" {
+		errChan <- errors.Errorf("missing bucket: %s,", s3path)
+		return
+	}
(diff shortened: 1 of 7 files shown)


cmd/opstools/s3queue/s3queue.go
```diff
@@ -0,0 +1,176 @@
...
+	})
+	if err != nil {
+		return errors.Wrapf(err, "could not get queue url for %s", queueName)
+	}
+
+	// the account id is taken from this  arn to assume role for reading in the log processor
+	topicARN := fmt.Sprintf(fakeTopicArnTemplate, account)
+
+	errChan := make(chan error)
+	notifyChan := make(chan *cloudTrailNotification, 1000)
```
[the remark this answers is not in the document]
kostaspap: Agree with Nick. no reason to just restrict it for CloudTrail. Better use "normal" S3 notifications to make it more generic.

cmd/opstools/s3queue/s3queue.go
```diff
@@ -0,0 +1,176 @@
...
+		}()
+	}
+
+	wg.Add(1)
+	go func() {
+		listPath(s3Client, s3path, limit, notifyChan, errChan, stats)
+		wg.Done()
+	}()
+
+	go func() {
```
kostaspap: I think we need to wait for this routine to finish before exiting the program.

cmd/opstools/s3queue/s3queue.go
```diff
@@ -0,0 +1,176 @@
...
+		ctnJSON, err := jsoniter.MarshalToString(cloudTrailNotification)
+		if err != nil {
+			errChan <- errors.Wrapf(err, "failed to marshal %#v", cloudTrailNotification)
+			return
+		}
+
+		// make it look like an SNS notification
+		snsNotification := events.SNSEntity{
+			Type:     "Notification",
+			TopicArn: topicARN,
```
kostaspap: I get why you add the Topic Arn here but you might want to add a comment in that regard. e.g. " We add fake topic ARNs since Log Processor expects it to be populated"

cmd/opstools/s3queue/s3queue.go
```diff
@@ -0,0 +1,176 @@
...
+			Message:  ctnJSON,
+		}
+		message, err := jsoniter.MarshalToString(snsNotification)
+		if err != nil {
+			errChan <- errors.Wrapf(err, "failed to marshal %#v", snsNotification)
+			return
+		}
+
+		sendMessageInput := &sqs.SendMessageInput{
```
kostaspap: You should use the BatchSend method. It's cheaper (less API calls) and faster. An easy way to make it the change would potentially making the `notifyChan ` contain array of notifications.

cmd/opstools/s3queue/s3queue.go
```diff
@@ -0,0 +1,176 @@
...
+				stats.NumBytes += (uint64)(*value.Size)
+				notifyChan <- &cloudTrailNotification{
+					S3Bucket:    &bucket,
+					S3ObjectKey: []*string{value.Key},
+				}
+				if stats.NumFiles >= limit {
+					break
+				}
+			}
+		}
```
kostaspap: Since this process can potentially take long, might make sense to post periodic updates to the screen e.g. every 1000 objects or so.

cmd/opstools/s3queue/s3queue/main.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/panther-labs/panther/cmd/opstools/s3queue"
+)
+
+const (
+	banner = "lists s3 objects and posts s3 notifications to log processor queue"
+)
+
+var (
+	REGION  = flag.String("region", "", "The AWS region where the queues and bucket exist (optional, defaults to session env vars).")
```
kostaspap: nitpick: I like to place `(optional)` in the beginning of a lengthy description (it might be broken in a new line in the console and people might miss it)

cmd/opstools/s3queue/s3queue.go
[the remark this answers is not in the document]
rleighton: It works with everything. If I recall, the reason I picked this notification is it was simpler (needed less  info). I'll look to at S3, just to not violate the principal of least surprise.

cmd/opstools/s3queue/s3queue.go
rleighton: I was on the fence about this ... I thought I put a comment in about that ... did not want to use batch to make debugging easier and if there is bad data we do not lose the whole set of files.

cmd/opstools/s3queue/s3queue.go
rleighton: Sure. Although it is very fast.

cmd/opstools/s3queue/s3queue.go
rleighton: You are right ... works most of the time but not guaranteed.

cmd/opstools/s3queue/s3queue.go
rleighton: I'll look into it. Makes the code a little more complex

kostaspap: If you send a batch and it fails, not everything fails actually: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html.

rleighton: Misspoke, I meant in the body of the message. I'll see about batching the sqs

cmd/opstools/s3queue/s3queue.go
```diff
@@ -0,0 +1,186 @@
...
+		for err := range errChan { // return last error
+			failed = err
+		}
+		errorWg.Done()
+	}()
+
+	queueWg.Wait()
+	close(errChan)
+	errorWg.Wait()
+
```
kostaspap: Can this lead to a deadlock? 
1. listPath routine starts listing s3 objects
2. All queueNotifications routines fail (SQS issues, permissions issues, wrong configuration). Noone is reading from the notifyChan 
3. listPath routine will block because the notifyChan is full and noone is reading from it? 


3.

cmd/opstools/s3queue/s3queue.go
austinbyers: for batching sqs, we have an [awsbatch](https://github.com/panther-labs/panther/tree/master/pkg/awsbatch) library that handles backoff and retrying failed messages and things

cmd/opstools/s3queue/s3queue.go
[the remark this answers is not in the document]
rleighton: Done

cmd/opstools/s3queue/s3queue.go
```diff
@@ -0,0 +1,176 @@
...
+	go func() {
+		for err := range errChan { // return last error
+			failed = err
+		}
+	}()
+
+	wg.Wait()
+	close(errChan)
+
+	return failed
```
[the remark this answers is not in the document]
rleighton: Fixed

cmd/opstools/s3queue/s3queue.go
rleighton: Fixed

cmd/opstools/s3queue/s3queue.go
rleighton: Done

cmd/opstools/s3queue/s3queue.go
rleighton: Fixed

cmd/opstools/s3queue/s3queue.go
```diff
@@ -0,0 +1,176 @@
...
+	limit uint64, stats *Stats) (failed error) {
+
+	queueURL, err := sqsClient.GetQueueUrl(&sqs.GetQueueUrlInput{
+		QueueName: &queueName,
+	})
+	if err != nil {
+		return errors.Wrapf(err, "could not get queue url for %s", queueName)
+	}
+
+	// the account id is taken from this  arn to assume role for reading in the log processor
```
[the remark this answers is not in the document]
rleighton: Fixed

cmd/opstools/s3queue/s3queue/main.go
rleighton: Fixed

PR #440, opened 2020-03-17
Modified rules-engine to use the dedup time period returned by the analysis-api

### LINKED ISSUE #348: Rules engine should use the dedup time interval configured for each rule
### Description

Rules engine currently dedups events for a fixed time interval (1hr). It should be modified to use the interval configured by the users. 

### Acceptance Criteria

- Rules engine updated to use the dedup time interval configured per rule

kostaspap:
## Background

Finalizing the backend changes that will make it possible to configure the dedup time period. 
Closes #348 

## Changes

- Modified the `GetEnabledPolicies` (used to return rules) to return the optional `dedupPeriodMinutes` property
- Modified rules-engine to use that property when doing the deduplication. In case no value is specified, it falls back to 60'. This gives us backwards compatibility with rules that have no value configured. 
- Updated the way Rule is initialized to avoid excessive number of constructor arguments (and make sure the python linter is not complaining)

## Testing

- Unit tests for new functionality
- Integration tests for new analysis-api functionality
- Deployed to my account and seen the new functionality working.

### DIFF
--- a/internal/log_analysis/rules_engine/src/__init__.py
+++ b/internal/log_analysis/rules_engine/src/__init__.py
@@ -17,6 +17,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from typing import Any, Dict, Optional
 
+
 # pylint: disable=too-many-instance-attributes
 @dataclass
 class EventMatch:
--- a/internal/log_analysis/rules_engine/src/__init__.py
+++ b/internal/log_analysis/rules_engine/src/__init__.py
@@ -17,7 +17,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from typing import Any, Dict, Optional
 
-
+# pylint: disable=too-many-instance-attributes
 @dataclass
 class EventMatch:
     """Represents an event that matched a rule"""
@@ -25,6 +25,7 @@ class EventMatch:
     rule_version: str
     log_type: str
     dedup: str
+    dedup_period_mins: int
     severity: str
     event: Dict[str, Any]
     title: Optional[str] = None
--- a/internal/log_analysis/rules_engine/src/output.py
+++ b/internal/log_analysis/rules_engine/src/output.py
@@ -115,7 +115,7 @@ class MatchedEventsBuffer:
 
 def _write_to_s3(time: datetime, key: OutputGroupingKey, events: List[EventMatch]) -> None:
     # 'severity', 'version', 'title', 'dedup_period' of a rule might differ if the rule was modified
-    # while the rules engine was running. We pick the first encountered values.
+    # while the rules engine was running. We pick the first encountered set of values.
     group_info = MatchingGroupInfo(
         rule_id=key.rule_id,
         rule_version=events[0].rule_version,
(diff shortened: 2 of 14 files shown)


internal/log_analysis/rules_engine/src/output.py
```diff
@@ -114,13 +114,14 @@ def flush(self) -> None:
 
 
 def _write_to_s3(time: datetime, key: OutputGroupingKey, events: List[EventMatch]) -> None:
-    # severity and version of a rule might differ if the rule was modified while the rules engine was running.
-    # We pick the first encountered severity, version and title.
+    # 'severity', 'version', 'title', 'dedup_period' of a rule might differ if the rule was modified
+    # while the rules engine was running. We pick the first encountered values.
```
rleighton: s/values/set of values

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -37,42 +37,61 @@
...
+DEFAULT_RULE_DEDUP_PERIOD_MINS = 60
+
 
 @dataclass
 class RuleResult:
     """Class containing the result of running a rule"""
     exception: Optional[Exception] = None
     matched: Optional[bool] = None
     dedup_string: Optional[str] = None
+    dedup_period_mins: Optional[int] = None
```
rleighton: Is RuleResult even needed? Can't run() just return an EventMatch?

internal/log_analysis/rules_engine/src/output.py
kostaspap: Fixed

internal/log_analysis/rules_engine/src/rule.py
kostaspap: `RuleResult` is the result of running a rule. So it has stuff like: did the rule match or not? Did the run succeed the run or not? And the results of any other rule-related functions (dedup, title). Prior to having the RuleResult, the `run` method was returning `Union[bool, Exception` and since we have also `dedup` and `title` functions, I created this struct. It's meant mostly to keep dynamically generated values not "rule configuration" parameters. 

But your comment is actually a good one: I misused the RuleResult here - I shouldn't add the `dedup_period_mins`  as a result since we are not running any python function to calculate it - it is part of the rule configuration, not dynamically calculated. Will update the code.

internal/log_analysis/rules_engine/src/__init__.py
```diff
@@ -18,13 +18,15 @@
 from typing import Any, Dict, Optional
 
 
+# pylint: disable=too-many-instance-attributes
```
austinbyers: I'm fine with removing this `pylint` check globally as well if you want

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -47,32 +51,46 @@ class RuleResult:
...
+        self.rule_id = config['id']
 
-        self._store_rule()
-        self._module = self._import_rule_as_module()
+        if not ('body' in config) or not isinstance(config['body'], str):
+            raise AssertionError('Field "body" of type str is required field')
+        self.rule_body = config['body']
+
+        if not ('severity' in config) or not isinstance(config['severity'], str):
+            raise AssertionError('Field "severity" of type str is required field')
```
austinbyers: Severity could have a reasonable default value too, though I think we require it in the API

kostaspap: Yeah, the check is there since the severity is actually required - if it's not it's actually an issue (it means we are not using it)

PR #442, opened 2020-03-17
Refactor users-api

austinbyers:
## Background

We completely rebuilt `users-api` in the enterprise edition, copying the same improved structure here to benefit everyone!

## Changes

- Delete `custom_message` Lambda function and move the functionality into `users-api` instead
    - This is used only when triggering the password reset flow, less infrastructure means faster tests and deploys
    - Its implementation is also more efficient - we don't have to ask Cognito for the user attributes when constructing the email, it already has them in the request
- API changes
    - Remove unused `onboarding` models
    - `InviteUser` can now invite users without sending the email, which we use in integration tests
    - All endpoints return non-empty structs with more information
    - `ListUsers` now lets you filter by name and Cognito status and sort by any attribute
- More comprehensive unit test coverage with more consistent mocking
- Added an integration test
- Renamed `gateway` package to `cognito` to more accurately describe what service it's talking to (it makes me think of API gateway otherwise)

## Testing

- `mage fmt test:ci`
- Deploy to dev account, `PKG=./internal/core/users_api/main mage test:integration`
- Triggered password reset flow, email was generated correctly, I was able to login with the new password


austinbyers:
`npm audit` is the only thing failing CI, I'm going to remove it - it shouldn't block all PRs when it's not related to changes in the PR

api/lambda/users/models/api.go
```diff
@@ -31,43 +41,105 @@ type LambdaInput struct {
...
-type ListUsersInput struct{}
+//
+// Example:
+// {
+//     "listUsers": {
+//         "contains": "austin"
+//     }
+// }
+type ListUsersInput struct {
+	// FILTERING
```
rleighton: I assume if you do Contains and Status filtering it is a logical and, right? Maybe add a comment explaining how these combine

internal/core/users_api/api/cognito_trigger.go
```diff
@@ -44,18 +44,37 @@ const template = `
 <br /><small>Copyright © 2020 Panther Labs Inc. All rights reserved.</small>
 `
 
+// Instead of the standard API call, the users-api was invoked by Cognito as a custom message trigger
+func CognitoTrigger(event *events.CognitoEventUserPoolsCustomMessage) (*events.CognitoEventUserPoolsCustomMessage, error) {
+	zap.L().Info("handling cognito trigger", zap.String("source", event.TriggerSource))
```
rleighton: Debug?

internal/core/users_api/api/cognito_trigger.go
```diff
@@ -19,17 +19,17 @@ package custommessage
...
 	"fmt"
 	"net/url"
 
 	"github.com/aws/aws-lambda-go/events"
-	"github.com/aws/aws-sdk-go/aws"
 	"go.uber.org/zap"
 )
 
 // This is similar to the template in deployments/core/cognito.yml for the invite email.
 const template = `
```
rleighton: Maybe be more descriptive, like ``` const passwordResetTemplate``` , makes later reading the code easier.

internal/core/users_api/api/remove_user.go
```diff
@@ -19,29 +19,24 @@ package api
...
-
 	"github.com/panther-labs/panther/api/lambda/users/models"
 	"github.com/panther-labs/panther/pkg/genericapi"
 )
 
 // RemoveUser deletes a user from cognito.
-func (API) RemoveUser(input *models.RemoveUserInput) error {
-	users, err := userGateway.ListUsers()
+func (API) RemoveUser(input *models.RemoveUserInput) (*models.RemoveUserOutput, error) {
+	users, err := userGateway.ListUsers(&models.ListUsersInput{})
```
rleighton: When we get to 1000 users this will get a little slow :)

internal/core/users_api/api/remove_user_test.go
```diff
@@ -23,51 +23,30 @@ import (
...
-	assert.Error(t, err)
-	assert.IsType(t, err, &genericapi.AWSError{})
-
-	mockGateway.AssertExpectations(t)
-}
+func TestRemoveUser(t *testing.T) {
+	userID := aws.String("user-remove")
+	otherUserID := aws.String("user-other")
 
-func TestRemoveLastUser(t *testing.T) {
```
rleighton: Why remove this test?

internal/core/users_api/cognito/list_users_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	mockCognitoClient.On("ListUsersPages", input, mock.Anything).Return(output, nil)
+
+	result, err := gw.ListUsers(&models.ListUsersInput{Status: aws.String("CONFIRMED")})
+	require.NoError(t, err)
+	assert.Len(t, result, 1)
+	mockCognitoClient.AssertExpectations(t)
+}
+
+func TestListUsersFilterByName(t *testing.T) {
```
rleighton: Maybe add ```TestListUsersFilterByNameAndStatus``` ?

api/lambda/users/models/api.go
austinbyers: Correct, filters are logical and. I'll add a comment

internal/core/users_api/api/cognito_trigger.go
austinbyers: Not in my opinion. Cognito triggers are pretty rare, I want to know when any of them happen even if we ignore them because it provides useful information about how cognito is processing updates

internal/core/users_api/api/remove_user.go
austinbyers: We intentionally do not expose paging here because we assume there will never be more than a few dozen users.

I find it incredibly unlikely that a security team will have more than 1000 people using the same deployment, but if so we can cross that bridge when we get to it. Paging now would be a very premature optimization and would require changes to the UI

internal/core/users_api/api/remove_user_test.go
austinbyers: I just copied our internal test suite, this case is covered by integration tests already. Unit tests are not very helpful here, tbh. The main handlers are usually only 3-5 lines that just call out to other mocked services.

internal/core/users_api/api/remove_user.go
austinbyers: My comment is about why `ListUsers` does not offer paging to the caller, but I realized you're asking about `RemoveUser`

I'll add a `limit` option so the list can stop early - we only care if there is at least one user

internal/core/users_api/cognito/list_users_test.go
austinbyers: No, those are two different tests - the `status` test is above that one

The mock results returned are the same, but the list calls are different:

```go
func TestListUsersFilterByStatus(t *testing.T) {
	result, err := gw.ListUsers(&models.ListUsersInput{Status: aws.String("CONFIRMED")})
}

func TestListUsersFilterByName(t *testing.T) {
	result, err := gw.ListUsers(&models.ListUsersInput{Contains: aws.String("example.com")})
}
```

austinbyers: Or are you suggesting combining the two unit tests? The name check is slightly more complex because it looks in multiple fields, so I'd rather keep that one separate

rleighton: You have a test for filter by name and one for status, you do not have one that test filtering by name and status at the same time (unless I missed that). I was suggesting adding that. Not a big deal, just a suggestion .

austinbyers: Oh, got it, I misunderstood. Yes, I can do that

Issue #359, opened 2020-03-06
Do not suppress prettier on `testdata`


austinbyers:
### Describe the bug

Running `prettier` on the `tools` directory breaks some tests

Context in https://github.com/panther-labs/panther/pull/321#issuecomment-595673614

We can update the tests to look for the new format, or change their equality checking - tests should not be so fragile as to depend on the exact indentation / formatting.

PR #444, opened 2020-03-18
Mage build:cfn and refactor monitoring stack

### LINKED ISSUE #435: CloudWatch Alarm Generation Should Resolve References
Currently CloudWatch alarm generation is done by parsing CloudFormation files but this parsing does not understand references (only constants). Ideally, this process should read all the CF files and resolve references.

austinbyers:
## Background
In preparation for reorganizing the CloudFormation stacks, I realized there's a subset of changes I can isolate for easier review. This PR focuses on refactoring the `deployments/monitoring` stack - most importantly that all dynamic values are passed as parameters instead of being written directly to the cloudformation.

I needed this to more effectively parallelize and flatten the stacks (coming soon), but this also has the nice property that `mage build:cfn` can now build all CloudFormation without a deployment. This makes it easier to develop and also we can add it to CI!

Closes: #359 (sort of accidentally, actually)
Closes: #435 

## Changes

- Refactor `monitoring` stack:
    - Move alarm topic ARN here
    - Remove doubly nested alarm stack for performance and simplicity
    - Use parameters for the alarm topic and any dynamic IDs in the monitoring resources
    - Rename `out/deployments/cloudwatch` to `out/deployments/monitoring` (to match the name of the stack)
- Fixed a small typo in SQS alarm messages (missing a space)
- Format all `tools` testdata (#359)
- Finish `build:cfn` and add a `build:all` command used for deploy and CI
- Move `config` models into their own package so it can be shared across all `tools` - resolves #435

## Testing

- Deployed to dev account, checked the dashboards and alarms and such and they all seem correct, but I definitely want @rleighton to deploy this branch and make sure they still look right

## Migration

- I recommend deleting the `panther-app-monitoring` stack before redeploying to avoid potential name conflicts


austinbyers:
Ah yeah, PR number 4️⃣ 4️⃣ 4️⃣

kostaspap:
> I needed this to more effectively parallelize and flatten the stacks (coming soon), but this also has the nice property that mage build:cfn can now build all CloudFormation without a deployment. This makes it easier to develop and also we can add it to CI!

This is really great, I was really missing this functionality. Thanks for fixing this!

tools/mage/swagger_test.go
```diff
@@ -56,8 +57,13 @@ func TestEmbedAPIs(t *testing.T) {
...
+	require.NoError(t, yaml.Unmarshal(transformed, &result))
 
-	expected, err := ioutil.ReadFile("testdata/valid-api-expected-output.yml")
+	expectedBytes, err := ioutil.ReadFile("testdata/valid-api-expected-output.yml")
 	require.NoError(t, err)
-	assert.Equal(t, string(expected), string(transformed))
+	var expected map[string]interface{}
+	require.NoError(t, yaml.Unmarshal(expectedBytes, &expected))
+
+	assert.Equal(t, expected, result)
```
kostaspap: You can probably use `require.YAMLEq()`

deployments/monitoring.yml
```diff
@@ -17,19 +17,67 @@
...
+    Type: String
+    Description: Appsync GraphQL ID
+  LoadBalancerFullName:
+    Type: String
+    Description: Full name of the application load balancer
+
+  AlarmTopicArn:
+    Type: String
+    Description: An existing SNS topic for CloudWatch alarms. If not specified, one is created for you.
+    Default: ''
```
kostaspap: Since this is the monitoring template, I am imagining it would be fine to also create the AlarmTopic inside here. I think that would further simplify significantly the process? Is there any real usecase where someone wouldn't like to have an SNS topic generated?

tools/cfngen/cloudwatchcf/alarms_test.go
```diff
@@ -19,22 +19,24 @@ package cloudwatchcf
...
 func TestGenerateAlarms(t *testing.T) {
-	stackOutputs := map[string]string{
-		"WebApplicationLoadBalancerFullName": "testLoadbalancer",
-		"WebApplicationGraphqlApiId":         "testGraphqlId",
-	}
-	_, cf, err := GenerateAlarms("my-sns-topic-arn", stackOutputs, "./testdata/cf.yml")
+	_, cf, err := GenerateAlarms("./testdata/cf.yml")
 	require.NoError(t, err)
-	const expectedFile = "./testdata/generated_test_alarms.json"
-	// uncomment to make a new expected file
```
rleighton: This is handy to have in the code when many things change. Had editing large generated files is not tenable. With this in the code you can uncomment and review the diff. It would be nice to keep it, unless you have a better way to make this efficient.

tools/cfngen/cloudwatchcf/dashboard.go
```diff
@@ -20,53 +20,53 @@ package cloudwatchcf
...
 		},
 	}
-	return
 }
 
 func GenerateDashboards(dashboards []*Dashboard) (cf []byte, err error) {
 	resources := make(map[string]interface{})
 	for _, dashboard := range dashboards {
-		resources[cfngen.SanitizeResourceName(dashboard.Properties.DashboardName)] = dashboard
+		resources[strings.TrimSuffix(dashboard.Properties.DashboardName.Sub, "-${AWS::Region}")] = dashboard
```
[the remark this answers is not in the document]
rleighton: Also, why don't you need to called "sanitize"? Since people can put anything in as the the name of the dashboard then there could be illegal resource names, right?

tools/cfngen/cloudwatchcf/metrics_test.go
```diff
@@ -19,18 +19,24 @@ package cloudwatchcf
...
 
+	jsoniter "github.com/json-iterator/go"
 	"github.com/stretchr/testify/require"
 )
 
 func TestGenerateMetrics(t *testing.T) {
 	cf, err := GenerateMetrics("./testdata/cf.yml")
 	require.NoError(t, err)
-	const expectedFile = "./testdata/generated_test_metrics.json"
-	// uncomment to make a new expected file
```
rleighton: I'd like to leave this (like above) for when there are large changes.

tools/cfngen/gluecf/db_test.go
```diff
@@ -19,37 +19,33 @@ package gluecf
...
 
 	cfTemplate := cfngen.NewTemplate("Test template", nil, resources, nil)
 
 	cf, err := cfTemplate.CloudFormation()
 	require.NoError(t, err)
+	var result map[string]interface{}
+	require.NoError(t, jsoniter.Unmarshal(cf, &result))
 
-	// uncomment to make a new expected file
-	// writeTestFile(cf, expectedFile)
```
rleighton: Please put back

tools/cfngen/gluecf/gluecf_test.go
```diff
@@ -41,20 +43,19 @@ type commonFields struct {
...
 	table := awsglue.NewGlueTableMetadata(models.LogData, "Log.Type", "dummy", awsglue.GlueTableHourly, &dummyParserEvent{})
 	tables := []*awsglue.GlueTableMetadata{table}
 
 	cf, err := GenerateTables(tables)
 	require.NoError(t, err)
+	var result map[string]interface{}
+	require.NoError(t, jsoniter.Unmarshal(cf, &result))
 
-	// uncomment to make a new expected file
-	//writeTestFile(cf, expectedFile)
```
rleighton: Please put back

tools/cfngen/gluecf/table_test.go
```diff
@@ -92,12 +94,13 @@ func TestTables(t *testing.T) {
 
 	cf, err := cfTemplate.CloudFormation()
 	require.NoError(t, err)
+	var result map[string]interface{}
+	require.NoError(t, jsoniter.Unmarshal(cf, &result))
 
-	// uncomment to make a new expected file
-	// writeTestFile(cf, expectedFile)
```
rleighton: Please put back

rleighton:
Deployed this and the dashboards looked correct and the alarm count was correct. Nice job.

tools/cfngen/cloudwatchcf/alarms_test.go
```diff
@@ -19,22 +19,24 @@ package cloudwatchcf
...
+	var result map[string]interface{}
+	require.NoError(t, jsoniter.Unmarshal(cf, &result))
+
+	expectedCf, err := ioutil.ReadFile("./testdata/generated_test_alarms.json")
 	require.NoError(t, err)
-	require.Equal(t, expectedCf, cf)
+	var expected map[string]interface{}
+	require.NoError(t, jsoniter.Unmarshal(expectedCf, &expected))
+
+	require.Equal(t, expected, result)
```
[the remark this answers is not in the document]
austinbyers: Ah neat! Will do

tools/cfngen/cloudwatchcf/dashboard.go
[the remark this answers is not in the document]
austinbyers: When I had it as sanitize, the CFN resource names (logical ID) turned out to all be `PantherAlertProcessingAWSRegion`

I just really didn't like that, and I figured since the names are few and coded by us in the same package, we didn't to sanitize them.

I can put it back if you're ok with that name ^

deployments/monitoring.yml
austinbyers: > Since this is the monitoring template, I am imagining it would be fine to also create the AlarmTopic inside here.

I do! The alarm topic is created in this template! I moved it here, it's just not created if they provided their own topic in the config

tools/cfngen/cloudwatchcf/alarms_test.go
austinbyers: Your `readFile` and `writeFile` tset utility functions actually already exist in the standard library:

```go
func readTestFile(filename string) ([]byte, error) {	
	fd, err := os.Open(filename)	
	if err != nil {	
		return nil, err	
	}	
	defer fd.Close()	
	return ioutil.ReadAll(fd)	
}
```

is almost identical to [`ioutil.ReadFile`](https://golang.org/pkg/io/ioutil/#ReadFile):

```go
// ReadFile reads the file named by filename and returns the contents.
// A successful call returns err == nil, not err == EOF. Because ReadFile
// reads the whole file, it does not treat an EOF from Read as an error
// to be reported.
func ReadFile(filename string) ([]byte, error) {
	f, err := os.Open(filename)
	if err != nil {
		return nil, err
	}
	defer f.Close()
	// It's a good but not certain bet that FileInfo will tell us exactly how much to
	// read, so let's try it but be prepared for the answer to be wrong.
	var n int64 = bytes.MinRead

	if fi, err := f.Stat(); err == nil {
		// As initial capacity for readAll, use Size + a little extra in case Size
		// is zero, and to avoid another allocation after Read has filled the
		// buffer. The readAll call will read into its allocated internal buffer
		// cheaply. If the size was wrong, we'll either waste some space off the end
		// or reallocate as needed, but in the overwhelmingly common case we'll get
		// it just right.
		if size := fi.Size() + bytes.MinRead; size > n {
			n = size
		}
	}
	return readAll(f, n)
}
```

Same for `ioutil.WriteFile`

tools/cfngen/cloudwatchcf/metrics_test.go
austinbyers: `ioutil.WriteFile` is exactly the same thing, in the standard library. I can leave a comment for that function if you like, but we don't need the handwritten functions

tools/cfngen/cloudwatchcf/dashboard.go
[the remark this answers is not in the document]
rleighton: Yeah, can you sanitize()? I worry about breakage that will puzzle someone later.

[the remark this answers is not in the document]
austinbyers: > Not clear to me, why do you have to trim the suffix?

The dashboard name is interpolated now: "MyName-${AWS::Region}"

So if you use the name of the dashboard for the CFN resource name as well, you'll have the `AWSRegion` suffix

tools/cfngen/cloudwatchcf/alarms_test.go
rleighton: That's a good change.  Can you just add to the test code using the std lib so we can just uncomment to regenerate the generated test files?

tools/cfngen/cloudwatchcf/dashboard.go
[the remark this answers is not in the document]
austinbyers: Reverted

tools/cfngen/cloudwatchcf/metrics_test.go
austinbyers: Putting the comment back, referencing `ioutil`

austinbyers:
This had some conflicting changes with #427 , I resolved them by using a config file instead of reading stack outputs.

Deployed again, still looks good

PR #447, opened 2020-03-18
Removed  'scanEnabled' property

kostaspap:
## Background

While working on the source_api found a property `scanEnabled` that is no longer used (double checked with @nhakmiller ) Removed it to cleanup a bit the code. 

## Changes

- Removed unused property. Small refactoring (renamings) in the code

## Testing

- unit tests.


internal/core/source_api/ddb/scan.go
```diff
@@ -28,35 +28,34 @@ import (
...
 
-// ScanEnabledIntegrations returns all enabled integrations based on type (if type is specified).
+// ScanIntegrations returns all enabled integrations based on type (if type is specified).
 // It performs a DDB scan of the entire table with a filter expression.
-func (ddb *DDB) ScanEnabledIntegrations(input *models.ListIntegrationsInput) ([]*models.SourceIntegration, error) {
-	filt := expression.Name("scanEnabled").Equal(expression.Value(true))
-	if input.IntegrationType != nil {
-		filt = expression.And(filt, expression.Name("integrationType").Equal(expression.Value(input.IntegrationType)))
-	}
+func (ddb *DDB) ScanIntegrations(input *models.ListIntegrationsInput) ([]*models.SourceIntegration, error) {
```
nhakmiller: There needs to be a corresponding frontend change for this re-naming, correct?

kostaspap: This is DDB operation - so it is used only internally!

PR #451, opened 2020-03-18
devtools:logprocessor - run log processor locally 

rleighton:
devtools:logprocessor - run log processor locally for instrumentation/benchmarks/debugging

## Background
This adds a notion of a mage build target called 'build:devtools' meant for building tools for developers.

The first tool is a 'main' for the logprocessor with flags to allow specifying inputs and also set up for pprof profiling output and to be used to tune performance.

## Changes
* mage target 'build:devtools'
* devtool 'logprocessor'

## Testing
mage build:logprocessor
ran locally


cmd/devtools/logprocessor/logprocessor/main.go
```diff
@@ -0,0 +1,98 @@
...
+			log.Fatal("could not create CPU profile: ", err)
+		}
+		defer f.Close() // error handling omitted for example
+		if err := pprof.StartCPUProfile(f); err != nil {
+			log.Fatal("could not start CPU profile: ", err)
+		}
+		defer pprof.StopCPUProfile()
+	}
+
+	config := zap.NewProductionConfig()
```
austinbyers: If running locally, `zap.NewDevelopmentConfig()` is usually a bit more friendly

rleighton: Good idea. Actually, it would be nice to have that settable by a flag. I'll do that .

rleighton: Done

Issue #446, opened 2020-03-18
PutIntegration returns empty list


3nvi:
### Describe the bug

Whenever I try to create a new integration, I'm getting back an empty list instead of getting the item (or the list of items) that got created.

### Steps to reproduce

1. Go to the lambda console for the `panther-sources-api`
2. Click on Test
3. Attempt to call `putIntegration` like this:

```json
{
  "putIntegration": {
    "integrations": [
      {
        "awsAccountId": "726103014039",
        "integrationLabel": "Test new compliancee",
        "remediationEnabled": true,
        "cweEnabled": true,
        "userId": "cc67e287-24b2-487b-864d-d72ec51080f7",
        "scanIntervalMins": 1440,
        "integrationType": "aws-scan",
        "scanEnabled": true
      }
    ]
  }
}
```
4. Inspect the result

### Expected behavior

A list should be returned with a single item; the item that just got created

### Additional context

I can verify that the data is being received correctly from the lambda

### Screenshots

<img width="1580" alt="Screen Shot 2020-03-18 at 1 31 41 PM" src="https://user-images.githubusercontent.com/10436045/76956701-37d69800-691d-11ea-9e4e-208ee58b93cd.png">

3nvi:
**UPDATE**
I saw that this behaviour happens when I attempt to add a compliance source when the related AWS ID already exists in another integration

3nvi:
@nhakmiller We need a way to properly document these behaviors since they slow down development when trying to understand/debug whether the lambda has a bug or the front-end sends the request wrongly. Cloudwatch sometimes doesn't log anything  neither for the appsync nor for the lambda.

kostaspap:
I'm doing some refactoring as part of my current task. I will deal with this.

Issue #394, opened 2020-03-11
Integration tests for sources_api fail


kostaspap:
### Describe the bug

Integration tests for sources_api fail. 

### Steps to reproduce

1. Deploy panther
2. Run `mage test:integration`

### Expected behavior

Tests should pass

### Environment

How are you deploying or using Panther?

- Panther version or commit: 9140b629242d15159400b2be23620e375129a398
- OS:  MacOS 10.15.3

### Additional context

Error
```
16:11:09        INFO    test:integration: go test ./internal/core/source_api/main -run=TestIntegration*
--- FAIL: TestIntegration (2.14s)
    --- FAIL: TestIntegration/API (1.78s)
        --- FAIL: TestIntegration/API/PutIntegrations (1.66s)
            integration_test.go:119: 
                        Error Trace:    integration_test.go:119
                        Error:          Received unexpected error:
                                        lambda error returned: panther-source-api: PutIntegration failed: invalid input: integration 888888888888 did not pass health check
                        Test:           TestIntegration/API/PutIntegrations
        --- FAIL: TestIntegration/API/GetEnabledIntegrations (0.12s)
            integration_test.go:132: 
                        Error Trace:    integration_test.go:132
                        Error:          "[]" should have 2 item(s), but has 0
                        Test:           TestIntegration/API/GetEnabledIntegrations
        --- FAIL: TestIntegration/API/DeleteIntegrations (0.00s)
panic: runtime error: index out of range [0] with length 0 [recovered]
        panic: runtime error: index out of range [0] with length 0

```

jacknagz:
@kostaspap is this still an issue? can we close it out?

kostaspap:
> @kostaspap is this still an issue? can we close it out?

This is still an issue

s0l0ist:
@kostaspap  what is the status of this issue? I'm punting to v1.12 in the meantime.

PR #461, opened 2020-03-19
Updated sources api to support new log onboarding operations

kostaspap:
## Background

 - Updated sources api to support new log onboarding operations
- We only put one integration at a time so updated the API to accept one onboard request at a time. This simplifies a lot the logic. 

## Notes
This is PR 1/3 for the new log onboarding process. This PR is focused on updating the `source_api` operations. 
The UI still doesn't support this new capability and the log processor will not be able to pull any data. This is a known issue and will be fixed by subsequent PRs. This PR is meant to be merged against feature branch `logonboarding`

## Changes

- PutIntegration accepts now one integration at a time
- We can now have only one CloudSec integration per account. If user tries to add a second one, we throw an error. 
- Log integrations for the same account need to have different labels. If user adds to add a log integration for an account with the same label, we throw an error. 
- Fix issue with Log onboarding template. See for more details the description of #465 
- Renamed log onboarding template to `panther-log-analysis-iam.yml`
- Fixed error messages returned to the user when PutIntegration fails. 
- PutIintegration request for LogOnboarding now has the following fields: 
  - S3Bucket 
  - S3Prefix 
  - LogTypes
  - KmsKey
- Added validation for IntegrationLabel. Integration label now can only have alphanumeric, empty space or dash. We put this limitation because now the label is used as suffix for CFN Stack name and role name.
- ListIntegrations is now serialized as `listIntegrations` instead of `getEnabledIntegrations`
- Fixed bug when removing SQS Queue policy would fail. 
- Updated CFN template for log processing onboarding. 
- Introducing `awsmocks.go` in the `testutils`. So many of my tests are using mocks of S3 clients, I thought it makes sense to define at one place and just re-use as appropriate. Didn't update all the code, to use them, but we can do it moving forward. 

## Testing

- Unit tests
- Deployed and verified in my account 
  - PutIntegration for log source/cloudsec
  - ListIntegration for log source/cloudsec
  - Delete Integration for log source/cloudsec
  - GetTemplate

@nhakmiller Can you take a look when you get some time? I know you've worked with the source_api so I'd like to know what you think of the changes. 
@3nvi I am updating the API to use new fields for Log Onboarding and also simplifying the PutIntegration. Also fixing #446


api/lambda/source/models/validator.go
```diff
@@ -19,20 +19,46 @@ package models
...
 	"gopkg.in/go-playground/validator.v9"
 )
 
+var (
+	integrationLabelValidatorRegex *regexp.Regexp
+)
+
+func init() {
+	var err error
+	integrationLabelValidatorRegex, err = regexp.Compile("^[0-9a-zA-Z- ]+$")
```
rleighton: You do not need the init() , do this in the var  above like:
```
var (
integrationLabelValidatorRegex = regexp.MustCompile("^[0-9a-zA-Z- ]+$")
)
```

api/lambda/source/models/validator_test.go
```diff
@@ -0,0 +1,48 @@
+package models
```
rleighton: Add tests for validateRoleArn() while you are in here?

internal/core/source_api/api/check_integration.go
```diff
@@ -156,20 +146,13 @@ func evaluateIntegration(api API, integration *models.CheckIntegrationInput) (bo
...
+	if *integration.IntegrationType == models.IntegrationTypeAWSScan {
+		passing := status.AuditRoleStatus.Healthy
+		// For these two, we are ok if they are not enabled or if they are passing
+		passing = passing && (!aws.BoolValue(integration.EnableRemediation) || status.RemediationRoleStatus.Healthy)
+		passing = passing && (!aws.BoolValue(integration.EnableCWESetup) || status.CWERoleStatus.Healthy)
+		return passing, nil
 	}
-
-	return passing, nil
+	// One of these will be nil, one of these will not. We only care about the value of the not nil one.
```
rleighton: This comment does not make sense, no nils here

internal/core/source_api/api/check_integration.go
```diff
@@ -156,20 +146,13 @@ func evaluateIntegration(api API, integration *models.CheckIntegrationInput) (bo
...
-	passing = passing && (!aws.BoolValue(integration.EnableRemediation) || aws.BoolValue(status.RemediationRoleStatus.Healthy))
-	passing = passing && (!aws.BoolValue(integration.EnableCWESetup) || aws.BoolValue(status.CWERoleStatus.Healthy))
-
-	// For these two, we are ok if none are set or all are passing
-	for _, bucket := range status.S3BucketsStatus {
-		passing = passing && aws.BoolValue(bucket.Healthy)
-	}
-	for _, key := range status.KMSKeysStatus {
-		passing = passing && aws.BoolValue(key.Healthy)
+	if *integration.IntegrationType == models.IntegrationTypeAWSScan {
```
rleighton: I found this hard to read.  The implicit 'else' for the 'if' is is a "processing" type. But I had to read this for a while to be sure. I think this would be more readable as a 'switch' on *integration.IntegrationType and the conditions for return under case 'case'.

internal/core/source_api/api/delete_integration.go
```diff
@@ -49,22 +54,45 @@ func (API) DeleteIntegration(input *models.DeleteIntegrationInput) (err error) {
...
 	}
 
 	if *integration.IntegrationType == models.IntegrationTypeAWS3 {
-		if err = RemovePermissionFromLogProcessorQueue(*integration.AWSAccountID); err != nil {
-			zap.L().Error("failed to remove permission from SQS queue for integration",
-				zap.String("integrationId", *input.IntegrationID),
-				zap.Error(errors.Wrap(err, "failed to remove permission from SQS queue for integration")))
-			return &genericapi.InternalError{Message: "failed to update integration"}
+		existingIntegrations, err := db.ScanIntegrations(&models.ListIntegrationsInput{IntegrationType: aws.String(models.IntegrationTypeAWS3)})
+
```
rleighton: extra \n

internal/core/source_api/api/delete_integration_test.go
```diff
@@ -227,13 +296,17 @@ func TestDeleteIntegrationDeleteRecoveryFails(t *testing.T) {
 	mockClient.AssertExpectations(t)
 }
 
-func getItem(integrationType string) *dynamodb.GetItemOutput {
+func generateGetItemOtput(integrationType string) *dynamodb.GetItemOutput {
```
rleighton: Typo in name

internal/core/source_api/api/get_integration_template.go
```diff
@@ -19,49 +19,60 @@ package api
...
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 	"go.uber.org/zap"
 
 	"github.com/panther-labs/panther/api/lambda/source/models"
 )
 
 const (
 	TemplateBucket           = "panther-public-cloudformation-templates"
+	TemplateBucketRegion     = endpoints.UsWest2RegionID
```
rleighton: This is fine for now, but will not work for China or any isolated region. It would be better if the code could take this as a parameter so in the future we can just deal with the config changes. Is that a hard change?

internal/core/source_api/main/integration_test.go
```diff
@@ -87,30 +87,47 @@ func TestIntegration(t *testing.T) {
 }
 
 func putIntegrations(t *testing.T) {
-	input := &models.LambdaInput{
-		PutIntegration: &models.PutIntegrationInput{
+	putIntegrations := []*models.PutIntegrationInput{
```
rleighton: The caller above is commented out. Did you test this? Perhaps fix this test. Also, integration tests should clean up after themselves.

api/lambda/source/models/validator.go
kostaspap: Nice! Updated

api/lambda/source/models/validator_test.go
kostaspap: Actually i searched the whole project and realized this is not used anywhere ^^ . The validation was also not complete e.g. it would validate successfully an IAM user. I removed it completely.

internal/core/source_api/api/check_integration.go
kostaspap: True, leftover from refactoring. Removed.

internal/core/source_api/api/check_integration.go
kostaspap: You are right. Updated.

internal/core/source_api/api/delete_integration.go
kostaspap: Removed.

internal/core/source_api/api/delete_integration_test.go
kostaspap: Fixed

internal/core/source_api/api/get_integration_template.go
kostaspap: This functionality doesn't support China right now. It's not just the region field - the templates are currently in one S3 Bucket in us-west-2. It would take more work to enable China. I would prefer to do that work in separate issue since my work in this PR is mostly unrelated (I just replaced a hardcoded `us-west-2` with the appropriate constant here).

internal/core/source_api/main/integration_test.go
kostaspap: The integration tests are currently not running - they started failing some time ago. We have an issue to fix this: https://github.com/panther-labs/panther/issues/394

We rely on unit tests and manual testing currently for this capability.

internal/core/source_api/api/get_integration_template.go
rleighton: Sorry, I was not clear. I was asking if we can pass-in the bucket and region in this PR (so later we can deal with isolated regions w/out changing this code).  Hard coding these things is just work/breakage later.

internal/core/source_api/main/integration_test.go
rleighton: Perhaps this is something Austin can prioritize post v1. Not having working integration tests is a big risk. Also, having them be self contain (clean up after them selves) is important.

api/lambda/source/models/api.go
```diff
@@ -61,21 +62,23 @@ type CheckIntegrationInput struct {
 
 // PutIntegrationInput is used to add one or many integrations.
 type PutIntegrationInput struct {
-	Integrations  []*PutIntegrationSettings `json:"integrations" validate:"required,dive"`
```
nhakmiller: What is the value of reducing this to just one integration at a time?

api/lambda/source/models/integration.go
```diff
@@ -56,23 +58,23 @@ type SourceIntegrationScanInformation struct {
...
-	KMSKeysStatus        map[string]SourceIntegrationItemStatus `json:"kmsKeysStatus"`
+	ProcessingRoleStatus SourceIntegrationItemStatus `json:"processingRoleStatus"`
+	S3BucketStatus       SourceIntegrationItemStatus `json:"s3BucketStatus"`
+	KMSKeyStatus         SourceIntegrationItemStatus `json:"kmsKeyStatus"`
 }
 
 type SourceIntegrationItemStatus struct {
-	Healthy      *bool   `json:"healthy"`
-	ErrorMessage *string `json:"errorMessage"`
+	Healthy      bool   `json:"healthy"`
```
nhakmiller: @3nvi said he was ok with these being `*bool`s when we talked yesterday, did you guys talk since then and decide differently?

internal/core/source_api/api/check_integration.go
```diff
@@ -156,20 +147,36 @@ func evaluateIntegration(api API, integration *models.CheckIntegrationInput) (bo
...
+
+	case models.IntegrationTypeAWS3:
+		if !status.ProcessingRoleStatus.Healthy || !status.S3BucketStatus.Healthy {
+			// If Log processing role is not healthy or S3 bucket status is not healthy return false
+			return false, nil
+		}
 
-	return passing, nil
+		if integration.KmsKey != nil {
+			// If the integratio has a KMS key and the keys is not healthy return false
```
nhakmiller: *integration

internal/core/source_api/api/get_integration_template.go
```diff
@@ -75,45 +86,50 @@ func (API) GetIntegrationTemplate(input *models.GetIntegrationTemplateInput) (*m
...
-	formattedTemplate = bytes.Replace(formattedTemplate, remediationFind,
-		[]byte(fmt.Sprintf(remediationReplace, *input.RemediationEnabled)), 1)
-
-	// Log Analysis replacements
-	formattedTemplate = bytes.Replace(formattedTemplate, s3BucketFind,
-		[]byte(fmt.Sprintf(s3BucketReplace, strings.Join(sliceStringValue(input.S3Buckets), ","))), 1)
-	formattedTemplate = bytes.Replace(formattedTemplate, kmsKeyFind,
-		[]byte(fmt.Sprintf(kmsKeyReplace, strings.Join(sliceStringValue(input.KmsKeys), ","))), 1)
+	if *input.IntegrationType == models.IntegrationTypeAWSScan {
+		formattedTemplate = strings.ReplaceAll(formattedTemplate, cweFind,
```
nhakmiller: Any reason `ReplaceAll` is preferable to `Replace` here? As far as I'm aware, we only ever have to replace one instance.

internal/core/source_api/api/get_integration_template.go
```diff
@@ -122,23 +138,28 @@ func getTemplate(integrationType *string) ([]byte, error) {
...
 	templateCache[*integrationType] = templateCacheItem{
 		Timestamp: time.Now(),
-		Body:      templateBody,
+		Body:      string(templateBody),
 	}
 
-	// Return the template
-	return templateBody, nil
+	// Return the s3Object
+	return string(templateBody), nil
```
nhakmiller: You could doing this conversion once then re-use it, so it doesn't have to traverse the entire thing twice.

api/lambda/source/models/integration.go
kostaspap: I think @3nvi prefers bool and he can deal with *bool even though it is not preferred, that's why I changed it. That's my recollection of yesterday's discussion. 
I'm fine either way. I can revert it back.

internal/core/source_api/api/check_integration.go
kostaspap: Eagle-eye! Fixed

internal/core/source_api/api/get_integration_template.go
kostaspap: I changed it during refactoring for no particular reason (I think I found the syntax shorter). But you are right, probably better to limit to 1. Changing.

internal/core/source_api/api/get_integration_template.go
kostaspap: By traversing you mean to avoid casting the templateBody to string twice? Sure, updated.

api/lambda/source/models/integration.go
nhakmiller: I don't have any preference, I just know on this specific point there was a lot of back and forth and I just want to make sure the front end and back end are aligned. A value of false (the default bool value) could be misinterpreted depending on how the frontend does things. As long as we are aligned with Aggelos' expectations either is fine with me.

api/lambda/source/models/api.go
3nvi: @kostaspap asked me if I ever want to support multiple integrations. I can pretty much guarantee that this will never be the case. It helps reduce boilerplate a bit from my side

api/lambda/source/models/api.go
```diff
@@ -51,8 +51,9 @@ type CheckIntegrationInput struct {
 	EnableRemediation *bool `json:"enableRemediation"`
 
 	// Checks for log analysis integrations
-	S3Buckets []*string `json:"s3Buckets"`
-	KmsKeys   []*string `json:"kmsKeys"`
+	S3Bucket *string `json:"s3Bucket,omitempty"`
+	S3Prefix *string `json:"s3Prefix,omitempty"`
```
3nvi: @kostaspap I thought we were gonna merge S3bucket + S3Prefix in a single field

api/lambda/source/models/integration.go
3nvi: I agree with @nhakmiller . Leave them as is for now.

kostaspap:
> I don't agree with moving the `PutIntegration` from accepting a list to only accepting a single item. Was there a specific reason for this? Batch adding integrations is something we will need to support at some point, and I haven't noticed it causing any issues. I worry we will just have to re-add this complex logic in the coming months, instead of having it baked in.

This is a reasonable concern. My worry about the current Batch method was that the code was a bit more complicated as it was trying to deal with an array of requests. This is reasonable of course because it has to deal with an array instead of having to deal with a single request. 
However, with the current code we can add a `BatchPutIntegration` or `PutIntegrations` relatively  easy: The Batch method will receive an array of `PutIntegration` requests and the method will delegate the handling of each onboarding request to the existing `PutIntegration` method. We can even run each request in its own go routine for increases performance if we wish to. I think with this approach the code will be overall small and simpler/easier to test. 
Also the above approach gives a couple more benefits: 
- We can easily support partially succeeding batches with appropriate failure messages for the integrations that failed. 
- Many of our systems (UI, onboarding during deployment) just need to add a single integration at a time so it kinda simplifies a bit calling it (admittedly very slightly, that's mostly a side effect)

I took the freedom to update it after talking with frontend and also seeing that no other part of the system is using this functionality currently. I can add a "PutIntegrations" method back following the approach I mentioned above once needed.

api/lambda/source/models/integration.go
kostaspap: Sure, I guess I misunderstood (or i mixed yesterday's discussion with older discussions somehow). I will revert back!

api/lambda/source/models/api.go
kostaspap: Chatted offline. Keeping it as is for now.

PR #463, opened 2020-03-19
Improve users in schema.graphql

3nvi:
## Background

This PR simplifies the graphql schema for user listing and simplifies the logic/component behind the UserInvitation & UserUpdate form.

Multiple open PRs currently have the user listing schema simplification, so instead of re-writing it in every new PR, it's time to isolate it in a single one.

**As always any file ending in `.generated.tsx` should not be reviewed** 

## Changes

- `inviteUser` now returns the created Userr
- `users` now returns a flattened list of the users
- Rename `BaseUserForm` to `UserForm`
- `UserInvitationForm` and `EditUserForm` are now simply `UserForm`
- `EditUser` and `EditUserSidesheet`'s code has been merged to a single component

## Testing


web/src/components/modals/ResetUserPasswordModal/graphql/resetUserPassword.generated.ts
```diff
@@ -17,6 +17,7 @@
  */
 
 /* eslint-disable import/order, import/no-duplicates */
+
```
alexmylonas: This extra line is from `generated-graphql` or `prettier`? Shoul we stick with the result from the first only?

codegen.yml
```diff
@@ -20,7 +20,26 @@ generates:
...
+           * but WITHOUT ANY WARRANTY; without even the implied warranty of
+           * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+           * GNU Affero General Public License for more details.
+           *
+           * You should have received a copy of the GNU Affero General Public License
+           * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+           */
+
+          /* eslint-disable import/order, import/no-duplicates */
     config:
```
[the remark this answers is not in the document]
3nvi: That's how we had it, to minimize changes I thought we would leave it as is. Eslint just wants to be **beforer any code**, not necessarily at the very top of the file

web/src/components/modals/ResetUserPasswordModal/graphql/resetUserPassword.generated.ts
3nvi: I have no idea. I was not able to remove it somehow. Would you like to try?

PR #466, opened 2020-03-19
Enhance behavior for User Invitation & Update

3nvi:
## Background

This PR makes the UX of updating a User and Inviting a new User better, by removing the need to refetch Users everytime an update and an invitation occurs. In addition, it fixes a minor bug when deleting a user.

## Changes

- Protect against empty `givenName` and `familyName`
- Force appsync to return `true` on user deletions
- Alter schema to return a `User` on updates & invites
- Optimistically add new user to the table
- Optimistically update new user on the table
- Disable eslint rule for `no-unused-vars` on autogen files

## Testing

- Locally


web/src/components/forms/UserForm/UserForm.tsx
```diff
@@ -87,14 +87,13 @@ const UserForm: React.FC<UserFormProps> = ({ initialValues, onSubmit }) => {
               mt={10}
               justifyContent="flex-end"
             >
-              <Flex>
```
3nvi: removed un-used flex

web/src/components/Header/Header.tsx
```diff
@@ -42,7 +42,9 @@ const Header = () => {
             />
             {userInfo && (
               <Text size="medium">
-                {userInfo.given_name} {userInfo.family_name[0]}.
+                {userInfo.given_name && userInfo.family_name
```
3nvi: protect against empty values

deployments/core/appsync.yml
```diff
@@ -290,7 +290,7 @@ Resources:
         #if($context.error)
           $util.error($context.error.errorMessage, $context.error.errorType, $ctx.args)
         #else
-          $util.toJson($context.result)
+          true
```
3nvi: Force a `true` on deletions

web/src/components/sidesheets/EditUserSidesheet/EditUserSidesheet.tsx
```diff
@@ -54,32 +50,36 @@ const EditUserSidesheet: React.FC<EditUserSidesheetProps> = ({ user }) => {
     givenName: user.givenName || '',
   };
 
+  const submitToServer = async (values: UserFormValues) => {
+    // optimistically hide the sidesheet
+    hideSidesheet();
+
+    await editUser({
```
3nvi: moved the `onSubmit` code from the handler to the component below

3nvi: also added optimistic response

Issue #456, opened 2020-03-18
Reorganize CFN stacks


austinbyers:
Our CloudFormation stacks are not well organized and not very efficient. It turns out nested stacks are not very performant - it forces round-trips to S3 (even if nothing changed) and forces more resources to be sequentially deployed that could in fact be parallel.

When we first open-sourced the repo, we pretty much just copied stacks from dozens of separate repos we used to have. They were never really organized.

The goal of reorganizing is to:

- Make deployments and teardowns significantly faster
- Allow devs to quickly deploy individual stacks
- Organize CFN templates so it's easier for developers to understand and work with
- Enable a number of other fixes: #197 , #443, #143 

Since CFN reorganization is a big breaking change, I'm going to try to ship it before v1

Issue #197, opened 2020-02-16
During initial deployment, invite email sent before webservice is available


stoggi:
### Describe the bug

During an initial deployment with `mage deploy`, after receiving the invite email clicking on the Dashboard link in the email will lead to a 503 error. The web service is deployed after setting up the Panther admin user, and took about 3-4 minutes before the page could be loaded.

### Steps to reproduce

1. Deploy a new Panther environment
2. Click on the `Dashboard` url in the admin user invite email
3. Observe a 503 error
4. Wait 3-4 minutes until the web service has been deployed
5. Reload dashboard and get presented with sign in page

### Expected behavior

While this is a very minor, people may think their deploy hasn't worked. It's not clear the web service had not yet been deployed or a delay is to be expected.

The email text could mention to wait until the deployment is complete, or delay sending the invite email until the web service has been deployed.

### Environment

How are you deploying or using Panther?

- Panther master at commit 48f2f244d453bb0cbaa3077c7d80f024b1a97a5c
- OS: Linux (Manjaro 19)
- Deployment method: docker container via ./dev.sh

3nvi:
It takes some time for ECS to reach consistency. We can definitely use the aws sdk to get this information and send the email after this check is made.

The reason though that you see it is because the invitation gets sent before the UI is created (for some reason).

I can easily revert the order and add a check to make sure that the email is sent after the related ECS service is stable, but @austinbyers might be making some changes in #213, so perhaps we can include it there?

austinbyers:
This is just because we setup the first user after the main backend stack, before the web has even been deployed.

The fix would be to just move the `postDeploySetup` after all stacks have deployed, which makes more sense anyway

austinbyers:
I'll put this in a separate PR, #213 is too big and is ready to merge

PR #467, opened 2020-03-20
Flatten CloudFormation stacks for faster deploys

austinbyers:
## Background
Panther's long deployment time is considerably slowing down the development cycle. The CloudFormation is also hard to understand, it was never organized after we smashed together the CFN from dozens of repos to form the repo you see today.

Unfortunately, existing deployments will need to start over completely after this change, which is why I'm landing it before v1.

Closes: #456 
Closes: #197 

### Nested Stacks
It turns out one of the biggest performance bottlenecks are nested stacks, for two main reasons:

1. Nested stacks break parallelization - the entire nested stack has to finish before you can reference any of its outputs. Some resources take far longer to create (e.g. KMS keys), so there is a lot of time spent waiting for one or two resources before CFN can continue
1. Nested stacks require round trips to S3 (you have to upload, CFN has to download). The S3 artifacts are not pinned to a specific version, which means at the time of change set creation CFN can't know whether there will be changes until it starts evaluating the template.

There are other good reasons to reduce our reliance on nested stacks besides improved performance:

- The default [limit](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html) is 200 CFN stacks, which includes nested stacks. Panther is already at 36 stacks, 18% of a user's entire quota
- Reliability: nested stacks seem to get stuck in UPDATE_ROLLBACK_FAILED states that can be difficult to recover from. With mostly flat singleton stacks, the majority of a deployment can still succeed even if a small part fails, and the failing stack can be deleted and recreated individually without having to tear down an entire deployment.

## Design
I've almost completely removed nested stacks (with a few small exceptions noted below). Panther now has 12 top-level stacks:

1. `bootstrap` - All S3 buckets, KMS keys, Appsync Api, Cognito user pool, load balancer / networking. Basically, everything that takes a long time to create and/or generates random IDs that we need later. 
1. `bootstrap-gateway` - API gateway and the Python layer - "bootstrap" resources that just require an S3 bucket to exist first (which is why they can't be in the first stack). 

These two stacks are the only ones with outputs! In fact, I actually combine the outputs from these two stacks and pass them through the entire`deploy` command. With the bootstrap stacks out of the way (~4 minutes), nearly everything else can be deployed in parallel:

- CloudWatch alarms: to avoid per-stack resource limits, we generate nested stacks for each top-level template. We could deploy each alarm stack separately, but here having a top-level stack with several nested stacks makes sense - it's a collection of many autogenerated substacks. They used to be doubly nested (monitoring => alarms => substacks), I got rid of the first layer of nesting
- CloudWatch dashboards
- CloudWatch metric filters (in their own stack because the associated log groups must exist first)
- Appsync schema + resolvers (the API ID is in bootstrap but nothing else related to appsync)
- Cloud Security
- Core (stuff shared between cloudsec and log analysis - analysis-api, users-api, etc)
- web server: this is only the ECS task (the part that actually changes when web src is modified). All the web network and load balancer is in bootstrap
- Glue databases (same as before, just deployed directly instead of as a nested stack)
- Log Analysis
- Onboarding (same as before). These are nested stacks that reference the `aux` templates, so I just left them

Stack Counts

## Changes

- On a fresh deploy, the invite for the first user has been moved to the very end (closing #197). This means the user prompt no longer blocks the deploy process - when you see the prompt, you're done!
- Reorganized the config file thematically instead of exposing stack details
- The CFN resources have not been modified (just moved) with a few exceptions:
    - Renamed the "access-logs" S3 bucket to "auditlogs" (per @rleighton's request). This bucket is now always created, but you can still route S3 access logs somewhere else per the config file.
    - I removed some parameters and outputs if they weren't needed
    - There are more generated CW alarms than before (e.g. SNS topics are now included)

## Testing

- Deployed to dev account, was able to login, update my profile, see the dashboard, resources, policies, etc. @kostaspap @rleighton can you deploy and make sure log processing / remediation still work as expected?

- Timed the deploy to `us-east-1` from commit `a748131` and this branch with a macbook on gigabit internet:

Deploy Benchmark

About twice as fast overall!


tools/mage/deploy.go
```diff
@@ -254,27 +298,156 @@ func uploadLayer(awsSession *session.Session, libs []string, bucket, key string)
...
+	}(finishedStacks)
+
+	// Web server
+	go func(result chan string) {
+		deployFrontend(awsSession, settings, accountID, sourceBucket, outputs)
+		result <- frontendStack
+	}(finishedStacks)
+
+	// Wait for stacks to finish
+	for i := 1; i <= 8; i++ {
```
nhakmiller: Not sure if there is a good/clean way to do this, but maybe you could have a counter that increases every time you call `go func` or something so that it's more clear where this 8 is coming from? My main concern is someone adding/removing a stack in the future and forgetting to update this number. Especially adding a stack would be problematic because the deploy would still finish (and probably be ok) in most cases, but in some rare cases might cause some weirdness.

docs/gitbook/operations/runbooks.md
```diff
@@ -96,9 +102,6 @@ This lambda implements the analysis API which is responsible for
 ## panther-analysis-api
 The `panther-analysis-api` API Gateway calles the `panther-analysis-api` lambda.
 
- Failure Impact
- * Failure of this API Gateway will prevent calls to the `panther-analysis-api` lambda.
-
```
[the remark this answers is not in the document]
austinbyers: Yeah, it just seemed redundant to see in the description "invokes the `panther-analysis-api` lambda" and then in the failure impact "fails to invoke the `panther-analysis-api` lambda"

I can put it back if we like

deployments/panther_config.yml
```diff
@@ -16,96 +16,84 @@
...
-  # Comma-delimited list of LayerVersions to attach to every Lambda function.
-  #
-  # For example, this could be a serverless monitoring/security service.
-  LayerVersionArns: ''
+  # XRay tracing mode for API Gateway and Lambda: '', 'Active', or 'PassThrough'
+  TracingMode: ''
 
-  # Custom layer attached to every Python Lambda function for analysis and remediation.
+Setup:
+  # List of policy/rule sets to install when Panther is FIRST deployed.
```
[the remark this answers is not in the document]
austinbyers: True, I can change the comment (I just copied what we had before)

tools/mage/deploy.go
austinbyers: Sure, fair point. A `sync.WaitGroup` would have the same problem (you have to set the count), I'll see what I can do

docs/gitbook/operations/runbooks.md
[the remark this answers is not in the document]
nhakmiller: No that's fine, I was just wondering if it was across the board or just in a few of them.

tools/mage/deploy.go
austinbyers: You can't update a counter in the goroutine (unless you protect it with a mutex lock)

So what I did instead is to just keep a running total of how many parallel stacks you started like this:

```go
var parallelStacks = 0

// Appsync
parallelStacks++
go func() { .... deploy .... }

// Glue
parallelStacks++
go func() { .... deploy ... }
```

This way, developers can add and remove stacks and the count will still be accurate as long as they follow that convention

nhakmiller: Right it is a bit weird either way, maybe even just a comment saying "hey be sure to update this number if you add/remove any stacks to the process". Also they'll have. to update both the two numbers in the below for loop, which might not be immediately intuitive either. Ultimately its probably a rare case, I wouldn't spend too much time on it.

docs/gitbook/operations/runbooks.md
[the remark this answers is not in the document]
austinbyers: Oh I may not have been consistent with them

austinbyers:
I've redeployed and tested this change even more thoroughly:

- Integration tests
- Clicking on everything in the web app
- Checking infra alarms (there were a few, but all related to stuff I was doing in the app)

I'd say it's ready to merge, but I will wait until @rleighton or @kostaspap can verify the log functionality so that they don't have to deal with the merge conflicts

deployments/bootstrap2.yml
```diff
@@ -0,0 +1,147 @@
+# Panther is a scalable, powerful, cloud-native SIEM written in Golang/React.
```
kostaspap: Can we change the name of it? It looks a bit like a typo to me :P 
Maybe make it more explicit and rename the two stacks from `bootstrap`, `bootstrap2` to:
- `bootstrap`, `pre-reqs`
- `bootstrap-stage-1`, `bootstrap-stage-2`
- `bootstrap-1`, `bootstrap-2`. 
- ...

deployments/bootstrap2.yml
```diff
@@ -0,0 +1,147 @@
...
+Conditions:
+  CreatePythonLayer: !Equals [!Ref PythonLayerVersionArn, '']
+
+Resources:
+  PythonLayer:
+    Condition: CreatePythonLayer
+    Type: AWS::Lambda::LayerVersion
+    Properties:
+      CompatibleRuntimes:
+        - python3.6
```
kostaspap: Not that it matters so much (this PR is massive already!) but I think it would be sufficient to have only `python-3.7` since this is what our lambda's use.

austinbyers: I'm going to leave it because people could re-use this layer in their own Python Lambda functions if they wanted

deployments/bootstrap2.yml
austinbyers: Sure, how about `bootstrap` and `bootstrap-gateway`?

austinbyers: I went with `bootstrap-gateway` because the main resource is API gateway and the variable names in the code tend to be `gatewayStack`, etc

deployments/bootstrap2.yml
kostaspap: Sure

deployments/bootstrap2.yml
kostaspap: works for me

austinbyers:
I've updated the branch with master. #489 and #491  were big changes that modified lots of CloudFormation, I manually reviewed both PRs and applied all the same changes here, with a few small changes:

- Apparently we do have to always supply the certificate arn, even if the bootstrap stack already exists. Not sure if that change happened recently or if I just missed this before
- I found two bugs in the deploy onboarding:
    - I already had guardduty enabled, but when `deploy_onboard.go` tried to describe the onboarding stack, it didn't exist. The error message we were looking for was not accurate. I fixed this
    - The other error is due to #494 , I just commented out the `PutIntegration` call for log processing onboarding
- Simplify onboarding teardown (it was in a separate method that can now be removed)
- Fixed integration tests broken by #482 

I went through a full testing cycle again:

- deploy to empty account
- login, click on everything in the web app
- deploy again (no change)
- integration tests
- teardown

@kostaspap @rleighton I'm going to need help testing this branch to make sure log processing still works end-to-end. There have been lots of CFN migrations which means I may have missed something.

tools/mage/deploy_onboard.go
```diff
@@ -54,38 +51,38 @@ func genLogProcessingLabel(awsSession *session.Session) string {
 }
 
 // onboard Panther to monitor Panther account
-func deployOnboard(awsSession *session.Session, settings *config.PantherConfig, bucketOutputs, backendOutputs map[string]string) {
-	deployOnboardTemplate(awsSession, settings, bucketOutputs)
-	registerPantherAccount(awsSession, bucketOutputs, backendOutputs) // this MUST follow the CloudSec roles being deployed
-	deployRealTimeStackSet(awsSession, backendOutputs["AWSAccountId"])
+func deployOnboard(awsSession *session.Session, settings *config.PantherConfig, accountID string, bootstrapOutputs map[string]string) {
+	deployOnboardTemplate(awsSession, settings, bootstrapOutputs)
```
kostaspap: I made the same comment to @rleighton and he recommend it that we fix the issue in this PR: We should not deploy just one template for the logs we are onboarding, we should deploy one CFN template per source: 
- One template for compliance
- One template per log type we onboard. 

This is the standard we are following when users onboard logs through the UI, we should do it here as well. This will make possible to also off-board individual sources if someone wants to.

austinbyers: That's too big a change for me to make, especially since I'm not familiar with the onboarding flow. If you want to make that change to this branch, please do

deployments/panther_config.yml
```diff
@@ -16,96 +16,85 @@
...
-  # If not specified, a layer is created for you based on the PipLayer setting below.
-  PythonLayerVersionArn: ''
+  # Entries can be URLs or file:// paths that point to a .zip file.
+  # If the analysis-api is non-empty, this setting is ignored and you can instead
+  # use the BulkUpload functionality from the web app to upload new or modified rule sets.
+  InitialAnalysisSets:
+    - https://github.com/panther-labs/panther-analysis/releases/latest/download/panther-analysis-all.zip
+
+  # Whether or not the Panther deployment should automatically onboard itself as a data source.
+  OnboardSelf: true
```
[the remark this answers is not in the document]
austinbyers: I'm just copying the same functionality we had before, changes like that should ideally not be part of this PR, but I will do so this time given the launch today.

- I agree with having it on by default because we want a quick time to value. Users will need to onboard some kind of log data to get a feel for the product
- I can add more comments to explain what is onboarded
- I'll log a warning during the onboarding that costs may be higher
- I'll update the "cost" section of the docs

deployments/panther_config.yml
```diff
@@ -16,96 +16,85 @@
...
-  - requests==2.22.0
-
-# List of policy/rule sets to install when Panther is FIRST deployed.
-#
-# Entries can be URLs or file:// paths that point to a .zip file.
-# If the analysis-api is non-empty, this setting is ignored and you can instead
-# use the BulkUpload functionality from the web app to upload new or modified rule sets.
-InitialAnalysisSets:
-  - https://github.com/panther-labs/panther-analysis/releases/latest/download/panther-analysis-all.zip
+  FargateTaskMemory: 1024
```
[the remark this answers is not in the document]
austinbyers: Sure, I can remove

deployments/panther_config.yml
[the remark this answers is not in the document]
austinbyers: Chatted offline, we'll also disable CloudTrail and GuardDuty by default, since they scan the entire account, not just the Panther deployment region

Issue #457, opened 2020-03-18
Enforce license headers and autogenerated files in CI


austinbyers:
Our `mage` tooling generally evolves faster than the corresponding CI enforcement. Besides Terraform (#356 ), we need to add the following to CI:

- License header enforcement (see https://github.com/panther-labs/panther/pull/433)
- Autogenerated files validation: docs, Swagger specs, and generated typescript should all be up to date with what's in the repo (e.g. https://github.com/panther-labs/panther/pull/396)

austinbyers:
Proposed solution:

1. Update `mage build:api` to generate the web source as well (typescript is generated from graphql)
2. Rather than trying to check for all possible formatting diffs from the many tools we use, a simpler way to enforce formatting is to:
    1. Copy the repo source to `/tmp`
    2. Run `mage build:api fmt`
    3. Look for any differences, and if so, fail CI

Another advantage of this approach is that running `mage test:ci` will automatically fix formatting and autogenerated files so devs don't have to reformat if ci fails

PR #468, opened 2020-03-20
Onboard Log Processing to Panther & Make Roles Regionally Namespaced

rleighton:
## Background
This PR onboards log processing to the Panther account for:
* s3 access logs
* alb
* vpc flow
* cloudtrail (optional, default on)
* guard duty (optional, default on)

I had big issues deploying to multiple regions due to the compliance and log roles that were global:
* PantherAuditRole
* PantherRemediationRole
* PantherCloudFormationStackSetExecutionRole
* PantherLogProcessingRole
This PR adds a region suffix to each to make them regionally name spaced.
Teardown was simplified by this change.

This means the Panther region is now a parameter to the templates. The UI does not seem to be using "get_integration_template" lambda, rather it seems to use the public bucket. If we switch to the API we do not have any UI changes. We need to talk about the UI and this change.

## Changes
* updated buckets.yml with config for services
* updated onboard.yml with config for cloudtrail and guardduty
* updated roles to regional suffixes
* removed need to read public bucket in "get_integration_template" by compiling in the templates into the source_api lambda
* improved logging for role errors in snapshot poller
* fixed bug listing integrations
* fixed bug that returned error when q was configured (should be a warn)
## Testing
mage test:ci
mage deploy


deployments/auxiliary/cloudformation/panther-cloudsec-iam.yml
```diff
@@ -75,7 +80,7 @@ Resources:
   AuditRole:
     Type: AWS::IAM::Role
     Properties:
-      RoleName: PantherAuditRole # DO NOT CHANGE! backend.yml CF depends on this name, set in mage deploy
+      RoleName: !Sub PantherAuditRole-${MasterAccountRegion} # DO NOT CHANGE! backend.yml CF depends on this name
```
austinbyers: Is there any reason we can't just let CloudFormation generate the name for us, like we do with S3 buckets?

deployments/backend.yml
```diff
@@ -53,9 +53,12 @@ Parameters:
     Default: ''
 
   # Set automatically by "mage deploy"
-  S3BucketAccessLogs:
+  AuditLogsBucket:
```
austinbyers: This is also added in #467

deployments/auxiliary/cloudformation/panther-cloudsec-iam.yml
austinbyers: I guess then we'd have to store it somewhere

internal/core/source_api/api/cloudsec_tmpl.go
```diff
@@ -0,0 +1,279 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+// DO NOT EDIT! generated by mage
+var cloudsecTemplate = `# Panther is a scalable, powerful, cloud-native SIEM written in Golang/React.
```
austinbyers: Why is this a Go string constant and not just a generated `.yml` file?

web/src/components/Wizard/Wizard.tsx
```diff
@@ -1,3 +1,21 @@
...
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
```
austinbyers: Remove the header below this one

web/src/components/Wizard/WizardPanelWrapper.tsx
```diff
@@ -1,3 +1,21 @@
...
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
```
austinbyers: Remove the header below this one

web/src/components/forms/UserForm/index.tsx
```diff
@@ -1,3 +1,21 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
```
austinbyers: Remove the header below this one (we'll add license enforcement to CI in #457)

deployments/auxiliary/cloudformation/panther-cloudsec-iam.yml
nhakmiller: In LogAnalysis they have to store the role name and look it up from the sources-api, I'd imaging cloud security could do the same. We used to do that when we were more hosted focused.

internal/core/source_api/api/get_integration_template.go
```diff
@@ -103,42 +83,13 @@ func sliceStringValue(stringPointers []*string) []string {
...
-		Timestamp: time.Now(),
-		Body:      templateBody,
+func getTemplate(integrationType *string) string {
+	switch *integrationType {
+	case models.IntegrationTypeAWSScan:
+		return cloudsecTemplate
+	case models.IntegrationTypeAWS3:
+		return logProcessingTemplate
+	default:
+		panic("unknown integration type: " + *integrationType)
```
nhakmiller: We very intentionally loaded the template from the public s3 bucket in order to ensure it would always be identical to the template that was used by the launch console button. Was there a specific reason you decided to change that decision?

tools/mage/deploy_onboard.go
```diff
@@ -40,50 +49,58 @@ const (
...
 func registerPantherAccount(awsSession *session.Session, pantherAccountID string) {
-	logger.Infof("deploy: registering account %s with Panther", pantherAccountID)
+	logger.Infof("deploy: registering account %s with Panther for monitoring", pantherAccountID)
+
+	// cloud sec
 	var apiInput = struct {
 		PutIntegration *models.PutIntegrationInput
 	}{
 		&models.PutIntegrationInput{
 			Integrations: []*models.PutIntegrationSettings{
```
nhakmiller: @kostaspap is changing this API to no longer support batch put operations, just a heads up for when you're merging the changes. You will have to do two sequential put operations.

tools/mage/util.go
```diff
@@ -100,6 +99,7 @@ func getSession() (*session.Session, error) {
 	return awsSession, nil
 }
 
+/* NOT USED but might be later
```
nhakmiller: Unless we have specific plans for using this in the near future, I'd say just remove it. If needed we can always go find it in the git hsitory.

deployments/auxiliary/cloudformation/panther-cloudsec-iam.yml
rleighton: We should likely move to how it is now done for log processing. Not urgent but would be consistent.

internal/core/source_api/api/cloudsec_tmpl.go
rleighton: Yeah, should be const

nhakmiller:
> ...
> ## Changes
> * removed need to read public bucket in "get_integration_template" by compiling in the templates into the source_api lambda
> ...

I commented on this in the code directly, but we had a reason for doing this. By going to the public bucket, we know that the launch console link in the UI and the get template API are always based on exactly the same base template. By generating it like you do in this change, the get template API can only be updated by re-deploying Panther but the launch console link could potentially be updated as we update the public template. I think this will be confusing and unintuitive for users that the template loaded may different based on which option they selected for setup.

web/src/components/Wizard/Wizard.tsx
rleighton: done

web/src/components/Wizard/WizardPanelWrapper.tsx
rleighton: done

web/src/components/forms/UserForm/index.tsx
rleighton: done

internal/core/source_api/api/get_integration_template.go
rleighton: Yes, we should remove the dependency on the public link. It can get out of sync. This way we are exactly current with the code. The UI should use the lambda to get "filled" templates.  I personally think we should remove that work flow and just have 2  mage targets `config:generate` and `config:apply` . The first you run with the Panther creds and it stores the templates under ./out, the second with the target account creds to apply them. I'm working on a proof of concept while I wait for test deploys to run :)

tools/mage/deploy_onboard.go
rleighton: Merged. Testing now before pushing an update.

tools/mage/util.go
rleighton: Sure

internal/core/source_api/api/get_integration_template.go
nhakmiller: So you want to remove the `Launch Stack` button from the UI that takes the users directly to CloudFormation with the correct stack and parameters all loaded? I think that is a very useful button for people new to AWS or new to Panther.

nhakmiller: I also like command line interfaces a lot, but plenty of people find them daunting. I think one of the awesome benefits of Panther is the clean and fully integrated web interface. Telling people they HAVE to go the command line (and with our mage tool installed as well) in order to carry out core functionality of the tool (adding sources) seems to be counter to that benefit. I have no problem adding that as an additional option for power users, but I think we absolutely should maintain a quick and easy way to progress solely through the web interface.

rleighton:
I moved to a new branch for merging work. I've incorporated suggested changes and will open a new PR for the merge.

PR #471, opened 2020-03-20
Log processor assumes role retrieved from sources api 

kostaspap:
## Background

The last backend change to support the new onboarding flow. 
In the new flow, we have one CFN stack per log source onboarded. This means we are having one IAM Role per log source. IAM roles have as suffix the integration label (lowercased, with empty spaces replaced with `-`)

This change modifies the log processor so that it assumes the appropriate role for the S3 object it is about to process. 

## Changes

- Sources API now stores the LogProcessing role ARN in the sources DB
- Log processor maintains a cache of log sources. When it receives an S3 Object it checks it if comes from a known S3 bucket+ S3 prefix. If it comes, it assumes the given role and reads data from S3. If it doesn't come from a known s3 bucket, we throw an exception. The message will eventually be retried. This protects us from the case where a log source has been onboarded but the cache in the log processor hasn't been updated with the latest integrations.  
- Added missing unit tests

## Testing

- Unit tests
- Deployed to my account. When I had configured no source the log processor was reporting errors. When the integration was added, log processor started processing events.  

@rleighton  do you mind taking a look at the log processor changes.

### DIFF
--- a/internal/log_analysis/log_processor/sources/s3_client_cache.go
+++ b/internal/log_analysis/log_processor/sources/s3_client_cache.go
@@ -29,6 +29,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/lambda"
 	"github.com/aws/aws-sdk-go/service/lambda/lambdaiface"
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 	lru "github.com/hashicorp/golang-lru"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
@@ -40,13 +41,20 @@ import (
 
 const (
 	// sessionDurationSeconds is the duration in seconds of the STS session the S3 client uses
-	sessionDurationSeconds  = 3600
-	sourceAPIFunctionName = "panther-source-api"
+	sessionDurationSeconds = 3600
+	sourceAPIFunctionName  = "panther-source-api"
+	// How frequently to query the sources_api for new integrations
+	sourceCacheDuration = 5 * time.Minute
 )
 
 type s3ClientCacheKey struct {
 	roleArn   string
 	awsRegion string
+}
+
+type sourceCacheStruct struct {
+	cacheUpdateTime time.Time
+	sources         []*models.SourceIntegration
 }
 
 var (
@@ -56,10 +64,16 @@ var (
 	// s3ClientCacheKey -> S3 client
 	s3ClientCache *lru.ARCCache
 
+	sourceCache = &sourceCacheStruct{
+		cacheUpdateTime: time.Unix(0, 0),
+	}
+
 	lambdaClient lambdaiface.LambdaAPI = lambda.New(common.Session)
+
+	//used to simplify mocking during testing
+	newCredentialsFunc = stscreds.NewCredentials
+	newS3ClientFunc    = getNewS3Client
 )
-
-
 
 func init() {
 	var err error
@@ -76,7 +90,7 @@ func init() {
 
 // getS3Client Fetches S3 client with permissions to read data from the account
 // that owns the SNS Topic
-func getS3Client(s3Object *S3ObjectInfo) (*s3.S3, error) {
+func getS3Client(s3Object *S3ObjectInfo) (s3iface.S3API, error) {
 	roleArn, err := getRoleArn(s3Object)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to fetch the appropriate role arn to retrieve S3 object %#v", s3Object)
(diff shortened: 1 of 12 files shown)


internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -19,44 +19,65 @@ package sources
...
 )
 
 const (
 	// sessionDurationSeconds is the duration in seconds of the STS session the S3 client uses
-	sessionDurationSeconds  = 3600
-	logProcessingRoleFormat = "arn:aws:iam::%s:role/PantherLogProcessingRole"
+	sessionDurationSeconds = 3600
+	sourceAPIFunctionName  = "panther-source-api"
+	// How frequently to query the sources_api for new integrations
+	sourceCacheDuration = 10 * time.Minute
```
austinbyers: I would do no more than 5 minutes

pkg/testutils/awsmocks.go
```diff
@@ -31,3 +33,18 @@ func (m *S3Mock) GetObject(input *s3.GetObjectInput) (*s3.GetObjectOutput, error
 	args := m.Called(input)
 	return args.Get(0).(*s3.GetObjectOutput), args.Error(1)
 }
+
+func (m *S3Mock) GetBucketLocation(input *s3.GetBucketLocationInput) (*s3.GetBucketLocationOutput, error) {
+	args := m.Called(input)
+	return args.Get(0).(*s3.GetBucketLocationOutput), args.Error(1)
+}
```
austinbyers: Unit tests all over define their own versions of these, it'd be nice to eventually consolidate all of them here, good idea!

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -19,44 +19,65 @@ package sources
...
+
+	//used to simplify mocking during testing
+	newCredentialsFunc = stscreds.NewCredentials
+	newS3ClientFunc    = getNewS3Client
+)
 
 func init() {
 	var err error
-	s3ClientCache, err = lru.NewARC(100)
+	s3ClientCache, err = lru.NewARC(1000)
```
rleighton: Add const above for the 1000

internal/log_analysis/log_processor/sources/s3_client_cache.go
rleighton: What's the risk for 10?

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -129,11 +152,48 @@ func getBucketRegion(s3Bucket string, awsCreds *credentials.Credentials) (string
...
+	return newCredentialsFunc(common.Session, roleArn, func(p *stscreds.AssumeRoleProvider) {
 		p.Duration = time.Duration(sessionDurationSeconds) * time.Second
 	})
 }
+
+// Returns the appropriate role arn for a given S3 object
+// It will return error if it encountered an issue retrieving the role.
+// It will return nil result if role for such object doesn't exist.
+func getRoleArn(s3Object *S3ObjectInfo) (*string, error) {
+	now := time.Now() // No need to be UTC. We care about relative time
```
rleighton: :)

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -129,11 +152,48 @@ func getBucketRegion(s3Bucket string, awsCreds *credentials.Credentials) (string
...
+	if sourceCache.cacheUpdateTime.Add(sourceCacheDuration).Before(now) {
```
rleighton: This gets done for each object, seems very heavy. What about starting a go routine in init() that manages the cache? The lru code is thread safe. I see no downside.

internal/log_analysis/log_processor/sources/s3_client_cache.go
kostaspap: Sure. Updated.

internal/log_analysis/log_processor/sources/s3_client_cache.go
kostaspap: Sure. Added.

internal/log_analysis/log_processor/sources/s3_client_cache.go
austinbyers: the lru is thread safe, the cache is not unless you lock it. Go maps are not concurrency safe

You could run into a situation where the cache is being updated while the main routine is using it, which would be bad. Checking a timestamp is probably similar overhead to checking a mutex

austinbyers: Goroutines in Lambda might be more overhead than checking a timestamp depending on how much memory is configured. You could do it once per invocation of the log processor instead of once per object

rleighton: A sleeping go routine is not expensive. And it makes the code cleaner. Separates concerns.

rleighton: We can do it later, I think it is a good change. Maybe make an issue.

kostaspap: Created issue: https://github.com/panther-labs/panther/issues/474

Issue #347, opened 2020-03-05
Modify Analysis API to allow users to modify custom event grouping/dedup 


kostaspap:
### Description

Modify Analysis API to allow users to modify custom event grouping/dedup parameters

### Acceptance Criteria

- Analysis API allows clients to store and retrieve the event grouping time period

PR #475, opened 2020-03-20
Deduplication time Period for Rules display & update enabled

### LINKED ISSUE #349: Rules page should include alert dedup time period
### Description

Events that match rules are grouped into alerts based on two criteria: 
- The `dedupString` associated with every event matched
- A time period (currently hardcoded to 60 minutes)
This means that if some events match a rule within the given time period and all have the same dedup strings, they will trigger only one Alert.

We want to make this time period configurable per rule.

### Acceptance Criteria

- Users can see the alert dedup time period that is configured for each rule
- Users can edit the alert dedup time period. Time period is in *minutes* and has minimum value of 15' and maximum of 1 day. 
- Default time period when someone creates a new rule is defined at 1 hr.

alexmylonas:
## Background

Rules have a deduplication time period property implemented by #347. 

This PR allows users to view and edit  `dedupPeriodMinutes` for each rule.
We are enforcing user input for `dedupPeriodMinutes` for both creation and editing of rules.

If it's not defined, backend defaults this to 60 minutes 

Closes #349, #476 

## Changes

- Included `dedupPeriodMinutes` in api `schema.graphql` types
- Included `dedupPeriodMinutes` to components queries related to Rules
- Updated Rule Form components to require Deduplication Period field on edit & creation
- Minor fix for policy array

## Testing

- Locally
- Dev Environment

Screenshot 2020-03-20 at 11 42 20 PM
Screenshot 2020-03-20 at 11 42 45 PM
Screenshot 2020-03-20 at 11 42 51 PM
Screenshot 2020-03-20 at 11 43 00 PM


alexmylonas:
@austinbyers Screenshots added!

Also a minor fix that closes #476

web/src/components/forms/BaseRuleForm/BaseRuleFormCoreFields/BaseRuleFormCoreFields.tsx
```diff
@@ -178,6 +179,13 @@ const BaseRuleFormCoreFields: React.FC<BaseRuleCoreFieldsProps> = ({ type }) =>
             inputProps={logTypesInputProps}
           />
         )}
+        <Field
+          as={FormikCombobox}
+          label="* Deduplication Period"
```
kostaspap: I know this is a required field so we have the `*` in front. But since there is no option for users to no fill it (which is good), maybe it's redundant and we can still remove it?

web/src/components/forms/BaseRuleForm/BaseRuleFormCoreFields/BaseRuleFormCoreFields.tsx
```diff
@@ -178,6 +179,13 @@ const BaseRuleFormCoreFields: React.FC<BaseRuleCoreFieldsProps> = ({ type }) =>
             inputProps={logTypesInputProps}
           />
         )}
+        <Field
```
3nvi: @alexmylonas This will be rendered for both PolicyForm and RoleForm. Is that what we want?

api/graphql/schema.graphql
```diff
@@ -165,6 +167,7 @@ type RuleSummary {
   lastModified: AWSDateTime
   logTypes: [String]
   severity: SeverityEnum
+  dedupPeriodMinutes: Int
```
3nvi: Can this field accept `null` values? If not we should add `!`

api/graphql/schema.graphql
```diff
@@ -149,6 +150,7 @@ input CreateOrModifyRuleInput {
   logTypes: [String]
   runbook: String
   severity: SeverityEnum!
+  dedupPeriodMinutes: Int!
```
3nvi: `dedupPeriodMinutes` do we need this field as input to a query?  Are we "filtering" based out of it?

api/graphql/schema.graphql
```diff
@@ -130,6 +130,7 @@ input ListRulesInput {
   enabled: Boolean
   logTypes: String # Comma-separated string
   severity: SeverityEnum
+  dedupPeriodMinutes: Int
```
3nvi: dedupPeriodMinutes do we need this field as input to a query? Are we "filtering" based out of it?

web/src/helpers/utils.tsx
```diff
@@ -38,6 +38,7 @@ import sum from 'lodash-es/sum';
 import { Box, ColumnProps, Label } from 'pouncejs';
 import { ErrorResponse } from 'apollo-link-error';
 import { ApolloError } from '@apollo/client';
+import { int } from 'aws-sdk/clients/datapipeline';
```
3nvi: don't make this import. It's un-necessary. Just use typescript's `number`

web/src/pages/EditRule/EditRule.tsx
```diff
@@ -60,17 +60,17 @@ const EditRulePage: React.FC = () => {
 
   const initialValues = React.useMemo(() => {
     if (queryData) {
-      const { tests, ...otherInitialValues } = pick(
```
3nvi: I believe you don't  need these changes since the value is "required" when creating a Rule (according to `schema.graphql`) . Thus the ` || 60` will never kick in. 

The existing code will work just fine.

P.S. Keep the `RuleDetails` change! Good catch on the TS issue :D

web/src/pages/ListPolicies/columns.tsx
```diff
@@ -84,7 +84,7 @@ const columns = [
   {
     key: 'severity',
     sortable: true,
-    flex: '0 0 100px',
+    flex: '0 0 110px',
```
3nvi: did it overflow?

web/src/pages/RuleDetails/RuleDetailsInfo/RuleDetailsInfo.tsx
```diff
@@ -172,6 +172,14 @@ const RuleDetailsInfo: React.FC<ResourceDetailsInfoProps> = ({ rule }) => {
             {formatDatetime(rule.lastModified)}
           </Text>
         </Box>
+        <Box my={1}>
+          <Label mb={1} is="div" size="small" color="grey300">
+            DEDUPLICATION PERIOD TIME
+          </Label>
+          <Text size="medium" color={rule.displayName ? 'black' : 'grey200'}>
+            {rule.dedupPeriodMinutes ? minutesToString(rule.dedupPeriodMinutes) : '1h'}
```
3nvi: again: according to this line `dedupPeriodMinutes: Int!` in `schema.graphql` this will never be empty/null

3nvi:
Hey @alexmylonas , great PR! Super solid work.  My only problem is the minor issue with the fact that the policies form is going to have the `dedup`-field. 

Other than that, I just commented on somme minor optimizations that are not affecting the productioon in any way, but may simplify the code a bit.

All in all, great!

api/graphql/schema.graphql
alexmylonas: Yes it can! Bulk uploading that is used for first deploy initializes with `null`.
@kostaspap is going to work on a fix for this.

kostaspap: The fix is already in place. https://github.com/panther-labs/panther/pull/481

api/graphql/schema.graphql
alexmylonas: I think it's necessary for the form to work seen it's can be modified. I can double check it

web/src/helpers/utils.tsx
alexmylonas: Correct will do

web/src/pages/ListPolicies/columns.tsx
alexmylonas: Yeap, issue here #476

web/src/pages/RuleDetails/RuleDetailsInfo/RuleDetailsInfo.tsx
alexmylonas: same as above comment!

PR #479, opened 2020-03-20
Log Onboarding with Log Processing Onboarding

rleighton:
## Merge of my log onboarding (approved) with current branch of log on boarding.


rleighton:
> I do still feel pretty strongly that we should NOT be loading the template files from disk during build/deployment, this leaves no method for updating the templates without re-deploying Panther and also means the launch console stack and the get template stack will not always be the same.
> 
> I'm approving in light of the v1 release on Monday, but if you decide to go forward with that part of the change I would like to have some internal discussions about this choice and its implications soon.

Agree. I feel pretty strongly the opposite way. We should figure out why. In the short run it does not matter.

internal/core/source_api/api/get_integration_template.go
```diff
@@ -129,41 +96,15 @@ func (API) GetIntegrationTemplate(input *models.GetIntegrationTemplateInput) (*m
...
-	if err != nil {
-		return "", err
+func getTemplate(integrationType *string) string {
+	switch *integrationType {
+	case models.IntegrationTypeAWSScan:
+		return cloudsecTemplate
+	case models.IntegrationTypeAWS3:
+		return logAnalysisTemplate
+	default:
+		panic("unknown integration type: " + *integrationType)
```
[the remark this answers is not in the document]
rleighton: This is an alternative. We can discuss pros/cons after v1 release party!

deployments/auxiliary/cloudformation/panther-log-processing-iam.yml
```diff
@@ -0,0 +1,92 @@
...
+    Type: CommaDelimitedList
+    Description: Allow Panther master account to get location of these bucket patterns.
+      E.g. "arn:aws:s3:::my-bucket,arn:aws:s3:::another-bucket*"
+    Default: '' # S3Buckets
+  S3ObjectPrefixes:
+    Type: CommaDelimitedList
+    Description:
+      Allow Panther master account access to access S3 objects that have the following patterns.
+      E.g. "arn:aws:s3:::my-bucket/prefix*,arn:aws:s3:::my-bucket/prefix-2*"
+    Default: '' # S3ObjectPrefixes
```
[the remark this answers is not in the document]
rleighton: Good catch! That file should be deleted. Done

internal/core/source_api/api/put_integration.go
```diff
@@ -58,6 +58,10 @@ func (api API) PutIntegration(input *models.PutIntegrationInput) (*models.Source
 		return nil, putIntegrationInternalError
 	}
 	if !passing {
+		zap.L().Warn("PutIntegration: resource is not healthy",
+			zap.Error(err),
+			zap.String("reason", reason),
+			zap.Any("input", input))
 		return nil, &genericapi.InvalidInputError{
```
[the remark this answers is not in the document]
rleighton: Done

tools/mage/deploy.go
```diff
@@ -108,16 +104,18 @@ func Deploy() {
 
 	logger.Infof("deploy: deploying Panther to %s", *awsSession.Config.Region)
 
-	// Deploy prerequisite bucket stack
+	// registerPantherAccount(awsSession, "050603629990")
```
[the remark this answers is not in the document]
rleighton: yes! good eyes.

deployments/core/buckets.yml
```diff
@@ -81,9 +137,57 @@ Resources:
...
+  AuditLogsBucket:
+    Description: S3 bucket name for Panther audit logs (include s3 access, alb, vpc)
+    Value: !Ref AccessLogs
   SourceBucketName:
     Description: S3 bucket name for Panther CloudFormation packaging
     Value: !Ref Source
+
+Mappings:
+  # OMG, I can't believe they make people do this ...
+  ELBAccounts: # https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-access-logs.html
```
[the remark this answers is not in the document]
rleighton: Thanks, yes. Updated.

rleighton:
I don't want to push this until tomorrow. I want to deploy/teardown over 3-4 regions a few times to be sure this is working.

deployments/onboard.yml
```diff
@@ -17,18 +17,184 @@
...
+      TargetKeyId: !Ref GuardDutyEncryptionKey
+  GuardDutyDetector:
+    Condition: EnableGuardDuty
+    Type: AWS::GuardDuty::Detector
+    Properties:
+      Enable: true
+      FindingPublishingFrequency: FIFTEEN_MINUTES
+
+  # Configure the Log processing role and access to AuditLogsBucket
+  LogProcessingRole:
```
kostaspap: You shouldn't repeat the text here - better use nested template and reference the actual CFN template we use for onboarding. If we update that template, we don't have to update it in both places.

deployments/onboard.yml
```diff
@@ -17,18 +17,184 @@
...
+                - Effect: Allow
+                  Action:
+                    - kms:Decrypt
+                    - kms:DescribeKey
+                  Resource: !Sub arn:${AWS::Partition}:kms:${AWS::Region}:${AWS::AccountId}:key/${GuardDutyEncryptionKey}
+                - !Ref AWS::NoValue
+
+  # Configure SNS Topic to receive bucket notifications
+  # This topic is used to notify the log processor queue whenever new data is written to the auditing bucket.
+  Topic:
```
kostaspap: This is the `panther-log-processing-infra.yml`. You shouldn't repeat the next - just reference using nested CFN stack.

deployments/web/networking.yml
```diff
@@ -95,6 +100,16 @@ Resources:
       SubnetId: !Ref PublicSubnetTwo
       RouteTableId: !Ref PublicRouteTable
 
+  # Panther monitors itself, the AuditLogsBucket is processed by the log processor
+  FlowLogs:
```
kostaspap: I guess this enables the logs always? There is no way to opt-out right? ( I see Austin is doing some work [here](https://github.com/panther-labs/panther/pull/467) to make it possible to opt-out but I guess you are not making this possible, right?)

internal/core/source_api/api/check_integration.go
```diff
@@ -140,42 +151,40 @@ func getCredentialsWithStatus(roleARN string) (*credentials.Credentials, models.
 	}
 }
 
-func evaluateIntegration(api API, integration *models.CheckIntegrationInput) (bool, error) {
+func evaluateIntegration(api API, integration *models.CheckIntegrationInput) (string, bool, error) {
 	status, err := api.CheckIntegration(integration)
 	if err != nil {
-		return false, err
+		zap.L().Error("integration failed health check",
```
kostaspap: This actually shouldn't be an error. Errors in the logs should indicate a problem with the system - in this case it is not really a problem of Panther, rather than a user providing invalid parameters. Should be Info at max.

internal/core/source_api/api/check_integration.go
```diff
@@ -140,42 +151,40 @@ func getCredentialsWithStatus(roleARN string) (*credentials.Credentials, models.
...
+			zap.Any("status", status))
+		return "", false, err
 	}
 
 	switch aws.StringValue(integration.IntegrationType) {
 	case models.IntegrationTypeAWSScan:
 		if !aws.BoolValue(status.AuditRoleStatus.Healthy) {
-			// If audit role is not healthy return false
-			return false, nil
+			return "unhealthy audit role", false, nil
```
kostaspap: It is actually less a "health" issue and more like a misconfiguration.... If we wanted to display something to the users that should be that they haven't configured their integration properly...

internal/core/source_api/api/cloudsec_tmpl.go
```diff
@@ -0,0 +1,281 @@
+package api
```
kostaspap: Why define it in Go? We currently support the "Launch Stack" button - and that uses a template in S3. By defining the template here you define it twice. Why is it better to define the template in two different places, I don't get it.

internal/core/source_api/api/sqs_utils.go
```diff
@@ -63,7 +63,9 @@ func AddPermissionToLogProcessorQueue(accountID string) (bool, error) {
 	}
 
 	if findStatementIndex(existingPolicy, accountID) >= 0 {
-		// if it already exists, no need to do anything
+		zap.L().Warn("queue has already been configured",
```
kostaspap: This is not a problem so it shouldn't be warning. This is absolutely normal - it can happen when people onboard multiple logs from the same account.

internal/core/source_api/api/update_integration.go
```diff
@@ -54,7 +56,12 @@ func (api API) UpdateIntegrationSettings(input *models.UpdateIntegrationSettings
 		return nil, err
 	}
 	if !passing {
-		return nil, &genericapi.InvalidInputError{Message: fmt.Sprintf("integration %s did not pass health check", *integration.AWSAccountID)}
+		zap.L().Warn("UpdateIntegration: resource is not healthy",
+			zap.Error(err),
+			zap.String("reason", reason),
+			zap.Any("input", input))
+		return nil, &genericapi.InvalidInputError{Message: fmt.Sprintf("integration %s did not pass health check because of %s",
```
kostaspap: Nice!

internal/core/source_api/api/vars_test.go
```diff
@@ -25,4 +29,8 @@ const (
 	testUserID           = "97c4db4e-61d5-40a7-82de-6dd63b199bd2"
 )
 
+func init() {
+	sess.Config.Region = aws.String("us-east-1")
```
kostaspap: I always prefer to use constants when they are available. You can use `endpoints.UsEast1RegionID`

internal/core/source_api/ddb/scan.go
```diff
@@ -26,22 +26,29 @@ import (
...
-		return nil, errors.Wrap(err, "failed to build scan expression")
-	}
-
-	scanInput := &dynamodb.ScanInput{
-		FilterExpression:          expr.Filter(),
-		ExpressionAttributeNames:  expr.Names(),
-		ExpressionAttributeValues: expr.Values(),
-		TableName:                 aws.String(ddb.TableName),
+	var scanInput *dynamodb.ScanInput
+	if input.IntegrationType != nil {
```
kostaspap: What is the usecase, where this is needed to be nil?

tools/dashboards/alerts.go
```diff
@@ -178,7 +178,7 @@ var alertsJSON = `
             "width": 9,
             "height": 6,
             "properties": {
-                "query": "SOURCE '/aws/lambda/panther-alert-forwarder' | SOURCE '/aws/lambda/panther-alert-processor' | filter level='error' or level='warn'\n| fields @timestamp, @message\n| sort @timestamp desc",
+                "query": "SOURCE '/aws/lambda/panther-alert-forwarder' | SOURCE '/aws/lambda/panther-alert-processor' | filter level='error' or level='warn'\n| fields @timestamp, @message\n| sort @timestamp desc | limit 20",
```
kostaspap: Good catch!

tools/mage/deploy_onboard.go
```diff
@@ -19,73 +19,148 @@ package mage
...
+	"github.com/aws/aws-sdk-go/service/guardduty"
+	"github.com/aws/aws-sdk-go/service/s3"
 
 	"github.com/panther-labs/panther/api/lambda/source/models"
+	"github.com/panther-labs/panther/tools/config"
 )
 
 const (
+	cloudSecLabel      = "panther-cloud-security"
+	logProcessingLabel = "panther-log-processing" // this must be lowercase, no spaces to work correctly, see genLogProcessingLabel()
```
kostaspap: In the new log onboarding flow, each template will have as suffix the integration label (lowercased). I think it would be good to follow the same practice here and call it e.g. panther-log-processing-onboarding or sth

tools/mage/deploy_onboard.go
```diff
@@ -19,73 +19,148 @@ package mage
...
+			registerCloudSec = false
+		}
+		if *integration.AWSAccountID == backendOutputs["AWSAccountId"] &&
+			*integration.IntegrationType == models.IntegrationTypeAWS3 &&
+			*integration.IntegrationLabel == genLogProcessingLabel(awsSession) {
+
+			logger.Infof("deploy: account %s is already registered for log processing", backendOutputs["AWSAccountId"])
+			registerLogProcessing = false
+		}
+	}
```
kostaspap: These checks are duplicated - the backend does them automatically if the when someone performs a PutIntegration. You can just invoke the PutIntegration operation - if it fails and the error is not InternalServerError, you can just continue. What I don't like about this is that it duplicates logic already present in the actual service.

web/src/components/Wizard/Wizard.tsx
```diff
@@ -1,10 +1,19 @@
 /**
+ * Panther is a scalable, powerful, cloud-native SIEM written in Golang/React.
```
kostaspap: I think this is already fixed in master.

api/lambda/source/models/api.go
```diff
@@ -88,7 +88,7 @@ type PutIntegrationSettings struct {
 
 // ListIntegrationsInput allows filtering by the IntegrationType or Enabled fields
 type ListIntegrationsInput struct {
-	IntegrationType *string `json:"integrationType" validate:"oneof=aws-scan aws-s3"`
+	IntegrationType *string `json:"integrationType" validate:"omitempty,oneof=aws-scan aws-s3"`
```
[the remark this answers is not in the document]
kostaspap: I figured it out later in the code. Looks good.

internal/core/source_api/ddb/scan.go
kostaspap: Saw later. Seems good.

deployments/backend.yml
```diff
@@ -53,9 +46,12 @@ Parameters:
...
 
   # Set automatically by "mage deploy"
-  S3BucketAccessLogs:
+  AuditLogsBucket:
     Type: String
-    Description: The name of the S3 bucket which stores access logs
+    Description: The name of the S3 bucket which stores Panther audit logs
+  S3AccessLogsBucket:
+    Type: String
+    Description: The name of the S3 bucket which stores S3 Access log (by default, same as AuditLogsBucket)
```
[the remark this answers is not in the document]
rleighton: Depends. If the customer gives THEIR bucket then no. We use the customer bucket for S3 logs but we use the AuditLog bucket for the other Panther logs (e.g., vpc flow, cloudtrail,etc). If the customer does NOT give us a bucket, then they are the same.

deployments/onboard.yml
rleighton: I wanted to avoid nested templates and allow us to do anything special we needed. I want to defer this suggestion since Austin's work will change it all anyway. He can decide.

deployments/onboard.yml
rleighton: Same comment as above.

tools/mage/deploy_onboard.go
rleighton: I was initially BUT when PutIntegration fails, that is an error and error means an alarm. I was getting annoyed with the alarms while doing many deployments. So, this avoids the alarms by checking. I'll add a comment explaining why this is done.

web/src/components/Wizard/Wizard.tsx
rleighton: I was just making the linter happy.

deployments/auxiliary/cloudformation/panther-cloudsec-iam.yml
```diff
@@ -36,7 +38,11 @@ Parameters:
   MasterAccountId:
     Type: String
     Description: DO NOT EDIT MANUALLY! Parameter is already populated with the appropriate value.
-    Default: ''
+    Default: '' # MasterAccountId
```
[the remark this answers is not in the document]
rleighton: I think this was left over when it was in the replacement logic.

internal/core/source_api/api/cloudsec_tmpl.go
rleighton: I'm going to revert this change because it causes confusion. I want to have a conversation _post_ v1 about removing the need to pull from a public s3 bucket. I feel pretty strongly this needs to be removed from the product for many reasons. Let's make that it's own discussion for later.

tools/mage/deploy_onboard.go
rleighton: This shows up in the UI and I want to make it clear to customers this is panther monitoring panther. I don't thing 'onboarding' in the name will convey that (might be confusing).  I think these should be `panther-account-...` so it is clear.  I'll make that change.

internal/core/source_api/api/cloudsec_tmpl.go
kostaspap: My personal preference would be to remove the launch stack button, have lambdas read everything from disk, remove the templates from S3 and just let people download the templates from UI and deploy it themselves through AWS Console/AWS CLI. But in the recent discussions the team decided to keep the launch stack button, in which case I think it's better to just read the template from S3. 
I don't like making this change as part of a PR without prior team agreement.

rleighton: Reverted

internal/core/source_api/api/vars_test.go
rleighton: Done

tools/mage/deploy_onboard.go
rleighton: Changed to start with 'panther-account-'

deployments/auxiliary/cloudformation/panther-cloudsec-iam.yml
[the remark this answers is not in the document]
rleighton: Removed

rleighton:
Replaced by https://github.com/panther-labs/panther/pull/491/

PR #482, opened 2020-03-21
Changing the way we serialize error messages

kostaspap:
## Background

Changing `errorMessage` to not be composed. So instead of this: 
```
{
  "errorMessage": "CheckIntegration failed: invalid input: Integration is not healthy",
  "errorType": "InvalidInputError"
}
```

we have this: 
```
{
  "errorMessage": "Integration is not healthy,
  "errorType": "InvalidInputError"
}
```

Some reasons for this: 
- Makes it easier to pass human readable messages to users
- The composed error message used to include the "errorType" which is anyway available as its own field. 
- Not sure including the method is needed anyway? The caller already knows the method that it has called. If we need it, we can add it as a separate field in the JSON. 

## Changes

- Changed error messages

## Testing

- Unit tests
- Deployed

@austinbyers  what do you think? I think it's a small change that doesn't break anything and would make it easier to display human readable text in the UI in case of error.


austinbyers:
> @austinbyers what do you think? I think it's a small change that doesn't break anything and would make it easier to display human readable text in the UI in case of error.

Glad it was an easy fix, works for me! We didn't originally intend to continue using this library, so we haven't spent much time on it so far, thanks for simplifying.

Did you run the integration tests though? This might have broken several of them

austinbyers:
This did in fact break integration tests for `outputs-api` and `users-api`. We were also sometimes relying on the error message generation, for example when a user doesn't exist, we returned just `Message: userId=...`, knowing that the "does not exist" was going to be added by the library.

Both issues have been fixed in #467 (https://github.com/panther-labs/panther/pull/467/commits/00b92188992dcd666ea92b43c2f956e8994b6a26)

PR #485, opened 2020-03-22
Remove cred cache in snapshot poller (keep client cache)

rleighton:
## Background
During testing of the release I was getting very repeatable InvalidClientTokenId errors from the snapshot poller. Looking at the code I notices 2 caches, 1 for creds and one for clients. Disabling the cred cache (but not the client cache) fixed the issue. Also, I noticed that most of the "list" operations (these were the ones failing) were not using the client cache, only the cred cache. I changed the code so that all uses of session client  call getClient() so they have an opportunity to get a cached client.

## Changes
* remove cred cache
* remove validation of assumed roles (not needed)
* update listing operations to use getClient()
* new single function to get core session object
* simplify tests

## Testing
mage test:ci
deployed to my dev account and exercised by driving snapshot poller via lambda test from snapshot_poller_scheduler.


internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -19,15 +19,48 @@ package aws
...
 
 	awsmodels "github.com/panther-labs/panther/internal/compliance/snapshot_poller/models/aws"
 )
 
-const MaxRetries = 6
+const (
+	// The amount of time credentials are valid
+	assumeRoleDuration = time.Hour
+	// Allows the credentials to trigger refreshing prior to the credentials actually expiring
+	assumeRoleExpiryWindow = 5 * time.Second
```
kostaspap: You can safely remove that, AWS SDK can deal with expiration. I don't think there is any real reason why this is preferable than the default value.

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -19,15 +19,48 @@ package aws
...
+
+	// assumeRoleFunc is the function to return valid AWS credentials.
+	assumeRoleFunc = assumeRole
+	// assumeRoleProviderFunc is the default function to setup the assume role provider.
+	assumeRoleProviderFunc = assumeRoleProvider
+)
+
+var sessionCache = make(map[string]*session.Session)
+
+func newSession(region string) (sess *session.Session) {
```
kostaspap: AWS SDK allows you to override the session region with the region you want on client creation (e.g. `s3.New(session, aws.NewConfig().WithRegion(...)`) so you don't need the session cache.

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -43,27 +76,50 @@ type cachedClient struct {
...
-	"iam":                    IAMClientFunc,
-	"kms":                    KmsClientFunc,
-	"lambda":                 LambdaClientFunc,
-	"rds":                    RDSClientFunc,
-	"redshift":               RedshiftClientFunc,
-	"s3":                     S3ClientFunc,
-	"waf":                    WafClientFunc,
-	"waf-regional":           WafRegionalClientFunc,
+// Functions used to build clients, keyed by service, use function not map to allow late binding by tests
+func lookupClientFunc(service string) func(session2 *session.Session, config *aws.Config) interface{} {
```
kostaspap: This can be simplified - you can just have one session and share it across the application. No need to keep multiple. I think it will simplify the code a bit.

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -80,22 +136,18 @@ func getClient(pollerInput *awsmodels.ResourcePollerInput, service string, regio
...
-		zap.L().Error("unable to assume role to build client cache", zap.Error(err))
-		return nil
-	}
+	sess := newSessionFunc(region)
+	creds := assumeRoleFunc(pollerInput, sess)
 
 	// Build the new client and cache it with the credentials used to build it
-	if clientFunc, ok := clientFuncs[service]; ok {
+	if clientFunc := lookupClientFunc(service); clientFunc != nil {
 		client := clientFunc(sess, &aws.Config{Credentials: creds})
```
kostaspap: I find it nicer to use the builder pattern given by the AWS SDK: `aws.NewConfig().WithCredentials(creds)`

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -109,3 +161,32 @@ func getClient(pollerInput *awsmodels.ResourcePollerInput, service string, regio
...
+func assumeRole(
+	pollerInput *awsmodels.ResourcePollerInput,
+	sess *session.Session,
+) *credentials.Credentials {
+
+	zap.L().Debug("assuming role", zap.String("roleArn", *pollerInput.AuthSource))
+
+	if pollerInput.AuthSource == nil {
+		panic("must pass non-nil authSource to AssumeRole")
+	}
```
kostaspap: Would it make sense to change the AuthSource to be `string` instead of `*string`?

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -51,26 +43,9 @@ const (
...
-	assumeRoleDuration = 15 * time.Minute
-	// Allows the credentials to trigger refreshing prior to the credentials actually expiring.
-	assumeRoleExpiryWindow = 5 * time.Second
-
-	// CredentialCache maps the integrationID to its assumed role credentials.
-	// The AssumeRole function will also check if the credential is expired before updating the cache.
-	CredentialCache = make(map[string]*credentials.Credentials)
-
 	// Default region to use when building clients for the individual resource poller
 	defaultRegion = "us-west-2"
```
kostaspap: nit: `endpoints.UsWest2RegionID`

internal/compliance/snapshot_poller/pollers/lambda_test.go
```diff
@@ -78,6 +50,27 @@ func TestBatchResources(t *testing.T) {
 	assert.Len(t, testBatches[2], 100)
 }
 
+/*
```
kostaspap: Is this meant to be fixed at some point? Do you mind making an issue and linking it here? Or just deleting the test if you think it doesn't give much value.

nhakmiller:
Did you test scanning multiple accounts in sequence? So finishing a scan of one account, then afterwards initiating a scan of another account?

internal/compliance/snapshot_poller/pollers/lambda_test.go
rleighton: Nick?

internal/compliance/snapshot_poller/pollers/aws/clients.go
rleighton: There is just 1 session, this is used as a "factory" method,. I'll change it look like you suggested below.

internal/compliance/snapshot_poller/pollers/aws/clients.go
rleighton: Done

internal/compliance/snapshot_poller/pollers/aws/clients.go
rleighton: Done

internal/compliance/snapshot_poller/pollers/aws/clients.go
rleighton: Done

internal/compliance/snapshot_poller/pollers/aws/clients.go
rleighton: Done

internal/compliance/snapshot_poller/pollers/aws/clients.go
rleighton: Nick?

internal/compliance/snapshot_poller/pollers/aws/poll.go
rleighton: Done

rleighton:
I'm going to wait to merge until the other PR with regional creds is in master and I can re-test.

internal/compliance/snapshot_poller/pollers/lambda_test.go
nhakmiller: Well this is in place for unit testing the handler. Seems a shame to get rid of it all just as we get the ability to easily mock the swagger APIs, which is what was blocking this. I would prefer to keep it if possible.

internal/compliance/snapshot_poller/pollers/aws/clients.go
nhakmiller: Some places require a `string`, some a `*string`. Just depends on who you want to be responsible for the dereference really.

rleighton: I'll leave it.

rleighton:
I need to do a few more changes based on testing, working those now.

internal/compliance/snapshot_poller/pollers/aws/cloudformation_stack.go
```diff
@@ -125,17 +139,17 @@ func getStack(svc cloudformationiface.CloudFormationAPI, stackName string) *clou
...
 			stacks = append(stacks, page.Stacks...)
 			return true
 		})
 
 	if err != nil {
-		utils.LogAWSError("CloudFormation.DescribeStacksPages", err)
+		return nil, errors.Wrap(err, "CloudFormation.DescribeStacksPages")
 	}
-	return
+	return stacks, nil
```
nhakmiller: If there is no naked return here, there is no need for named return values.

internal/compliance/snapshot_poller/pollers/aws/acm_certificate.go
```diff
@@ -162,15 +175,11 @@ func PollAcmCertificates(pollerInput *awsmodels.ResourcePollerInput) ([]*apimode
 	acmCertificateSnapshots := make(map[string]*awsmodels.AcmCertificate)
 
 	for _, regionID := range utils.GetServiceRegions(pollerInput.Regions, "acm") {
-		sess := session.Must(session.NewSession(&aws.Config{Region: regionID}))
-
-		creds, err := AssumeRoleFunc(pollerInput, sess)
+		acmSvc, err := getAcmClient(pollerInput, *regionID)
 		if err != nil {
-			return nil, err
```
nhakmiller: Continuing instead of returning here means that transient errors will not be retried by lambda automatically. I'm not sure if this is desired behavior, as re-scans will not occur until up to 24 hours later, which means if a region gets skipped once due to a temporary error there could be a 48 hour gap between scans, which is a moderately long time for something to be out of compliance without an alert.

rleighton: Good point .. I'll go back and review similar loops and fix.

rleighton: Fixed all the loops.

PR #486, opened 2020-03-22
Tune noisy alarms for too-old-sqs and lambda warns

rleighton:
## Background
During testing these alarms proved too sensitive:
1) "too old in sqs q" : snapshot poller would set off 
2) "lambda warns": compliance_api would set off when there were checks for new infra

## Changes
1) "too old in sqs q": threshold changed from 5min to 15min
2) "lambda warns": threshold changed from 5 in 5min to 25 in 5min

## Testing
mage test:ci
ran in dev account


tools/cfngen/cloudwatchcf/alarms_sqs.go
```diff
@@ -51,8 +51,8 @@ func generateSQSAlarms(resource map[interface{}]interface{}) (alarms []*Alarm) {
 		alarms = append(alarms, NewSQSAlarm(queueName, "SQSDeadLetters", "ApproximateNumberOfMessagesVisible",
 			"has failed items from "+referenceQueue).SumCountThreshold(0, 60*5))
 	} else { // regular q's
-		// nothing in our queues should be older than 5min
-		const tooOldSec float32 = 60.0 * 5.0
+		// nothing in our queues should be older than 15min
```
kostaspap: Seems good. I would be fine to increase it even further if feel we want to e.g. 30 mins.

tools/cfngen/cloudwatchcf/testdata/generated_test_alarms.json
```diff
@@ -81,16 +78,15 @@
...
-            "Value": { "Ref": "AppsyncId" }
+            "Ref": "AlarmTopicArn"
           }
         ],
+        "TreatMissingData": "notBreaching",
+        "Namespace": "AWS/AppSync",
+        "MetricName": "4XXError",
+        "Dimensions": [{ "Name": "GraphQLAPIId", "Value": { "Ref": "AppsyncId" } }],
         "ComparisonOperator": "GreaterThanThreshold",
         "EvaluationPeriods": 1,
```
kostaspap: This also seems it can be too sensitive? 4XX are errors caused by user activity e.g. invalid user input. I would raise it to a bit more e.g. 3 periods of 300 each instead of 1 period of 300. 1 period of 300 means that a sudden spike of errors can trigger an alert. It would be better to trigger an alert in case of persistent such errors though.

tools/cfngen/cloudwatchcf/testdata/generated_test_alarms.json
```diff
@@ -744,16 +714,15 @@
...
-            "Value": "test-lambda"
+            "Ref": "AlarmTopicArn"
           }
         ],
+        "TreatMissingData": "notBreaching",
+        "Namespace": "AWS/Lambda",
+        "MetricName": "Throttles",
+        "Dimensions": [{ "Name": "FunctionName", "Value": "test-lambda" }],
         "ComparisonOperator": "GreaterThanThreshold",
         "EvaluationPeriods": 1,
```
kostaspap: See comment above. Such issues can appear at any moment - AWS gets temporary issues all the time. I think it's good to increase the periods we are evaluating over to e.g. 3 periods (15').

tools/cfngen/cloudwatchcf/alarms_lambda.go
```diff
@@ -74,7 +74,7 @@ func generateLambdaAlarms(resource map[interface{}]interface{}, settings *config
 	// warns from metric filter (application logs)
 	// NOTE: it is important to not set units because the metric filter values have no units
 	alarms = append(alarms, NewLambdaMetricFilterAlarm("LambdaApplicationWarns", lambdaWarnsMetricFilterName,
-		"is warning", resource).SumNoUnitsThreshold(5, 60*5) /* tolerate a few warnings before alarming */)
+		"is warning", resource).SumNoUnitsThreshold(25, 60*5) /* tolerate a few warnings before alarming */)
```
kostaspap: I think it's better to increase the evaluation periods instead of increasing the threshold to be honest. I think we should care about persistent issues tbh.

rleighton:
Agree with your suggestions. I'll do them in another PR where I review all the settings. Having them sensitive has been very helpful with testing but we should dampen some of them before release.

PR #488, opened 2020-03-22
Add new Log Analysis onboarding screens 

3nvi:
## Background

This PR adds the new log onboarding screens to the application. These mimic 1:1 the new screens that were already added for compliance.

## Changes

- Migrate to new sources-api
- Add LogAnalysis CRUD screens
- Bugfixes on both compliance and loganalysis screens

## Testing

- Locally. Counting on @kostaspap as well :P


api/gateway/resources/client/operations/delete_resources_responses.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/go-openapi/runtime"
 	strfmt "github.com/go-openapi/strfmt"
 
-	"github.com/panther-labs/panther/api/gateway/resources/models"
+	models "github.com/panther-labs/panther/api/gateway/resources/models"
```
3nvi: @kostas should these file changes be here? For some reason they were autogenerated for me :S

kostaspap: It could be that swagger changed version. Since compilation passes, we are fine :)

kostaspap:
Actually, updating the account ID shouldn't be possible (like we do in cloudsec onboarding). Can you grey that out in the "edit" page?

kostaspap:
Another small issue: If someone tries to update the label, say from `test` to `test2`, the instructions to run the update are including the new label: 
```
Alternatively, you can download it and deploy it through the AWS CLI with the stack name panther-log-analysis-test2
```
They should either include the old one or just not let people update the label.

kostaspap:
Chatted offline with @3nvi . Merging now to unblock work, can fix small issues tomorrow.

PR #495, opened 2020-03-23
Doc Review [Pt 4] + More

jacknagz:
to @kostaspap @3nvi 

## Background

Continued documentation review. I also found some small edits to the alert delivery and wording in the UI.

## Changes

- Refresh log onboarding docs (almost done, just need to review the SNS part)
- Prepend `New Alert` to alerts to match the cloud security format of `Policy Failure` and to distinguish the two
- Small tweaks in the UI wording for the log analysis workflow
- Show the Integration Label if the health check fails vs the account Id 

## Testing

- mage fmt test:go


docs/gitbook/log-analysis/rules/README.md
```diff
@@ -55,6 +53,8 @@ Python provides high flexibility in defining your rules, and the following libra
 | `policyuniverse` | `1.3.2.1` | Parse AWS ARNs and Policies | Apache v2 |
 | `requests`       | `2.22.0`  | Easy HTTP Requests          | Apache v2 |
 
+If you would like to request a new library, [please file a feature request](https://github.com/panther-labs/panther/issues/new/choose)!
```
kostaspap: * Users can define their own Lambda [layer](https://github.com/panther-labs/panther/blob/master/deployments/panther_config.yml#L70)  if they want more libraries - Python libraries are fully customizable
* They can specify the libraries they want as part of the configuration: https://github.com/panther-labs/panther/blob/master/deployments/panther_config.yml#L106-L109

web/src/components/wizards/LogSourceWizard/SourceConfigurationPanel/SourceConfigurationPanel.tsx
```diff
@@ -91,7 +91,7 @@ const SourceConfigurationPanel: React.FC = () => {
             <Field
               name="s3Prefix"
               as={FormikTextInput}
-              label="S3 Object Prefix"
```
kostaspap: I am not sure I like this rename. I think this might "hint" that they can introduce a regex, while they should just put the Prefix of the S3 objects. 
Something like how AWS does it for S3 event notifications: 
Screen Shot 2020-03-23 at 11 29 26 AM

internal/core/alert_delivery/outputs/outputs.go
```diff
@@ -117,7 +117,7 @@ func generateDetailedAlertMessage(alert *alertmodels.Alert) string {
 
 func generateAlertTitle(alert *alertmodels.Alert) string {
 	if alert.Title != nil {
-		return *alert.Title
+		return "New Alert: " + *alert.Title
```
kostaspap: Nice

internal/core/source_api/api/put_integration.go
```diff
@@ -64,7 +64,7 @@ func (api API) PutIntegration(input *models.PutIntegrationInput) (*models.Source
 			zap.Any("input", input))
 		return nil, &genericapi.InvalidInputError{
 			Message: fmt.Sprintf("source %s did not pass configuration check because of %s",
-				*input.AWSAccountID, reason),
```
nhakmiller: `IntegrationLabel` is not a required field, account ID is. I think we should stick with account ID, or make `IntegrationLabel` a required field of `PutIntegrationInput` in `api/lambda/source/models/api.go`.

kostaspap: It actually is required.

nhakmiller: I see, that has changed recently 👍

kostaspap: yup

web/src/components/wizards/LogSourceWizard/SourceConfigurationPanel/SourceConfigurationPanel.tsx
jacknagz: I think `Filter` is clear, it's not `Pattern`

PR #496, opened 2020-03-23
Adding further validations for fields

kostaspap:
## Background

* Fixing deployment issues.
* Added extra validations for API.

## Changes

- Added kmsKeyArn validator. If KmsKey is provided, it has to adhere to KMS Key ARN standard
- Added basic validation for LogTypes, S3Bucket, S3Prefix, KmsKey fields of sources_api. The most appropriate to handle this in the log term is that we create separate endpoints to the backend for Log and CloudSec onboarding but it is a bigger change to do atm. 
- Updated self-onboarding flow to populate appropriate values for onboarding: 
 - Log types for log onboarding
 - Marking remediations and real-time events as enabled since the CFN stack deployed anyway has it. 
- Fixing issue where when a user would delete the `panther-app-onboard`, re-deployment would fail. 

## Testing

- Deployed to my account. 
- Unit tests pass


tools/mage/deploy_onboard.go
```diff
@@ -152,6 +154,7 @@ func registerPantherAccount(awsSession *session.Session, bucketOutputs, backendO
 					IntegrationType:  aws.String(models.IntegrationTypeAWS3),
 					UserID:           aws.String(mageUserID),
 					S3Bucket:         aws.String(bucketOutputs["AuditLogsBucket"]),
+					LogTypes:         aws.StringSlice([]string{"AWS.S3ServerAccess"}),
```
rleighton: This should be a list of the "enabled" log types OR better, loop over each "enabled" log type , put in a single integration for each. Right now only CT and GD on configurable , so you can look at the "settings" for those and put the rest into a single integration , or separate as you like. We have support for: CT, GD, ALB, VPC, S3Access. One thing check on is getting the "prefix" right. S3Access has a prefix of the whole bucket (no prefix). The others follows the AWS standard of <bucket>/AWSLogs/<account>/<service prefix>.
image

kostaspap: To fix this we have to fix the whole deployment process really (our discussion of onboading each log as separate CFN stack and resource). 
I will just add all the log sources under this prefix for now, I don't want to increase the scope of this PR to update that part of the code, it will require extended updates to CFN templates.

kostaspap: @rleighton  take a look again when you get some time. I think the onboarding flow needs to be re-worked to deploy individual CFN stacks. When this is done, it will be easier to add individual entries now. The fact that currently in any case the users can only offboard all log at once and not individually, so by having one log source with multiple log types we actually representing more accurately the current onboarding mechanism.

kostaspap: Screen Shot 2020-03-23 at 4 40 59 PM
This is how it looks for me (I have GD enabled that goes into different bucket, that's why it was not created)

PR #498, opened 2020-03-23
Improving stability when running rules

### LINKED ISSUE #460: Alerts should use default title and dedup string in case of an exception
### Description

In case the `title` or `dedup` Python methods fail, the event fails to match a rule. We should instead fallback to default values. 

### Acceptance Criteria

- If "title" method of a rule fails, fallback to default value
- If "dedup" method of a rule fails, fallback to default value

kostaspap:
## Background

Fixes #460 
In case title or dedup functions throw exception, we use default values instead of dropping event. 
This is meant to improve system stability and to make sure that no crucial events will be ignored. 

## Changes

- Switching to use default values for alert title and dedup string in case the dedup function or title function throw exception

## Testing

- Unit testing.

### DIFF
--- a/internal/log_analysis/rules_engine/src/rule.py
+++ b/internal/log_analysis/rules_engine/src/rule.py
@@ -1,19 +1,3 @@
-# Panther is a scalable, powerful, cloud-native SIEM written in Golang/React.
-# Copyright (C) 2020 Panther Labs Inc
-#
-# This program is free software: you can redistribute it and/or modify
-# it under the terms of the GNU Affero General Public License as
-# published by the Free Software Foundation, either version 3 of the
-# License, or (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU Affero General Public License for more details.
-#
-# You should have received a copy of the GNU Affero General Public License
-# along with this program.  If not, see <https://www.gnu.org/licenses/>.
-
 import os
 import sys
 import tempfile
--- a/internal/log_analysis/rules_engine/src/rule.py
+++ b/internal/log_analysis/rules_engine/src/rule.py
@@ -95,7 +95,7 @@ class Rule:
         if not hasattr(self._module, 'rule'):
             raise AssertionError("rule needs to have a method named 'rule'")
 
-        if hasattr(self._module, 'dedup'):
+        if hasattr(self._module, 'dedup_string'):
             self._has_dedup = True
         else:
             self._has_dedup = False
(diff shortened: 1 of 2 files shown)


kostaspap:
> How will users know if their dedup/title functions are throwing exceptions? Could we perhaps use the defaults to send an alert, but still set the status to exception? Does log processing normally create an alert if a rule body throws an exception, or just drops it?

Actually the rules don't have a "status" like policies have... if the `rule` method throws an exception, we log it and continue. We consider that the event didn't match a rule. 
Note that the event will still be stored in Athena (we store all events). But it will be considered that it didn't match the rule.

nhakmiller:
> > How will users know if their dedup/title functions are throwing exceptions? Could we perhaps use the defaults to send an alert, but still set the status to exception? Does log processing normally create an alert if a rule body throws an exception, or just drops it?
> 
> Actually the rules don't have a "status" like policies have... if the `rule` method throws an exception, we log it and continue. We consider that the event didn't match a rule.
> Note that the event will still be stored in Athena (we store all events). But it will be considered that it didn't match the rule.

Hmm then it seems quite likely users will never know when their rules are throwing exceptions. Might be worth considering triggering an alert on exceptions in general then.

kostaspap:
> > > How will users know if their dedup/title functions are throwing exceptions? Could we perhaps use the defaults to send an alert, but still set the status to exception? Does log processing normally create an alert if a rule body throws an exception, or just drops it?
> > 
> > 
> > Actually the rules don't have a "status" like policies have... if the `rule` method throws an exception, we log it and continue. We consider that the event didn't match a rule.
> > Note that the event will still be stored in Athena (we store all events). But it will be considered that it didn't match the rule.
> 
> Hmm then it seems quite likely users will never know when their rules are throwing exceptions. Might be worth considering triggering an alert on exceptions in general then.

When this happens we log at error level: https://github.com/panther-labs/panther/blob/83c6ed86044aa65e05a4e4f32895890a80c8c0bb/internal/log_analysis/rules_engine/src/engine.py#L51 . If this happens repeatedly a CW Alarm will trigger. That's the best mechanism we have right now. But we can definitely improve, I'm happy to hear suggestions :)

nhakmiller:
> When this happens we log at error level:
> 
> https://github.com/panther-labs/panther/blob/83c6ed86044aa65e05a4e4f32895890a80c8c0bb/internal/log_analysis/rules_engine/src/engine.py#L51
> 
> . If this happens repeatedly a CW Alarm will trigger. That's the best mechanism we have right now. But we can definitely improve, I'm happy to hear suggestions :)

I think most users who are writing rules will not look at AWS to make sure the rules are working, if the issue is not displayed in the UI it will likely be missed for a long time. My suggestion is to still create an alert in the case of an exception, but with some indication that an exception was thrown during rule evaluation. Maybe even just replacing the dedup string/title with "An exception was thrown during rule evaluation".

Ideally this means a rule with an exception in it makes a lot of noise so that someone goes to look at it and fixes the exception.

PR #499, opened 2020-03-23
Clean up top-level web-related files

3nvi:
## Background

This PR doesn't add a feature or mess with the application code. All it does is move our development tools inside the `/web` folder, so that the top-level directory is more clean, while front-end related configuration files are only present in the `/web` where they belong. For example, the un-needed top-level `webpack.config.js` has been removed since it was polluting dirs without adding value in its location. 

With the exception of webpack, all the other files are dev-only migrations that will have no effect in a production deployment, thus the changes are considered safe. 

## Changes

- Convert `.husky.js` to `.huskyrc`
- Move eslint, babel, lint-staged, tsConfig & graphlq-codegen files inside the `/web` repo
- Update commands to reference the update location of the config files
- Delete top level `webpack.config.js` and migrate its check to `web/scripts`

## Testing

- Local prod deployment
- Local dev deployment
- Dev deployment


.huskyrc
```diff
@@ -0,0 +1,5 @@
+{
```
3nvi: @rleighton re-enabling hooks that will **only** run prettier on YAML/JSON files that are being committed (and nothing else) after our conversation here:

https://github.com/panther-labs/panther/pull/321

All the test cases where prettier created issues have been remediated (see [here](https://github.com/panther-labs/panther/pull/321#issuecomment-595673614) to understand why those issues occurred in the first place). In addition, if you want to opt out of it for certain files, please add an entry to the top-level `.prettierignore` file so that `prettier` won't affect you

docs/gitbook/development.md
```diff
@@ -79,9 +79,8 @@ You can easily chain `mage` commands together, for example: `mage fmt test:ci de
 
 ## Testing
 
-1. Run backend test suite: `mage test:ci`
-2. Run frontend test suite: `npm run lint`
-3. Run integration tests against a live deployment: `mage test:integration`
```
3nvi: that doc was wrong

package.json
```diff
@@ -121,8 +121,8 @@
     "build": "node web/scripts/build.js",
     "start": "node web/scripts/start.js",
     "serve": "node web/scripts/serve.js",
-    "generate-gql": "graphql-codegen",
```
3nvi: rename commands and explicitly add the config file

tools/mage/test_namespace.go
```diff
@@ -236,9 +236,9 @@ func testWeb() bool {
 		return false
 	}
 
-	logger.Info("test:web: npm run lint")
-	if err := sh.RunV("npm", "run", "lint"); err != nil {
-		logger.Errorf("npm lint failed: %v", err)
+	logger.Info("test:web: npm run validate")
```
3nvi: renamed `npm run lint` to `npm run validate`, since it doesn't "change" code, it just "validates" that everything is in order

web/.eslintrc.js
```diff
@@ -0,0 +1,72 @@
+const path = require('path');
```
3nvi: This file is a port from `.eslintrc`. No new code was added, I just converted the JSON config file to a JS config file, while moving it inside `/web`

web/lint-staged.config.js
```diff
@@ -25,15 +25,15 @@ module.exports = {
   /*
    * Run prettier TS, JS, JSON, YAML and Markdown files found anywhere in the project
    */
-  '*.{ts,tsx,js,md,yaml,yml,json}': ['prettier --write', 'git add'],
```
3nvi: `git add` is now automatically ran by husky (since recent versions). No need to add it here as well

PR #500, opened 2020-03-23
Generating and storing CFN stack name

kostaspap:
## Background

The frontend is currently generating the CFN stack name for an integration when a user clicks the "Launch Stack" button. This stack name is used subsequently in order to provided guidance to the users when they need to edit/update their sources. 
The problem is, the frontend was currently generating the stack name using the IntegrationLabel. If a user however updates the label of an integration, then the instructions provided by the frontend will be wrong since they will be using the *current* label, not the label with which the source was created. 

After discussion with @3nvi  we decided to move the stack-name generate mechanism to the backend and make sure the backend keeps the CFN stack name that will be used when instructing users to update their templates. 

## Changes

- Sources_api now generating the CFN stack name
- Sources_api now returning CFN stack name in the "getTemplate" method

## Testing

- Unit tests. 
- Deployed and verified it works in my account.

### DIFF
--- a/internal/core/source_api/api/get_integration_template.go
+++ b/internal/core/source_api/api/get_integration_template.go
@@ -37,8 +37,12 @@ const (
 	TemplateBucket           = "panther-public-cloudformation-templates"
 	TemplateBucketRegion     = endpoints.UsWest2RegionID
 	CloudSecurityTemplateKey = "panther-cloudsec-iam/v1.0.0/template.yml"
-	LogProcessingTemplateKey = "panther-log-analysis-iam/v1.0.0/template.yml"
-	cacheTimeout             = time.Minute * 30
+	LogAnalysisTemplateKey   = "panther-log-analysis-iam/v1.0.0/template.yml"
+
+	LogAnalysisStackNameTemplate = "panther-log-analysis-setup-%s"
+	CloudSecStackName            = "panther-cloudsec-setup"
+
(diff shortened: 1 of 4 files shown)


internal/core/source_api/api/get_integration_template.go
```diff
@@ -37,8 +37,12 @@ const (
 	TemplateBucket           = "panther-public-cloudformation-templates"
 	TemplateBucketRegion     = endpoints.UsWest2RegionID
 	CloudSecurityTemplateKey = "panther-cloudsec-iam/v1.0.0/template.yml"
-	LogProcessingTemplateKey = "panther-log-analysis-iam/v1.0.0/template.yml"
-	cacheTimeout             = time.Minute * 30
+	LogAnalysisTemplateKey   = "panther-log-analysis-iam/v1.0.0/template.yml"
+
+	LogAnalysisStackNameTemplate = "panther-log-analysis-setup-%s"
```
rleighton: Do we validate/constrain length of stackname? I'm sure there is a limit.

kostaspap: I'm putting a limit in the Integrationlabel here: https://github.com/panther-labs/panther/pull/496  , in order to avoid hitting the 64 character limit for IAM Roles. 
CFN Stack name limit is 128 characters.

austinbyers:
PR number 💯 ✖️ 5️⃣ !

PR #501, opened 2020-03-23
Allow server to calculate CFN stack name for Integrations

3nvi:
## Background

This PR makes sure to allow the server to provide the `stackName` instead of thee front-end calculating it. This helps us in cases where we prompt the user to Delete or Update a particular CFN stack when updating/deleting an Integration.

The corresponding CFN stackname was not known to the front-end 

## Changes

- Add `stackName` to graphql schema
- Utilize it in all places where the front-end was using a placeholder value for it

## Testing

- Deployed to dev
- Added & Updated cloud-sec integration
- Added & Updated log-analysis integration
- Counting on @kostaspap to validate

### DIFF
--- a/web/src/components/wizards/LogSourceWizard/StackDeploymentPanel/StackDeployment.tsx
+++ b/web/src/components/wizards/LogSourceWizard/StackDeploymentPanel/StackDeployment.tsx
@@ -102,7 +102,7 @@ const StackDeployment: React.FC = () => {
             <a
               href="#"
               title="Download Cloudformation template"
-              download={`${stackName}.yaml`}
+              download={`${stackName}.yml`}
               ref={downloadRef}
               onClick={() => setStatus({ cfnTemplateDownloaded: true })}
             >
@@ -124,7 +124,7 @@ const StackDeployment: React.FC = () => {
               <a
                 href="#"
                 title="Download Cloudformation template"
-                download={`${initialValues.initialStackName}.yaml`}
+                download={`${initialValues.initialStackName}.yml`}
                 ref={downloadRef}
                 onClick={() => setStatus({ cfnTemplateDownloaded: true })}
               >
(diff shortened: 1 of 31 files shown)


kostaspap:
Another change we wanted is that KmsKey and S3Prefix are not passed to backend to the `getIntegrationTemplate` when they are not populated. Is this being addressed in this PR, I am not sure I can see the relevant changes.

web/src/components/wizards/LogSourceWizard/StackDeploymentPanel/StackDeployment.tsx
```diff
@@ -112,7 +98,17 @@ const StackDeployment: React.FC = () => {
             Alternatively, you can download it and deploy it through the AWS CLI with the stack name{' '}
             <b>{stackName}</b>
           </Text>
-          {downloadTemplateLink}
+          <Text size="large" color="blue300" is="span">
+            <a
+              href="#"
+              title="Download Cloudformation template"
+              download={`${stackName}.yaml`}
```
kostaspap: We tend to name them `.yml` in our codebase, I think better follow same convention.

web/src/components/wizards/LogSourceWizard/StackDeploymentPanel/StackDeployment.tsx
```diff
@@ -124,7 +120,17 @@ const StackDeployment: React.FC = () => {
             <Text size="large" color="grey200" mr={1}>
               1.
             </Text>
-            {downloadTemplateLink}
+            <Text size="large" color="blue300" is="span">
+              <a
+                href="#"
+                title="Download Cloudformation template"
+                download={`${initialValues.initialStackName}.yaml`}
```
kostaspap: Same here. Rename to `.yml`

3nvi:
> getIntegrationTemplate

You are right! Added.

kostaspap:
We limit the integration label to 32 characters (due to restrictions in the IAM role name size). I think it's better to have such check in the UI as well.

3nvi:
> We limit the integration label to 32 characters (due to restrictions in the IAM role name size). I think it's better to have such check in the UI as well.

Done

PR #533, opened 2020-03-23
Fix/rules dedup

alexmylonas:
## Background

Fixing changes for #475

## Changes

- Fixes for #475 
## Testing

- Locally
- Deployed env


api/graphql/schema.graphql
```diff
@@ -119,7 +119,7 @@ type AlertDetails {
   eventsMatched: Int!
```
3nvi: `dedupPeriodMinutes` should never be optional. @kostaspap how come we have it required in `AlertsDetails`, but optional in `RuleDetails`? Seems like a bug

3nvi:
@alexmylonas Are we missing the showcasing of dedup information in the Rules details page? I'm referring to the first criteria of https://github.com/panther-labs/panther/issues/349.

api/graphql/schema.graphql
kostaspap: We always had an dedupPeriodMinutes that was used when generating alarms and that period was 1hr. This hardcoded 1hr value was used in a different part of the system, the one responsible for generating alerts. We recently made the change to make it configurable on Rule level and thus we introduced this value to the AnalysisAPI. 
The reason it was optional for Rule was because until recently the Rule's didn't really have this field so I had made it optional for backwards compatibility reasons mainly. For the reason mentioned above, the field was always present in the AlertDetails. 

I agree that can be updated to reflect the current reality. But is it a bug aka causing the software to fail/crash or is it an inconsistency you are referring to?

3nvi: Currently, we have `dedupPeriodMinutes` as required in `AlertDetails` objects & optional in `RuleDetails` objects.

How can an alert have it if a rule doesn't have it. This for me and according to @nhakmiller is a bug. The behavior should be consistent, in that the alert should always **mimic** what the rule has

3nvi:
> @alexmylonas Are we missing the showcasing of dedup information in the Rules details page? I'm referring to the first criteria of #349.

@kostaspap can you validate?

api/graphql/schema.graphql
kostaspap: I will fix it. But I want to be sure on this: Does this somehow affect the product behavior or not?

3nvi:
> kostaspap can you validate?

Turns out it was my mistake, it's there

api/graphql/schema.graphql
kostaspap: @3nvi I fixed it with https://github.com/panther-labs/panther/pull/543

PR #536, opened 2020-03-23
Returning original error

kostaspap:
## Background

Deploying on a clean account was not succeeding. 

## Changes

- Mage uses a utility method to identify if the error returned when describing a stack is that a stack doesn't exist. The utility method is here: https://github.com/panther-labs/panther/blob/master/tools/mage/util_cfn.go#L131-L138

Updated code to return original exception

## Testing

- Was able to deploy


kostaspap:
> Ah, I made a similar fix in #467, adding this utility method:
> 
> https://github.com/panther-labs/panther/blob/175fe02ebda5ac124561e5ccb9ff74ed407e6598/tools/mage/util.go#L102-L110
> 
> Feel free to merge this as-is, and #467 already applies the fix more generally

Nice! Yeah, was trying to fix another bug today and introduced that method. Seems that I caused another one.

PR #544, opened 2020-03-24
Added default functions for dedup and title

kostaspap:
## Background

Adding default rule functions when users decided to create a new rule. This is meant to motivate people to user these functions instead of relying just in docs. 

## Changes

- Adding `dedup` and `title` functions. 
- Added missing licence headers. 

## Testing

- Deployed and verified that they show up properly. 
Screen Shot 2020-03-24 at 2 07 48 PM

@3nvi  can you check that the code looks good? 
@jacknagz can you check the wording?


nhakmiller:
Should we inform users to remove these if they're not going to use them? Leaving them blank may have some strange results right? Or will it go to defaults if these return empty strings?

web/src/constants.ts
```diff
@@ -34,7 +34,14 @@ export const INCLUDE_SPECIAL_CHAR_REGEX = new RegExp('(?=.*[!@#\\$%\\^&\\*;:,.<>
...
 
-export const DEFAULT_RULE_FUNCTION = 'def rule(event):\n\t# Write your code here.\n\treturn False';
+export const DEFAULT_RULE_FUNCTION =
+  'def rule(event):\n\t# Write your rule code here.\n\treturn False';
+
+export const DEFAULT_TITLE_FUNCTION =
+  "def title(event):\n\t# Write custom code for alert title generation here.\n\treturn ''";
+
+export const DEFAULT_DEDUP_FUNCTION =
+  "def dedup(event):\n\t# Write custom code for deduplication string generation here.\n\treturn ''";
```
jacknagz: we can drop `custom code`

jacknagz: In this and the above line

PR #545, opened 2020-03-24
Fix clearing of s3 event notifications and add delay before deleting bucket in mage:teardown

rleighton:
Fix clearing of s3 event notifications and add delay before deleting bucket bucket in mage:teardown

## Background
The api to clear event notifications requires an empty config struct to clear (not nil). Documentation was not clear. 

## Changes
* Added empty config doc to api call, moved to section of code where we leave the bucket
* Added short delay before deleting bucket as I have experienced errors here saying bucket was not empty (but it was). This forced me to run teardown again (annoying). This delay make this much less likely.

## Testing
Changed code to _just_ remove notifications (not delete bucket), ran, checked bucket in console to confirm notifications were not configured.
mage:teardown

### DIFF
--- a/tools/mage/teardown.go
+++ b/tools/mage/teardown.go
@@ -352,12 +352,6 @@ func destroyPantherBuckets(awsSession *session.Session, bucketNames []*string) {
 //
 // Or, if there are too many objects to delete directly, set a 1-day expiration lifecycle policy instead.
 func removeBucket(client *s3.S3, bucketName *string) {
-	// remove any notifications (best effort)
-	notificationInput := &s3.PutBucketNotificationConfigurationInput{
-		Bucket: bucketName,
-	}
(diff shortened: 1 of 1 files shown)


tools/mage/teardown.go
```diff
@@ -424,7 +428,7 @@ func removeBucket(client *s3.S3, bucketName *string) {
 	if err != nil {
 		logger.Fatalf("failed to batch delete objects: %v", err)
 	}
-
+	time.Sleep(time.Second / 4) // short pause since S3 is eventually consistent to avoid next call from failing
```
kostaspap: Is that enough time? What about leaving it wait a bit more just to be more sure (e.g. 1 second? 5 seconds)? Given that it's a teardown command, I'm thinking it's better to be more reliable than try to save few seconds.

rleighton: I think this is fine. Used to work _most_ of the time

PR #547, opened 2020-03-24
Baby you' re a firework

3nvi:
## Background

You just gotta ignite the light and let it shine

download (2)

Just own the night like the 4th of July

ezgif com-video-to-gif

**P.S. It looks better than this crappy GIF**

## Changes

- Show fireworks for the **first** user during his **first** login

## Testing

- Local


austinbyers:
🎆 amazing

PR #548, opened 2020-03-24
Refactor mage show:schemas so it stays up to date and conforms to other targets

rleighton:
## Changes
Refactor mage show:schemas so it stays up to date and conforms to other targets

## Testing
mage show:schemas


jacknagz:
@rleighton thank you :) wasn't sure how to do this at the time

rleighton:
> @rleighton thank you :) wasn't sure how to do this at the time

This feature!

PR #551, opened 2020-03-24
Rule Field Mappings in BulkUpload

nhakmiller:
## Background

In order to support a more intuitive rule writing experience, the BulkUpload endpoint now supports users specifying the fields `RuleID` and `LogType` for rule configs, and `Log` and `LogType` for rule test configs when uploading YAML/JSON specification files.

## Changes

- BulkUpload now re-maps `RuleID` and `LogType` to `PolicyID` and `ResourceType` when constructing test cases

## Testing

- Tested in my dev account


internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -213,33 +213,39 @@ func extractZipFile(input *models.BulkUpload) (map[models.ID]*tableItem, error)
...
+			}
+			if len(config.ResourceTypes) == 0 {
+				analysisItem.ResourceTypes = config.LogTypes
 			}
 		}
 
 		for i, test := range config.Tests {
-			resource, err := jsoniter.MarshalToString(test.Resource)
+			// Same re-mappings as above
+			if test.Resource == nil {
```
rleighton: Maybe add a comment explaining why test.Resource == nil means "rule"

nhakmiller:
The comments I added caused CI to fail because the function became too long...

Issue #492, opened 2020-03-23
Docs: Update Versions for V1.0


jacknagz:
### Description

Make sure to update all references to tagged versions in docs

### Acceptance Criteria

N/A

PR #556, opened 2020-03-24
Fix swagger/build race, add v1 tags to docs

austinbyers:
## Background

If Swagger needs to be built, sometimes there will be a race condition on a fresh deploy:

```
08:07:46        INFO    build:api: generating Go SDK for 4 APIs (api/gateway/*/api.yml)
08:07:46        FATAL   failed to read deployments/bootstrap.yml: open deployments/bootstrap.yml: no such file or directory
```

This happens because `build:api` (swagger) has to change the working directory (due to a bug in swagger). If the bootstrap stack is trying to deploy while the directory is different, it will fail.

The fix is to build swagger apis before starting the bootstrap goroutine

Closes: #492 

## Changes

- Update v1 in documentation
- Build swagger before starting bootstrap goroutine
    - Technically, `build:api` now happens twice, since `build:lambda` also calls it, but the second time no changes will need to be made so it will return immediately (without changing directories)

## Testing

- `touch api/gateway/*/api.yml`
- `mage -v deploy`

No error in the deploy, first swagger build happens by itself, second swagger build shows no changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,6 @@
 </p>
 
 <p align="center">
-  <i>Panther is currently in beta</i>
-</p>
-
-<p align="center">
   <a href="https://docs.runpanther.io">Documentation</a> |
   <a href="https://docs.runpanther.io/quick-start">Quick Start</a> |
   <a href="https://blog.runpanther.io">Blog</a> |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -6,10 +6,6 @@
   <b>A Cloud-Native SIEM for the Modern Security Team</b>
 </p>
 
-<p align="center">
-  <i>Panther is currently in beta</i>
-</p>
-
```
nhakmiller: 🎊

PR #559, opened 2020-03-24
Fix remediation role construction

nhakmiller:
## Background

Remediation role ARN construction was not updated after changing to regionally suffixed roles.

## Changes

- Change role generation to take suffix into account

## Testing

- Deployed internally and observed remediations work

### DIFF
--- a/deployments/cloud_security.yml
+++ b/deployments/cloud_security.yml
@@ -701,7 +701,6 @@ Resources:
       Environment:
         Variables:
           LOGGING_LEVEL: !If [DebugEnabled, DEBUG, INFO]
-          ROLE_FORMAT: !Sub arn:${AWS::Partition}:iam::{}:role/PantherRemediationRole-${AWS::Region}
       Layers: !If
         - AttachLayers
         - !Split # CFN doesn't have list append, so convert to/from CSV string to prepend base layers
--- a/deployments/cloud_security.yml
+++ b/deployments/cloud_security.yml
@@ -701,6 +701,7 @@ Resources:
       Environment:
         Variables:
           LOGGING_LEVEL: !If [DebugEnabled, DEBUG, INFO]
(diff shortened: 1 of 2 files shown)


deployments/cloud_security.yml
```diff
@@ -701,6 +701,7 @@ Resources:
       Environment:
         Variables:
           LOGGING_LEVEL: !If [DebugEnabled, DEBUG, INFO]
+          ROLE_FORMAT: !Sub arn:${AWS::Partition}:iam::{}:role/PantherRemediationRole-${AWS::Region}
```
austinbyers: `iam` is actually not partitioned, just `arn:aws:iam` would suffice

It's a little confusing to me to have the `{}` from the Python format string here in the CloudFormation, because it's not obvious what it's replaced with and it looks like it should be part of the `! Sub`

Let's change this to

```
REMEDIATION_ROLE_NAME: !Sub PantherRemediationRole-${AWS::Region}
```

austinbyers:
Oh, and actually I think we want the region from the client, not the one in the cloudformation

austinbyers:
In `deployments/auxiliary/cloudformation/panther-remediations-iam.yml`, the template still defines a remediation role without a suffix:

```yaml
AWSTemplateFormatVersion: 2010-09-09
Transform: AWS::Serverless-2016-10-31
Description: Deploys the IAM role that allows Panther to remediate resources in an account
Metadata:
  Version: 1.0

Parameters:
  MasterAccountId:
    Type: String
    Description: The account ID where Panther is deployed in

Resources:
  RemediationRole:
    Type: AWS::IAM::Role
    Properties:
      RoleName: PantherRemediationRole
```

Is this a leftover template that can be deleted @kostaspap ?

austinbyers:
I deployed and was able to click the "remediate" button and re-encrypt some buckets

nhakmiller:
> Oh, and actually I think we want the region from the client, not the one in the cloudformation

This is incorrect, we want the region from CloudFormation.

PR #560, opened 2020-03-25
Fix deployment role

austinbyers:
## Background

We haven't updated the deployment role in awhile, and there are some new resources involved in deploy and teardown that we need to add permissions for.

## Changes

- Add support for GuardDuty, CloudFormation stack sets, VPC flow logs, and IAM server certificates in the deployment role
- Troubleshooting documentation
    - The error mentioned in the deploy process was actually fixed by #556 
    - I mistyped an email in the deploy process, adding some clarification for how to remove the user from Cognito. They can re-add the user in Cognito, as originally suggested, but that won't add custom attributes (name), and in enterprise it won't work at all (they need a role). So it's easier to just people to redeploy in this case, it only takes about a minute now.

## Testing

- Deploy and teardown using the deployment policy

### DIFF
--- a/docs/gitbook/troubleshooting.md
+++ b/docs/gitbook/troubleshooting.md
@@ -6,20 +6,14 @@ This section acts as a quick Q&A on issues that you may face during your Panther
 
 Don't worry. You can safely re-deploy Panther by running `mage deploy` and it will pick up where it left off.
 
-Alternatively, if you 're using temporary credentials, please add a longer timeout.
-
-## Deployment Failed
-
-`open deployments/bootstrap.yml: no such file or directory`
-
-This is an issue with Docker and the way volumes work.
-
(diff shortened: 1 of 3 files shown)


docs/gitbook/troubleshooting.md
```diff
@@ -6,20 +6,14 @@ This section acts as a quick Q&A on issues that you may face during your Panther
...
-
-To resolve that, please step close the container (Ctrl+D on Mac) and reconnect to it by typing `./dev.sh` so that the container can pick up all the necessary files.
-
-Running `mage deploy` should resolve the issue.
+Alternatively, if you're using temporary credentials, please add a longer timeout.
 
 ## Mistyped Email During `mage deploy`
 
-Don't worry, these things happen!
+Don't worry, these things happen! From the AWS console in the deployed region:
```
jacknagz: Thanks for adding this!

Issue #558, opened 2020-03-24
Page Crash When Remediate Button is Pressed


nhakmiller:
### Describe the bug

When the remediate button is pressed, the page crashes.

### Steps to reproduce

1. Go to policies
2. Click on a policy with auto remediations enabled
3. Scroll down to a failing resource
4. Click on remediate
5. Observe

### Expected behavior

Page does not crash when I click remediate.

### Environment

How are you deploying or using Panther?

- Panther version or commit: 
Version v0.3.0-187-g3a605bd

### Additional context

Network tab of my browser reports:
```
4 requests
1.5 KB transferred
30.9 KB resources
{data: {remediateResource: null},…}
data: {remediateResource: null}
remediateResource: null
errors: [{path: ["remediateResource"], data: null, errorType: "500", errorInfo: null,…}]
0: {path: ["remediateResource"], data: null, errorType: "500", errorInfo: null,…}
path: ["remediateResource"]
0: "remediateResource"
data: null
errorType: "500"
errorInfo: null
locations: [{line: 2, column: 3, sourceName: null}]
0: {line: 2, column: 3, sourceName: null}
line: 2
column: 3
sourceName: null
message: "A custom error was thrown from a mapping template."
```

### Screenshots

<img width="1535" alt="Screen Shot 2020-03-24 at 4 24 08 PM" src="https://user-images.githubusercontent.com/49166439/77486259-79b78080-6dec-11ea-8b77-a3d82b15b9c6.png">

PR #562, opened 2020-03-25
Remediation role correct region

nhakmiller:
## Background

The remediation role suffix was being set incorrectly.

## Changes

- Set suffix correctly

## Testing

- Deployed in dev account


nhakmiller:
Closes #558.

PR #564, opened 2020-03-25
Fixing stability issues

kostaspap:
## Background

While testing log alerting I encountered two issues: 
- If the rule was deleted before the alert was propagated, the alert-log-forwarder would keep retrying to fetch the rule .
- Rule's engine was not handling properly the case were a user had written an invalid dedup function. 
- Data Catalog updater was not updating rules matches properly

## Changes

- Fixed above issues. 
- Improved logging in some cases
- Moved onboarding configuration for S3 Server Access logs to be in the same section as CT and GD onboarding since this seems more appropriate. It is not about "monitoring" of the app (in the sense that it monitors the app stability) it is more about log onboarding/auditing of accesses to panther resources&the account where panther is deployed. 
- Fixed data catalog updater

## Testing

- Unit tests
- Deployed and verified normal behavior in my account
- Did a teardown. Set `EnableS3AccessLogs` to `false`. Deployed fresh and seen that audit logs where not enabled. 
- Did a teardown. Set `EnableS3AccessLogs` to `true`. Deployed fresh and seen that audit logs where  enabled.

### DIFF
--- a/internal/log_analysis/alert_forwarder/forwarder/forwarder.go
+++ b/internal/log_analysis/alert_forwarder/forwarder/forwarder.go
@@ -63,6 +63,9 @@ func SendAlert(event *AlertDedupEvent) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to get alert information")
 	}
+	if alert == nil {
+		return nil
+	}
 	msgBody, err := jsoniter.MarshalToString(alert)
 	if err != nil {
 		return errors.Wrap(err, "failed to marshal alert notification")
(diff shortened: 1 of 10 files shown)


internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -92,6 +95,10 @@ func getAlert(alert *AlertDedupEvent) (*alertModel.Alert, error) {
 	})
 
 	if err != nil {
+		if err.Error() == (&policiesoperations.GetRuleNotFound{}).Error() {
```
austinbyers: Better to just check the type of the error:

```go
if _, ok := err.(*policiesoperations.GetRuleNotFound); ok {
    return nil, nil
}
```

next time

PR #567, opened 2020-03-25
Add fluentd parsers to UI and run fmt

rleighton:
## Changes
- Add fluentd parsers to UI
- mage fmt

## Testing
mage test:ci
deployed to dev account, saw tables in Athena


rleighton:
> This looks good. Shouldn't we add it in the docs as well?

He is working on that in a separate PR now.

Issue #356, opened 2020-03-06
Formal Terraform support for aux templates


austinbyers:
### Describe the ideal solution

- All CloudFormation templates in `deployments/auxiliary` have a Terraform equivalent
- `mage fmt` adds license headers to `.tf` files
- `mage fmt` runs `terraform fmt` on `.tf` files
- `mage test:ci` runs `terraform validate` on `.tf` files
- Use `jsonencode` instead of raw JSON in Terraform source

### Describe your use cases

See #337 

### How frequently would you use such feature

Every new user that has a Terraform workflow will need these templates. We need the formatting and testing for every change to the files.

PR #575, opened 2020-03-25
Upgrade go-swagger

austinbyers:
## Background

As I'm getting ready to lightly refactor the `mage fmt` and `mage test` commands (for #356 and #457), I noticed the `go-swagger` library updated, which fixed some of the problems we were seeing before.

The result is slightly simpler and faster building and formatting

## Changes

- Upgrade `go-swagger` to `0.23.0`
- Update all go modules
- Use `swagger generate` to embed the license header instead of doing it ourselves. This saves us a bit of time formatting all the autogenerated files (go formatting now ignores `api/gateway` completely)
- `build:api` no longer has to change directories, which also means bootstrap can run concurrently with building (essentially reverting #556)

## Testing

- `mage build:api`
- `mage fmt` shows no additional changes to swagger files


kostaspap:
Nice!

PR #588, opened 2020-03-26
Autogenerate documentation for log parsers 

rleighton:
Autogenerate documentation for log parsers (including schema and column descriptions)

## Background
Log parser writers often forget to add the documentation. Also, when schemas change it may not be updated.  This addresses the problem by generating the documentation from the source code. Schema is included.

## Changes
* Add mage target doc:logs
* Add mage target doc:all
* Update mage target fmt to generate all documentation
* Update SUMMARY to point to generated file
* Refine PR process in writing parsers doc

## Testing
mage fmt


jacknagz:
Is there also a way to show the required fields? It would help a lot for writing rules

rleighton:
> Hey @rleighton this is awesome!
> 
> Can we retain the separate page structures per major log type? I'd also like to keep the pattern we had before with subheaders for each type therein.
> 
> # Osquery Logs
> ## Osquery.Differential
> ## Osquery.Snapshot
> ...

It does have that pattern (but in 1 page). Hard to read as HTML.

I can do multiple pages IF I can figure out how to get the SUMMARY to from a single reference dir. I can't update the SUMMARY.md file. If I do 
```
[Supported Logs](log-analysis/log-processing/supported-logs/)
```

Will gitbook find and index all the files in the dir?

rleighton:
> Is there also a way to show the required fields? It would help a lot for writing rules

Good idea. I think so, yes.

rleighton:
> > Is there also a way to show the required fields? It would help a lot for writing rules
> 
> Good idea. I think so, yes.

This is done

docs/gitbook/log-analysis/log-processing/writing-parsers.md
```diff
@@ -38,12 +38,20 @@ Be sure to populate the fields.
...
 
-Before making a pull-request:
+### Before making a pull-request
 
 * Write [unit tests](https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_test.go) for your parser.
-* Deploy Panther. You should be able to see a new table with your added parser in Glue Data Catalog! ![Log List from Glue Catalog](../../.gitbook/assets/glue-catalog.png)
-* Use [s3queue](../operations/ops-home.md#tools) to copy test files into the `panther-bootstrap-auditlogs-<id>` bucket to drive log processing. Query Athena to confirm your data is available.
-* Update the [documentation](./supported-logs) with your new parser.
+* Ensure your code is formatted, run `mage fmt`
+* Ensure all tests run `mage test:ci`
```
austinbyers: ensure all tests *pass*

tools/mage/doc_namespace.go
```diff
@@ -21,17 +21,29 @@ package mage
...
+	"github.com/panther-labs/panther/tools/cfngen/gluecf"
 )
 
 // targets generating documentation
 type Doc mg.Namespace
 
+// Generate all documentation
+func (d Doc) All() {
+	d.Cfn()
+	d.Logs()
```
austinbyers: I don't see any reason people would run one command and not the other, unless it takes a very long time?

This is now 3 targets: `doc:all, doc:cfn, doc:logs`, I would just collapse them into a single `mage doc` command, which autogenerates docs

austinbyers: Something should only be a separate target if developers or CI would explicitly use it. For the same reason, I'll be getting rid of some of our existing ones (e.g. `test:cover`)

docs/gitbook/log-analysis/log-processing/writing-parsers.md
rleighton: Done

tools/mage/doc_namespace.go
rleighton: Done

rleighton:
> I love autogenerated docs, thanks for adding! We may also want to consider using `godoc` for doc generation at some point instead of rolling it ourselves.
> 
> When you preview the generated docs, the `type` column can be very obtuse:
> 
> `struct<type:string,principalId:string,arn:string,accountId:string,accessKeyId:string,userName:string,sessionContext:struct<attributes:struct<mfaAuthenticated:string,creationDate:string>,sessionIssuer:struct<type:string,principalId:string,arn:string,accountId:string,userName:string>,webIdFederationData:struct<federatedProvider:string,attributes:string>>,invokedBy:string,identityProvider:string>`
> 
> I don't think that's helping anyone. Can we replace it with something formatted nicer or even just a link to the struct in the source?

I changed this to break on the spaces so it is more readable. Not much to be done about this. Complex struct.

rleighton:
I can't really tell what this _final_ result will look like. Can we push this and then do another iteration?

docs/gitbook/log-analysis/log-processing/supported-logs/Fluentd.md
```diff
@@ -0,0 +1,90 @@
+
+<!-- This document is generated by "mage doc:logs". DO NOT EDIT! -->
+# Fluentd
+Required fields are in <b>bold</b>.
```
jacknagz: Let's wrap this with a hint:

{% hint style="info" %}
Required fields are in **bold**
{% endhint %}

docs/gitbook/log-analysis/log-processing/supported-logs/Fluentd.md
```diff
@@ -0,0 +1,90 @@
...
+##Fluentd.Syslog3164
+Fluentd syslog parser for the RFC3164 format (ie. BSD-syslog messages)
+Reference: https://docs.fluentd.org/parser/syslog#rfc3164-log
+<table>
+<tr><th align=center>Column</th><th align=center>Type</th><th align=center>Description</th></tr>
+<tr><td valign=top>pri</td><td>smallint</td><td valign=top>Priority is calculated by (Facility * 8 + Severity). The lower this value, the higher importance of the log message.</td></tr>
```
jacknagz: Can we wrap all field names and types with backticks so they show as `code`?

docs/gitbook/log-analysis/log-processing/supported-logs/Fluentd.md
rleighton: Done!

docs/gitbook/log-analysis/log-processing/supported-logs/Fluentd.md
rleighton: Done!

PR #589, opened 2020-03-26
Clickable Tags on Rule Details page

### LINKED ISSUE #436: Clickable Rule Tags

From the Rule Details page, the Tags should be clickable to a filtered List Rules page:

### Describe your use cases

Pivoting from a single alert -> rule -> related running rules

Daily

None

### References

None

### Additional context

None

alexmylonas:
## Background

As a user, i want to easily switch from a single Rule to a group of rules in the same category

Closes #436 

## Changes

- Changed Tags on Rules details to be clickable and move user to a filtered ListRules page based on clicked tag

## Testing

- Locally

Screenshot 2020-03-26 at 5 02 50 PM


web/src/pages/RuleDetails/RuleDetailsInfo/RuleDetailsInfo.tsx
```diff
@@ -145,10 +145,14 @@ const RuleDetailsInfo: React.FC<ResourceDetailsInfoProps> = ({ rule }) => {
           </Label>
           {rule.tags.length ? (
             rule.tags.map((tag, index) => (
-              <Text size="medium" color="black" key={tag} is="span">
-                {tag}
+              <React.Fragment key={tag}>
+                <Link key={tag} to={`${urls.logAnalysis.rules.list()}?page=1&tags[]=${tag}`}>
+                  <Text size="medium" color="blue300" is="span">
```
alexmylonas: Tried with` <Text is="Link" ..../>` but the result component had a padding can we override? or should we keep this way?

3nvi: It should be 

`<Text is={Link} to="#" color="blue300">...</Text>`

web/src/pages/ListRules/ListRulesActions/ListRulesActions.tsx
```diff
@@ -91,6 +91,9 @@ const ListRulesActions: React.FC = () => {
   const filterKeys = Object.keys(filters) as (keyof ListRulesInput)[];
   const filtersCount = filterKeys.filter(key => !isEmpty(requestParams[key])).length;
 
+  // This should only trigger when users have filters and a re-render occurs
+  if (areFiltersVisible === false && filtersCount > 0) setFiltersVisibility(true);
```
3nvi: This should be already handled by showcasing the number of active filters, through this line:

`Filter Options {filtersCount ? `(${filtersCount})` : ''}`

But if the requirement says that it **must** be open on-initial-render, then you should modify the initial `useState` hook:

```tsx
  const [areFiltersVisible, setFiltersVisibility] = React.useState(false);
```

and have it not start with a `false` value, but with a dynamically calculated boolean value, depending on whether there are filters active or not

alexmylonas: It's not an AC but since most filters return many results it wasn't clear to me that they were filtered because i didn't actively created them and I was just presented with a page.

Not strong opinion about this what do you think?

3nvi: Me neither. Go ahead and do it :)

P.S. Make  sure you update the initial React state like I suggested

3nvi:
@alexmylonas Can we prioritize closing this?

web/src/pages/ListRules/ListRulesActions/ListRulesActions.tsx
alexmylonas: No need to change initial state to `true`, we just need to changed to `true` if there are filters IMO

web/src/pages/RuleDetails/RuleDetailsInfo/RuleDetailsInfo.tsx
alexmylonas: nvm my comment for padding a `p` prop was passed by mistake

alexmylonas:
@3nvi Latest changes applied!

web/src/pages/ListRules/ListRulesActions/ListRulesActions.tsx
3nvi:
```jsx
  if (areFiltersVisible === false && filtersCount > 0) setFiltersVisibility(true);
```


This is a side-effect and should be defined in a `useEffect` with the proper arguments. Also the check `areFiltersVisible === false` is not needed, since React won't update the state from `true` to `true` even if you tell it to. It's smart enough to understand that.

web/src/pages/RuleDetails/RuleDetailsInfo/RuleDetailsInfo.tsx
3nvi: 👍

web/src/pages/ListRules/ListRulesActions/ListRulesActions.tsx
3nvi: @alexmylonas

PR #611, opened 2020-03-30
Fix incorrect endpoint in log processing SNS setup doc

rleighton:
Fix incorrect endpoint in log processing SNS setup doc reported by customer.


rleighton:
> Should we consider updating the v1 docs with this change as well? Not sure if we can still do that at this point, but it seems a shame to have incorrect docs like this in the v1 tagged docs. I suppose if we follow a two week release cycle then it should be updated soon.

ideally we would update v1 docs. That said, folks with issues will reach out and it will be corrected soon. I could go either way.

nhakmiller:
In this case since it's a simple typo I updated by pushing to the `v1-docs` branch directly with just this change manually copied in. I realize now I should have created a branch of that branch, then put up a PR for that, which I can perhaps do in the future if needed. Hopefully these occasions should be rare though.

Issue #609, opened 2020-03-30
LogSubscriptions


rleighton:
### Description
Enable outside sources to subscribe via S3 event notifications for data in the log processing bucket.

### Acceptance Criteria
* Update panther_config.yml to allow a list of principals that can read from the bucket and subscribe to SNS from log processor.

austinbyers:
@rleighton I'm closing this because it seems #612 satisfies the acceptance criteria. Feel free to re-open if it's incomplete with more context about what still needs to happen.

PR #612, opened 2020-03-30
Log subscriptions

rleighton:
## Background

This change allows users to configure principals in panther_config.yml to subscribe to events on the log processing bucket and subscribe to the sns topic. Useful for people to integrate Panther data into down-stream systems.

See: https://github.com/panther-labs/panther/issues/609

## Changes
* Make internal notification message format compatible with s3 event notifications
* Add section to setup in panther_config.yml to allow configuration.

## Testing
mage test:ci
ran in dev env
configured 3rd party app to consume data


internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -135,6 +128,31 @@ func Handle(lc *lambdacontext.LambdaContext, batch *events.SQSEvent) (err error)
 	return err
 }
 
+func handleLogProcessorCloudTrail(messageBody string, changes map[string]*resourceChange) (ok bool, err error) {
+	notification := &models.S3Notification{}
+	if err := jsoniter.UnmarshalFromString(messageBody, notification); err != nil {
+		return false, errors.Wrap(err, "failed to unmarshal record")
+	}
```
nhakmiller: In this function we're not marshaling/unmarshaling. Can we maintain the `gjson.Get` pattern from above?

internal/log_analysis/rules_engine/src/output.py
```diff
@@ -146,26 +146,55 @@ def _write_to_s3(time: datetime, key: OutputGroupingKey, events: List[EventMatch
...
             'id': {
                 'DataType': 'String',
-                'StringValue': notification.id
+                'StringValue': key.rule_id
             }
         }
     )
 
 
+# pylint: disable=invalid-name
```
nhakmiller: Is there a reason this function has to be named this way instead of in a way that pylint won't complain about? Personally I try to only override the pylint checks if there is a specific reason we are goin against python standards.

internal/compliance/aws_event_processor/processor/handler.go
austinbyers: Right `gjson` is much faster than unmarshaling everything

internal/log_analysis/rules_engine/src/output.py
austinbyers: Agreed, we have control over the name here so let's just follow pylint guidelines

deployments/log_subscription.yml
```diff
@@ -0,0 +1,78 @@
...
+    Description: The list of Principal ARNs to allow read access to the ProcessedDataBucket and subscribe access to ProcessedDataTopicArn
+
+Conditions:
+  ExistsPrincipals: !Not [!Equals [!Select [0, !Ref LogSubscriptionPrincipals], '']]
+
+Resources:
+  # Read access on processed data bucket
+  ProcessedDataBucketPolicy:
+    Condition: ExistsPrincipals
+    Type: AWS::S3::BucketPolicy
```
austinbyers: All buckets and policies are currently in the bootstrap stack, can we just move these there?

We don't need a separate stack for just these two separate resources in my opinion. The parameters you need are already in the bootstrap stack, easier (and faster) to just put them in the same stack, we have plenty of room before we hit the size limit

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -60,6 +61,11 @@ const (
 
 	bytesPerMB           = 1024 * 1024
 	maxS3BufferSizeBytes = 50 * bytesPerMB // the largest we let any single buffer get
+
+	// for the sns s3 events
```
austinbyers: Are these constants really necessary? They'll be shared across the entire `destinations` package and don't appear anywhere else

I'd just hardcode them in your event below

tools/mage/deploy.go
```diff
@@ -416,6 +418,13 @@ func deployMainStacks(awsSession *session.Session, settings *config.PantherConfi
 		result <- frontendStack
 	}(finishedStacks)
 
+	// Log subscriptions
```
austinbyers: Again, much simpler to just put your two resources in the bootstrap stack since they are directly related to them (s3 policy and SNS policy)

api/lambda/core/log_analysis/log_processor/models/api.go
```diff
@@ -20,19 +22,8 @@ package models
...
-	Events *int `json:"events"`
-	// Bytes is the uncompressed size in bytes of the S3 object
-	Bytes *int `json:"bytes"`
-	// Type is the type of data available in the S3 object (LogData,RuleOutput)
-	Type *string `json:"type" validate:"required"`
-	// ID is an identified for the data in the S3 object. In case of LogData this will be
-	// the Log Type, in case of RuleOutput data this will be the RuleID
-	ID *string `json:"id" validate:"required"`
+	// https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html
+	Records []events.S3EventRecord
```
kostaspap: We are actually relying on the other fields. Both ID and Type fields are used. For example: 
1. RulesEngine only listens to new Log notifications and skipping notifications for rules output : https://github.com/panther-labs/panther/blob/master/deployments/log_analysis.yml#L502-L504
2. Snapshot Poller only listers to CloudTrail notifications: https://github.com/panther-labs/panther/blob/master/deployments/cloud_security.yml#L306-L318

I don't like it so much that we lose the capability to filter notifications on the subscription level. I also don't like that we lose the flexibility of adding our own fields to notifications - that was one of the main reasons we wanted to have custom notifications instead of using S3 Event notifications. 

I am thinking that a more appropriate solution would be just to have a Lambda that does translation of our notifications to S3 for wherever this is needed. I don't like that we compromise our design and opt for a less flexible one.

internal/compliance/aws_event_processor/processor/handler.go
kostaspap: I think the benefit of taking this approach vs to using `gjson.Get` is that we get compile-time guarantees for the S3Notification message. Without such, it is hard to identify all the parts of the code that are consuming the Log Processor notifications.

rleighton: Exactly what Kostas said. If we change anything then all places we use it change, tests will break,, etc. In fact, if Kostas did not remind me of the event processor, I would have missed it. It is important to make this dependency clear.

internal/log_analysis/rules_engine/src/output.py
rleighton: Good catch, that was left over. Will remove.

deployments/log_subscription.yml
rleighton: Sure. I did it as a separate stack when I was uncertain if this was "enterprise", in which case having another stack that "layers on" was a nice solution.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: I need them for the tests. That said, it would be cleaner to make a constructer to encapsulate these.

api/lambda/core/log_analysis/log_processor/models/api.go
rleighton: Actually, this PR did not change the SNS envelope at all , the filtering still works.  

We also do not lose the capability to add our own fields, we just add them in the SNS envelope.

I don't think we lose any flexibility for the use cases I can imagine (adding more meta data,  like tags for counts).

Thoughts?

kostaspap: I am not sure I understand... There is no SNS envelope, we send just one message to SNS, this  message is `S3Notification` that contains the fields `ID` and `Type` which no longer exist. The filtering is performed based on these fields and I don't see these fields being added anywhere else.... 
If we want the notifications though, why not just add notifications to the bucket itself - and create a new SNS topic where these notifications will be forwarded?

rleighton: No, there is an SNS envelope and _that_ is what is used for filtering: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/destinations/s3.go#L290 . There is similar code in the log processor. The "S3 notification" is just a blob.

The reason not to add directly to the bucket is there are a bunch of restrictions on how you can configure notifications. I was going to do that originally, but I worried about getting stuck. This way (because we have logType and ID in the SNS message) we can control who gets what data using filtering.

internal/log_analysis/log_processor/destinations/s3.go
kostaspap: Actually probably these constants should probably be in the same file as `S3Notification`. 
A maybe cleaner way would be to just have a builder pattern for the S3Notification - and that will make sure to set these constants when building the notification. That avoid the problem when a producer of the message forgets to set the properly.

api/lambda/core/log_analysis/log_processor/models/api.go
rleighton: Also, we talked about allowing users to add their own "tags" when onboarding a log source. The would be added to records with something like  "p_customer_tags"  When we do that, we should add those to the SNS message attributes so consumers can see these and use them for routing.

internal/log_analysis/rules_engine/src/output.py
rleighton: done

deployments/log_subscription.yml
rleighton: done

internal/log_analysis/log_processor/destinations/s3.go
rleighton: I made it a method but I'll do what you suggest (and have the method call it).

tools/mage/deploy.go
rleighton: done

internal/log_analysis/log_processor/destinations/s3.go
rleighton: done

api/lambda/core/log_analysis/log_processor/models/api.go
```diff
@@ -20,19 +22,35 @@ package models
...
+		eventName    = "ObjectCreated:Put"
+	)
+	return &S3Notification{
+		Records: []events.S3EventRecord{
+			{
+				EventVersion: eventVersion,
+				EventSource:  eventSource,
+				EventName:    eventName,
+				S3: events.S3Entity{
+					ConfigurationID: logType,
```
kostaspap: ConfigurationID is the ID of the S3Event notification configured for an S3 bucket. We can probably just leave it out...
I wouldn't like that our code starts to depend on this being the log type (or ruleID). We will never/shouldn't ever send in a single notification notifications for different types of logs. This would hinter our ability to do appropriate filtering on SNS subscription level. 

Any listeners of these notifications can always get this info from the MessageAttributes (which actually in case of SNS->SQS get merged in the actual SQS message).

deployments/bootstrap.yml
```diff
@@ -299,6 +303,37 @@ Resources:
...
+            Action: s3:ListBucket
+            Resource: !Sub arn:aws:s3:::${ProcessedData}
+          - Sid: ReadLogData
+            Effect: Allow
+            Principal:
+              AWS: !Ref LogSubscriptionPrincipals
+            Action:
+              - s3:GetObject
+              - s3:GetObjectVersion
+            Resource: !Sub arn:aws:s3:::${ProcessedData}/logs/*
```
kostaspap: Shouuld be just `!Sub arn:aws:s3:::${ProcessedData}/*` I think. We already write to `/rules/` and soon might write in different paths as well.

deployments/bootstrap.yml
```diff
@@ -726,6 +761,22 @@ Resources:
...
+    Properties:
+      Topics:
+        - !Ref ProcessedDataNotifications
+      PolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Sid: ReadTopic
+            Effect: Allow
+            Principal:
+              AWS: !Ref LogSubscriptionPrincipals
```
kostaspap: This has to be a `root` principal anything else wouldn't work for e.g. SNS->SQS subscription cross-account. This has to be clear in the documentation. But then again, why not just require in the configuration users to add the account IDs and let us generate the appropriate `root` role ARNs? That would minimize the chance for misconfiguration from customer side.

internal/log_analysis/rules_engine/src/main.py
```diff
@@ -127,8 +118,32 @@ def log_analysis(event: Dict[str, Any]) -> None:
...
+
+# Reads S3 notifications and return tuples of (bucket, key, configurationId)
+def _load_s3_notifications(records: List[Dict[str, Any]]) -> List[Tuple[str, str, str]]:
+
+    events: List[Tuple[str, str, str]] = []
+    for s3event in records:
+        # https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html
+        bucket = s3event['s3']['bucket']['name']
+        object_key = s3event['s3']['object']['key']
+        configuration_id = s3event['s3']['configurationId']
```
kostaspap: Same comment I did above. We shouldn't rely on this field but instead parse it from the SNS envelope. The field is already there (named "Id").  
This information is already in the SNS Envelope so I'd prefer to take it from there instead of having it in two places.

deployments/panther_config.yml
```diff
@@ -94,6 +94,14 @@ Setup:
   # You may want this off if you have org-level GD configured for Panther.
   EnableGuardDuty: false
 
+  # Grant external access to the Panther processed data.
+  LogSubscriptions:
+    # A list of ARNs of Principals that want to subscribe to log data.
+    # These ARNs will be granted read permission on the processed log bucket and
+    # SNS subscribe permission to the panther-processed-data-notifications topic.
+    PrincipalARNs:
+      -
```
kostaspap: nitpicks: Why not flatten it? e.g. SubscriptionPrincipalARNs? Are we going to have more fields here?

api/lambda/core/log_analysis/log_processor/models/api.go
kostaspap: Ah, true, the MessageAttributes. That's true,!we can use these attributes to pass custom information. Also in case e.g. of SNS->SQS subscription, the MessageAttributes get merged with the original SQS message. 
Sounds good!

deployments/bootstrap.yml
rleighton: Third parties have custom ARNs. This is what is needed for the SNS topic for snowflake (a particular IAM user): arn:aws:iam::343182919852:user/k8m1-s-v2st0722 .

internal/log_analysis/rules_engine/src/main.py
rleighton: Agree. I was leaving it the way it was to minimize changes , but I'll do that.

deployments/panther_config.yml
rleighton: Initially I thought so.  Turns out I did not. This at least future proofs a name space. I don't feel strongly either way right now.

deployments/bootstrap.yml
rleighton: Initially I did, but keeping with the principle of least privilege I scoped it. It may be that we want to have data in the bucket NOT available to others. At least this way we can control that going fwd.

api/lambda/core/log_analysis/log_processor/models/api.go
rleighton: Sure

rleighton: Ok, it turns out for JSON style messages you can't get the attributes. You can use them for filtering but they won't be passed down. Yuck.

So we have to leave this. It we need extra attributes we can add them but if 3rd parties verify schemas it could break them. Nobody does this, so I suspect would be fine.

internal/log_analysis/rules_engine/src/main.py
rleighton: See above. Can't change.

internal/compliance/aws_event_processor/processor/handler.go
```diff
@@ -83,17 +84,9 @@ func Handle(lc *lambdacontext.LambdaContext, batch *events.SQSEvent) (err error)
...
-			zap.L().Debug("processing s3 notification")
-			object := &sources.S3ObjectInfo{
-				S3Bucket:    results[0].Str,
-				S3ObjectKey: results[1].Str,
-			}
-			err := handleS3Download(object, changes)
-			if err != nil {
-				return err
-			}
+		if isLogProcessorCloudTrail, err := handleLogProcessorCloudTrail(record.Body, changes); err != nil {
```
rleighton: This looks wrong to me (wrong part of message). Let me check...

deployments/bootstrap.yml
kostaspap: Interesting! Fair enough

internal/log_analysis/rules_engine/src/main.py
rleighton: Nevermind. This will work.

api/lambda/core/log_analysis/log_processor/models/api.go
rleighton: Nevermind. This will work. Fixing...

internal/log_analysis/rules_engine/src/main.py
rleighton: done

api/lambda/core/log_analysis/log_processor/models/api.go
rleighton: done

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: Sure, the benefit is compile-time guarantees. But the cost is time. We did the benchmarking, and determined it was worth the tradeoff of compile time guarantees for speed. This change forces every single SNS wrapped CloudTrail event to first be unsuccessfully unmarshalled into this S3 notification struct. The entire rest of this lambda function very specifically avoids unmarshalling for this exact reason.

The reason we have tests is so they'll break when things change, the reasons we have code reviews is so people can remind each other of things. I don't think these are valid reasons to not follow this pattern here.

internal/log_analysis/rules_engine/src/output.py
```diff
@@ -146,26 +146,54 @@ def _write_to_s3(time: datetime, key: OutputGroupingKey, events: List[EventMatch
...
+    """
+    return {
+        'Records':
+            [
+                {
+                    'eventVersion': '2.0',
+                    'eventSource': 'aws:s3',
+                    'eventName': 'ObjectCreated:Put',
+                    's3': {
+                        'configurationId': rule_id,
```
kostaspap: I think you can remove it this line - since it is not used in Golang neither.

rleighton: done

deployments/log_subscription.yml
austinbyers: It's still a separate stack, did you forget to push a change?

internal/compliance/aws_event_processor/processor/handler.go
austinbyers: json unmarshaling is relatively slow because it has to use reflection to understand the struct you're trying to unmarshal into.

If you have data that can be of different types, `gjson` can very quickly check individual fields to distinguish them. Here's the type of benchmark Nick is referring to - using the sample event from one of the tests, I construct a message which is wrapped and would not have records when parsed:

```go
var sampleSNSEvent = `{"arn": "my-topic-arn", "sender": "abcdefg", "message": ` + sampleS3Event + "}"

func BenchmarkJson(b *testing.B) {
	for i := 0; i <= b.N; i++ {
		var result models.S3Notification
		if err := jsoniter.UnmarshalFromString(sampleSNSEvent, &result); err != nil {
			panic(err)
		}
		if len(result.Records) > 0 {
			panic("found records unexpectedly")
		}
	}
}

func BenchmarkGson(b *testing.B) {
	for i := 0; i <= b.N; i++ {
		records := gjson.Get(sampleSNSEvent, "Records")
		if records.Exists() && records.IsArray() {
			panic("found records unexpectedly")
		}
	}
}
```

```
BenchmarkJson-12    	  397207	      2836 ns/op
BenchmarkGson-12    	  973148	      1226 ns/op
```

As you can see, `gjson` can discard the event twice as quickly. The actual performance difference may be negligible and we may not care, but it does need to be a conscious decision why we are suddenly discarding the more performant method

austinbyers: I'm fine with trading performance for clarity if that's what we decide, I just wanted to give more context about why we were doing it the way that we are

rleighton: Yes, clarity and safety is preferred here.

deployments/log_subscription.yml
rleighton: Hmmm...

rleighton: It is gone. Look at the file list. Comment says "Outdated"

internal/compliance/aws_event_processor/processor/handler.go
nhakmiller: At the very least can we use gjson to determine whether or not the event is an s3Notification, then if you really want to unmarshal it afterwards? That way at least the change is not hurting the performance of any other use cases for this one.

deployments/log_subscription.yml
austinbyers: Huh, I thought I refreshed before reviewing, but clearly not. LGTM

rleighton:
> Would like to see a resolution to this thread before merging: [#612 (comment)](https://github.com/panther-labs/panther/pull/612#discussion_r400536716)

Actually gjson was not necessary. I changed to usedSNS message attributes to dispatch. I would check the other types of messages to see if they have reliable SNS attributes that might avoid doing json parsing for dispatch.

Issue #605, opened 2020-03-30
Panther failing to build


kostaspap:
### Describe the bug

Panther build fails
```
10:44:30        INFO    fmt: license headers
10:44:30        INFO    fmt: gofmt api/lambda internal pkg tools cmd magefile.go
10:44:33        INFO    fmt: python yapf internal/compliance/remediation_aws internal/compliance/policy_engine internal/log_analysis/rules_engine
10:44:33        INFO    fmt: prettier
10:44:42        INFO    doc: generating operational documentation from cloudformation
10:44:42        INFO    doc: generating documentation on supported logs
10:44:42        INFO    build:api: generating Go SDK for 4 APIs (api/gateway/*/api.yml)
10:44:46        INFO    build:lambda: compiling 20 Go Lambda functions (internal/.../main) using go1.14.1
# github.com/panther-labs/panther/api/gateway/analysis/client
api/gateway/analysis/client/panther_analysis_client.go:73:19: undefined: operations
api/gateway/analysis/client/panther_analysis_client.go:118:13: undefined: operations
# github.com/panther-labs/panther/api/gateway/compliance/client
api/gateway/compliance/client/panther_compliance_client.go:73:19: undefined: operations
api/gateway/compliance/client/panther_compliance_client.go:118:13: undefined: operations
# github.com/panther-labs/panther/api/gateway/remediation/client
api/gateway/remediation/client/panther_remediation_client.go:73:19: undefined: operations
api/gateway/remediation/client/panther_remediation_client.go:118:13: undefined: operations
10:44:46        FATAL   go build out/bin/internal/compliance/alert_processor/main/main failed: running "go build -ldflags -s -w -o out/bin/internal/compliance/alert_processor/main ./internal/compliance/alert_processor/main" failed with exit code 2

```

### Steps to reproduce

1. Checkout panther
2. `mage setup:all test:ci`

### Expected behavior

Panther should build

### Environment

How are you deploying or using Panther?

- Panther version or commit: 1c738d9fa723c9efc543e04d1f10307fde38b0b3
- OS: MacOS 10.15.4
- Browser: Not relevant

### Additional context

* It seems that `mage build:api` is modifying the swagger files and this leads to build failures. 
* The problem seems to be due to some inconsistent behavior of `go-swagger`. I see the same problem when running directly swagger installed in the `.setup` folder:
```
$ .setup/swagger version
version: v0.23.0
commit: 3596b40bedd4f06cc918767098568456d2ec2370
[...]

PR #614, opened 2020-03-31
Build swagger in CI

austinbyers:
## Background

Add `build:api` to CI (trying to reproduce #605 )

## Changes

- Add `build:api` to CI (it used to be part of `test:go`, when we moved it out we didn't update CI)

## Testing

- CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,9 +49,6 @@ jobs:
     steps:
       - setup_backend
       - run:
-          name: Touch Swagger SDKs to prevent regeneration
-          command: touch api/gateway/*/{client,models}
-      - run:
           name: Lint and test the backend infrastructure
           command: mage build:api test:cfn test:go test:python
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ jobs:
       - setup_backend
       - run:
           name: Lint and test the backend infrastructure
-          command: mage build:api bulid:cfn test:cfn test:go test:python
+          command: mage build:api build:cfn test:cfn test:go test:python
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ jobs:
       - setup_backend
       - run:
           name: Lint and test the backend infrastructure
-          command: mage build:api test:cfn test:go test:python
+          command: mage build:api bulid:cfn test:cfn test:go test:python
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ jobs:
           command: touch api/gateway/*/{client,models}
       - run:
           name: Lint and test the backend infrastructure
-          command: mage test:cfn test:go test:python
+          command: mage build:api test:cfn test:go test:python
 
 workflows:
   version: 2


.circleci/config.yml
```diff
@@ -48,12 +48,9 @@ jobs:
     resource_class: xlarge
     steps:
       - setup_backend
-      - run:
-          name: Touch Swagger SDKs to prevent regeneration
-          command: touch api/gateway/*/{client,models}
       - run:
           name: Lint and test the backend infrastructure
-          command: mage test:cfn test:go test:python
+          command: mage build:api test:cfn test:go test:python
```
kostaspap: We are missing the `build:opstools`, `build:devtools`, `build:cfn`. Any of these could be failing and CI would never report it - bad commits can make it through. 
Why not just `mage test:ci`? It is hard to otherwise keep track of changes and we might miss something.

austinbyers: `build:cfn` is simple enough I can go ahead and add it now

I don't include `build:lambda`, `build:opstools` and `build:devtools` because they take a long time and `go test` will already fail if any of the Go source can't compile. So it significantly slows down CI for little value. If they are broken into different tasks, the slowdown wouldn't be so problematic

We don't run `mage test:ci` for the same reason - it takes a long time and the idea was to break apart the tests into different categories to get feedback to the developer more quickly. Right now, we have frontend and backend tests separated, but I'd like to break them apart even more.

So I see two options:

1. Continue breaking apart the test tasks based on the setup required. E.g. `test:cfn, test:go, test:python, test:tf, test:web` might all be separate test tasks that each install only the libraries it needs. This gets feedback to the developer quickly and individual jobs can be restarted if something goes wrong.
1. Consolidate everything and just run `mage test:ci`, eventually spend some time making this run faster (parallelize when possible, for example). This avoids the problem where we forget to update the CI jobs relative to the test suite, but at the cost of slower builds and less timely feedback for developers

Which would you prefer?

austinbyers: Either way, for this PR I'll just add build:api and build:swagger

austinbyers:
> Dunno if we gonna keep it. I'm cool eitherway

Right, we may do away with swagger this quarter if we can, but we still need to test it until then

.circleci/config.yml
rleighton: I would prefer the safer , more complete option to avoid pushing bad code.

kostaspap: TLDR:
I vote for a minimum number of splits (3rd option?) I don't think that it needs to be either 1 task or "many tasks".  There is the middle ground option as well which for me is preferable. For me personally the current split is fast enough, but if people think that this is affecting the productivity, we could create another tasks if it is expected to give a significant speed boost. 

More detailed thoughts: 
From a developer perspective, I want to have a simple command that verifies that my code is working as expected:
* Makes sure that the code in the project compiles. Including generated code. 
* Make sure the unit tests pass
* Make sure my code matches team standards (linting)
We should make it dead simple for developers to run the above when needed. Ideally that would be one command(or "target"). As developer I don't always know all the different ways that the code is split ("what is opstools? what is devtools?Is it only test:cfn and test:tf or did we add another project one in the latest commit?"), but I still want to make sure my change didn't break anything (e.g. renaming a mistyped struct can have impact e.g .in Terraform generation). 

If running everything starts being slow and impacts productivity, we want to create commands that verify individual subsystems. If running everything was taking few seconds, we wouldn't even care to make this split!!
But I think it's better to make this split gradually as the need raises. Creating multiple commands increases the complexity of the system and makes it harder to verify that it is indeed working properly. E.g. I am completely against having `test:cfn` and `test:tf` targets. It just adds complexity (minor but everything adds up) and doesn't give any tangible productivity benefit as I see it. The different mage targets should form a nice acyclic graph that should be easy for anyone to follow. Including us, maintainers of this project! If we don't understand it clearly, we can't optimize it and we can even miss some validations. 

In general, I think we should be viewing multiple mage targets as a necessary "evil" rather than desired state.

austinbyers: i've found it useful to be able to run `test:web`, `test:go`, and `test:python` separately, but that's just because it's much faster than `test:ci`. I, too, would like to consolidate targets

Thanks for the feedback guys! I'll be refactoring the tests as part of enabling #356 and #457

PR #620, opened 2020-03-31
Add docker docs

### LINKED ISSUE #603: Add swagger to public development image

Currently all users need to run `setup:swagger` before they deploy. We should consider adding it to our development image so that users can simply run `mage deploy` to deploy for the first time.

### Describe your use cases

Users will have to:

- wait less for a deployment since swagger will already be present
- have a single memorable command to remember (i.e. `mage deploy`)
- docs will be simplified

On each update

The current alternative is to run `mage setup:swagger deploy`.

### References

https://github.com/panther-labs/panther/issues/369

### Additional context

If we migrate to the proposed solution, we will have to update the public image each time we bump the swagger dependency. Bumping a version shouldn't be something that happens on a daily basis (more like monthly), so although it's not a big thing, it's something we should think about.

3nvi:
## Background

Early exit on image build. Add docs

Closes #603


build/README.md
```diff
@@ -0,0 +1,123 @@
+# Panther Docker Images
```
kostaspap: These instructions, although very useful, cannot be used by the majority of the users using this repo. These are instructions for only the panther team, so I think it's more appropriate to move this to internal docs.

3nvi: The only limitation to using that is pushing it to the official repo. A use might still want to build & tag an image locally to use for their purposes. For example, users might build their own versions of the `development-pack` should they wish to. These docs help on how to do this.

Lastly, we might want to add some contributors to our dockerhub without them being affiliated to Panther Labs.

build/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Updating Panther Images
+
+To update a Panther image you need to:
+
+- build it locally
+- tag it
+- push the updated tags to the remote
+
+This is noot handled by any automation. This is something that needs to be done manually
```
nhakmiller: *not

build/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Building an image
+
+To produce a local image out of a Dockerfile, make surer that Docker is running and type:
```
nhakmiller: *make sure

build/README.md
austinbyers: It's ok to have things in the repo that are primarily for our own internal use in my opinion. We already do this with the public cloudformation template makefile, for example, and many of the mage commands

PR #621, opened 2020-03-31
Separated Delete Rule Query in schema & AppSync

### LINKED ISSUE #464: Rules should be deleted by a separate query instead of using the deletePolicy query
### Description

Deleting a Rule in Log Analysis is using a deletePolicy query that is defined from `DeletePolicyResolver` in appsync right now. We need to split this to a separate resolver `DeleteRuleResolver` and a separate query for clarity and consistency. 

### Acceptance Criteria

- Add new `DeleteRuleResolver` in `appsync.yaml`
- Add graphl query for deleting rules based on above resolver
- Update `DeleteRuleModal` to use the new query

alexmylonas:
## Background

Deleting a Rule in Log Analysis is using a deletePolicy query that is defined from DeletePolicyResolver in appsync right now. We need to split this to a separate resolver DeleteRuleResolver and a separate query for clarity and consistency.

Closes #464 

## Changes

- Added `DeleteRuleResolver` in Appsync
- Added new graphql queries in web and schema
- Update `DeleteRuleModal` to use new query
## Testing

- Locally
- Dev enviroment

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -173,7 +173,7 @@ input GetRuleInput {
 }
 
 input DeleteRuleInput {
-  policies: [DeleteRuleInputItem]
+  rules: [DeleteRuleInputItem!]!
 }
 
 input DeleteRuleInputItem {
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -18,6 +18,7 @@ type Mutation {
   deleteComplianceIntegration(id: ID!): Boolean
   deleteLogIntegration(id: ID!): Boolean
   deletePolicy(input: DeletePolicyInput!): Boolean
+  deleteRule(input: DeleteRuleInput!): Boolean
   deleteUser(id: ID!): Boolean
   inviteUser(input: InviteUserInput): User!
   remediateResource(input: RemediateResourceInput!): Boolean
@@ -169,6 +170,14 @@ input CreateOrModifyRuleInput {
 input GetRuleInput {
   ruleId: ID!
   versionId: ID
+}
+
+input DeleteRuleInput {
+  policies: [DeleteRuleInputItem]
+}
+
+input DeleteRuleInputItem {
+  id: ID!
 }
 
 type RuleSummary {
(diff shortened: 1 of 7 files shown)


api/graphql/schema.graphql
```diff
@@ -171,6 +172,14 @@ input GetRuleInput {
   versionId: ID
 }
 
+input DeleteRuleInput {
+  policies: [DeleteRuleInputItem]
```
3nvi: let's rename this key to `rules`

deployments/appsync.yml
```diff
@@ -1391,6 +1391,44 @@ Resources:
...
+    Type: AWS::AppSync::Resolver
+    DependsOn: GraphQLSchema
+    Properties:
+      ApiId: !Ref ApiId
+      TypeName: Mutation
+      FieldName: deleteRule
+      DataSourceName: !GetAtt AnalysisAPIHttpDataSource.Name
+      RequestMappingTemplate: |
+        #set ($policiesInput = [])
+        #foreach($item in $ctx.args.input.policies)
```
3nvi: rename the final key to `.rules` if you change it in `schema.graphql`

deployments/appsync.yml
```diff
@@ -1391,6 +1391,44 @@ Resources:
...
+  DeleteRuleResolver:
+    Type: AWS::AppSync::Resolver
+    DependsOn: GraphQLSchema
+    Properties:
+      ApiId: !Ref ApiId
+      TypeName: Mutation
+      FieldName: deleteRule
+      DataSourceName: !GetAtt AnalysisAPIHttpDataSource.Name
+      RequestMappingTemplate: |
+        #set ($policiesInput = [])
```
3nvi: Hmm.. 

It has been at least 8 months since this code has been touched (The Delete Policy one), but I feel that lines 1403-1408 simply recreate what already exists in `$ctx.args.input.policies`

Do  you agree? Am I missing something?

alexmylonas: I think you are correct, seems like it serves the purpose of only getting the `id` for policy item.
Which we already assure in graphql, i 'll test it out along with the other changes

3nvi: Perfect! Thanks

3nvi: @alexmylonas please do (and test) the  same in the original `DeletePolicyResolver`

alexmylonas:
@3nvi Tested with latest changes, everything seems in order!

PR #632, opened 2020-04-01
Updated latency threshold

nhakmiller:
## Background

Observed that some app sync queries (especially on a cold start) were taking longer than 1 second to respond under normal circumstances, but our alarms trigger when more than 5 queries hit 1 second latency in a short period. Increasing the latency required to trigger an alarm to 2 seconds.

## Changes

- Bumped latency threshold for app sync alarms from 1 second to 2 seconds

## Testing

- None

### DIFF
--- a/tools/cfngen/cloudwatchcf/alarms_appsync.go
+++ b/tools/cfngen/cloudwatchcf/alarms_appsync.go
@@ -58,7 +58,7 @@ func generateAppSyncAlarms(resource map[interface{}]interface{}) (alarms []*Alar
 
 	// latency
 	alarms = append(alarms, NewAppSyncAlarm("AppSyncHighLatency", "Latency",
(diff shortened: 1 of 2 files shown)


tools/cfngen/cloudwatchcf/alarms_appsync.go
```diff
@@ -58,7 +58,7 @@ func generateAppSyncAlarms(resource map[interface{}]interface{}) (alarms []*Alar
 
 	// latency
 	alarms = append(alarms, NewAppSyncAlarm("AppSyncHighLatency", "Latency",
-		"is experience high latency", resource).MaxNoUnitsThreshold(1000, 60).EvaluationPeriods(5))
+		"is experience high latency", resource).MaxNoUnitsThreshold(2000, 60).EvaluationPeriods(5))
```
austinbyers, line 61: nit: "is experiencing high latency"

PR #642, opened 2020-04-02
Filtering based on "free" text field passed encode data to API

### LINKED ISSUE #576: Filtering Resources with quotes in the name returns all results
### Describe the bug

When you are on the resource list page, if you type a filter in the filter box that includes quotes, it just returns everything.

### Steps to reproduce

1. Go to resource list page
2. Click Filter Options
3. In ID / Name filter box put something in in quotes
4. Click the Apply button
5. Observe that results are not filtered
6. remove quotes
7. Observe that results are filtered

### Environment

- Panther version v0.3.0-190-g28769b4

### Additional context

See screenshots.

### Screenshots

Screen Shot 2020-03-25 at 5 56 45 PM
Screen Shot 2020-03-25 at 5 56 54 PM

alexmylonas:
## Background

Initial problem from issue #576 : 

> Filtering Resources with quotes in the name returns all result

This PR addresses this issue by url encoding all free text fields for `names` and `ids`
Specifically: 
 - Cloud Security -> Policies, Resources
 - Log Analysis -> Rules

Closes #576
## Changes

- Encode params `idContains` & `nameContains` on `updateRequestParams`

## Testing

- Locally

### DIFF
--- a/web/src/hooks/useRequestParamsWithPagination.tsx
+++ b/web/src/hooks/useRequestParamsWithPagination.tsx
@@ -30,8 +30,8 @@ function useRequestParamsWithPagination<
   const updateRequestParamsAndResetPaging = React.useCallback(
     (newParams: Partial<AvailableParams>) => {
       const params = { ...urlParams, ...newParams, page: 1 };
-      if (params.idContains) params.idContains = encodeURI(newParams.idContains);
-      if (params.nameContains) params.nameContains = encodeURI(newParams.nameContains);
+      if (params.idContains) params.idContains = encodeURIComponent(newParams.idContains);
+      if (params.nameContains) params.nameContains = encodeURIComponent(newParams.nameContains);
       updateUrlParams(params);
     },
     [urlParams]
--- a/web/src/hooks/useRequestParamsWithPagination.tsx
+++ b/web/src/hooks/useRequestParamsWithPagination.tsx
@@ -19,7 +19,9 @@
 import React from 'react';
 import useUrlParams from 'Hooks/useUrlParams';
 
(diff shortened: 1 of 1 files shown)


web/src/hooks/useRequestParamsWithPagination.tsx
```diff
@@ -19,15 +19,20 @@
...
+>() {
   const { urlParams, updateUrlParams } = useUrlParams<Partial<AvailableParams>>();
 
   // This is our typical function that updates the parameters with the addition of resetting the
   // page to `1`. The only time where we don't wanna do that, is when changing pages. In this
   // scenario, we want to change the params but not reset the page.
   const updateRequestParamsAndResetPaging = React.useCallback(
     (newParams: Partial<AvailableParams>) => {
-      updateUrlParams({ ...urlParams, ...newParams, page: 1 });
+      const params = { ...urlParams, ...newParams, page: 1 };
```
3nvi, line 32: I don't think this is the correct place to put it. 

This changes a fully abstract and re-usable pattern to handle a specific requirement. What if the field `idContains` changes name? What if we have another field with `nameContains` where the choices are "fixed' and it's not free text.

This should occur directly on the forms that define a "free search" field by modifying the payload sent to the server, not on the abstraction itself

3nvi: We need an approach similar to this? https://github.com/panther-labs/panther-enterprise/pull/59/files

Perhaps kill two birds with one stone by adding a utility function to do that? It should be field-name agnostic

alexmylonas:
Closing this in favor of #653

PR #644, opened 2020-04-02
Change onboarding check

nhakmiller:
## Background

When upgrading Panther, the self onboarding step assumes that users have not changed the label for the self onboarded cloud security source. In at least one of our internal deployments, this is not the case. This assumption doesn't break anything, but it causes an error to be logged in the backend which sets off alarms and pages people. Updating this check to be more accurate to how the backend performs the same check.

## Changes

- Update the checks performed to determine whether self onboarding is necessary

## Testing

- Deployed in dev account, upgraded, observed error. Made changes, deployed, no error observed.


nhakmiller:
> It's still checking the AWS account ID, so we still have a way of knowing which account it is, it seems

This is actually the only thing the backend checks, which is why it is error'ing. These deploy checks were being more specific in what they considered a duplicate than the backend checks, so the deploy process thought it was not a duplicate but the backend thought it was. Now they are making the exact same checks.

Issue #656, opened 2020-04-03
Fix data races reported by testing with -race


rleighton:
### Describe the bug
@dev reported on the Panther OSS Slack channel that running the test suite with -race flags issues.

The following files have reported issues:
internal/log_analysis/log_processor/destinations/s3_test.go
pkg/awsglue/glue.go

### Steps to reproduce
Add -race in the mage test_namespace.go
```args := []string{"test", "-race", "-vet", "", "-cover", "./..."} add -race flag test_namespace.go 98 ```
 run mage test:go

rleighton:
Fixed

Issue #458, opened 2020-03-18
Dev and deployment tooling


austinbyers:
A collection of mostly small improvements to the developer and deployment flows that we need before v1 or shortly thereafter

PR #657, opened 2020-04-03
Fix data races found by testing with -race

rleighton:
## Background
@dev reported on the Panther OSS Slack channel that running the test suite with -race flags issues.

The following files have reported issues:
internal/log_analysis/log_processor/destinations/s3_test.go
pkg/awsglue/glue.go

See: https://github.com/panther-labs/panther/issues/656

### internal/log_analysis/log_processor/destinations/s3_test.go
This test file had a test that was timing related. The -race made the timing related test not work. 
I change the test to use a signal on a channel rather than sleep and -race was happy.

### pkg/awsglue/glue.go  CreateJSONPartition()
This was a real data race caught, where the function was mutating shared data passed by the caller.
The fix was to copy , then mutate.

### pkg/awsglue/glue.go  SyncPartitions
The first change in this function was to copy data returned from glueClient.GetPartition() before mutation.
The issue here is not the code,  but the test driver which returns the same results for all calls. I  could have changed the test driver, but that was more complex than just doing a copy and this is not a performance sensitive place in the code.

### pkg/awsglue/glue.go  SyncPartitions
The second change to this function was needed due to a share variable used to detect and return errors.
While this sharing works correctly, it drives `-race` batty. I changed to using a channel for errors.

## Testing
mage test:ci (with -race flag)

### DIFF
--- a/internal/log_analysis/log_processor/destinations/s3_test.go
+++ b/internal/log_analysis/log_processor/destinations/s3_test.go
@@ -313,6 +313,7 @@ func TestSendDataIfTimeLimitHasBeenReached(t *testing.T) {
 
 	destination := newS3Destination()
 	eventChannel := make(chan *parsers.PantherLog, 2)
+	doneChannel := make(chan bool, 1)
 
 	const nevents = 7
 	testEvent := newSimpleTestEvent()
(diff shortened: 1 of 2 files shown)


rleighton:
> Thanks for tracking those down!
> 
> You can add `-race` to the go test command now if you want, or I'll do so as part of #458

I think you may want Race() as target like for Cover(). It _really_ slows things down. I think we should run Race() in CircleCI. Play with it.

Issue #615, opened 2020-03-31
Show Alert Title in Alert Tables


3nvi:
### Description

This ticket makes sure to add the `Alert Title` in all the alert tables found throughout the application. Alert Title is a prominent thing and should be added as the first column in all related tables

### Acceptance Criteria

In the RuleDetails page, the alerts should:

- Add alert title to schema (`AlertSummary` and `AlertDetails`)
- Add the alert title as the first column in the table found in RuleDetails page
- Add the alert title as first column in the ListAlerts page
- Show the dedup string in the table found in RuleDetails page

kostaspap:
This requires backend changes - the backend doesn't return the alert title. I would recommend making an epic for this and adding all the necessary backend changes that you require.

kostaspap:
The more I was working on this task, the more I am thinking that we might want to display something slightly different...
### List Alerts Page ###

I am thinking we should display 
1. Alert Title. This is aligned with the initial suggestion in this issue. Be aware that alert title can be potentially a pretty long string so we need to display only the first X characters. I will let UI guys decided how many characters makes sense to display
2. Rule. This is the Rule Display Name if it exists (returned by the Alerts API) or Rule ID otherwise. This is similar to what is displayed in the List Rules (`/log-analysis/rules/`) page. 
3. Severity (thinking might be better to move "earlier" since it is actually more important field IMO than the following ones)
4. Number of events (maybe just rename from existing?)
5. Created At (as is now)
6. Last matched at (as is now)

So, i think we can skip display the "dedupString" in the List Alerts page. This is a string meant to be consumed by the system and is not necessarily intended for "human" consumption.  The Alert Title actually can include all necessary information about an alert that the author of the rule meant to display. 
The following screenshot is from current solution, I don't it is easy to see that the dedup string is indeed not intended for "easy" consumption by the users. 

<img width="1243" alt="Screen Shot 2020-04-06 at 12 28 56 PM" src="https://user-images.githubusercontent.com/2652630/78553378-6ef7e500-7809-11ea-9d8b-0178f6ed89a9.png">


### Rule Details Page ###
I am thinking we should display the following columns:
1. Alert Title. See similar comments above
2. Created At
3. Last Matched At. 

Makes sense @jacknagz @3nvi ?

kostaspap:
@jacknagz What should be the default alert title in case user hasn't specified `title` method in their Python rules? 
What should be the default title if they have specified one - should it be just the title they specified or should we put any prefix/suffix?

[...]

PR #659, opened 2020-04-06
Alerts API can now return Alert Title and Rule display name

### LINKED ISSUE #617: Expose Alert Title from alerts-api

The front-end requires to show the title of an alert in the corresponding Table, but it currently doesn't have a way of fetching it from the API.

This ticket will make sure to expose this information through the Alerts-api

### Describe your use cases

Showcasing `title` in:

- RuleDetails page
- AlertsList page

Daily

Not showing it. This is not considered a solution

kostaspap:
## Background

Closes #617 

* Alerts API can now return the title returned by the Python Rule's `title` method. If the Python rule has no `title` method, the default title is returned `<rule display name/rule id> failed`
* Alerts API can now return the Rule Display name if it exists, returns nil otherwise. 
* Refactor in the log alert forwarding logic and rule engine: 
 * The only rule-related fields the rules-engine stores is the ruleID, ruleVersion and the fields generated dynamically (alert title and alert dedup). RuleID and RuleVersion can be used by any downstream consumers to retrieve any specific about the rule. This means that the rules-engine no longer stores the rule severity. This makes the design makes it easier to downstream consumers to retrieve any rule-related information they might require. 
  * The log-alert-forwarder that consumes the rules-engine output is now making use of the ruleID and ruleVersion to fetch any fields related to a Rule. 

## Changes

- GetAlert and ListAlerts now includes the Rule display name (if available) and the Alert title. The alert Title will always be populated
- All rules that are running in the rules-engine are not expected to have a Rule Version. Until now, all rules in the system had a version, but we were not enforcing it. 

## Testing

- Unit tests
- Deployed to account: 
 * Created Rule with title method. Seen it trigger and was able to list alerts in the UI. I was able to see the new fields when invoking the Lambda from AWS Console. 
 * Created Rule without title method. Seen it trigger and was able to list alerts in the UI. I was able to see the new fields when invoking the Lambda from AWS Console.


api/lambda/alerts/models/api.go
```diff
@@ -83,23 +77,29 @@ type ListAlertsOutput struct {
...
+	DedupString     *string    `json:"dedupString,omitempty"`
+	CreationTime    *time.Time `json:"creationTime" validate:"required"`
+	UpdateTime      *time.Time `json:"updateTime" validate:"required"`
+	EventsMatched   *int       `json:"eventsMatched" validate:"required"`
+	Severity        *string    `json:"severity" validate:"required"`
+	Title           *string    `json:"title" validate:"required"`
 }
 
 // Alert contains the details of an alert
 type Alert struct {
```
rleighton, line 93: This looks like it shares most of AlertSummary. Consider composing from AlertSummary:
```
type Alert struct {
 AlertSummary
 Events ...
```

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -82,37 +148,40 @@ func SendAlert(event *AlertDedupEvent) error {
 	return nil
 }
 
+func getAlertTitle(rule *models.Rule, alertDedup *AlertDedupEvent) string {
+	if alertDedup.GeneratedTitle != nil {
+		return *alertDedup.GeneratedTitle
+	}
+	ruleDisplayName := getRuleDisplayName(rule)
+	if ruleDisplayName != nil {
+		return *ruleDisplayName + " failed"
```
rleighton, line 157: Can you add a comment explaining why you add "failed" to the end .

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -104,6 +106,18 @@ func (API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOutput,
...
 }
 
+// Method required for backwards compatibility
+// In case the alert title is empty, return custom title
+func getAlertTitle(alert *table.AlertItem) *string {
+	if alert.Title != nil {
+		return alert.Title
+	}
+	if alert.RuleDisplayName != nil {
+		return aws.String(*alert.RuleDisplayName + " failed")
```
rleighton, line 116: Comment here too why "failed" is added to the end.

api/lambda/alerts/models/api.go
kostaspap: Good idea! Merged

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
kostaspap: This is a default value i specified - we might want to display a different value though. I made a comment here: https://github.com/panther-labs/panther/issues/615#issuecomment-609793892 and will update this PR according to the outcome of that discussion.

internal/log_analysis/alerts_api/api/get_alert.go
kostaspap: See above.

api/lambda/alerts/models/api.go
```diff
@@ -33,23 +33,17 @@ type LambdaInput struct {
...
 //     "getAlert": {
-// 	    "alertId": "ruleId-2"
+// 	    "alertId": "ruleId-2",
+//         "eventsPageSize": 20
 //     }
 // }
 type GetAlertInput struct {
 	AlertID                 *string `json:"alertId" validate:"required,hexadecimal,len=32"` // AlertID is an MD5 hash
-	EventsPageSize          *int    `json:"eventsPageSize,omitempty"  validate:"omitempty,min=1,max=50"`
+	EventsPageSize          *int    `json:"eventsPageSize"  validate:"required,min=1,max=50"`
```
austinbyers, line 42: Why this change? Seems reasonable to have a default page size instead of forcing the caller to specify it every time

api/lambda/alerts/models/api.go
```diff
@@ -83,23 +77,21 @@ type ListAlertsOutput struct {
...
 type Alert struct {
-	AlertID                *string    `json:"alertId"`
-	RuleID                 *string    `json:"ruleId"`
-	DedupString            *string    `json:"dedupString"`
-	CreationTime           *time.Time `json:"creationTime"`
-	UpdateTime             *time.Time `json:"updateTime"`
-	EventsMatched          *int       `json:"eventsMatched"`
-	Events                 []*string  `json:"events"`
-	EventsLastEvaluatedKey *string    `json:"eventsLastEvaluatedKey,omitempty"`
+	*AlertSummary
```
austinbyers, line 94: My personal preference is to keep embedded structs as non-pointers unless they can be optional

internal/log_analysis/alert_forwarder/forwarder/forwarder_test.go
```diff
@@ -18,6 +18,24 @@ package forwarder
...
+* License, or (at your option) any later version.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU Affero General Public License for more details.
+*
+* You should have received a copy of the GNU Affero General Public License
+* along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
```
austinbyers, line 37: Remove extra license header

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -75,14 +72,9 @@ def __init__(self, config: Dict[str, Any]):
...
 
-        if not ('severity' in config) or not isinstance(config['severity'], str):
-            raise AssertionError('Field "severity" of type str is required field')
-        self.rule_severity = config['severity']
-
         if not ('versionId' in config) or not isinstance(config['versionId'], str):
-            self.rule_version = DEFAULT_RULE_VERSION
-        else:
-            self.rule_version = config['versionId']
+            raise AssertionError('Field "versionId" of type str is required field')
```
austinbyers, line 76: Was the versionID always specified when invoking this function? If not, this may cause problems updating existing deployments depending on what order the components update in - events that were previously valid might suddenly get dropped

api/lambda/alerts/models/api.go
kostaspap: Sure, no strong preference on that. Will update.

internal/log_analysis/rules_engine/src/rule.py
kostaspap: Yup yup, it was always specified: 

https://github.com/panther-labs/panther/blob/master/internal/log_analysis/rules_engine/src/engine.py#L94
(the `raw_rule` above is actually the results from the `GetRule` operation from the Analysis API that always contains a version).

kostaspap: The previous check was coming mostly from my misunderstanding about Rule Version. I was under the impression that the Rule Version was not officially supported by the Analysis API but I realized that the backend fully supports rule versioning - it is just something we haven't integrated in the UI. 
From our recent discussions it became clear that Rule Version is well supported in the backend so in this PR I am making sure to update log analysis to treat Rule Version as a required field - it also simplifies a lot the way we deal with rules.

api/lambda/alerts/models/api.go
kostaspap: Actually the backend was requiring this value - if someone wouldn't specify it, it would panic with nil pointer error. The frontend which is calling this method was specifying always the page size. 
I didn't remove functionality but updated the API validation to reflect the way backend was treating this value. 
I like the idea of defaulting to a value for API calls that require some page size - I would rather put it in a separate PR though and only if we decide to apply same logic to all similar APIs.

api/lambda/alerts/models/api.go
kostaspap: Updated

internal/log_analysis/alert_forwarder/forwarder/forwarder_test.go
kostaspap: Fixed

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
kostaspap: Following the [discussion](https://github.com/panther-labs/panther/issues/615#issuecomment-609793892) , in the cases where there is no dynamically generated alert title, we will use the  Rule Display name (if one is available) or the Rule ID.

PR #660, opened 2020-04-06
Fix Password Validation for Special characters

### LINKED ISSUE #597: Password Validation: Special Characters
### Describe the bug

The frontend validation regex for checking user password requirements is missing some special characters. 

### Steps to reproduce

1. Set the user password
2. Add a `+` or `-`
3. See error message: "Include at least 1 special character"

### Expected behavior

All special characters should be expected, or a message should display which characters are available for use.

### Environment

Panther version or commit: v1.0

### Additional context

None

alexmylonas:
## Background

Password regex for special characters didn't include all special characters (i.e `+`,`-`)

Closes #597 

## Changes

- Updated regex to check if password has anything else except alphanumeric characters

## Testing

- Locally
- You can also validate new regex [here](https://www.regextester.com/?fam=115596)


web/src/constants.ts
```diff
@@ -29,7 +29,7 @@ export const SOURCE_LABEL_REGEX = new RegExp('^[a-zA-Z0-9- ]+$');
 
 export const INCLUDE_UPPERCASE_REGEX = new RegExp('(?=.*[A-Z])');
 
-export const INCLUDE_SPECIAL_CHAR_REGEX = new RegExp('(?=.*[!@#\\$%\\^&\\*;:,.<>?/])');
+export const INCLUDE_SPECIAL_CHAR_REGEX = new RegExp('[^\\d\\sA-Za-z]');
```
kostaspap, line 32: 👍

alexmylonas:
I am 50-50 about this is a negative pattern check the alternative would be to list all "common" special characters like this `[±§!@#$%^&*()_+=\-{}[\]/\\?,.;'"]`

But other special characters like `◕◡◕  ☻` would be disregarded. If you add an all greek password you would still required to an english character to satisfy the other check.

Post that made me decide: https://stackoverflow.com/questions/18057962/regex-pattern-including-all-special-characters

3nvi:
> I am 50-50 about it. This is a negative pattern check the alternative would be to list all "common" special characters like this `[±§!@#$%^&*()_+=\-{}[\]/\\?,.;'"]`
> 
> But other special characters like `◕◡◕ ☻` would be disregarded. If you add an all greek password you would still required to an english character to satisfy the other check.
> 
> Post that made me decide: https://stackoverflow.com/questions/18057962/regex-pattern-including-all-special-characters

Convinced me as well. Let's go ahead and keep your implementation :)

alexmylonas:
Who can resist Baby Angel 👼

PR #661, opened 2020-04-06
Checking existing user emails on UserForm validation

### LINKED ISSUE #631: User which fails to invite still added to user list
### Describe the bug

If you invite a user with an email that already exists, the backend rejects the request, but the frontend still shows the newly added user:

Screen Shot 2020-04-01 at 12 19 56 PM

If you refresh the page, it will go away, which proves the duplicate user was never added in the backend

### Steps to reproduce

1. Invite a user with an email that already exists
2. See both the error message and the user in the list
3. Refresh the page and see the "duplicate" user removed

### Expected behavior

Either don't add the new user to the list until the response is successful or remove it from the list after a failed response

### Environment

- Latest master

alexmylonas:
## Background

Users shouldn't be able to invite that already exist or change an email of a user to one that already exists. This functionality is implemented correctly in the backend but there is an issue on the frontend that displays wrong results.

The root cause is an Apollo issue described [here](https://github.com/apollographql/apollo-client/issues/5790) 

This is a temporal fix that performs a validation on the UserForm to prohibit users from typing an email that already exists.

Closes #631 

## Changes

- Added a conditional validation for input emails to not match an email of an existing user

## Testing

- Locally


web/src/components/forms/UserForm/UserForm.tsx
```diff
@@ -37,15 +38,23 @@ export interface UserFormProps {
...
-  email: Yup.string()
-    .email('Must be a valid email')
-    .required('Email is required'),
-  familyName: Yup.string().required('Last name is required'),
-  givenName: Yup.string().required('First name is required'),
-});
-
 const UserForm: React.FC<UserFormProps> = ({ initialValues, onSubmit }) => {
+  /*
+   This is temporal fix for inviting OR editing users that already exist
```
alexmylonas, line 43: Should we remove this eventually? It seems like a good validation

web/src/components/forms/UserForm/UserForm.tsx
```diff
@@ -37,15 +38,23 @@ export interface UserFormProps {
...
+   when this is fixed we should revert it: https://github.com/apollographql/apollo-client/issues/5790
+   */
+  const { data } = useListUsers({ fetchPolicy: 'standby' });
```
alexmylonas, line 46: I am not entirely sure about the `standby` policy, but every other policy triggers rendering twice. Any suggestions?

3nvi: Standby is not a proper policy: https://github.com/apollographql/apollo-client/blob/f08ab2a6d39ca12b1618a1fcabd468fe2a1fb055/src/core/watchQueryOptions.ts#L15

Can you explain what do you mean by "rendering twice"?

web/src/components/forms/UserForm/UserForm.tsx
3nvi: Nitpick: since cache-first is the default we can event remove it completely

web/src/components/forms/UserForm/UserForm.tsx
alexmylonas: Updated to `cache-first` policy but `data` seems to causing a re-rendering similar to https://github.com/apollographql/react-apollo/issues/3906
Doens't affect UX in any way so we can leave it as it is

web/src/components/forms/UserForm/UserForm.tsx
alexmylonas: Done!

PR #667, opened 2020-04-07
Add enterprise docs, tweak default python, show AWS region/account in settings

austinbyers:
## Background

Add enterprise documentation for how to use roles and other minor documentation tweaks.

## Changes

- Add an `enterprise` section of documentation which explains how roles work
- Modify the default Python bodies to explain what needs to be returned. I couldn't remember whether to return True or False for policies vs. rules, I think more context there would be helpful
- Add the AWS region and account ID to the general settings view. When using a custom domain, there's no indication in Panther itself about where it is deployed. This information is especially helpful for service providers managing multiple deployments (ourselves included)
- Update Go modules

## Testing

- `mage fmt test:ci`

Deployed to dev account:

Screen Shot 2020-04-07 at 3 05 04 PM
Screen Shot 2020-04-07 at 3 05 16 PM

Screen Shot 2020-04-07 at 3 17 27 PM


docs/gitbook/enterprise/rbac.md
```diff
@@ -0,0 +1,26 @@
+# Role-Based Access Control
+Role-Based Access Control (RBAC) allows enterprise deployments to enforce least privilege access -
+users are assigned roles with fully customizable permission sets.
+
+## Roles
+In Panther Enterprise, a *role* is a configurable set of permissions and every user is assigned
+exactly one role. By default, the following 3 roles are created for you:
+
+<img src="docs/img/rbac-default-roles.png" alt="Default Roles"/>
```
jacknagz, line 9: These images aren't loading, and can you put them into the same folder as the rest of the gitbook assets?

docs/gitbook/enterprise/README.md
```diff
@@ -0,0 +1,16 @@
+# Panther Enterprise
```
nhakmiller, line 1: Should we put some contact information in here somewhere so that someone can email/call/whatever us with enquiries about upgrading to enterprise?

docs/gitbook/enterprise/rbac.md
```diff
@@ -0,0 +1,26 @@
+# Role-Based Access Control
+Role-Based Access Control (RBAC) allows enterprise deployments to enforce least privilege access -
```
nhakmiller, line 2: Maybe specify `least privilege access within the Panther app itself`, Panther infra already runs with least privilege principles being respected by the backend infrastructure, this is just for the Panther application level access. Or maybe that's inferred, not sure.

docs/gitbook/enterprise/rbac.md
```diff
@@ -0,0 +1,26 @@
...
+users are assigned roles with fully customizable permission sets.
+
+## Roles
```
jacknagz, line 5: Let's change this to `Built-in Roles`

web/src/constants.ts
```diff
@@ -32,16 +32,16 @@ export const INCLUDE_UPPERCASE_REGEX = new RegExp('(?=.*[A-Z])');
 export const INCLUDE_SPECIAL_CHAR_REGEX = new RegExp('[^\\d\\sA-Za-z]');
 
 export const DEFAULT_POLICY_FUNCTION =
-  'def policy(resource):\n\t# Write your code here.\n\treturn True';
+  'def policy(resource):\n\t# Return True if the resource is passing/compliant.\n\treturn True';
 
 export const DEFAULT_RULE_FUNCTION =
-  'def rule(event):\n\t# Write your rule code here.\n\treturn False';
+  'def rule(event):\n\t# Return True to match the event and trigger an alert.\n\treturn False';
```
nhakmiller, line 38: Great user focused improvement! I think we should be consistent on recommending to return the state that triggers an alert or the state that doesn't trigger an alert.

So either policy could be changed to `# Return False if the resource is non-compliant and an alert should be triggered` or the rule could be changed to `# Return False if the event should not trigger an alert`.  Between the two, I kind of prefer the first.

docs/gitbook/enterprise/rbac.md
```diff
@@ -0,0 +1,26 @@
...
+
+All roles (including the default ones above) are fully customizable by any user with UserModify permissions:
+
+* You can create as many roles as you want
+* Roles can be renamed as long as the names are unique
+* Role permissions can be changed as long as at least one user has UserModify permissions
+* Roles can be deleted as long as no users are currently assigned to them
+
+<img src="docs/img/rbac-role-edit.png" alt="Role Edit"/>
+<p align="center">A subset of the role edit page</p>
```
jacknagz, line 26: Can we add a `## Create a Role` section?

docs/gitbook/enterprise/rbac.md
austinbyers: Ah, I think I might have the wrong relative path

docs/gitbook/enterprise/rbac.md
austinbyers: "Default Roles" maybe, but sure

docs/gitbook/enterprise/README.md
austinbyers: The pricing page I link to already has the "contact sales" buttons, I can add that here as well if we want? Just don't want it to be out of date when the main website changes

docs/gitbook/enterprise/rbac.md
austinbyers: Sure, I'll reword

docs/gitbook/enterprise/README.md
nhakmiller: That's probably fine then, I guess its pretty natural to go to the pricing page as the first step to signing up for enterprise.

docs/gitbook/enterprise/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+In addition to the open-source community edition, Panther offers an enterprise product with all
+community features plus:
+
+* 24x7 Live Support
```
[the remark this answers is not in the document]
austinbyers: Will do.

FYI, I was just matching what was on the [pricing page](https://runpanther.io/pricing/)

docs/gitbook/enterprise/README.md
austinbyers: I nonetheless made it a bit more prominent: moved it to its own line and added a link to request a demo (which is where the pricing page ultimately directs as well)

docs/gitbook/enterprise/rbac.md
austinbyers: Moved to assets, images render now

docs/gitbook/enterprise/rbac.md
austinbyers: Restructured this section slightly

docs/gitbook/enterprise/rbac.md
austinbyers: I added a "customize roles" section which talks about both creating and editing roles (which is exactly the same workflow). Added more language to clarify it is both create and edit

austinbyers:
@jacknagz comments addressed, ready for re-review

docs/gitbook/enterprise/rbac.md
```diff
@@ -0,0 +1,32 @@
+# Role-Based Access Control
+Role-Based Access Control (RBAC) gives Panther Enterprise deployments fine-grained access control for
+their user accounts. Each user is assigned a role with a fully customizable permission set.
+
+## Roles
```
jacknagz, line 5: I don't think we need this header

Let's just use H2 for the subsequent ones too.

Issue #408, opened 2020-03-13
Optimize UX for updates


3nvi:
### Description

Just like with deletions, we don't need to perform a network request for a refetch when something gets updated. We know the values aforehand. In fact, Apollo [**automatically** updates the cache](https://www.apollographql.com/docs/react/data/mutations/#updating-a-single-existing-entity) when an item gets updated, if the mutation returns an `id`.We should leverage that and make sure that all updates have optimistic responses and that they get updated automatically.  You can find related code [here](https://github.com/panther-labs/panther-enterprise/pull/16/files#diff-963bc10850addb399bf425e12946bc44R51-R67).

There are some specific parts that are going to be tough, since not all modification mutations are going to return proper responses containing `id`s. Some of them require the assignee to dig and understand what kind of cache modification would need to happen. These in particular are:

* usePolicySuppression
* RemediationButton

For those particular modifications you can't rely on Apollo (since no `id` is returned), but you need to understand what they do & write an appropriate [update function](https://www.apollographql.com/docs/angular/features/cache-updates/#update) in order to update the cache in the places that would normally be updated from the - soon to be removed - `refetchQueries`.

### Acceptance Criteria

This task should make sure that:

- Modification mutations (a.k.a. updates) never have a `refetchQueries` associated with them.  The locations are:

* ResetUserPasswordModal
* UpdateSourceSidesheet
* EditUser
* usePolicySuppression
* RemediationButton

- Modification mutations (a.k.a. updates) that can't return `id`s  have  an `update` function in order to perform a manual update to the cache.  The locations are:

* ResetUserPasswordModal
* usePolicySuppression
* RemediationButton

- All modification mutations leverage Apollo's internals to perform a cache update automatically (by returning an `id` field). This should happen wherever there is an update in the project (not only in the places mentioned in the bullet above). Check the `schema.graphql` to see which mutations modify data.
 
- An `optimisticResponse` gets added for all the modification mutations (same places as the bullet right above that added the `id` field)

alexmylonas:
By @3nvi in #672 
> Hmm I see that we don't get the "new status" of the user from the server. If we do, we can remove the refetchQueries call that is present on a successful "reset password" operation on the ResetUserPasswordModal.
> 
> This is optional since I think it's covered by #408

3nvi:
> By @3nvi in #672
> 
[...]

PR #672, opened 2020-04-08
Force Change password for Users more than once

### LINKED ISSUE #397: Resetting User not CONFIRMED throws error in platform
### Describe the bug

After resetting a user's password and change it's status in FORCE_CHANGE_PASSWORD and you attempt to reset it again, you 'll get an error.

### Steps to reproduce

1. Go to Users Page
2. Click on "Reset password" for a user
3. Click on it again

### Expected behavior

a. Option should grayed out or hidden for status not CONFIRMED
b. Resending email to that user

### Environment

- Panther version or commit: 9140b62
- Browser: all

alexmylonas:
## Background

Resetting users password more than once is now supported in the backend, this PR handles changes to enable this on frontend too.

Current behavior: 
 - You can reset a user's password as many time as you want but there is time limit from cognito
-  You can't reset the password for a user that hasn't set already a password (invited users)

Note: MFA is not reseted when a user is forced to change password, which to my understanding is valid. 

Closes #397 
## Changes

- Changed schema to return `User`
- Changed `resetUserPassword` to expect `id` of user
- Hide "Force Change password" for invited users

## Testing

- Locally
- Deployed env


web/src/components/modals/ResetUserPasswordModal/graphql/resetUserPassword.graphql
```diff
@@ -1,3 +1,5 @@
 mutation ResetUserPassword($id: ID!) {
-    resetUserPassword(id: $id)
-}
\ No newline at end of file
+  resetUserPassword(id: $id) {
+    id
```
3nvi, line 3: Hmm I see that we don't get the "new status" of the user from the server.  If we do, we can remove the `refetchQueries` call that is present on a successful "reset password" operation on the `ResetUserPasswordModal`.

This is optional since I think it's covered by #408

alexmylonas: I have tested it and seems like the backend doesn't return anything else than ID -> https://github.com/panther-labs/panther/blob/75a96feda33f8e70a8359de33e54740e801faa19/internal/core/users_api/api/reset_user_password.go#L32

@austinbyers is that correct?

alexmylonas:
Merging this since it's blocking my other task, @3nvi i am adding your comment to #408

Issue #583, opened 2020-03-26
clarify license in readme


kapilt:
### Describe the bug

From the readme, "Panther is dual-licensed under the AGPLv3 and Apache-2.0 licenses."

The common usage of dual-license in oss is for source code that is available wholly under either license, except a cursory look at your license file makes it clear that not the case.
https://github.com/panther-labs/panther/blob/master/LICENSE

Instead an accurate phrase would be "parts of the code base are selectively licensed under either the AGPLv3, Apache-2.0, or Panther Commercial License"

jacknagz:
Thanks @kapilt! That makes sense.

austinbyers:
@kapilt great feedback! I'll clarify in the license

Most of the code is AGPLv3, the idea was to allow common utility code that other people might find especially useful to live under a less restrictive license (Apache v2) - this is the `pkg` folder. But this distinction might do more harm than good in terms of the confusion it causes

For now, I'll just clarify the license text as you suggest, but we'll also sync as a team and see if we can just remove this distinction somehow to simplify things

PR #679, opened 2020-04-09
Use AGPLv3 license for the entire repo

austinbyers:
## Background

Panther already uses AGPLv3 for the majority of the source code - the one exception is the `pkg` folder, which contains some commonly used Go utilities. This exception forces a non-standard LICENSE file and can cause confusion (#583), so we decided to simplify to one license across the whole repo.

The idea with `pkg` was that it might provide more generically useful code that other projects could use directly, and we felt Apache v2 might help encourage that. But we realized that our `pkg` utilities are probably not yet in a state where they're useful for others yet - they need more complete functionality, documentation, and test coverage before they could be standalone Go packages that would have value outside of Panther.

We still hope to move in that direction in the long term and make some of those libraries robust and generic enough to release on their own. If and when that happens, we will likely move them to their own repo, at which time we'll re-evaluate which license would make the most sense for each package.

In the meantime, let's keep things clear and simple: AGPLv3 for all Panther source!

## Changes

- Change LICENSE file to https://www.gnu.org/licenses/agpl-3.0.txt
- Adjust `mage fmt`, README, and documentation
- Change Apache license headers to AGPL
- Added license headers for TF files (#356) and all web src (some was previously missing)

## Testing

- `mage test:ci`
- Searched the repo for all references to "Apache"


pkg/gatewayapi/replace_nils_test.go
```diff
@@ -1,19 +1,21 @@
 package gatewayapi
 
 /**
- * Copyright 2020 Panther Labs Inc
+ * Panther is a scalable, powerful, cloud-native SIEM written in Golang/React.
```
jacknagz, line 4: Can we modify this first line to `A Cloud-Native SIEM for the Modern Security Team`?

austinbyers: Sure

PR #687, opened 2020-04-10
Optimize UX & Cache for addition of certain items

### LINKED ISSUE #409: Optimize UX for addition of certain items
### Description

Adding items is hard when the items are paginated. This is because the front-end never knows where the newly added items should end up. For those items, we need to keep having a `refetchQueries` present, but we also need to invalidate all existing queries that show these items (through `cache.evict`) prior to refetching new ones from the server.

### Acceptance Criteria

We should make sure that:

* When items are added in paginated queries, the existing queries get invalidated before the existing `refetchQueries` kicks in. 

* We review whether we should retain `awaitRefetchQueries` depending on the query (in some we need to wait for it, while in others the queries that are refetched can be a background process that we shouldn't wait for)

* When items are added in non-paginated queries & queries that , no `refetchQueries` is called and we simply use `cache.modify` to add the newly created item to the associated cacheEntries. This includes destinations, sources, users, etc.

alexmylonas:
## Background

Optimising cache data for adding new items to reduce wait time, improve user experience and show valid results as described on #409 : 

Non-paginated data: 
 - All except `Destinations` seems to already follow the optimised pattern

Paginated data: 
- Invalidated cache when creating a new `Rules`, `Policies`

Closes #409 

## Changes

- Updated `AddDestinationSidesheet` to use `cache.modify`
- Updated `CreatePolicy` to use `cache.evict`
- Updated `CreateRule` to use `cache.evict`
- Minor fixes for related files

## Testing

- Locally


web/src/components/sidesheets/AddDestinationSidesheet/graphql/addDestination.generated.ts
```diff
@@ -110,6 +110,7 @@ export const AddDestinationDocument = gql`
       }
       verificationStatus
       defaultForSeverity
+      __typename
```
alexmylonas, line 113: Added __typename to please `ts` that property was missing from object

web/src/components/sidesheets/AddDestinationSidesheet/graphql/addDestination.graphql
```diff
@@ -1,49 +1,50 @@
...
+      asana {
+        personalAccessToken
+        projectGids
+      }
     }
-}
\ No newline at end of file
+    verificationStatus
+    defaultForSeverity
+    __typename
```
alexmylonas, line 48: Added __typename to please ts that property was missing from object

web/src/pages/CreatePolicy/CreatePolicy.tsx
```diff
@@ -49,8 +47,11 @@ export const initialValues: PolicyDetails = {
 const CreatePolicyPage: React.FC = () => {
   const { history } = useRouter();
   const [createPolicy, { error }] = useCreatePolicy({
-    refetchQueries: [getOperationName(ListPoliciesDocument)],
```
alexmylonas, line 52: This query seems to never trigger before. Am I missing something?

web/src/pages/CreateRule/CreateRule.tsx
```diff
@@ -51,8 +49,11 @@ export const initialValues: RuleDetails = {
 const CreateRulePage: React.FC = () => {
   const { history } = useRouter();
   const [createRule, { error }] = useCreateRule({
-    refetchQueries: [getOperationName(ListRulesDocument)],
```
alexmylonas, line 54: Same as ^^

web/src/components/sidesheets/AddDestinationSidesheet/graphql/addDestination.graphql
3nvi: yeah sometimes we need it sometimes not. You are correct to add it :)

web/src/pages/CreatePolicy/CreatePolicy.tsx
3nvi: chatted offline about the fact that only observable queries are refetched when passing a string as a parameter to `refetchQueries`

web/src/pages/CreateRule/CreateRule.tsx
3nvi: same as well :P

web/src/components/sidesheets/AddDestinationSidesheet/AddDestinationSidesheet.tsx
```diff
@@ -86,20 +84,14 @@ const AddDestinationSidesheet: React.FC<AddDestinationSidesheetProps> = ({ desti
...
-              ...existingData,
-              destinations: [data.addDestination, ...existingData.destinations],
+        update: (cache, { data: { addDestination: newDestination } }) => {
+          cache.modify('ROOT_QUERY', {
+            destinations: (queryData, { toReference }) => {
+              const addDestinationRef = toReference(newDestination);
+              return queryData ? [addDestinationRef, ...queryData] : [addDestinationRef];
             },
           });
+          cache.gc();
```
3nvi, line 94: this is not needed since we are adding data and not removing. Garbage collection is needed for deletions :)

web/src/pages/CreateComplianceSource/CreateComplianceSource.tsx
```diff
@@ -38,7 +38,7 @@ const CreateComplianceSource: React.FC = () => {
       cache.modify('ROOT_QUERY', {
         listComplianceIntegrations: (queryData, { toReference }) => {
           const addedIntegrationCacheRef = toReference(addComplianceIntegration);
-          return queryData ? [addedIntegrationCacheRef, ...queryData] : queryData;
+          return queryData ? [addedIntegrationCacheRef, ...queryData] : [addedIntegrationCacheRef];
```
3nvi, line 41: oh wow! Thanks for that! Definitely a bug

web/src/pages/CreatePolicy/CreatePolicy.tsx
```diff
@@ -49,8 +47,11 @@ export const initialValues: PolicyDetails = {
...
     onCompleted: data => history.push(urls.compliance.policies.details(data.addPolicy.id)),
+    update: cache => {
```
3nvi, line 51: why are we evicting the entire cache? I think that just refetching the 1st page would be enough. Am I missing something?

web/src/pages/CreateRule/CreateRule.tsx
```diff
@@ -51,8 +49,11 @@ export const initialValues: RuleDetails = {
...
     onCompleted: data => history.push(urls.logAnalysis.rules.details(data.addRule.id)),
+    update: cache => {
+      cache.evict('ROOT_QUERY', 'rules');
```
3nvi, line 54: same comment as above

web/src/pages/CreatePolicy/CreatePolicy.tsx
alexmylonas: Thanks for the comment, made dive deeper on the `refetchQueries` and made it work, with small caveat. I 'll tag you on the change!

web/src/pages/CreatePolicy/CreatePolicy.tsx
```diff
@@ -49,7 +48,7 @@ export const initialValues: PolicyDetails = {
 const CreatePolicyPage: React.FC = () => {
   const { history } = useRouter();
   const [createPolicy, { error }] = useCreatePolicy({
-    refetchQueries: [getOperationName(ListPoliciesDocument)],
+    refetchQueries: [{ query: ListPoliciesDocument, variables: { input: {} } }],
```
alexmylonas, line 51: @3nvi This works well but refetches data only for first page, do you have any insight or how to refetch every existing query of this type?

web/src/pages/CreateRule/CreateRule.tsx
```diff
@@ -51,7 +50,7 @@ export const initialValues: RuleDetails = {
 const CreateRulePage: React.FC = () => {
   const { history } = useRouter();
   const [createRule, { error }] = useCreateRule({
-    refetchQueries: [getOperationName(ListRulesDocument)],
+    refetchQueries: [{ query: ListRulesDocument, variables: { input: {} } }],
```
alexmylonas, line 53: Same as above

web/src/pages/CreatePolicy/CreatePolicy.tsx
3nvi: we can't right now. This will do just fine (either way the `refetchQueries` was not even working before ^^)

PR #690, opened 2020-04-13
Support Global Analysis Type

nhakmiller:
NOTE: `aws_globals` files will no longer be treated as globals after this! Be sure to convert them to explicit `global` analysis types. See the `panther_analysis_tool` [documentation](https://github.com/panther-labs/panther_analysis_tool) for details.

## Background

Previously we supported global helper functions by allowing a user to create a policy or rule name `aws_globals`, and all global helpers were defined within that policy or rule. This methodology had a number of issues, so we decided to more formally support a `global` analysis type.

These `global` analysis types are implemented as lambda layers, which simplify the backend policy/rule engine infrastructure while also guaranteeing that the global modules are always up to date across all lambdas. They also simplify sharing code between the policy and rule engines.

## Changes

- Added support for a `global` analysis type to the `analysis-api`
  - Added CRUD actions for `global` type
  - Added communication with new `layer-manager` lambda via SQS when globals are changed
- Created the `layer-manager` lambda function, to be responsible for managing layers attached to the policy and rule engine. This is in support of this change, but also in anticipation of broader changes to support both multiple globals in one deployment as well as possibly implementing all policies and rules as layers
- Removed code responsible for managing global helper functions from the `policy-engine`, `rules-engine`, and the `resource-processor`
- Changed the `panther` helper functions to be uploaded as the default layer during initial onboarding, as opposed to being imported explicitly by the `policy-engine`

## Testing

- Manually deployed to dev account and tested
- Integration tests for new `analysis-api` endpoints


internal/core/analysis_api/handlers/get_enabled_policies.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/panther-labs/panther/pkg/gatewayapi"
 )
 
-// GetEnabledPolicies fetches all enabled policies from an organization for backend processing.
+// GetEnabledPolicies fetches all enabled policies or rules from an organization for backend processing.
```
rleighton, line 35: Does it also do "global"?

Could the name be changed from GetEnabledPolicies to GetEnabledPythonModule (or something ... confusing to use word "policies")

internal/core/layer_manager/manager/handle.go
```diff
@@ -0,0 +1,173 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/lambda"
+	"go.uber.org/zap"
+
+	analysisoperations "github.com/panther-labs/panther/api/gateway/analysis/client/operations"
+	"github.com/panther-labs/panther/api/gateway/analysis/models"
+)
+
+/**
```
rleighton, line 19: Remove copyright and re-run 'fmt' to get in the right place.

internal/core/analysis_api/handlers/get_enabled_policies.go
nhakmiller: It does not do global, I didn't think a concept of enabled vs. disabled was particularly useful when talking about globals since there will only ever be a handful (one right now) and so really they should just be deleted/removed if not in use. I will see about renaming this particular function, it's just a holdover from when everything was a policy.

deployments/core.yml
```diff
@@ -291,6 +292,13 @@ Resources:
                 - kms:Decrypt
                 - kms:GenerateDataKey
               Resource: !Sub arn:${AWS::Partition}:kms:${AWS::Region}:${AWS::AccountId}:key/${SqsKeyId}
+        - Id: PublishToLayerManagerQueue
+          Version: 2012-10-17
+          Statement:
+            - Effect: Allow
+              Action:
+                - sqs:SendMessage
```
austinbyers, line 300: Nit: we usually put singleton actions on a single line:

```yaml
Action: sqs:SendMessage
```

docs/gitbook/operations/runbooks.md
```diff
@@ -41,6 +34,13 @@ The `panther-alert-forwarder` lambda reads from the ddb stream for the table `pa
  * Failure of this lambda will stop delivery of alerts to destinations.
  * There will be no data loss until events are purged from the ddb stream (24 hours).
 
+## panther-alert-forwarder
```
austinbyers, line 37: @rleighton there still seems to be indeterminism in the generated doc ordering

internal/compliance/policy_engine/src/policy.py
```diff
@@ -76,16 +72,6 @@ def __init__(self, policies: List[Dict[str, Any]]) -> None:
         self._policies_by_type: Dict[str, List[Policy]] = collections.defaultdict(list)
         self._global_policies: List[Policy] = []  # List of policies that apply to all log types
 
-        # Import panther helper functions, and add them to sys.modules so they can be imported by
-        # user defined policies
-        sys.modules['panther'] = Policy.import_module(helpers.__name__, helpers.__file__)
```
austinbyers, line 81: Is anything special required to update existing deployments? If someone has `v1.0.1` and then deploys `v1.1.0`, will their existing policies still work? I see that you added the global layer setup as part of the deploy process, have you tested the migration?

internal/core/layer_manager/main/lambda.go
```diff
@@ -0,0 +1,51 @@
...
+	"go.uber.org/zap"
+
+	"github.com/panther-labs/panther/internal/core/layer_manager/manager"
+	"github.com/panther-labs/panther/pkg/lambdalogger"
+	"github.com/panther-labs/panther/pkg/oplog"
+)
+
+func lambdaHandler(ctx context.Context, event events.SQSEvent) (err error) {
+	lc, _ := lambdalogger.ConfigureGlobal(ctx, nil)
+	operation := oplog.NewManager("cloudsec", "layer_manager").Start(lc.InvokedFunctionArn).WithMemUsed(lambdacontext.MemoryLimitInMB)
```
austinbyers, line 36: Not sure if "cloudsec" is the right component name here

internal/core/layer_manager/manager/handle.go
austinbyers: I think this is the right place - after the "package" statement but before the imports

internal/core/layer_manager/manager/handle.go
```diff
@@ -0,0 +1,173 @@
...
+// that AWS is expecting.
+func packageLayer(analyses map[string]string) ([]byte, error) {
+	zap.L().Debug("packaging lambda layer")
+	buf := new(bytes.Buffer)
+	w := zip.NewWriter(buf)
+
+	for id, body := range analyses {
+		f, err := w.Create(layerPath + id + ".py")
+		if err != nil {
+			log.Fatal(err)
```
austinbyers, line 106: Why `log.Fatal` instead of returning the error?

tools/mage/deploy.go
```diff
@@ -572,5 +575,27 @@ func initializeAnalysisSets(awsSession *session.Session, endpoint string, settin
...
 	logger.Infof("deploy: initialized with %d policies and %d rules", newPolicies, newRules)
+
+	// Setup the initial helper layer
+	content, err := ioutil.ReadFile(defaultGlobalLocation)
+	if err != nil {
+		return fmt.Errorf("failed to read default globals file: %v", err)
+	}
+
+	logger.Infof("deploy: uploading initial global helper module")
+	_, err = apiClient.Operations.CreateGlobal(&operations.CreateGlobalParams{
```
austinbyers, line 586: This will happen on every deploy, is that what we want? Will this create a new layer or overwrite the existing one?

internal/core/analysis_api/handlers/get_enabled_policies.go
nhakmiller: Renamed

docs/gitbook/operations/runbooks.md
nhakmiller: I'm not quite on master, if a change was pushed recently to fix this issue I may still need to pull it in. I was waiting to merge master until after PR comments but I can do that now.

internal/compliance/policy_engine/src/policy.py
nhakmiller: If the global is created properly, then everything will just work. I'll look at smartening up the deploy step so that upgrades work.

tools/mage/deploy.go
nhakmiller: This should only happen in cases where no policies or rules were found, but I'm going to change the logic so that it checks for the existence separately so that upgrades work in place.

internal/core/layer_manager/manager/handle.go
nhakmiller: Good point, I'll return the error.

internal/core/layer_manager/main/lambda.go
nhakmiller: None of the other `core` lambdas had this set that I could find, so I went with `core` as the component.

internal/core/layer_manager/manager/handle.go
nhakmiller: I deleted it and re-ran `mage fmt`, either way it should be in the right place now.

nhakmiller:
> A huge change across many components, congrats on the broad work!
> 
> Please add integration tests now for the global endpoints. I know you're not using them to update the layers yet, so it's even easier to get the tests in place before rather than after. (Plus, I want to make sure we didn't break any existing integration tests)
> 
> Also double-check the policy-engine integration tests, which include an `aws_globals` policy

All current integration tests are passing. But it seems that's more by luck than anything, as an `aws_globals` policy is provided so it must just be imported first by good luck. I'll update that test to refer to import `panther`, which will assume that the layer-manager is deployed.

docs/gitbook/operations/runbooks.md
austinbyers: I think we fixed some of these issues, but not all of them

internal/core/analysis_api/handlers/create_global.go
```diff
@@ -1,5 +1,23 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
 /**
```
austinbyers, line 21: Be sure to remove the old license headers too (it's possible I missed some). Just search for "cloud-native SIEM written in Golang/React"

nhakmiller: Good catch.

internal/core/analysis_api/handlers/delete_global.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// DeleteGlobal deletes existing globals.
+func DeleteGlobal(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyResponse {
+	input, err := parseDeletePolicies(request)
+	if err != nil {
+		return badRequest(err)
+	}
+
+	if err = dynamoBatchDelete(input); err != nil {
```
rleighton, line 52: Can we get into an inconsistent state if this fails part the way thru?

nhakmiller: If dynamo fails halfway through deleting stuff? Or this whole function fails half way through? Not sure what we can do about dynamo failing partway through, but if say the dynamo delete succeeds but the s3 delete fails we return an error message to the user. At that point they should be able to either re-try the operation or resolve it some other way if they want. Delete operations should be fairly idempotent, from my understanding requesting a deletion for a non-existing resource will still go through the entire function and not return an error.

internal/core/analysis_api/handlers/create_global.go
nhakmiller: Updated.

tools/mage/deploy.go
nhakmiller: Updated.

internal/core/analysis_api/handlers/delete_global.go
rleighton: We have 3 deletes, I guess the question is about failure and order. What happens if ddb, s3 work and update layer fails? Do we can "stuck" in any way? Do we need "undo" logic in a defer function? I don't know, I'm just wondering if all the implications have been thought through. Does deleting S3 before updating the layer matter? If updating the layer fails will we have lambda pointing at delete files (again, I don't know, just asking)?

nhakmiller: Gotcha. The dynamo copy is the current working configuration. The S3 version is the history of that configuration. The layer is what is actually deployed. The three failure scenarios:

1. Dynamo delete fails. The whole process is cancelled and nothing changes, no inconsistent state. The user sees an error and needs to try and delete the global again.
2. S3 delete fails. At this point, the global has already been deleted from dynamo, but the history is still present and the layer is not updated, so there is an inconsistency between what is in dynamo and what is actually running. The next time any change happens to globals successfully (create, update, or delete), the layer manager will re-create the whole thing and the inconsistent state will be resolved. I suppose we could try updating the layer anyways, even though the s3 delete failed, and at least the state would be consistent.
3. Layer update fails. At this point, there is no memory of the global anywhere except in the running layer. The state is inconsistent, but again the next time any change happens to a global it will be made consistent. Unfortunately there will be no indication of this to the user, and the layer will just quietly keep running the older global.

I think failure 1 is a non-issue, the user will be informed and simply have to press the delete button again. Since right now we really don't care if the s3 delete is successful, I can re-order it so that first the layer update happens then the s3 delete. Then we can only get an inconsistent state if the update layer fails, which is just a call to SQS send message. If that fails, it would be reasonable to try rolling back.

rleighton: Nice explanation. Perhaps cut-paste it into a comment here for posterity.

nhakmiller:
> The new lambda could use some integration tests.

I have added integration tests for the new lambda. Although it seems integration tests only pass on occasion, I repeatedly had integration test failures for the resources-api, compliance-api, and especially the opstools s3queue and requeue that were seemingly totally unrelated to my code, and would happen only sometimes.

tools/mage/teardown.go
```diff
@@ -146,6 +151,43 @@ func destroyEcrRepos(awsSession *session.Session, repoNames []*string) {
 	}
 }
 
+// Remove layers created for the policy and rules engines
+func destroyLambdaLayers(awsSession *session.Session) {
```
austinbyers, line 155: :+1:

tools/mage/teardown.go
```diff
@@ -146,6 +151,43 @@ func destroyEcrRepos(awsSession *session.Session, repoNames []*string) {
...
+
+	// Find the layers that need to be destroyed
+	var layersToDestroy []*string
+	for _, layer := range layers.Layers {
+		if aws.StringValue(layer.LayerName) == globalLayerName {
+			layersToDestroy = append(layersToDestroy, layer.LayerName)
+		}
+	}
+
+	// Find and destroy each version of each layer that needs to be destroyed
```
austinbyers, line 171: Maybe a brief log message here:

```go
logger.Infof("Removing %d versions of Lambda layer %s", len(layersToDestroy), globalLayerName)
```

internal/core/layer_manager/manager/handle.go
```diff
@@ -0,0 +1,240 @@
...
+			return nil, nil
+		}
+		return nil, err
+	}
+	return packageLayer(map[string]string{globalModuleName: string(global.Payload.Body)})
+}
+
+// packageLayer takes a mapping of filenames to function bodies and constructs a zip archive with the file structure
+// that AWS is expecting.
+func packageLayer(analyses map[string]string) ([]byte, error) {
```
rleighton, line 112: Since we do this in memory, should we spec a "BIG" lambda (e.g., 1G)  for the layer manager? Since this is rarely run, there is no cost issue.

internal/core/layer_manager/main/integration_test.go
```diff
@@ -0,0 +1,499 @@
...
+
+	// The test global
+	globalAnalysisID          = "panther"
+	globalAnalysisBody        = "def custom_always_true():\n\treturn True"
+	globalAnalysisBodyUpdated = "def custom_always_true():\n\treturn True\n\ndef more_logic():\n\treturn False"
+)
+
+var (
+	integrationTest bool
+	awsSession      = session.Must(session.NewSession())
```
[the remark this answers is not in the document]
nhakmiller: Not sure, but I updated it to be safe.

internal/core/layer_manager/manager/handle.go
nhakmiller: The maximum size of the layers is ~225 MB (cap enforced by AWS), assuming very little compression when zipping we're looking at around 400MB memory usage. Looks like I had it set to 128 MB, I'll bump it to 512 MB to be safe.

PR #692, opened 2020-04-13
Use of Fragments for common fields in queries & mutations

### LINKED ISSUE #411: Create Fragments for common fields
### Description

We have a lot of queries and mutations that repeatedly fetch the same fields for certain types. For example, when fetching a `User` or a `Destination` we have to copy and paste the same fields over and over among multiple queries and mutations (i.e. during listing of destinations, during update of a destination, etc.)

This ticket will focus on creating fragments for **certain** types, that will help us not have to re-type the same fields for a particular type over and over again. The Fragments should have a verbose name( i.e. `DestinationDetails` or `DestinationTeaser` or `DestinationSummary`) and should be based according to the use cases.

Fragments should live under `web/src/graphql/fragments` and each fragment should be in its own file (i.e. `web/src/graphql/fragments/DestinationDetails.graphql`. They will be automatically picked up by other graphql files, so you don't have to import them manually anywhere.

### Acceptance Criteria

- As a minimum, let's create Fragments for the following types:
  *  AlertDetails
  *  RuleDetails
  *  RuleTeaser
  * PolicyDetails
  *  PolicyTeaser
  * Destination
  * LogSource (describes aws-scan Integrations)
  * ComplianceSource (describes aws-s3 integrations) 
  * User
  * General Settings

- Fragments should live under `web/src/graphql/fragments` and each fragment should be in its own file 

- We **should not** create Fragments for everything. Only create Fragments for **repetitive** stuff. Things that are not repetitive can be left as is. 

*Tip: If you need you can also create a Fragment, by extending another Fragment*

alexmylonas:
## Background

We have duplicate code for multiple queries and mutations that can be isolated and be re-usable. Solving this will reduce copy-pasting of fields and provide consistency for data returned from graphql calls.

Fragment that have already implemented: 
`LogSource`
`ComplianceSource`
`User`

Closes #411 

## Changes

- Fragment for `GeneralSettings`
- Fragment for `RuleDetails`
- Fragment for `RuleSummary`
- Fragment for `Rule` extending `RuleDetails`
- Fragment for `RuleTeaser`
- Fragment for `AlertDetails`
- Fragment for `PolicyDetails`
- Fragment for `PolicySummary`
- Fragment for `Policies`
- Fragment for `Destinations`

## Testing

- Locally


web/src/components/sidesheets/AddDestinationSidesheet/graphql/addDestination.graphql
```diff
@@ -1,49 +1,5 @@
...
-                projectGids
-            }
-        }
-        verificationStatus
-        defaultForSeverity
-    }
-}
\ No newline at end of file
+  addDestination(input: $input) {
+    ...DestinationFragment
```
3nvi, line 3: let's rename to something more verbose like `DestinationDetails`. Generally, the word `Fragment` shouldn't be included in the name, since it's clear that this is a fragment (it follows `...`)

web/src/graphql/fragments/AlertDetailsFragment.graphql
```diff
@@ -0,0 +1,10 @@
+fragment AlertDetailsFragment on AlertDetails {
```
3nvi, line 1: again the word `Fragment` should be omitted. We should also select a proper name for the fragment, like `AlertDetailsComplete` or something better than that :P

web/src/graphql/fragments/DestinationFragment.graphql
```diff
@@ -0,0 +1,47 @@
+fragment DestinationFragment on Destination {
```
3nvi, line 1: same about name conventions :)

web/src/pages/EditPolicy/graphql/policyDetails.graphql
```diff
@@ -1,23 +1,12 @@
...
-        tags
-        body
-        tests {
-            expectedResult
-            name
-            resource
-            resourceType
-        }
+  policy(input: $input) {
+    ...PolicyDetailsFragment
```
3nvi, line 3: why are we not adding the body & tests here?

web/src/pages/PolicyDetails/graphql/policyDetails.graphql
```diff
@@ -1,54 +1,40 @@
...
-        suppressions
-        id
-        lastModified
-        reference
-        resourceTypes
-        runbook
-        severity
-        tags
+  policy(input: $policyDetailsInput) {
+    ...PolicyDetailsFragment
```
3nvi, line 6: ah perhaps that's why you didn't include the body & tests :)

PR #694, opened 2020-04-13
Add support for step functions in alarm generation

rleighton:
## Background
The Athena API will make use of step functions. This PR adds support for generating alarms for step functions.

## Changes
- add sfn support
- update tests

## Testing
mage test:ci
mage build:cfn


tools/cfngen/cloudwatchcf/alarms_sfn.go
```diff
@@ -0,0 +1,57 @@
...
+	},
+	})
+	return alarm
+}
+
+func generateSFNAlarms(resource map[interface{}]interface{}) []*Alarm {
+	return []*Alarm{
+		NewSFNAlarm(
+			"SFNError",
+			"ExecutionsFailed",
```
austinbyers, line 52: If a Lambda function in the state machine fails, we'll get an alarm for that, too? Will the step function alarms catch something the lambda alarms do not or vice versa?

rleighton: This will catch when step code fails itself. For example, passing bad arguments. If the lambda called in the step function fails we will get 2 alarms.

tools/cfngen/cloudwatchcf/alarms_sfn.go
```diff
@@ -0,0 +1,57 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import "fmt"
+
+type SFNAlarm struct {
```
kostaspap, line 23: What does SFN stand for? `State Function <?>`?

PR #695, opened 2020-04-13
Refactor pkg/awsathena

rleighton:
## Background
As part of the new Athena api, refactor pkg/awathena which is used to make the core api calls to Athena.

## Changes
- Made package function oriented.
- Update use in generating athena views

## Testing
mage test:ci
mage deploy


pkg/awsathena/query.go
```diff
@@ -31,83 +30,101 @@ const (
...
 
 	var startContext athena.QueryExecutionContext
-	startContext.SetDatabase(aq.Database)
+	startContext.SetDatabase(database)
 	startInput.SetQueryExecutionContext(&startContext)
 
 	var resultConfig athena.ResultConfiguration
-	if aq.S3ResultsPath != nil {
-		resultConfig.SetOutputLocation(*aq.S3ResultsPath)
+	if s3Path != nil {
```
kostaspap, line 51: What happens if this is nil? What's the behavior of the system?

pkg/awsathena/query.go
```diff
@@ -31,83 +30,101 @@ const (
...
-			return
+			return executionOutput, true, nil
+		default:
+			return executionOutput, false, nil
+		}
+	}
+
+	poll := func() (*athena.GetQueryExecutionOutput, error) {
+		for {
+			executionOutput, done, err := isFinished()
```
kostaspap, line 79: If the query has failed this will not return an error and there is no way for the caller to get this information, the `WaitForResults` never returns this information. I see why this can be restrictive for many users of this API.  
If we are to return athena.GetQueryResultsOutput we should also return the Status of the query.

tools/athenaviews/athenaviews.go
```diff
@@ -46,12 +47,7 @@ func CreateOrReplaceViews(athenaResultsBucket string) (err error) {
 		return err
 	}
 	for _, sql := range sqlStatements {
-		q := awsathena.NewAthenaQuery(sess, awsglue.ViewsDatabaseName, sql, &s3Path) // use default bucket
-		err = q.Run()
-		if err != nil {
-			return errors.Wrap(err, "CreateOrReplaceViews() failed")
-		}
-		err = q.Wait()
+		_, err := awsathena.RunQuery(athena.New(sess), awsglue.ViewsDatabaseName, sql, &s3Path) // use default bucket
```
kostaspap, line 50: See comment above: This query can fail due to e.g invalid SQL syntax and we will never be notified.

pkg/awsathena/query.go
rleighton: Defaults to the configured work space.

pkg/awsathena/query.go
rleighton: We do , the status of the query is in executionOutput.QueryExecution.Status.State which is returned.

tools/athenaviews/athenaviews.go
rleighton: Good catch, I changed the semantics in the new code and this is old code. Will fix.

rleighton: Nevermind, this does work in fact because if you pull results for a failed query you get the error:

=== RUN   TestIntegrationAthenaQueryBadSQL
    TestIntegrationAthenaQueryBadSQL: integration_test.go:66: InvalidRequestException: line 1:1: mismatched input 'wwwww' expecting {'(', 'SELECT', 'DESC', 'USING', 'WITH', 'VALUES', 'CREATE', 'TABLE', 'INSERT', 'DELETE', 'DESCRIBE', 'GRANT', 'REVOKE', 'EXPLAIN', 'SHOW', 'USE', 'DROP', 'ALTER', 'SET', 'RESET', 'START', 'COMMIT', 'ROLLBACK', 'CALL', 'PREPARE', 'DEALLOCATE', 'EXECUTE'}
        {
          RespMetadata: {
            StatusCode: 400,
            RequestID: "b877ceba-c824-40e8-9596-d3482e323c28"
          },
          AthenaErrorCode: "MALFORMED_QUERY",
          Message_: "line 1:1: mismatched input 'wwwww' expecting {'(', 'SELECT', 'DESC', 'USING', 'WITH', 'VALUES', 'CREATE', 'TABLE', 'INSERT', 'DELETE', 'DESCRIBE', 'GRANT', 'REVOKE', 'EXPLAIN', 'SHOW', 'USE', 'DROP', 'ALTER', 'SET', 'RESET', 'START', 'COMMIT', 'ROLLBACK', 'CALL', 'PREPARE', 'DEALLOCATE', 'EXECUTE'}"
        }

Added integration test to be sure.

pkg/awsathena/query.go
kostaspap: This method returns `athena.GetQueryResultsOutput` which has the following fields: 

```
type GetQueryResultsOutput struct {
	_ struct{} `type:"structure"`

	// A token to be used by the next request if this request is truncated.
	NextToken *string `min:"1" type:"string"`

	// The results of the query execution.
	ResultSet *ResultSet `type:"structure"`

	// The number of rows inserted with a CREATE TABLE AS SELECT statement.
	UpdateCount *int64 `type:"long"`
}
```

I don't see the field you are referring to.

rleighton: Sorry, I was not clear. IsFinished() uses that field to detect failure to stop, poll() exits, we call GetResults() which returns the error in err. I tested this in the new integration test I put in.

Issue #691, opened 2020-04-13
Mage command to deploy Panther modifies source files


kostaspap:
### Describe the bug

Mage command to deploy Panther modifies source files.

### Steps to reproduce

1. Checkout master branch
2. Run `mage deploy`
3. Run `git status` to see changed files

### Expected behavior

Running `mage deploy` command doesn't modify source files. 

### Environment

How are you deploying or using Panther?

- Panther version or commit: https://github.com/panther-labs/panther/commit/45017102d989ea7c7cef9e9139594c35be39cce2
- OS: MacOS 10.15.4
- Browser: Irrelevant

### Additional context

Might worth writing down/documenting the DAG of the different operations that mage performs - and how the are depending on each other. Similar to the graph in [this](https://proandroiddev.com/understanding-gradle-the-build-lifecycle-5118c1da613f) post.

austinbyers:
I see two types of changes (testing commit 7a87e60a)

1. `go.mod` and `go.sum` change after `mage setup`
    - go can freely change these files when it builds or tests, I'll do some research to see if I can make these more stable.
2. Generated API files (`api/gateway/*/{client,models}`, `web/__generated__/`) change after `mage build`
    - This fix is easy, we just need to `mage fmt` the generated files

alexmylonas:
@austinbyers This seem to still occur on c8125e9ce6460ec12ad703e266c9e0dca6189027

Example diff: 
![Screenshot 2020-04-14 at 6 22 47 PM](https://user-images.githubusercontent.com/14179917/79242421-fcc47780-7e7c-11ea-86ab-66ab0d0941c1.png)

austinbyers:
@alexmylonas that commit is on a branch, are you sure your branch is up to date with master?

From master (7fa37cc2b), I did a brand new clone of the panther repo. `mage setup` showed no diffs, `mage build:api` no diffs, `mage build:lambda` no diffs, and `mage fmt` still no diffs.

PR #696, opened 2020-04-13
Stabilize Go modules, format generated API files

austinbyers:
## Background

A few small fixes to setup/build/fmt which will help those commands run faster and avoid changes when building a fresh clone.

Closes: #691 

## Changes

- Remove the https://github.com/fatih/color package - it is no longer maintained. We were only using it for one line of the deploy output; removing it clears several modules that users and CI no longer need to install.
- `mage setup` now pins the version of `goimports` and runs `go mod download` instead of `go get./...`. This will download exactly the modules in `go.mod`, preventing changes to that file during setup (#691). This is also faster than asking go to scan the whole repo looking for which modules to download.
- `mage fmt` now includes `go mod tidy` to cleanup `go.mod` and `go.sum` on every PR
- `mage build:api` adds license headers and formats API files after generating them. Without this, `build:api` (and therefore `deploy` as well) results in dirty repo state (#691)
- `mage build:lambda` again builds Lambda functions in parallel. We originally built in parallel, but reverted to sequential builds when we realized that `go build` will modify the `go.mod` file when it needs new modules, which would cause builds to fail. Now that all users are pre-installing go modules (`mage setup`) before compilation, this will only cause issues when developers add a new package for the first time. In that case, devs just need to run `go get ./...` before building.

## Testing

- `mage clean setup build:lambda` on a fresh clone of this branch shows no pending changes
- `mage build:lambda` is safe from fresh clones and is now 62% faster (22.2 s => 8.5s)

### DIFF
--- a/tools/mage/build_namespace.go
+++ b/tools/mage/build_namespace.go
@@ -112,8 +112,23 @@ func (b Build) lambda() error {
 
 	logger.Infof("build:lambda: compiling %d Go Lambda functions (internal/.../main) using %s",
 		len(packages), runtime.Version())
+
+	// "go build" in parallel for each Lambda function.
+	//
+	// If you don't already have all go modules downloaded, this may fail because each goroutine will
+	// automatically modify the go.mod/go.sum files which will cause conflicts with itself.
+	//
+	// Run "go mod download" or "mage setup" before building to download the go modules.
+	// If you're adding a new module, run "go get ./..." before building to fetch the new module.
+	errs := make(chan error)
 	for _, pkg := range packages {
-		if err := buildPackage(pkg); err != nil {
+		go func(pkg string, errs chan error) {
+			errs <- buildPackage(pkg)
+		}(pkg, errs)
+	}
+
+	for range packages {
+		if err := <-errs; err != nil {
 			return err
 		}
 	}
--- a/tools/mage/build_namespace.go
+++ b/tools/mage/build_namespace.go
@@ -75,11 +75,23 @@ func (b Build) API() {
 		}
 		walk(client, handler)
 		walk(models, handler)
+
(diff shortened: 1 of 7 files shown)


austinbyers:
> Seems fine, although I suppose now developers will need to commit their go.mod and go.sum files

Right, as I've been doing more research and understanding better, these should always be committed.

> Perhaps with this new stuff those conflicts will be less frequent.

Exactly! With this change, they should *never* change unless you add a new module the codebase hasn't used before

austinbyers:
Going to go ahead and merge now because I have more `mage` changes that will go on top of this, but @kostaspap feel free to comment if you see something and I can address it in a future PR

tools/mage/build_namespace.go
```diff
@@ -100,8 +112,23 @@ func (b Build) lambda() error {
...
+	//
+	// If you don't already have all go modules downloaded, this may fail because each goroutine will
+	// automatically modify the go.mod/go.sum files which will cause conflicts with itself.
+	//
+	// Run "go mod download" or "mage setup" before building to download the go modules.
+	// If you're adding a new module, run "go get ./..." before building to fetch the new module.
+	errs := make(chan error)
 	for _, pkg := range packages {
-		if err := buildPackage(pkg); err != nil {
+		go func(pkg string, errs chan error) {
```
kostaspap, line 125: A big problem we had with this is that the laptop would "choke" doing 100% utilization for all the cores in their workstation. 
Would be good to set a limit to the parallelization, either a fixed number or  `max(1, runtime.NumCPU() -1)` so at least people have one core free in their workstations and avoid system freezes.

austinbyers: Sure, I'll add that to the mage backlog

PR #697, opened 2020-04-13
Cache dependencies in CI

austinbyers:
## Background

CircleCI currently spends 40 seconds installing python/go/npm/other dependencies on every build. But most of the time these dependencies don't change, so this is a waste of time and compute. Thankfully, CircleCI has native support for [dependency caching](https://circleci.com/docs/2.0/caching/) for exactly this reason!

## Changes

- Cache each language's dependencies in CircleCI. Any changes to dependencies will rebuild just the dependencies for that language
- Simplify `build:tools` and build in parallel (saves 10-30 seconds in CI)
- Added `--no-audit` to npm install (saves a few seconds, we have a separate test for auditing)

## Testing

- Dozens of CI builds
- Restoring from the cache usually saves ~20 seconds during the setup phase, though CircleCI cache performance seems to vary. We can always revert this if it proves to be slower on average

CircleCI steps now look like this:
Screen Shot 2020-04-13 at 4 58 07 PM

You can see in this case that setup only took ~20 seconds total (twice as fast as before)

### DIFF
--- a/tools/mage/build_namespace.go
+++ b/tools/mage/build_namespace.go
@@ -144,13 +144,10 @@ func (b Build) Tools() {
 func (b Build) tools() error {
 	// cross compile so tools can be copied to other machines easily
 	buildEnvs := []map[string]string{
-		{"GOOS": "darwin", "GOARCH": "386"},
 		// darwin:arm is not compatible
 		{"GOOS": "darwin", "GOARCH": "amd64"},
-		{"GOOS": "linux", "GOARCH": "386"},
 		{"GOOS": "linux", "GOARCH": "amd64"},
 		{"GOOS": "linux", "GOARCH": "arm"},
-		{"GOOS": "windows", "GOARCH": "386"},
 		{"GOOS": "windows", "GOARCH": "amd64"},
 		{"GOOS": "windows", "GOARCH": "arm"},
 	}
--- a/tools/mage/build_namespace.go
+++ b/tools/mage/build_namespace.go
@@ -34,8 +34,6 @@ import (
 
 const swaggerGlob = "api/gateway/*/api.yml"
 
-var buildEnv = map[string]string{"GOARCH": "amd64", "GOOS": "linux"}
-
 // Build contains targets for compiling source code.
 type Build mg.Namespace
 
@@ -123,7 +121,7 @@ func (b Build) lambda() error {
 	errs := make(chan error)
 	for _, pkg := range packages {
 		go func(pkg string, errs chan error) {
-			errs <- buildPackage(pkg)
+			errs <- buildLambdaPackage(pkg)
 		}(pkg, errs)
 	}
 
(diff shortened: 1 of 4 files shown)


austinbyers:
In aggregate, the latest [build for this branch](https://app.circleci.com/pipelines/github/panther-labs/panther/3257/workflows/7b9ca34c-b03c-40c3-9e9d-0820019c179d) is 22% (40 seconds) faster than the latest [master build](https://app.circleci.com/pipelines/github/panther-labs/panther/3240/workflows/268b4e56-6f49-49e6-b5f9-6703fdd2d26b)!

This translates to improved developer productivity and lower CI costs

tools/mage/build_namespace.go
```diff
@@ -144,80 +142,63 @@ func (b Build) Tools() {
...
-}
-
-func buildTools(tools, binDir, sourceDir string) error {
 	// cross compile so tools can be copied to other machines easily
-	archs := []string{"amd64", "386", "arm"} // yes arm, AWS is now supporting arm processors and they are cheap!
-	oses := []string{"linux", "darwin", "windows"}
-	blacklist := map[string]bool{ // incompatible combinations
-		"darwin:arm": true,
+	buildEnvs := []map[string]string{
+		{"GOOS": "darwin", "GOARCH": "386"},
```
kostaspap, line 147: nit: If were were to vote, I would vote for kicking out `386` architectures. I doubt anyone is running on this type of architecture anymore.

tools/mage/setup.go
```diff
@@ -92,7 +92,8 @@ func Setup() {
 // Fetch all Go modules needed for tests and compilation.
 //
 // "go test" and "go build" will do this automatically, but putting it in the setup flow allows it
-// to happen in parallel with the rest of the downloads.
+// to happen in parallel with the rest of the downloads. Pre-installing modules also allows
```
kostaspap, line 95: Comments goes for the full file: `installPythonEnv` and  `installNodeModules` become no-op if running in CI and files already exist. What about the rest - swagger, terraform,  goimports etc It seems we will still do the setup even if we are running in CI and files already exist, right?

tools/mage/build_namespace.go
austinbyers: Saves some time, I'm fine with removing it. We can re-add it if someone needs it

tools/mage/setup.go
austinbyers: > What about the rest - swagger, terraform, goimports etc It seems we will still do the setup even if we are running in CI and files already exist, right?

Nothing is installed if it already exists in CI. Swagger/Terraform/Goimports/Golangci-lint are individual binaries which we execute to get their version. If the version already exists and matches the constant in "setup.go", we skip installing it. That check works both in and out of CI. Example:

```go
func installTerraform(uname string) error {
	if output, err := sh.Output(terraformPath, "-version"); err == nil && strings.Contains(output, terraformVersion) {
		logger.Infof("setup: %s v%s is already installed", terraformPath, terraformVersion)		
		return nil
         }
}
```

There just isn't a similar "generic" check for python and node to see if the dependencies are up to date without running the install commands , which is why those two commands have to explicitly check if they're running in CI

austinbyers: Here's the full output of the `mage setup` command in CI for this PR:

```
00:57:50	INFO	setup: skipping npm install
00:57:50	INFO	    √ npm install finished (1/6)
00:57:50	INFO	setup: download go modules...
00:57:50	INFO	setup: skipping pip install
00:57:50	INFO	    √ pip install finished (2/6)
00:57:50	INFO	setup: .setup/swagger v0.23.0 is already installed
00:57:50	INFO	    √ download go-swagger finished (3/6)
00:57:50	INFO	setup: .setup/terraform v0.12.24 is already installed
00:57:50	INFO	    √ download terraform finished (4/6)
00:57:50	INFO	setup: .setup/golangci-lint v1.23.6 is already installed
00:57:50	INFO	    √ download golangci-lint finished (5/6)
go: found golang.org/x/tools/cmd/goimports in golang.org/x/tools v0.0.0-20200407144507-5fc56a9a2104
00:57:51	INFO	    √ download go modules finished (6/6)
```

As you can see, it finishes in 1 second and doesn't actually do anything

Issue #443, opened 2020-03-18
Component Tag for Cost Estimates


jacknagz:
### Describe the ideal solution

In addition to the Application tag we currently use to group Panther infrastructure from the rest, we should add one for Component. Categories include: "Cloud Security", "Core", "Log Analysis", and "Web".

### Describe your use cases

Enable more. accurate cost estimations.

### How frequently would you use such a feature

Weekly/Monthly

### Describe alternatives you have considered

None, there's no easy way to get this breakdown.

### References

None

### Additional context

None

PR #698, opened 2020-04-13
[deployments] rename alb and sg

jacknagz:
## Background

To better identify Panther infra, we should make sure everything is prefixed with `panther-`. While analyzing the ALB logs, I noticed two components were not prefixed. 

## Changes

- Prefix the ALB and the SG

## Testing

- None


3nvi:
That naming must be related to template nesting. I'm sure that I've added `panther-XXX` in the past, but perhaps we thought it was un-needed for one reason or another. 

In any case, I agree with Austin that we need to test it first. I'm not 100% sure if the resource would be deleted, since the resource name remains the same, but we should check it

austinbyers:
I tested this against an existing deployment (of 7fa37cc2). The security group name can be safely updated in place, but the load balancer cannot:

Screen Shot 2020-04-14 at 11 08 50 AM

We could work around this by manually removing/detaching the target groups, but then this will definitely destroy the existing load balancer. Unfortunately, I don't think changing the alb name is worth the developer and customer effort

3nvi:
> I tested this against an existing deployment (of [7fa37cc](https://github.com/panther-labs/panther/commit/7fa37cc2bba50d250a68e3970b4a214e4c294145)). The security group name can be safely updated in place, but the load balancer cannot:
> 
> Screen Shot 2020-04-14 at 11 08 50 AM
> 
> We could work around this by manually removing/detaching the target groups, but then this will definitely destroy the existing load balancer. Unfortunately, I don't think changing the alb name is worth the developer and customer effort

Even updating one (sec group) is better than nothing.

Since we are there, may I recommend to add proper panther [tags](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-loadbalancer.html#cfn-elasticloadbalancingv2-loadbalancer-tags) to the resources? I don't know if "search by tags" is available, but it's still helpful for the user to understand that the "web" resource is associated to Panther by looking at the tags.

austinbyers:
> Even updating one (sec group) is better than nothing.

Agreed, I'll just update the PR with that change and revert the other.

> Since we are there, may I recommend to add proper panther tags to the resources?

Tags are added automatically by CloudFormation when you deploy. We had "Application: Panther" tags to all resources as well as "Stack: (panther-cfn-stack-name)" See #443

PR #703, opened 2020-04-14
Shorten default CloudWatch log retention to be frugal

rleighton:
## Background
The previous default retention for CW logs was 365 days. This is expensive and likely unneeded for most users.

## Changes
- Change default CW retention to 30 days

### DIFF
--- a/deployments/panther_config.yml
+++ b/deployments/panther_config.yml
@@ -52,7 +52,7 @@ Monitoring:
   AlarmSnsTopicArn: ''
 
   # Retention period for all Panther CloudWatch log groups.
-  CloudWatchLogRetentionDays: 365
+  CloudWatchLogRetentionDays: 30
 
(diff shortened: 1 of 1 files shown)


rleighton:
> Fine by me! On a note though, I doubt though that this is going to lead to significant cost savings. Storing logs is pretty [cheap](https://aws.amazon.com/cloudwatch/pricing/) compared to putting or querying them. In my personal account storage costs are 0 all the cost comes from putlog, alarms and metrics.

This came from a discussion with Jack about overall pricing and I looked at my bill and it was $64 so far this month. That said, I just expanded the details and 1/2 the price is from Dashboards!  I have Panther deployed in 3 regions and it is 3$ / dashboard. Give that is an anomaly and there is no concern about scaling Dashboard, alarms or metrics I'm going to close this.

PR #705, opened 2020-04-15
Minor refactoring

kostaspap:
## Background

Ran across some minor issues while working on a recent task

## Changes

- Removed Lambda#SQSEvent dependency from S3 Source public API. No need to tie it to SQS specifically since events can come from any source. 
- Defining constant for event delimiter

## Testing

- `mage test:ci`

### DIFF
--- a/internal/log_analysis/log_processor/sources/s3.go
+++ b/internal/log_analysis/log_processor/sources/s3.go
@@ -42,12 +42,12 @@ const (
 	cloudTrailValidationMessage = "CloudTrail validation message."
 )
 
-// ReadNotifications reads incoming messages containing SNS notifications and returns a slice of DataStream items
-func ReadS3Events(events []string) (result []*common.DataStream, err error) {
-	zap.L().Debug("reading data for messages", zap.Int("numMessages", len(events)))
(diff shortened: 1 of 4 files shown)


internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -42,12 +42,12 @@ const (
 	cloudTrailValidationMessage = "CloudTrail validation message."
 )
 
-// ReadSQSMessages reads incoming messages containing SNS notifications and returns a slice of DataStream items
-func ReadSQSMessages(messages []events.SQSMessage) (result []*common.DataStream, err error) {
-	zap.L().Debug("reading data for messages", zap.Int("numMessages", len(messages)))
+// ReadMessages reads incoming messages containing SNS notifications and returns a slice of DataStream items
+func ReadMessages(messages []string) (result []*common.DataStream, err error) {
```
rleighton, line 46: I think this should be ReadSNSMessages, since that is what struct they are being parsed into (and the comment)

kostaspap: Makes sense. Updating

PR #706, opened 2020-04-15
Suricata DNS parser

kostaspap:
## Background

Adding support for Suricata DNS 

## Changes

- Added Suricata DNS parser
- Added unit test. Got samples from their documentation: https://github.com/OISF/suricata/blob/master/doc/userguide/output/eve/eve-json-format.rst#event-type-dns
## Testing

- Unit tests
- Deployed to AWS account


internal/log_analysis/log_processor/parsers/suricatalogs/dns.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (event *DNS) updatePantherFields(p *DNSParser) {
+	event.SetCoreFields(p.LogType(), (*timestamp.RFC3339)(event.Timestamp), event)
+	event.AppendAnyIPAddressPtr(event.SrcIP)
+	event.AppendAnyIPAddressPtr(event.DestIP)
+
+	event.AppendAnyIPAddressPtr(event.DNS.RData)
+	event.AppendAnyDomainNamePtrs(event.DNS.Rrname)
+
+	for _, answer := range event.DNS.Answers {
```
rleighton, line 138: Hmmm. No sure this is precise enough. Can you add to the test 3 more record types: MX, PTR and TEXT. Those should make it clear if this extraction needs some qualifications.

kostaspap: In those cases the Rdata will not be IP correct (even in CNAME it will not be) - I am just relying that the `AppendAnyIPAddressPtr` will only add the provided argument if it is an IPv4 or IPv6. I could add a check to add the addresses only in case of `A` or `AAAA` record, just didn't perform this check it to simplify the logic. I could add a comment about this or add the check

rleighton: It is the DomainNamePtr I'm worried about. MX, PTR and TEXT are very important record types and I think we should have at least one example of each in the tests anyway. The examples should tell you if you can use Rrname as domain name unconditionally.

kostaspap: Ok, I see what you mean. Actually good catch: the PTR records will include by definition IP since they are there for reverse lookups so I need to exclude them. MX and TEXT will however include a domain, also by definition. 

I am thinking to follow whitelist approach: Add the domains of A, AAAA, MX, TEXT and just drop everything else. Add the PTR as an IP.

internal/log_analysis/log_processor/parsers/suricatalogs/dns.go
```diff
@@ -0,0 +1,160 @@
...
+		switch aws.StringValue(answer.Rrtype) {
+		case "A", "AAAA":
+			event.AppendAnyIPAddressPtr(answer.Rdata)
+			event.AppendAnyDomainNamePtrs(answer.Rrname)
+		case "CNAME", "TXT", "MX":
```
rleighton, line 144: For CNAME, PTR and MX (but not TXT) we also want answer.Rdata for AppendAnyDomainNamePtrs(), right?

kostaspap: They will not necessarily include top level domains, but they will be domains names. You are right, i'll add them.

kostaspap: Added

PR #714, opened 2020-04-15
Add note about rbac permission changes

austinbyers:
## Background

In Panther enterprise, changing a user's role / role permissions will not take effect until the next time the user logs in. This is because their [Cognito groups are encoded in a token sent to AppSync](https://docs.aws.amazon.com/appsync/latest/devguide/security.html#amazon-cognito-user-pools-authorization) and so persist for the duration of their session.

> When using Amazon Cognito User Pools, you can create groups that users belong to. This information is encoded in a JWT token that your application sends to AWS AppSync in an authorization header when sending GraphQL operations.

## Changes

Add a note so enterprise users understand when permission changes will take effect


3nvi:
@austinbyers It was also mentioned [on the form itself](https://github.com/panther-labs/panther-enterprise/blob/0541e38fac9213e9cf92f554122e343f4957bede/web/src/components/forms/UserForm/UserForm.tsx#L109-L112)

PR #716, opened 2020-04-16
Documentation link in UI should lead to an appropriate docs version

### LINKED ISSUE #665: Documentation should link to specific version
### Description

Currently in the web app, the documentation links to https://runpanther.io/docs, which will redirect to whichever is the "main" version.

Instead, the documentation should link to the version relevant to that specific deployment: `https://docs.runpanther.io/v/v1.0`

### Acceptance Criteria

- Clicking on "documentation" from the Panther web app directs to a specific version of documentation

### Implementation
We already report the "version" of Panther in the general setting from `git describe --tags` - the version number included in that can be used for this purpose.

Or we can hardcode a version number somewhere if we have to

alexmylonas:
## Background

Documentation link in UI right now leads to the latest master docs. We are changing this, so if that if you running an exact version, this link should guide you to the appropriate docs.

Some examples: 
Running `v1.0.1 `will guide to https://docs.runpanther.io/v/v1.0.1
Running `v1.0` will guide to https://docs.runpanther.io/v/v1.0
Running `v1.0.1-xxx` will guide you to https://docs.runpanther.io (master)
**Caveat**: Running `v1.0-xxx` will guide you again to https://docs.runpanther.io (master)

Closes #665
## Changes

- Added a function that check `process.env.PANTHER_VERSION` if it's a proper version
## Testing

- Locally with using above environment variable
- Cant be tested on a deployed env easily


alexmylonas:
@austinbyers @jacknagz how do you feel about the caveat mentioned above? An alternative would be to always point users to their "base" version docs.

austinbyers:
I would prefer if running `v1.0.1-xxx` redirects you to `v1.0.1`, i.e. the latest tagged release on which your deployment is based

austinbyers:
The reason is that if someone just makes a small change (e.g. a config file) that they commit locally, we still want to show the docs that are most likely relevant to their deployment

3nvi:
> The reason is that if someone just makes a small change (e.g. a config file) that they commit locally, we still want to show the docs that are most likely relevant to their deployment

I agree. Stripping the `-` is the  simple yet valid solution. No need to redirect to master docs

alexmylonas:
Implemented the suggested fix and merged it!

PR #717, opened 2020-04-16
Conform Panther to latest Pounce API

3nvi:
## Background

The API of pounce is evolving and Panther needs to keep up with that. In the spirit of progressive changes, this PR performs mass-replacements in order to align with the latest Pounce API

**Disclaimer: 85% of the changes are mass replacements**

## Changes

- Bump pounce
- Mass-replace all `is` props with `as`
- Mass-replace all `<Flex>` components to use the new props
- Mass-replace all `<Grid>` components, to use the nifty `SimpleGrid` component
- Use the new `Link` wherever we wanted to render a Link
- Use `AbtractButton` wherever we wanted something to function as a button **but not look like one**
- Update `Linkify` (the ones that creates links from texts like runbook) to use the new `Link` component from pounce

## Testing

- Locally

At the end of the day, the app should be exactly the same. Functionality is not changed in any way. **If the app looks the same, then these changes are correct**


3nvi:
@alexmylonas I'll wait for a review from you before I merge

3nvi:
@alexmylonas Thanks so much! will address both of them

PR #718, opened 2020-04-16
Use worker model to limit max number of parallel go builds

austinbyers:
## Background

#696 started compiling Go binaries in parallel. While faster, this will completely consume all available compute resources, and @kostaspap started noticing his laptop struggling to do other things while building/testing.

This PR introduces a compromise - instead of launching a new goroutine for every unit of work, we launch a goroutine for each logical CPU (-1) and feed a stream of work units. In this model, you have to introduce a "poison pill" into the stream so the workers know when to stop.

## Changes

- `mage build:lambda` and `mage build:tools` will only use `runtime.numCPU() - 1` concurrent goroutines

Since these commands themselves run in parallel in `mage test:ci`, it's still possible to be consuming all cores, but hopefully by reducing the parallelism of both commands the system can keep up now. If not, it's easy to tune the "maxWorkers" parameter as needed.

## Testing

- Temporarily added log messages to see worker start + stop
- Tested with 1 - 30 workers (below, at, and above my system CPU)

### DIFF
--- a/tools/mage/build_namespace.go
+++ b/tools/mage/build_namespace.go
@@ -100,6 +100,13 @@ func (b Build) Lambda() {
 	}
 }
 
+// "go build" in parallel for each Lambda function.
+//
+// If you don't already have all go modules downloaded, this may fail because each goroutine will
+// automatically modify the go.mod/go.sum files which will cause conflicts with itself.
+//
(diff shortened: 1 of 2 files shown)


austinbyers:
See also https://github.com/panther-labs/panther/pull/696#discussion_r407978732

PR #732, opened 2020-04-17
Fix output path for mage build:tools

rleighton:
## Before
Output path for tools executable was deeply nested and confusing:
```
./out/bin/cmd/opstools/s3queue/s3queue/darwin/amd64/s3queue 
```
## After
Now path starts with tool name: 
```
./out/bin/s3queue/darwin/amd64/s3queue
```

## Testing
mage test:ci
mage build:tools


rleighton:
> I'm ok with this, but it's just using the package path for the binary, which is the same thing we do for all `out` artifacts.
> 
> The better fix, in my opinion, is to fix the paths of the tool source. Why `cmd/opstools/requeue/requeue/main.go` instead of `cmd/tools/requeue/main.go`?

We can, but then the bin path will still have /cmd/tools/. Most people expect a flat or shallow bin dir . I don't know that making the source code path itself related to the bin path makes sense for everything.

Issue #708, opened 2020-04-15
Support CloudTrail Insights logs


kostaspap:
### Description

Panther should support CloudTrail Insights logs. 
- General information: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/logging-insights-events-with-cloudtrail.html
- Sample events: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference.html

### Acceptance Criteria

- Panther has a new Parser for CloudTrail Insights events
- Unit tests for the parser
- Classifier uses the new parser
- Appropriate tables are created in Glue.

PR #733, opened 2020-04-17
[awslogs] AWS CloudTrailInsight parser

alxarch:
## Background

Adds parser for AWS CloudTrail Insights records #708

## Changes

- Adds and registers a new parser at `internal/log_analysis/parsers/awslogs` to parse AWS CloudTrail Insights events.

## Testing

- Added unit tests and a test helper to verify multiple log output from parser (each Insight event has 2 records)
- Tested with `mage test:ci`


alxarch:
For some reason I get the following error when trying to deploy:
```
failed to create change set for stack panther-glue: ValidationError: [/Resources/pantherlogsawscloudtrailinsight/Type/TableInput/StorageDescriptor/Columns] 'null' values are not allowed in templates
```
Rings any bells?

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights_test.go
```diff
@@ -0,0 +1,196 @@
...
+	expectMore = append([]SetEventer{expect}, expectMore...)
+	require.Equal(t, len(expectMore), len(results), "Invalid number of pather logs produced by parser")
+	for i, result := range results {
+		expect := expectMore[i]
+		plog := expect.Log()
+		plog.SetEvent(expect)
+		CheckJSONEqual(t, expect.Log(), plog, "Invalid panther log entry %d", i)
+		require.Equal(t, parser.LogType(), *result.PantherLogType)
+	}
+}
```
kostaspap, line 183: There as some practices in this method that I really like - and that we should really adopt in other parts of the code:
1. The use of `t.Helper()` - we should be doing it in other methods we use as part of testing validations
2. The check for invalid number of panther logs has nice output in case of failure
3. The initialization of the parser with `parser.New()`. This is something we should be doing in all of our tests
4. By making the `CheckPantherParser` accept a `parsers.LogParser` argument, we essentially verify that an event parser adheres to the `parsers.LogParser` interface and we can probably get rid of the slightly awkward `// NOTE: guard to ensure interface implementation
var _ parsers.LogParser = (*CloudTrailInsightParser)(nil)` in the parser code. 

However, what I don't particularly like about this code is that it follows different logic for running tests compared to the rest of our parsers. I think we will be better of if we do the following: 
1. Move `CheckPantherParser` method to `testutils` package. 
2. Update the loop in lines 176-182 to use the [EqualPantherLog](https://github.com/panther-labs/panther/blob/9757c067a89873d35818ca1b53d4c7d9f756f4f2/internal/log_analysis/log_processor/parsers/testutil/testutil.go#L33) method. That method already defines a lot of the checks that have to pass for a PantherLog event to be considered valid. 
3. In this (or another PR might be better) updater our existing tests to use this method. 

Let me know what you think. We can even chat offline about this if you want me to clarify something.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// panther fields
+	expectedEventStart.PantherLogType = aws.String("AWS.CloudTrailInsight")
+	expectedEventStart.PantherEventTime = (*timestamp.RFC3339)(&expectedDateStart)
+	expectedEventStart.AppendAnyAWSAccountIds("123456789012")
+	expectedEventEnd.PantherLogType = aws.String("AWS.CloudTrailInsight")
+	expectedEventEnd.PantherEventTime = (*timestamp.RFC3339)(&expectedDateEnd)
+	expectedEventEnd.AppendAnyAWSAccountIds("123456789012")
+
+	CheckPantherParser(t, log, new(CloudTrailInsightParser), expectedEventStart, expectedEventEnd)
```
kostaspap, line 157: We generally avoid using `new()` in the codebase - update to `&CloudTrailParser{}`

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights_test.go
```diff
@@ -0,0 +1,196 @@
...
+func TestCloudTrailInsightLogType(t *testing.T) {
+	parser := &CloudTrailInsightParser{}
+	require.Equal(t, "AWS.CloudTrailInsight", parser.LogType())
+}
+
+type SetEventer interface {
+	Log() *parsers.PantherLog
+}
+
+func CheckPantherParser(t *testing.T, log string, parser parsers.LogParser, expect SetEventer, expectMore ...SetEventer) {
```
kostaspap, line 169: I assume the method could be changed to:
```
func CheckPantherParser(t *testing.T, log string, parser parsers.LogParser, expect ...SetEventer) {
```
Any benefit of the current approach? Is it to force the caller to provide on `SetEventer` argument at least or maybe something else that I don't see

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights_test.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func TestCloudTrailInsightLogType(t *testing.T) {
+	parser := &CloudTrailInsightParser{}
+	require.Equal(t, "AWS.CloudTrailInsight", parser.LogType())
+}
+
+type SetEventer interface {
+	Log() *parsers.PantherLog
+}
```
kostaspap, line 167: Not clear to me, what is the benefit of introducing this interface? Why not just pass around `*parsers.PantherLog`?

internal/log_analysis/log_processor/registry/registry.go
```diff
@@ -77,6 +78,20 @@ var (
...
+		(&gitlablogs.IntegrationsParser{}).LogType(): DefaultLogParser(&gitlablogs.IntegrationsParser{},
+			&gitlablogs.Integrations{}, gitlablogs.IntegrationsDesc),
+		(&gitlablogs.GitParser{}).LogType(): DefaultLogParser(&gitlablogs.GitParser{},
+			&gitlablogs.Git{}, gitlablogs.GitDesc),
+		(&gitlablogs.AuditParser{}).LogType(): DefaultLogParser(&gitlablogs.AuditParser{},
+			&gitlablogs.Audit{}, gitlablogs.AuditDesc),
+		(&gitlablogs.ExceptionsParser{}).LogType(): DefaultLogParser(&gitlablogs.ExceptionsParser{},
+			&gitlablogs.Exceptions{}, gitlablogs.ExceptionsDesc),
+		(&awslogs.CloudTrailInsightParser{}).LogType(): DefaultLogParser(&awslogs.CloudTrailInsightParser{},
+			&awslogs.CloudTrailInsightParser{}, awslogs.CloudTrailInsightDesc),
```
kostaspap, line 94: Should be: 
```suggestion
		(&awslogs.CloudTrailInsightParser{}).LogType(): DefaultLogParser(&awslogs.CloudTrailInsightParser{},
			&awslogs.CloudTrailInsight{}, awslogs.CloudTrailInsightDesc),
```
That's the reason the generation of the table is failing

alxarch: Nice catch, fixing

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights.go
```diff
@@ -0,0 +1,122 @@
...
+	// NOTE: added to end of struct to allow expansion later
+	AWSPantherLog
+}
+
+// nolint:lll
+type InsightDetails struct {
+	State          *string         `json:"state" validate:"required" description:" Shows whether the event represents the start or end of the insight (the start or end of unusual activity). Values are Start or End."`
+	EventSource    *string         `json:"eventSource" validate:"required" description:"The AWS API for which unusual activity was detected."`
+	EventName      *string         `json:"eventName" validate:"required" description:"The AWS API for which unusual activity was detected."`
+	InsightType    *string         `json:"insightType" validate:"required,eq=ApiCallRateInsight" description:"The type of Insights event. Value is ApiCallRateInsight. "`
```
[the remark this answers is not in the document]
alxarch: Does this affect the Glue tables somehow? I can remove redundant required if not

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights_test.go
alxarch: This was my intention but I hit errors with the type system. The `EqualPantherLog` method will need some refactoring itself (and also handle the new LogParser interface)

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights_test.go
alxarch: Yes it's implemented like this to force the passing of at least one argument

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights_test.go
alxarch: I wanted the parsing to be done inside the test, The name is wrong as initally it was supposed to ensure the `SetEvent` method. Will look into it.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights.go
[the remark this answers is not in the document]
kostaspap: Actually, I looked again in the codebase. It seems that we are checking for the existence of the "required" field when generating documentation (to mark a field as required or not) and also when creating Glue Columns. Good catch, better to leave it there indeed.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights_test.go
alxarch: One question I have is if the JSON-based test is sufficient for the usage scenario of *parsers.PantherLog. Is it Good Enough (TM) to just check the JSON output for equality. I used this because `reflect.Equal` does not seem to work with pointer fields.

kostaspap: Testing for JSON equality is a good way to test for equality of two PantherLog events, this is what we are doing in the [testutils](https://github.com/panther-labs/panther/blob/9757c067a89873d35818ca1b53d4c7d9f756f4f2/internal/log_analysis/log_processor/parsers/testutil/testutil.go#L52-L58) as well. 
It might be counter-intuitive, but in the context of the Panther parsers, this is what we want in the end. 

It might have been more "clean" to verify the equality of the structures (if the structures are equal their JSON output should be equal) but it was complex to get that done so we opted for JSON Eq. If we figure out a better way to do that in the future should be easy to implement since it would require changes only in one place (in the testutils)

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights_test.go
alxarch: The issue seems to be that `SetEvent` needs the internal event not the *parsers.PantherLog. That was the reason I used an ad-hoc interface. So that SetEvent needn't be called before checking and thus avoid separate helper for each log type. TBH I think that if we stick to the struct/tags-based approach for logs then It's worth the effort to add a `panther` tag to the structs and mark any fields that hold `Timestamp` `IP`, `SHA1` etc. That would solve this circular relation that exists between `parsers.PantherLog` and any event struct.

kostaspap: > TBH I think that if we stick to the struct/tags-based approach for logs then It's worth the effort to add a `panther` tag to the structs and mark any fields that hold `Timestamp` `IP`, `SHA1` etc. That would solve this circular relation that exists between `parsers.PantherLog` and any event struct.

Ok, that's an interesting approach. The first thing that comes to mind is that such a design would have to account for scenarios like this one: 
https://github.com/panther-labs/panther/blob/89b6dd7cfdfc97afa9007ecac1738057b3f37841/internal/log_analysis/log_processor/parsers/fluentdsyslogs/rfc3164.go#L81-L85

I like the idea, we should definitely explore it. The way to move forward as I see it: 
1. Go for the minimum changes required for this PR. 
2. We can chat offline about the details of what you had in mind about the event tagging. After that maybe you can come up with a small PoC/change in a single parser to see how this could look like and well it performs and if we like the approach we can apply this to all parsers.

alxarch: I reviewed the code and better understood the relationship between PathnerLog and the inne event. Will use the existing testutil. It was a bit confusing because the embedded struct is used in the test but the embedded struct cannot access the event struct.

web/src/constants.ts
```diff
@@ -81,6 +81,9 @@ export const LOG_TYPES = [
   'AWS.ALB',
   'AWS.AuroraMySQLAudit',
   'AWS.CloudTrail',
+  'AWS.CloudTrailInsight',
+  'Fluentd.Syslog3164',
+  'Fluentd.Syslog5424',
```
kostaspap, line 86: These two seem to have slipped in by mistake? They are already defined below.

alxarch: It figures, somehow the merge had duplicate values got GitLab logs also

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights.go
```diff
@@ -0,0 +1,122 @@
...
+
+// nolint:lll
+type InsightAverage struct {
+	Average *float64 `json:"average,omitempty" description:"Average value for the insight metric"`
+}
+
+type CloudTrailInsightParser struct{}
+
+// NOTE: guard to ensure interface implementation
+var _ parsers.LogParser = (*CloudTrailInsightParser)(nil)
```
kostaspap, line 83: I am thinking probably we don't need this? The unit tests verify for this. Thoughts?

alxarch: It's a good practice I've picked up. As soon as you know a struct should implement an Interface you can have it pop up as a mistake if it doesn't and also writes down the info that the struct implements the interface comewhere near the struct definition

alxarch: And it's compiled away by go compiler so it doesn't affect the binary

kostaspap: I have to say the aesthetics of it bother me, but I know it's kinda standard Go practice. I usually prefer to find ways to avoid it, but the fact that it is unit tested "only" seems a bit weak. I imagine we can get rid of it if we end up enforcing this check in different part of the application (e.g. the registry ;) )

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_insights_test.go
```diff
@@ -0,0 +1,176 @@
...
+// 	ex, err := json.Marshal(expect)
+// 	if err != nil {
+// 		t.Fatalf("Failed to serialize to JSON %q", err)
+// 	}
+// 	ac, err := json.Marshal(actual)
+// 	if err != nil {
+// 		t.Fatalf("Failed to serialize to JSON %q", err)
+// 	}
+// 	require.JSONEq(t, string(ex), string(ac), msgAndArgs...)
+// }
```
kostaspap, line 176: Leftovers?

3nvi:
@all-contributors please add @alxarch for code and security

PR #737, opened 2020-04-18
Configuring AppSync logging

kostaspap:
## Background

I noticed that our current AppSync had logging on by default.  
I made AppSync logging configurable - it will be enabled only when Debug is set to true. Even in debug mode though we will exclude HTTP header logging to avoid leaking authentication tokens in the logs. 

## Changes

- Made AppSync logging configurable. 
- Updated Go dependencies.

## Testing

- Deployed in my account both scenarios. 
when Debug is set to *True* the AppSync configuration looks like this: 
Screen Shot 2020-04-18 at 4 46 42 PM

When Debug is set to *False* the AppSync configuration looks like this:
Screen Shot 2020-04-18 at 4 45 56 PM


deployments/bootstrap.yml
```diff
@@ -698,9 +704,12 @@ Resources:
         AwsRegion: !Ref AWS::Region
         UserPoolId: !Ref UserPool
         DefaultAction: ALLOW
-      LogConfig:
-        CloudWatchLogsRoleArn: !GetAtt AppsyncServiceRole.Arn
-        FieldLogLevel: ALL
+      LogConfig: !If
+        - EnableDebug
+        - ExcludeVerboseContent: TRUE
```
austinbyers, line 709: Maybe logging can always be enabled, but this setting is TRUE/FALSE depending on the DEBUG setting? Do you have an example of sensitive data you are worried about?

Other ideas we can eventually do here:
* CloudWatch log groups aren't encrypted, because it's not supported by CloudFormation, but we could add encryption manually. That would help alleviate concerns about sensitive data winding up in log groups
* We've talked about having multiple log levels (DEBUG, INFO, WARN, etc) instead of just Debug on/off

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/stretchr/testify v1.5.1
 	github.com/tidwall/gjson v1.6.0
 	go.uber.org/zap v1.14.1
+	golang.org/x/tools v0.0.0-20200407144507-5fc56a9a2104 // indirect
```
austinbyers, line 28: Did you run a specific command to change these files? A standard `mage clean setup deploy` shouldn't typically change these, nor do we necessarily need to. I don't fully understand when or why these files update

3nvi:
@kostaspap I'm dependent on a log of these CW logs, since it's my only means of doing debugging. Do you have an example on how these will change  after your PR?

go.mod
kostaspap: These got updated when I did the `mage setup` command.

deployments/bootstrap.yml
kostaspap: Basically when logging is enabled (even with the `ExcludeVerboseContent: TRUE`) AppSync logs the full body of the response. Depending on the endpoint that will the full rule body or actual log events. So even the "basic" logging actually emulated the "Debug" logging of the rest of the application...

kostaspap:
> @kostaspap I'm dependent on a log of these CW logs, since it's my only means of doing debugging. Do you have an example on how these will change after your PR?

@3nvi Sure. To get AppSync logging of requests/responses you will have to set the debug to true: https://github.com/panther-labs/panther/blob/89b6dd7cfdfc97afa9007ecac1738057b3f37841/deployments/panther_config.yml#L57-L58

This will still not contain HTTP Headers. If you ever need them for some debugging purposes you can enable them in AppSync settings, though I can imagine it is rare that this will be needed.

deployments/bootstrap.yml
kostaspap: Even if we end up encrypted the CW logs, I would prefer if customer data don't end up in there. S3 gives us more fine-grained control over the data, CW Logs doesn't. 
Some support engineer might require access to CW Logs to debug an issue, this shouldn't grant him automatically access to all the data in the data lake.

PR #738, opened 2020-04-19
Increase delay before deleting bucket in teardown to avoid errors

rleighton:
## Background
S3 delete bucket calls were failing during teardown, requiring a re-run.

## Changes
Increase delay to 2sec to help avoid this irritation.

## Testing
mage test:ci
mage teardown


austinbyers:
I'll be working on a proper fix for that this week, unless you get to it first

PR #739, opened 2020-04-19
Tag all Panther aws objects with the version of the deployed sw

rleighton:
## Background
It is difficult to be 100% sure what software is deployed into an account, especially if the circumstances  of the deployment are uncertain.

This PR tags all objects with the git version. When helping a customer troubleshoot we can be certain of the deployed sw and also find any inconsistencies.

## Changes
- Tag all sources with the git version
- Print git version on deployment

## Testing
mage test:ci
mage deploy, confirmed version is set


tools/mage/deploy_template.go
```diff
@@ -180,6 +184,10 @@ func createChangeSet(
 				Key:   aws.String("Application"),
 				Value: aws.String("Panther"),
 			},
+			{
+				Key:   aws.String("PantherVersion"),
+				Value: &gitVersion,
```
austinbyers, line 189: This has a major downside: every developer change will force every CFN resource to be updated. I haven't tested it yet, but it seems this might significantly slow down deploys when you're making only small changes because every stack has to be re-updated.

We already store the version in the general settings page which we can lookup when troubleshooting, why do we need it on every CFN resource as well?

rleighton: Good point. I updated the code to strip the hash and just do the version in the tag (which is all that is important). This way it will not affect devs (this tag will not change per commit).
image

The reason I want it in all objects is when troubleshooting with a customer using the Console, I do not want to have them go to the UI. Also, since the UI does this itself in an unrelated code path, if there are build issues there is a small chance this not the same (bug in UI).   Since this is basically free, I do not see the downside and it gives us really good tracking data to use for validation / troubleshooting.

kostaspap:
I am not sure this is helpful in any way? The customers can see the deployed version in their General Settings. Since panther is in a monorepo, all the resources deployed will always be tagged with the same version. Even if e.g. one Lambda is modified locally, it will be tagged with the same version as all other resources.

tools/mage/deploy_template.go
kostaspap: The problem with this is that it will be misleading if someone deployed from a non-tagged commit... and we will have no way to find out they did.

austinbyers:
Have to agree with Kostas, this doesn't really hurt anything but it also doesn't help. If all things are tagged with the same version, and the version is visible in the UI, what's the added value?

> The reason I want it in all objects is when troubleshooting with a customer using the Console, I do not want to have them go to the UI.

If that's the case, you can see the version in the settings Dynamo table through the AWS console.

tools/mage/deploy_template.go
rleighton: Then it is even more important. I can make an empty string say "untagged"

rleighton:
> Have to agree with Kostas, this doesn't really hurt anything but it also doesn't help. If all things are tagged with the same version, and the version is visible in the UI, what's the added value?
> 
> > The reason I want it in all objects is when troubleshooting with a customer using the Console, I do not want to have them go to the UI.
> 
> If that's the case, you can see the version in the settings Dynamo table through the AWS console.

Where is it in ddb? I looked and could not find in it.

That said, in an emergency, while doing troubleshooting I cannot overstate how important concise and rapid decision making is important. Even if this is in ddb and the UI, why make hard? If I'm troubleshooting , having a tag tells on the page I'm looking at about the object (e.g., SQS). It also allows customer to screenshot easily. This is zero cost and the kind of info that tags are intended to be used for.

austinbyers:
> Where is it in ddb? I looked and could not find in it.

Oh, you're right, I forgot we bundle the version in the .env file in the web image, it's not in the settings table. 

In that case, I can see the value of showing the version somewhere in the infra itself. Then we should show exactly the same version we do in the web UI, no need to truncate it. (Otherwise, the discrepancy would be confusing). I confirmed the `git describe --tags` only changes on an actual commit, so rapid developer iteration won't be affected by passing the version through.

Sorry for the back and forth, but it was a good discussion! Approved assuming we'll use the full version

tools/mage/deploy_template.go
```diff
@@ -164,6 +168,12 @@ func createChangeSet(
 		})
 	}
 
+	// add version tag to all objects ("untagged" if not set)
+	pantherVersion := gitVersion
+	if pantherVersion == "" {
+		pantherVersion = "untagged"
+	}
```
austinbyers, line 175: This actually isn't possible right now, I don't think. I haven't tested it yet, but I think the web docker image will fail to build if the `git describe` call fails

This is fine to keep, just an FYI

PR #740, opened 2020-04-19
Tune Syslog, Pull IPs from Syslog Messages

jacknagz:
## Background

It's common for IPs to land in the authentication information generated from the SSH daemon and collected with syslog. We want to search this easily with the standardized data fields, so this change extracts them out.

```
Accepted publickey for ubuntu from 8.103.222.11 port 51454 ssh2: RSA SHA256:...
```

## Changes

- Remove the validate:"required" tag on the FluentdSyslog.RFC3164 struct
- Pull out IPs from log messages using a regexp
- Enable IP extraction on all syslog log messages
- Update tests

## Testing

- Unit testing

### DIFF
--- a/internal/log_analysis/log_processor/parsers/pantherlog.go
+++ b/internal/log_analysis/log_processor/parsers/pantherlog.go
@@ -147,17 +147,12 @@ func (pl *PantherLog) AppendAnyIPAddressInFieldPtr(value *string) bool {
 
 // AppendAnyIPAddressInField extracts all IPs from the value using a regexp
 func (pl *PantherLog) AppendAnyIPAddressInField(value string) bool {
-	matchedIPs := ipRegex.FindAll([]byte(value), -1)
-	if len(matchedIPs) > 0 {
-		if pl.PantherAnyIPAddresses == nil { // lazy create
-			pl.PantherAnyIPAddresses = NewPantherAnyString()
-		}
-		for _, match := range matchedIPs {
-			AppendAnyString(pl.PantherAnyIPAddresses, string(match))
-		}
-		return true
+	result := false
+	matchedIPs := ipRegex.FindAllString(value, -1)
+	for _, match := range matchedIPs{
+		result = result || pl.AppendAnyIPAddress(match)
 	}
-	return false
+	return result
 }
 
 func (pl *PantherLog) AppendAnyIPAddress(value string) bool {
--- a/internal/log_analysis/log_processor/parsers/pantherlog.go
+++ b/internal/log_analysis/log_processor/parsers/pantherlog.go
@@ -33,7 +33,7 @@ const (
 )
 
 var (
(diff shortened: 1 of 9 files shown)


internal/log_analysis/log_processor/parsers/pantherlog.go
```diff
@@ -32,6 +33,7 @@ const (
 )
 
 var (
+	ipRegex, _ = regexp.Compile(`(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])`)
```
kostaspap, line 36: nits:
* Rename to ipv4Regex. In general in this file when we refer to IP we refer to a field that could be either IPv4 or IPv6
* Change to `ipv4Regex = regexp.MustCompile('.....')`

kostaspap:
```suggestion
	ipv4Regex = regexp.MustCompile(`(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])`)
```

internal/log_analysis/log_processor/parsers/pantherlog.go
```diff
@@ -126,14 +128,38 @@ func (pl *PantherLog) SetCoreFields(logType string, eventTime *timestamp.RFC3339
...
+		if pl.PantherAnyIPAddresses == nil { // lazy create
+			pl.PantherAnyIPAddresses = NewPantherAnyString()
+		}
+		for _, match := range matchedIPs {
+			AppendAnyString(pl.PantherAnyIPAddresses, string(match))
+		}
+		return true
+	}
+	return false
+}
```
kostaspap, line 161: nits: 
* Replace `FindAll` with `FindAllString` to avoid casting
* Can make the code more compact and avoid writing some logic that already exists by changing to something like below:
```suggestion
func (pl *PantherLog) AppendAnyIPAddressInField(value string) bool {
	result := false
	matchedIPs := ipRegex.FindAllString(value, -1)
	for _, match := range matchedIPs{
		result = result || pl.AppendAnyIPAddress(match)
	}
	return result
}
`
```

internal/log_analysis/log_processor/parsers/pantherlog.go
```diff
@@ -126,14 +128,33 @@ func (pl *PantherLog) SetCoreFields(logType string, eventTime *timestamp.RFC3339
...
+	}
+	return pl.AppendAnyIPAddressInField(*value)
+}
+
+// AppendAnyIPAddressInField extracts all IPs from the value using a regexp
+func (pl *PantherLog) AppendAnyIPAddressInField(value string) bool {
+	result := false
+	matchedIPs := ipv4Regex.FindAllString(value, -1)
+	for _, match := range matchedIPs {
+		result = result || pl.AppendAnyIPAddress(match)
```
rleighton, line 153: Hmmm, I _think_ go does "short cutting" and if you have more than 1 ip in a string you will get only the first.
Easy to test, in TestAppendAnyIPsInField(t *testing.T) can you add a string with 2 or 3 distinct ips and keep sure it works?

PR #741, opened 2020-04-20
[log_processor] Refactor `LogParser` to return errors

alxarch:
## Background

Returning errors from `LogParser.Parse()` will allow easier testing for parsers (on non valid log input) and centralizes parse error handling in the classifier, decoupling parsers from zap logger 

## Changes

- Change `LogParser#Parse` interface method signature from `Parse(log string) []*PantherLog` to `Parse(log string) ([]*PantherLog, error)`
- Change all implemented parsers to the new signature, returning the errors instead of logging
- Log the error in `classification.Classifier.safeLogParse`
- Modify `parsers/testutil/EqualPantherLog` to check for errors
- Update all tests to the new signature

## Testing

- Tested with `mage test:ci`


internal/log_analysis/log_processor/classification/classifier_test.go
```diff
@@ -38,13 +38,13 @@ func (m *mockParser) New() parsers.LogParser {
 	return m // pass through (not stateful)
 }
 
-func (m *mockParser) Parse(log string) []*parsers.PantherLog {
+func (m *mockParser) Parse(log string) ([]*parsers.PantherLog, error) {
 	args := m.Called(log)
 	result := args.Get(0)
 	if result == nil {
-		return nil
+		return nil, nil
```
kostaspap, line 45: Method should be returning the "real" error back. Something like; 

```
func (m *mockParser) Parse(log string) ([]*parsers.PantherLog, error) {
 args := m.Called(log)
 result := args.Get(0)
 err := args.Error(1)
 if result == nil {
  return nil, err
 }
 return result.([]*parsers.PantherLog), err

```

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -87,28 +88,24 @@ func (p *ALBParser) New() parsers.LogParser {
...
 	if err != nil {
-		zap.L().Debug("failed to parse the log as csv")
-		return nil
+		return nil, err
 	}
 
 	if len(record) < albMinNumberOfColumns {
-		zap.L().Debug("failed to parse the log as csv (wrong number of columns)")
-		return nil
+		return nil, err
```
kostaspap, line 98: `err` will be nil. You need to return an error with custom message.

internal/log_analysis/log_processor/parsers/awslogs/aurora_mysql_audit.go
```diff
@@ -59,28 +57,28 @@ type AuroraMySQLAuditParser struct {
...
 	if err != nil {
-		zap.L().Debug("failed to parse the log as csv")
-		return nil
+		return nil, err
 	}
 
 	if len(record) < auroraMySQLAuditMinNumberOfColumns {
-		zap.L().Debug("failed to parse the log as csv (wrong number of columns)")
-		return nil
+		return nil, err
```
kostaspap, line 76: `err` will be nil, we need to populate it.

internal/log_analysis/log_processor/parsers/awslogs/vpc_flow_test.go
```diff
@@ -143,6 +144,7 @@ func TestVpcFlowLogType(t *testing.T) {
 func checkVPCFlowLog(t *testing.T, header, log string, expectedEvent *VPCFlow) {
 	expectedEvent.SetEvent(expectedEvent)
 	parser := (&VPCFlowParser{}).New() // important to call New() to initialize reader
-	parser.Parse(header)
-	testutil.EqualPantherLog(t, expectedEvent.Log(), parser.Parse(log))
+	_, _ = parser.Parse(header)
```
kostaspap, line 147: Probably better to check the result here (even though the code previously was not doing it). 
We should verify that the parser returns an empty array and no error.

internal/log_analysis/log_processor/parsers/nginxlogs/access.go
```diff
@@ -56,46 +55,46 @@ type Access struct {
...
+var _ parsers.LogParser = (*AccessParser)(nil)
+
 func (p *AccessParser) New() parsers.LogParser {
 	return &AccessParser{}
 }
 
 // Parse returns the parsed events or nil if parsing failed
-func (p *AccessParser) Parse(log string) []*parsers.PantherLog {
+func (p *AccessParser) Parse(log string) ([]*parsers.PantherLog, error) {
 	reader := csv.NewReader(strings.NewReader(log))
```
kostaspap, line 66: Huh, this seems like a leftover. We should be using the streaming CSV parser (like we do in AWS ALB logs). If it's not too much of a hustle maybe you can do in this PR or we can do in a subsequent PR.

web/src/constants.ts
```diff
@@ -103,6 +103,12 @@ export const LOG_TYPES = [
   'Suricata.DNS',
   'Syslog.RFC3164',
   'Syslog.RFC5424',
+  'GitLab.Rails',
+  'GitLab.API',
+  'GitLab.Audit',
+  'GitLab.Integrations',
+  'GitLab.Exceptions',
+  'GitLab.Git',
```
kostaspap, line 111: This PR doesn't add new parsers. Merge conflict?

internal/log_analysis/log_processor/destinations/s3_test.go
```diff
@@ -67,13 +67,15 @@ type mockParser struct {
...
 
-func (m *mockParser) Parse(log string) []*parsers.PantherLog {
+var _ parsers.LogParser = (*mockParser)(nil)
+
+func (m *mockParser) Parse(log string) ([]*parsers.PantherLog, error) {
 	args := m.Called(log)
 	result := args.Get(0)
 	if result == nil {
-		return nil
+		return nil, nil
```
kostaspap, line 76: We should return the error returned by the parser. 
Should be changed to: 
```
args := m.Called(log)
result := args.Get(0)
err := args.Error(1)
if result == nil {
  return nil, err
}
return result.([]*parsers.PantherLog),err

```

web/src/constants.ts
alxarch: Yep, I have a conflict in my master branch. Fixed

internal/log_analysis/log_processor/destinations/s3_test.go
alxarch: Fixed

internal/log_analysis/log_processor/parsers/nginxlogs/access.go
alxarch: I'd prefer a separate PR unless it's already checnged in upstream and the merge reverted it

kostaspap: Fine by me. No, this looks like a miss upstream.

Issue #136, opened 2020-02-07
mage deploy: implement our own cfn packaging


austinbyers:
**Is your feature request related to a problem? Please describe.**
We shell out to `aws cloudformation package` when running `mage deploy`. The AWS CLI unfortunately has a bug where the generated template doesn't always reference the correct region, so we have to post-process the generated templates.

We also have to revert timestamps on generated Go binaries if they haven't changed so that the `package` command treats them as unchanged. Identical binaries with different timestamps are considered different by the AWS CLI.

**Describe the solution you'd like**
We can implement the same functionality as `aws cloudformation package` in the Go code instead of shelling out. Besides avoiding the two bugs mentioned above, this would also:

-eliminate the AWS CLI completely from the Panther workflow
-probably allow faster deploys with parallel packaging

**Describe alternatives you've considered**
There is no direct equivalent of `aws cloudformation package` in the Go SDK

PR #742, opened 2020-04-20
Refactor cfn deploy process

austinbyers:
## Background

A number of substantial improvements to the way we handle CloudFormation templates and deployment. The end result is fewer trips to S3 (which leads to slightly faster deployments), significantly easier debugging, and reduction in tech debt (no more CloudFormation hacks!)

The templates themselves do not change, just how we process them. (One small exception: a few ints had to be converted to strings to avoid being marshaled in scientific notation, very strange)

## Changes

- Parse CloudFormation correctly instead of doing string find+replace. We do increasingly more CFN introspection (doc generation, alarms, swagger embedding, etc), so string/regex hacks are going to be increasingly fragile. The problem is that Go YAML parsers ignore the CFN short-form functions (`!Sub`, `!If`, etc) - the trick I use is [cfn-flip](https://github.com/awslabs/aws-cfn-template-flip) to convert to json and then parse the json.
    - Swagger embedding is now based on the YAML structure instead of string patterns
    - Another advantage of parsing the templates is to strip away comments/spaces before deploying, which makes the resulting template smaller and therefore more likely to be uploaded directly instead of going to S3
    - The `<cfndoc>` tags are still implemented as comments for now - this is the only place left that uses string regexes to parse templates. We can leave it that way, or we can promote them to "real" YAML fields so that they can be parsed by the yaml parsers. If we opt for the latter, I'll make that change in a subsequent PR (the doc fields would be stripped away before deploying)
- Implement asset packaging ourselves (closes: #136)
    - Lambda code source includes S3 versions, which means CFN updates are slightly faster when the compiled binary hasn't changed
    - We no longer have to revert binary timestamps nor fix S3 URLs (both of which were hacks to get around aws CLI limitations)
    - The aws CLI is no longer required to deploy Panther
    - The final generated template is always saved locally for easier debugging and only uploaded to S3 if absolutely necessary (the template is too large and doesn't already exist on S3)
- ROLLBACK_COMPLETE and other bad stack states are automatically detected and corrected, if possible (closes: #143)
- When a deploy fails, the stack events are listed to find the actual error message (#143). We already do something similar for `mage teardown`, now we have it for deploy as well.
    - All ongoing stacks are allowed to finish so error messages can be collected together

## Testing

- `mage test:ci`
- Many deploys, some intentionally broken, to my dev account
- Fresh deploy as well as upgrade existing deployment
- Integration tests
- Teardown

Example of a deployment with failures in multiple stacks:

```
20:54:05	INFO	    √ panther-web finished (2/10)
20:54:08	INFO	    √ panther-glue finished (3/10)
20:54:29	ERROR	stack panther-core: AWS::SQS::Queue AlertDLQ UPDATE_FAILED: Encountered non numeric value for property MessageRetentionPeriod
20:54:29	ERROR	stack panther-core: AWS::SQS::Queue LayerDLQ UPDATE_FAILED: Encountered non numeric value for property MessageRetentionPeriod
20:54:29	ERROR	    X panther-core failed (4/10): UPDATE_ROLLBACK_COMPLETE
20:54:55	ERROR	stack panther-appsync: AWS::AppSync::DataSource AlertsAPILambdaDataSource UPDATE_FAILED: Property validation failure: [Value for property {/Type} does not match pattern {(AWS_LAMBDA|AMAZON_DYNAMODB|AMAZON_ELASTICSEARCH|NONE|HTTP|RELATIONAL_DATABASE)}]
20:54:55	ERROR	    X panther-appsync failed (5/10): UPDATE_ROLLBACK_COMPLETE
20:54:55	ERROR	stack panther-cloud-security: AWS::SQS::Queue MyNewQueue CREATE_FAILED: panther-alert-processor-queue already exists in stack arn:aws:cloudformation:us-west-2:1234:stack/panther-cloud-security/3d603c50-7f5b-11ea-a41f-02e60452d3a4
20:54:55	ERROR	    X panther-cloud-security failed (6/10): UPDATE_ROLLBACK_COMPLETE
20:55:05	ERROR	stack panther-log-analysis: AWS::Lambda::Function UpdaterFunction UPDATE_FAILED: 1 validation error detected: Value '10' at 'memorySize' failed to satisfy constraint: Member must have value greater than or equal to 128
20:55:05	ERROR	    X panther-log-analysis failed (7/10): UPDATE_ROLLBACK_COMPLETE
20:55:15	ERROR	stack panther-cw-alarms-CloudSecurityAlarms-1JPRY92ADP5Y9: AWS::CloudWatch::Alarm PantherAlarmSQSTooOldpantherremediationqueue UPDATE_FAILED: 1 validation error detected: Value 'GraterThun' at 'comparisonOperator' failed to satisfy constraint: Member must satisfy enum value set: [GreaterThanThreshold, LessThanThreshold, LessThanOrEqualToThreshold, GreaterThanOrEqualToThreshold]
20:55:15	ERROR	stack panther-cw-alarms: AWS::CloudFormation::Stack CloudSecurityAlarms UPDATE_FAILED: Embedded stack arn:aws:cloudformation:us-west-2:1234:stack/panther-cw-alarms-CloudSecurityAlarms-1JPRY92ADP5Y9/3f86c5d0-7f5b-11ea-a676-068a25e2e0ea was not successfully updated. Currently in UPDATE_ROLLBACK_IN_PROGRESS with reason: The following resource(s) failed to update: [PantherAlarmSQSTooOldpantherremediationqueue].
20:55:15	ERROR	    X panther-cw-alarms failed (8/10): UPDATE_ROLLBACK_COMPLETE
20:55:15	FATAL	deploy failed: panther-core, panther-appsync, panther-cloud-security, panther-log-analysis, panther-cw-alarms
```


deployments/cloud_security.yml
```diff
@@ -102,7 +102,7 @@ Resources:
       # When the system has recovered they should be re-queued to the `panther-alert-processor-queue` using
       # the Panther tool `requeue`.
       # </cfndoc>
-      MessageRetentionPeriod: 1209600 # Max duration - 14 days
+      MessageRetentionPeriod: '1209600' # Max duration - 14 days
```
austinbyers, line 105: Without this, `yaml.Marshal` saves it as 1.2096e6, which is technically valid yml, but CFN doesn't like it

tools/cfngen/cloudwatchcf/util.go
```diff
@@ -32,15 +32,17 @@ func getResourceFloat32Property(key string, resource map[interface{}]interface{}
...
-	switch props := resource[(interface{})("Properties")].(type) {
-	case map[interface{}]interface{}:
-		switch val := props[(interface{})(key)].(type) {
+func getResourceProperty(key string, resource map[string]interface{}) string {
+	switch props := resource["Properties"].(type) {
+	case map[string]interface{}:
+		switch val := props[key].(type) {
 		case string:
 			return val
+		case float32, float64:
```
austinbyers, line 41: Ints were apparently being parsed as floats somewhere in the yaml/json conversion

tools/mage/deploy.go
```diff
@@ -322,39 +347,39 @@ func uploadLayer(awsSession *session.Session, libs []string, bucket, key string)
...
-	parallelStacks := 0
+	results := make(chan goroutineResult)
+	count := 0
 
 	// Alarms
-	parallelStacks++
-	go func(result chan string) {
-		deployTemplate(awsSession, alarmsTemplate, sourceBucket, alarmsStack, map[string]string{
+	count++
+	go func(c chan goroutineResult) {
```
austinbyers, line 356: sharing a common parallelization pattern with `mage fmt`, `mage setup` and `mage test:ci`

tools/mage/deploy_template.go
```diff
@@ -19,152 +19,313 @@ package mage
...
-	if strings.HasPrefix(strings.TrimSpace(line), "TemplateURL: ") {
-		// Break the line down to the pieces we need
-		lineParts := strings.Split(line, "https://")
-		uriParts := strings.Split(lineParts[1], "/")
-		prefixParts := strings.Split(uriParts[0], ".")
-
-		// Check if the format is already correct
-		if prefixParts[1] == "amazonaws" {
-			return line
+// If the stack is ROLLBACK_COMPLETE or otherwise failed to create, it will be deleted automatically.
```
kostaspap, line 218: Not sure, why do we want to delete these stacks automatically? I am thinking someone might want to investigate the reason that creating the stack failed, shouldn't they be able to look at it?

austinbyers:
> My question is, does this change mean someone no longer requires AWS CLI in order to deploy panther?

@kostaspap correct. From the (admittedly lengthy) PR description:

> Implement asset packaging ourselves (closes: #136)
...
    * The aws CLI is no longer required to deploy Panther

tools/mage/deploy_template.go
austinbyers: > Not sure, why do we want to delete these stacks automatically?

It's the only way to continue a deployment in one of these states. This is a common scenario I know at least @3nvi, @rleighton and myself have encountered:

* Deploy a change in a fresh deployment
* It fails (now, with this PR, you'll immediately know why! No need to go to the console)
* Find and fix the bug
* Deploy again
* Deploy fails - stack is ROLLBACK_COMPLETE
* (the developer is now annoyed) Manually delete stack and wait for it to finish
* Deploy again

> I am thinking someone might want to investigate the reason that creating the stack failed, shouldn't they be able to look at it?

This PR finds all relevant error messages for a failed stack, so you typically won't need the console anymore to troubleshoot failed deployments.

Even so, deleting a stack does NOT delete its history. It's retained for 90 days, you can still see it by filtering for deleted stacks in the AWS console. Every instance of a stack has a unique ID, which is why you can still see deleted stacks even if they have the same name as an active stack.

(There's also a comment in the source that explains this)

austinbyers:
I found a new edge case: if the very first `panther-bootstrap` stack fails to create, we need to do a full teardown to delete the orphaned S3 buckets and ACM cert (@nhakmiller)

That change has been added

tools/mage/deploy_template.go
```diff
@@ -19,150 +19,322 @@ package mage
...
+			return nil, fmt.Errorf("failed to describe stack %s: %v", stack, err)
 		}
-		changeSetType = "UPDATE"
+
+		status := *detail.Stacks[0].StackStatus
+		if _, ok := terminalStackStatus[status]; ok {
+			return detail.Stacks[0], nil
+		}
+
+		time.Sleep(pollInterval)
```
nhakmiller, line 320: No timeouts here? It seems like if nothing has changed for 45 minutes users may want to know that either something is still happening or something failed. Maybe every 60 (5 minutes) - 120 (10 minutes) pollInterval's a little message just saying what stack is being waited on so a user can go manually check stuff out if needed?

tools/cfnparse/parse.go
```diff
@@ -0,0 +1,97 @@
...
+			pythonVirtualEnvPath, err = filepath.Abs(filepath.Join(path, ".setup", "venv"))
+			if err != nil {
+				panic(err)
+			}
+			return
+		}
+
+		// Work our way up the file hierarchy
+		path = filepath.Join(path, "..")
+	}
```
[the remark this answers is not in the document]
rleighton: Agree with Nick. This works, but is fragile. If mage drives this, perhaps it could set an env var that this code uses (and panics if not set).

tools/mage/deploy.go
```diff
@@ -305,7 +330,7 @@ func uploadLayer(awsSession *session.Session, libs []string, bucket, key string)
 	// └ python/policyuniverse-VERSION.dist-info/
 	//
 	// https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html#configuration-layers-path
-	if err := shutil.ZipDirectory(filepath.Dir(layerSourceDir), layerZipfile); err != nil {
+	if err := shutil.ZipDirectory(filepath.Dir(layerSourceDir), layerZipfile, false); err != nil {
```
rleighton, line 333: Maybe add comment why 'false' is used.

tools/mage/deploy_certificate.go
```diff
@@ -60,25 +60,24 @@ func certificateArn(awsSession *session.Session, settings *config.PantherConfig)
...
+	// CFN makes us re-specify the parameter if it already had a non-default value.
+	stack, err := waitForStack(cfn.New(awsSession), bootstrapStack)
+	if err != nil {
+		// If the stack doesn't exist yet, its reported status will be DELETE_COMPLETE (not an error)
+		logger.Fatal(err)
 	}
 
-	if errStackDoesNotExist(err) {
-		// The stack doesn't exist yet - upload a new certificate
+	status := *stack.StackStatus
```
rleighton, line 71: A switch stmt would be easier to read

tools/mage/deploy_template.go
```diff
@@ -19,150 +19,322 @@ package mage
...
 
-// Upload resources to S3 and return the path to the modified CloudFormation template.
-// TODO - implement this directly to avoid the aws cli (https://github.com/panther-labs/panther/issues/136)
-func cfnPackage(templatePath, bucket, stack string) string {
-	outputDir := filepath.Dir(templatePath)
-	if !strings.HasPrefix(outputDir, "out") {
-		outputDir = filepath.Join("out", outputDir)
+// Upload resources to S3 and return the modified CloudFormation template.
+//
+// This is similar to the "aws cloudformation package" CLI command, but our implementation
```
rleighton, line 94: Nice comment

tools/mage/deploy_template.go
```diff
@@ -19,150 +19,322 @@ package mage
...
+//    - We rely only on asset hashes - AWS CLI will re-upload identical files with different timestamps.
+//    - AWS CLI suffers from a region bug when writing S3 URLs
+//        (possibly related to https://github.com/aws/aws-cli/issues/4372)
+//
+// Resources currently supported:
+//    - AWS::AppSync::GraphQLSchema - DefinitionS3Location
+//    - AWS::CloudFormation::Stack - TemplateURL
+//    - AWS::Serverless::Function - CodeUri
+//
+// The bucket can be empty to skip S3 packaging (e.g. for the bootstrap stack) - in that case,
```
rleighton, line 107: Does this comment belong here?

tools/mage/deploy_template.go
```diff
@@ -19,150 +19,322 @@ package mage
...
+
+	if bucket == "" {
+		// No S3 packaging - just emit the template in standard form
+		return yaml.Marshal(cfnBody)
 	}
-	if err := os.MkdirAll(outputDir, 0755); err != nil {
-		logger.Fatalf("failed to create directory %s: %v", outputDir, err)
+
+	logger.Debugf("deploy: packaging %s assets", templatePath)
+	for _, resource := range cfnBody["Resources"].(map[string]interface{}) {
```
rleighton, line 121: This is pretty cool!

tools/mage/deploy_template.go
rleighton: I like the idea that every 5min you get a message like "Waiting on CF to complete" or sth. Not sure we need a timeout, we have ^C

tools/mage/util_cfn.go
```diff
@@ -148,6 +178,61 @@ func walkPantherStack(client *cfn.CloudFormation, stackID *string, handler func(
 	return nil
 }
 
+// Log failed resources from the stack's event history.
+//
+// Use this after a stack create/update fails to understand why the stack failed.
```
rleighton, line 183: Excellent.

rleighton:
I suggest waiting to push this until after this week's release so we have a full sprint to burn this in.

tools/cfnparse/parse.go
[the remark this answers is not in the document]
austinbyers: I agree it's awkward, this is one thing that's easier in Python lol. We could use an env variable, yes: `mage test:ci` invokes `VENV=... go test` and then this code changes to use that env variable to lookup where `cfn-flip` is

The reason I chose not to do that: it would make it much more difficult to run tests manually. In other words, running `go test` directly (or from your IDE) would no longer work - you'd be forced to either always run tests with mage or just remember to keep that environment variable set constantly. This would really hurt my personal workflow - I like to run `go test` for various packages individually and I use the IDE test feature quite a lot.

While awkward, I don't agree that it's *fragile* - `magefile.go` is a specific name that will never exist anywhere else in the repo unless we substantially refactor how mage works, at which point this won't matter. This code runs only for mage and all code paths are exercised in CI.

Another option I tried was to have unit tests parse cfn directly to avoid `cfn-flip` completely. But then we can't ever have cloudformation unit tests with `!Ref` or other short-form functions. And it's easier to have unit tests parse cfn exactly the same way mage does, because there are subtle but important format differences that we need to check for (e.g. `map[interface{}]interface{}` vs `map[string]interface{}` when parsing yaml to json)

So I agree it's awkward and I'm to other ideas if you can think of one that passes tests with an unmodified `go test ./...` call

tools/mage/deploy_template.go
austinbyers: Terraform does something similar with periodic status messages for operations that are taking more than a few minutes, good idea, I can do that.

As for a timeout, I could go either way. We know that in general, it's possible for stacks to take 30-60 minutes or more (e.g. CloudFront distributions). But Panther's specific stacks should never take more than a few minutes, so perhaps it's prudent to stop after 10 mins or so because we know Panther should never be taking that long. Lets the user know sooner that something is probably wrong

tools/mage/deploy.go
austinbyers: In this specific place, it actually doesn't matter either way. We only need it `false` when we want to ignore modification times. Since we're building a layer here, it's actually probably better to keep the "originial" packaging, I'll change it to true

tools/mage/deploy_template.go
austinbyers: Yes, by "bucket" I mean literally the parameter in the function can be `""` haha not the bucket itself. I'll clarify, I see how that can be confusing

tools/mage/deploy_template.go
nhakmiller: Thinking on int, the message is probably sufficient no timeout needed. Letting users know something is happening, then they can manually check on whatever the stack is being worked on if it takes too long maybe.

tools/mage/deploy_certificate.go
austinbyers: Actually, after taking another look, I don't think I need to care about the stack status at all here

austinbyers:
I made a few more adjustments based on feedback and more testing:

- Every few minutes, show a message for stacks still in progress
- Create custom waiters `waitForStackCreate`, `waitForStackUpdate`, etc, which will fail as soon the stack enters an invalid state. For example, if a stack is UPDATE_ROLLBACK_IN_PROGRESS, we don't need to wait until it finishes before reporting the error to the user
- Change teardown stack deletion to use the same stack wait flow as the deploy process
- Fix a panic when trying to delete buckets which have already been deleted outside the stack

Given the fairly large changes, @nhakmiller can I get another quick review? The easiest way is to checkout the branch, deploy it, make sure everything's still good. Then deliberately break the CloudFormation, deploy again, see if the error is more helpful

rleighton:
> I made a few more adjustments based on feedback and more testing:
> 
> * Every few minutes, show a message for stacks still in progress
> * Create custom waiters `waitForStackCreate`, `waitForStackUpdate`, etc, which will fail as soon the stack enters an invalid state. For example, if a stack is UPDATE_ROLLBACK_IN_PROGRESS, we don't need to wait until it finishes before reporting the error to the user
> * Change teardown stack deletion to use the same stack wait flow as the deploy process
> * Fix a panic when trying to delete buckets which have already been deleted outside the stack
> * Removed AWS CLI from "development" documentation
> 
> Given the fairly large changes since the last review, @nhakmiller or @rleighton can I get another quick review? The easiest way is to checkout the branch, deploy it, make sure everything's still good. Then deliberately break the CloudFormation (e.g. set an S3 bucket version status to "FAKE"), deploy again, see if the error is more helpful

Tell folks they need to `mage clean setup` first, I did not have `cfn-flip` and got an error.

Love the progress messages!

Got the errors: ```
07:43:56	ERROR	stack panther-bootstrap: AWS::S3::Bucket AuditLogs UPDATE_FAILED: Status FAKE is not in [Enabled, Disabled]
07:43:56	ERROR	    X bootstrap: stacks failed (2/2): UPDATE_ROLLBACK_IN_PROGRESS
```

Fixed that , and re-deployed, worked.

Ran `mage teardown`. Nice INFO messages.  No issues.

I would suggest that we get this out and use the rest of the sprint to fine tune. Nice job!

PR #743, opened 2020-04-21
Globals Bulk Upload Support

nhakmiller:
## Background

While implementing support for globals, I did not add bulk upload support because I was not sure how CLI based global management would work. Now that we've figured it out (change to the panther_analysis_tool coming soon), this change will make sure the backend can handle bulk upload requests that included globals in them.

## Changes

- Modified the bulk upload endpoint of the analysis-api to better handle globals
- Modified the bulk upload endpoint to more accurately report what analyses were actually modified during a bulk upload operation

## Testing

- Existing unit / integration testing
- Testing manually in dev account


internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -112,13 +116,20 @@ func BulkUpload(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyR
 			} else if result.changeType == updatedItem {
 				*counts.ModifiedPolicies++
 			}
-		} else {
+		} else if result.item.Type == typeRule {
```
austinbyers, line 119: Nit: consider a `switch` on the item type instead of if / else if / else

internal/core/analysis_api/handlers/util.go
```diff
@@ -196,6 +198,51 @@ func writeItem(item *tableItem, userID models.UserID, mustExist *bool) (int, err
...
+		oldItem.Description == newItem.Description && oldItem.DisplayName == newItem.DisplayName &&
+		oldItem.Enabled == newItem.Enabled && oldItem.Reference == newItem.Reference &&
+		oldItem.Runbook == newItem.Runbook && oldItem.Severity == newItem.Severity &&
+		oldItem.DedupPeriodMinutes == newItem.DedupPeriodMinutes &&
+		setEquality(oldItem.ResourceTypes, newItem.ResourceTypes) &&
+		setEquality(oldItem.Suppressions, newItem.Suppressions) && setEquality(oldItem.Tags, newItem.Tags) &&
+		len(oldItem.AutoRemediationParameters) == len(newItem.AutoRemediationParameters) &&
+		len(oldItem.Tests) == len(newItem.Tests)
+
+	if itemsEqual {
```
austinbyers, line 218: Nit: instead of nesting so much in this if block, just return early if you know you can:

```go
if !itemsEqual {
    return true
}
```

internal/core/analysis_api/handlers/util.go
```diff
@@ -196,6 +198,51 @@ func writeItem(item *tableItem, userID models.UserID, mustExist *bool) (int, err
...
+		}
+		// Check Tests for equality
+		oldTests := make(map[models.TestName]*models.UnitTest)
+		for _, test := range oldItem.Tests {
+			oldTests[test.Name] = test
+		}
+		for _, newTest := range newItem.Tests {
+			oldTest, ok := oldTests[newTest.Name]
+			// The test Resource gets sorted by Dynamo in some fashion that is hard to replicate without converting
+			// the whole string to a struct, so we just check if they're approximately equal
```
austinbyers, line 234: You can json marshal the struct for comparison if you want, the extra few ms of overhead isn't a big deal in the relatively rare bulk upload operation

internal/core/analysis_api/handlers/util.go
```diff
@@ -196,6 +198,51 @@ func writeItem(item *tableItem, userID models.UserID, mustExist *bool) (int, err
 	return changeType, nil
 }
 
+// itemUpdated checks if ANY field has been changed between the old and new item. Only used to inform users whether the
+// result of a BulkUpload operation actually changed something or not.
+//
+// DO NOT use this for situations the items MUST be exactly equal, this is a "good enough" approximation for the
+// purpose it serves, which is informing users that their bulk operation did or did not change something. Some fields
+// are not checked, and some are not checked adequately (e.g. test bodies)
```
austinbyers, line 206: Could be confusing if a user just changes tests and then the BulkUpload says nothing changed?

It doesn't seem too hard to do the full equality checking, don't worry about the performance of `reflect.DeepEqual` or json serialization. I can help with this too, if you like

PR #745, opened 2020-04-21
Aws cloudtrail digest

### LINKED ISSUE #709: Panther supports CloudTrail digest logs
### Description

Panther should support CloudTrail Digest log files. 
- Documentation with samples: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-log-file-validation-digest-file-structure.html

### Acceptance Criteria

- Panther has a new Parser for CloudTrail Insights events
- Unit tests for the parser
- Classifier uses the new parser
- Appropriate tables are created in Glue.

alxarch:
## Background

Implement parsers for AWS CloudTrail digest logs to resolve #709 

## Changes

- Adds `CloudTrailDigest` parser and event
- <strike>Adds a new field `p_any_hashes` to `PantherLog`  to support   arbitrary hash values</strike>

## Testing

- Go unit tests
- mage test:ci


alxarch:
The initial intention was to add a `p_any_hashes` field to `parsers.PantherLog` but this conflicts with tests in other parts of the project. Let's reevaluate the usefulness of such a field and implement this in a separate PR.

alxarch:
I would like to point out as well that the AWS docs and example seem inconsistent and it'd be best to find actual samples of digest logs to make sure that the parser will work correctly in the Real World (TM)

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_digest.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (event *CloudTrailDigest) updatePantherFields(p *CloudTrailDigestParser) {
+	// Use end time as it's the time the digest was actually delivered
+	event.SetCoreFields(p.LogType(), event.DigestEndTime, event)
+	event.AppendAnyAWSAccountIdPtrs(event.AWSAccountID)
+	// event.AppendAnyHashesPtr(event.PreviousDigestHashValue)
+	// for i := range event.LogFiles {
+	// 	lf := &event.LogFiles[i]
+	// 	event.AppendAnyHashesPtr(lf.HashValue)
+	// }
```
kostaspap, line 106: Can leave out.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_digest_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	// panther fields
+	expectedEvent.PantherLogType = aws.String("AWS.CloudTrailDigest")
+	expectedEvent.PantherEventTime = (*timestamp.RFC3339)(&expectedDateEnd)
+	expectedEvent.AppendAnyAWSAccountIds("111122223333")
+	// expectedEvent.AppendAnyHashes(
+	// 	"97fb791cf91ffc440d274f8190dbdd9aa09c34432aba82739df18b6d3c13df2d",
+	// 	"9bb6196fc6b84d6f075a56548feca262bd99ba3c2de41b618e5b6e22c1fc71f6",
+	// )
```
kostaspap, line 100: Again, better leave out.

internal/log_analysis/log_processor/parsers/pantherlog.go
```diff
@@ -55,6 +55,7 @@ type PantherLog struct {
 	PantherAnyDomainNames *PantherAnyString `json:"p_any_domain_names,omitempty" description:"Panther added field with collection of domain names associated with the row"`
 	PantherAnySHA1Hashes  *PantherAnyString `json:"p_any_sha1_hashes,omitempty" description:"Panther added field with collection of SHA1 hashes associated with the row"`
 	PantherAnyMD5Hashes   *PantherAnyString `json:"p_any_md5_hashes,omitempty" description:"Panther added field with collection of MD5 hashes associated with the row"`
+	// PantherAnyHashes      *PantherAnyString `json:"p_any_hashes,omitempty" description:"Panther added field with collection of hashes of any algorithm associated with the row"`
```
kostaspap, line 58: Better leave out. We can add it back when needed

web/src/constants.ts
```diff
@@ -82,6 +82,7 @@ export const LOG_TYPES = [
   'AWS.AuroraMySQLAudit',
   'AWS.CloudTrail',
   'AWS.CloudTrailInsight',
+  'AWS.CloudTrailDigest',
```
kostaspap, line 85: nit: Let's keep them in alphabetical order, better put it above AWS.CloudTrailInsight

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_digest.go
```diff
@@ -0,0 +1,107 @@
...
+
+// LogType returns the log type supported by this parser
+func (p *CloudTrailDigestParser) LogType() string {
+	return "AWS.CloudTrailDigest"
+}
+
+func (event *CloudTrailDigest) updatePantherFields(p *CloudTrailDigestParser) {
+	// Use end time as it's the time the digest was actually delivered
+	event.SetCoreFields(p.LogType(), event.DigestEndTime, event)
+	event.AppendAnyAWSAccountIdPtrs(event.AWSAccountID)
```
rleighton, line 101: Can use `HashAlgorithm` to lookup the hashes we support and then populate?

alxarch: Haven't seen any samples not using SHA256 so far

rleighton: We should add SHA256 as a panther log type

alxarch: I also attempted to add a `p_any_sha256_hashes` to `PantherLog` but it conflicted with other parts of the code we decided with kostas to leave it out for now.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_digest.go
alxarch: Fixed

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_digest.go
alxarch: We were writing at the same time, didn't see your comment before I posted mine :)
As I said above adding it to the 'base' struct created issues with Python code that had to sth to do with the `rules`. I'm not familiar with that part of the project to make this change.

rleighton: Ok, for now but that is an important indicator (not for this log source) but for host-based agent logs. So it can wait until we have a log source with some juicy file related data.

kostaspap: Jumping in because maybe I created some confusion with miscommunication from my side.
We shouldn't remove/alter any of the existing fields so we should't remove the `md5`, `sha1` etc fields. That could cause backwards incompatibility issues. 

We can however add a SHA256 any field. I didn't feel strongly about it so I initially I told @alxarch that he could leave it out. If @rleighton things it's appropriate, we can add it.

internal/log_analysis/log_processor/parsers/pantherlog.go
kostaspap, line 58: Should remove commented code.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_digest.go
rleighton: We will want a SHA256 "p_any_" field, if this is the first source to use it, you might as well add it. For this source (I think) the values are uninteresting from a security POV, but for consistency we should put them in.  I do like the idea of also having a "p_any_file_hashes" that also hold the union of all the hashes for a file. Many host-agent logs have typically 3 hashes for all files: md5, sha1 and sha256.

rleighton: @alxarch, for context, I'm not sure how much security related work you have done, so I'lll explain the use of the file hashes in the logs. Host based agent logs (for example on your laptop) will report manifests of files with hashes, sometime as a result of a security events, sometimes just to report a file inventory. They often have 3 or more hashes for each file. The reason for more than 1 is to use as keys to join to threat indicator sources.  You. can buy (and get for free on the Internet) lists of "bad" files or use a service like VirusTotal to lookup a hash to see if it is bad. The threat lists are very inconsistent, some only have 1 hash as a key for a "bad" file, others have more. md5 is still used but is being phased out. So we want the type of the hash to do that join efficiently to see if a file is on a "bad" list automatically. The hashes are also used to join between different log sources that concern files (not necessarily threat feeds). The "p_any_file_hashes" field will be useful for a different case from joining (btw, you could use this to join also, just not as efficiently as if you know the type), this is useful if someone gives you a list of "bad' file hashes and you want know "have I ever seen these on my network?", so having them all in 1 field makes that query easy.

alxarch: @rleighton Thanks! I didn't know about these `threat indicator sources`. I added the sha256 field, it had a conflict in an unrelated test about athena views that I mended to pass with the new field but someone should review that test because I don't know what it was supposed to be checking.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail_digest.go
```diff
@@ -0,0 +1,104 @@
...
+	event.AppendAnySHA256HashesPtr(event.PreviousDigestHashValue)
+	for i := range event.LogFiles {
+		lf := &event.LogFiles[i]
+		event.AppendAnySHA256HashesPtr(lf.HashValue)
+	}
```
kostaspap, line 103: You could make that slightly more compact: 
```suggestion
	for _, file := range event.LogFiles {
		event.AppendAnySHA256HashesPtr(file.HashValue)
	}
```

alxarch: Will change, it's more readable that way. It's an optimization habit, getting just the pointer is faster in slices of `[]Struct` because `for _, foo := range foos` copies the struct to the stack so if the slice element is *not* a pointer it's a tiny bit heavier

kostaspap:
@rleighton  not sure if there is anything else? Should I merge this?

PR #759, opened 2020-04-22
Remove un-used subscriptions

3nvi:
## Background

This PR removes `queryDone` subscription from the schema, along with its resolvers. Until we pursue a subscription implementation for the Log Query results (thus replacing polling), we shouldn't have it available in the schema. Currently:

- the `queryDone` is not utilized
- the `@aws_iam` breaks current deployments on the `graphql-codegen` level

**The related work by @rleighton is still valid and will be used as reference for the soon-to-come implementation** 

## Changes

- Remove `queryDone` subscription
- Remove related resolver


rleighton:
I'll comment out the integration tests in this PR so they still run and we can resume this work easily.

Coming later today.

api/graphql/schema.graphql
```diff
@@ -719,17 +719,17 @@ input InviteUserInput {
...
-  queryId: String!
-  workflowId: String!
-}
-
-type QueryDone @aws_iam {
-  userData: String!
-  queryId: String!
-  workflowId: String!
-}
+#input QueryDoneInput {
```
austinbyers, line 722: My preference is to remove unused code completely instead of leaving it commented out - we can always find it again from the version history

austinbyers: But if we actually plan to re-add it soon, comments are fine. The tendency is just people think they will re-add it soon and then never do and then it gets confusing

rleighton: I'm worried about losing the working backend which was implemented so it was ready for Aggelos to use, but then he was reprioritized.  I hope we get to this, because it is an important feature that sets us apart (so modern! using graphql subscriptions). 

I have a suggestion to make this easier, let's close this PR and just let me do a new one with this code _removed_ BUT I'll copy it into the integration test  as a comment  (disabled). This way when we want to resume we just follow those instructions in the the comment showing how to add the resolvers. The we can be running again in 5min.

Thoughts?


Thoughts?

austinbyers: Up to you! I'm fine with however you want to do it, it's just a habit of mine after years of code reviews to cringe at commented out code lol but if we plan to add it soon, I don't mind

rleighton: Ok, working on this now. Sorry for the delay but had meetings. I'm closing this, new PR coming.

rleighton:
Closing this PR and will open a different one soon.

PR #766, opened 2020-04-23
[WIP] Pantherlog decouple

alxarch:
## Background

In preperation for dynamic log schema definition some modifications seem to be needed in the interface between `parsers` module and `destinations`. Although custom logs should allow for the definition of any log type there will be many cases where struct-based parsing will be required for custom parsing rules, multi-event log lines, text formats etc. To be able to support both, a common *minimal interface* that both methods can easily implement must be implemented. At the same time parsers should be easier to test more thoroughly given the complexity of input that needs to be tested against multiple samples to avoid missclassifications and errors in parsing.

Currently parsed logs are written to S3 through passing of a `*PantherLog` struct. `PantherLog` struct has an unexported field `event` that holds the actual parsed event. And all `event` structs must embed a  `PantherLog` to be able to serialize to JSON properly. This creates a circular dependency that obfuscates parser code and *makes  testing harder* as a series of methods need to be called in *a specific order* on the embedded `PantherLog` before handing it over to be processed or tested. 

#### `parsers.Result`
Going over the code and looking at the relations between packages I realized that once a log line is parsed, all we need to know to deliver it to a destination is it's `type`, `time` and `json` form. The main reason `event` was embedded seems to be so that a `destination` could call `jsoniter.Marshal` on the event and get a blob to store in a bucket. This PR introduces `parsers.Result` that acts as the *minimal interface* between parsers and consumers (`destination`). This interface can easily by implemented by both the current struct-based system and the dynamic one. This contains the deserialize/serialize logic inside the `parsers` package and specifically inside the logic provided by a parser. 

#### `parsers.Interface`
The `parsers.Parser` interface was simplified to only provide the actual parsing method with the implementation-agnostic `parsers.Interface` that defines a single `Parse(log string) ([]*parsers.Result, error)` method. The creation of new parsers was moved to a new registerable type called `parsers.LogType` and the log type information is provided by the `parsers.Result` type opening the way for the possibility of parsers that can return multiple/different log types from a single log input type`.

#### `parsers.LogType`
`parsers.LogType` defines a log type with a name, description, parser factory and schema struct. A concurrent-safe registry of LogTypes is added to the `parsers` package both at package-level with a `defaultRegistry` global var and as a standalone registry that can be used for concurrent tests without requiring resetting of global variables. Registration of log types can happen in an init block or at any point during the runtime of the program. All current log parsers have init blocks that register their log types at startup and a Compatibility layer is introduced in the `registry` module for controlling project-wide registration of log types.

#### `parsers.LogEventer` - `logs.Event`
The draft of the implementation meant more code had to be written for all parsers since the JSON 'blob' had to have the fields in the correct way to be used as a glue table. To solve this and drastically reduce the code needed to write a new struct based parser at the same time a new interface was created `parsers.PantherEventer` which returns a `logs.Event` instance. Parsed struct events can implement this interface to take advantage of parser building/testing helpers in the `logs` package. The interface returns all info required to populate a `PantherLog` from a parsed event. A pool of `logs.Event` objects is used to minimize allocations across the log processor pipeline. At the same time in order to 'pack' a parsed event into `parsers.Result` the `logs.PackResult` helper is introduces. In this draft all parsers at the time of branching have been ported to use the `logs.Event` way of parsing logs to demonstrate how the packages work together and get tested.

#### `logs.Field` - `logs.FieldKind`
Panther fields (ie `p_any_...` fields) use the `logs.Field` type that is a pair of `logs.FieldKind` - value (string). The `logs.Field` approach paves the way for custom logs to collect fields from any JSON input and makes testing of parsers more declarative. Each `logs.Event` object contains a (reusable) map of `FieldKind -> []string` values that can be easily checked for complete field extraction of expected fields. Field parsers follow the interface `logs.FieldParser` that provides an append method `ParseFields(value string, fields []logs.Field) ([]logs.Field, error) that can handle extraction of zero, one or multiple fields and of fields of different kinds with a single testable interface. (This is used extensively in the custom logs code to extract panther fields using `gjson` at any position in an input `JSON` line.

## Changes

- Introduce a new 'parsed' struct `parsers.Result` used as a minimal interface between parsers and destinations.
- Add `parsers.LogType` as the main way to register log types and a concurrent-safe registry for them.
- Change `parsers.Parser` to return `parsers.Result` (renamed to `parsers.Interface` to avoid name conflicts when this PR is split)
- Introduce a new event interface `logs.Event` that exprorts all needed fields from a parsed log event
- Introduce the struct `logs.Field` that can hold any string field that needs to be added as meta to a parsed log event.
- Add methods and helper functions to handle `logs.Field` extraction in a unified way across parsers. 
-  Add helpers to easily parse and pack panther logs into `parsers.Result`
- Add test utilities to test both the `parsers.PantherEventer` interface and the `parsers.Interface` easily and flexibly
- Add `parsers.Metered` parser wrapper to have metrics on any parser

## Testing

- This is a draft PR, all go unit tests pass (but some tests required modifications) 

## Issues to be resolved

- The code that produces the schemas for Glue tables should use the struct synthesized by `parsers.ComposeStruct` to include the fields inside `parsers.ParserLog`
- <strike>The current demo implementation doesn't handle the case of `AWSParserLog` that is used as a base log type extending `ParserLog` with extra optional fields. This is a pattern that is definitely useful to use in cases where there are a lot of common fields in a group of logs. One (not elegant) solution to this would be to extend the `enum` of `PantherFieldKind` to include all the field kinds and add fields with `omitempty` to all tables. Another approach would be to extend `PantherLog` to `AWSPantherLog` and override the `SetFields` to recognize more enum values and then delegate to `PantherLog#SetFields`. If we're careful about enum boundaries this will work without conflicts (ie AWS kind values could start at 1000, GCP kind values could start at 2000 etc)</strike> Solved using  `logs.RegisterMetaPrefix`.


alxarch:
Again be aware that this is completely WIP to be able to discuss this approach. I'm leaving this as a proof of concept to be reviewed and taken into consideration for the discussion about dynamic log schema definition and will be working on some parsers that were prioritized.

internal/log_analysis/log_processor/parsers/gitlablogs/audit.go
```diff
@@ -62,28 +62,15 @@ func (p *AuditParser) New() parsers.LogParser {
...
 
 // LogType returns the log type supported by this parser
 func (p *AuditParser) LogType() string {
 	return TypeAudit
 }
 
-func (event *Audit) updatePantherFields(p *AuditParser) {
-	event.SetCoreFields(p.LogType(), event.Time, event)
+func (event *Audit) PantherEvent() (string, time.Time, []parsers.PantherField) {
+	return TypeAudit, event.Time.UTC(), nil
```
kostaspap, line 75: So basically every event implements it's own method that decorates it/populates it with Panther events. Interesting.

internal/log_analysis/log_processor/parsers/pantherlog.go
```diff
@@ -57,6 +60,32 @@ type PantherLog struct {
...
+
+// var (
+// 	typPantherLog = reflect.TypeOf(PantherLog{})
+// )
+
+func NewPantherLog(logType string, tm time.Time, fields ...PantherField) *PantherLog {
+	now := time.Now()
+	if tm.IsZero() {
+		tm = now
+	}
```
kostaspap, line 77: You could be passing the `tim` as `*time.Time`. You told me however there was a specific reason why you didn't want to have a pointer here?

internal/log_analysis/log_processor/parsers/pantherlog.go
```diff
@@ -57,6 +60,32 @@ type PantherLog struct {
...
+	LogType   string
+	EventTime time.Time
+	JSON      []byte
+}
+
+// var (
+// 	typPantherLog = reflect.TypeOf(PantherLog{})
+// )
+
+func NewPantherLog(logType string, tm time.Time, fields ...PantherField) *PantherLog {
```
kostaspap, line 73: By passing `time.Time` it means we basically don't need the "custom" timestamp structs right? Any parser can just implement it's own parsing logic and we enforce the serialization format of the time here, centrally. Correct?

internal/log_analysis/log_processor/parsers/pantherlog.go
```diff
@@ -94,24 +123,121 @@ func (any *PantherAnyString) UnmarshalJSON(jsonBytes []byte) error {
...
+		case KindMD5Hash:
+			pl.AppendAnyMD5Hashes(field.Value)
+		case KindSHA1Hash:
+			pl.AppendAnySHA1Hashes(field.Value)
+		case KindDomainName:
+			pl.AppendAnyDomainNames(field.Value)
+		}
+	}
+}
+func PantherLogEvent(event PantherEventer) PantherLog {
```
kostaspap, line 226: Is this used somewhere?

internal/log_analysis/log_processor/parsers/pantherlog.go
alxarch: It's a general rule that interfaces are easier to mock/test/implement if they use the 'core' types in their arguments/return values.
Since the parser already has parsed the timestamp somehow and the `PantherLog` struct will serialize it to the proper format there's no need to use a specialized value here and it's *a lot* easier to test.

internal/log_analysis/log_processor/parsers/pantherlog.go
alxarch: In general I'm not a big fan of pointer `args` being used to define 'emptiness' of a value. It has the negative effect of filling the code with `nil` checks which in turn forces functions to either return errors unnecessarily, being forced to panic or even worse panicking in places you didn't expect. As a rule of thumb function arguments should either use values by copy where values are 'small' (ie numbers, times, strings,maps) or use interface args that clearly define how the value will be used in the context of the function call. All other cases should be implemented as methods where the receiver is a pointer to the struct. This of course is a generic rule for API design and rules are made to be broken :) Just not where there's no benefit from breaking them.

kostaspap:
How would this work with the equivalent of AWSPantherLog?

internal/log_analysis/log_processor/parsers/pantherlog.go
alxarch: No that was an attempt to use the interface initially when `PantherLog` was still embedded in the event struct. To assign all panther values easily. Will remove if not used elsewhere.

internal/log_analysis/log_processor/parsers/pantherlog.go
kostaspap: In general I agree. In the cases of the parsers though, we have to be extra careful since the semantics of a `nil` can be different than the "default" value. E.g. a boolean field in parsed event being `nil` means it was not included upstream, "false" means it was included and it was false. 
Same with time, some upstream provides might send `nil` timestamp while others might for whatever reason sent epoch `0`. We have to be careful not to alter the semantics of what is send to us from upstream.

alxarch:
@kostaspap as mentioned in the description this is one of the issues to be resolved. I'm leaning on the second solution described. Will implement it in one of the aws logs to demonstrate

cmd/devtools/logprocessor/logprocessor/main.go
```diff
@@ -86,7 +87,8 @@ func main() {
 	}
 	zap.ReplaceGlobals(logger)
 
-	err = processor.Process(dataStreams, destinations.CreateDestination())
+	logTypes := parsers.DefaultRegistry()
```
alxarch, line 90: Pass log types registry as an explicit dependency. This helps with tests and provides a clear way to choose local over global (singleton-like) scope wherever needed.

internal/log_analysis/log_processor/classification/classifier.go
```diff
@@ -38,26 +37,25 @@ type ClassifierAPI interface {
 	// aggregate stats
 	Stats() *ClassifierStats
 	// per-parser stats, map of LogType -> stats
-	ParserStats() map[string]*ParserStats
+	ParserStats() map[string]parsers.Stats
```
alxarch, line 40: `parsers.Stats` Uses new stats struct based on `parsers.Metered` parser wrapper to be able to get metrics on any parser.

internal/log_analysis/log_processor/classification/classifier.go
```diff
@@ -38,26 +37,25 @@ type ClassifierAPI interface {
...
 }
 
 // ClassifierResult is the result of the ClassifierAPI#Classify method
 type ClassifierResult struct {
 	// Events contains the parsed events
 	// If the classification process was not successful and the log is from an
 	// unsupported type, this will be nil
-	Events []*parsers.PantherLog
+	Events []*parsers.Result
```
alxarch, line 48: Events from parsers are already serialized using `parsers.Result` type

internal/log_analysis/log_processor/classification/classifier.go
```diff
@@ -38,26 +37,25 @@ type ClassifierAPI interface {
...
 	// LogType is the identified type of the log
 	LogType *string
 }
 
 // NewClassifier returns a new instance of a ClassifierAPI implementation
-func NewClassifier() ClassifierAPI {
+func NewClassifier(logTypes ...parsers.LogType) ClassifierAPI {
```
alxarch, line 54: Log types for this classifier are passed as explicit dependencies to avoid global state in tests and allow updating of the classifier over the lifetime of a lambda.

internal/log_analysis/log_processor/classification/parser_queue.go
```diff
@@ -32,18 +31,25 @@ type ParserPriorityQueue struct {
...
 			penalty: 1,
 		})
 	}
 }
 
 // ParserQueueItem contains all the information needed to initialize a schema.
 type ParserQueueItem struct {
-	parser parsers.LogParser
+	logType string
+	parser  *parsers.Metered
```
alxarch, line 52: Use `parsers.Metered` to have stats over each parser inside the queue.

internal/log_analysis/log_processor/destinations/destinations.go
```diff
@@ -33,16 +33,20 @@ import (
 
 // Destination defines the interface that all Destinations should follow
 type Destination interface {
-	SendEvents(parsedEventChannel chan *parsers.PantherLog, errChan chan error)
+	SendEvents(parsedEventChannel chan *parsers.Result, errChan chan error)
```
alxarch, line 36: Destinations are insulated from log parser logic by using a `parsers.Result` that has already serialized the log event to `JSON`

internal/log_analysis/log_processor/destinations/destinations.go
```diff
@@ -33,16 +33,20 @@ import (
...
 }
 
 // CreateDestination the method returns the appropriate Destination based on configuration
-func CreateDestination() Destination {
+// If no registry is specified the default registry is used
+func CreateDestination(r *parsers.Registry) Destination {
```
alxarch, line 41: A log type registry is passed as an explicit dependency to help with tests. If `nil` the global `parsers.DefaultRegistry()` is used

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -299,11 +292,29 @@ func (destination *S3Destination) sendSNSNotification(key string, buffer *s3Even
 	return err
 }
 
-func getS3ObjectKey(logType string, timestamp time.Time) string {
+// The timestamp format in the S3 objects with second precision: yyyyMMddTHHmmssZ
+const S3ObjectTimestampFormat = "20060102T150405Z"
+
+func getS3ObjectKey(r *parsers.Registry, logType string, timestamp time.Time) string {
```
alxarch, line 298: A reistry of parsers is passed as an argument. (Can also be changed into a method of `S3Destination` to avoid direct dependency.

internal/log_analysis/log_processor/main/lambda.go
```diff
@@ -56,6 +57,6 @@ func process(lc *lambdacontext.LambdaContext, event events.SQSEvent) (err error)
 	if err != nil {
 		return err
 	}
-	err = processor.Process(dataStreams, destinations.CreateDestination())
+	err = processor.Process(dataStreams, destinations.CreateDestination(parsers.DefaultRegistry()))
```
alxarch, line 60: Use global registry

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -25,16 +25,26 @@ import (
...
 
-var ALBDesc = `Application Load Balancer logs Layer 7 network logs for your application load balancer.
-Reference: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html`
-
 const (
+	TypeALB               = "AWS.ALB"
 	albMinNumberOfColumns = 25
 )
 
+var LogTypeALB = parsers.LogType{
```
alxarch, line 37: Declarative definition of a new LogType at the same place as the parser code. (Easier for maintenance)

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -66,19 +76,18 @@ type ALB struct {
 	ActionsExecuted        []string           `json:"actionsExecuted,omitempty" description:"The actions taken when processing the request. This value is a comma-separated list that can include the values described in Actions Taken. If no action was taken, such as for a malformed request, this value is set to NULL."`
 	RedirectURL            *string            `json:"redirectUrl,omitempty" description:"The URL of the redirect target for the location header of the HTTP response. If no redirect actions were taken, this value is set to NULL."`
 	ErrorReason            *string            `json:"errorReason,omitempty" description:"The error reason code. If the request failed, this is one of the error codes described in Error Reason Codes. If the actions taken do not include an authenticate action or the target is not a Lambda function, this value is set to NULL."`
-
-	// NOTE: added to end of struct to allow expansion later
-	AWSPantherLog
 }
 
+var _ parsers.PantherEventer = (*ALB)(nil)
```
alxarch, line 81: The log event struct returns a `logs.Event` instance with all panther log related info`.
Tests for proper field extraction can test `logs.Event` directly.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -155,25 +164,15 @@ func (p *ALBParser) Parse(log string) ([]*parsers.PantherLog, error) {
...
-		return nil, err
-	}
-
-	return event.Logs(), nil
-}
-
-// LogType returns the log type supported by this parser
-func (p *ALBParser) LogType() string {
-	return "AWS.ALB"
+	return parsers.PackResults(event)
```
alxarch, line 167: Since `ALB` implements the `parsers.PantherEventer` interface it can take advantage of the `parsers.PackResults` helper for boilerplate code.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -155,25 +164,15 @@ func (p *ALBParser) Parse(log string) ([]*parsers.PantherLog, error) {
...
 }
 
-func (event *ALB) updatePantherFields(p *ALBParser) {
-	event.SetCoreFields(p.LogType(), event.Timestamp, event)
-	event.AppendAnyIPAddressPtr(event.ClientIP)
-	event.AppendAnyIPAddressPtr(event.TargetIP)
-	event.AppendAnyDomainNamePtrs(event.DomainName)
-	event.AppendAnyAWSARNPtrs(event.ChosenCertARN, event.TargetGroupARN)
+func (event *ALB) PantherEvent() *logs.Event {
```
alxarch, line 170: Panther fields are passed to the `logs.Event` in a declarative way. Notice that both common field types (IPAddress, DomainName) and aws specific fields use the same interface.

internal/log_analysis/log_processor/parsers/awslogs/alb_test.go
```diff
@@ -63,20 +63,17 @@ func TestHTTPLog(t *testing.T) {
...
-
-	// panther fields
-	expectedEvent.PantherLogType = aws.String("AWS.ALB")
-	expectedEvent.PantherEventTime = (*timestamp.RFC3339)(&expectedTime)
-	expectedEvent.AppendAnyIPAddress("192.168.131.39")
-	expectedEvent.AppendAnyIPAddress("10.0.0.1")
-	expectedEvent.AppendAnyAWSARNs("arn:aws:elasticloadbalancing:us-east-2:123456789012:targetgroup/my-targets/73e2d6bc24d8a067")
-
-	checkALBLog(t, log, expectedEvent)
+	testutil.CheckPantherEvent(t, event, TypeALB, tm,
```
alxarch, line 71: Testing of proper field extraction is straight forward allowing for creation of utilities to test any `logs.Event` parser easily.

internal/log_analysis/log_processor/parsers/awslogs/alb_test.go
```diff
@@ -63,20 +63,17 @@ func TestHTTPLog(t *testing.T) {
...
+		logs.IPAddress("192.168.131.39"),
+		logs.IPAddress("10.0.0.1"),
+		ARN("arn:aws:elasticloadbalancing:us-east-2:123456789012:targetgroup/my-targets/73e2d6bc24d8a067"),
+	)
+	testutil.CheckParser(t, log, TypeALB, event)
```
alxarch, line 76: Comprehensive test of parser output regardless of parsing logic. It checks existence of required fields in the resulting JSON, parser registration and more.

internal/log_analysis/log_processor/parsers/awslogs/awslogs.go
```diff
@@ -0,0 +1,56 @@
...
+	logs.RegisterPrefixMeta("AWS", NewMeta)
+
+	// Register AWS fields
+	logs.RegisterField(KindAWSARN, "aws_arn", ARN)
+	logs.RegisterField(KindAWSInstanceID, "aws_instance_id", InstanceID)
+	logs.RegisterField(KindAWSAccountID, "aws_account_id", AccountID)
+	logs.RegisterField(KindAWSTag, "aws_tag", Tag)
+}
+
+func NewMeta(event *logs.Event) (interface{}, error) {
```
alxarch, line 48: `logs.Meta` factory function that embeds the default panther meta fields (`logs.Meta`) and includes custom panther fields for AWS based logs.

internal/log_analysis/log_processor/parsers/awslogs/awslogs.go
```diff
@@ -0,0 +1,56 @@
...
+		LogTypeALB,
+		LogTypeAuroraMySQLAudit,
+		LogTypeCloudTrail,
+		LogTypeCloudTrailInsight,
+		LogTypeGuardDuty,
+		LogTypeS3ServerAccess,
+		LogTypeVPCFlow,
+	)
+
+	// Register AWS meta struct
```
alxarch, line 38: Registration of specific panther log meta for log types with an `AWS.`

internal/log_analysis/log_processor/parsers/awslogs/awslogs.go
```diff
@@ -0,0 +1,56 @@
...
+	logs.RegisterPrefixMeta("AWS", NewMeta)
+
+	// Register AWS fields
+	logs.RegisterField(KindAWSARN, "aws_arn", ARN)
```
alxarch, line 42: Registers default field factories for aws specfic fields to be used in other parsers, documentation, custom logs etc.

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
```diff
@@ -116,64 +129,49 @@ type CloudTrailSessionContextWebIDFederationData struct {
...
-func (event *CloudTrail) updatePantherFields(p *CloudTrailParser) {
-	event.SetCoreFields(p.LogType(), event.EventTime, event)
-
-	// structured (parsed) fields
-	event.AppendAnyIPAddressPtr(event.SourceIPAddress)
-	event.AppendAnyAWSAccountIdPtrs(event.RecipientAccountID)
-
+func (event *CloudTrail) PantherEvent() *logs.Event {
+	e := logs.NewEvent(TypeCloudTrail, event.EventTime.UTC(),
+		logs.HostnameP(event.SourceIPAddress),
```
alxarch, line 154: `logs.HostnameP` helper handles IP or Domain field creation transparently

internal/log_analysis/log_processor/parsers/gitlablogs/api.go
```diff
@@ -54,50 +62,27 @@ type API struct {
...
+func NewAPIParser() parsers.Interface {
+	return &APIParser{}
 }
 
-func (event *API) updatePantherFields(p *APIParser) {
-	event.SetCoreFields(p.LogType(), event.Time, event)
-	event.AppendAnyIPAddressPtr(event.RemoteIP)
+// Parse returns the parsed events or nil if parsing failed
+func (p *APIParser) Parse(log string) ([]*parsers.Result, error) {
+	return parsers.QuickParseJSON(&API{}, log)
```
alxarch, line 87: Use of `parsers.QuickParseJSON` helper to avoid boilerplate for all the cases where a single JSON input line needs to be parsed,validated and packed into a result.

internal/log_analysis/log_processor/parsers/logs/meta.go
```diff
@@ -0,0 +1,129 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+const PantherFieldPrefix = "p_"
+
+// All log parsers should extend from this to get standardized fields (all prefixed with 'p_' as JSON for uniqueness)
+// NOTE: It is VERY important that fields are added to END of the structure to avoid needed to re-build existing Glue partitions.
+//       See https://github.com/awsdocs/amazon-athena-user-guide/blob/master/doc_source/updates-and-partitions.md
+// nolint(lll)
+type Meta struct {
```
alxarch, line 37: `logs.Meta` is the struct to use to serialize and validate a `logs.Event` into JSON for `parsers.Result`. Custom meta implementations such as `AWS` can be registered with the `logs` package at init.

internal/log_analysis/log_processor/parsers/logs/meta.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Meta uses the logType prefix to construct a value containing all panther meta fields and info.
+// If a custom factory is registered for the event.LogType it is used to create the value.
+// Otherwise the default Meta struct is returned.
+// A custom MetaFactory can be registered for a log type prefix with `RegisterPrefixMeta`.
+func (e *Event) Meta() (interface{}, error) {
+	if e == nil {
+		return nil, errors.Errorf("nil event")
+	}
+	prefix := logTypePrefix(e.LogType)
```
alxarch, line 115: Every event checks it's log type prefix for a custom meta struct

internal/log_analysis/log_processor/parsers/metered.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// AvgParseTimeSeconds returns the average number of seconds spent parsing a single line
+func (s *Stats) AvgParseTimeSeconds() float64 {
+	return s.TotalTimeSeconds / s.NumLines
+}
+
+// Metered is a log parser wrapper that keeps track of metrics.
+// Metrics can be retrieved safely from different goroutines.
+type Metered struct {
```
alxarch, line 49: A wrapper around a parser to provide stats for parsing.

alxarch: It uses `AtomicFloat` to easily provide rates of stats. The same logic is used by well-estabished metrics packages such as Prometheus.

internal/log_analysis/log_processor/parsers/metered.go
```diff
@@ -0,0 +1,124 @@
...
+	if m, ok := parser.(*Metered); ok {
+		return m
+	}
+	return &Metered{
+		parser: parser,
+	}
+}
+
+// Parse implements parsers.Interface
+func (m *Metered) Parse(log string) (results []*Result, err error) {
```
alxarch, line 90: Parse increments stats counters and delegates parsing to the wrapped parser.

internal/log_analysis/log_processor/parsers/awslogs/guardduty.go
```diff
@@ -66,45 +76,30 @@ type GuardDutyService struct {
 // VPCFlowParser parses AWS VPC Flow Parser logs
```
rleighton, line 76: ???

alxarch: It's like that on master as well

3nvi:
Hey @alxarch is this PR still valid? I see that it has been open for quite a while and that it a lot of conflicts. I'm just wondering whether it's something that we want to eventually merge or if we are just keeping it  for reference

alxarch:
> Hey @alxarch is this PR still valid?

It's here to help discussions we are having on the general direction we will follow in upcoming changes.

> I see that it has been open for quite a while and that it a lot of conflicts.

The conflicts are mostly due to the PoC parsers that follow a different interface.

> I'm just wondering whether it's something that we want to eventually merge or if we are just keeping it for reference

This PR will be split into smaller (easier to review) ones and be merged with the master at some point. All parsers will be conflict-free at that point.

Issue #650, opened 2020-04-03
execute change set for stack panther-web is ROLLBACK_COMPLETE


kbroughton:
### Describe the bug
The following command fails.

mage -debug setup:swagger deploy

03:44:15	DEBUG	mage/deploy_template.go:280	deploy: ExecuteChangeSet for stack panther-web is now ROLLBACK_COMPLETE
03:44:15	FATAL	mage/deploy_template.go:290	execute change set for stack panther-web is ROLLBACK_COMPLETE:

In the CloudFormation Events I see

WebApplicationServer | CREATE_FAILED | Unable to assume the service linked role. Please verify that the ECS service linked role exists. (Service: AmazonECS; Status Code: 400; Error Code: InvalidParameterException; Request ID: 150230c7-0d46-4353-ad8d-7afd580f7c47)
-- | -- | --

It seems to be the following IAM role causing problems. Not sure why.

```
  WebApplicationServerExecutionRole:
    Type: AWS::IAM::Role
    Properties:
      AssumeRolePolicyDocument:
        Statement:
        - Effect: Allow
          Principal:
            Service: ecs-tasks.amazonaws.com
          Action: sts:AssumeRole

```

The first time running dev.sh I tried using the minimal role. It failed because of missing some permissions. Eventually I went back to using my admin role. After hitting this error, I deleted the panther-web stackset and tried again. The error was reproducible every time for me.

### Steps to reproduce
Follow the instructions at https://docs.runpanther.io/quick-start#step-3

./dev.sh
mage -debug setup:swagger deploy

### Expected behavior
completes successfully

### Environment

How are you deploying or using Panther?

commit c7c33378de3765c958cadc1b844eda101e2911aa
Author: Kostas Papageorgiou <kostas@runpanther.io>
Date:   Wed Mar 25 13:52:09 2020 +0100

    Fixing minor stability issues (#564)

austinbyers:
Thanks for the report!

- Not that it should matter, but what region were you deploying to?
- Did it work for you once you deployed with an admin role instead of the deployment role?

kbroughton:
deploying to us-east-1
I switched mid way between CFN creation after the first several fails to my admin creds. However, I did delete the stack each time. I will try a complete clean and redeploy next.

austinbyers:
I haven't been able to reproduce this so far, are you sure you're using the latest deployment role?

[...]

PR #771, opened 2020-04-23
v1.1.1 release - critical deploy fixes

austinbyers:
## Background

@nhakmiller found a number of issues when trying to deploy Panther to a fresh AWS account. Some of these had been previously reported but we had been unable to reproduce until now.

A few quick security fixes are included as well!

This will be the v1.1.1 release, a quick patch after v1.1.0 released yesterday

## Bug Fixes

- Change `./dev.sh` to use the newest development image. #685 added Terraform, which requires `unzip` to extract, so that PR already published a new deployment image. But I didn't realize I had to update the script as well to point to the new version. This fixes `mage setup` when running in Docker
- Update deployment role to have permission to create and use service roles as needed. Our dev accounts already had all the service roles we needed, so we didn't see permission errors associated with IAM service roles in a fresh account until now. Closes: #650
    - The ECS service role was created for me automatically when I had the correct permissions in the deploy role from the beginning, @nhakmiller to double-check this indeed solves the issue
- Fix the same swagger bug we've seen before (where imports are generated incorrectly). This time we literally just rewrite the import ourselves to guarantee it's correct no matter your go environment. Closes: #605 
- Another swagger bug: in docker, sometimes models are being deleted even when they shouldn't, causing `go build` to fail with 

```
go: finding module for package github.com/panther-labs/panther/api/gateway/remediation/models
internal/compliance/remediation_processor/main/lambda.go:32:2: no matching versions for query "latest"
```

`swagger generate` does not delete models once you remove them from the spec, so we were doing this ourselves based on timestamps. This logic isn't working in Docker, so this PR just removes the auto-deletion completely. Developers will have to manually remove generated api files if they delete a model.

## Security Fixes

- Require TLS 1.2 (and no earlier versions) on the load balancer listener

## Testing

- Deleted all AWSServiceRoles and Panther deployments from my dev account
- `cd /tmp`
- `git clone git@github.com:panther-labs/panther --branch austin-v1.1.1`
- `./dev.sh`
- export AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY with deployment user
- `mage setup deploy`
    - Verified I could login to the web app
- `mage teardown`
- `mage clean setup deploy`
    - To ensure a deploy with the service roles already existing still works

- `sslscan -u https://web-(accountid).us-east-1.elb.amazonaws.com` shows only TLS v1.2 is accepted:

```
  Supported Server Cipher(s):
Preferred TLSv1.2  128 bits  ECDHE-RSA-AES128-GCM-SHA256   Curve P-256 DHE 256
Accepted  TLSv1.2  128 bits  ECDHE-RSA-AES128-SHA256       Curve P-256 DHE 256
Accepted  TLSv1.2  128 bits  ECDHE-RSA-AES128-SHA          Curve P-256 DHE 256
Accepted  TLSv1.2  256 bits  ECDHE-RSA-AES256-GCM-SHA384   Curve P-256 DHE 256
Accepted  TLSv1.2  256 bits  ECDHE-RSA-AES256-SHA384       Curve P-256 DHE 256
Accepted  TLSv1.2  256 bits  ECDHE-RSA-AES256-SHA          Curve P-256 DHE 256
Accepted  TLSv1.2  128 bits  AES128-GCM-SHA256
Accepted  TLSv1.2  128 bits  AES128-SHA256
Accepted  TLSv1.2  128 bits  AES128-SHA
Accepted  TLSv1.2  256 bits  AES256-GCM-SHA384
Accepted  TLSv1.2  256 bits  AES256-SHA256
Accepted  TLSv1.2  256 bits  AES256-SHA
```


austinbyers:
We're going to add service role creation to the bootstrap stack as well

deployments/bootstrap.yml
```diff
@@ -647,6 +647,7 @@ Resources:
       LoadBalancerArn: !Ref PublicLoadBalancer
       Port: 443
       Protocol: HTTPS
+      SslPolicy: ELBSecurityPolicy-TLS-1-2-Ext-2018-06
```
jacknagz, line 650: 👏

deployments/auxiliary/cloudformation/panther-deployment-role.yml
```diff
@@ -18,7 +18,7 @@ AWSTemplateFormatVersion: 2010-09-09
 Description: IAM role for deploying Panther
 
 Metadata:
-  Version: v1.1.0
+  Version: v1.1.1
```
rleighton, line 21: One day this would be nice to automate setting.

PR #776, opened 2020-04-23
Bump template version in source api backend

nhakmiller:
## Background

Previous PR forgot to bump version key in backend.


austinbyers:
You can have the template match the repo version, i.e. 1.1.1 , if you like

That way, we'll know which release last modified the template

PR #778, opened 2020-04-24
Use new Pounce Table API in Panther

3nvi:
## Background

This PR bumps pounce and introduces a new way to "write" tables into Panther in a more declarative way. This is going to help us combat the multiple unforeseen scenarios where we want a "custom" behavior in a particular table.

Specifically, it will help add virtualization in the table rows (only render what the user sees in their screen) and stretching in the table cells (don't restrict width of a cell).

This involved multiple changes, but at the end of the day **everything should appear the way it used to and everything should work the way it used to**. 

## Changes

- Bump pounce to `0.18.x`
- Use the new Table API in every table within Panther
- Add `SeverityBadge` component 
- Remove all `columns.tsx` files
- The  table row is no longer clickable in its entirety
- The table columns have now a "dynamic" width according to their content
- Allow  to open links from table rows into new tabs (I think @nhakmiller wanted that at one point, but we couldn't support it in the previous table API)

**Apart from the last 3 bullets, everything else should look the same**

## Testing

All the changes have been tested locally extensively. For testing, **no deployment** is needed; one can test by simply running `npm i && npm run start`.

I just need a second pair of eyes in case I missed something. One nice way to test it is to go side-by-side with the deployed version and check the local version against it


web/src/pages/Users/ListUsersTable/ListUsersTableRowOptions/ListUsersTableRowOptions.tsx
```diff
@@ -18,14 +18,14 @@
...
-import { User } from 'Generated/schema';
 import useModal from 'Hooks/useModal';
 import { MODALS } from 'Components/utils/Modal';
 import { SIDESHEETS } from 'Components/utils/Sidesheet';
 import useSidesheet from 'Hooks/useSidesheet';
+import { UserDetails } from 'Source/graphql/fragments/UserDetails.generated';
 
 interface ListUsersTableRowOptionsProps {
-  user: User;
+  user: UserDetails;
```
alexmylonas, line 28: I like this! Should we do this for everything else too? Props from fragments instead of schema

kostaspap:
A couple of minor issues:
1. The name of the Rule in the Rule list displays the RuleID and not the Display Name
2. In the Alerts list, in `EVENTS` the number indentation is to the right, should be towards the left. I know it's tiny but all other fields seem to be indented to the left.

web/src/pages/Users/ListUsersTable/ListUsersTableRowOptions/ListUsersTableRowOptions.tsx
3nvi: Ideally yeah! We just shouldn't invest time to do **just that**. It should come gradually as part of other changes :)

3nvi:
> 1. The name of the Rule in the Rule list displays the RuleID and not the Display Name
100% right, thanks for catching this

> 2. In the Alerts list, in EVENTS the number indentation is to the right, should be towards the left. I know it's tiny but all other fields seem to be indented to the left.
This is intentional actually. Number fields should be aligned to the right for easier reading; It just happens that we have only 1 number field :D

PR #779, opened 2020-04-24
[gcplogs] AuditLog parsers

alxarch:
## Background

Implement GCP logs AuditLog parsers.

## Changes

- Adds `gcplogs` module
- Adds parsers for AuditLogs (Activity/System/Data)
- Modifies athena schema tool slightly to handle enum types (not necessary can revert)

## Testing

- `mage test:ci`


internal/log_analysis/log_processor/parsers/numerics/numerics.go
```diff
@@ -33,11 +34,14 @@ func (i *Integer) String() string {
 	return strconv.Itoa((int)(*i))
 }
 
+// FIXME: `nil` is not valid JSON data is this intended?
```
kostaspap, line 37: I am not sure this is intended. I see that it returns `nil` while it should be returning `null`. We should fix that. @rleightons is it possible this is a requirement of Athena or can we update this?

internal/log_analysis/log_processor/parsers/numerics/numerics.go
```diff
@@ -33,11 +34,14 @@ func (i *Integer) String() string {
...
 func (i *Integer) MarshalJSON() ([]byte, error) {
 	return ([]byte)(i.String()), nil
 }
 
+// FIXME: unmarshalling to a nil target fails silently here, this is a weird behavior probably not intentional
```
kostaspap, line 42: Yes, I think it is a miss. Would you mind updating it? I'd prefer that our numerics display the same behavior.

internal/log_analysis/log_processor/parsers/numerics/numerics.go
```diff
@@ -33,11 +34,14 @@ func (i *Integer) String() string {
...
 func (i *Integer) UnmarshalJSON(jsonBytes []byte) (err error) {
+	// FIXME: The strings.Trim call accepts invalid JSON input (ie `"42""``) as shown in the corresponding test case
```
kostaspap, line 44: I think it's probably more appropriate to just fail in that case. @rleighton any thoughts on this one?

internal/log_analysis/log_processor/parsers/numerics/numerics.go
```diff
@@ -46,3 +50,42 @@ func (i *Integer) UnmarshalJSON(jsonBytes []byte) (err error) {
...
+		if n := len(data) - 1; 0 <= n && n < len(data) && data[n] == '"' {
+			return data[:n]
+		}
+	}
+	return data
+}
+
+func (i *Int64) UnmarshalJSON(data []byte) error {
+	if i == nil {
+		return fmt.Errorf("nil target")
```
kostaspap, line 82: Can switch to `errors.New("nil target")` using `"github.com/pkg/errors"` (so it will include stacktrace)

internal/log_analysis/log_processor/parsers/gcplogs/gcplogs.go
```diff
@@ -0,0 +1,115 @@
...
+	SeverityDefault   LogSeverity = 0
+	SeverityDebug     LogSeverity = 100
+	SeverityInfo      LogSeverity = 200
+	SeverityNotice    LogSeverity = 300
+	SeverityWarning   LogSeverity = 400
+	SeverityError     LogSeverity = 500
+	SeverityCritical  LogSeverity = 600
+	SeverityAlert     LogSeverity = 700
+	SeverityEmergency LogSeverity = 800
+)
```
kostaspap, line 88: The enum will not change anything in the way the field is serialized. If anything, if GCP adds some new severity level, we would have to update our parsers to support it... I think you can remove this enum and just mark the `Severity` as an `int` number.

internal/log_analysis/log_processor/parsers/gcplogs/gcplogs.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if entry.LogName == nil {
+		return ""
+	}
+	name := *entry.LogName
+	if pos := strings.LastIndexByte(name, '/'); 0 <= pos && pos < len(name) {
+		return name[pos+1:]
+	}
+	return ""
+}
```
kostaspap, line 59: Looking at this PR, I understand the purpose of this method. However for the future maintainer of this code it might not be so obvious. Could you add some documentation about this method, why we need it and maybe a sample return value.

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
```diff
@@ -0,0 +1,181 @@
...
+func (p *auditLogParser) Parse(log string) ([]*parsers.PantherLog, error) {
+	entry := LogEntryAuditLog{}
+	if err := jsoniter.UnmarshalFromString(log, &entry); err != nil {
+		return nil, err
+	}
+	if id := entry.LogID(); id != p.LogID {
+		return nil, fmt.Errorf("invalid LogID %q != %q", id, p.LogID)
+	}
+	ts := entry.Timestamp
+	if ts == nil || ((*time.Time)(ts)).IsZero() {
```
kostaspap, line 112: If upstream source really sent timestamp of UnixTime 0, we should use that, we shouldn't alter incoming log values.

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
```diff
@@ -0,0 +1,181 @@
...
+		ts = entry.ReceiveTimestamp
+	}
+	entry.SetCoreFields(p.Type, ts, &entry)
+	if entry.HTTPRequest != nil {
+		entry.AppendAnyIPAddressPtr(entry.HTTPRequest.RemoteIP)
+	}
+	if meta := entry.Payload.RequestMetadata; meta != nil {
+		entry.AppendAnyIPAddressPtr(entry.Payload.RequestMetadata.CallerIP)
```
kostaspap, line 120: * I assume this line could be: 
```
entry.AppendAnyIPAddressPtr(meta.CallerIP)
```

* LogEntry also has the `HTTPRequest.RemoteIP` and `HTTPRequest.ServerIP` which we should add to the list of IPs.

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
```diff
@@ -0,0 +1,181 @@
...
+	MethodName         *string             `json:"methodName" validate:"required" description:"The name of the service method or operation. For API calls, this should be the name of the API method."`
+	ResourceName       *string             `json:"resourceName" validate:"required" description:"The resource or collection that is the target of the operation. The name is a scheme-less URI, not including the API service name."`
+	NumResponseItems   *numerics.Int64     `json:"numResponseItems,omitempty" description:"The number of items returned from a List or Query API method, if applicable."`
+	Status             *Status             `json:"status,omitempty" description:" The status of the overall operation."`
+	AuthenticationInfo *AuthenticationInfo `json:"authenticationInfo,omitempty" description:"Authentication information."`
+	AuthorizationInfo  []AuthorizationInfo `json:"authorizationInfo,omitempty" description:"Authorization information. If there are multiple resources or permissions involved, then there is one AuthorizationInfo element for each {resource, permission} tuple."`
```
kostaspap, line 137: We need the `validate:"omitempty,dive"` tag. 
https://godoc.org/gopkg.in/go-playground/validator.v9#hdr-Dive

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// Parse implements parsers.LogParser interface
+func (p *auditLogParser) Parse(log string) ([]*parsers.PantherLog, error) {
+	entry := LogEntryAuditLog{}
+	if err := jsoniter.UnmarshalFromString(log, &entry); err != nil {
+		return nil, err
+	}
+	if id := entry.LogID(); id != p.LogID {
+		return nil, fmt.Errorf("invalid LogID %q != %q", id, p.LogID)
```
kostaspap, line 109: Should use `errors.Errorf` from `"github.com/pkg/errors"`

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
```diff
@@ -0,0 +1,181 @@
...
+
+type AuditLogActivityParser struct{}
+type AuditLogSystemParser struct{}
+type AuditLogDataParser struct{}
+
+const AuditLogActivityLogID = "cloudaudit.googleapis.com%2Factivity"
+const AuditLogDataLogID = "cloudaudit.googleapis.com%2Fdata_access"
+const AuditLogSystemLogID = "cloudaudit.googleapis.com%2Fsystem_event"
+
+var _ parsers.LogParser = (*auditLogParser)(nil)
```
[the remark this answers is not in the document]
alxarch: I agree that it's simpler with a single log type. We should evaluate if searching for example `SystemEvents` is crucial enough to 'partition' the data by creating separate table. From the docs it's clear that `Activity` and `SystemEvent` are always logged by google. `DataAccess` is only logged if enabled at the project level. I can merge the separate tables but keep a check to verify the `LogID` to distinguish from other GCP log types.

internal/log_analysis/log_processor/parsers/numerics/numerics.go
alxarch: Should this be done for all errors? I've been using `errors.New` without checking the imported package name all along.

internal/log_analysis/log_processor/parsers/gcplogs/gcplogs.go
alxarch: Yep, will add comments and explanation from the GCP logs docs

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
alxarch: I wasn't sure if we'll end up storing `HTTPRequest` as `jsoniter.RawMessage` in the end. Any thoughts on this concerning table sizes etc?

kostaspap: Nope, no size concern for this one. If the IP is in the log, it should be in the "Any IP address" column.

internal/log_analysis/log_processor/parsers/numerics/numerics.go
kostaspap: Yup, should be done for all errors. I am guessing we might have missed this in a few(potentially a lot) cases. Feel free to update any occurrence you encounter.

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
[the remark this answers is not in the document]
kostaspap: I agree, both approach are valid. I think however that it's better just to start with one table and we can update our code to split it to multiple tables depending on users feedback. 

> I can merge the separate tables but keep a check to verify the LogID to distinguish from other GCP log types.

Sounds good!

internal/log_analysis/log_processor/parsers/numerics/numerics.go
rleighton: We should have proper JSON , looks like a bug.

internal/log_analysis/log_processor/parsers/numerics/numerics.go
rleighton: Agree

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
alxarch: I kept only the nil check and fall back to `ReceiveTimestamp` if it's `nil`.

NOTE: `Time#IsZero()` is not the zero unix timestamp. 
> IsZero reports whether t represents the zero time instant, January 1, year 1, 00:00:00 UTC. 
Which is the first day of year 0, 2020 years ago :)

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
alxarch: Swapped all instances of `"errors"` and `"fmt.Errorf"` with the github pkg.

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
alxarch: Fixed

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
alxarch: Fixed

internal/log_analysis/log_processor/parsers/numerics/numerics.go
alxarch: Switched all instances of `errors` and `fmt.Errorf` to github package

internal/log_analysis/log_processor/parsers/numerics/numerics.go
alxarch: Fixed

internal/log_analysis/log_processor/parsers/numerics/numerics.go
alxarch: Fixed

internal/log_analysis/log_processor/parsers/numerics/numerics.go
alxarch: Fixed

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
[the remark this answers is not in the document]
alxarch: Fixed

internal/log_analysis/log_processor/parsers/gcplogs/gcplogs.go
alxarch: Fixed

internal/log_analysis/log_processor/parsers/gcplogs/gcplogs.go
```diff
@@ -0,0 +1,105 @@
...
+	Referer       *string         `json:"referer" validate:"required" description:"The referer URL of the request"`
+	// FIXME Parse duration?
+	Latency        *string         `json:"latency" validate:"required" description:"The request processing latency in seconds on the server, from the time the request was received until the response was sent."`
+	CacheLookup    *bool           `json:"cacheLookup,omitempty"  description:"Whether or not a cache lookup was attempted."`
+	CacheHit       *bool           `json:"cacheHit,omitempty"  description:"Whether or not an entity was served from cache (with or without validation)."`
+	CacheValidated *bool           `json:"cacheValidatedWithOriginServer,omitempty" description:"Whether or not an entity was served from cache (with or without validation)."`
+	CacheFillBytes *numerics.Int64 `json:"cacheFillBytes,omitempty" description:"Whether or not an entity was served from cache (with or without validation)."`
+	Protocol       *string         `json:"protocol" validate:"required" description:"Protocol used for the request."`
```
kostaspap, line 97: Most of the fields here are marked as `required`. Is this true or are they required? I was checking the documentation and I couldn't see where it was mentioned that they were required: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest

internal/log_analysis/log_processor/parsers/numerics/numerics.go
```diff
@@ -33,16 +34,89 @@ func (i *Integer) String() string {
...
+	n, err := strconv.ParseInt(string(data), 10, 64)
+	if err != nil {
+		return err
+	}
+	if n > int64(MaxInt) {
+		return errors.Errorf("integer overflow")
+	}
+	if n < int64(MinInt) {
+		return errors.Errorf("integer underflow")
 	}
```
kostaspap, line 77: Could you add unit test cases for these scenarios? I would personally expect that parsing might fail for this scenarios with an error...But might be the case. Better unit test :)

internal/log_analysis/log_processor/parsers/numerics/numerics.go
```diff
@@ -33,16 +34,89 @@ func (i *Integer) String() string {
...
+
+func unquoteJSON(data []byte) []byte {
+	if len(data) > 1 && data[0] == '"' {
+		data = data[1:]
+		if n := len(data) - 1; 0 <= n && n < len(data) && data[n] == '"' {
+			return data[:n]
+		}
+	}
+	return data
+}
```
kostaspap, line 108: Better move private method in the bottom - below from where it is used (there is one usage in line 115)

internal/log_analysis/log_processor/parsers/gcplogs/gcplogs.go
alxarch: From what I have seen from the samples these are filled  from HTTP LoadBalancer service and have all fields apart from the cache stuff. I can revert them all to `omitempty`. But the only one I think could be missing would be the Referrer field. I will try to find samples to include in the tests.

internal/log_analysis/log_processor/parsers/numerics/numerics.go
alxarch: It's now used by both numerics in the latest commits, will move to bottom

internal/log_analysis/log_processor/parsers/gcplogs/gcplogs.go
kostaspap: I see. My concern is that if a field is marked as "required" while it actually is not, it means that we will fail to process this event as it will be not accepted as "valid" GCP log. 
The downside of marking a field as optional while it is not could lead to wrongly classifying a log to be of the wrong log type... 
It seems that the latter is not so possible since GCP logs already have some distinct fields. The former is a dangerous scenario that we should really try to avoid... Unless we are sure that some fields are required (GCP documentation should clarify that), it is preferable to just marked them as optional.

internal/log_analysis/log_processor/parsers/numerics/numerics.go
alxarch: I added unit test for these but as it is platform dependent the test fails in `strconv.ParseInt` so it is not added in coverage

internal/log_analysis/log_processor/parsers/numerics/numerics.go
alxarch: Fixed

internal/log_analysis/log_processor/parsers/numerics/numerics.go
alxarch: I have no idea how to properly test this

internal/log_analysis/log_processor/parsers/gcplogs/gcplogs.go
alxarch: I marked all non-URL based fields as optional. Should I mark them all?

internal/log_analysis/log_processor/parsers/numerics/numerics.go
kostaspap: Shouldn't it be possible to test with something like:
```
i.UnmarshalJSON(([]byte)(`1234567890123456789012345678....`))
```
and 
```
i.UnmarshalJSON(([]byte)(`-1234567890123456789012345678....`))
```
?

internal/log_analysis/log_processor/parsers/gcplogs/gcplogs.go
kostaspap: I think the documentation doesn't explicitly mention anything as required. I'd say mark them all to be on the safe side - the GCP logs include the log type field that is pretty unique

internal/log_analysis/log_processor/parsers/numerics/numerics.go
alxarch: Nope, because the test fails during `strconv.ParseInt` since `ParseInt` uses `int64` and since the value doesn't fit in `int64` either and since `int` *is* `int64` the overflow/underflow ifs would never be triggered. Still casting to `int` from `int64` should be checked.

kostaspap: >  because the test fails during strconv.ParseInt since ParseInt uses int64 and since the value doesn't fit in int64 either and since int is int64 the overflow/underflow ifs would never be triggered. 

So, we don't need the checks above `(n> int64(MaxInt)` and `n < int64(MinInt)` as well as the associated constants, correct?

internal/log_analysis/log_processor/parsers/gcplogs/gcplogs.go
alxarch: It's a mess, on `system_event` samples I have there are some other undocumented or rather deprecated fields. Fortunately the names don't collide so I put them in with `omitempty` and reverted a lot of other fields to `omitempty` also. The *actual* rule that detects an `AuditLog` is the `eq` on `protoPayload.@type`. That should at least be consistent.

internal/log_analysis/log_processor/parsers/numerics/numerics.go
kostaspap: I see your point. Worth putting a comment on the checks to say that it e.g. it will always succeed in `x64` architectures

alxarch: As a bonus knowledge on the topic `strconv.Atoi` could work but doesn't since it a branch for when the platform size is `int32` but falls back to blindly converting the `int64` to `int` without checks which is wrong. This would be considered a bug in any serious systems language but this is `go` we're talking about...

alxarch: I avoid messing with `int` and `usize` in any part of the code that should be used outside go (ie RDBMS, binary protocols etc.) I just stick to  `int64` and maybe `int32` if there are size concerns and I know that the value will not be over/underflow (again with checks). That being said I think I know how to handle the specific case a bit more elegantly. Will commit a change.

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
```diff
@@ -0,0 +1,190 @@
...
+	Fragment *string            `json:"fragment,omitempty" description:"The HTTP URL fragment. No URL decoding is performed."`
+	Time     *timestamp.RFC3339 `json:"time,omitempty" description:"The timestamp when the destination service receives the first byte of the request."`
+	Size     *int64             `json:"size,omitempty" description:"The HTTP request size in bytes. If unknown, it must be -1."`
+	Protocol *string            `json:"protocol,omitempty" description:"The network protocol used with the request, such as 'http/1.1', 'spdy/3', 'h2', 'h2c', 'webrtc', 'tcp', 'udp', 'quic'."`
+	Reason   *string            `json:"reason,omitempty" description:"A special parameter for request reason. It is used by security systems to associate auditing information with a request."`
+	Auth     *Auth              `json:"auth,omitempty" description:"A special parameter for request reason. It is used by security systems to associate auditing information with a request."`
+}
+
+type Peer struct {
+	IP         *string           `json:"ip,omitempty" description:"The IP address of the peer."`
```
rleighton, line 165: This should be extracted into AppendAnyIPAddressPtr()

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
```diff
@@ -0,0 +1,190 @@
...
+	RequestAttributes       *RequestAttributes `json:"requestAttributes,omitempty" description:"Request attributes used in IAM condition evaluation. This field contains request attributes like request time and access levels associated with the request."`
+	DestinationAttributes   *Peer              `json:"destinationAttributes,omitempty" description:"The destination of a network activity, such as accepting a TCP connection."`
+}
+type RequestAttributes struct {
+	ID       *string            `json:"id,omitempty" description:"The unique ID for a request, which can be propagated to downstream systems."`
+	Method   *string            `json:"method,omitempty" description:"The HTTP request method, such as GET, POST."`
+	Headers  map[string]string  `json:"headers,omitempty" description:"The HTTP request headers. If multiple headers share the same key, they must be merged according to the HTTP spec. All header keys must be lowercased, because HTTP header keys are case-insensitive"`
+	Path     *string            `json:"path,omitempty" description:"The HTTP URL path."`
+	Host     *string            `json:"host,omitempty" description:"The HTTP request host header value."`
```
rleighton, line 153: Should this be extracted into panther log domain field?

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
```diff
@@ -0,0 +1,190 @@
...
+	ResourceName       *string             `json:"resourceName,omitempty" description:"The resource or collection that is the target of the operation. The name is a scheme-less URI, not including the API service name."`
+	NumResponseItems   *numerics.Int64     `json:"numResponseItems,omitempty" description:"The number of items returned from a List or Query API method, if applicable."`
+	Status             *Status             `json:"status,omitempty" description:" The status of the overall operation."`
+	AuthenticationInfo *AuthenticationInfo `json:"authenticationInfo,omitempty" description:"Authentication information."`
+	AuthorizationInfo  []AuthorizationInfo `json:"authorizationInfo,omitempty" validate:"omitempty,dive" description:"Authorization information. If there are multiple resources or permissions involved, then there is one AuthorizationInfo element for each {resource, permission} tuple."`
+	RequestMetadata    *RequestMetadata    `json:"requestMetadata,omitempty" description:"Metadata about the request"`
```
kostaspap, line 112: I'm seeing that the `RequestMetadata` object is rather complex. Considering the following:
1. Maintaining such a complex struct can be really challenging moving forward. My concern is the complexity of maintaining the actual Glue table schema. 
2. `Request` is already mapped as a `jsoniter.RawMessage`

I think it's better to just store the `RequestMetadata` field as `jsoniter.RawMessage`. 

There is a lot of work to create the `RequestMetadata` struct though and all the nested structures. This work can be used to eventually extract useful fields to common fields. I think it's fine to just leave the structs as is, just put a comment on top of them and make them package visible only (instead of public visible)

alxarch: As the problematic ones seem to be the extra fields inside `RequestMetadata` I did that for `Peer` and `RequestAttributes` objects and their dependencies but kept the `CallerIP` stuff.
The issue now seems to be that by removing the structs we will need to use an `extractor` for the IP in `Peer`. Should I implement that? I don't have any log sample that *actually* has the `Peer` struct populated. We're chasing our tails here.

kostaspap: Looks good. 
I think we can implement this as part of a separate issue, I have created #801

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
kostaspap: I have created a separate issue to keep track of this: https://github.com/panther-labs/panther/issues/801

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
kostaspap: I have created a separate issue to keep track of this: https://github.com/panther-labs/panther/issues/801

PR #783, opened 2020-04-24
Minor fix for AlertId on breadcrumb

### LINKED ISSUE #751: Alert details breadcrumb capitalizes first letter
### Describe the bug

It's more like a "nuance" than a bug. The top of the Alert Details page capitalizes the first letter of the Alert ID. 

### Steps to reproduce

1. Go to `Log Analysis` -> `Alerts`
1. Click on any alert. 
1. See the AlertID in the top of the page (see screenshot below)

### Expected behavior

AlertID should be lowercase. 

### Environment

- Panther version or commit: https://github.com/panther-labs/panther/commit/d81bdd6307cffc211f921d8dcd987fb39b0141cd
- OS: MacOS 10.15.3
- Browser: Chrome 81.0.4044.113 

### Screenshots

As you can see below, the first `F` is capitalized.

alexmylonas:
## Background

AlertId on breadcrumb is capitalized & fails to pass to `isHash` test.

Closes #751
## Changes

- Lowercase alertId before test and after shortening

## Testing

- Locally


web/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -22,7 +22,8 @@ import { isGuid, capitalize, shortenId, isHash } from 'Helpers/utils';
 import { Link as RRLink } from 'react-router-dom';
 import useRouter from 'Hooks/useRouter';
 
-const transformBreadcrumbText = text => (isHash(text) ? shortenId(text) : text);
+const transformBreadcrumbText = text =>
+  isHash(text.toLowerCase()) ? shortenId(text).toLowerCase() : text;
```
kostaspap, line 26: Is there a reason why we try to lowercase only if it's hash? How do we do upper case first letter at first place, is it needed?

alexmylonas: The breadcrumb generator is a bit generic to handle every breadcrumb on UI, having capitalized first letter is easier on the eye.
A couple of screenshots without capitlization and with lowercase for all! 

Screenshot 2020-04-24 at 2 31 37 PM
Screenshot 2020-04-24 at 2 32 35 PM

kostaspap: Ok, makes sense. Fine to ship from me, I will let @3nvi  just the actual code ^^

PR #784, opened 2020-04-27
In case Alert with the specified ID does not exist, return DoesNotExist error

### LINKED ISSUE #768: Invalid AlertID still returns data
### Describe the bug

When querying `getAlert` with an invalid AlertID, I expect to get a `null` back or (not preferable) an error. Instead I get the following response back

```
alertId: ""
ruleId: ""
title: ""
creationTime: "0001-01-01T00:00:00Z"
eventsMatched: 0
updateTime: "0001-01-01T00:00:00Z"
eventsLastEvaluatedKey: "eyJsb2dUeXBlVG9Ub2tlbiI6e319"
events: []
```

### Steps to reproduce

1. Invoke `getAlert` supplying an invalid `alertId` parameter
2. Observe the response

### Expected behavior

Return `null` or error out

### Environment

- Panther version or commit: [e.g. v0.1.1]
- OS: [e.g. MacOS 10.15.3]
- Browser: [e.g. Chrome 80.0.3987.87, Safari Version 13.0.5 (15608.5.11)]

kostaspap:
## Background

In case Alert with the specified ID does not exist, return DoesNotExist error.

Fixes #768 

## Changes

- Updated Alerts API
- Updated Alerts table

## Testing

- Unit testing
- Deployed in my account and verified it works as expected.


kostaspap:
@alexmylonas @3nvi I deployed to my account and I verified that the Alerts API now returns `nil` Alert and an `DoesNotExist` error in case a non-existing AlertID is requested. This all according to #768  description. 

When tested it on my dev account I noticed that with this change, the UI breaks
Screen Shot 2020-04-27 at 3 34 17 PM

It seems that the frontend is not taking care of this case properly. I think it might be more appropriate that you update the FE UI to handle the new Alerts API response in this PR?

3nvi:
Yeah we need to alter some things, since we adjusted the UI code to the response itself. Keep this branch open and I'll make the necessary updates

3nvi:
@kostaspap I don't have a way of parsing the error in order to understand what kind of error it was. That means that I can't distinguish an actual error from a `DoesNotExist` error.

We should either:

- Add an error code (currently I'm getting`Lambda:unhandled`)
- Remove the error altogether and just count on the fact that the data would be `nil`

kostaspap:
> * Remove the error altogether and just count on the fact that the data would be `nil`

You can count on the fact that the data is `nil`.

3nvi:
> > * Remove the error altogether and just count on the fact that the data would be `nil`
> 
> You can count on the fact that the data is `nil`.

By default in GraphQL, data is `nil` whenever there is any error. If I depend on that, I'll be masking other errors and translate them all to "404" (although the actual error might be unrelated)

kostaspap:
Closing in favor of https://github.com/panther-labs/panther/pull/829

PR #786, opened 2020-04-27
Source Onboarding UX Improvements

### LINKED ISSUE #598: Source Onboarding UX Improvements
### Description

While adding/removing/editing many cloudsec/log analysis sources, I identified a couple of UX and functionality improvements. 

### Acceptance Criteria
- **Users should be able to quickly rename a source.** Currently, users have to download templates and click all the way through just to fix a typo.
- **Users should not be blocked to download/run CloudFormation to add a source.** They should be able to click through to the validation because sometimes roles are created out of band with either Terraform or with nested stacks.
- S3 bucket names should be [validated](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html) to prevent ARNs from accidentally being pasted in the field.

alexmylonas:
## Background

When trying to edit/add new source in Log Analysis or Cloud Security, we require users to download a template. That can be optionally, since users could create and upload templates they created.

To address this: 
- Added a "Skip" button on the wizard so users can proceed to validation without downloading the pre-made template. 
- Added an S3 Bucket name regex to validate format and prevent copying pasting of ARNs

Closes #598,#800

**Update:** Removed skip button and just enable Next button immediately + updated content based on #800 

_Note: Test cases for new regex [here](https://regex101.com/r/ovY9Hf/4)_

## Changes

- Added `S3_BUCKET_NAME_REGEX`
- Added Skip action in `WizardPanelWrapper`
- Added Skip button in Compliance and Log Wizard (screenshots)

## Testing

- Locally

Screenshot 2020-04-29 at 9 07 08 PM


alexmylonas:
I think it makes some sense to move the steps 2-5 to next page for Log Analysis. What do you think?

3nvi:
> I think it makes some sense to move the steps 2-5 to next page for Log Analysis. What do you think?

Let's not change anything now, since we have an updated & verified new flow coming

alexmylonas:
> I think just "enabling" the next button there would be the simpler solution (than adding an extra button). Perhaps though adding an explicit `Skip` might be better. What do you think?

If we don't need any user action, this screen seems redundant, i am thinking as the "Skip" button as a confirmation that the user read the instruction but has already done it. Not a strong opinion whatsoever though. 

> If we decide to keep the skip button, it should be in the bottom left corner, away from the other 2. I'm approving, but let's figure it out before merging .

Agree with that!


> P.S. @alexmylonas The related issue also mentions S3 bucket front-end validation. Let's add this here

Yeap, already implemented on `LogSourceWizard.tsx` L63!

jacknagz:
@alexmylonas the other case I identified was leading spaces in the bucket, we should make sure that's not part of the regular expression

alexmylonas:
Removed `Skip` button based on feedback and enable `Next` by default
> @alexmylonas the other case I identified was identifying and preventing leading spaces in the bucket name, which can cause issues. We should make sure that it's accounted for in the regular expression

@jacknagz Yeap, that regex prevents that case as well! @nhakmiller could you take a quick look at this test cases to see if something obvious. https://regex101.com/r/ovY9Hf/4/tests

nhakmiller:
> @jacknagz Yeap, that regex prevents that case as well! @nhakmiller could you take a quick look at this test cases to see if I am missing something obvious. https://regex101.com/r/ovY9Hf/4/tests

Technically `192.0.0` is a valid bucket name, because it's not an IP address. Any of these approaches would be valid:

- Ignore this inconsistency, I think it will be extremely rare to enter something like `100.100.10` as a bucket name
- Update the regex to handle this edge case, which sounds difficult and not worth the time IMO but would be the most correct
- simplify the regex to just not care if someone adds an IP address, technically it's not a valid bucket name but it's a somewhat rare edge case and really just validating everything else is a big improvement already

alexmylonas:
> * simplify the regex to just not care if someone adds an IP address, technically it's not a valid bucket name but it's a somewhat rare edge case and really just validating everything else is a big improvement already

@nhakmiller I followed this suggestion for the moment!

PR #793, opened 2020-04-28
Enable Filtering by 'enabled' field for Policies/Rules

### LINKED ISSUE #666: Filter rules/policies by whether they are enabled

I want to be able to filter rules/policies based on whether they are currently enabled

### Describe your use cases

* Filtering to just show enabled makes it easier to see what analysis is actually running
* Filtering to just show disabled makes it easy to find rules which need adjustment before they can be re-enabled

Weekly

### Implementation

This functionality already exists in the backend: https://github.com/panther-labs/panther/blob/master/api/gateway/analysis/api.yml#L307-L310

We just need to offer it as an additional filter option in the web app

alexmylonas:
## Background

We allow users to filter Rules and Policies by whether are enabled or not. Functionality is already implemented in the backend.

We added field `Enabled` on Filters component with `Yes` and `No` options.

### Describe use cases
Filtering to just show enabled makes it easier to see what analysis is actually running
Filtering to just show disabled makes it easy to find rules which need adjustment before they can be re-enabled

Closes #666 
## Changes

- Added `Enabled` field on Filtering component for Policies
- Added `Enabled` field on Filtering component for Rules
- Fixed a minor issue on pounce & bump to `pouncejs@v0.18.5`

Check screenshots

## Testing

- Locally

Screenshot 2020-04-28 at 3 35 44 PM
Screenshot 2020-04-28 at 3 36 06 PM


3nvi:
> Looks good! There is still an option to see all rules/policies without the "enabled" filter, right? By default, we show both enabled and disabled, then the user can filter down to one or the other?

Actually @alexmylonas Austin is right. We should provide a third option to "reset" the filter. The options should be like "Enabled", "Disabled", "All" or something similar (wording can change)

alexmylonas:
> Actually @alexmylonas Austin is right. We should provide a third option to "reset" the filter. The options should be like "Enabled", "Disabled", "All" or something similar (wording can change)

"All" options has the same result as not having this filter and that case also is needed on `Severity`  and `Status` filters. 

I added a solution but doesn't seem very clean, should we update the underlying pounce component to handle this cleanly?

3nvi:
> > Actually @alexmylonas Austin is right. We should provide a third option to "reset" the filter. The options should be like "Enabled", "Disabled", "All" or something similar (wording can change)
> 
> "All" options has the same result as not having this filter and that case is also needed on `Severity` and `Status` filters.
> 
> @3nvi I added a solution but doesn't seem very clean, should we update the underlying pounce component to handle this cleanly?

The thing is that there is no easy way to do this on Pounce as well. A combobox normally has values to switch from; Pounce shouldn't care if those values include an empty string or not. This is our business logic which shouldn't interfere with something Pounce doesn't know about.

I'd say the solution seems ok, just make sure that `itemToString` shows a "proper text" for when the option is `''`. Also, we can make this particular "resetting option" be `undefined` instead of `''` (so it would be `[undefined, true, false]` which makes more sense to me since `undefined` is actually what it says it is - not defined).

Also let's make sure we add this on both Rules and Policies

PR #797, opened 2020-04-28
Update deployment role

nhakmiller:
## Background

The main functional change is adding some CloudTrail permissions so that the deployment role can succesfully deploy an account where `SelfOnboard` and `EnableCloudTrail` are both set equal to `true` in the `deployments/panther_config.yml` file. I also merged two permission blocks that I didn't actually understand why they were separate, but I can revert that change if need be.

## Changes

- Added `cloudtrail:CreateTrail` and `cloudtrail:DescribeTrails` actions to the `Resource: *` permission
- Created a new permission that grants `cloudtrail:DeleteTrail`, `cloudtrail:StartLogging`, `cloudtrail:StopLogging`, and `cloudtrail:AddTags` on the CloudTrail that is created by Panther for self-onboarding
- Condensed two separate permissions that allowed various actions on `Resource: *` into one permission

## Testing

- Did a fresh deployment with this role where CloudTrailEnabled and SelfOnboard were both equal to `true`

### DIFF
--- a/deployments/auxiliary/cloudformation/panther-deployment-role.yml
+++ b/deployments/auxiliary/cloudformation/panther-deployment-role.yml
@@ -57,37 +57,11 @@ Resources:
               - athena:*
               - cloudformation:Describe*
               - cloudformation:List*
+              - cloudtrail:DescribeTrails
+              - cloudtrail:CreateTrail
               - cloudwatch:*
               - cognito-idp:*
               - dynamodb:List*
-              - ecr:GetAuthorizationToken
-              - ecs:*
-              - events:*
-              - glue:*
-              - guardduty:CreatePublishingDestination
-              - guardduty:ListDetectors
-              - kms:CreateKey
-              - kms:List*
-              - lambda:*EventSourceMapping
-              - lambda:List*
-              - logs:*
-              - sns:List*
-              - sqs:List*
-              - states:CreateStateMachine
-              - states:TagResource
-              - states:UntagResource
-            Resource: '*'
-          - Effect: Allow
-            Action: cloudformation:*
-            Resource:
-              - !Sub arn:${AWS::Partition}:cloudformation:*:${AWS::AccountId}:stack/panther-*
-              - !Sub arn:${AWS::Partition}:cloudformation:*:${AWS::AccountId}:stackset/panther-*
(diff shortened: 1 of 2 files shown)


deployments/auxiliary/cloudformation/panther-deployment-role.yml
```diff
@@ -123,7 +97,40 @@ Resources:
               - ec2:UpdateSecurityGroupRuleDescriptionsEgress
               - ec2:UpdateSecurityGroupRuleDescriptionsIngress
               - elasticloadbalancing:*
+              - ecr:GetAuthorizationToken
```
austinbyers, line 100: I just had the EC2/networking stuff separate because it was easier to read, this is fine

deployments/auxiliary/cloudformation/panther-deployment-role.yml
```diff
@@ -123,7 +97,40 @@ Resources:
...
+              - states:UntagResource
             Resource: '*'
+          - Effect: Allow
+            Action: # permissions for handling self-onboarding CloudTrail
+              - cloudtrail:DeleteTrail
+              - cloudtrail:StartLogging
+              - cloudtrail:StopLogging
+              - cloudtrail:AddTags
+            Resource:
+              - !Sub arn:${AWS::Partition}:cloudtrail:${AWS::Region}:${AWS::AccountId}:trail/panther-cloudtrail-${AWS::Region}
```
austinbyers, line 124: Nit: single resource could be on the same `Resource: ` line

austinbyers:
Alternatively, we can bump all aux template versions as part of each release (I'm not really sure the best way to handle the versioning)

deployments/auxiliary/cloudformation/panther-deployment-role.yml
austinbyers: I like it better consolidated

nhakmiller: Ah I was confused because I was assuming that since it was in a separate block the "These are the actions where we cannot specify a resource" comment might not apply to these actions. I'll leave them condensed since you say it's fine.

nhakmiller:
> Please (a) bump the template version (or it will overwrite the previous version) and (b) update the tf version of the deployment role

Ok I did these. Although I could not find a version for the terraform template, I assume its not applicable?

nhakmiller:
Template uploaded.

PR #798, opened 2020-04-28
HTML escape user name in password reset email

austinbyers:
## Background

If a username embeds HTML, e.g. a first name of `<a href="https://google.com">click</a>`, that HTML will be rendered in the password reset email:

Screen Shot 2020-04-28 at 11 20 16 AM

As a result, _if you can modify someone's first name in Panther_, it's possible to capture the password reset token and takeover the account by triggering a password reset email for that user! Clever

This is not a high severity issue because if you have the ability to change user details besides your own, you're already a user admin anyway and can take over their account by just changing their email, for example.

Nonetheless, Panther should not be forwarding unsanitized HTML in emails!

## Changes

- `hmtl.EscapeString` for user names in password reset email

A future PR will consider denying/replacing special characters `<>"'&` (the html escape set that golang uses) from all user input at the api level, but that's a bigger change that needs more planning

## Testing

With the HTML escaping, the email renders the name literally, as you would see it in Panther:

Screen Shot 2020-04-28 at 11 18 19 AM


internal/core/users_api/api/cognito_trigger.go
```diff
@@ -71,7 +72,7 @@ func handleForgotPassword(event *events.CognitoEventUserPoolsCustomMessage) (*ev
 	}
 
 	event.Response.EmailMessage = fmt.Sprintf(passwordResetTemplate,
-		givenName,
+		html.EscapeString(givenName),
```
kostaspap, line 76: Should we just restrict the characters that we accept in the user name? Just thinking where else this can can be rendered and what would be the implications there.

austinbyers: Like I said in the PR description:

> A future PR will consider denying/replacing special characters <>"'& (the html escape set that golang uses) from all user input at the api level, but that's a bigger change that needs more planning

Yes, we should consider doing that as well, but this is a quick fix to solve the immediate concern

austinbyers: And we should always have both, since it's theoretically possible for the user name to be set directly in AWS

PR #803, opened 2020-04-29
[mage] Mage ci task queue

alxarch:
## Background

Running `mage test:ci` on my laptop overwhelms the CPU and grinds everything to a halt for almost 10 minutes on each run. This might not be noticeable on better hardware but introducing a simplistic task queue to run `NumCPU - 1` tasks at a time greatly improves the overall time and responsiveness of the system.

## Changes

- Adds a `runTask` helper in `tools/mage` that enforces a limit on parallel running tasks based on `runtime.NumCPU` 
- Use the `runTask` to limit concurrent tasks running on `test:ci` task

## Testing

- Ran `mage test:ci`


alxarch:
Let's wait to see how long it takes on CircleCI :)

tools/mage/util.go
```diff
@@ -44,7 +44,33 @@ const (
...
+	if n > 1 {
+		return n - 1
+	}
+	return 1
+}()
+
+// Queue limiting concurrent tasks when using `runTask`
+var queue = make(chan struct{}, maxWorkers)
+
+// Ugly task queue hack to limit concurrent tasks
```
austinbyers, line 63: `build:lambda` and `build:tools` already use explicit worker pools:

https://github.com/panther-labs/panther/blob/cc049e9f82b3611353333d1adf9c2c8172554e30/tools/mage/build_namespace.go#L154-L180

It's a bit more verbose, but also more clean because it creates exactly as many goroutines as there are workers. Here, you create as many goroutines as _tasks_, which means many goroutines are blocked, doing nothing. Goroutines are cheap, but not free

Can we follow that same pattern in `test:ci` for consistency instead of the "ugly task queue hack"? I'm open to ideas for more generic worker pools to better share the code chunk I pasted, but let's do it the same way in all places and let's not have unnecessarily blocked goroutines

tools/mage/util.go
```diff
@@ -44,7 +44,33 @@ const (
...
+	}
+	// Ensure we don't set maxWorkers to zero
+	if n > 1 {
+		return n - 1
+	}
+	return 1
+}()
+
+// Queue limiting concurrent tasks when using `runTask`
+var queue = make(chan struct{}, maxWorkers)
```
austinbyers, line 61: If we were to keep this, the variable name needs to be less generic because it's shared across the entire `mage` namespace

tools/mage/util.go
```diff
@@ -44,7 +44,33 @@ const (
 )
 
 // For CPU-intensive operations, limit the max number of worker goroutines.
-var maxWorkers = runtime.NumCPU() - 1
+var maxWorkers = func() int {
+	n := runtime.NumCPU()
+	// Use all CPUs on CI environment
+	if runningInCI() {
```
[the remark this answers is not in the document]
austinbyers: Clever way to initialize a variable without needing an `init` function!

austinbyers:
Nice to have someone else looking at the `mage` code! Thanks @alxarch , I like the idea of limiting `test:ci`

tools/mage/util.go
alxarch: There are packages that offer this kind of functionality with a consistent interface. One I had reviewed for a project in the past was [tunny](https://github.com/Jeffail/tunny). It creates a (resizable) pool of goroutine ready to accept tasks and takes a `func() interface{}` as a task.
I quickly went over `mage` code and saw that it provides some form of parallelism in the form of `mg.Deps` and `mg.SerialDep`. That mechanism though provides no way to limit concurrent tasks, it's either serial or full parallel (limited by `GOMAXPROCS`. It's also missing the 'progress' layer (ie 'foo finished i/n')  that `logResults` provides on top of each task's logging output. 
I think the real issues come from the fact that most of the tasks we're running are spawning separate processes of external tools and these tools assume that they 'own' all processors so while we try to limit the tasks to let's say 4 cores each of these might spawn 4 threads/goroutine processors and that might cascade further. TBH I don't think it's worth the effort to guard against this, the tasks will finish eventually and all will be fine.

tools/mage/util.go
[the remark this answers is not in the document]
alxarch: Also note that it guards against the case of single core returning `maxworkers=0`

tools/mage/util.go
alxarch: Fixed

tools/mage/util.go
alxarch: I changed the `runTask` helper to spawn the goroutine *after* it pushes to the queue. That could not be done using the `count++` way so  I also used a slice of anonymous tasks to be more declarative about adding new tasks and be able to know the number of tasks to be run *before* we actually run them.

tools/mage/test_namespace.go
```diff
@@ -51,85 +51,39 @@ var (
...
-	count := 0
-
+	tasks := []struct {
+		Name string
+		Task func() error
+	}{
+		{"fmt", func() error { return fmtErr }},
+		{"build:cfn", build.cfn},
+		{"build:lambda", build.lambda},
+		{"build:tools", build.tools},
```
austinbyers, line 65: To avoid the process explosion problem you were talking about, let's run `build.lambda` and `testGoUnit` sequentially (and possibly `build.tools` as well) - each of those tasks already independently make full use of all available cores and are likely the cause of issues. Then the rest of the tasks can probably run in parallel without problems

tools/mage/util.go
austinbyers: I agree we don't need an external library for this, that would be way overkill and introduces an unnecessary extra dependency.

My concern was more that this introduces a new parallelism paradigm when other `mage` commands already have defined one. Why should we use this "ugly task queue hack" when we already have a worker pool model implemented elsewhere?

The answer, from what I can tell, is that this can be more easily re-used and requires less boilerplate to run a set of tasks in parallel. With the changes you made, I do like being able to read the slice of tasks, that's neat

kostaspap:
Seems that this PR has already gotten 3 approvals, merging

PR #805, opened 2020-04-29
Several minor improvements

kostaspap:
## Background

Several style improvements

## Changes

- Several style improvements 

## Testing

- `mage test:ci`
- `mage deploy`

### DIFF
--- a/api/lambda/source/models/api.go
+++ b/api/lambda/source/models/api.go
@@ -43,7 +43,7 @@ type LambdaInput struct {
 
 // CheckIntegrationInput is used to check the health of a potential configuration.
 type CheckIntegrationInput struct {
-	AWSAccountID     *string `genericapi:"redact" json:"awsAccountId" validate:"required,len=12,numeric"`
+	AWSAccountID     *string `genericapi:"redact" json:"awsAccountId" validate:"omitempty,len=12,numeric"`
 	IntegrationType  *string `json:"integrationType" validate:"required,oneof=aws-scan aws-s3"`
 	IntegrationLabel *string `json:"integrationLabel" validate:"required,integrationLabel"`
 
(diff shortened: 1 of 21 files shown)


internal/core/source_api/api/check_integration.go
```diff
@@ -49,37 +49,47 @@ var (
...
+}
 
-	return out, nil
+func checkAwsScanIntegration(input *models.CheckIntegrationInput) *models.SourceIntegrationHealth {
+	out := &models.SourceIntegrationHealth{
+		AWSAccountID:    aws.StringValue(input.AWSAccountID),
+		IntegrationType: aws.StringValue(input.IntegrationType),
+	}
+	_, out.AuditRoleStatus = getCredentialsWithStatus(fmt.Sprintf(auditRoleFormat,
+		*input.AWSAccountID, *awsSession.Config.Region))
```
nhakmiller, line 68: `input.AWSAccontID` could be nil now right? Won't this panic?

internal/core/source_api/api/check_integration.go
```diff
@@ -49,37 +49,47 @@ var (
...
+	return out
+}
+
+func checkAwsS3Integration(input *models.CheckIntegrationInput) *models.SourceIntegrationHealth {
+	out := &models.SourceIntegrationHealth{
+		AWSAccountID:    aws.StringValue(input.AWSAccountID),
+		IntegrationType: aws.StringValue(input.IntegrationType),
+	}
+	var roleCreds *credentials.Credentials
+	logProcessingRole := generateLogProcessingRoleArn(*input.AWSAccountID, *input.IntegrationLabel)
```
nhakmiller, line 86: Same question as above about a nil account ID

api/lambda/source/models/api.go
```diff
@@ -69,7 +69,7 @@ type PutIntegrationInput struct {
 
 // PutIntegrationSettings are all the settings for the new integration.
 type PutIntegrationSettings struct {
-	AWSAccountID       *string   `genericapi:"redact" json:"awsAccountId" validate:"required,len=12,numeric"`
+	AWSAccountID       *string   `genericapi:"redact" json:"awsAccountId" validate:"omitempty,len=12,numeric"`
```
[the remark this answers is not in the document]
kostaspap: There are some integrations that don't require this... 
It is similar to the scenario that we have where the S3Bucket is marked `omitempty`, however it should never be empty in an integration of type `aws-s3`. In both cases we expect that frontend sends the appropriate request.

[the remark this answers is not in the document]
kostaspap: I reverted back this change as it was not required. Was afraid of scope creep if I were to implement input validation for the API inputs... as it was not the purpose of my PR. I am planning to introduce some changes to the API anyway to address this specific issue but on a separate PR I am working on.

internal/core/source_api/api/check_integration.go
kostaspap: Added back the "required" check for the AWS Account ID

internal/core/source_api/api/check_integration.go
kostaspap: Added back the check in the API.

PR #808, opened 2020-04-29
Clarify deployment docs

nhakmiller:
## Background

The current deploy docs do not specify that `mage setup` needs to be run before `mage test:ci`, even though they suggest users may run `mage test:ci` before deploying. This small update just clarifies that `mage setup` is a requirement for `mage test:ci`.

## Changes

- Clarified some docs

## Testing

- Viewed in markdown renderer of IDE

### DIFF
--- a/docs/gitbook/quick-start.md
+++ b/docs/gitbook/quick-start.md
@@ -65,7 +65,7 @@ Run the following command to deploy Panther:
 mage setup deploy
 ```
 
-- Optionally, you can run `mage setup test:ci` before deploying to confirm that all tests are passing. If you perform this step, then you can just deploy with `mage deploy` as setup has already been completed.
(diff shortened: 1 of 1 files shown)


docs/gitbook/quick-start.md
```diff
@@ -65,7 +65,7 @@ Run the following command to deploy Panther:
 mage setup deploy
 ```
 
-- Optionally, you can `mage test:ci` before deploying to confirm that all tests are passing
+- Optionally, you can run `mage setup test:ci` before deploying to confirm that all tests are passing. If you perform this step, then you can deploy with just `mage deploy` as setup has already been completed.
```
austinbyers, line 68: Maybe change the last sentence to `(Setup only needs to be run once in your repo)`

Or we can just omit it completely, `setup` is safe to run multiple times and won't redownload stuff that already exists

PR #810, opened 2020-04-30
Reject HTML characters in most user-facing APIs

austinbyers:
## Background

#798 shows the danger of storing user input that may contain HTML. While we fixed that specific issue by escaping before sending the email, Panther may send more emails in the future that could embed user names, policy names, organization display name, etc. Alert outputs also may be sent to email or other services that may or not render the resulting text

To be as safe as possible, we should completely reject HTML characters on all user-facing APIs where it makes sense to do so. If we don't store them in the first place, we won't risk sending them unsafely. We might not want to reject HTML everywhere - for example, the "reference" in a policy may want a real link (with `&` characters)

I imagine (and encourage) this PR will engender considerable discussion.

## Changes

Reject the character set `< > & " '` from the following:

* User first/last name
* Organization display name
* Policy/Rule display name
* Source label
* Output display name

That character set was chosen because it is exactly what `html.EscapeString` uses. We don't want to call `html.EscapeString` on these fields because it is not idempotent - calling it multiple times on the same string will keep changing it. We also don't want to be silently modifying user input, it's better to just reject the request

## Testing

@3nvi @alexmylonas I noticed I can still add HTML characters to the user first/last name when I go to edit my own profile from the profile icon, even though the `users-api` rejects `UpdateUser` requests with HTML characters (I tested). Is this update going to Cognito directly? If so, we need to re-route it through the `users-api` like all other user changes

Otherwise, the errors look good:

Screen Shot 2020-04-29 at 5 25 42 PM

Screen Shot 2020-04-29 at 5 26 23 PM

Policies show the error in a strange way, but an error nonetheless:

Screen Shot 2020-04-29 at 5 53 18 PM

### DIFF
--- a/api/lambda/organization/models/api.go
+++ b/api/lambda/organization/models/api.go
@@ -34,7 +34,7 @@ type UpdateSettingsInput = GeneralSettings
 
 // GeneralSettings defines basic settings for a Panther deployment.
 type GeneralSettings struct {
-	DisplayName           *string `json:"displayName" validate:"omitempty,min=1"`
+	DisplayName           *string `json:"displayName" validate:"omitempty,min=1,excludesall='<>&\""`
 	Email                 *string `genericapi:"redact" json:"email" validate:"omitempty,email"`
 	ErrorReportingConsent *bool   `json:"errorReportingConsent"`
 }
--- a/api/lambda/outputs/models/api.go
+++ b/api/lambda/outputs/models/api.go
@@ -43,7 +43,7 @@ type LambdaInput struct {
 // }
 type AddOutputInput struct {
 	UserID             *string       `json:"userId" validate:"required,uuid4"`
-	DisplayName        *string       `json:"displayName" validate:"required,min=1"`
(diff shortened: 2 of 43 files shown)


austinbyers:
> My slight concern is about verbosity, I am thinking it might be easier to use if we had a validator for this?

Validator tags are only applicable to Lambda-style APIs (`genericapi`), it doesn't apply to API gateway (analysis-api) because we don't directly control those Go structs. If and when we can consolidate to a consistent API format, I agree a validator would be the way to go

austinbyers:
I updated all the lambda-based APIs to just add an `excludesall` validation tag. The gateway-based APIs still have to explicitly check.

The challenge with struct validation is that the generated errors are completely unhelpful for the user:

Screen Shot 2020-04-30 at 1 23 42 PM

So I also made a change to the core router to restructure that error message:

Screen Shot 2020-04-30 at 3 36 54 PM

As an added bonus, this will improve *all* validation error messages from lambda-based APIs. It's still not perfect, we can spend more time in the future generating custom error messages for every type of validation failure. But this is definitely a step in the right direction, better than we had before

PR #813, opened 2020-04-30
updated alerts folder name

willow9886:
changed the docs folder name from "setup" to "destinations" .

## Background

Currently, the top level folder for destinations is called setup, which is unintuitive and ambiguous. 

## Changes

I deleted the setup folder, and moved all docs under /destinations. 

## Testing


kostaspap:
Are we still going ahead with this change? @willow9886  ?

3nvi:
> Are we still going ahead with this change? @willow9886 ?

Was just gonna ask the same

PR #816, opened 2020-04-30
Refactor datacatalog-updater and pkg/awsglue for Parquet work (minimize conflicts)

rleighton:
## Background
We will be supporting automatic JSON to Parquet conversion for processed log data in `panther-enterprise`. 

In order to minimize merge conflicts, the  supporting code has been refactored.

There is only minor change to the functionality in this PR.

## Changes
- Conform to structure in `panther-enterprise`
- Optimize for easy merges (e.g., isolate code unchanged in `panther-enterprise` into files)
- CreatePartition() now returns a boolean to indicate if a partition was created or already existed
- Added GetPartition() method to look up a Glue partition
- Added new "panther_temp" Athena database

## Testing
mage test:ci
mage deploy
testing in `panther-enterprise`


internal/log_analysis/datacatalog_updater/process/glue_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (m *mockGlue) GetPartition(input *glue.GetPartitionInput) (*glue.GetPartitionOutput, error) {
+	args := m.Called(input)
+	return args.Get(0).(*glue.GetPartitionOutput), args.Error(1)
+}
+
+func (m *mockGlue) UpdatePartition(input *glue.UpdatePartitionInput) (*glue.UpdatePartitionOutput, error) {
+	args := m.Called(input)
+	return args.Get(0).(*glue.UpdatePartitionOutput), args.Error(1)
+}
```
kostaspap, line 109: Move this to `pkg/testutils/awsmocks.go`.

internal/log_analysis/datacatalog_updater/process/init_test.go
```diff
@@ -0,0 +1,25 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+func initTest() {
+	partitionPrefixCache = make(map[string]struct{})
+	mockGlueClient = &mockGlue{}
```
kostaspap, line 23: The mock is a stateful object, so we need to a new one in each test.

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+
+			// remember in cache
+			partitionPrefixCache[partitionLocation] = struct{}{}
+		}
+	}
+	return nil
+}
+
+func getPartition(bucketName, key string) (existsInCache bool, partitionLocation string, gluePartition *awsglue.GluePartition, err error) {
```
kostaspap, line 75: You don't have to return the `partitionLocation` since this is already available in the `gluePartition` object.

pkg/awsglue/glue.go
```diff
@@ -347,16 +382,16 @@ func (gm *GlueTableMetadata) partitionHasData(client s3iface.S3API, t time.Time,
...
 	}
-	if gm.timebin >= GlueTableDaily {
+	if timebin >= GlueTableDaily {
 		values = append(values, aws.String(fmt.Sprintf("%02d", t.Day())))
 	}
-	if gm.timebin >= GlueTableHourly {
+	if timebin >= GlueTableHourly {
 		values = append(values, aws.String(fmt.Sprintf("%02d", t.Hour())))
 	}
 	return
```
kostaspap, line 397: This looks like it could be a method of the `GlueTableTimeBin`struct, it is not dependent on any of the fields for the GlueTableMetadata.

pkg/awsglue/partition.go
```diff
@@ -54,6 +55,10 @@ func (gp *GluePartition) GetTable() string {
 	return gp.tableName
 }
 
+func (gp *GluePartition) GetHour() time.Time {
+	return gp.hour
+}
```
kostaspap, line 60: The partition could be yearly, monthly, daily (and hourly). How does "GetHour" makes sense for such partitions?

pkg/awsglue/partition.go
```diff
@@ -104,10 +109,15 @@ type PartitionColumnInfo struct {
...
 // Creates a new partition in Glue using the client provided.
-func (gp *GluePartition) CreatePartition(client glueiface.GlueAPI) error {
+func (gp *GluePartition) CreatePartition(client glueiface.GlueAPI) (created bool, err error) {
 	return NewGlueTableMetadata(gp.datatype, gp.tableName, "", GlueTableHourly, nil).CreateJSONPartition(client, gp.hour)
 }
 
+// Gets the Glue partition using the client provided.
+func (gp *GluePartition) GetPartition(client glueiface.GlueAPI) (output *glue.GetPartitionOutput, err error) {
+	return NewGlueTableMetadata(gp.datatype, gp.tableName, "", GlueTableHourly, nil).GetPartition(client, gp.hour)
+}
```
kostaspap, line 119: This method seems that it is not needed. You just delegate the call to the equivalent `GlueTableMetadata` method which is also public. 
The caller can just call the actual method on the GlueTableMetadata. If anything you want callers to be able to go from GluePartition to the GlueTableMetadata, you can just add a pointer to it (e.g. `func (gp *GluePartition) GetTableMetadata() *glue.GlueTableMetadata`).

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/panther-labs/panther/pkg/awsglue"
+)
+
+var (
+	// partitionPrefixCache is a cache that stores all the prefixes of the partitions we have created
+	// The cache is used to avoid attempts to create the same partitions in Glue table
+	partitionPrefixCache = make(map[string]struct{})
+)
+
+func SQS(event events.SQSEvent) error {
```
kostaspap, line 37: This method is not dependent on the input being an SQS message in any case. It is an opportunity to create an abstraction layer here to hide the "client" used to fetch the data. 

You can just change this to `func Handle(events []string` and convert the input in the actual Lambda initialization code.

pkg/awsglue/glue.go
```diff
@@ -165,6 +168,17 @@ func GetTableName(logType string) string {
...
 }
 
+func GetDataPrefix(databaseName string) string {
+	switch databaseName {
+	case LogProcessingDatabaseName:
+		return logS3Prefix
+	case RuleMatchDatabaseName:
+		return ruleMatchS3Prefix
+	default:
+		return logS3Prefix // assume logs, used for integration tests
```
kostaspap, line 178: This comment is not accurate, we also have `ViewsDatabaseName` and `TempDatabaseName`. 
What is the prefix used by the tests, we could add it explicitly. I'd prefer to have a `default` that panics, this code could be used to read or even worse write data and could lead to data overwriting.

pkg/awsglue/glue.go
```diff
@@ -303,18 +320,36 @@ func (gm *GlueTableMetadata) createPartition(client glueiface.GlueAPI, t time.Ti
...
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == glue.ErrCodeAlreadyExistsException {
+			return false, nil // no error
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+// get partition, return nil if it does not exist
+func (gm *GlueTableMetadata) GetPartition(client glueiface.GlueAPI, t time.Time) (output *glue.GetPartitionOutput, err error) {
```
kostaspap, line 332: This method isn't used anywhere in the code...? Also it is a bit confusing, this method returns actual "glue" class while the whole package deals with a different level of abstraction.

internal/log_analysis/datacatalog_updater/process/glue_test.go
rleighton: Good suggestion, will minimize conflicts. Done.

pkg/awsglue/glue.go
rleighton: Indeed. Move to a method of GlueTableTimebin.

internal/log_analysis/datacatalog_updater/process/init_test.go
rleighton: I put this in a separate file because initTest() will be used for both process_test.go and compaction_test.go in `panther-enterprise`.  So it is shared. I could have initProcessTest() in process_test.go and initCompactionTest() in compaction_test.go. That would have 0 conflicts but they are 99% the same, the only difference is 1 line to get another mock. Thoughts?

internal/log_analysis/datacatalog_updater/process/process.go
rleighton: It is not actually, it is computed with string manipulation. Since this string is used immediately after this call it is more efficient to return it.

pkg/awsglue/partition.go
rleighton: This is a "getter" for gp.hour. The other members have getters (see surrounding code), this one was just forgotten because it was unused. It is used in `panther-enterprise`

pkg/awsglue/partition.go
rleighton: That's a good suggestion. When a GluePartition is created it makes a *glue.GlueTableMetadata. GluePartition needs one interface to get to the gluecatlog to create/get partitions. Done.

internal/log_analysis/datacatalog_updater/process/process.go
kostaspap: It is not such an expensive call though, it's just string manipulation. It adds more verbosity and it is kinda confusing for the caller - the caller function has access to a "partitionLocation" string and to a `gluePartition.GetPartitionLocation` method - but objects returned from the same function.

internal/log_analysis/datacatalog_updater/process/process.go
rleighton: I would prefer not to do this until there is a need. I know we did this for the log processor, but we waiting until we had other events. For the data catalog updater I'm not sure we will, since the definition of its work is to take S3 notifications and reflect them in the glue catalog.

pkg/awsglue/partition.go
kostaspap: What does it represent though? Should it be "GetTime"?

pkg/awsglue/glue.go
rleighton: Views and temp databases do not have s3 prefixes, but I see your point. I changed this to add a special case for integration tests, else panic.

pkg/awsglue/glue.go
rleighton: In `panther-enterprise` I need a way to go from an s3 path to a glue partition. This is the "bridge". The change you suggested above where the glue metadata table is a member of the GluePartition I think makes this more natural.

internal/log_analysis/datacatalog_updater/process/init_test.go
kostaspap: I am just nitpicking here... personally I would prefer to have two different methods. The test for each "unit" (`process`, `compaction`) has its own requirements... and it happens that for now they are the same. For all we know, each unit might require different initialization in the future.

internal/log_analysis/datacatalog_updater/process/process.go
rleighton: I would push back if this is in a "hot" path, but it is not. We only need this when we create partitions which happens (mostly) a few times an hour per lambda per table. I'll change it.
Having written a lot of low level code where this kind of waste is an issue, it is not in my initial instinct to do things like this. Death by a thousand cuts. This will make the code a little prettier. I do admire your aesthetic  :)

pkg/awsglue/partition.go
```diff
@@ -104,8 +110,13 @@ type PartitionColumnInfo struct {
...
-func (gp *GluePartition) CreatePartition(client glueiface.GlueAPI) error {
-	return NewGlueTableMetadata(gp.datatype, gp.tableName, "", GlueTableHourly, nil).CreateJSONPartition(client, gp.hour)
+func (gp *GluePartition) CreatePartition(client glueiface.GlueAPI) (created bool, err error) {
+	return gp.gm.CreateJSONPartition(client, gp.hour)
+}
+
+// Gets the Glue partition using the client provided.
+func (gp *GluePartition) GetPartition(client glueiface.GlueAPI) (output *glue.GetPartitionOutput, err error) {
+	return gp.gm.GetPartition(client, gp.hour)
 }
```
kostaspap, line 120: The caller, if it has already the GluePartition, can do `partition.GetTableMetadata().CreatePartition(...)` and equivalent.  Adding these methods means our API now has two more methods that the caller needs to be aware of and we have two more methods we have to test.

internal/log_analysis/datacatalog_updater/process/init_test.go
kostaspap: Not a big deal though, you can do as you prefer.

internal/log_analysis/datacatalog_updater/process/process.go
kostaspap: I'm not really a fan of this, but I see your perspective. I'm fine to leave as is.

pkg/awsglue/partition.go
rleighton: Sure. It is the time of the partition (truncated appropriately). I named it Hour() because the member is `hour`.  I'll change the member and getter.

pkg/awsglue/glue.go
kostaspap: We already have the `GetPartitionFromS3`, isn't it sufficient?

internal/log_analysis/datacatalog_updater/process/init_test.go
rleighton: True, but while we are in here might as well. Usually we try not to duplicate code but to keep these merges clean, splitting test code is a good idea.

pkg/awsglue/partition.go
rleighton: Done

pkg/awsglue/glue.go
rleighton: Nope, it does not read from the Glue catalog. The glue catalog as info not stored in the s3 path (like the schema and the serde), which I need.

pkg/awsglue/partition.go
rleighton: True. I can do that.

rleighton: Alright, did that. I have to say this was one of your more pedantic requests. It does make the interface clear between GluePartition and GlueTableMetadata, but at the expense of slightly more verbose call signature. It did not help any with testing, since those code paths all have tests anyway.  But it is done.

pkg/awsglue/glue.go
kostaspap: Hmmm, I see.... This is actually information we have already and we wouldn't have to query Glue for that: 
1. If an S3 object is .json.gz, we know that it is of JSON Serde information
2. If an S3 object is .parquet, we know it belongs to a Parquet Serde
3. From the table name we can infer the LogType and thus fetch the schema from the registry. 

We can be 100% sure that this info is accurate because we generate them from our code. If some table are managed outside our code, well ... we have bigger problems :P 

Anyway, I see that this would require some bigger change, so I don't suggest doing it now. It seems that we could fix that eventually.

pkg/awsglue/glue.go
```diff
@@ -262,29 +279,31 @@ func (gm *GlueTableMetadata) SyncPartitions(glueClient glueiface.GlueAPI, s3Clie
...
 	tableOutput, err := client.GetTable(tableInput)
 	if err != nil {
-		return err
+		return false, err
 	}
 
 	// ensure this is a JSON table, use Contains() because there are multiple json serdes
 	if !strings.Contains(*tableOutput.Table.StorageDescriptor.SerdeInfo.SerializationLibrary, "json") {
-		return errors.Errorf("not a JSON table: %#v", *tableOutput.Table.StorageDescriptor)
+		return false, errors.Errorf("not a JSON table: %#v", *tableOutput.Table.StorageDescriptor)
```
kostaspap, line 295: It's not for the current PR, but I assume this will not work with the parquet work. There is no such thing as "table serde", it's more about partition Serde...

pkg/awsglue/glue.go
rleighton: I thought about that but decided that was risky since we could have "skew" related bugs w.r.t. things working with the Glue catalog. For example on a deploy there could be races between schema changes and dependent Athena processes (which only sees the Glue catalog). So for this purpose I really need the catalog.

pkg/awsglue/glue.go
rleighton: Not relevant to the Parquet work. Parquet code (currently) never creates JSON partitions.

PR #820, opened 2020-05-03
Add PantherEdition tag so it easy to tell apart Enterprise vs OpenSource

rleighton:
## Background
When having more than one deployment, it easy to forget which is Enterprise and which is OpenSource.

This can also be useful for troubleshooting customer issues to definitively know the edition deployed.

## Changes
* Add a "PantherEdition" tag.

## Testing
* mage test:ci
* mage deploy

### DIFF
--- a/tools/mage/deploy_template.go
+++ b/tools/mage/deploy_template.go
@@ -304,6 +304,7 @@ func createChangeSet(
 		StackName:     &stack,
 		Tags: []*cfn.Tag{ // Tags are propagated to every supported resource in the stack
 			{Key: aws.String("Application"), Value: aws.String("Panther")},
+			{Key: aws.String("PantherEdition"), Value: aws.String("OpenSource")},
 			{Key: aws.String("PantherVersion"), Value: &pantherVersion},
 			{Key: aws.String("Stack"), Value: &stack},
 		},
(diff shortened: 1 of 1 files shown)


tools/mage/deploy_template.go
```diff
@@ -304,6 +304,7 @@ func createChangeSet(
 		StackName:     &stack,
 		Tags: []*cfn.Tag{ // Tags are propagated to every supported resource in the stack
 			{Key: aws.String("Application"), Value: aws.String("Panther")},
+			{Key: aws.String("PantherEdition"), Value: aws.String("OpenSource")},
```
austinbyers, line 307: This is already shown in the general settings page and on the login pages. It's a constant in the web src folder

Do see some additional value pinning it to all of the backend infra?

rleighton: Absolutely, a few reasons:
* when troubleshooting, you are in the infra and not likely using the UI and being really sure what is out there quickly is important (same reason we added version)
* customers will likely have sysadmins doing installation or troubleshooting, they may not have access to the UI
* as a backed dev, primarily, I rarely use the UI and often deploy w/out configuring access to the UI. I have regions with OSS and Enterprise and it good to have a definitive way to know exactly what is deployed.

austinbyers: Let's change its name to `Community` 
(https://runpanther.io/pricing/)

rleighton: Done

PR #825, opened 2020-05-05
Updated EditProfile to use 'users-api' instead of Cognito

### LINKED ISSUE #817: Route user updates to users-api instead of Cognito directly
### Description

When you edit your first/last name by clicking on your profile icon on the upper right, that request is sent to Cognito directly.

It needs to go through the users-api instead so we can enforce important validation (see https://github.com/panther-labs/panther/pull/810). We should only be communicating with Cognito directy during the login flow

There is already an `UpdateUser` endpoint in the backend

### Acceptance Criteria

- If I change my own first name by clicking on my profile, I can't embed HTML (which is enforced by `users-api`)

alexmylonas:
## Background

For security reasons, we need to block updates on user's first and last name that include HTML fragments. This check is supported from `users-api` but from cognito which we are currently usign for user updates.

This PR addresses this issue by update the `EditProfileForm` to use `users-api`

Closes #817 

## Changes

- Updated `EditProfileForm` to use `useEditUser` instead of cognito's `updateUserInfo`

## Testing

- Locally

### DIFF
--- a/web/src/components/forms/EditProfileForm/EditProfileForm.tsx
+++ b/web/src/components/forms/EditProfileForm/EditProfileForm.tsx
@@ -22,8 +22,8 @@ import { Field, Formik } from 'formik';
 import FormikTextInput from 'Components/fields/TextInput';
 import SubmitButton from 'Components/buttons/SubmitButton';
 import useAuth from 'Hooks/useAuth';
-import { useState } from 'react';
 import { useEditUser } from 'Components/sidesheets/EditUserSidesheet/graphql/editUser.generated';
+import { extractErrorMessage } from 'Helpers/utils';
 
 interface EditProfileFormProps {
   onSuccess: () => void;
@@ -37,17 +37,17 @@ interface EditProfileFormValues {
 
 const EditProfileForm: React.FC<EditProfileFormProps> = ({ onSuccess }) => {
   const { userInfo, refetchUserInfo } = useAuth();
-  const [status, setStatus] = useState(null);
+  const [status, setStatus] = React.useState(null);
 
   const [editUser] = useEditUser({
     onCompleted: () => {
       onSuccess();
       return refetchUserInfo();
     },
-    onError: ({ message }) =>
+    onError: updateUserError =>
       setStatus({
         title: 'Unable to update profile',
-        message,
+        message: extractErrorMessage(updateUserError) || 'An unknown error occurred',
       }),
   });
 
--- a/web/src/components/forms/EditProfileForm/EditProfileForm.tsx
+++ b/web/src/components/forms/EditProfileForm/EditProfileForm.tsx
@@ -17,11 +17,13 @@
  */
 
 import * as React from 'react';
-import { Alert, Box, Flex, useSnackbar } from 'pouncejs';
+import { Alert, Box, Flex } from 'pouncejs';
 import { Field, Formik } from 'formik';
 import FormikTextInput from 'Components/fields/TextInput';
 import SubmitButton from 'Components/buttons/SubmitButton';
 import useAuth from 'Hooks/useAuth';
+import { useState } from 'react';
+import { useEditUser } from 'Components/sidesheets/EditUserSidesheet/graphql/editUser.generated';
 
 interface EditProfileFormProps {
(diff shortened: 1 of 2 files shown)


web/src/components/forms/EditProfileForm/EditProfileForm.tsx
```diff
@@ -34,8 +36,20 @@ interface EditProfileFormValues {
...
 const EditProfileForm: React.FC<EditProfileFormProps> = ({ onSuccess }) => {
-  const { userInfo, updateUserInfo } = useAuth();
-  const { pushSnackbar } = useSnackbar();
+  const { userInfo, refetchUserInfo } = useAuth();
+  const [status, setStatus] = useState(null);
+
+  const [editUser] = useEditUser({
+    onCompleted: () => {
+      onSuccess();
+      return refetchUserInfo();
```
3nvi, line 45: no need for `return`

3nvi, line 45: we are missing 

```tsx
            pushSnackbar({ title: 'Successfully updated profile!', variant: 'success' });
```

that the existing code had

web/src/components/forms/EditProfileForm/EditProfileForm.tsx
```diff
@@ -17,11 +17,13 @@
  */
 
 import * as React from 'react';
-import { Alert, Box, Flex, useSnackbar } from 'pouncejs';
+import { Alert, Box, Flex } from 'pouncejs';
 import { Field, Formik } from 'formik';
 import FormikTextInput from 'Components/fields/TextInput';
 import SubmitButton from 'Components/buttons/SubmitButton';
 import useAuth from 'Hooks/useAuth';
+import { useState } from 'react';
```
3nvi, line 25: no need to import react twice. Just use `React.useState` instead of doing this import

web/src/components/forms/EditProfileForm/EditProfileForm.tsx
```diff
@@ -34,8 +36,20 @@ interface EditProfileFormValues {
...
-  const { pushSnackbar } = useSnackbar();
+  const { userInfo, refetchUserInfo } = useAuth();
+  const [status, setStatus] = useState(null);
+
+  const [editUser] = useEditUser({
+    onCompleted: () => {
+      onSuccess();
+      return refetchUserInfo();
+    },
+    onError: ({ message }) =>
```
3nvi, line 47: I'm not 100% sure that `message` is returned from the `onError`. The signature is this

```ts
    onError?: (error: ApolloError) => void;
```

We should treat it the same way we do in other components

web/src/components/forms/EditProfileForm/EditProfileForm.tsx
alexmylonas: This is just to suppress a minor warning about `Promise returned ... is ignored`
Should I remove it?

web/src/components/forms/EditProfileForm/EditProfileForm.tsx
alexmylonas: Correct!

web/src/components/forms/EditProfileForm/EditProfileForm.tsx
3nvi: oh gotcha. Whatever you feel then

alexmylonas: Indeed, but it was never displayed since this code is live and dies within the sidesheet, which `onSuccess` we stop rendering it.

web/src/components/forms/EditProfileForm/EditProfileForm.tsx
alexmylonas: Seems to be working ok but I updated to use `extractErrorMessage` as we do in several other places

web/src/components/forms/EditProfileForm/EditProfileForm.tsx
3nvi: I think now that you added it to Apollo, it will properly be executed

PR #827, opened 2020-05-05
Grant rules engine permission to query resources table

nhakmiller:
## Background

Now that `globals` are supported and shared across policies and rules, there is no reason not to let the `rules-engine` query the `resources-table` in the same manner that the `policy-engine` is able to. This change grants the `rules-engine` these permissions.

## Changes

- Granted the `rules-engine` permission to communicate with the `resources-table`

## Testing

- `mage test:ci`
- Deployed to my dev account and tested manually


nhakmiller:
> I suppose this could be abused but we should get lots of alarms if that happens, so there should be no silent failures.

We originally blocked this due to concerns about cost due to the high volume of rules/events being processed and the (relatively) high latency of dynamodb reads. But some users have requested this feature, and really there's nothing stopping somebody from creating long running rules if that is their goal. I'll merge for now, feel free to ping me if you think there are some compensating controls we can put in to address the abuse you mention.

deployments/log_analysis.yml
```diff
@@ -661,3 +661,10 @@ Resources:
             - Effect: Allow
               Action: execute-api:Invoke
               Resource: !Sub arn:${AWS::Partition}:execute-api:${AWS::Region}:${AWS::AccountId}:${AnalysisApiId}/v1/GET/enabled
+        - Id: ResourceLookup
+          Version: 2012-10-17
+          Statement:
+            - Effect: Allow
+              Action:
+                - dynamodb:GetItem
+              Resource: !Sub arn:${AWS::Partition}:dynamodb:${AWS::Region}:${AWS::AccountId}:table/panther-resources
```
kostaspap, line 670: This is really interesting! So we can already write rules that interact with resources :D

PR #839, opened 2020-05-06
Aggelos add missing alerts

### LINKED ISSUE #794: Rule Details Page doesn't list all related alerts
### Describe the bug

In a details page of a rule, only  the first 10 alerts are listed and the rest (that belong to separate pages are never loaded 

### Steps to reproduce

1. Go to a rule's page with more than 25 alerts
2. See how many you are viewing
3. Observe that there is no way too load more

### Expected behavior

All alerts that are related to a rule should be listed

### Environment

- Panther version or commit: [e.g. v0.1.1]
- OS: [e.g. MacOS 10.15.3]
- Browser: [e.g. Chrome 80.0.3987.87, Safari Version 13.0.5 (15608.5.11)]

### Additional context

Only the 1st page is listed and the froont-end doesn't handle the loading of more pages

3nvi:
## Background

Currently, when you are viewing a rule, you only see the last 10 alerts and not the full alert list. This was an old bug that was just noticed by me, so this PR addresses it and makes sure that the more you scroll the more alerts you will see.

Closes #794 

## Changes

- Add endless pagination to alerts in Rule Details page

## Testing

- Added small `pageSize` and made sure that all pages are fetched as you scroll


kostaspap:
> ## Background
> Currently, when you are viewing a rule, you only see the last 10 alerts and not the full alert list. 

I have deployed version `v1.2.0-2-g3a279b91` in my account (master branch). I don't see the issue you mention here - I am able to see a list of 316 alerts associated with a rule from the rule details page.

3nvi:
> > ## Background
> > Currently, when you are viewing a rule, you only see the last 10 alerts and not the full alert list.
> 
> I have deployed version `v1.2.0-2-g3a279b91` in my account (master branch). I don't see the issue you mention here - I am able to see a list of 316 alerts associated with a rule from the Rule Details page.


@kostaspap  You are right. The behavior (up until now) was to "fetch all alerts in one go" just as a temporary hack until the current PR landed. Out of curiosity, can you see how much time the request took (and how much KB got transferred)? :P

kostaspap:
> > > ## Background
> > > Currently, when you are viewing a rule, you only see the last 10 alerts and not the full alert list.
> > 
> > 
> > I have deployed version `v1.2.0-2-g3a279b91` in my account (master branch). I don't see the issue you mention here - I am able to see a list of 316 alerts associated with a rule from the Rule Details page.
> 
> @kostaspap You are right. The behavior (up until now) was to "fetch all alerts in one go" just as a temporary hack until the current PR landed. Out of curiosity, can you see how much time the request took (and how much KB got transferred)? :P

Ok so the fix is that we now fetch them gradually instead of all at once, right?
Not sure where I can see the amount of data, the query to GraphQL took 620ms from a warm lambda.

3nvi:
> Ok so the fix is that we now fetch them gradually instead of all at once, right?

That's right.

> Not sure where I can see the amount of data, the query to GraphQL took 620ms from a warm lambda.

If you open the `network tab` on Chrome's devtools and inspect the **XHR** tab, you can see the "Response" of a request and see the number of alerts returned

kostaspap:
> > Ok so the fix is that we now fetch them gradually instead of all at once, right?
> 
> That's right.
> 
> > Not sure where I can see the amount of data, the query to GraphQL took 620ms from a warm lambda.
> 
> If you open the `network tab` on Chrome's devtools and inspect the **XHR** tab, you can see the "Response" of a request and see the number of alerts returned

The number of alerts is 316 - I mentioned above. I thought you were looking to see how many bytes were transferred.

Issue #834, opened 2020-05-06
Panther tables not shipping with default views


nhakmiller:
### Describe the bug

On a fresh deployment of Panther, there are supposed to be a number of views available on the panther tables in Athena to assist with making generalized search (views such as `panther_views.all_logs` and `panther_views.all_rule_matches`). These views are no longer being created by default.

### Steps to reproduce

1. Do a fresh deployment of Panther
2. Go to the AWS Athena console
3. Scroll down to where the views are listed on the bottom left, observe that there are none

### Expected behavior

Panther should deploy with a handful of default views.

### Environment

v1.2.0 candidate commit.

### Additional context

@rleighton confirmed this in his accounts.

### Screenshots

<img width="368" alt="Screen Shot 2020-05-05 at 7 04 22 PM" src="https://user-images.githubusercontent.com/49166439/81132420-3ed45c80-8f03-11ea-84ff-6d4399f2dacb.png">

rleighton:
This is due to 2 issue: 
1) missing error check 
2) the issue discussed at standup with gcp schemas
The bad schema caused creating the view to fail and the missing error check caused it to go un-noticed.

rleighton:
Resoved

PR #841, opened 2020-05-06
Fix error handing in awsathena.WaitForResults

rleighton:
## Background
We discovered during release testing that athena views were not being deployed. Digging into the problem there were 2 issues:
 1) the new gcp logs have a type declaration that Athena does not like, causing view creation to fail
 2) an error path was not handled in the Athena query code, so deploys succeeded

See issue: https://github.com/panther-labs/panther/issues/834

This PR fixes the second issue.

There was a test for execution failures in this code, however, it was validating the wrong error path (failure in SQL planning). The proper test is to force a failure AFTER the planner. This was achieved by creating a table with a  non-existent bucket. The test was further updated to validate the error message from the check for execution failures to confirm the proper code path.

The Athena API does not use awsathena.WaitForResults() and is not affected by this issue.

## Changes
- Add code to check for execution failures
- Change execution failure test 

## Testing
mage test:ci
ran integration tests for awsathena
ran mage deploy, this now fails due to issue 1 above

### DIFF
--- a/pkg/awsathena/query.go
+++ b/pkg/awsathena/query.go
@@ -84,7 +84,7 @@ func WaitForResults(client athenaiface.AthenaAPI, queryExecutionID string) (quer
 				return nil, err
 			}
 			if done {
-				return executionOutput, err
+				return executionOutput, nil
 			}
 			time.Sleep(pollDelay)
 		}
--- a/pkg/awsathena/query.go
+++ b/pkg/awsathena/query.go
@@ -81,7 +81,11 @@ func WaitForResults(client athenaiface.AthenaAPI, queryExecutionID string) (quer
 				return nil, err
 			}
 			if done {
-				return executionOutput, nil
+				if *executionOutput.QueryExecution.Status.State == athena.QueryExecutionStateFailed {
+					err = errors.Errorf("query execution failed: %s",
+						*executionOutput.QueryExecution.Status.StateChangeReason)
+				}
+				return executionOutput, err
 			}
 			time.Sleep(pollDelay)
 		}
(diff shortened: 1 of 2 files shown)


pkg/awsathena/query.go
```diff
@@ -81,7 +81,11 @@ func WaitForResults(client athenaiface.AthenaAPI, queryExecutionID string) (quer
 				return nil, err
 			}
 			if done {
-				return executionOutput, nil
+				if *executionOutput.QueryExecution.Status.State == athena.QueryExecutionStateFailed {
+					err = errors.Errorf("query execution failed: %s",
+						*executionOutput.QueryExecution.Status.StateChangeReason)
+				}
+				return executionOutput, err
```
kostaspap, line 88: nit: The code in the switch statement of lines 66 is already handling the different `QueryExecution` states. 
It seems to me it would be more appropriate to change that switch statement so in case of `athena.QueryExecutionStateFailed` we return the error, sth like:

```
isFinished := func() (executionOutput *athena.GetQueryExecutionOutput, done bool, err error) {
  switch *executionOutput.QueryExecution.Status.State {
    case athena.QueryExecutionStateSucceeded, athena.QueryExecutionStateCancelled:
       return executionOutput, true, nil
    case athena.QueryExecutionStateFailed:
      return executionOutput, errors.Errorf("query execution failed: %s",....)
....
}
poll := func() (*athena.GetQueryExecutionOutput, error) {
 for {
   executionOutput, done, err := isFinished()
   if done {
       return executionOutput, err
   }
  time.Sleep(pollDelay)
}
```

It's less lines and I think more clear?

rleighton: Good point. Fixed.

PR #853, opened 2020-05-07
Add titles to Panther pages

### LINKED ISSUE #840: Add title to Panther tabs

Pages don't have a title. Instead it always says "Panther Dashboard". It would be helpful if we could at least show some basic information, so that when users see their history (or have multiple tabs open), they can understand what's what

### Describe your use cases

Everyday use of Panther

As an analyst, daily

Leaving everything as is

### References

Are there any other GitHub issues, whether open or closed, that are related to the problem you've described above or to the suggested solution? If so, please create a list below that mentions each of them.

### Additional context

It involves adding some super basic meta tags on every page

3nvi:
## Background

Currently if you have multiple tabs open, they all say "Panther Dashboard". This is the same when you are viewing your browser's history. This can be a bit confusing, especially when you continuously want to tab between different pages in Panther

This PR adds titles to all pages in Panther. Closes #840 

## Changes

- Add HOC for wrapping pages with SEO-related data
- Wrap all pages with `withSEO` HOC

## Testing

- Went through all pages and verified that each had a title


3nvi:
> LGTM! Only minor comment which probably we don't have control over it, is that when opening a new tab, the title changes only when user first visits the tab.

@alexmylonas  Unfortunately we don't yet do SSR so we can't serve the HTML with the correct title to begin with. We could if we wanted to, but I feel we have more important things to worry about.

Currently the title gets added when JS runs, so if chrome "suspends" a tab when it opens (i.e. doesn't run its JS until its in the foreground), then I can't do much about it :/

PR #856, opened 2020-05-07
Refactor log processor moving sqs functions to awsbatch/sqsbatch for reuse

rleighton:
## Background
The parquet work in `panther-enterprise` needs to use sqs batch functions. This PR moves the code used by the log processor to `pkg/awsbatch` to enable reuse.

## Changes
- move sqs functions from log processor to `pkg/awsbatch` 

## Testing
mage test:ci
deployments in dev account from version in `panther-enterprise`


internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -102,27 +99,18 @@ func streamEvents(sqsClient sqsiface.SQSAPI, deadlineTime time.Time, event event
...
 
 			// keep reading from SQS to maximize output aggregation
-			messages, messageReceipts, overLimit, err := readSqsMessages(sqsClient)
+			messages, messageReceipts, err := sqsbatch.ReceiveMessage(sqsClient, common.Config.SqsQueueURL, sqsWaitTimeSeconds)
 			if err != nil {
 				readEventErrorChan <- err
 				return
 			}
 
-			// just stop processing if the queue has too many requests in flight (and delete from queue)
```
rleighton, line 111: This was removed because after reading the docs it only applies to short-poll calls not log poll calls (we do long poll).

pkg/awsbatch/sqsbatch/delete_message_batch.go
```diff
@@ -0,0 +1,85 @@
...
+func DeleteMessageBatch(sqsClient sqsiface.SQSAPI, queueURL string, messageReceipts []*string) {
+	messageReceiptChan := make(chan *string, maxMessages)
+	go func() {
+		for _, messageReceipt := range messageReceipts {
+			messageReceiptChan <- messageReceipt
+		}
+		close(messageReceiptChan) // tells streamDeleteSqsMessages() we are done
+	}()
+
+	streamDeleteSqsMessages(sqsClient, queueURL, messageReceiptChan) // blocks until done
```
kostaspap, line 39: I'm thinking that using channels and a separate Go routine might be overkill here..;
The only thing the go routine does is push messages to the channel - where the main routine is reading the messages from. The actual API calls are performed in serial fashion in the main program. 

I think you could get the code of the `streamDeleteSqsMessages` and put it in here. You just have to change the loop (line 52) to iterate over the `messageReceipts` and that should be it?

pkg/awsbatch/sqsbatch/delete_message_batch.go
```diff
@@ -0,0 +1,85 @@
...
+	for i := range deleteMessageBatchRequestEntries {
+		deleteMessageBatchRequestEntries[i] = &sqs.DeleteMessageBatchRequestEntry{
+			Id: aws.String(strconv.Itoa(i)), // preset ids within batch
+		}
+	}
+
+	for messageReceipt := range messageReceiptChan {
+		deleteMessageBatchRequestEntries = deleteMessageBatchRequestEntries[:batchCounter+1] // extend
+		deleteMessageBatchRequestEntries[batchCounter].ReceiptHandle = messageReceipt        // set
+		batchCounter++
```
kostaspap, line 55: This is not the number of batches - I think it is the number of events.

pkg/awsbatch/sqsbatch/delete_message_batch.go
```diff
@@ -0,0 +1,85 @@
...
+		if batchCounter == maxMessages {
+			deleteMessageBatch(sqsClient, queueURL, deleteMessageBatchRequestEntries)
+			// reset
+			batchCounter = 0
+			deleteMessageBatchRequestEntries = deleteMessageBatchRequestEntries[:0]
+		}
+	}
+	// the rest
+	if batchCounter > 0 {
+		deleteMessageBatch(sqsClient, queueURL, deleteMessageBatchRequestEntries)
```
kostaspap, line 65: If I understand this correctly the last batch of events can contain events from the "next-to-last" batch. I think if e.g. the last batch has 1 event, the remaining 19 events (receipt handles) will be from the batch before.

internal/log_analysis/log_processor/processor/stream.go
kostaspap: Nice!

pkg/awsbatch/sqsbatch/delete_message_batch.go
rleighton: True, the was left over from when the receipts were in sets. Will change

pkg/awsbatch/sqsbatch/delete_message_batch.go
rleighton: I'll rename to `messagesInBatchCounter`

pkg/awsbatch/sqsbatch/delete_message_batch.go
rleighton: Can you elaborate? 

This statement is to get the the last buffered batch. We fill the buffer, a write when it is full. If the the counter = 0 , then we wrote and the buffer is empty, if not we filled the buffer but not completely and we need to write it. The tests exercise this, for example 11 events will be 2 writes, the first of 10 and the last of 1. At least that is what is intended.

kostaspap: You are correct, I misread the code. Due to the line
```
deleteMessageBatchRequestEntries = deleteMessageBatchRequestEntries[:messagesInBatchCounter+1]
```
the `deleteMessageBatchRequestEntries` will only have as many entries as receipt handles.

Issue #847, opened 2020-05-06
Custom cfn resource: certificates


austinbyers:
### Description

In order to achieve an installable deployment package, self-signed certificates need to be generated as a custom resource and not via `mage deploy`

PR #857, opened 2020-05-07
Add custom certificate CFN resource

austinbyers:
## Background

Use a CloudFormation custom resource for the default self-signed certificate to enable a self-contained deployment model.

Closes: #847 

This is a much more natural model in any case - our current way of handling certificates requires adding them as a parameter and an output to the same stack and then listing all ACM/IAM certificates when manually tearing down. Now CloudFormation neatly tracks which cert is actually used.

## Changes

- Add a `panther-cfn-custom-resources` Lambda function which supports the `Custom::Certificate` resource type (more coming soon). This function is created in the `bootstrap-gateway` stack (it can't be in the first stack because an S3 bucket needs to exist first so `mage deploy` can package it when deploying from source)
    - Moved certificate create/destroy out of `mage` and into the Lambda function
- Moved the load balancer listener (the thing which actually uses the cert) from `bootstrap` to the `web` stack because now it depends on the custom-resource
- Bugfix: Create the correct type of changeset after a bad stack has been automatically deleted by `mage deploy` (I know @nhakmiller ran into this issue recently)
- Upgrade our `backoff` to v4 - we were still on v2
- Fix deploy/teardown progress indicators to be consistent

## Testing

The resource shows up in CloudFormation like this:

Screen Shot 2020-05-04 at 5 56 09 PM

- Deploy with cert specified in the config file
- Deploy + teardown with generated IAM certs
- Deploy + teardown with generated ACM certs
- Generated errors when creating and deleting certs to see how they look in CloudFormation:

Screen Shot 2020-05-04 at 5 56 01 PM

### Migration
- Deployed this branch on top of a `master` deployment to make sure the migration is smooth

The web URL will not change, but it will be unreachable during the deploy as the listener moves to a different stack. If someone specifies a certificate in the settings file, it is used exactly the same as before.

If a user was relying on the default self-signed certificate, that cert will no longer be used - a new one will be generated instead. They will need to manually delete their old cert if they care to do so.

We could "import" their existing self-signed cert if we want by changing the "CREATE" functionality of the Custom::Certificate resource to scan for existing certs before generating a new one. I chose not to do this for a few reasons:

- Production deployments shouldn't be relying on self-signed certs anyway, no need to have lots of code for backwards-compatibility with the non-secure default
- Here we have a guarantee that the private key is completely secure - it's generated in Lambda memory (never stored on disk), then sent to ACM/IAM and never seen again. This is more secure for default deployments because now the key can't be stolen from a developer laptop.


deployments/web_server.yml
```diff
@@ -56,10 +63,49 @@ Parameters:
     Default: 80
     Description: The exposed port of the application, that will be used by the VPC & Container
 
+Conditions:
+  CreateCertificate: !Equals [!Ref CertificateArn, '']
+
 Resources:
+  SelfSignedCertificate:
+    Condition: CreateCertificate
+    Type: Custom::Certificate
```
nhakmiller, line 72: I'm not sure what the right phrasing is, but is the namespace of custom resources restricted to one stack, or should we maybe name this `Custom::PantherCertificate` so if someone already has a `Custom::Certificate` resource there won't be any overlap?

deployments/bootstrap_gateway.yml
```diff
@@ -69,6 +85,58 @@ Resources:
...
+              # not support using resource tags as a condition.
+              # So this is as narrow as this resource can get:
+              Resource: !Sub arn:${AWS::Partition}:acm:${AWS::Region}:${AWS::AccountId}:certificate/*
+            - Effect: Allow
+              Action:
+                - iam:DeleteServerCertificate
+                - iam:UploadServerCertificate
+              Resource:
+                - !Sub arn:${AWS::Partition}:iam::${AWS::AccountId}:server-certificate/panther/*
+                - !Sub arn:${AWS::Partition}:iam::${AWS::AccountId}:server-certificate/Panther*
```
nhakmiller, line 132: Do we still need this permission? It appears we've standardized on `/panther/{region}/Panther-{timestamp}`, I think we only had this other permission for backwards compatibility with before we were using a path but it sounds like this update breaks that backwards compatibility anyways.

deployments/bootstrap_gateway.yml
```diff
@@ -69,6 +85,58 @@ Resources:
...
       LayerName: panther-analysis
 
+  CustomResourceFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: ../bin/internal/core/custom_resources/main
+      Description: Custom CloudFormation resources when deploying Panther
+      Environment:
+        Variables:
+          DEBUG: !Ref Debug
```
kostaspap, line 95: Should we always have debug mode "on" for this function?

internal/core/custom_resources/resources/certificate.go
```diff
@@ -0,0 +1,250 @@
...
+	default:
+		// There is nothing to update on an existing certificate.
+		certArn := event.PhysicalResourceID
+		return certArn, map[string]interface{}{"Arn": certArn}, nil
+	}
+}
+
+// Generate a self-signed certificate and private key.
+func generateKeys() ([]byte, []byte, error) {
+	now := time.Now()
```
kostaspap, line 74: `time.Now().UTC()` just to be sure.

internal/core/custom_resources/main/lambda.go
```diff
@@ -0,0 +1,52 @@
+package main
```
kostaspap, line 1: Should we just move this Lambda to the tools section? Thinking that "internal" is really the actual panther logic, this is no more than a deployment tool.

deployments/bootstrap_gateway.yml
rleighton: I like that, since this just runs on deployments.

deployments/bootstrap_gateway.yml
```diff
@@ -69,6 +85,58 @@ Resources:
       Description: Pip libraries available to the Python analysis/remediation functions
       LayerName: panther-analysis
 
+  CustomResourceFunction:
```
rleighton, line 88: I'm assuming that this lambda will just keep being extend with functionality, shouldn't it be in `bootstrap.yml` ? I know the first use is certs (hence in `bootstrap_gateway.yml` but will add more stuff here.

Does it make sense to move this to its own stack that is deployed first called `bootstrap_cfn.yml` ?

deployments/web_server.yml
austinbyers: The resource type is just passed as a parameter to the custom resource handler (in this case, our Lambda function). It's not registered with CloudFormation or anything; CFN completely ignores the type except to show it in status messages

A conflict isn't possible even if someone made changes to one of the Panther stacks, as long as they provided a different custom resource handler:

```yml
Resources:
  SelfSignedCertificate:
    Type: Custom::Certificate
    Properties:
      ServiceToken: .../panther-cfn-custom-resources

  MyOwnSpecialCertificate:
    Type: Custom::Certificate
    Properties:
      ServiceToken: .../a-completely-different-lambda-function
```

austinbyers: Since we control the namespacing, having everything prefixed with `Panther` is unnecessarily verbose, in my opinion

deployments/bootstrap_gateway.yml
austinbyers: I'm convinced this is actually a bug with IAM itself - yes we do need both. A fresh deploy/teardown without that second permission will fail with permission denied.

My best guess why - API calls for server certificates reference the certificate name, not the path nor arn:

```
dev -- aws iam delete-server-certificate --server-certificate-name arn:aws:iam::XXX:server-certificate/panther/us-west-2/PantherCertificate-2020-04-29T16-56-35

An error occurred (ValidationError) when calling the DeleteServerCertificate operation: The specified value for serverCertificateName is invalid. It must contain only alphanumeric characters and/or the following: +=,.@_-
```

The certificate name has to be specified:

```
dev -- aws iam delete-server-certificate --server-certificate-name PantherCertificate-2020-04-29T16-56-35
```

So, even though the arn includes the `/panther/` path, the *name* starts with `Panther` and I'm guessing IAM isn't smart enough to realize the arn includes a path which is authorized.

------
I'll add a comment briefly explaining why we need it

deployments/bootstrap_gateway.yml
austinbyers: Sure, I can keep it on, but there are no debug messages in any case

internal/core/custom_resources/resources/certificate.go
austinbyers: I was just copying the same code we had in `mage deploy`, which used the local timezone. It turns out the x509 library does the [conversion](https://golang.org/src/crypto/x509/x509.go?s=66383:66500#L2141) for you, so it makes no difference.

I'll change it just to reduce confusion

internal/core/custom_resources/main/lambda.go
austinbyers: `internal` is really a Go idiom that means "all of the application code which shouldn't be directly importable by other projects"

tools are meant to be more standalone binaries like `mage`. This Lambda function is deployed with Panther itself, it has to be managed just like the rest of the application. I'd prefer to keep it with all the other Lambda functions

austinbyers: See https://github.com/golang-standards/project-layout for more context

deployments/bootstrap_gateway.yml
austinbyers: Yes, I wanted to put it in the very first stack (`bootstrap.yml`) but we have a chicken-and-the-egg problem.

When deploying from SAR, everything is pre-packaged so it doesn't matter where the function is. BUT, when deploying from source, you need an S3 bucket to exist before the stack containing the function can be deployed because the function source has to be uploaded somewhere. Packaging happens locally - the Lambda zipfile must be uploaded to S3 before you can even start CloudFormation.

If we put it in bootstrap, then, we'd have to deploy that whole stack twice: once for just the S3 buckets and then again with the custom resource source packaged in the "source" S3 bucket. Every resource except S3 would have to be tagged with a "NotFirstTime" condition.

So one solution could be to move all S3 buckets to their own initial stack which is perhaps what we should have done in the first place. But that of course breaks existing deployments unless we try to fix it using the new CFN import functionality.

Instead, I put this function in the earliest possible stack - the stack right after `bootstrap`. `bootstrap-gateway` is a bit of a misnomer, but it essentially contains all "bootstrap" resources that require the source S3 bucket to exist ahead of time. This will work as long as nothing in "bootstrap" requires custom resources, and I've already checked that any such resources (for now) can be safely moved to another stack without issue.

Fortunately, lambda functions are easy to move between stacks, so if we need to move it in the future I think we can do so fairly easily. It's just an S3 bucket has to exist first

austinbyers: I added a comment in cfn explaining why it's placed there

austinbyers:
Deployed again and verified still works

deployments/bootstrap_gateway.yml
austinbyers: The "source" bucket is actually not required at all for standard packaged deployments. I'm open to the idea of moving it out of CloudFormation entirely, since it will only be used by developers in the very near future

PR #858, opened 2020-05-08
Fix for mapping JSON field names to Glue table column names and Athena views

alxarch:
## Background

There seem to be some (undocumented?) restrictions on field names in columns of type `struct` in Glue/Athena.
The issue came up while writing the parser for `GCP.AuditLog` with a field that conatined the `@` character (`@type`) that was found in a nested struct (`@payloadProto`) of a log entry.

 
## Changes

The proposed solution adds a `jsoniter.Extension` that maps field names of structs on JSON encoding only. Structs are deserialized using the field name specified in `json` struct tag and are remapped to 'normalized' field names upon JSON serialization.
The extension approach was chosen to have a generic solution to the problem without requiring per-case solutions on every log parser.

- Adds a `jsonutil` module to define the extension and registration helper
- Adds an `init()` block to `parsers` module to register the extension so that every other part of the code that uses `parsers` behaves correctly on `jsoniter.Marshal`
- Adds the same mapping function to `tools/cfngen/gluecf` so that glue tables use the proper field name
- Adds tests

## Testing

- Unit tests and `mage test:ci`
- [pending] mage deploy
- [pending] E2E tests


alxarch:
Related comment: https://github.com/panther-labs/panther/pull/842#discussion_r421539314

tools/cfngen/gluecf/schema_test.go
```diff
@@ -220,11 +236,9 @@ func TestInferJsonColumns(t *testing.T) {
 	resetField(cols) // reset the Field, not needed for tests
 
 	// uncomment to see results
-	/*
-		for _, col := range cols {
-			fmt.Printf("{Name: \"%s\", Type: \"%s\",Comment: "test field"},\n", col.Name, col.Type)
-		}
-	*/
+	for _, col := range cols {
```
rleighton, line 239: Put comment back

alxarch: > I would suggest the documentation changes go in this PR so we can merge it all at once.
I am unsure as to how to proceed with the docs changes. The whole `jsonschema` approach is incompatible with this change and feels like a quick fix to the complexity of the recursion into structs. The actual 'mistake' in the docs currently is one letter in one field in one log entry and it's blocking AuditLogs for no reason. I would suggest we describe the situation about the field in the `description` struct tag so that is shows up in the docs for now and have a separate PR that removes the jsonschema generator in favor of an actual recursive docs generator.

internal/log_analysis/log_processor/jsonutil/jsonutil.go
```diff
@@ -0,0 +1,78 @@
...
+	"strings"
+	"sync"
+
+	jsoniter "github.com/json-iterator/go"
+)
+
+// TODO: [parsers] Add more mappings of invalid Athena field name characters here
+// NOTE: The mapping should be easy to remember (so no ASCII code etc) and complex enough
+// to avoid possible conflicts with other fields.
+var athenaStringReplacer = strings.NewReplacer(
```
kostaspap, line 31: I would change the name to a more generic "fieldNameReplacer" or similar. We encountered the issue while working with Athena but now this change will apply universally (either if we use Athena to query the data, or S3 Select or any other type of backend).

This comment applies to all "Athena"-named structs/functions in this file.

internal/log_analysis/log_processor/jsonutil/jsonutil.go
```diff
@@ -0,0 +1,78 @@
...
+	})
+}
+
+type encoderRewriteStrategyExtension struct {
+	jsoniter.DummyExtension
+	translate func(string) string
+}
+
+// UpdateStructDescription maps output field names to
+func (extension *encoderRewriteStrategyExtension) UpdateStructDescriptor(structDescriptor *jsoniter.StructDescriptor) {
```
kostaspap, line 61: Is the scope of this global? Can we restrict it somehow e.g. to apply only for parser outputs?

alxarch: The scope is global. I can look into adding a `parsers.JSONIter()` or a `parsers.MarshalJSON` function that returns a configured pool and users would use that. It could also be made local if the log_processor was reworked to have explicit dependencies. (ie `jsoniter.API` argument instead of `jsoniter.Marshal`, `*zap.Logger` argument instead of `zap.L()` etc). I think the second approach is more thorough but as the code base stands we can go for the first solution.

alxarch: I tried introducing a scoped `parsers.JSON` variable of type `jsoniter.API` to add the extension only there. It complicates things since the `jsoniter.Marshal` calls are scattered around the code base. I think the only call that really matters is the one on `destinations/s3.go`. I think that with the current approach of using global variables instead of passing dependencies to instances (ie `S3Destination`) such a variable should be put in `log_processor` common and should be used everywhere that it's needed.

internal/log_analysis/log_processor/jsonutil/jsonutil.go
alxarch: I think this discussion is linked with the scope related comment.

tools/cfngen/gluecf/schema_test.go
rleighton: Can you explain? Are you sure it does not work and your issue was not the issue you had with `mage`. I looked at the code and it uses the function you changed to generate the schemas:
https://github.com/panther-labs/panther/blob/master/tools/mage/doc.go#L138

I don't see why it does not work for the docs.

alxarch: It's got to do with this https://github.com/panther-labs/panther/blob/be8576758acd9780cc464d82fa9b8be0213a0674/tools/mage/doc.go#L211
The `InferJSONColumns` is only used for the top level fields. The fields in nested structs use `jsonschema.Reflector`

alxarch: The `@type` field that revealed the bug (and is still wrong in the generated docs) was not at top level but nested inside `payloadProto` field

rleighton: Sorry, still confused. This the same way we generate schemas:
https://github.com/panther-labs/panther/blob/master/tools/cfngen/gluecf/gluecf.go#L150

alxarch: The problem is not with `InferJSONColumns`, that works fine. The issue comes up in the `mage/doc.go` code because to describe the nested fields it re-parses the struct using `jsonschema.Reflector` and that ignores the names of struct fields in the glue column definitions as returned by `InferJSONColumns`. And then the info is serialized to JSON using `json.Marshal` to be written in the  in the doc buffer.

internal/log_analysis/log_processor/destinations/s3_test.go
```diff
@@ -230,7 +229,7 @@ func TestSendDataToS3BeforeTerminating(t *testing.T) {
 	expectedS3Notification := models.NewS3ObjectPutNotification(destination.s3Bucket, *uploadInput.Key,
 		len(expectedBytes))
 
-	marshaledExpectedS3Notification, _ := jsoniter.MarshalToString(expectedS3Notification)
+	marshaledExpectedS3Notification, _ := parsers.JSON.MarshalToString(expectedS3Notification)
```
kostaspap, line 232: Actually this could stay as it is - this is the marshaling of the notification sent when we create new object in S3 (https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/destinations/s3.go#L288)

internal/log_analysis/log_processor/jsonutil/jsonutil.go
```diff
@@ -0,0 +1,56 @@
...
+	jsoniter.DummyExtension
+	translate func(string) string
+}
+
+// UpdateStructDescription maps output field names to
+func (extension *encoderNamingStrategy) UpdateStructDescriptor(structDescriptor *jsoniter.StructDescriptor) {
+	for _, binding := range structDescriptor.Fields {
+		tag, hastag := binding.Field.Tag().Lookup("json")
+
+		// toName := binding.Field.Name()
```
kostaspap, line 43: Can remove?

internal/log_analysis/log_processor/parsers/parsers.go
```diff
@@ -35,3 +42,32 @@ type LogParser interface {
...
+
+// TODO: [parsers] Add more mappings of invalid Athena field name characters here
+// NOTE: The mapping should be easy to remember (so no ASCII code etc) and complex enough
+// to avoid possible conflicts with other fields.
+var fieldNameReplacer = strings.NewReplacer(
+	"@", "_at_sign_",
+	",", "_comma_",
+	"`", "_backtick_",
+	"'", "_apostrophe_",
+)
```
kostaspap, line 54: I know it's nitpicking, but since this change is "here to stay", it would be better to avoid having underscores that are not necessarily needed. It would be better if the behavior would be the following:  
`@type` -> `at_sign_type`
`type@` -> `type_at_sign`
`t@pe` -> `t_at_sign_pe`

internal/log_analysis/log_processor/destinations/s3_test.go
alxarch: Fixed

internal/log_analysis/log_processor/parsers/parsers.go
alxarch: I don't think this is nitpicking. It's user-facing so it should be as convenient as possible for the *user* not us :).
The `strings.Replacer` is one of the most efficient ways to replace predefined mappings of text in `go`. It internally compiles a tree of possible paths (like a state machine) to have good performance. That was the reason it was chosen instread of a `regexp` based approach. If my understanding of how `jsoniter` works internally is correct we don't really care about performance though since this mapping should happen only once when the `StructDescriptor` is registered. The only way I can think of to implement such logic without using a brute-force `strings.Trim` after replacement (that could modify unmatched json fields that end/start with '_') would require multiple (similar) regexp replacements. Any suggestions as to how to implement this without having to define each mapping in 3 different places?

alxarch: Ah! I got it, we can check if the name was modifield and only then apply the `strings.Trim`. On it.

tools/cfngen/gluecf/schema_test.go
rleighton: Ah, now I remember. Jack wanted a JSON schema in the docs not the HIVE/Presto struct format. Let me look at the (very ugly) code and see if we can do something about it.

rleighton: I never liked the JSON schema and we really do not need specifically that, just a human readable format for nested arrays/maps/structs. Let me spend a few hours changing the doc code to re-format the Glue types to something less ugly, then this is a non-issue.

internal/log_analysis/log_processor/parsers/parsers.go
```diff
@@ -35,3 +42,36 @@ type LogParser interface {
...
+
+func RewriteFieldName(name string) string {
+	result := fieldNameReplacer.Replace(name)
+	if result == name {
+		return name
+	}
+	return strings.Trim(result, "_")
```
kostaspap, line 61: We want to remove the `_` that exists the beginning or at the end. We don't want to remove all intermediary `_`. I think you need strings.TrimSuffix , strings.TrimPrefix

tools/cfngen/gluecf/schema_test.go
alxarch: See if you can come up with something. I think the task of recursively walking a struct through reflection is used extensively in various parts of tools. Some of the logic is overlapping and maybe the is some library that can help to structure the process. Maybe this could help https://godoc.org/github.com/jmoiron/sqlx/reflectx

internal/log_analysis/log_processor/parsers/parsers.go
alxarch: `strings.Trim` doesn't affect 'inside' the string.

kostaspap:
Code looks good to me. Can you make sure that an E2E test works? (Pushing some GCP Audit logs to S3 and being able to query them in Athena)

alxarch:
> Can you make sure that an E2E test works? (Pushing some GCP Audit logs to S3 and being able to query them in Athena)
Unfortunately the GCP Audit log parser is not included in this branch.

kostaspap:
Already resolved

PR #859, opened 2020-05-08
Several updates to source API

kostaspap:
## Background

Bringing on several improvements to source API. Improvements were inspired by enterprise version. 

## Changes

- Several refactorings of the source_api

## Testing

- mage test:ci
- mage deploy


internal/core/source_api/api/delete_integration_test.go
```diff
@@ -315,22 +296,22 @@ func generateDDBAttributes(integrationType string) map[string]*dynamodb.Attribut
...
+		statements := make([]awssqs.SqsPolicyStatement, len(accountIDs))
 		for i, accountID := range accountIDs {
-			statements[i] = SqsPolicyStatement{
+			statements[i] = awssqs.SqsPolicyStatement{
 				SID:       fmt.Sprintf("PantherSubscriptionSID-%s", accountID),
 				Effect:    "Allow",
 				Principal: map[string]string{"AWS": "*"},
 				Action:    "sqs:SendMessage",
-				Resource:  env.LogProcessorQueueArn,
+				Resource:  "*",
```
rleighton, line 306: Can you add comment about why this needs to be '*' and the safety of that?

kostaspap: The policy is attached to an SQS queue - so the policy applies only to that SQS queue. I used to put the SQS Queue ARN in there but I realized the '*' would work as well. I did it so I could get rid of an environment variable (`LogProcessorQueueArn`)

PR #860, opened 2020-05-08
Refactor pkg/awsglue

rleighton:
## Background
As part of the Parquet implementation, the pkg/awsglue was refactored to make it better organized, remove redundant code and add useful entry points.

## Changes
- Refactored code, no substantive functionality changes

## Testing
mage test:ci
mage deploy, checked dashboards
mage deploy in Parquet branch


pkg/awsglue/glue_table_metadata.go
```diff
@@ -19,48 +19,19 @@ package awsglue
  */
```
kostaspap, line 19: nit: Since the package/folder is `awsglue` you could just name this file `table_metadata`.

pkg/awsglue/glue_table_metadata.go
```diff
@@ -242,21 +175,15 @@ func (gm *GlueTableMetadata) SyncPartitions(glueClient glueiface.GlueAPI, s3Clie
 					continue
 				}
 
-				// leave _everything_ the same except the schema, and the serde info
+				// leave _everything_ the same except the schema, and the serde info to get column mappings
 				storageDescriptor := *getPartitionOutput.Partition.StorageDescriptor // copy because we will mutate
 				storageDescriptor.Columns = columns
```
kostaspap, line 180: This two lines could be inside the `IsJSONPartition` method

pkg/awsglue/glue_timebin.go
```diff
@@ -77,3 +82,42 @@ func (tb GlueTableTimebin) PartitionValuesFromTime(t time.Time) (values []*strin
...
+		MaxKeys: aws.Int64(1), // look for at least 1
+	}
+	var hasData bool
+	err = client.ListObjectsV2Pages(inputParams, func(page *s3.ListObjectsV2Output, isLast bool) bool {
+		for _, value := range page.Contents {
+			if *value.Size > 0 { // we only care about objects with size
+				hasData = true
+			}
+		}
+		return false // "To stop iterating, return false from the fn function."
```
kostaspap, line 119: I am wondering whether this should be `!hasData`. Is it possible that the list operation returns 0 elements in the first page, but 1+ elements in the second page of results? 
DDB Queries + Filtering can display the above behavior, where you might need to go through several empty pages to find a page with data, wondering if S3 with Prefix works like that.

pkg/awsglue/partition.go
```diff
@@ -79,30 +68,19 @@ func (gp *GluePartition) GetGlueTableMetadata() *GlueTableMetadata {
...
+func (gp *GluePartition) GetPartitionLocation() string {
+	return "s3://" + gp.s3Bucket + "/" + gp.gm.GetPartitionPrefix(gp.time)
 }
 
-func (gp *GluePartition) GetPartitionLocation() string {
-	tablePrefix := getTablePrefix(gp.datatype, gp.tableName)
-	prefix := "s3://" + gp.s3Bucket + "/" + tablePrefix
-	for _, partitionField := range gp.partitionColumns {
-		prefix += partitionField.Key + "=" + partitionField.Value + "/"
+func GetPartitionLocation(s3Path string) (string, error) {
```
kostaspap, line 78: A bit confused about this method's functionality.  We pass in an S3Path to get back...  an s3 path? We also have the method in lines 173-180, do we really need this method as well?

pkg/awsglue/wrappers.go
```diff
@@ -0,0 +1,146 @@
...
+
+func DeleteTable(client glueiface.GlueAPI, databaseName, tableName string) (*glue.DeleteTableOutput, error) {
+	deleteTableInput := &glue.DeleteTableInput{
+		DatabaseName: aws.String(databaseName),
+		Name:         aws.String(tableName),
+	}
+	return client.DeleteTable(deleteTableInput)
+}
+
+func TableHasPartitions(client glueiface.GlueAPI, databaseName, tableName string) (hasData bool, err error) {
```
kostaspap, line 53: This method is only called by another part of the code, right?  As I see it is called only in `HasPartitions`. `HasPartitions` is already a wrapper for the Glue call. Not sure we need a second wrapper here... We add another layer of abstraction without real need as I see. Same goes for `GetTable`, `DeleteTable`, `CreatePartition`. I feel they complicate the codebase and will make any future refactoring more challenging.

pkg/awsglue/partition.go
kostaspap: Ah I think I figured out. Could you add a comment on the method to clarify how it works?

pkg/awsglue/awsglue.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/panther-labs/panther/api/lambda/core/log_analysis/log_processor/models"
+)
+
+// This file registers the Panther specific assumptions about tables and partition formats with associated functions.
+
+const (
+	logS3Prefix       = "logs"
+	ruleMatchS3Prefix = "rules"
+
+	LogProcessingDatabaseName        = "panther_logs"
```
austinbyers, line 33: In theory, Go packages within `pkg` are meant to be completely generic - other projects could import and use them without having anything to do with Panther. See https://github.com/golang-standards/project-layout for more context.

That's why we have our `testutils` and `genericapi` and `awsbatch` libraries here - those could all be useful outside the context of Panther. This library, however, is necessarily only relevant for a Panther deployment. It just needs to be shared among Panther components.

We don't have a specific place for sharing internal packages, I'm open to ideas. I'm fine with leaving it here, it just means we're no longer following the standard we originally were. None of the other packages in this folder are ready for external use anyway, that would likely be a separate effort and then they might move to their own repo.

pkg/awsglue/glue_timebin.go
austinbyers: I had the same thought. Reading through the SDK, it seems fine the way it is implemented, but to be safe/consistent let's not make the assumption that an empty page means no more results.

Since you already set `MaxKeys=1`, just `return !hasData` to let the paginator keep going. If there are potentially empty pages, this guarantees we go through them. If there are not, then it doesn't cost anything extra

pkg/awsglue/wrappers.go
austinbyers: Similarly, `DeleteTable` above is exactly the same as using the client directly, the wrapper is not doing any extra boilerplate work. Using the wrapper function might make it fit on fewer lines is all, but I agree that in that case the abstraction may do more harm than good

pkg/awsglue/glue_table_metadata.go
rleighton: Can you explain? Those 2 lines mutate the storageDescriptor while `IsJSONPartition` is simply a predicate. Also, `IsJSONPartition`  is used in other contexts in the Parquet code.

pkg/awsglue/wrappers.go
rleighton: These wrappers are used by the Parquet code and make that code much smaller and easier to read  by avoiding all the setup.

pkg/awsglue/glue_table_metadata.go
rleighton: Done

pkg/awsglue/glue_timebin.go
rleighton: Done

pkg/awsglue/awsglue.go
rleighton: Yeah, this evolved from a generic utility package to a _very_ opinionated package. I've moved it under log_analysis

pkg/awsglue/glue_table_metadata.go
kostaspap: I thought that the only mutation was changing SerDe to be JSON SerDe, in which. case the line `storageDescriptor := *getPartitionOutput.Partition.StorageDescriptor ` could be inside the `IsJSONPartition` condition. 
But I see that we also modify the Columns (I didn't notice this when I wrote this comment). Is there a reason why we modify the columns?

pkg/awsglue/partition.go
rleighton: done

pkg/awsglue/glue_table_metadata.go
rleighton: This is SyncPartitions (`mage glue:sync`) , so that is the purpose of this code , to set partition columns to match parent table.

pkg/awsglue/wrappers.go
kostaspap: This can be argued for all AWS SDK calls though. The wrapper doesn't remove the complexity, it hides it under another layer of abstraction. That extra layer itself however has its own maintenance cost. 

I would much prefer if we don't do that, it makes the overall code more complex than it needs to be at this moment.

internal/log_analysis/awsglue/glue_timebin.go
```diff
@@ -77,3 +82,42 @@ func (tb GlueTableTimebin) PartitionValuesFromTime(t time.Time) (values []*strin
...
+		return fmt.Sprintf("year=%d/month=%02d/day=%02d/hour=%02d/", t.Year(), t.Month(), t.Day(), t.Hour())
+	case GlueTableDaily:
+		return fmt.Sprintf("year=%d/month=%02d/day=%02d/", t.Year(), t.Month(), t.Day())
+	default:
+		return fmt.Sprintf("year=%d/month=%02d/", t.Year(), t.Month())
+	}
+}
+
+// PartitionHasData checks if there is at least 1 s3 object in the partition
+func (tb GlueTableTimebin) PartitionHasData(client s3iface.S3API, t time.Time, tableOutput *glue.GetTableOutput) (bool, error) {
```
kostaspap, line 99: This method is mixing different abstraction layers and clients/apis (`s3iface.S3API`, `glue.GetTableOutput`, `time.Time`, `GlueTableTimebin`) .
I am thinking that the `awsglue.GluePartition` already has the location, time, GlueTableTimebin, so this could be be a method of that like `func (p *GluePartition) HasData(client s3iface.S3API) (bool, error)`. Makes sense?

pkg/awsglue/glue_table_metadata.go
kostaspap: Ah ok, was not sure why this was required

internal/log_analysis/awsglue/glue_timebin.go
rleighton: This is used when the caller looks up a table in the glue catalog, has a timebin (hourly) and needs to know if there are partitions.  We have no constructor today that will build a awsglue.GluePartition from a timebin and a table path. We could add one, just to support this call but that will make the caller more verbose, because it needs to make a partition and then call this code. While this makes the pkg a tiny bit cleaner, this is now being pedantic.  I'll do it because I'm always happier with the results when I do your pedantic requests (despite the marginal value, and a bit of grumbling at the extra work).

pkg/awsglue/wrappers.go
rleighton: You have not seen the Parquet code which makes extensive use of these and does not use the other abstractions as much because it is dealing directly with partitions. Without the wrappers the calling code functions get 2X longer for no good reason (all of this filling out inputs). Makes them hard read. Rather than:
```
bar := foo()
```
we have:
```
fooInput := {
qsqw
qwq
qwqw
qqw}
bar := foo(fooInput)
```
If you do a few of these calls it makes a simple function very hard to read.

rleighton: Also, this package is our very opinionated code. It makes sense here.

internal/log_analysis/awsglue/table_metadata.go
```diff
@@ -242,21 +175,15 @@ func (gm *GlueTableMetadata) SyncPartitions(glueClient glueiface.GlueAPI, s3Clie
...
-					TableName:          aws.String(gm.tableName),
-					PartitionInput:     partitionInput,
-					PartitionValueList: values,
+				// we need to update the SerDeInfo for JSON partitions to get the column mappings
+				if IsJSONPartition(&storageDescriptor) {
+					storageDescriptor.SerdeInfo = tableOutput.Table.StorageDescriptor.SerdeInfo
 				}
-				_, err = glueClient.UpdatePartition(updatePartitionInput)
+				_, err = UpdatePartition(glueClient, gm.databaseName, gm.tableName, values,
+					&storageDescriptor, nil)
```
kostaspap, line 186: I am thinking that we could check if the storage descriptor has been modified and issue the "UpdatePartition" only if it was updated? This could provide significant speedup to generally expensive `glue:sync` process.

pkg/awsglue/wrappers.go
kostaspap: I made my initial comment in the assumption that this was used only once - I haven't seen how it is used in the Parquet code. Let's move forward with this then .

PR #861, opened 2020-05-08
Create universal web-image

### LINKED ISSUE #846: Universal web images
### Description

We currently build a custom docker image for each new deployment which includes string constants for the user's environment (AWS account ID, GraphQL API, etc). In order to have a static deployment package, we need a way to build the web image such that it can be published and shared and then accept configuration dynamically.

### Acceptance Criteria

- `mage deploy` continues to build the image from source. We need this for developers and users who choose to modify their source.
- The built image is generic and could be published to DockerHub/ECR

This issue is just about changing the built image itself, don't worry about automating its publication

3nvi:
## Background

This is a really significant PR since it affects both development & deployment of the panther UI. 

In short, we are no longer creating a "personalized" web image for each deployment, but we are using the same web image for all users. Personalization will come through the ENV variables passed to the front-end (NodeJS) server.

Closes #846 

## Context

One of the hardest bits  with front-end is that there is no concept of runtime. If you add an ENV var in the server that serves the front-end, the pre-compiled JS code **cannot** alter itself and adapt to the dynamic ENV. The front-end JS code **is what it is when it was built**. That's  why we had personalized images, since we needed to make sure that  the front-end JS assets are  personalized on "build time".

To bypass that, we did a small hack, that's actually a pretty common technique. We took the ENV vars that are passed on the server, add them as a `<script>` tag to the HTML that is  gonna be served by the  server and  then have the front-end JS assets  "read"  this script tag and "boot" their environment/configuration using  the data found there.

In short, previously all AWS related data  was provided on build-time. Now  the front-end reads the AWS  related stuff from an HTML inlined script tag.

This way, the JS assets are the same for everyone, while the `<script>` tag  contains the ENV information that the NodeJS server has (which is indeed dynamic, since it's a proper server :) ). This worked well for production deployments, since we could add the proper ENV vars to the ECS service, which the NodeJS server would then read, add them on a `<script>` tag in the HTML that it's going to serve and then have the front-end read them from this tag.

Unfortunately, the development side-of-things didn't include a NodeJS server so far. I tried to use the same server for development purposes, but ran into a lot of issues with live-updates. Thus, I kept development as is. This means that during development, the ENV vars are added to the `<script>` tag on **build-time** (using the contents of  the local `/out/.env.aws` file), while during production the build script doesn't add anything to the `<script>` tag, but lets the  NodeJS server handle it. This means that we have personalized development builds and universal production builds. We only care about production being  universal (since development is inherently personalized), so this works well.

In this PR I had to handle some edge cases and I've added comments as to why I did some "hacky" stuff.

## Changes

- Make NodeJS server pass ENV vars to a script tag during production
- Make webpack pass ENV vars to a script tag during development (not production anymore)
- replace all `process.env.` to `pantherConfig.` since we will be gathering all dynamic ENV vars in an object (can't use `process.env` anymore)
- Boot `pantherConfig` from `<script>` tag
- Adjust development & production scripts to remove personalization code

## Testing

- Successfully run development environment
- Successfully deployed using `mage setup deploy`


web/public/index.ejs
```diff
@@ -1,42 +1,43 @@
...
+                        fill="#5478F7"
+                    />
+                    <path
+                        d="M6.91481 2.70154L16 0V30C14.451 29.7599 11.2538 29.2642 10.8569 29.2024C10.4601 29.1407 10.065 28.2676 9.9171 27.8388L7.41084 22.2298C5.7139 21.0806 2.20514 18.6895 1.74566 18.319C1.28618 17.9485 1.25833 17.2384 1.30184 16.9297C1.60642 14.8628 2.2208 10.6467 2.24169 10.3173C2.26257 9.98799 2.19818 9.80274 2.16337 9.75129C2.00673 9.47684 1.43237 8.43911 0.388101 6.48371C-0.917242 4.03945 1.45848 0.797599 1.82398 0.437393C2.11637 0.149228 2.60718 0.317324 2.81604 0.437393C3.86031 1.07204 6.01674 2.38765 6.28825 2.5729C6.55976 2.75815 6.81909 2.73585 6.91481 2.70154Z"
+                        fill="#3561FE"
+                    />
+                </svg>
+            </div>
+        </div>
+        <script id="__PANTHER_CONFIG__" type="application/json"><%-JSON.stringify(PANTHER_CONFIG)%></script>
```
3nvi, line 41: Here is where both  Webpack (during Development) and Express (during Production) add  their variables.

web/scripts/build.js
```diff
@@ -17,21 +17,12 @@
...
 process.env.NODE_ENV = 'production';
-// Generate  a `PANTHER_VERSION` that the javascript error logging function running in the browser
-// is going to reference when reporting a crash
-process.env.PANTHER_VERSION = getPantherDeploymentVersion();
 
 // Add all the sentry-related ENV vars to process.env
 loadDotEnvVars('web/.env.sentry');
 
-// Add all the aws-related ENV vars to process.env
-loadDotEnvVars('out/.env.aws');
```
3nvi, line 32: We don't want to load the user's variables during build-time anymore

web/scripts/serve.js
```diff
@@ -78,8 +80,9 @@ app.get('/healthcheck', (req, res) => {
 });
 
 // Resolve all other requests to the index.html file
+app.set('view engine', 'ejs');
 app.get('*', (req, res) => {
-  res.sendFile(path.resolve(__dirname, '../dist/index.html'));
+  res.render(path.resolve(__dirname, '../dist/index.ejs'), getAppTemplateParams());
```
3nvi, line 85: `getAppTemplateParams()` is where we gather all the ENV and add them to the script tag above. If you observe, the script tag had a `<%-PANTHER_CONFIG%>` entry and  `getAppTemplateParams` returns an object  containing the key `PANTHER_CONFIG`

web/scripts/utils.js
```diff
@@ -27,31 +26,22 @@ function loadDotEnvVars(path) {
...
-function validateRequiredEnv() {
-  const requiredEnvs = [
-    'AWS_ACCOUNT_ID',
-    'AWS_REGION',
-    'WEB_APPLICATION_GRAPHQL_API_ENDPOINT',
-    'WEB_APPLICATION_USER_POOL_CLIENT_ID',
-    'WEB_APPLICATION_USER_POOL_ID',
-  ];
+function getAppTemplateParams() {
+  const PANTHER_CONFIG = {
```
3nvi, line 30: this is the aggregated keys we were talking about

web/src/apollo/authLink.ts
```diff
@@ -18,15 +18,16 @@
...
 import { createAuthLink, AUTH_TYPE } from 'aws-appsync-auth-link';
 
 /**
  * This link is here to add the necessary headers present for AMAZON_COGNITO_USER_POOLS
  * authentication. It essentially signs the Authorization header with a JWT token
  */
 const authLink = (createAuthLink({
-  region: process.env.AWS_REGION,
-  url: process.env.WEB_APPLICATION_GRAPHQL_API_ENDPOINT,
+  region: pantherConfig.AWS_REGION,
```
3nvi, line 29: we need to change all `process.env` to `pantherConfig` since we are gonna "gather" the injected ENVs in a variable and not read them as build-time ENVs

web/src/config.ts
```diff
@@ -19,11 +19,25 @@
...
 import Auth from '@aws-amplify/auth';
 
+// read the NodeJS-injected configuration and parse it back from JSON -> JS object. During development,
+// the NodeJS server doesn't contribute in that at all, but instead the HtmlWebpackLoader injects
+// them directly. There is a long-run inconsistency with EJS templates and webpack's ejs-loader,
+// where the "symbol" for unescaped entities is different when using ejs-loader (development) and
+// when you are just using EJS templates without a webpack loader (production). Just because we
+// don't know whether we arrived here from development or production, we are doing the `&quot;`
+// replace you see to make sure we convert an escaped entity back to its original value
+export const pantherConfig = JSON.parse(
```
3nvi, line 29: and that's where we boot the config from. We read the contents  of the script with the ID `__PANTHER__CONFIG__` and parse its contents (it's safe to parse, since it's not user input)

web/webpack.config.js
```diff
@@ -223,37 +224,36 @@ module.exports = {
...
+      inject: true,
+      // We are using `html-loader` here for an EJS template (instead of the `ejs-loader`).
+      // The reason for that is that we have an EJS template, filled with template parameters
+      // that are going to be replaced in runtime. HtmlWebpackPlugin throws an error if those
+      // parameters are not provided during build time and there was no way to get around it.
+      // Basically, if those were undefined during build time, the EJS failed to compile. The
+      // only way to bypass that is to force HtmlWebpackPlugin to treat this template as a
+      // simple HTML and leave those template parameters untouched. Of course, we couldn't just
+      // remove this plugin entirely, since we need it for the CSS/JS tag injection
+      template: isEnvDevelopment
```
3nvi, line 237: in this line we are choosing to:

* replace vars at  build-time during development
* ignore vars and let NodeJS handle them during production

web/webpack.config.js
```diff
@@ -223,37 +224,36 @@ module.exports = {
...
+        ? path.resolve(__dirname, 'public/index.ejs')
+        : `html-loader!${path.resolve(__dirname, 'public/index.ejs')}`,
+      filename: isEnvDevelopment ? 'index.html' : 'index.ejs',
+      templateParameters: isEnvDevelopment ? getAppTemplateParams() : undefined,
```
3nvi, line 241:
```js
templateParameters: isEnvDevelopment ? getAppTemplateParams() : undefined,
```

If we chose to do replacement, that's where the values of the vars are gonna be found

tools/mage/deploy_frontend.go
```diff
@@ -64,13 +64,22 @@ func deployFrontend(
 		return nil, err
 	}
 
+	gitVersion, err := sh.Output("git", "describe", "--tags")
+	if err != nil {
+		return nil, err
```
austinbyers, line 69: Instead of returning an error and blocking the whole deploy here, let's just log a warning and then set a default:

```go
if err != nil {
    logger.Warnf("git describe --tags failed, Panther version unknown: %v", err)
    gitVersion = "(unknown)"
}
```

austinbyers: Actually, `gitVersion` should be a package variable that is already set for you (it happens at the very beginning of the deploy process)

3nvi:
> Awesome! I like how you comment on your own PR to add context that doesn't necessarily make sense to be in a code comment. I've started doing that as well

Yeah I've been doing that for a  while in PRs where the solution might seem weird, to explain my though process or the reason why.

tools/mage/deploy_frontend.go
3nvi: I find it so weird that variables get hoisted like that (without an explicit import). I mean how can you track what's already defined? Is there a list of "inherited vars"?

web/scripts/utils.js
```diff
@@ -27,31 +26,22 @@ function loadDotEnvVars(path) {
...
+    WEB_APPLICATION_GRAPHQL_API_ENDPOINT: process.env.WEB_APPLICATION_GRAPHQL_API_ENDPOINT,
+    WEB_APPLICATION_USER_POOL_CLIENT_ID: process.env.WEB_APPLICATION_USER_POOL_CLIENT_ID,
+    WEB_APPLICATION_USER_POOL_ID: process.env.WEB_APPLICATION_USER_POOL_ID,
+  };
 
-  const unsetVars = requiredEnvs.filter(env => process.env[env] === undefined);
-  if (unsetVars.length) {
-    throw new Error(chalk.red(`Couldn't find the following ENV vars: ${unsetVars.join(', ')}`));
+  const keys = Object.keys(PANTHER_CONFIG).filter(key => PANTHER_CONFIG[key] === undefined);
+  if (keys.length) {
```
alexmylonas, line 40: Change `keys` to `missingKeys`!

web/webpack.config.js
alexmylonas: Do we really need both `isEnvProduction` and `isEnvDevelopment`? What's the use case or having a NODE_ENV not `production` and not `development`

web/src/helpers/utils.tsx
```diff
@@ -237,7 +237,7 @@ export const getStableVersion = (version: string) =>
   version.indexOf('-') > 0 ? version.substring(0, version.indexOf('-')) : version;
 
 export const generateDocUrl = (baseUrl: string, version: string) => {
-  if (version) {
+  if (version && version !== 'custom') {
```
alexmylonas, line 240: More question than a comment, can some1 have a version so that is divert that master docs makes more sense than their base version?

web/scripts/utils.js
3nvi: 👍

web/webpack.config.js
3nvi: it's for two reasons:

* it's easier to read `isEnvProduction && ...` rather than `!isEnvDevelopment && ...` 
* we have the ability  to add `NODE_ENV=staging` at a given time (or `NODE_ENV=test`), although that's pretty rare

web/src/helpers/utils.tsx
3nvi: @alexmylonas can you rephrase that? I didn't understand

tools/mage/deploy_frontend.go
austinbyers: Every folder in our source tree is its own Go package. A package is a namespace - you could pretend all of `tools/mage/*.go` is in a single file, which means all global variables, functions, etc are shared with the entire package automatically.

We always define package constants and variables at the top of source files, but these could be spread out over multiple files within the same package.

In this case, most of the package-wide variables relevant to the deploy process are defined at the top of `tools/mage/deploy.go`

web/src/helpers/utils.tsx
austinbyers: Not sure if we're asking the same question, but I know we link to the version of documentation relevant to each deployment. If there is no version available, will that link be broken? Or do we redirect to the latest stable documentation? We should do the latter

tools/mage/deploy_frontend.go
3nvi: I understand, but I still find it weird having variables available to you without an explicit import

Issue #849, opened 2020-05-06
Pre-built Python layer generation


austinbyers:
### Description

We currently build the default Python layer locally (with `boto3`, `policyuniverse`, and `requests`) during `mage deploy`. To be able to have a self-contained deployment, the layer needs to support pre-packaging and needs to be small enough to fit in SAR's 54MB package limit.

### Acceptance Criteria

- Python layer is built in a way that is compatible with SAR. In particular, its total size is < 54 MB

Issue #850, opened 2020-05-06
Custom cfn resources: post-deploy setup


austinbyers:
### Description

We make a number of API calls manually after a deployment that need to be moved into custom CFN resources to easily package the entire Panther app in SAR.

### Acceptance Criteria

Some combination of custom resources for all of the post-deploy process:

- Inviting the first admin user
- Bulk uploading the initial analysis pack
- Requiring TOTP for Cognito's 2FA
- Self-onboarding

PR #863, opened 2020-05-08
Custom resource for Cognito MFA, use sam packaging

austinbyers:
## Background

Changes to the bootstrap and packaging process to bring us closer to SAR. With this change, the `bootstrap-gateway` stack can now be published in SAR, and the `bootstrap` stack won't be too far behind.

Closes: #849
Part of #850 

## Changes

- Move Cognito MFA configuration from `mage deploy` into a `Custom::CognitoUserPoolMfa` custom CFN resource

Another big advantage of moving these setup API calls from `mage` into custom CFN resources is that they run in the same AWS region as the resources they need access to, which should be lower latency than running from developer/user laptops

- Use `sam package` to package templates before uploading
    - `aws-sam-cli` has been added to the python requirements

We used to use "aws cloudformation package", then #742 implemented our own packaging because of some limitations with the aws cli. But since we need Panther to be compatible with SAR, I'd rather have the developer workflow match the SAR publishing workflow.

Thankfully, "sam package" supports more resources than the AWS CLI and does not have the `us-east-1` region error we encountered before. The only downside (?) is that it outputs progress tickers to stderr (not stdout), so I can't easily suppress them with mage:

```
15:30:45	INFO	    √ panther-bootstrap-gateway finished (2/12)
Uploading to 51fdb42e2ebee248f96eb563ec99dfdc.template  6261 / 6261.0  (100.00%)
Uploading to 9fc4c45fbaf1abde3d0f1e9ffc6449bc  16459 / 16459.0  (100.00%)
15:30:46	INFO	deploy: docker build web server (deployments/Dockerfile)
Uploading to d2077873bd708758a06e60b40e9f6ae3.template  10278 / 10278.0  (100.00%)
15:30:47	INFO	deploy: create CloudFormation stack panther-appsync
Uploading to b65d2ed33d1044c2b540f2dac358aa2e.template  92839 / 92839.0  (100.00%)
15:30:49	INFO	deploy: create CloudFormation stack panther-cw-dashboards
Uploading to 2a6d330b5fcde1a9320cf4ad1a4f6656.template  78264 / 78264.0  (100.00%)
Uploading to a10aa107033f9bd9bc42f75e5359ffe9.template  49349 / 49349.0  (100.00%)
Uploading to b555489e39ae5e87c73b7e56299b4a65.template  771 / 771.0  (100.00%)
Uploading to 3bd90fa6fb6e50a3d411c3d793ba950b  524288 / 4811637.0  (10.90%)15:30:51	INFO	deploy: create CloudFormation stack panther-cw-alarms
```

Not a big deal, and might actually be helpful for those with a slow internet connection to see how quickly their S3 assets are being published. I'm sure we can figure out a way to suppress these later if needed

- Removed `boto3` from the default Python layer so that the layer fits within SAR's ~54MB package size limit (the layer was 57 MB with boto3!)

We used to pin our own newer version of boto3 because at the time I believe we needed a remediation that was not available in the default Lambda version of boto3. Today, the current [lambda default version](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html) (1.12.22) is from March 16, which should easily be recent enough for us.

Removing this from the default layer will speed up deployment and Python Lambda initialization as well since it's ~50MB less data to upload/download from S3

I've pinned `boto3` in the requirements to match the version in Lambda, which should ensure unit tests will fail if we're using an unsupported function at which point we can re-evaluate how to proceed.

## Testing

- Deployed to `us-west-2` on top of `master`
- Deployed to `us-east-1` from scratch (to ensure packaging still worked in a formerly problematic region)
- `sam publish -t out/package.embedded.bootstrap_gateway.yml` to confirm the `bootstrap-gateway` stack can be published to SAR


deployments/bootstrap_gateway.yml
```diff
@@ -16,12 +16,21 @@
 
 AWSTemplateFormatVersion: 2010-09-09
 Transform: AWS::Serverless-2016-10-31
-Description: API gateway and Python layer
+Description: Bootstrap resources that require the source S3 bucket to exist first
+
+Metadata:
```
austinbyers, line 21: This metadata is used by "sam publish" to setup the SAR app. We haven't published anything publicly yet, but when we do, this metadata will be attached to it

We'll add this metadata to every stack as I update each one for SAR compatibility

internal/core/custom_resources/main/lambda.go
```diff
@@ -41,7 +41,7 @@ func customResourceHandler(ctx context.Context, event cfn.Event) (string, map[st
 
 	handler, ok := resources.CustomResources[event.ResourceType]
 	if !ok {
-		return "", nil, fmt.Errorf("unsupported resource type: %s", event.ResourceType)
+		return "-", nil, fmt.Errorf("unsupported resource type: %s", event.ResourceType)
```
austinbyers, line 44: It turns out the `physicalResourceId` returned by a custom resource should *always* be set, even if you're returning an error.

If it's blank, CloudFormation will report "invalid resourceID" instead of the actual error message

tools/mage/build_namespace.go
```diff
@@ -194,6 +196,21 @@ func buildLambdaPackage(pkg string) error {
 		return fmt.Errorf("go build %s failed: %v", binary, err)
 	}
 
+	if statErr == nil && hashErr == nil {
```
austinbyers, line 199: Restored from #742 with a small adjustment to the comment (now that I better understand why this happens)

tools/mage/deploy.go
```diff
@@ -226,38 +222,20 @@ func deployBoostrapStacks(
 		return nil, err
 	}
 
-	// Enable only software MFA for the Cognito user pool - enabling MFA via CloudFormation
```
austinbyers, line 229: We will continue to remove functionality from `mage deploy` until all custom logic can run from within a SAR package / CFN template

austinbyers:
No changes are required for existing deployments, but developers will note that you will need a fresh `mage setup` to install the new Python libraries

deployments/panther_config.yml
```diff
@@ -30,15 +30,18 @@ Infra:
   # the larger sizes may be required for adequate performance or large files.
   LogProcessorLambdaMemorySize: 1024 # 256 - 3008, in 64MB increments
 
-  # Create a Python layer with these pip library versions.
+  # Create a Python layer with these pip library versions for analysis and remediation.
```
nhakmiller, line 33: Should we add a warning about the max compressed size here?

austinbyers: The 54MB limit only applies to us when SAR packaging a release, users can put as many libraries as they want here.

Still, I suppose we can add a warning to our future selves

nhakmiller: Ah I misunderstood. Although I suppose we could add a warning that the max layer + function size can't exceed 250MB, but that's not very useful without knowing the function size (which is ~6MB) and also not related to this change. Up to you!

internal/core/custom_resources/main/lambda.go
rleighton: Good to know!

PR #869, opened 2020-05-12
[parsers] Parser for Apache HTTP server access logs

### LINKED ISSUE #867: Support Apache Access Logs
### Description

Support Apache Access Logs

### Acceptance Criteria

- Parsers are in place
- Unit tests for parsers
- Log type selectable in UI
- Glue tables created

alxarch:
## Background

Resolves #867 

## Changes

- Adds and registers parsers for Apache `common` log format
- Adds and registers parsers for Apache `combined` log format

## Testing

- mage test:ci
- unit tests for samples found in the docs
- TODO: get more samples for unit tests from apache instances


internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
```diff
@@ -0,0 +1,131 @@
...
+	AccessCommonLog
+	UserAgent *string `json:"user_agent,omitempty" description:"The User-Agent HTTP header"`
+	Referer   *string `json:"referer,omitempty" description:"The Referer HTTP header"`
+}
+
+type AccessCombined struct {
+	AccessCombinedLog
+
+	parsers.PantherLog
+}
```
kostaspap, line 49: Why do we need both? Seems simpler to just have one of them, will save a few lines.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
```diff
@@ -0,0 +1,131 @@
...
+		common, ref, ua := row[:numFieldsAccessCommon], row[fieldIndexReferer], row[fieldIndexUserAgent]
+		if err := log.AccessCommonLog.SetRow(common); err != nil {
+			return err
+		}
+		log.Referer = nonEmptyLogField(stripQuotes(ref))
+		log.UserAgent = nonEmptyLogField(stripQuotes(ua))
+		return nil
+	}
+	return errors.Errorf("invalid number of fields %d", len(row))
+}
```
kostaspap, line 71: The way I see it, the Parser's `Parse` method accepts a string, applies all the necessary logic to extract useful fields out of it and returns a -dummy- struct with appropriate fields populated. 

This code takes a different approach. It seems that the responsibility for parsing a string lays on the output struct(AccessCombinedLog) itself and the Parser is responsible for just invoking the appropriate method on the struct. 
If we follow this logic though, the existence of the "Parser" struct itself doesn't make sense - the output struct (AccessCombinedLog) can implement the Parser interface. But I am not sure I like it that much. 

My thought is to keep the `AccessCombinedLog` simple and keep the logic for field extraction in the `AccessCombinedParser`, similar to how we do in the rest of the parsers. Same applies for the rest of the parsers in this PR.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
```diff
@@ -0,0 +1,131 @@
...
+
+type AccessCombinedParser struct{}
+
+var _ parsers.LogParser = (*AccessCombinedParser)(nil)
+
+func NewAccessCombinedParser() parsers.LogParser {
+	return &AccessCombinedParser{}
+}
```
kostaspap, line 79: Not really needed - `New()` function is sufficient.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
```diff
@@ -0,0 +1,131 @@
...
+	rxQuoted,     // userAgent
+))
+
+const (
+	rxUnquoted   = `[^\s]+`
+	rxBrackets   = `\[[^\]]+\]`
+	rxQuoted     = `"[^"]+"`
+	rxStatusCode = `\d{3}`
+	rxSize       = `-|\d+`
+)
```
kostaspap, line 115: We usually prefer to put constants/vars at the top of the file.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
```diff
@@ -0,0 +1,131 @@
...
+	rxUnquoted,   // remoteIP
+	rxUnquoted,   // clientID,
+	rxUnquoted,   // userID,
+	rxBrackets,   // requestTime
+	rxQuoted,     // requestLine
+	rxStatusCode, // responseStatus
+	rxSize,       // responseSize
+	rxQuoted,     // referer
+	rxQuoted,     // userAgent
+))
```
kostaspap, line 107: We have a `StreamingCSVReader` that you can use to parse CSV logs.  This is how we do it in e.g. ALB parser: https://github.com/panther-labs/panther/blob/7e97d75cab0fdbef7cabe729cb6e731a66ddaccf/internal/log_analysis/log_processor/parsers/awslogs/alb.go#L82-L89

We also have utility methods like `CsvStringToPointer` that you can could use, I think it can simplify the logic a bit. Any specific reason you picked regexes? In general, I'd keep regex as a ... last resort due to potential of something going terribly wrong performance-wise. This code will be touched by make devs in the future with varying levels of regex familiarity.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
```diff
@@ -0,0 +1,205 @@
...
+// zone = (`+' | `-') 4*digit
+const layoutApacheTimestamp = `[02/Jan/2006:15:04:05 -0700]`
+
+func parseAccessLogTimestamp(s string) (time.Time, error) {
+	tm, err := time.Parse(layoutApacheTimestamp, s)
+	if err != nil {
+		return time.Time{}, err
+	}
+	return tm.UTC(), nil
+}
```
kostaspap, line 172: We have the `timestamp.Parse(layoutApacheTimestamp, s)` that does the same operation.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
alxarch: We discussed this on slack and I thought the conclusion was to support both (they are both quite common). The `combined` is used for analytics (ie user browsers) and the `common` is the default.

kostaspap: Sorry I meant that instead of:
```
type AccessCombinedLog struct {
	AccessCommonLog
	UserAgent *string `json:"user_agent,omitempty" description:"The User-Agent HTTP header"`
	Referer   *string `json:"referer,omitempty" description:"The Referer HTTP header"`
}

type AccessCombined struct {
	AccessCombinedLog

	parsers.PantherLog
}
```

we could have
```
type AccessCombined struct {
	AccessCommonLog
	UserAgent *string `json:"user_agent,omitempty" description:"The User-Agent HTTP header"`
	Referer   *string `json:"referer,omitempty" description:"The Referer HTTP header"`
	parsers.PantherLog
}
```

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
alxarch: The parser handles the assignment of panther fields  and initialization of PantherLog struct.
The ParseString was put as a method so it returns a single error and Parse would not have multiple return points  that obfuscate it's debugging. I'm OK if you want to move the ParseString method  to the parser but then it would need an extra argument or return a pointer.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
alxarch: It's to be streamlined with the upcoming changes and to avoid the ugly `(&apachelogs.Parser{}).New()` wherever we need to use a new parser. The call is inlined by the compiler anyway.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
alxarch: The grouping at the bottom has the purpose of having all regexp related code close together so it is easy to reason about what ` buildRx` does. I can move the `buildRx` in a new file to make it more apparent since it is used by both logs.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
alxarch: CSV is not working in this case because the timestamp field is grouped by square brackets `[`, `]` and has whitespace inside. It would be much harder to understand what the code does (also requires more allocations) if we had to concat fields after parsing to re-create the timestamp string value. Was my first approach to this and it didn't work.

alxarch: Regexes are quite efficient if they are compiled before they are used (ie MustCompile). Most of the time the result is close to hand written state machine code. They also have the benefit of being somewhat composable as the `buildRx` function shows.  The performance problems with regexes start when compile them at runtime or use a lot of nested groups. For extra performance the `*Index` variances of the methods can be used to avoid allocation on subroups return slice but this is a mute point when we use `*string` all over the place which creates new allocations for every small string any way.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
alxarch: Will use that

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
alxarch: As for the devs issue that was the reason I opted for the `buildRx` approach instead of using a long obfuscated regexp. I also get confused with regexes all the time but breaking them up in parts reduces cognitive load to almost zero. :)

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
alxarch: It was to avoid duplication of field definitions and to reuse the `AccessCommonLog.SetRow()` method without duplication of logic.

alxarch: The second one can be changed, indeed. It's just for keeping the structure similar in both cases (ie setting the ParseString and SetRow methods at the same level in both structs).

kostaspap: I had similar comment for the `AccessCommonLog` and `AccessCommon`. It seems that one of them should be enough. We can embed the PantherLog directly in the `AccessCommonLog` and get rid of the `AccessCommon`.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
kostaspap: Can't we just inline the calls in the Parse method? Seems to be simpler and more straightforward. My bigger problem while reading the code is that I had to scroll up and down several times trying to follow the code logic (could be an issue of positioning of the functions though)

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
kostaspap: Alright.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
```diff
@@ -0,0 +1,196 @@
...
+	rxBrackets,   // requestTime
+	rxQuoted,     // requestLine
+	rxStatusCode, // responseStatus
+	rxSize,       // responseSize
+))
+
+func (log *AccessCommonLog) ParseString(s string) error {
+	match := rxAccessCommon.FindStringSubmatch(s)
+	if len(match) > 1 {
+		return log.SetRow(match[1:])
```
kostaspap, line 93: Not sure, why `log.SetRow(match[1:])` and not `log.SetRow(match)`?

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
```diff
@@ -0,0 +1,196 @@
...
+		remoteIP, idRFC1413, userID, requestTime, requestLine, responseStatus, responseSize := row[0], row[1], row[2], row[3], row[4], row[5], row[6]
+		tm, err := timestamp.Parse(layoutApacheTimestamp, requestTime)
+		if err != nil {
+			return err
+		}
+		req, err := parseRequestLine(requestLine)
+		if err != nil {
+			return err
+		}
+		n, err := strconv.ParseInt(responseStatus, 10, 16)
```
kostaspap, line 113: Better use https://github.com/panther-labs/panther/blob/89b6dd7cfdfc97afa9007ecac1738057b3f37841/internal/log_analysis/log_processor/parsers/csv_utils.go#L46-L55

It can deal with `-` and already returns a pointer.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
```diff
@@ -0,0 +1,196 @@
...
+		if err != nil {
+			return err
+		}
+		statusCode := int16(n)
+		numBytes, err := strconv.ParseInt(responseSize, 10, 64)
+		if err != nil {
+			return err
+		}
```
kostaspap, line 121: You can use https://github.com/panther-labs/panther/blob/89b6dd7cfdfc97afa9007ecac1738057b3f37841/internal/log_analysis/log_processor/parsers/csv_utils.go#L34-L44 . 
It deals with the `-` character that appears in such logs when a field is not populated. You can introduce an Int64 version if you wish.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
```diff
@@ -0,0 +1,196 @@
...
+
+// 	[day/month/year:hour:minute:second zone]
+// day = 2*digit
+// month = 3*letter
+// year = 4*digit
+// hour = 2*digit
+// minute = 2*digit
+// second = 2*digit
+// zone = (`+' | `-') 4*digit
+const layoutApacheTimestamp = `[02/Jan/2006:15:04:05 -0700]`
```
kostaspap, line 163: Since it's used in both parsers, you can define this as a constant in a `apachelogs.go` file.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
```diff
@@ -0,0 +1,196 @@
...
+
+type requestLine struct {
+	Method   string
+	URI      string
+	Protocol string
+}
+
+var rxSplitSpace = regexp.MustCompile(`\s+`)
+
+func stripQuotes(line string) string {
```
kostaspap, line 173: Looks like `strings.Trim(.., "\"")` ?

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
alxarch: regexp fist match is the whole string match.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
alxarch: Same as above

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
alxarch: `strings.Trim` trims multiple characters off both sides so a quote in the string would be chopeed off. This helper function has a clear name for the op and gets inlined by the compiler due to it's simplicity.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
alxarch: They are separate functions in order to reuse the logic of common for combined. Scrolling is more of a github review UI issue than a code issue. I can duplicate all the code if you want but I fail to see the reasoning.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
kostaspap: Ok, some good points raised! 
Putting my thoughts here: 

1. AWS Dependency: We could get rid of the `aws` dependency in the csv_utils.go package. From a fresh perspective, it seems like an overkill. The truth is, we use this utility pretty extensively throughout the code because it is ... handy. We can get rid of the dependency in the code in csv_utils. 
2.  Error handling: I realize the issue, it's true that the csv_utils is not handling conversions properly. It should be either panicking (Must...) or return an error. Seems like something we should fix (not in this PR, we should probably do in separate)
3. The `CsvStringToPointer` has the same functionality as `nonEmptyLogField`. It is used 56 times overall by aws/alb, aws/vpcflow, ngxinx logs... It is very usual that an empty value in CSV (TSV/...) logs is represented by `-`. So we need a method like this to be available in more parsers. 
4. We started with just the above method, `CsvStringToPointer`. Then we encountered cases in other logs where  number could be `nil` (represented by `-` in the logs) e.g. TargetPort in ALB logs. So the CSV Utils got extended with more utilities. 

The Apache log is another log type that seems to have similar characteristics (e.g. `-` signifying `nil`).  I am also in general looking to minimize inter-package and inter-file dependencies even, but in this cases it seems that some of the functionality required by the Apache parsers was already developed for other parsers.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
kostaspap: The comment about scrolling up/down is not specific to Github reviewing, I think it is a general readability issue. I was making a reference to the "stepdown rule" for writing code.

kostaspap: I was trying to see how we could improve readability. We should not duplicate code, though, I agree on that. If it's too complicated you can leave out.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
kostaspap: Alright, didn't realize that we could have more than one `"` in the prefix/suffix. Sounds good then.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
alxarch: It's a very interesting topic :) Although `nonEmptyField` *seems* to be doing the same but `nonEmptyLogField` is *inlined* by the compiler avoiding the function call overhead (death by a 1000 cuts) and it can be composed with other inlineable helpers (such as `stripQuotes`). The csvutil functions are calling non-inlineable function from `strconv` to convert the values which results in an actual function call. Also the fact that they do 2 things (check for empty and convert) leads to them having long names to explain their functionality. The fact that other logs use the `-` convention for empty fields is more of a coincidence they could be using `N/A` or `unknown` for example. Repeating the helper in the log parser package itself allows for modifying the specific helper in the context where it is used. The actual code is just a switch statement packaged as a function so reviewing it is not time consuming. I'm not suggesting changing function names or how the functions currently work since they are reviewed and happily running for some time as they are (although to be safe we should have a PR switching all `strconv.Atoi` to safer code). 
My suggestion is that going forward this kind of no-op helpers that have logic defined by the requirements of the parser to be close to the point that they are used to avoid inter-package dependencies of logic that is trivial. OTOH helpers that could be useful to all parsers should exist at the `parsers` module for example that is already a must have dependency for all parsers and they should be very clear about their purpose (like the strings package for example). As for the `aws.X` helpers It's ok to keep using them in test code since it's not included at the production build but avoid them if we can in parser code  especially when the to-be-pointed-at variable is right there on the stack, only a `&value` away.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
alxarch: Changed

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
kostaspap: I agree with this approach in general. And thanks for the interesting discussion!
Some thoughts on the above: 
* The parsers have largely dependencies only in the parsers package (`csv_utils.go` is inside the parsers package) and I agree that we should try to keep that. The `aws.String`, `aws.Int` etc dependency is something we discussed in the past and decided to make an exception since it tends to be pretty ... handy. But whenever it's not required, we can skip using it indeed. Feel free to fix this in this PR, or we can do it at a subsequent PR. 
* The CSV functions having the "Pointer" suffix to make clear the return type, but they didn't have to do that indeed. They could return the values instead of pointers... In the context of the parsers however as see it appropriate. It would save from repetition. 
* It's fair to say that we should define such utility functions as close to the caller as possible. It happens pretty frequently that parsers of the same family (e.g. AWS) have shared characteristics e.g. the share the same timestamp format or they define "nil" with the same character. For such a scenario we could/should share some code in a `xyzlogs/xyzlogs.go` file like you did in GCP.Audit logs. For Apache, it seems fair to keep the function you already defined though, I see your point.

internal/log_analysis/log_processor/parsers/apachelogs/access_combined.go
alxarch: Moved

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
alxarch: Moved

Issue #848, opened 2020-05-06
Custom cfn resource: alarms


austinbyers:
### Description

SAR does not supported nested cfn stacks, which is how our CloudWatch alarms are currently implemented. We need to generate alarms in a SAR-compatible way.

We could publish deeply nested SAR apps for each nested alarm stack, but a cleaner solution (if it's not too much work) is to create a custom resource for each group of alarms. Then alarms can be connected directly to their associated resource. For example,

```yaml
Resources:
  MyLambdaFunction:
    Type: AWS::Serverless::Function
    # ...
     
  # Creates alarms for memory, timeout, errored invocations, logged errors, logged warnings
  MyLambdaFunctionAlarms:
    Type: Custom::LambdaAlarmGroup
    Properties:
      FunctionName: !Ref MyLambdaFunction
      # Optional
      LogWarnThreshold: 5
      LogErrorThreshold: 2
```

austinbyers:
Closed by #937

PR #872, opened 2020-05-12
Custom resource: metric filters

austinbyers:
## Background

Metric filters are currently their own stack autogenerated by `build:cfn`. This complicates the SAR packaging - since the alarms will move into a custom resource, why not the metric filters as well?

Besides simplifying packaged deployments, building metric filters next to its associated log group makes it easier to add future customization and makes deploys faster! We have to wait until the very end to deploy the `metric-filters` stack today because all other log groups have to exist first. But now, metric filters can be applied to each log group immediately after it is created.

Related to #848 

## Changes

- Add a `Custom::LamdaMetricFilters` resource which adds memory, warning, and error metric filters with the same exact configuration we have today
    - This is our first "full" custom resource that needs create, update, and delete implementations and has parameters. As such, I've added parsing/validation of parameters into structs for the custom resource handlers
- Remove `metric-filters` stack
- Bugfix: custom resource deletion no longer fails when rolling back a failed create
- Bugfix: if a deploy has to wait on a stack in progress, show a message to the user immediately

## Testing

- Fresh deploy + teardown
- Deploy this PR on top of master
- Added a check to `test:ci` that confirms all lambda functions have an associated log group and metric filter in the same template with the same logicalID prefix
- Verified all lambda log groups have 3 metric filters after deployment
- Verified metric filters have the exact same configuration as before:

Screen Shot 2020-05-12 at 11 12 00 AM

- Tested custom resource update by changing runtime (update in place) and/or log group (create + destroy)


internal/core/custom_resources/main/lambda.go
```diff
@@ -41,10 +41,18 @@ func customResourceHandler(ctx context.Context, event cfn.Event) (string, map[st
 
 	handler, ok := resources.CustomResources[event.ResourceType]
 	if !ok {
-		return "-", nil, fmt.Errorf("unsupported resource type: %s", event.ResourceType)
+		return "error", nil, fmt.Errorf("unsupported resource type: %s", event.ResourceType)
```
kostaspap, line 44: nit: You can also just define a constant "DefaultPhysicalID" or similar in the package and reference it here and line 52.

internal/core/custom_resources/resources/parser.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return fmt.Errorf("parameter json marshal failed: %v", err)
+	}
+
+	if err = jsoniter.Unmarshal(json, out); err != nil {
+		return fmt.Errorf("parameter validation failed: %v", err)
+	}
+
+	if err = inputValidator.Struct(out); err != nil {
+		return fmt.Errorf("parameter validation failed: %v", err)
```
kostaspap, line 52: nit: You might want to clean the `out` in case the validation has failed.

internal/core/custom_resources/resources/parser.go
```diff
@@ -0,0 +1,56 @@
...
+		return nil
+	}
+
+	// We could manually use reflection on the output struct, which would be marginally faster
+	// but unnecessarily complex and error-prone.
+	//
+	// Instead, we'll just marshal the map as json and then unmarshal into the struct
+	json, err := jsoniter.Marshal(&params)
+	if err != nil {
+		return fmt.Errorf("parameter json marshal failed: %v", err)
```
rleighton, line 44: I would add a `%#v` to the error message with `params` to make debugging easier.

tools/mage/deploy.go
```diff
@@ -146,7 +146,13 @@ func Deploy() {
 		logger.Fatal(err)
 	}
 
-	logger.Infof("deploy: finished successfully in %s", time.Since(start))
+	// ***** Step 4: migrations
+	// Starting in v1.3.0, the metric-filter stack is no longer used. It can be safely deleted
+	if err := deleteStack(cloudformation.New(awsSession), aws.String(metricFilterStack)); err != nil {
+		logger.Warnf("failed to delete deprecated %s stack: %v", metricFilterStack, err)
```
rleighton, line 152: This will only trigger on failure, not if the stack does not exist, right?

internal/core/custom_resources/resources/lambda_metric_filters.go
```diff
@@ -0,0 +1,188 @@
...
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to put %s metric filter: %v", metricName, err)
+	}
+	return nil
+}
+
+func deleteMetricFilterGroup(physicalID string) error {
+	split := strings.Split(physicalID, ":")
```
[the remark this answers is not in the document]
austinbyers: Good idea, we have that comment in a different file but not this one, will add

internal/core/custom_resources/main/lambda.go
austinbyers: Sure, but it's only two instances within the same function, constant seems overkill to me at that point

internal/core/custom_resources/resources/parser.go
austinbyers: Not sure what you mean by this? `out` is a pointer to a struct which is a different instance on every lambda invocation

internal/core/custom_resources/resources/parser.go
austinbyers: At the very beginning of the Lambda function, we always log the full event. This error message will show up directly in CloudTrail, it can't be too big and we want it to be user friendly.

In any case, this particular error is nearly impossible to trigger - you'd have to be passing a very strange struct for json marshaling to fail

tools/mage/deploy.go
austinbyers: Correct - the `deleteStack` helper function returns `nil` if the stack does not exist

internal/core/custom_resources/main/lambda.go
kostaspap: The reason I said this: I read this code and started writing a comment in the likes of "Why do we need this to be 
'error' and not empty string?". 
Then I continued reading and saw that there was same value (with a comment explaining why it was needed). I thought a local constant with a comment on it would clarify or you can also move the comment up. Up to you, it's nitpicking at this point.

austinbyers: Sure, yeah I was going to add a local constant

internal/core/custom_resources/resources/parser.go
kostaspap: You are right. You can ignore this comment.

internal/core/custom_resources/main/lambda.go
austinbyers: ah, actually I forgot this is the separate `main` package, the constant is fine. Good idea! thanks

PR #879, opened 2020-05-13
Custom resource: ApiGatewayAlarms

austinbyers:
## Background

We currently implement alarms as nested stacks, which is not compatible with SAR. We could transform these nested stacks into nested SAR apps, but it's simpler to just use a custom resource for each alarm group that we care about. This also has the benefit of keeping monitoring next to the infra itself and makes it much easier to configure alarm thresholds.

This PR is the first such alarm group to make sure we like this approach before I convert all alarms. The idea is that each resource type (SNS, API Gateway, Lambda, Dynamo, etc) gets a custom resource which defines all alarms for a resource of that type.

A potential disadvantage is that this adds dozens of new resources to existing stacks, which brings us closer to the 200 resource limit. But it's much better than adding alarms individually and our biggest stack so far (by resource count) is cloud-security, with only 74 resources so far. Adding an alarm group for every function, queue, and table in that stack would add 25 resources, meaning we'd still only be at half the limit. We're also well below the template size limit

Part of #848 

### Alternative Design
Instead of adding one alarm group per resource, we could add one alarm group per resource *type* in each stack:

```yml
ApiGatewayAlarms:
    Type: Custom::ApiGatewayAlarms
    Properties:
        ApiNames:
            - panther-analysis-api
            - panther-compliance-api
            # ...
```

This reduces the total number of resources, but makes the update logic in the custom-resource handler more complex and makes it harder to configure each alarm individually.

For these reasons, I prefer the one-alarm-group-per-resource design laid out here

## Changes

- Add `Custom::ApiGatewayAlarms` custom resource
    - I wanted `Custom::Alarms::ApiGateway`, but you can't add extra colons
    - Fixed a typo in the IntegrationLatency alarm name
    - Changed the Error alarm unit from "None" to "Count" (confirmed in the [spec](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-metrics-and-dimensions.html))
- Set the maxRetries in AWS config to 10 - metric filter deletes were getting throttled during a teardown
- Move alarm topic ARN to `bootstrap` stack and renamed it to avoid conflicts when upgrading an existing deployment. Since we're creating alarms as we go, we have to create the alarm topic at the very beginning of the deploy instead of at the end

## Testing

- Added a CI check which enforces all API gateways have an associated alarm group in the same stack
- Fresh deploy + teardown
- Upgrade from `master`
- Change all alarm properties to force updates in place and update/destroy for new resources

### Migration
The `panther-alarms` topic will be deleted and replaced with `panther-cw-alarms`. If you specify a topic in the config file, this change has no effect. If users want to keep their existing `panther-alarms` topic, they can add `DeletionPolicy: Retain` in the alarms stack, deploy, then apply this change with the old topic in the config file.

But most users (ourselves included) are already using the config file to route alarms in production, so this change shouldn't cause much trouble


austinbyers:
> You are going to have some "checker" so we can't forget to add alarms, correct?

Correct, that's already included! I'm adding them as I go to make sure I don't miss anything, which will also then enforce them in the future

> I would encourage you as you add each new alarm to set the threshold very low, deploy and confirm it works, then reset the threshold before checkin. It is very easy to misconfigure an alarm such that it never fires.

Good idea, I'll make sure I can trigger the alarms before merging. I've also been comparing the settings side-by-side to make sure they are identical before and after

austinbyers:
Testing revealed that these alarms have actually never worked! We were using `Name` as the metric dimension, but it should have been `ApiName`. Fixed!

PR #887, opened 2020-05-14
Custom resources: ELB and AppSync alarms

austinbyers:
## Background

Creating custom alarm groups for ELB and AppSync. We only have one load balancer and one AppSync API, so we could have just put the alarms directly in the CloudFormation instead of having a custom resource.

I'm open to that idea if we find it preferable, but alarms are very verbose to define, it's certainly easier to build them in Go so we can share common fields between each alarm creation.

Part of #848 

## Changes

- Bugfix: Don't error when deleting a resource type that doesn't exist
    - This can happen when rolling back a new resource type. Without this fix, CloudFormation spends several minutes trying to repeatedly delete a custom resource type which no longer exists
- Add some shared alarm create/delete methods to simplify the handlers of each type
- Added "Count" as the unit for the applicable alarms
- Add `Custom::AppSyncAlarms` and `Custom::ElbAlarms`
    - Remove corresponding alarm generation from `mage`

## Testing

- Deployed to dev account

These alarms are difficult to trigger, but I double-checked that the config is (a) the same as before and (b) matches the alarm documentation for the two services.

Both the load balancer and the appsync API are defined in the `bootstrap` stack, but their alarms can't be added until later (after the `custom-resources` function exists in `bootstrap-gateway`). For this reason, I didn't add the CI enforcement of these alarm groups, which should be fine because we only have one of each anyway


austinbyers:
It always makes me happy when the code added is the same or less than the code deleted :)

internal/core/custom_resources/resources/alarms.go
```diff
@@ -0,0 +1,113 @@
...
+
+	if input.TreatMissingData == nil {
+		input.TreatMissingData = aws.String("notBreaching")
+	}
+
+	if input.EvaluationPeriods == nil {
+		var periods int64 = 1
+		if aws.StringValue(input.Unit) != cloudwatch.StandardUnitCount {
+			periods = 5
+		}
```
kostaspap, line 61: I'm not sure I understand this logic... If the metric unit is not `Count`, we increase the evaluation periods? What for? 
Same comment for the statements in lines 67, 75

rleighton: Yeah, some comments around this. Might be scary for someone later to change

austinbyers: This is just following the standard we apparently have for all of our alarms - if we care about "Count", we always do "more than X in a single 5-minute window"

If we care about latency or anything else, it's always "more than X 5 times in a row in a 1-minute window".

But since it's confusing I'll just remove the default and we'll again specify it for every alarm individually

PR #888, opened 2020-05-14
[deployment] bucket hardening

nhakmiller:
## Background

Enforce secure access to s3 buckets via conditions on the s3 access policy.

## Changes

- Update the s3 access policy of all buckets to block insecure access

## Testing

- Deployed to dev account with self onboarding enabled, no errors appeared in cloudwatch so I don't think it caused any access issues


deployments/bootstrap.yml
```diff
@@ -227,13 +227,21 @@ Resources:
...
             Effect: Allow
             Principal:
               Service: guardduty.amazonaws.com
             Action: s3:PutObject
             Resource: !Sub arn:aws:s3:::${AuditLogs}/*
+          - Sid: ForceSSL
+            Effect: Deny
+            Principal: '*'
+            Action: s3:GetObject
+            Resource: !Sub arn:aws:s3:::${AuditLogs}/*
```
austinbyers, line 241: Not sure if we should be specifying the partition:

```
Resource: !Sub arn:${AWS::Partition}:s3:::${AuditLogs}/*
```

PR #890, opened 2020-05-15
Add simpleDB key-value store

austinbyers:
## Background

@jacknagz asked for a simple key-value DB which Python rules and policies could read from and write to.

## Changes

- Add a DynamoDB key-value store
- Allow the two Python engines to read from and write to the store

## Testing

- Deployed
- `mage test:ci`


deployments/bootstrap.yml
```diff
@@ -369,6 +369,12 @@ Resources:
       VersioningConfiguration:
         Status: Enabled
 
+  ########## Storage ##########
+  PythonKeyValueDB:
+    Type: AWS::SDB::Domain
+    Properties:
+      Description: Simple key/value store for Python policies and rules to use however they like
```
austinbyers, line 376: There literally aren't [any other properties](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-simpledb.html) here

kostaspap:
SimpleDB was one of the first AWS services, it was built before DynamoDB. It is considered a legacy service, it has stopped getting new features and serious support years ago. Not clear to me why we would prefer to expose this over a DDB table?  If the users find confusing to use DDB, we could write a couple of wrapper functions to make calling DDB easier, it sounds like 1hr work at max?

I am missing a lot of context about this decision but if others approve it feel free to merge.

kostaspap:
I think we might want to consider DDB over SimpleDB since SimpleDB is an older service and in keep-the-light-on mode. AWS has stopped developing features for it that's why it doesn't show up in the AWS Console for example.

austinbyers:
> I think we might want to consider DDB over SimpleDB since SimpleDB is an older service and in keep-the-lights-on mode

Ah, I thought SimpleDB was a wrapper around DynamoDB that made it easier to use. Very well, will revert to Dynamo

> Better than this would be some abstraction we give customers to put/get from the "PantherStore"

Agreed, but that's outside the scope of this change. @jacknagz just needs something immediately for demo deployments and such.

PR #896, opened 2020-05-15
Panther managed globals

nhakmiller:
## Background

We want to be able to ship some global helper functions that users can't change to have them be a part of the core panther experience.

## Changes

- When constructing the global layer, add a package for the `panther-helpers`, helper functions that panther manages

## Testing

- Deployed to dev account, confirmed it works and confirmed updating existing deployments works


nhakmiller:
This does not meet the requirements actually, closing this now.

PR #899, opened 2020-05-16
Add support for multiple globals

nhakmiller:
## Background

This PR adds support for creating multiple global files. At this time, there is no way to delete globals except via API call or direct access to dynamodb. Perhaps we can update the `panther_analysis_tool` logic to handle this case for CLI users.

## Changes

- Added `ListGlobals` API endpoint to the analysis-api
- Updated the `BulkUpload` endpoint of the analysis-api to only re-create the layer once regardless of how many globals were uploaded
- Updated the layer manager to use all globals when creating the layer

## Testing

- Deployed to dev account


nhakmiller:
Did some quick manual testing, looks good so far. Will give it some more thorough testing later before merging.

internal/core/analysis_api/handlers/list_globals.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/aws/aws-sdk-go/service/dynamodb/expression"
+
+	"github.com/panther-labs/panther/api/gateway/analysis/models"
+	"github.com/panther-labs/panther/pkg/gatewayapi"
+)
+
+// ListGlobals pages through globals from a single organization.
+func ListGlobals(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyResponse {
```
nhakmiller, line 37: I tried to work this in to the same methodology as list rules and list policies, where they share one `handleList` function that just does some switching based on type. But there were a lot of assumptions there that did not hold true for globals (such as having a `severity` or `enabled` field) and it seemed more straightforward just to bring the applicable logic over here instead of adding a ton of switching/branching/conditional lookup logic to the other function.

internal/core/analysis_api/handlers/util.go
```diff
@@ -187,7 +187,7 @@ func writeItem(item *tableItem, userID models.UserID, mustExist *bool) (int, err
 		// When policies and rules are also managed by globals, this can be moved out of the if statement,
 		// although at that point it may be desirable to move this to the caller function so as to only make the call
 		// once for BulkUpload.
-		return changeType, updateLayer(item.Type)
+		return changeType, nil
```
nhakmiller, line 190: Moved this out to the calling function so it only happens once per BulkUpload operation, as rebuilding the layer is moderately time consuming.

internal/core/layer_manager/manager/handle.go
```diff
@@ -91,20 +90,59 @@ func buildLayer() ([]byte, error) {
...
+
+	// If there are no globals, delete the layer
+	if len(globals) == 0 {
+		return nil, nil
+	}
+
+	zap.L().Debug("getting each global")
+	nameBodyMap := make(map[string]string)
+	// Iterate through each global and retrieve its body
+	for _, globalName := range globals {
```
nhakmiller, line 106: I considered updating the `ListGlobals` function to return the body, but there should not be more than 3-5 globals in most cases so this should not be too expensive.

internal/core/analysis_api/handlers/list_globals.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// Parse the input
+	ascending := defaultSortAscending
+	if sortDir := request.QueryStringParameters["sortDir"]; sortDir != "" {
+		switch sortDir {
+		case "ascending":
+			ascending = true
+		case "descending":
+			ascending = false
+		}
```
austinbyers, line 48: perhaps more simply,

```suggestion
		ascending = sortDir == "ascending"
```

internal/core/analysis_api/handlers/list_globals.go
```diff
@@ -0,0 +1,145 @@
...
+		globals = append(globals, &models.GlobalSummary{
+			ID:           item.ID,
+			LastModified: item.LastModified,
+			Tags:         item.Tags,
+		})
+		return nil
+	})
+
+	if err != nil {
+		return &events.APIGatewayProxyResponse{StatusCode: http.StatusInternalServerError}
```
austinbyers, line 105: Make sure that each of these internal server errors are logged somewhere. Either here, or in the `scanPages` function deeper in the call stack

internal/core/analysis_api/handlers/list_globals.go
```diff
@@ -0,0 +1,145 @@
...
+		expression.Name("tags"),
+	)
+	filter := expression.Equal(expression.Name("type"), expression.Value(typeGlobal))
+
+	expr, err := expression.NewBuilder().
+		WithFilter(filter).
+		WithProjection(projection).
+		Build()
+
+	if err != nil {
```
austinbyers, line 81: Log an error before returning 500 so we can troubleshoot

internal/core/analysis_api/handlers/list_globals.go
nhakmiller: Feels very pythonic to me, gives me shivers. updated.

internal/core/analysis_api/handlers/list_globals.go
nhakmiller: Updated here and elsewhere in this function.

nhakmiller:
Tested with new `panther_analysis_tool` in my dev account and it's working swimmingly 👍

internal/core/analysis_api/handlers/list_globals.go
austinbyers: Haha what is Go but a faster and safer Pythonic language?

PR #903, opened 2020-05-18
Provisioned capacity and TTL for key-value store

austinbyers:
## Background

Since the key-value store is designed to be accessed from user-defined Python, it would be very easy for DynamoDB costs to skyrocket. To prevent that, we're adding provisioned capacity with autoscaling.

This also adds a TTL column `expiresAt` so Python rules/policies can choose to expire items automatically if they want to.

## Changes

- Add provisioned capacity to panther-kv-store
- Add autoscaling IAM role and targets
- Add TTL column to panther-kv-store

## Testing

Deployed

Confirmed capacity settings are as expected:

Screen Shot 2020-05-18 at 10 51 10 AM

### DIFF
--- a/deployments/bootstrap.yml
+++ b/deployments/bootstrap.yml
@@ -938,7 +938,7 @@ Outputs:
   # Dynamo
   DynamoScalingRoleArn:
     Description: IAM role arn for DynamoDB auto-scaling
-    Value: !Ref DynamoScalingRole
+    Value: !GetAtt DynamoScalingRole.Arn
 
   # KMS
   OutputsEncryptionKeyId:
--- a/deployments/bootstrap.yml
+++ b/deployments/bootstrap.yml
@@ -498,6 +498,33 @@ Resources:
         - given_name
         - family_name
 
+  ########## Dynamo ##########
+  DynamoScalingRole:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Principal:
+              Service: application-autoscaling.amazonaws.com
+            Action: sts:AssumeRole
+      Policies:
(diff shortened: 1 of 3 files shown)


deployments/bootstrap.yml
```diff
@@ -498,6 +498,33 @@ Resources:
...
+              - Effect: Allow
+                Action:
+                  - dynamodb:DescribeTable
+                  - dynamodb:UpdateTable
+                  - cloudwatch:PutMetricAlarm
+                  - cloudwatch:DescribeAlarms
+                  - cloudwatch:GetMetricStatistics
+                  - cloudwatch:SetAlarmState
+                  - cloudwatch:DeleteAlarms
+                Resource: '*'
```
jacknagz, line 526: Should we break this into two policies and only give the DDB permissions on the kv-store table?

austinbyers: I could see us adding more provisioned tables in the future, but I suppose we can limit it to tables with a `panther-` prefix

deployments/core.yml
```diff
@@ -809,6 +812,7 @@ Resources:
       ServiceToken: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-cfn-custom-resources
 
   ##### Key-Value store for Python policies/rules #####
+  # TODO: build conditionally (disabled by default) since you have to pay for provisioned capacity
```
nhakmiller, line 815: Whenever this gets added as an option to `panther_config.yml`, it would be nice to also be able to set the capacity that is provisioned.

austinbyers: Yep, agreed. I added another TODO comment, just didn't implement it yet because of time constraints

deployments/bootstrap.yml
jacknagz: Perfect!

PR #906, opened 2020-05-18
downgraded to Info lvl messages

nhakmiller:
## Background

Currently when the `aws-event-processor` encounters an unknown event, it logs a WARN message. This causes alarms to go off indicating there is a problem, when in reality this is generally expected behavior. We used WARN messages to help remind ourselves to go add support or explicitly ignore irrelevant messages, but turns out 99% of the time it's an irrelevant message and we just whitelist it in the ignored event name logic.

This change downgrades these messages from WARN to INFO level. This way we are still logging the information and can periodically scan our logs looking for things we missed, but we don't alert on call rotations etc. for non impactful information.

## Changes

- Downgraded messages from WARN to INFO level

## Testing


jacknagz:
Could we capture these in a dashboard similar to the way we pull errors? I'd imagine we could use the error string pattern cc @rleighton

rleighton:
> Could we capture these in a dashboard similar to the way we pull errors? I'd imagine we could use the error string pattern cc @rleighton

Yes, that is a good idea. The we do not need to remember how to hunt for them.
I'll do that tomorrow, it will only take a few minutes.

PR #910, opened 2020-05-19
Restore form state in case of accidents

### LINKED ISSUE #855: Improvements for Global Python Module
### Description
Currently, the module is viewable/editable by navigating through the `Settings` menu. We also have a link on Policy/Rule details that opens the module on a new tab. This is happening because if we open the link on the same page, users will lose all their changes.

### Acceptance Criteria
- Don't open a new tab for global module but redirect in the same page.
- When a user leave a Policy/Rule form, save data on sessionStorage that way when the user comes back to the same Policy/Rule sees the changes that he hasn't yet saved.

### LINKED ISSUE #900: Accidental navigation leads to loss of edits

While editing a form, a user can accidentally navigate away and lose all of his work. We should figure out a way of either "warn" them against leaving or "restore" their session where  they left off

### Describe your use cases

People performing mass edits on Rule, Policies, GlobalModule, etc.

ie. Weekly

* Alert saying "are you sure you want to leave"?
* Mechanism for session restoration

### References

Are there any other GitHub issues, whether open or closed, that are related to the problem you've described above or to the suggested solution? If so, please create a list below that mentions each of them.

### Additional context

I think that  ideally the session restoration is a better option, because it's more passive (not an alert in your face) and doesn't require the user to think of whether he's sure that he wants to leave (which would mean trying to think what  changes he has got)

3nvi:
## Background

This PR adds persistence in our complex forms. This adds a mechanism where if you leave the page accidentally or hit refresh, you can continue where you left off with regards to  the form changes. That's especially useful for when you move from RuleForm -> Global Module -> RuleForm, since it allows you to move freely across  the app without fearing that your state is going to get lost.

This is currently implemented in:

- RuleForm
- PolicyForm
- GlobalModuleForm

as these are the most crucial & complex forms of the app.

Closes #900 
Closes #855 

## Changes

- Add `useFormSessionRestoration` hook and `FormSessionRestoration` component
- Use `<FormSessionRestoration>` in related forms
- Make global module open in the same tab as the one the user's currently in (currently spawns a new tab)

## Testing

- Refreshing while making changes to Policy/Rule, persists & restores your changes
- Restoring a closed tab restores your pending changes
- Navigating away and coming back, restores your form session


web/src/components/forms/GlobalModuleForm/GlobalModuleForm.tsx
```diff
@@ -58,49 +59,51 @@ const GlobalModuleForm: React.FC<GlobalModuleFormProps> = ({ initialValues, onSu
...
-            <Flex borderTop="1px solid" borderColor="grey100" pt={6} mt={10} justify="flex-end">
-              <Flex>
-                <Button variant="default" size="large" onClick={history.goBack} mr={4}>
-                  Cancel
-                </Button>
-                <SubmitButton>{initialValues.id ? 'Update' : 'Create'}</SubmitButton>
+          <FormSessionRestoration sessionId="global-module">
+            <Form>
+              <Text size="large" color="grey300" mb={4}>
+                The global module allows you to define a set of re-usable functions, variables and
```
jacknagz, line 65:
```suggestion
                The global module allows you to define a set of reusable functions, variables and
```

web/src/components/forms/GlobalModuleForm/GlobalModuleForm.tsx
```diff
@@ -58,49 +59,51 @@ const GlobalModuleForm: React.FC<GlobalModuleFormProps> = ({ initialValues, onSu
...
+                classes which can be directly imported to your Rule or Policy definition. Anything
+                defined below can later be imported through the aws_imports module
```
jacknagz, line 67:
```suggestion
                defined below can later be imported through the `panther` module
```

right @nhakmiller ?

alexmylonas: i can confirm too this is name of the new default global moduel

PR #916, opened 2020-05-20
Change webpack EJS loader

3nvi:
## Background

This PR changes the webpack loader of EJS files in order to align with the rest of our products. This PR doesn't add functionality, but syncs with changes in other repos.

## Changes

- Change ejs loader for development 

## Testing

- `npm run start` works


3nvi:
> I am approving, but a small point of interest is that if you check `package-lock.json` a lot of modules, moved out of `dev: true`.
> 
> Just have a quick look, don't think we can/should do anything.

You are  right! It's because I accidentally installed normal dependency instead of a dev one. Will fix it asap :)

Thanks for noticing!

PR #926, opened 2020-05-21
Custom resource: dynamo alarms

austinbyers:
## Background

Continuing our alarm migration, Dynamo alarms are next.

We currently have 150 total Dynamo alarms: 3 alarms (SystemErrors, Throttles, Integration) for each of 5 operations (PutItem, GetItem, etc) for each of our 10 tables. I feel this is excessive - having an alarm for every individual operation adds multiplicative alarm growth without adding much extra value.

So I've reduced it to just 4 alarms per table, essentially the same we had before just not broken down by request type:

* UserErrors (we didn't have this before)
* SystemErrors
* RequestLatency
* Throttles

If these trigger, we can still check the dashboards/metrics to understand which operations are contributing to the problem.

## Changes

- Add `Custom::DynamoDBAlarms`
- Remove DynamoDB alarm generation from `mage`
- Add CI enforcement so that all tables have associated alarms

## Testing

- Deploy + teardown, verified alarm configuration matches what we had before


austinbyers:
I'll wait for the current release to finish before merging

internal/core/custom_resources/resources/map.go
```diff
@@ -47,6 +47,15 @@ var CustomResources = map[string]cfn.CustomResourceFunction{
 	// PhysicalId: custom:alarms:appsync:$API_ID
 	"Custom::AppSyncAlarms": customAppSyncAlarms,
 
+	// CloudWatch alarms for AppSync 4XX and 5XX errors
```
rleighton, line 50: Cut-n-paste error?

austinbyers: Yep, you're right! I actually already fix it in the next PR (#932), so I'll just handle it there

PR #929, opened 2020-05-21
Storing reports in Athena

kostaspap:
## Background

Storing Rules reports in AWS Athena

## Changes

- Updated rules engine to store rules reports in Athena
- Rules reports are stored in Athena as `map[string][]string
- Currently it is only possible to manipulate the reports only through Analysis Tool. Updating a rule through  the UI will result to setting the Reports to `{}`. 

## Testing

- mage test:ci
- Deployed in AWS and verified everything works. 
Screen Shot 2020-05-22 at 10 17 18 AM

@jacknagz Let me know if the structure (map[string][]string) works for the queries you plan to perform. 
Also, Reports are only manageable through the Panther Analysis Tool. If someone tries to update the Rule through the UI, the Reports will be set to `{}` (the UI sends `{}` for Reports)


rleighton:
I don't understand the term "reports" or how this is meant to be used. Looks like tags to me.

nhakmiller:
> I don't understand the term "reports" or how this is meant to be used. Looks like tags to me. Why not generalize tags?

Reports and tags are similar in functionality, but have slightly different purposes. I think the distinction is useful. Generally tags are intended to be used for administrative purposes, whereas reports and intended to be used specifically for generating some sort of report card for how your organization is doing against some standard (such as CIS or PCI).

One example is we have the `Configuration required` tag that is used to indicate users need to go manually configure & enable a policy/rule. You likely would not desire a report on how many `Configuration required` rules are passing. A report will be something like `CIS` or `PCI`, which is something you would want a report on.

And of course reports can carry more context since it's a `map[string][]string` instead of just a `[]string`. This context is useful for reports, as you may wish to indicate that in one report the rule/policy meets multiple controls. For tags this would rarely be useful and just add complexity I think.

We are still very early in the implementation of repots though, so I'm not sure what exactly they will look like.

rleighton:
> > I don't understand the term "reports" or how this is meant to be used. Looks like tags to me. Why not generalize tags?
> 
> Reports and tags are similar in functionality, but have slightly different purposes. I think the distinction is useful. Generally tags are intended to be used for administrative purposes, whereas reports and intended to be used specifically for generating some sort of report card for how your organization is doing against some standard (such as CIS or PCI).
> 
> One example is we have the `Configuration required` tag that is used to indicate users need to go manually configure & enable a policy/rule. You likely would not desire a report on how many `Configuration required` rules are passing. A report will be something like `CIS` or `PCI`, which is something you would want a report on.
> 
> And of course reports can carry more context since it's a `map[string][]string` instead of just a `[]string`. This context is useful for reports, as you may wish to indicate that in one report the rule/policy meets multiple controls. For tags this would rarely be useful and just add complexity I think.
> 
> We are still very early in the implementation of repots though, so I'm not sure what exactly they will look like.

So this new attribute is not _the_ reports, rather it is the name(s) of reports that can _later_ be generated and this is how the code will find the policies and rules associated with the compliance standards. That makes sense. I think a comment or two is in order :)

Maybe change the name to `compliance_reports`?

Having this activity in the database will allow easy BI reports showing where there is activity w.r.t. a standard over time.

kostaspap:
> I think this may break integration tests. Can you sync with master and confirm they still work and/or update.

Synced and ran integration tests. You are right, there were some issues that I fixed.

PR #932, opened 2020-05-22
Custom resource: lambda alarms

austinbyers:
## Background

Moving lambda alarms into a custom resource. I kept the structure essentially the same: alarms for logged warnings, logged errors, memory, duration, and execution errors. I did standardize the interval on all of them to 5 minutes just to be consistent

This includes dynamo changes from #926, that PR should merge first (and you need not review dynamo changes here)

## Changes

- Replaced alarm CF generation with `Custom::LambdaAlarms`
- Add CI enforcement to guarantee every Lambda function in our CloudFormation has an associated alarm group in the same template
- Bugfix: Custom resource properties do not need to be parsed for DELETE operations. Not only does this waste a few cycles, but it causes problems if an alarm is invalid:
    1. CREATE fails because you forgot a field
    2. CloudFormation rolls back, which then triggers a DELETE on the custom resource
    3. The DELETE fails for the same reason CREATE failed - invalid field

## Testing

- Compared Lambda alarms in the CloudWatch console side-by-side with the old and the new to confirm same settings applied
- Verified duration and memory by clicking through to the alarm details, where CloudWatch graphs the tracked metric - it had the expected data
- Triggered errors on a function


deployments/alarms.yml
```diff
@@ -37,13 +37,6 @@ Resources:
         AlarmTopicArn: !Ref AlarmTopicArn
       TemplateURL: ../out/deployments/monitoring/bootstrap_alarms.json
 
-  GatewayAlarms:
-    Type: AWS::CloudFormation::Stack
-    Properties:
-      Parameters:
-        AlarmTopicArn: !Ref AlarmTopicArn
-      TemplateURL: ../out/deployments/monitoring/bootstrap_gateway_alarms.json
```
austinbyers, line 45: This nested stack is now completely empty (all alarms from that stack have migrated)

internal/core/custom_resources/resources/parser.go
```diff
@@ -31,8 +31,8 @@ var inputValidator = validator.New()
 //
 // Out must be a pointer to a struct with appropriate `validate` tags
 func parseProperties(params map[string]interface{}, out interface{}) error {
-	if len(params) == 0 {
-		return nil
+	if params == nil {
+		params = make(map[string]interface{})
```
austinbyers, line 35: We still need to apply the struct validation on an empty map, we shouldn't have returned early

internal/core/custom_resources/resources/alarms_lambda.go
```diff
@@ -0,0 +1,178 @@
...
+	switch event.RequestType {
+	case cfn.RequestCreate, cfn.RequestUpdate:
+		var props LambdaAlarmProperties
+		if err := parseProperties(event.ResourceProperties, &props); err != nil {
+			return "", nil, err
+		}
+
+		// Set defaults
+		// TODO - better parsing of properties with defaults
+		if props.FunctionMemoryMB == 0 {
```
austinbyers, line 63: I debated adding a `default` tag to the properties struct above, but then we'd have to use reflect to fill it in manually (unless there's a library we could use). I'm happy to do that, but it could get fairly complex if we wanted nested properties, lists, etc. Filling in the defaults manually like this is by far the simplest solution

Using reflect directly would have some other advantages over our current json marshal/unmarshal parsing - it'd be a bit faster (not very important) and it would allow us to warn about unused fields. Today, any mistyped custom resource property is simply ignored. This would presumably fail validation on your properties, but it'd be nice to have a warning

internal/core/custom_resources/resources/alarms_lambda.go
```diff
@@ -0,0 +1,177 @@
...
+	input.AlarmName = aws.String(
+		fmt.Sprintf("Panther-%s-%s", lambdaThrottleAlarm, props.FunctionName))
+	input.MetricName = aws.String("Throttles")
+	input.Threshold = aws.Float64(float64(*props.ThrottleThreshold))
+	if err := putMetricAlarm(input); err != nil {
+		return err
+	}
+
+	input.AlarmDescription = aws.String(fmt.Sprintf(
+		"Lambda function %s is using more than 90%% of its allotted execution time. See: %s#%s",
```
nhakmiller, line 168: Is the `90%%` a typo?

austinbyers: Nope, that's how to get a literal `%` sign in a format string

fmt.Printf("Give 110%% \n")
> Give 110%

deployments/cloud_security.yml
```diff
@@ -862,6 +918,15 @@ Resources:
       LogGroupName: !Ref ResourcesApiLogGroup
       ServiceToken: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-cfn-custom-resources
 
+  ResourcesApiAlarms:
+    Type: Custom::LambdaAlarms
+    Properties:
+      AlarmTopicArn: !Ref AlarmTopicArn
+      FunctionMemoryMB: 512
```
rleighton, line 925: Having this constants in 2 places is dangerous. Can we move all of them to parameters and use !Ref ?

internal/core/custom_resources/resources/alarms_lambda.go
rleighton: Why not just fail if not set? See my comment about using !Ref for these values for safety.

deployments/cloud_security.yml
austinbyers: I'd rather not move them all to parameters or we'd get too close to the [60 parameter limit](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html) (timeout + memory param for 10-15 functions per stack)

We can, however, make a Mapping for them, like a constant at the top of each template. Mappings have higher limits and we don't use them as often. I'll do that in a follow-up PR

Alternatively, we could use CI to enforce the values in the alarm match the value in the function. But I like the mapping idea because then it's easy to tweak memory/timeout for an entire stack at once, and we sometimes need these values in several places (e.g. the timeout is also used for SQS queues)

internal/core/custom_resources/resources/alarms_lambda.go
austinbyers: Since 90% of our Lambda functions use the default memory (128 MB), the idea was to use that as the default so we didn't have to specify the same value 20 times. This helps keep the template size smaller, for example

But per your earlier comment, we do need a way to keep these values aligned, in which case it would then be a required parameter, yes

PR #947, opened 2020-05-26
Types and helpers for `null` JSON value handling

alxarch:
## Background

The current method of handling `null` values in input JSON is to use pointer types in struct fields (ie *string). This comes with an allocation penalty during unmarshal for every such field that has a value. This cost adds up in big log structs and is paid for every line of log that is parsed. The changes in this PR provide a group of packages with types and helpers to handle these values. This PR is a partial PR of changes related to parsers refactoring as described in #902. Changes will be merged and reviewed incrementally to ease the transition. 

## Changes

- Adds `null` package that defines types for nullable JSON values that omit zero values in the output.
- Adds `strictnull` package that defines types for nullable JSON values that don't omit zero values in the output.
- Adds `box`, 'unbox` packages to replace helper functionality found in the `aws` package to avoid unnecessary dependency for parsers for handling of pointer values.

## Testing

- Includes extensive unit tests. Since code in this PR is not currently used by parsers it does not affect anything to need E2E tests.


internal/log_analysis/log_processor/common/box/box.go
```diff
@@ -0,0 +1,64 @@
+// Package box provides boxing helpers for scalar values.
+// This package exists to help the transition form pointer based fields to `null` fields
+package box
```
kostaspap, line 3: Can you put this in the `pkg` folder? https://github.com/panther-labs/panther/tree/master/pkg 

We could start using these methods in other parts of the code, instead of putting an AWS dependency. 

Same comment for `unbox`

internal/log_analysis/log_processor/common/strictnull/strictnull.go
```diff
@@ -0,0 +1,2 @@
+// Package strictnull provides types for lossless handling of nullable JSON values.
+package strictnull
```
alxarch, line 2: This package provides separation of logic of whether we want to include blank string values in the output JSON.
It uses the package name as a namespace to distinguish itself from the `null` package. Another solution would be to define the types and tests in the `null` package and use a `Strict` prefix for all types. Please review to decide on the final naming strategy.

internal/log_analysis/log_processor/common/box/box.go
alxarch: I can move the packages there. Does this apply for the `null`, `strictnull` packages as well?

kostaspap: While I think that the `box` and `unbox` we could start using in other projects right away, I am not sure we will do that for the `null` and `strictnull` . We might do that in the future, but until that happens I think we can keep them here since they will be used mainly by the `log_processor`.

alxarch: fixed

PR #949, opened 2020-05-26
Remove glue table stack

rleighton:
## Before
Glue tables were declared via CF and generated from the registry of all log parsers. 
As the number of log parser grew we exceeded the max CF template size.
See: https://github.com/panther-labs/panther-enterprise/issues/187

Also, we are moving to a SAR deployment model which means that generating stacks is not an option. 

The solution is to create tables only when they are needed during log source onboarding.

## After

- The source-api creates tables needed for the specified log sources when they are registered and the views updated.
- The glue stack and glue stack generation has been removed.
- During deployment all _deployed_ tables are updated to get any schema changes and their partitions sync'd.
- The legacy glue stack is NOT removed during deployment to avoid disruption to log processing, it is removed during a teardown.

## Testing

- mage test:ci
- mage deploy + mage teardown
- mage deploy v1.3.0 + mage deploy master + mage teardown


tools/mage/deploy_onboard.go
```diff
@@ -114,6 +114,18 @@ func registerPantherAccount(awsSession *session.Session, settings *config.Panthe
 		return fmt.Errorf("error calling lambda to register account: %v", err)
 	}
 
+	// collect the log types configured in panther_config.yml
```
rleighton, line 117: Moving this here and using it to compare fixes a bug where if you added Cloudtrail AFTER having onboarded then the source integration was not being updated.

tools/mage/deploy.go
```diff
@@ -435,9 +435,15 @@ func deployMainStacks(awsSession *session.Session, settings *config.PantherConfi
 		c <- goroutineResult{summary: frontendStack, err: err}
 	}(results)
 
+	// Glue (not a stack!, moving to custom resource)
```
austinbyers, line 438: Is this a TODO - we need to add a custom resource?

If so, I'm a bit confused - I thought `source-api` was handling all of the glue tables? Can you help clarify:

* What does `source-api` now do with Glue tables?
* What needs to happen on every deploy with a custom resource? Is that coming in the next PR?
* What does `glue:sync` do and when should it be run?

rleighton:
> > The legacy glue stack is NOT removed during deployment to avoid disruption to log processing, it is removed during a teardown.
> 
> So this means all existing deployments need to manually delete the glue stack? Or they should leave it indefinitely and it won't ever cause problems? Neither option seems very appealing.
> 
> My understanding was that log processing works fine without the glue tables, but historical searches wouldn't work until the glue catalog is configured correctly? If that's the case, could we check for the glue stack and, if it exists, delete it / update glue before continuing the deploy?

The existing log processing needs to interact with the glue catalog to manage partitions. Dropping a table w/out stopping log processing will cause errors.  Leaving  the stack causes no harm and will be removed on teardown.  This is the simplest and most robust "migration".

tools/mage/deploy.go
rleighton: The source-api will add new tables now rather than adding all tables at once via CF.

The code executed above needs to be executed on on each deploy to manage any schema changes to existing tables. Before the CF took care of that. This code also to sets up Athena (the Athena part is not new). This the part I was hoping you would migrate to a custom resource. 

`glue:sync` should not be needed anymore except in cases of migration, so I want to leave it.

austinbyers: > The code executed above needs to be executed on on each deploy to manage any schema changes to existing tables. Before the CF took care of that. This code also sets up Athena

I see, so if an existing log schema adds a new field for a table that's already created, this code is what will update the table definition?

One caveat: custom resources will never be updated unless one of their parameters changes. You want it to trigger on every deploy, right? We could add a dummy parameter which is a random string or maybe a hash of the schema definitions

As for ownership, I'm happy to try it, but packaging is a hard requirement for the next release, so depending on timeline I may need your help.

rleighton: Exactly, this is needed on each deploy to deal with changes (which were handled in the CF).

I just figured since time is short and you create the other custom resources you would be able to do it most quickly (just wrap that one function). I can give it a try in this PR if you like, just may take me a little longer.

rleighton: I did 2min of googling, some folks just add a timestamp as a parameter to ensure the resource is invoked each time.

rleighton: How would that work in SAR (in mage we can just add a timestamp to the calling stack)?

rleighton: Also, there is this (which I hope we never have to use): https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros.html

rleighton: Actually, for SAR we could just use the Panther version, which should change each time. For mage we would set the parameter to the version+timestamp (so devs can test changes).

rleighton: I'll go ahead now and take a crack at the custom resource in this PR.

austinbyers: > I did 2min of googling, some folks just add a timestamp as a parameter to ensure the resource is invoked each time.

Makes sense. Yes, in SAR it can just be the Panther version and then mage can be whatever we want

But is this a fast operation if there are no changes? We don't want to invoke a custom resource on every deploy (for developers) if it takes several minutes but doesn't do anything. Even if it is fast, ideally we can skip the custom resource if it didn't change. That's why I had suggested a hash of the log schemas, so it only changes when the logs themselves have actually changed. That hash can be hardcoded when publishing to SAR, but I'm not sure if that's an easy thing to compute right now?

Also, will the 15-minute Lambda timeout be sufficient? If not, we'll have to use SNS custom resource invocation, which I'm not as familiar with

tools/mage/deploy.go
```diff
@@ -450,25 +456,62 @@ func deployMainStacks(awsSession *session.Session, settings *config.PantherConfi
...
+	deployedLogTables, err := gluetables.DeployedTables(glueClient)
+	if err != nil {
+		logger.Fatal(err)
+	}
+	for _, logTable := range deployedLogTables {
+		logger.Infof("deploy: creating/updating glue table %s.%s", logTable.DatabaseName(), logTable.TableName())
+		err := logTable.CreateOrUpdateTable(glueClient, outputs["ProcessedDataBucket"])
+		if err != nil {
+			return errors.Wrapf(err, "could not create glue log table for %s.%s", logTable.DatabaseName(), logTable.TableName())
+		}
```
kostaspap, line 496: I am not sure I understand, why is this needed? The tables are already deployed, why do we need to run "CreateOrUpdate" on them?

tools/mage/deploy.go
```diff
@@ -450,25 +456,62 @@ func deployMainStacks(awsSession *session.Session, settings *config.PantherConfi
...
+		err = logTable.SyncPartitions(glueClient, s3Client, zeroStartTime)
+		if err != nil {
+			logger.Fatalf("deploy: failed syncing %s.%s: %v", logTable.DatabaseName(), logTable.TableName(), err)
+		}
```
kostaspap, line 500: This can be time-consuming, it could delay deployments especially during development. Seems it would be required only on specific version migrations?

tools/mage/deploy.go
```diff
@@ -450,25 +456,62 @@ func deployMainStacks(awsSession *session.Session, settings *config.PantherConfi
...
+		logger.Infof("deploy: creating/updating glue table %s.%s",
+			ruleTable.DatabaseName(), ruleTable.TableName())
+		err = ruleTable.CreateOrUpdateTable(glueClient, outputs["ProcessedDataBucket"])
+		if err != nil {
+			return errors.Wrapf(err, "could not create glue rule table for %s.%s", ruleTable.DatabaseName(), ruleTable.TableName())
+		}
+		err = ruleTable.SyncPartitions(glueClient, s3Client, zeroStartTime)
+		if err != nil {
+			logger.Fatalf("deploy: failed syncing %s.%s: %v", ruleTable.DatabaseName(), ruleTable.TableName(), err)
+		}
```
kostaspap, line 514: Same as above. Seems it would be required only when migrating between specific versions.

tools/mage/glue_namespace.go
```diff
@@ -85,16 +63,25 @@ func (t Glue) Sync() {
 		logger.Infof("syncing %s", name)
 		err := table.SyncPartitions(glueClient, s3Client, startDate)
 		if err != nil {
-			logger.Fatalf("failed syncing %s: %v", name, err)
+			if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == glue.ErrCodeEntityNotFoundException {
```
kostaspap, line 66: * This code seems to be pretty identical to what is running in the `deploy.deployGlue` command. Might be easy to extract to common function? 
* Line 58 should be `gluetables.DeployedTables()`, right?

internal/log_analysis/awsglue/integration_test.go
```diff
@@ -103,15 +82,25 @@ func TestIntegrationGlueMetadataPartitions(t *testing.T) {
 		removeTables(t)
 	}()
 
-	gm := NewGlueTableMetadata(models.LogData, testTable, "test table", GlueTableHourly, &testEvent{})
+	// get the meta data, note we update the schema from the one used in setupTables()
+	gm := NewGlueTableMetadata(models.RuleData, testTable, "test table", GlueTableHourly, &testEventModified{})
```
kostaspap, line 86: For my own understanding, is there a reason it was changed from `LogData` to `RuleData`?

kostaspap:
Nice cleanup, very happy to see the CFN generation gone :D

internal/core/source_api/api/put_integration.go
```diff
@@ -89,25 +89,25 @@ func (api API) PutIntegration(input *models.PutIntegrationInput) (*models.Source
...
-	}
-
 	switch aws.StringValue(input.IntegrationType) {
 	case models.IntegrationTypeAWS3:
 		permissionAdded, err = AllowExternalSnsTopicSubscription(*input.AWSAccountID)
 		if err != nil {
 			err = errors.Wrap(err, "Failed to add permissions to log processor queue")
 			return nil, putIntegrationInternalError
 		}
+		err = addGlueTables(input)
```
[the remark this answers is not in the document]
rleighton: Good point.

internal/core/source_api/api/put_integration.go
```diff
@@ -148,8 +148,9 @@ func (api API) integrationAlreadyExists(input *models.PutIntegrationInput) error
 				return nil
 			case models.IntegrationTypeAWS3:
 				if *existingIntegration.AWSAccountID == *input.AWSAccountID &&
-					*existingIntegration.IntegrationLabel == *input.IntegrationLabel {
-					// Log sources for same account need to have different labels
+					*existingIntegration.IntegrationLabel == *input.IntegrationLabel &&
+					len(existingIntegration.LogTypes) == len(input.LogTypes) {
```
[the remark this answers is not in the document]
rleighton: Yeah, that was me "fixing" a bug , but it is not a bug. The onboarding code has a bug where if you add CT in the panther config after deploying it is not picked up.  The real fix is for that code to call update, in that case since put does not update. Will fix.

internal/log_analysis/awsglue/awsglue.go
```diff
@@ -49,6 +49,7 @@ var (
 		LogProcessingDatabaseName: LogProcessingDatabaseDescription,
 		RuleMatchDatabaseName:     RuleMatchDatabaseDescription,
 		ViewsDatabaseName:         ViewsDatabaseDescription,
+		TempDatabaseName:          TempDatabaseDescription,
```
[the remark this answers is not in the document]
rleighton: Yes, used for Parquet CTAS and declared here to minimize merge changes. Originally it was not put in this list to keep it out of the UI, but we need a master list so  put it back in.

internal/log_analysis/awsglue/integration_test.go
rleighton: Yes, to exercise all the code paths in the test. For rule data we append some columns and I wanted to cover that.

internal/log_analysis/awsglue/schema.go
```diff
@@ -26,13 +26,107 @@ import (
...
+			Comment: "The tags of the rule that generated this alert",
+		},
+	}
+)
+
+type Column struct {
+	Name     string
+	Type     string // this is the Glue type
+	Comment  string `json:",omitempty"`
+	Required bool   `json:"-"` // do NOT serialize! Not used for Glue types (used for doc).
```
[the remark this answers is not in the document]
rleighton: Good point, yes.

internal/log_analysis/gluetables/gluetables.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/glue"
+	"github.com/aws/aws-sdk-go/service/glue/glueiface"
+	"github.com/pkg/errors"
+
+	"github.com/panther-labs/panther/internal/log_analysis/awsglue"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/registry"
+)
+
+func DeployedTables(glueClient glueiface.GlueAPI) (deployedLogTables []*awsglue.GlueTableMetadata, err error) {
```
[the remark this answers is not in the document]
rleighton: I wish, import cycle issues.

tools/mage/deploy.go
```diff
@@ -435,9 +435,15 @@ func deployMainStacks(awsSession *session.Session, settings *config.PantherConfi
...
+	// Glue (not a stack!, moving to custom resource)
+	count++
+	go func(c chan goroutineResult) {
+		c <- goroutineResult{summary: "panther-glue-tables", err: deployGlue(awsSession, outputs)}
+	}(results)
+
 	// Wait for stacks to finish.
 	// There are two stacks before and one stack after
-	logResults(results, "deploy", 3, count+2, len(allStacks))
+	logResults(results, "deploy", 3, count+2, len(allStacks)+1)
```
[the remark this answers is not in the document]
rleighton: Moving to custom resource, so this is going to be deleted.

tools/mage/deploy.go
rleighton: Schema change (e.g., add a Panther column).

tools/mage/deploy.go
rleighton: I'll be computing a hash of the deployed tables, so this will only run when needed in the custom resource.

tools/mage/deploy.go
rleighton: See above, next update should be more clever.

tools/mage/deploy_onboard.go
```diff
@@ -126,7 +138,8 @@ func registerPantherAccount(awsSession *session.Session, settings *config.Panthe
 		}
 		if aws.StringValue(integration.AWSAccountID) == accountID &&
 			*integration.IntegrationType == models.IntegrationTypeAWS3 &&
-			*integration.IntegrationLabel == genLogProcessingLabel(awsSession) {
+			*integration.IntegrationLabel == genLogProcessingLabel(awsSession) &&
+			len(integration.LogTypes) == len(logTypes) {
```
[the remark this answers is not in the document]
rleighton: Yeah, will fix. The _bug_ is we need to do an update if it already exists and the log types have changed. Will fix.

tools/mage/glue_namespace.go
rleighton: I wrestled with that for a bit and moved on ... might make more sense now that the code is in better shape. I'll give it a try.

tools/mage/deploy_onboard.go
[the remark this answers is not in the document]
rleighton: Fixed

tools/mage/deploy_onboard.go
```diff
@@ -185,7 +196,25 @@ func registerPantherAccount(awsSession *session.Session, settings *config.Panthe
 
 			return fmt.Errorf("error calling lambda to register account for log processing: %v", err)
 		}
-		logger.Infof("deploy: account %s registered for log processing", accountID)
+		logger.Infof("deploy: account %s registered for log processing on bucket %s for %v",
+			accountID, auditLogsBucket, logTypes)
+	}
+
+	if logProcessingIntegration != nil { // need to update
```
rleighton, line 203: This fixes the bug where we were not updating log types when changes were made in `panther_config.yml`

tools/mage/deploy.go
rleighton: Added custom resources

internal/log_analysis/awsglue/table_metadata.go
```diff
@@ -114,6 +118,80 @@ func (gm *GlueTableMetadata) PartitionKeys() (partitions []PartitionKey) {
...
+	_, err := glueClient.CreateTable(createTableInput)
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == glue.ErrCodeAlreadyExistsException {
+			// need to do an update
+			updateTableInput := &glue.UpdateTableInput{
+				DatabaseName: &gm.databaseName,
+				TableInput:   tableInput,
+			}
+			_, err := glueClient.UpdateTable(updateTableInput)
+			return err
```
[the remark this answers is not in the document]
rleighton: Done

internal/log_analysis/awsglue/table_metadata.go
```diff
@@ -114,6 +118,80 @@ func (gm *GlueTableMetadata) PartitionKeys() (partitions []PartitionKey) {
...
+		}
+		return err
```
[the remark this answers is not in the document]
rleighton: Done

internal/log_analysis/awsglue/wrappers.go
```diff
@@ -30,16 +30,29 @@ import (
 
 // Wrapper functions to reduce boiler-plate code in callers
 
+func CreateDatabase(client glueiface.GlueAPI, databaseName, databaseDescripiton string) (*glue.CreateDatabaseOutput, error) {
```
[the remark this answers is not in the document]
rleighton: Done

internal/core/source_api/api/put_integration.go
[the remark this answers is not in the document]
rleighton: Fixed

internal/log_analysis/awsglue/schema.go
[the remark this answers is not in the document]
rleighton: Changed

tools/mage/deploy.go
rleighton: Added use of table "signatures" for change detection.

tools/mage/glue_namespace.go
rleighton: Not in deploy anymore, in custom resource.

internal/log_analysis/awsglue/table_metadata.go
```diff
@@ -114,6 +125,104 @@ func (gm *GlueTableMetadata) PartitionKeys() (partitions []PartitionKey) {
...
+				Parameters:           descriptorParameters,
+			},
+		},
+		TableType: aws.String("EXTERNAL_TABLE"),
+	}
+}
+
+func (gm *GlueTableMetadata) Signature() (string, error) {
+	tableInput := gm.glueTableInput("")
+	tableInputJSON, err := json.MarshalIndent(tableInput, "", "") // Indent forces sorting for consistency
```
rleighton, line 192: I did not use `jsoniter` here because it did not produce consistent data.

deployments/log_analysis.yml
```diff
@@ -87,6 +93,22 @@ Conditions:
...
+    Properties:
+      AthenaResultsBucket: !Ref AthenaResultsBucket
+      ServiceToken: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-cfn-custom-resources
+
+  ###### Update Glue Table Schemas for Deployed Tables #####
+  UpdateGlueTables:
+    DependsOn: AthenaConfigure
+    Type: Custom::UpdateGlueTables
+    Properties:
+      TablesSignature: !Ref TablesSignature # this should change when tables need to be updated, for CF this can be the Panther version
```
austinbyers, line 108: Nice! We can also just compute the hash as part of the release process and hardcode it in the final master template, right? Either way, the version will be available

deployments/master.yml
```diff
@@ -110,7 +110,8 @@ Resources:
         # If the AlarmTopicArn parameter is blank, Bootstrap will create the topic.
         # That's why this uses the Bootstrap output instead of the parameter.
         AlarmTopicArn: !GetAtt Bootstrap.Outputs.AlarmTopicArn
-
+        AthenaResultsBucket: !GetAtt Bootstrap.Outputs.AthenaResultsBucket
+        ProcessedDataBucket: !GetAtt Bootstrap.Outputs.ProcessedDataBucket
```
austinbyers, line 114: Yay! Thanks for remembering to update the master stack, I'll use CI to help enforce this in the future

internal/core/custom_resources/resources/athena.go
```diff
@@ -0,0 +1,57 @@
...
+				workgroup, props.AthenaResultsBucket, err)
+		}
+
+		return resourceID, nil, nil
+	case cfn.RequestDelete: // noop
+		return event.PhysicalResourceID, nil, nil
+
+	default:
+		return "", nil, fmt.Errorf("unknown request type %s", event.RequestType)
+	}
```
austinbyers, line 56: Since it isn't really possible to have an invalid request type, I usually just collapse these:

```suggestion
	default:
    		// ignore deletes
   		return event.PhysicalResourceID, nil, nil
```

Whichever you prefer

internal/core/custom_resources/resources/glue.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+
+	"github.com/panther-labs/panther/internal/log_analysis/athenaviews"
+	"github.com/panther-labs/panther/internal/log_analysis/awsglue"
+	"github.com/panther-labs/panther/internal/log_analysis/gluetables"
+)
+
+type UpdateGlueTablesProperties struct {
+	// TablesSignature should change every time the tables change (for SAR this can be the Panther version)
```
austinbyers, line 38: Nit: Replace "SAR" references to "pre-packaged" or "master stack" or similar

internal/core/source_api/api/glue_utils.go
```diff
@@ -18,3 +17,25 @@ package gluecf
...
+		}
+	}
+
+	// update the views with the new tables
+	err := athenaviews.CreateOrReplaceViews(glueClient, athenaClient)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
austinbyers, line 40:
```suggestion
	return athenaviews.CreateOrReplaceViews(glueClient, athenaClient)
```

internal/log_analysis/awsglue/table_metadata.go
austinbyers: Ah, there's a setting for that:

```go
json := jsoniter.ConfigCompatibleWithStandardLibrary
b, err := json.Marshal(m)
```

This will serialize in lexicographic order like the standard library. Not as fast as the default jsoniter speed, but still faster than the stdlib

tools/mage/deploy_onboard.go
austinbyers: nit: can we elaborate on the comment? It looks like a leftover TODO instead of "this setting is defined, we need to update the integration settings..."

tools/mage/teardown.go
```diff
@@ -212,6 +212,11 @@ func destroyCfnStacks(awsSession *session.Session, identity *sts.GetCallerIdenti
 		logger.Fatal(err)
 	}
 
+	// In v1.4.0 we removed the stack `panther-glue`, delete it (we can remove this after a few more releases)
+	if err := deleteStack(client, aws.String("panther-glue")); err != nil {
```
austinbyers, line 216: I thought you wanted to leave it for stability?

internal/core/custom_resources/resources/glue.go
```diff
@@ -0,0 +1,123 @@
...
+
+		// update the views with the new tables
+		err = athenaviews.CreateOrReplaceViews(glueClient, athenaClient)
+		if err != nil {
+			return "", nil, errors.Wrap(err, "failed creating views")
+		}
+
+		return resourceID, nil, nil
+
+	case cfn.RequestDelete:
```
austinbyers, line 105: Since there's deletion code in the custom resource, does that mean we can remove anything from `mage teardown`?

tools/mage/deploy.go
rleighton: Done

deployments/log_analysis.yml
rleighton: Either way.

tools/mage/teardown.go
rleighton: Not when we teardown, nothing is left :)

internal/core/custom_resources/resources/glue.go
rleighton: Already done. The stack is gone from the slice of stacks. Just the explicit delete is there to get rid of legacy stacks. We can probably remove that in a month or two.

tools/mage/teardown.go
austinbyers: (I thought this was in `deploy.go` for some reason, my bad lol)

Issue #851, opened 2020-05-06
Master Panther template


austinbyers:
### Description

For automated and pre-packaged deployments, we need to be able to offer a single master template which can complete an end-to-end deployment.

This template would contain a reference to the main SAR app (with compiled Lambda source) as well as resources that SAR does not currently support:

* `AWS::GuardDuty::Detector` (self-onboarding)
* `AWS::EC2::FlowLog`
* `AWS::ECS::*`

(Alternatively, those resources could be implemented as custom resources in the SAR app if it makes sense to do so.)

### Acceptance Criteria

- A single CloudFormation template can deploy all of Panther (preferably without needing to spin up an EC2 instance / Fargate container)
- Deploying from source still works as before without changing the stack structure (using the same templates)

Issue #864, opened 2020-05-08
SAR CI enforcement


austinbyers:
### Description

Not all CloudFormation resources are supported in SAR, but "sam package" will not warn you about them. You will only realize when you try to "sam publish" the application, which we would do as part of a release process.

We need a simple CI check to warn developers if they're about to add a resource which wouldn't be supported in the applicable stacks.

### Acceptance Criteria

- CI fails if a developer tries to add a resource not supported by SAR in a stack that functions as a SAR app

austinbyers:
In addition to unsupported SAR resources, it'd be nice to check for SAR/CloudFormation limits:

* Number of resources
* Size of template
* Size of packaged SAR app

etc. These could be added as comments to templates so we know how close we are to hitting the limits. The SAR limits in particular we will not be aware of until we start prepping the release.

austinbyers:
Since we are no longer using SAR, closing this issue as not relevant.

Checking for CloudFormation limits more generally could still be useful, but we have no specific need for it at the moment

PR #950, opened 2020-05-26
Pre-packaged deployment: initial master stack

austinbyers:
## Background

This is the majority of the master stack that users will deploy in future releases (#851). The master stack could be deployed directly from the AWS console or CLI or (more commonly), it can be deployed as a nested CloudFormation stack

```yaml
Resources:
  Panther:
    Properties:
      TemplateURL: panther-s3-public/panther-community-master.yml
      Parameters:
        CloudWatchLogRetentionDays: 90
        # ... other Panther settings ...
      Tags:
        Application: Panther
      TimeoutInMinutes: 60
```

That's it! Your entire Panther deployment in a single 10-20 line template

### Deploy-From-Source vs Deploy-Master-Stack
The current deploy-from-source process is unchanged and for the foreseeable future we will have to support both workflows:

* When deploying from source, we use `panther_config.yml` for settings and `mage` to deploy stacks in parallel and pass stack outputs around
* When deploying the master stack, all settings are top-level parameters (no config file) and CloudFormation itself manages the nested stack deployment

This split is not ideal - if you add a new parameter to a stack, you now have to update both mage and the master stack to pass in that new parameter. Eventually, I'd like deploy-from-source to switch to use the same master stack. Then we could remove almost all of the `mage deploy` code and have nearly identical packaging/deploy processes.

Unfortunately, deploying from the master template results in a different stack structure (our current top-level stacks are nested one level deeper) which means it isn't trivial to migrate from one to the other. Since we have existing deployments from source, we will have to keep them that way for now. It is theoretically possible to use CloudFormation imports to do a migration in the future, but with several hundred resources it will take more developer time than we can currently afford.

### Dropping SAR
The initial vision was to use SAR to contain the entire application so we didn't have to manage assets in a public S3 bucket. SAR is just a CloudFormation template which bundles Lambda binaries in itself. We'd still like to do that eventually, but for now we're just going to rely on standard CloudFormation packaging with a public S3 bucket for the following reasons:

* SAR has a ~50MB package limit which some of our stacks are dangerously close to hitting. Our Go binaries are 5-12 MB each.
* Both the `onboard` and `web_server` stacks contain resources that are not supported by SAR (GuardDuty and ECS, respectively). Without migrating these to custom resources, we have to use a public bucket for these templates; they cannot be packaged in SAR. Since we can't avoid a public bucket, might as well use it for everything
* It turns out SAR only packages Lambda source into itself. Other packaged assets (e.g. GraphQL schema) remain pointing to the original S3 bucket.

If we used SAR for the majority of stacks, the release process would look like this:

1. Publish assets to public bucket / ECR
2. Publish nested SAR apps
3. Delete Lambda source from S3 (no longer needed)

Or, if we don't use SAR, we can stop after the first step. Essentially, doing half-and-half SAR and public bucket doesn't add any value.

The work we've done so far has not gone to waste - we still needed custom resources for all of the mage code. And we can still move to SAR in the future with a bit more work if and when we want to. But for now, it's simpler to just package in a public S3 bucket.

## Changes

- Add `deployments/master.yml` with all main stacks except glue (being removed in #949 ), onboarding, and web (the latter two coming in a future PR)
- Move auto-generated dashboard stack into `deployments` (instead of hiding in out/) so we can enforce linting/formatting like all other stacks
- Remove SAR headers from stacks

## Testing

`mage deploy` (to make sure the dashboards still deployed correctly)

```
sam package -t deployments/master.yml --output-template-file pkg.master.yml --s3-bucket y-bucket

# 'aws cloudformation deploy' would also work
sam deploy -t pkg.master.yml --stack-name panther
```


deployments/master.yml
```diff
@@ -0,0 +1,225 @@
...
+  SAR:
+    Constants:
+      Version: 1.4.0
+
+Resources:
+  Bootstrap:
+    Type: AWS::Serverless::Application
+    Properties:
+      Location:
+        ApplicationId: arn:aws:serverlessrepo:us-east-1:349240696275:applications/panther-community-bootstrap
```
austinbyers, line 82: The accountID is Panther's public account (the app has not been published there yet, but will be)

I'm hardcoding the region for now, in a future PR I'll figure out how to handle deployments in other regions (does SAR support global apps? can we interpolate a region here? etc)

deployments/master.yml
```diff
@@ -0,0 +1,225 @@
...
+        Debug: !Ref Debug
+        EnableS3AccessLogs: !Ref EnableS3AccessLogs
+        LogSubscriptionPrincipals: !Join [',', !Ref LogSubscriptionPrincipals]
+        TracingMode: !Ref TracingMode
+      Tags:
+        Application: Panther
+        # SAM will automatically apply the following tags:
+        #   lambda:createdBy = SAM
+        #   serverlessrepo:applicationId
+        #   serverlessrepo:semanticVersion
```
austinbyers, line 98: `mage deploy` adds tags for Panther edition (community), version (1.3.0) and stack name (panther-bootstrap)

All of that information is automatically included when deploying a SAR app as a nested stack like this, so I've elided our own custom tags.

tools/mage/cfngen.go
```diff
@@ -44,10 +46,44 @@ func generateGlueTables() error {
...
+				"Name":   "panther-community-dashboards",
+				"Author": "Panther Labs",
+				"Description": "Panther Community (nested app) - " +
+					"Detect threats with log data and improve cloud security posture",
+				"HomePageUrl":   "https://runpanther.io",
+				"SpdxLicenseId": "AGPL-3.0-or-later",
+				"SourceCodeUrl": "https://github.com/panther-labs/panther",
+			},
+		},
+	}
```
austinbyers, line 65: Changed dashboard template generation from json to yml and added the metadata section required for SAR publication

austinbyers:
Update: Lambda assets are indeed bundled with the SAR app itself - they can be removed after publication. When deployed, it turns into:

```yaml
  CustomResourceFunction:
    Properties:
      CodeUri:
        Bucket: awsserverlessrepo-changesets-plntc6bfnfj
        Key: 111122223333/arn:aws:serverlessrepo:us-east-1:111122223333:applications-panther-community-bootstrap-gateway-versions-1.4.0/d3360920-ff3b-440c-b87d-632215f032cc
```

However, other assets (like GraphQL schema) remain as their original S3 URL and whatever bucket they are published to must remain accessible to SAR.

This means not everything is self-contained in the SAR app like I thought, but it also means that it's possible to "bundle" everything into a SAR app, even if it's not natively supported! So the onboard and web stacks will just be CloudFormation templates, included like so (next PR):

```yaml
  LogAnalysis:
    Type: AWS::Serverless::Application
    Properties:
      Location: web_server.yml
      Parameters:
        AlarmTopicArn: !GetAtt Bootstrap.Outputs.AlarmTopicArn
      Tags:
        Application: Panther
```

austinbyers:
> This means not everything is "self-contained" in the SAR app like I thought, but that's ok - we have to rely on public buckets to distribute the non-SAR templates anyway

And because of this, I'm questioning whether we actually need SAR at first. It seems confusing and problematic to have assets in a public S3 bucket _and_ in SAR. Since we can't package everything in SAR just yet, maybe just nested CloudFormation templates are sufficient.

Either way, our work so far was still necessary (we need custom resources for many things) and this PR will be almost identical with or without SAR (just replace `AWS::Serverless::Application` with `AWS::CloudFormation::Stack`)

I'll hold off on merging until I chat with @jacknagz about it

tools/mage/release.go
```diff
@@ -0,0 +1,145 @@
...
+`
+
+// Release Publish nested SAR apps
+func Release() {
+	version := os.Getenv("SEMVER")
+	if version == "" {
+		logger.Fatal("please set SEMVER env variable with the version you want to publish")
+	}
+
+	awsSession, err := session.NewSession(&aws.Config{Region: aws.String("us-east-1")})
```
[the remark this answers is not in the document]
austinbyers: This was going to be fixed in a subsequent PR (TODO: publish to multiple regions), but after dropping SAR, this is no longer relevant (code has been removed)

deployments/master.yml
austinbyers: After moving away from SAR, we do indeed need the explicit tags again. Tags applied to the master stack would propagate, but we can't always assume users will deploy the master stack with a specific tag set

tools/mage/release.go
```diff
@@ -0,0 +1,145 @@
...
+            },
+            "Action": "s3:GetObject",
+            "Resource": "arn:aws:s3:::%s/*"
+        }
+    ]
+}
+`
+
+// Release Publish nested SAR apps
+func Release() {
```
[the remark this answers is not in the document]
austinbyers: Yes, this would only be used by the Panther team. Removing this for now since dropping SAR makes it easier.

I will add this command back in a future PR. It will automate as much of the release process as possible:

* Updating any version numbers in the codebase
* Publishing assets and the master stack to our public S3 bucket
* Building and publishing the pre-built docker image
* Publishing aux templates to our public S3 bucket

deployments/master.yml
```diff
@@ -0,0 +1,225 @@
...
+      TimeoutInMinutes: 20
```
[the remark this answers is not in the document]
austinbyers: This is actually a native CloudFormation setting (not SAR-specific). I was mostly worried about custom resources never returning, but it turns out CFN has its own timeout for that.

And in that case, I agree timeouts for each stack are more likely to cause problems than add value. Removing

austinbyers:
> I think migration would really require importing 9 Panther DDB tables, 2 S3 buckets and Glue database. I am not sure we need to migrate anything else?

I was assuming we'd want to keep log processing running so that deployments don't lose any data. If you're willing to stop log processing and delete all of the queues, you're right we wouldn't have to migrate much else. But we still have to be careful about the order of operations, my point was simply that it would take awhile to develop and test a smooth migration. It's certainly doable, just not trivial. Outside the scope of pre-packaged deployments, but I would like to get there at some point.

> Anyone developing panther locally would use mage and mage deploy command that uses the "old" parent stack. mage deploy uploads the nested templates to "bootstrap" S3 bucket, it DOESN'T publish them in SAR.

Correct, our current `mage deploy` isn't going to change at all.

> Anyone not interested into developing in Panther, they will just go to SAR UI directly.
Is my understanding correct? 

Correct, although now we're going to drop SAR and just rely on a public S3 bucket (see updated PR description for more details). They'll just use the master template directly either from the CloudFormation console or embedded as a nested stack or any other CFN deployment method.

> Can you give some more information about what are the differences in the two approaches?

`mage deploy` packages assets to the "source" bucket, then deploys each stack manually. Parallelism and stack parameters are handled explicitly by mage, and top-level parameters are managed with a config file.

Deploying the master stack from our public bucket means assets are pre-built and CloudFormation manages the entire deploy. No mage, no config file

You can see the difference when adding new stack parameters. Take #952 , for example. @nhakmiller added a new bootstrap parameter and passed it in with mage. Then I had to update this PR to also include the new parameter in the master stack.

The disadvantage of this split is the possibility for inconsistency - a deploy from source may not necessarily work when deploying the master template if we forgot to add the right parameters. We can use CI to help mitigate this problem (related to #864) and otherwise our release process just has to test for this

That said, being able to deploy our current stacks individually has advantages for local development. @3nvi and @alexmylonas, for example, often need to deploy only the appsync stack. Forcing all deployments to always use the master stack does complicate and slow down developer iteration.

So for now, we're supporting both of these deploy flows. They both use the same core templates (bootstrap, appsync, core, cloud_security, etc), but deploy-from-source deploys those stacks directly while deply-from-master-template wraps those templates in a single master stack

rleighton:
Are you saying the for existing deployments we require a teardown/rebuild?

austinbyers:
@nhakmiller @kostaspap @rleighton I updated the PR and description - we've decided to drop SAR for now. The master stack is still essentially the same though, please take another look and let me know if it makes sense

austinbyers:
> Are you saying the for existing deployments we require a teardown/rebuild?

Nope, I'm saying existing deployments will continue to `mage deploy` from source as they always have. The question is whether and how to migrate them in the future to use the top-level master stack. New deployments will use the master template for simplicity. 

@kostaspap was suggesting a migration where we teardown everything except data stores, and then redeploy with the data stores imported into the new stacks. Some kind of migration is eventually possible, but not our priority right now.

PR #952, opened 2020-05-27
Added support for backing up processed data

nhakmiller:
## Background

This change adds support for deploying the `processed-data` bucket with data replication. Currently the data replication is for same account GLACIER storage class data replication. This is the use case we have for this feature internally and seems like the most common use case, but could be expanded to cross account or different status data replication in the future if needed.

I considered making more parameters configureable, such storage class or cross account or even a prefix list (if for example you only want to replicate some of the processed data), but I decided against over complicating the feature and polluting the config file and bootstrap stack parameters for a feature that will likely only be used lightly.

Question to reviewers: should we replicate the data in other buckets? I thought about perhaps the analysis versions or audit logs buckets, but the analysis versions bucket is already a backup of data in dynamo and the audit logs buckets are generally ingested by Panther, and so end up (in a normalized form) in the processed data bucket and get replicated eventually anyways.

## Changes

- Updated bootstrap.yml to accept a `DataReplicationBucket` parameter, and if supplied appropriately configure the `processed-data` bucket to replicate data there
- Updated the deploy process to add the `DataReplicationBucket` parameter to `panther_config`

## Testing

- `mage test:ci`
- Deployed into my dev account and verified that data replication was working

### DIFF
--- a/deployments/bootstrap.yml
+++ b/deployments/bootstrap.yml
@@ -51,6 +51,10 @@ Parameters:
     Type: String
     Description: Optional external bucket for storing S3 access logs. If not specified, the Panther audit bucket is used.
     Default: ''
+  DataReplicationBucket:
+    Type: String
+    Description: Optional external bucket for replicating processed data in s3 glacier.
+    Default: ''
   CloudWatchLogRetentionDays:
     Type: Number
     Description: CloudWatch log retention period
@@ -140,6 +144,7 @@ Conditions:
   ConfigureLogSubscriptions: !Not [!Equals [!Select [0, !Ref LogSubscriptionPrincipals], '']]
   EnableAccessLogs: !Equals [!Ref EnableS3AccessLogs, true]
   ExternalAccessLogs: !Not [!Equals [!Ref AccessLogsBucket, '']]
+  ReplicateData: !Not [!Equals [!Ref DataReplicationBucket, '']]
   TracingEnabled: !Not [!Equals [!Ref TracingMode, '']]
   UseCustomDomain: !Not [!Equals [!Ref CustomDomain, '']]
   EnableDebug: !Equals [!Ref Debug, true]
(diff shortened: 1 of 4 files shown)


deployments/bootstrap.yml
```diff
@@ -352,6 +357,16 @@ Resources:
         - DestinationBucketName: !If [ExternalAccessLogs, !Ref AccessLogsBucket, !Ref AuditLogs]
           LogFilePrefix: !Sub panther-processed-data-${AWS::AccountId}-${AWS::Region}/
         - !Ref AWS::NoValue
+      ReplicationConfiguration: !If
+        - ReplicateData
+        - Role: !Sub arn:aws:iam::${AWS::AccountId}:role/panther-data-replication-role
```
nhakmiller, line 362: Can't `!Ref` the role resource here as that would cause a circular dependency.

deployments/bootstrap.yml
```diff
@@ -352,6 +357,16 @@ Resources:
...
+          Rules:
+            - Destination:
+                Bucket: !Ref DataReplicationBucket
+                StorageClass: GLACIER
+              Prefix: ''
```
nhakmiller, line 367: Required field, even if empty.

deployments/bootstrap.yml
```diff
@@ -51,6 +51,10 @@ Parameters:
     Type: String
     Description: Optional external bucket for storing S3 access logs. If not specified, the Panther audit bucket is used.
     Default: ''
+  DataReplicationBucket:
+    Type: String
+    Description: Optional external bucket for replicating processed data in s3 glacier.
```
austinbyers, line 56: This bucket has to be in the same account? If so, please mention that in the description

austinbyers:
> Question to reviewers: should we replicate the data in other buckets? 

I think it's fine to just replicate the processed data bucket, you make good points about the other buckets

nhakmiller:
> If it's relatively easy to support cross-account replication, I would strongly suggest adding it.

Looking into it again this morning, I think as long as the users properly setup the permissions on the other account's bucket to allow data replication this should work fine for that.

Issue #640, opened 2020-04-02
Editing the Source refers to a non existing stack - Log Analysis


alexmylonas:
### Describe the bug

As user I want to edit the initial Source, when I am prompted to upload a template the stack displayed name doesn't exist on AWS.

Check screenshot

### Steps to reproduce

1. Go to `Sources` pages on `Log Analysis`
2. Click on `Edit` the initial Source
3. Make a small change i.e name
4. Click `Next`
5. You are instructed to upload a template to a stack that doesn't exist

### Expected behavior

I am expecting for the instructions to refer me to an existing stack that I can update

### Environment

How are you deploying or using Panther?

- Panther version or commit: 5c032b0c582a598d6dad9a86eef52dc2ef687857

### Screenshots
![Screenshot 2020-04-02 at 3.16.24 PM.png](https://images.zenhubusercontent.com/5e4fb724a541ae524a4864cc/d616cfa7-c10c-4024-aed3-bd2dd92f8802)

alexmylonas:
The issue is described mostly from the frontend perspective, feel free to add more info for the backend affecting this! @kostaspap @austinbyers @nhakmiller

nhakmiller:
This is because we follow a different methodology for naming the first "self onboarding" stack than we follow for future onboarding stacks.

We could either update this so it displays the correct name for the self-onboarded stack, change the self-onboarded stack to follow the same pattern as future onboarded stacks, or perhaps disallow editing of the self onboard stack.

Displaying the correct name is difficult first of all because it's randomly generated, and second of all because then we would have to have some way of differentiating the self onboard stack from the rest.

Changing the self-onboarded stack to follow the same pattern as the others would require us to change how we do the self-onboarding flow (e.g. not do it as a nested stack), but I don't think that would be too difficult and removes any other special logic.

Disallowing editing the self-onboarding stack seems like a not very good idea, but it may be the simplest to do.

austinbyers:
The [onboarding stack](https://github.com/panther-labs/panther/blob/master/deployments/onboard.yml) contains two nested stacks with IAM roles, which is why we don't know their names

I believe @rleighton and/or @kostaspap suggested we just deploy those nested stacks directly, in which case their stack names would be known. That is a fairly easy change, although we have to do it in a way that is compatible with existing v1 deployments

alexmylonas:
Is there an easy way to identify a stack as "self-onboarded" and then disable edit for that? Seems like the cleaner solution if I understand correclty.

[...]

PR #953, opened 2020-05-28
Add onboarding to master stack

austinbyers:
## Background

There was quite a bit of `mage` code surrounding the onboard stack, both before and after deploying it. These have all been moved to custom resources

Will likely conflict with #949 , I'll wait for that to merge first

## Changes

- Add 3 new custom resources for the onboarding stack
    - `Custom::SelfLogConfig` - S3/GuardDuty settings which aren't supported in CloudFormation
    - `Custom::SelfRegistration` - Hits the source-api to register Panther cloud security / log analysis as sources
    - `Custom::StackSet` - Deploy/Teardown a stack set with a single instance in the current region. This is used for real-time cloud security events, but is generic enough to have other potential applications in the future

## Testing

- Deploy/teardown with mage
- Deploy/teardown with master template
- Toggled OnboardSelf, EnableGuardDuty, etc


internal/core/custom_resources/resources/self_registration.go
```diff
@@ -0,0 +1,161 @@
...
+			registerCloudSec = false
+		}
+		if aws.StringValue(integration.AWSAccountID) == props.AccountID &&
+			*integration.IntegrationType == models.IntegrationTypeAWS3 &&
+			*integration.IntegrationLabel == genLogProcessingLabel() {
+
+			zap.L().Info("account already registered for log processing",
+				zap.String("accountID", props.AccountID))
+			registerLogProcessing = false
+		}
```
kostaspap, line 95: This logic could cause issues in the following scenario: 
1. Someone deploys panther and enables the ALB and VPC FL. 
1. They decided to turn S3 Server Access on. They re-deploy Panther. 

This logic will not update the the log types of the integration, right?

deployments/onboard.yml
```diff
@@ -109,7 +113,7 @@ Resources:
   LogProcessingRole:
     Type: AWS::IAM::Role
     Properties:
-      RoleName: !Sub PantherLogProcessingRole-${LogProcessingRoleLabel}
+      RoleName: !Sub PantherLogProcessingRole-panther-account-${AWS::Region}
```
kostaspap, line 116: If you could take this role out of this template and create it in a new stack using `panther/deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml` with appropriate parameters, I think it would solve this long standing issue as well: https://github.com/panther-labs/panther/issues/640

Not sure if it's easy to do as part of this PR.

internal/core/custom_resources/resources/self_log_config.go
```diff
@@ -0,0 +1,104 @@
...
+	case cfn.RequestCreate, cfn.RequestUpdate:
+		var props SelfLogConfigProperties
+		if err := parseProperties(event.ResourceProperties, &props); err != nil {
+			return "", nil, err
+		}
+		return "custom:self-log-config:singleton", nil, configureLogProcessingUsingAPIs(props)
+
+	default:
+		// ignore deletes
+		return event.PhysicalResourceID, nil, nil
```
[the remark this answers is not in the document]
rleighton: Good point. I think today if you disable the self onboarding the registrations are not removed. The CF took care of the permissions and CF constructed entities, so as a practical matter it may not matter.

We need to preserve at least the same clean up as what was happening in CF. For example, un-registering GuardDuty.

internal/core/custom_resources/resources/self_registration.go
```diff
@@ -0,0 +1,161 @@
...
+	switch event.RequestType {
+	case cfn.RequestCreate, cfn.RequestUpdate:
+		var props SelfRegistrationProperties
+		if err := parseProperties(event.ResourceProperties, &props); err != nil {
+			return "", nil, err
+		}
+		return "custom:self-registration:singleton", nil, registerPantherAccount(props)
+
+	default:
+		// ignore deletes
```
rleighton, line 60: Per Kostas' note above, don't we need to delete the self integrations when they disable the self monitoring?

internal/core/custom_resources/resources/self_registration.go
rleighton: My PR handles that, I fixed that. Austin can copy that part into this.

internal/core/custom_resources/resources/stack_set.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		return event.PhysicalResourceID, nil, deleteStackSet(split[3], split[4])
+
+	default:
+		return "", nil, fmt.Errorf("unknown request type %s", event.RequestType)
+	}
+}
+
+// Returns once parent stack set has started creation (does not wait for it to finish).
```
rleighton, line 70: Is that what mage does today? How do we know about errors?

deployments/onboard.yml
austinbyers: Interesting. I don't fully understand what's needed for that issue, what would I use for the `RoleSuffix`? Can you show exactly how I should be instantiating that stack?

```yaml
# auxiliary/cloudformation/panther-log-anslysis-iam.yml
Mappings:
  # DO NOT EDIT PantherParameters section. Panther application relies on the exact format (including comments)
  # in order to replace the default values with an appropriate ones.
  PantherParameters:
    MasterAccountId:
      Value: '' # MasterAccountId
    RoleSuffix:
      Value: '' # RoleSuffix

Parameters:
  # Required parameters
  MasterAccountId:
    Type: String
    Default: ''
  RoleSuffix:
    Type: String
    Default: ''

Conditions:
  # Condition to define if the template is generated by panther backend
  IsGenerated: !Not [!Equals ['', !FindInMap [PantherParameters, MasterAccountId, Value]]]

Resources:
  LogProcessingRole:
    Type: AWS::IAM::Role
    Properties:
      RoleName: !If
        - IsGenerated
        - !Sub
          - 'PantherLogProcessingRole-${Suffix}'
          - Suffix: !FindInMap [PantherParameters, RoleSuffix, Value]
        - !Sub 'PantherLogProcessingRole-${RoleSuffix}'
```

If the IAM role will have the same name, then we'll have a conflict - CloudFormation will destroy this IAM role and create a new one in the nested stack with the same name at the same time, which will probably fail. We'd have to add a migration to `mage deploy` to delete the existing role before deploying the stacks, which will also temporarily break log processing I assume?

If the role name will be different, there's no conflicts and I'm happy to make that change now. If the role name is the same and we need a migration, I'll just add a comment and we'll do that in a different PR

austinbyers: The issue you cited (#640) mentions unpredictable/inconsistent stack names, how would this help solve that?

kostaspap: You can pass in the parameters to the CFN stack. 
- MasterAccountId: The account id where panther is deployed
- RoleSuffix : The `IntegrationLabel`, lowercased and with `-` instead of ` `. In this case `panther-account`.

You are right, to do the migration we should delete first the old role then create the new one. Or change the label (and `RoleSuffix` as a consequence) to something else e.g. `panther-account-onboard` and then delete the old integration (with label `panther-account`). If a role doesn't exist for a bit is fine - there will be no data loss, the log processor retries to pull the logs for a while (a couple of hours). Feel free to skip these changes and add a TODO or something if you think it's more appropriate. 

This template has a very very ugly fix to allow us to create/update templates both from UI and templates that are downloaded from the UI then deployed manually :(

internal/core/custom_resources/resources/self_log_config.go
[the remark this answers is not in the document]
austinbyers: > Is the "delete" going to be invoked when someone switches from from OnboardSelf: true to OnboardSelf: false?

When using the master stack, yes. The nested `onboard` stack would be automatically deleted, which would then delete all custom resources as well. 

> I think today if you disable the self onboarding the registrations are not removed.

Corect - when deploying from mage, we just say "if OnboardSelf, deploy(onboardStack)". The stack in that case would NOT be destroyed if OnboardSelf is toggled off, but now that's a fairly easy change I can make.

This is a great example of how using a custom resources makes our cleanup automatically better than it was before - today, `mage` only deletes onboarding resources during a teardown.

However, these steps do not yet have special cleanup:

* `Custom::SelfRegistration` - we could deregister Panther from the source-api
* `Custom::SelfLogConfig`
    * GuardDuty publishing destination - when the detector is destroyed, this presumably won't matter anyway
    * S3 bucket notification - we could remove the notification

I may actually split the `SelfLogConfig` into two separate resources, then, to better handle cleanup.

-----
In summary:

* Using a custom resource means our cleanup is already better than it used to be
* I'll fix `mage deploy` to teardown self-onboarding if `OnboardSelf=false`
* I'll add cleanup for the other custom resources (which we did not have before in `mage`)

internal/core/custom_resources/resources/self_registration.go
austinbyers: Sure, good idea. We don't have this in mage today, and I just ported what we had before, but now that's a relatively easy change I can make

deployments/onboard.yml
kostaspap: > The issue you cited (#640) mentions unpredictable/inconsistent stack names, how would this help solve that?

With the change above, the stack names will be with what we do normally during S3 onboarding!
Normally, the IAM role is deployed to its own stack and has a specific name. In case of the "onboard" stack, the IAM role was(is) nested

internal/core/custom_resources/resources/stack_set.go
austinbyers: Hmm good point. `mage` does wait for the stack set to finish creating and destroying.

I wasn't sure if it was a good idea to have the custom resource waiting while the stack sets finished, because of the timeout and because you maxed out the memory. If there was an error *starting* the stack set, the deploy would fail. But you're right that if there's an error finishing it, the deploy process would never know.

I'll go ahead and put the wait back in, but then I will also reduce the memory to, say, 1024 so we're not paying the maximum price when waiting for 10-120 seconds for every deploy + teardown

deployments/onboard.yml
austinbyers: Ok, if deleting the role and re-creating it is fine, that's fairly easy and I can probably add that.

> With the change above, the stack names will be with what we do normally during S3 onboarding!

With the introduction of the master stack, we don't have any control over the stack names. Someone can deploy that master template with any stack name they choose

internal/core/custom_resources/resources/stack_set.go
austinbyers: Ah, I was wrong - `mage deploy` today does not wait for the stack set instances to finish (but it does wait for teardown). I'll add it nonetheless

deployments/onboard.yml
austinbyers: More precisely, to answer this:

> If you could take this role out of this template and create it in a new stack 

We have to give people a single master template, we can't have some IAM roles created in a different top-level stack for packaged deployments

internal/core/custom_resources/resources/self_registration.go
austinbyers: Done

deployments/onboard.yml
austinbyers: I have switched the log role to use the aux template. Because the logicalID is the same, I actually don't think any special migration is necessary (but I will test)

internal/core/custom_resources/resources/stack_set.go
austinbyers: On second thought, I've actually removed the stackset altogether (see updated PR description / comment)

austinbyers:
@rleighton @kostaspap please take another look! Significant improvements since your last review:

- Better teardown/cleanup:
    - `mage deploy` now deletes the onboard stack if OnboardSelf=false
    - When `Custom::SelfRegistration` is deleted, it will remove the sources from source-api
    - `Custom::SelfLogConfig` has been split into its fundamental components: `Custom::S3BucketNotification` and `Custom::GuardDutyDestination`. This way, the custom resources are more generic, we can use the AWS Go structs as the custom resource properties, and there is a clean delete implementation for each
- Log processing role has been switched to use the aux template (instead of being defined directly in the onboarding stack)
- The real-time events stack set has been replaced with a simple nested template. Stack sets help with deploying across multiple accounts or regions, but the default self-onboarding does neither of these things - it uses a single stack instance in a single region. There are only 4 resources in the `cloudwatch-events` stack, all of which are safe to be deployed in multiple regions.
    - Trying to manage the stack set with a custom resource is possible, but way overkill for our needs. We have to manage the stack set instance and the parent and wait for both to finish - hundreds of lines of code and wasted custom-resource lambda cycles
- Moved migrations into a separate file (and added the stack set deletion as a new migration)

## Testing
- Deployed from mage to us-west-2
- Deployed from master template to us-east-1
- Upgrade us-east-2 1.3.0 deployment (to test migration)

deployments/onboard.yml
austinbyers: And I was wrong:

```
failed to create change set for stack panther-onboard: ValidationError: Update of resource type is not permitted. The new template modifies resource type of the following resources: [LogProcessingRole]
```

Migration added

tools/mage/migrations.go
```diff
@@ -0,0 +1,225 @@
...
+	// Removed stacks
+	alarmsStack            = "panther-cw-alarms"
+	metricFilterStack      = "panther-cw-metric-filters"
+	realTimeEventsStackSet = "panther-real-time-events"
+)
+
+// Migrations which run before the main deploy.
+//
+// These can be removed a few releases after they have been added.
+func migrate(awsSession *session.Session, accountID string) {
```
kostaspap, line 42: Nice! I was thinking that the migrations in the future could be split in chunks. E.g. the code would be checking the already deployed release (visible as a tag in the deployed CFN stacks) and running the appropriate migration required between the deploy and the latest available. A "migration" would be basically valid only for one updating from one version to the next. 

E.g. 
We can have something like:
```
type Migration interface{
  GetVersion() string
  Migration func() error
}
```
and a different migration per version. Or just `version_1_3_0.Migrate()` etc and the "main" migration code would be invoking the above in sequence. 


I know it's too early for this. This PR already improves a lot the current process, just thinking out loud how this could look in the future.

austinbyers: >  A "migration" would be basically valid only for one updating from one version to the next.

The problem with this approach is that developers need the migrations also. As soon as a breaking change lands in master, all developers immediately need the migration. That's why I prefer to check for the actual infrastructure markers rather than a specific version.

This method is also more robust if, say, an upgrade fails halfway through - some stacks might be v1.3.0 and other stacks might be v1.4.0, what then? It's not clear whether to run the migration.

Checking the infra the way we do now does not take more than a second or two and can be removed after a few releases

kostaspap: I see. Yeah indeed for such usecases this looks more robust. 

The usecases I had in mind were:
1. Managing migration long-term. E.g. not sure when is really a good time to drop some migration code (if ever for enterprise customers)
2. Migrations that are not dependent on infra e.g. running `glue:sync` equivalent code when needed, database migrations etc

internal/core/custom_resources/resources/self_registration.go
```diff
@@ -0,0 +1,275 @@
...
+		zap.String("accountID", props.AccountID))
+
+	// avoid alarms/errors and check first if the integrations exist
+	var listOutput []*models.SourceIntegration
+	var listInput = &models.LambdaInput{
+		ListIntegrations: &models.ListIntegrationsInput{},
+	}
+	if err := genericapi.Invoke(getLambdaClient(), "panther-source-api", listInput, &listOutput); err != nil {
+		return fmt.Errorf("error calling source-api to list integrations: %v", err)
+	}
```
nhakmiller, line 80: Is this code accomplishing anything? In the `getSelfIntegrations` call below you do the exact same thing anyways.

austinbyers: You're right! I moved it to the helper function and forgot to delete it here. Will do

PR #956, opened 2020-05-28
Fixing issue with url-encoded S3 object keys

kostaspap:
## Background

S3 Event notifications contain the S3 keys URL encoded. Trying to use URL encoded keys with the S3 client results to issues. 

## Changes

- URL decoding S3 object key. 

## Testing

- Unit test
- Ran in my account for few hours, verified it works fine.

### DIFF
--- a/internal/log_analysis/log_processor/sources/s3.go
+++ b/internal/log_analysis/log_processor/sources/s3.go
@@ -188,13 +188,15 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
 
 // ParseNotification parses a message received
 func ParseNotification(message string) ([]*S3ObjectInfo, error) {
-	s3Objects, err := tryParseCloudTrailNotification(message)
-	if err == nil {
+	s3Objects := parseCloudTrailNotification(message)
+
+	// If the input was not a CloudTrail notification, s3Objects will be nil
+	if s3Objects != nil {
 		return s3Objects, nil
 	}
 
-	s3Objects, err = tryParseS3Event(message)
-	if err == nil {
+	s3Objects = parseS3Event(message)
+	if s3Objects != nil {
 		return s3Objects, nil
 	}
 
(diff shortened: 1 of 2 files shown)


internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -236,9 +237,13 @@ func parseS3Event(message string) (result []*S3ObjectInfo) {
 	}
 
 	for _, record := range notification.Records {
+		urlDecodedKey, err := url.PathUnescape(record.S3.Object.Key)
+		if err != nil {
+			return nil
```
rleighton, line 242: nil?

kostaspap: Check again, I refactored this piece of code a bit.

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -187,15 +188,13 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
...
-	// If the input was not a CloudTrail notification, s3Objects will be empty slice
-	if len(s3Objects) > 0 {
+	s3Objects, err := parseCloudTrailNotification(message)
+	if err == nil {
 		return s3Objects, nil
 	}
 
-	s3Objects = parseS3Event(message)
-	if len(s3Objects) > 0 {
+	s3Objects, err = parseS3Event(message)
```
rleighton, line 196: I think we need to dispatch on the errors and decide which ones to return and when to ignore with lots of comments.

Look carefully, you added error handling in the called functions but it is all ignored in this function.

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -209,11 +208,15 @@ func ParseNotification(message string) ([]*S3ObjectInfo, error) {
 
 // parseCloudTrailNotification will try to parse input as if it was a CloudTrail notification
 // If the input was not a CloudTrail notification, it will return a empty slice
```
rleighton, line 210: That does not seem true any more.

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -209,11 +208,15 @@ func ParseNotification(message string) ([]*S3ObjectInfo, error) {
...
-func parseCloudTrailNotification(message string) (result []*S3ObjectInfo) {
+func parseCloudTrailNotification(message string) (result []*S3ObjectInfo, err error) {
 	cloudTrailNotification := &cloudTrailNotification{}
-	err := jsoniter.UnmarshalFromString(message, cloudTrailNotification)
+	err = jsoniter.UnmarshalFromString(message, cloudTrailNotification)
```
rleighton, line 213: I think we are conflating issues ... errors should be a real failure in processing , while discovering this is actually NOT a CT notification is a different case. Can we make this really clear in the code? Perhaps return a boolean? Previously we used the slice len.

internal/log_analysis/log_processor/sources/s3.go
kostaspap: The logic here is that we try parsing the input in different ways. We try to parse it as a CT notification, then S3 notification, etc. If none works, we throw an error. 

The only change I did is that while previous code was relying on `len(_)==0` fo decide if a notification is valid, I moved that code inside each individual function. Seems like a better design in Go - the caller can decide how to treat the error.

internal/log_analysis/log_processor/sources/s3.go
kostaspap: True, updating!

internal/log_analysis/log_processor/sources/s3.go
kostaspap: You are right, it's not clear. I think renaming the function would make it more clear? Take a look in the code now.

PR #959, opened 2020-05-29
Update Docker Development Image

### LINKED ISSUE #936: [Enhancement] Dockerfile for CI is using Debian oldstable instead of stable

Since Debian has released Buster (current stable, initial release July 6th, 2019) it is recommended to use this version instead of Stretch (oldstable, initial release June 17th, 2017).

To give a simple example, this is the current state of apt and in Sid/Buster/Stretch

**Package apt**
```
sid (unstable) (admin): commandline package manager
2.1.4: alpha amd64 arm64 armel armhf hppa i386 m68k mips64el mipsel ppc64 ppc64el riscv64 s390x sh4 sparc64 x32

buster (stable) (admin): commandline package manager
1.8.2.1 [security]: amd64 arm64 armel armhf i386 mips mips64el mipsel ppc64el s390x

stretch (oldstable) (admin): commandline package manager
1.4.10 [security]: amd64 arm64 armel armhf i386 mips mips64el mipsel ppc64el s390x
```
Since Debian stable is always stable by definition, there is no need to use oldstable in terms of stability. I don't know if you follow Debian closely but the stable versions already are way behind
the latest package versions. This is why they are super stable.

Tagging @austinbyers since he seems like the Maintainer based on commits.

### Describe your use cases

This can be used for deployments, CI and development.

This image can be very useful for development as it creates an identical local environment across teammates. As long as the docker user is using the current OS user(UID:GID) and not root (docker default user). This is not required for Docker on OSX but it's required on Linux.

Right now for example, all mage builds are being run by root and not by the actual linux user executing the command. So you end up with [node_modules, .setup, .terraform, out] paths owned by root:root.

Nice base images to consider
https://hub.docker.com/_/debian?tab=tags&page=1&name=buster

It's a good practice to use "slim" images so by adding only the necessary packages
you end up with a small image. Right now the image is quite large

```
$ docker images | grep pantherlabs
pantherlabs/panther-development-pack   1.1.2   9c2e0860d978   5 weeks ago   1.94GB
```

### References

You can check the Debian release lifecycle here.
https://www.debian.org/releases/

3nvi:
## Background

This PR updates the Docker image used by both our CI and the users that wish to deploy Panther. The main change is a switch to the newest stable release of Debian (codenamed `buster`) which includes security updates and benefits from greater maintainability.

While there, I also bumped:

- Golang to latest patch version
- NodeJS to LTS

Closes #936 

## Changes

- Remove un-needed debian packages
- Bump Languages
- Move to Debian 10
- Bump CI image
- Bump dev bash script
- Push images to dockerhub as `v1.2.0`

## Testing

- Managed to run `mage setup`
- Managed to run `mage deploy`
- CI checks pass

**I'd like to have one extra person validate proper deployment as well**


3nvi:
> If the AWS CLI is still installed, that could be removed if you want (not required for deploys).
> 
> I know you considered combining the dev and CI images, is that still something we want to do in the future or no?

Are we not dependant on AWS CLI anywhere anymore?

Yeah, we still want to do it, but we have the limitation of needing docker to build the images from source during development

austinbyers:
> Are we not dependant on AWS CLI anywhere anymore?

Correct. Ever since #742 (last month). Today we use `sam` instead, which is installed via pip by `mage setup`

PR #960, opened 2020-05-29
Custom resources: python analysis set, panther user/settings

austinbyers:
## Background

The last of the `mage deploy` API calls moved to custom resources! 

Closes: #850 

## Changes

- `Custom::AnalysisSet` handles the BulkUpload for the initial analysis pack (we assume there is a global present, enabled by https://github.com/panther-labs/panther-analysis/pull/61)
    - This also means we can remove the `helpers.py` from the policy-engine source
- `Custom::PantherUser` resource with full create/update/delete support. We only use this for creating the initial admin user, but it's interesting to note that Panther deployments could now manage all of their users with CloudFormation if they wanted!
    - Also adds support for defining the first user in the settings file so that you can start a deploy without having to worry about answering a terminal prompt
- `Custom::PantherSettings` updates "organization" settings (today only the display name and email, neither of which are actually used for anything)

- Bugfix: removed `appsync` permissions from `users-api` (it hasn't needed those for a long time)
- Bugfix: removed dead code from #953 
- Bugfix: `analysis-api` CreateGlobal endpoint updates the global layer
- Bugfix: The `Custom::S3BucketNotification` resource no longer fails when trying to delete a notification from an already deleted bucket

## Testing

- Deploy/teardown from mage
- Deploy/teardown from master template (in progress)
- Deploy with migration from previous version (in progress)


internal/core/custom_resources/resources/analysis_set.go
```diff
@@ -0,0 +1,157 @@
...
+
+		exists, err = lambdaFunctionExists("panther-rules-engine")
+		if err != nil {
+			return err
+		}
+		if !exists {
+			continue
+		}
+
+		return nil // both exist
```
nhakmiller, line 92: If the `panther-policy-engine` completes long before the `panther-rules-engine`, it will keep getting checked every loop. Probably not a big deal but a small thing.

austinbyers: True, though we'd have to be waiting anyway, so there's no extra cost, really

austinbyers:
Migration and deploy-from-master stack succeeded!

PR #962, opened 2020-05-30
Changing serde to org.apache.hive.hcatalog.data.JsonSerDe

kostaspap:
## Background

Fixing an issue where camelCase fields in a struct would be always null. 

## Changes

- Changed serde

## Testing

- Deployed. Verified that camelCase fields inside structs are now populated:

```
SELECT useridentity.principalid from panther_logs.aws_cloudtrail WHERE useridentity.principalid IS NOT NULL LIMIT 100
```

### DIFF
--- a/internal/log_analysis/awsglue/table_metadata.go
+++ b/internal/log_analysis/awsglue/table_metadata.go
@@ -179,7 +179,7 @@ func (gm *GlueTableMetadata) glueTableInput(bucketName string) *glue.TableInput 
 			InputFormat:  aws.String("org.apache.hadoop.mapred.TextInputFormat"),
 			OutputFormat: aws.String("org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat"),
 			SerdeInfo: &glue.SerDeInfo{
-				SerializationLibrary: aws.String("org.openx.data.jsonserde.JsonSerDe"),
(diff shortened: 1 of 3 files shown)


deployments/panther_config.yml
```diff
@@ -103,7 +103,7 @@ Setup:
 
   # Whether or not to create a CloudTrail in the Panther account and monitor.
   # You may want this off if you have org-level CT configured for Panther to avoid ingesting the same data twice.
-  EnableCloudTrail: false
```
nhakmiller, line 106: We don't want this enabled out of the box I think, CloudTrail is expensive and commonly enough enabled already by an organization trail.

internal/log_analysis/awsglue/table_metadata.go
```diff
@@ -180,7 +173,8 @@ func (gm *GlueTableMetadata) glueTableInput(bucketName string) *glue.TableInput
 			OutputFormat: aws.String("org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat"),
 			SerdeInfo: &glue.SerDeInfo{
 				SerializationLibrary: aws.String("org.openx.data.jsonserde.JsonSerDe"),
-				Parameters:           descriptorParameters,
+				// SerializationLibrary: aws.String("org.apache.hive.hcatalog.data.JsonSerDe"),
```
nhakmiller, line 176: What is this commented out line for?

PR #964, opened 2020-06-01
Fixing `null` values inside structs

kostaspap:
## Background

@nhakmiller  and @rleighton  identified an issue where a struct column would return always `null` for CT fields. 
This issue was a result of defining `case.insensitive=false` for the OpenX SerDe but not specifying the appropriate mappings for the struct field names. 

## Changes

- Adding mappings for struct field names. 

## Testing

- Unit tests
- Deployed and verified that nested fields now get populated appropriately

Screen Shot 2020-06-01 at 12 26 35 PM
Screen Shot 2020-06-01 at 12 40 11 PM


internal/log_analysis/awsglue/schema.go
```diff
@@ -134,8 +134,10 @@ type CustomMapping struct {
 	To   string       // glue type to emit
 }
 
-// Walk object, create columns using JSON Serde expected types, allow optional custom mappings
-func InferJSONColumns(obj interface{}, customMappings ...CustomMapping) (cols []Column) {
+// Walks the object creating Glue columns using JSON Serde expected types.
+// It returns a slice with the names of all nested fields.
```
rleighton, line 138: Can you elaborate on this comment and explain this is a global list for the table and why that is needed ... probably reference openx serde, since this change is specific to that.

internal/log_analysis/awsglue/schema.go
```diff
@@ -149,16 +151,30 @@ func InferJSONColumns(obj interface{}, customMappings ...CustomMapping) (cols []
...
 
-	return inferJSONColumns(objType, customMappingsTable)
+	cols, names := inferJSONColumns(objType, customMappingsTable)
+
+	// Removing duplicates
+	stringSet := map[string]struct{}{}
+	for _, name := range names {
+		stringSet[name] = struct{}{}
+	}
+	structFieldNames := make([]string, 0)
```
rleighton, line 161: `make([]string, 0, len(stringSet))`

internal/log_analysis/awsglue/schema.go
```diff
@@ -149,16 +151,30 @@ func InferJSONColumns(obj interface{}, customMappings ...CustomMapping) (cols []
...
+	for key := range stringSet {
+		structFieldNames = append(structFieldNames, key)
+	}
+
```
rleighton, line 165: Since this is not compute intesive code, I would add `sort.Strings(structFieldNames)` to both make it more readable and avoid the sorts in the tests.

internal/log_analysis/awsglue/schema.go
kostaspap: Done

internal/log_analysis/awsglue/schema.go
kostaspap: Done

internal/log_analysis/awsglue/schema.go
kostaspap: Added more comments

PR #968, opened 2020-06-01
Use client helper to get session for custom resource analysis set

rleighton:
## Before
The analysis set custom resource was not using the client helper get the session causing nil ptr exceptions.

## After
Use the helper

## Testing

- mage test:ci


austinbyers:
I'll also make a change to prevent this from happening again - either pre-creating the clients like we do elsewhere or moving the clients to a sub-package so the private variables are not accessible

PR #970, opened 2020-06-01
Add web stack to master template

austinbyers:
## Background

Add the web stack to the master template, thus completing it!

Closes: #851 

## Changes

- Add `web_server.yml` to the master template
    - The image repository defaults to the Panther public account, but can be overridden to a developer ECR when testing
    - The image tag is assumed to be the same as the Panther version
- Add `mage master` command to deploy the master template
    - This will only be used by Panther developers when testing the master template, usually as part of a release
    - I'll add a similar `mage release` command in a future PR which will build and publish the appropriate assets to our public S3 buckets / ECR repository
- Upgrade Go modules to the latest versions
- Clients are pre-created in the custom-resources Lambda (per @rleighton )
    - This prevents problems like the one fixed in #968 
    - Creating all of these clients takes only 150 microseconds on my laptop, so lazy creation turned out to be an unnecessary optimization
- `mage deploy` now errors if the installed version of node is different from the one in the web server - several developers are reporting consistent "Internal Server Errors" when they access a web app deployed in an environment with a different version of node
- Bugfix: User first/last name is actually required
- Bugfix: The S3 bucket notification in the onboard stack depends on the SNS topic policy

## Testing

- `mage master`


tools/mage/deploy.go
```diff
@@ -133,18 +133,27 @@ func deployPreCheck(awsRegion string) {
 		logger.Fatalf("go %s not supported, upgrade to 1.13+", version)
 	}
 
+	// Check the node version (must match the version in the web server)
+	const requiredNodeVersion = "v12.16" // web server has 12.16.1, to be exact
```
kostaspap, line 137: Latest is `12.17` and @3nvi  is setting it here: https://github.com/panther-labs/panther/pull/959 . You might want to coordinate on this.

tools/mage/deploy_frontend.go
```diff
@@ -111,7 +111,10 @@ func buildAndPushImageFromSource(awsSession *session.Session, imageRegistry stri
 	}
 
 	localImageID := strings.Replace(dockerBuildOutput, "sha256:", "", 1)
```
3nvi, line 113: Optional: Let's put L113 inside the `if` statement below

rleighton:
> > ## Background
> > Add the web stack to the master template, thus completing it!
> > Closes: #851
> > ## Changes
> > 
> > * Add `web_server.yml` to the master template
> >   
> >   * The image repository defaults to the Panther public account, but can be overridden to a developer ECR when testing
> >   * The image tag is assumed to be the same as the Panther version
> > * Add `mage master` command to deploy the master template
> >   
> >   * This will only be used by Panther developers when testing the master template, usually as part of a release
> >   * I'll add a similar `mage release` command in a future PR which will build and publish the appropriate assets to our public S3 buckets / ECR repository
> > * Upgrade Go modules to the latest versions
> > * Clients are pre-created in the custom-resources Lambda (per @rleighton )
> >   
> >   * This prevents problems like the one fixed in #968
> >   * Creating all of these clients takes only 150 microseconds on my laptop, so lazy creation turned out to be an unnecessary optimization
> > * `mage deploy` now errors if the installed version of node is different from the one in the web server - several developers are reporting consistent "Internal Server Errors" when they access a web app deployed in an environment with a different version of node
> > * Bugfix: User first/last name is actually required
> > * Bugfix: The S3 bucket notification in the onboard stack depends on the SNS topic policy
> > 
> > ## Testing
> > 
> > * `mage master`
> 
> Can you add more information about the difference between `mage master` and `mage deploy`? When you say it is meant to be used only during release, what's the difference of running `mage deploy` from master branch compared to running `mage master`?
> 
> Could we also update the name a bit, `master` seems a bit ambiguous imo... `mage release:deploy` if it is meant to be used only as part of the release process?

Maybe `mage master:deploy` or `mage master:package`

tools/mage/deploy.go
```diff
@@ -133,18 +133,27 @@ func deployPreCheck(awsRegion string) {
 		logger.Fatalf("go %s not supported, upgrade to 1.13+", version)
 	}
 
+	// Check the node version (must match the version in the web server)
```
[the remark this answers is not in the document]
rleighton: This is a good idea also just from the pov of "discovery". It may not be very obvious where to go to change these values but a well named file would be obvious.

tools/mage/master.go
```diff
@@ -0,0 +1,120 @@
...
+		logger.Fatal(err)
+	}
+
+	// Use the version from the master template for the docker image tag
+	cfn, err := cfnparse.ParseTemplate("deployments/master.yml")
+	if err != nil {
+		logger.Fatal(err)
+	}
+	type m = map[string]interface{}
+	version := cfn["Mappings"].(m)["Constants"].(m)["Panther"].(m)["Version"].(string)
```
rleighton, line 112: Check for version == "" and error?

tools/mage/deploy.go
austinbyers: This was when we thought the "internal server error" had to do with a mismatched node version, and for a moment I thought I had it working with a very specific node version.

But the problem turned out to be a bug in the web code, so I'll just reduce this to the major version (node12) like you suggest

tools/mage/deploy.go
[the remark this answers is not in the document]
austinbyers: I'm just going to reduce this check to node v12, @3nvi already fixed the underlying bug that was causing the "internal server error" I was experiencing. Then we won't need to update it until we make a big change at which point it's a quick fix.

Since this code will just be checking major versions, no need for a shared file - it's not the versions we will be installing. For example, we check only that "go > 1.12", but it could be go 1.13 or go 1.14, etc.

tools/mage/deploy_frontend.go
austinbyers: We still need it for the first argument of `docker tag` (line 119)

tools/mage/master.go
austinbyers: Sure, I can check for a blank version, good call.

If there's a type error, this will panic, which I'm fine with. It'd be very tedious to check the type of all 4 nested keys

austinbyers:
> Can you add more information about the difference between mage master and mage deploy? When you say it is meant to be used only during release, what's the difference of running mage deploy from master branch compared to running mage master?

* `mage master` - deploy `deployments/master.yml`
* `mage deploy` - deploy 9 top-level stacks directly as we always have

I will be updating our release notes and documentation as part of the next release, yes.

> Could we also update the name a bit, master seems a bit ambiguous imo... mage release:deploy if it is meant to be used only as part of the release process?

Perhaps I misspoke: it's just a different way of deploying. You can deploy one template (`mage master`), or you can deploy 9 (`mage deploy`). Both are technically valid ways of deploying from source, it's just `mage deploy` is a better developer experience.

What I meant as "part of the release process" was that it's an easy way to test the master template, which we could and should do at any time.

As for the name, I wanted to put it in the `deploy` namespace, e.g. `mage deploy:master`, but then we'd have to change the name of the normal deploy command. 

> Maybe mage master:deploy or mage master:package

Sure, I'll go with `master:deploy` for now

austinbyers:
>  I just want to verify that all these changes are only related to deploying from source.

Well, sort of. The master template is what we will give users in the next release - it is the pre-packaged deployment. The changes I've been making are to align our standard deploy-from-source model with the deploy-from-master-template model. They share the same code as much as possible

But you're right that `mage` won't even be needed for users to install Panther in the future. It's for us to package it

tools/mage/master.go
austinbyers: Ah, I see what you mean.

"Check for blank version, and then raise an error" - yes I will do that

I read this at first as "check for blank version and also check for type errors"

Issue #951, opened 2020-05-27
Teardown pre-packaged deployment


austinbyers:
### Description

Even with pre-packaged deployments, users will still need to run `mage teardown` to fully cleanup a Panther deployment (S3 buckets, leftover log groups, etc), at least for now.

Right now, `mage teardown` looks for specific stack names which will no longer be applicable. We need to fix teardown to work for either style of deployment

### Acceptance Criteria

- `mage teardown` works when deploying Panther from a single master stack, regardless of that stack's name
- `mage teardown` works as before for deployments from source

PR #971, opened 2020-06-02
Teardown supports master stack

austinbyers:
## Background

Moved much of `mage teardown` into a custom resource, and made sure the rest of teardown works both when deploying from source and when deploying the pre-packaged master stack.

Closes: #951 

## Changes

- New `Custom::PantherTeardown` resource which is only triggered on its own deletion
    - Handles removing lambda layers and ECR images
- `mage teardown` now checks for a STACK environment variable
    - When deploying master template, we have no control over the stack name. Rather than checking every stack and trying to guess, just require the user to specify it for safety
- When deleting log groups and S3 buckets, `mage teardown` relies on names and tags (respectively) instead of describing specific stack names
    - Again, when deploying via the master template, users can choose any stack name they want.
    - I wanted to delete leftover log groups with the custom resource, but its own log group is virtually guaranteed to remain regardless

With this change, after deploying the master template, people can just delete their main stack to clean up the vast majority of Panther's infrastructure. If they want to clean up their buckets and empty log groups, they can easily do so manually or run `mage teardown`

## Testing

- `STACK=panther-master mage teardown` (on a `mage master` deployment)


internal/core/custom_resources/resources/panther_teardown.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+const globalLayerName = "panther-engine-globals"
+
+type PantherTeardownProperties struct {
+	EcrRepoName string
+}
+
+func customPantherTeardown(_ context.Context, event cfn.Event) (string, map[string]interface{}, error) {
+	resourceID := "custom:panther-teardown:singleton"
```
rleighton, line 40: nit: const?

tools/mage/migrations.go
```diff
@@ -55,21 +55,20 @@ func migrate(awsSession *session.Session, accountID string) {
 	// In v1.4.0, the ECS cluster moved from the bootstrap stack to the web stack.
 	// Enumerate the bootstrap stack to see if this migration is necessary.
 	clusterInBootstrap := false
-	err := walkPantherStack(cfnClient, aws.String(bootstrapStack), func(r cfnResource) {
-		if aws.StringValue(r.Resource.ResourceType) == "AWS::ECS::Cluster" {
+	listStackResources(cfnClient, aws.String(bootstrapStack), func(r *cfn.StackResourceSummary) bool {
+		if aws.StringValue(r.ResourceType) == "AWS::ECS::Cluster" {
```
rleighton, line 59: nit: `*r.ResourceType` like you did below?

PR #982, opened 2020-06-03
fix: removes the pip upgrade warning when running `mage setup`

s0l0ist:
## Background

After running `mage setup`, there's a superfluous warning to upgrade `pip`.

image

## Changes

Specified `pip` with a version snapshot in the `requirements.txt` file.

## Testing

Ran `mage setup` results in no warning about a stale version of pip.


kostaspap:
Nice!

PR #985, opened 2020-06-04
v1.4.0 release candidate

austinbyers:
## Background

v1.4.0 release candidate

## Changes

- Added `mage master:publish` command to publish Panther assets to the public account as part of the release process
- Updated quick-start and development documentation with the new deploy process and mage commands, respectively

- bugfix: stackset migration - log a warning (instead of fatal) if the parent stack set cannot be deleted during the migration. There could be stack set instances in other regions; we only care about deleting the instance in the deployment region
- bugfix: In the master template, the web stack needs to wait for the custom-resources and the users-api to finish first
- bugfix: Moved the first user custom resource to the web stack so that it is created only after the web service is running. Otherwise, the invite email is sent too early (the link would be invalid)
- bugfix: Teardown was failing because the ECR repo was not empty. We deleted the ECR repo in a custom resource, but it apparently takes a bit to finish. So instead, the custom resource now *empties* the ECR repo rather than deleting it
- bugfix: Empty analysis sets no longer raise an error (empty strings are ignored in the custom resource)

## Testing

- deploy (v1.3.0), migrated to this branch, then teardown
- `mage deploy` and `mage teardown` from this branch
- `mage master:deploy` and `STACK=panther mage teardown`
- `mage master:publish` and deployed directly from the CloudFormation console in my dev account


docs/gitbook/development.md
```diff
@@ -14,38 +14,43 @@ For a more detailed architecture diagram, see the bottom of this page.
...
 
-### Dependencies
+For the remaining dependencies, you can either use our development image or install development dependencies locally.
 
-Install [Go](https://golang.org/doc/install#install) 1.13+, [Node](https://nodejs.org/en/download/) 10+, and [Python](https://www.python.org/downloads/) 3.7+.
+### Development Image
+This is the easier option, but will also lead to much slower builds.
 
-For MacOS w/ homebrew:
+Simply export your AWS credentials as environment variables, and then run `./dev.sh`
```
jacknagz, line 24: We had a question in OSS about this process. So let's give a full example in a code block below.

docs/gitbook/development.md
```diff
@@ -63,33 +68,16 @@ Targets:
   doc                 Auto-generate specific sections of documentation
   fmt                 Format source files
   glue:sync           Sync glue table partitions after schema change
-  glue:update         Updates the panther-glue cloudformation template (used for schema migrations)
+  master:deploy       Deploy single master template (deployments/master.yml) nesting all other stacks
+  master:publish      Publish a new version of Panther to the public account
```
nhakmiller, line 72: Should we make a note that users won't be able to use this command? Or can they use it to publish to their own app?

docs/gitbook/development.md
```diff
@@ -106,6 +94,63 @@ Since the majority of Panther is written in Go, the repo follows the standard [G
 | [**tools**](https://github.com/panther-labs/panther/tree/master/tools)  | Magefile source and other build infrastructure  |
 | [**web**](https://github.com/panther-labs/panther/tree/master/web)   | Source for the Panther web application  |
 
+## Testing
+
+1. Run our test suite: `mage test:ci`
+2. Run integration tests against a live deployment: `mage test:integration`
+   - **WARNING**: integration tests will erase all Panther data stores
```
jacknagz, line 101: {% hint style="danger" %}
Integration tests will erase all Panther data stores
{% endhint %}

We can probably remove the ordered list too

docs/gitbook/development.md
```diff
@@ -106,6 +94,63 @@ Since the majority of Panther is written in Go, the repo follows the standard [G
...
+
+### From an EC2 Instance
+
+You can also deploy from an EC2 instance with Docker and git installed in the same region you're deploying Panther to.
+Instead of exporting your AWS credentials as environment variables, you will need to attach the [deployment IAM role](quick-start.md#prerequisites) to your EC2 instance profile.
+Your EC2 instance needs at least 1 vCPU and 2GB of memory; the cheapest suitable instance type is a `t2.small`.
+
+## Teardown
+Run `mage teardown` to remove all Panther infrastructure, including S3 buckets that are normally retained if just the CloudFormation stack were deleted.
+If you used a master stack (i.e. did *not* `mage deploy`), then `STACK=your-stack-name mage teardown`
```
jacknagz, line 144: Can we say 'cloudformation stack' instead of 'master stack'? It's a bit clearer.

austinbyers:
ECR teardown is still failing, even though the repo is correctly emptied. I might have to add a wait to that resource

docs/gitbook/development.md
austinbyers: You're right, it will only work for us, though it wouldn't be hard to extend it to any other customer (you just need an S3 bucket and an ECR repo)

I'll just add a note for now

docs/gitbook/development.md
austinbyers: Honestly, we could even remove the warning completely - `mage` already warns you about it and the user has to explicitly confirm

docs/gitbook/development.md
austinbyers: Hmm but there's always a cloudformation stack. If you `mage deploy`, you get 12 stacks, if you deploy from the pre-packaged template, you get 13 (one master stack + all the nested stacks)

I'll change to "if you did not deploy from source", maybe?

docs/gitbook/development.md
austinbyers: There is a full example of the AWS environment variables in its own section in the same file, I'll link to it

docs/gitbook/development.md
austinbyers: I wound up doing both - adding a comment in the Go source about the purpose of the command and also adding environment variables so users could publish their own internal versions if they wanted.

docs/gitbook/development.md
austinbyers: Never hurts to have the warning in more than one place, I went ahead and added it

docs/gitbook/development.md
austinbyers: I went with "if you have a single top-level Panther stack", since you can easily check that by looking at your stacks in the CloudFormation console

austinbyers:
Cross-region pre-packaged deploys are failing at the Python layer:

```
Error occurred while GetObject. S3 Error Code: PermanentRedirect. S3 Error Message: The bucket is in this region: us-east-1. Please use this region to retry the request (Service: AWSLambdaInternal; Status Code: 400; Error Code: InvalidParameterValueException; Request ID: 57a7cf92-ed31-4345-abc3-8280ec2433f8)
```

CloudFormation loads the nested templates just fine from another region, but apparently AWS Lambda is refusing to load the layer from a bucket in a different region. In that case, we may just have to package in multiple regions

austinbyers:
Everything looks good now, running tests one last time before merging

docs/gitbook/development.md
austinbyers: Strike that - now that we need multi-region packaging, this command now has to loop over every region, so I just made it clear this command is for Panther use

PR #986, opened 2020-06-04
New log type registry and log parser interface

alxarch:
## Background

To support custom log parsers it will be required to be able to register new log types at runtime.
This PR introduces a thread safe registry for log types and a declarative way for defining new log types.

## Changes

- Adds `logtypes` module to handle log type definition and registration.
- Adds `logtypes.Config` to define a log type in a declarative and extensible way.
- Adds `logtypes.Entry` interface to define the read-only API for registered log types.
- Adds `logtypes.Registry` which is a concurrent safe registry of `logtype.Entry` values.
- Adds `parsers.Result` which becomes the interface between parsers and the S3 destination.
- Adds `parsers.Interface` interface that will be the 'universal' parser interface to support custom log parsers in the future.
- Adds helpers to create 'adapters' for parsers using the old `parsers.Parser` interface.
- Uses the adapters to register all existing parsers to the new registry.
- Modifies destinations, processor and classifier to use the new `logtypes.Entry` and logtypes.Registry` types.
   
## Testing

- mage test:ci


alxarch:
The reason this PR is marked as draft is that I think we should discuss how to merge it. In the conversation yesterday @rleighton mentioned that he thinks we should port all the parsers to the new format and then merge. That would require to add the `pantherlog.Event` stuff here as well and port the parsers in this PR. I have some reservations on this for 3 reasons.
- The extend of the PR will once again be huge.
- If we get errors while testing E2E it would be very hard to pinpoint if the problem comes from the modifications in the parser or the modifications in the overall log processing pipeline.
-  It could mean that the merging of these changes could be delayed further blocking customlog stuff from progressing on the enterprise branch. 

My suggestion is to deploy and E2E test this PR in it's current state *without* modifying the parsers.
This way we know that if there are errors they are caused by the changes in the registry-processor-classifier-destination layer of the code (or maybe the adapter but it's straight forward the code there) and not in the parsers. We merge this PR and on the next one we introduce the `pantherlog.Event` stuff with 'real' parsers for AWS logs (since they cover a lot of cases and are on by default on a new deploy). After we merge that one the road is open to both port other parsers at separate PRs and start using the changes in the enterprise repo to implement custom logs.

internal/log_analysis/log_processor/pantherlog/pantherlog.go
```diff
@@ -0,0 +1,108 @@
...
+		u, err := url.Parse(t.ReferenceURL)
+		if err != nil {
+			return errors.Wrapf(err, "invalid reference URL for log type %q", t.Name)
+		}
+		switch u.Scheme {
+		case "http", "https":
+		default:
+			return errors.Wrapf(err, "invalid reference URL scheme %q for log type %q", u.Scheme, t.Name)
+		}
+	}
```
kostaspap, line 87: You could use validator package & tags for these checks.

internal/log_analysis/log_processor/classification/classifier.go
```diff
@@ -41,22 +41,31 @@ type ClassifierAPI interface {
...
+//type LogParser interface {
+//	pantherlog.LogParser
+//	markClassifier()
+//}
+//
+//// LogParserMarker can be embedded into pantherlog.LogParsers to mark the parser as a classifying parser.
+//type LogParserMarker struct{}
+//
+//func (m *LogParserMarker) markClassifier() {}
+
```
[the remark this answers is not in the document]
alxarch: yup

internal/log_analysis/log_processor/pantherlog/rowid.go
```diff
@@ -48,32 +48,19 @@ var (
...
-	ifName, addr := getHardwareInterface()
-	if ifName == "" { // should never happen ... but just in case
-		err := errors.Errorf("Could not find hardware interface, generating random addr for uuid prefix") // to get stacktrace
-		zap.L().Error(err.Error(), zap.Error(err))
+	nif, err := getHardwareInterface()
+	if err != nil { // should never happen ... but just in case
+		zap.L().Error("Could not find hardware interface, generating random addr for uuid prefix", zap.Error(err))
 		noise := make([]byte, nodeIDSize)
-		rand.Read(noise) // nolint (errcheck) , not checking error because there is noting else to do
+		_, _ = rand.Read(noise) // nolint (errcheck) , not checking error because there is noting else to do
```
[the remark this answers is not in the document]
alxarch: the linter complains in go land

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
```diff
@@ -28,12 +28,6 @@ import (
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/parsers/timestamp"
 )
 
-const TypeAccessCommon = `Apache.AccessCommon`
-
-const AccessCommonDesc = `Apache HTTP server access logs using the 'common' format
-
-Reference: https://httpd.apache.org/docs/current/logs.html#common`
-
```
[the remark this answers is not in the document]
alxarch: I think that having the package 'index' file (a go file with the same name as the package) handle all registration is much cleaner. All log types for a module are defined in one place so discrepancies in log type prefix, registration, missing docs etc are all handled in a single place instead of chasing one const here, one var there and one  function call at a different module. We minimize tha number for `init()` calls to 1 per log parser module. The `const` 'block' definition at the top of the file is a good starting point to see which parsers are defined in this module for larger modules and for new developers.

internal/log_analysis/log_processor/registry/registry.go
```diff
@@ -19,138 +19,36 @@ package registry
  */
```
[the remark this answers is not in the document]
alxarch: The 'third' module is required to not run into cyclical dependencies. The `pantherlog` module cannot `import` modules that use `pantherlog`. The `registry` module with the `_ "pkgpath/.../foologs"` can import both. It ensures that by going through this module all imported parsers are available.

internal/log_analysis/log_processor/pantherlog/pantherlog.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/panther-labs/panther/api/lambda/core/log_analysis/log_processor/models"
+	"github.com/panther-labs/panther/internal/log_analysis/awsglue"
+)
+
+const FieldPrefix = "p_"
+
+// LogType describes a log type.
+// It provides a method to create a new parser and a schema struct to derive tables from.
+// LogTypes can be grouped in a `Registry` to have an index of available log types.
+type LogType struct {
```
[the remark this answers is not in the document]
alxarch: The interface is not the way to go here. It required ugly stuff like `New() Parser` method to work and is not much extensible since to add a method you would need to add it to all parsers to satisfy the interface while a zero value field can be handled in code in a backwards compatible way. The whole point is to become declarative about the log types. I would prefer  `pantherlog.EventType`  to avoid the `logType` naming conflict and remove the `...log.Log...` duplication in the name.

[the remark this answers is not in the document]
alxarch: The parser interface is generic now with only one method `ParseLog(log string) ([]*pantherlog.Result, error)`. It can even return different event types from a single log line. The 'type' of the log event is part of the `pantherlog.Result` struct. The definition that describes the log type should be a struct with (mostly) public fields so it can be passed around to all modules that need to now that info.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -314,11 +312,17 @@ func (destination *S3Destination) sendSNSNotification(key string, buffer *s3Even
 	return err
 }
 
-func getS3ObjectKey(logType string, timestamp time.Time) string {
+func (destination *S3Destination) getS3ObjectKey(logType string, timestamp time.Time) (string, error) {
+	typ := destination.logTypes.Get(logType)
```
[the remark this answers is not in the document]
alxarch: With dynamic log types at runtime it could very well be the case it happens that a polling for new log types has updated the available types while an event is in the buffer waiting to be written.

internal/log_analysis/log_processor/pantherlog/pantherlog.go
alxarch: I thought about it but seemed way harder to achieve and require reading validator package docs to understand corner cases, too much trouble (how do you check for schema or allow `-` *and* a valid url?). Also keeping simple checks like this allows for better to understand error messages on the specific case. Using both validator tags *and* a custom check seemed like an even worse solution. Doing it this way involves 3 simple ifs with a clear uniformity in error messages.

internal/log_analysis/log_processor/destinations/s3.go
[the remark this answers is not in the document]
kostaspap: Got it. Yeah, for that case it makes sense.

internal/log_analysis/log_processor/pantherlog/pantherlog.go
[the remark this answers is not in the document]
kostaspap: Got it. `EventType` seems better for me

rleighton:
78 files! OMG ... starting reading now :)

internal/log_analysis/awsglue/schema.go
```diff
@@ -28,64 +28,38 @@ import (
...
+	GlueStringArrayType = "array<string>"
+	GlueBigintType      = "bigint"
 )
 
+// RegisterMapping should be run in an init() block to register a custom mapping for a value.
+func RegisterMapping(x interface{}, glueType string) {
+	// Get non-pointer value
+	from := reflect.Indirect(reflect.ValueOf(x)).Type()
+
+	GlueMappings = append(GlueMappings, CustomMapping{
```
rleighton, line 47: I think we need to dedup/error-check, just in case 2 parsers register the same or conflicting types

internal/log_analysis/log_processor/pantherlog/pantherlog.go
```diff
@@ -0,0 +1,108 @@
...
+	Name         string
```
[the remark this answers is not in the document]
alxarch: Yes it's the identifier for the <strike>`Log`</strike>`EventType`

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -79,14 +76,15 @@ func init() {
...
-func CreateS3Destination() Destination {
+func CreateS3Destination(logTypes *pantherlog.Registry) Destination {
 	return &S3Destination{
 		s3Uploader:          common.S3Uploader,
 		snsClient:           common.SnsClient,
 		s3Bucket:            common.Config.ProcessedDataBucket,
 		snsTopicArn:         common.Config.SnsTopicARN,
 		maxBufferedMemBytes: maxS3BufferMemUsageBytes(common.Config.AwsLambdaFunctionMemorySize),
 		maxDuration:         maxDuration,
+		logTypes:            logTypes,
```
rleighton, line 87: `logRegistry`? ... we use `logTypes` all over the place to mean a slice of names of logTypes.

internal/log_analysis/log_processor/pantherlog/pantherlog.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	data, err := jsoniter.Marshal(schema)
+	if err != nil {
+		return errors.Errorf("invalid schema struct for log type %q: %s", logType, err)
+	}
+	var fields map[string]interface{}
+	if err := jsoniter.Unmarshal(data, &fields); err != nil {
+		return errors.Errorf("invalid schema struct for log type %q: %s", logType, err)
+	}
+	// TODO: [parsers] Use reflect to check provided schema struct for required panther fields
```
rleighton, line 106: Should this be added to this PR?

internal/log_analysis/log_processor/pantherlog/registry.go
```diff
@@ -0,0 +1,132 @@
...
+func (r *Registry) Get(name string) *LogType {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return r.entries[name]
+}
+
+// LogTypes returns all available log types in a registry
+func (r *Registry) LogTypes() (logTypes []LogType) {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
```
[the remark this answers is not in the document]
alxarch: Append cannot be avoided when dumping map keys because the `for` is on string keys. Preallocation to `make([]LogType, 0, len(r.entries))`  and then append would work. TBH I think the best way to do this would be to allocate to `make([]LogType, 0, 16)` (or some other const minSize) *before* acquiring the lock to avoid delays.

internal/log_analysis/awsglue/schema.go
alxarch: This part of the PR was required to resolve some cyclic references issues and was done in a bit of haste. I saw you use the mappings through a map anyway so I thought that duplicates wouldn't hurt. I will revisit to see if I can turn it into a `map[reflect.Type]string`

internal/log_analysis/log_processor/pantherlog/pantherlog.go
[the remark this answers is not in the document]
alxarch: Renamed to `EventType`

internal/log_analysis/log_processor/destinations/s3.go
alxarch: Renamed to `eventTypes`

internal/log_analysis/log_processor/pantherlog/registry.go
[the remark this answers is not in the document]
alxarch: The conversation here made me rethink a bit the `pantherlog.Registry` interface. There is an issue with mutability outside a lock on the returned `*EventType`. I changed the `LogTypes()` method to return a `[]string` of log type names to be more compatible with the current usage of the term 'LogType'. I am also considering defining an interface for event types and a separate `EventOptions` struct for registration so the read-only aspect of a registered event type is enforced by the type system.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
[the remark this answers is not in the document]
kostaspap: > All log types for a module are defined in one place so discrepancies in log type prefix, registration, missing docs etc are all handled in a single place instead of chasing one const here, one var there

I disagree. When a developer works on a debugging a parser or works on developing a new parser, it is much better that the references to e.g. log format are in the same file as the parser. If someone wants to see all the parsers in a module, they can just look at the files in the directory. Personally with this change, I find it harder to understand what a parser does when looking in the parser code. In the same way we improve readability by co-locating the field description with the field name and type, the same I see it for parser code, type, description and reference. 

In any case this is a readability issue, so my thoughts above are subjective. I will defer to @rleighton for a tie-breaker.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -79,14 +77,15 @@ func init() {
 	memUsedAtStartupMB = (int)(memStats.Sys/(bytesPerMB)) + 1
 }
 
-func CreateS3Destination() Destination {
+func CreateS3Destination(logTypes *parsers.Registry) Destination {
```
kostaspap, line 80: Renamed `logTypes` to `registry`

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
```diff
@@ -102,10 +94,10 @@ func (p *AuditLogParser) Parse(log string) ([]*parsers.PantherLog, error) {
...
-	ResourceName       *string             `json:"resourceName,omitempty" description:"The resource or collection that is the target of the operation. The name is a scheme-less URI, not including the API service name."`
-	NumResponseItems   *numerics.Int64     `json:"numResponseItems,omitempty" description:"The number of items returned from a List or Query API method, if applicable."`
+	ServiceName        *string             `json:"serviceName,omitempty" description:"The name of the LogTypeEntry service performing the operation"`
+	MethodName         *string             `json:"methodName,omitempty" description:"The name of the service method or operation. For LogTypeEntry calls, this should be the name of the LogTypeEntry method."`
+	ResourceName       *string             `json:"resourceName,omitempty" description:"The resource or collection that is the target of the operation. The name is a scheme-less URI, not including the LogTypeEntry service name."`
+	NumResponseItems   *numerics.Int64     `json:"numResponseItems,omitempty" description:"The number of items returned from a List or Query LogTypeEntry method, if applicable."`
```
kostaspap, line 100: These descriptions shouldn't be modified, right? IDE mass renaming? :)

internal/log_analysis/log_processor/parsers/pantherlog.go
```diff
@@ -62,6 +66,10 @@ type PantherAnyString struct { // needed to declare as struct (rather than map)
 	set map[string]struct{} // map is used for uniqueness, serializes as JSON list
 }
 
+func init() {
+	awsglue.RegisterMapping(PantherAnyString{}, awsglue.GlueStringArrayType)
+}
```
kostaspap, line 71: I am not so sure I like this approach tbh. This increases the coupling between Parser code and AWS Glue code. If anything, we want the parsers code to be decoupled from Glue (and any other system that we used for storing/querying/indexing data). Parsers should just take an input (string, Reader, etc) and returns a JSON output. 
Having a centralized mapping can help keeping the AWS Glue logic inside the AWS Glue modules. I think a more centralized approach is going play out better long term, keeping into consideration that Glue mappings are only "one type" of mappings we need. Depending on the engine we use for querying this data, we will need more mappings e.g. we already know we need mapping for Snowflake types.

internal/log_analysis/log_processor/parsers/parsers.go
```diff
@@ -19,17 +19,20 @@ package parsers
...
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog"
 )
 
 // LogParser represents a parser for a supported log type
+// NOTE: We will be transitioning parsers to the `pantherlog.LogParser` interface.
+// Until all parsers are converted to the new interface the `AdapterFactory()` helper should be used
+// when registering a `pantherlog.EventTypeEntry` that uses this interface.
 type LogParser interface {
-	// LogType returns the log type supported by this parser
+	// EventTypeEntry returns the log type supported by this parser
```
kostaspap, line 35: Should stay `LogType`

internal/log_analysis/log_processor/parsers/registry.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+func checkLogEntrySchema(logType string, schema interface{}) error {
+	if schema == nil {
+		return errors.Errorf("nil schema for log type %q", logType)
+	}
+	data, err := jsoniter.Marshal(schema)
+	if err != nil {
+		return errors.Errorf("invalid schema struct for log type %q: %s", logType, err)
+	}
```
kostaspap, line 110: This code should be using the field-name re-writing logic (`@` -> `_at_sign_`, etc)

internal/log_analysis/log_processor/parsers/timestamp/timestamp.go
```diff
@@ -41,6 +43,15 @@ const (
...
 )
 
+func init() {
+	awsglue.RegisterMapping(RFC3339{}, awsglue.GlueTimestampType)
+	awsglue.RegisterMapping(ANSICwithTZ{}, awsglue.GlueTimestampType)
+	awsglue.RegisterMapping(UnixMillisecond{}, awsglue.GlueTimestampType)
+	awsglue.RegisterMapping(UnixFloat{}, awsglue.GlueTimestampType)
+	awsglue.RegisterMapping(FluentdTimestamp{}, awsglue.GlueTimestampType)
+	awsglue.RegisterMapping(SuricataTimestamp{}, awsglue.GlueTimestampType)
+}
```
kostaspap, line 53: See comment above.

internal/log_analysis/log_processor/parsers/registry.go
```diff
@@ -0,0 +1,290 @@
...
+		DefaultRegistry().MustRegister(entry)
+	}
+}
+
+// LogTypeEntry describes a registered log event type.
+// It provides a method to create a new parser and a schema struct to derive tables from.
+// Entries can be grouped in a `Registry` to have an index of available log types.
+type LogTypeEntry interface {
+	Describe() Desc
+	NewParser(params interface{}) pantherlog.LogParser
```
kostaspap, line 64: `interface{}` seems.... to broad. What type of parameters are you expecting to pass? Would be better to have a concrete type if possible.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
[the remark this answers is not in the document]
rleighton: Hmmm, I see both points. The most important thing for me is a logical structure and consistency, either approach can have these properties. One advantage of the package level organization (rather than per-parser) is in cases where there is common code, then the package level organization makes it more clear to view the code holistically rather than as a set of independent parsers. This is happening more and more in the recent work.  I really do not feel strongly about this.  Given that, and the work done so far I vote to just leave the current organization in the PR using a package level theme. If it becomes an issue we can easily refactor since this is simply an organization of the code (not the actual code), so not a 1-way door.

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
rleighton: Good catch, that's a mistake.

internal/log_analysis/log_processor/parsers/pantherlog.go
rleighton: Currently I derive the Snowflake mappings from the glue mappings, but your point still stands. I think we should centralize the glue mappings, this also helps with my previous comment about panic'ing if the same mapping is registered two times ... these mappings are often shared between parsers so having a central registry makes sense. The doc generation and deployment code should fail hard (panic) if a new type is used and not registered, so we should be safe in that way.

internal/log_analysis/log_processor/parsers/apachelogs/access_common.go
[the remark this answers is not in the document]
kostaspap: Fair point. In the near future we will be defining parsers using a different format anyway, and this discussion might not be relevant then. Let's leave it as is in this PR.

internal/log_analysis/log_processor/parsers/pantherlog.go
alxarch: I didn't want this to be part of this PR either. The thing is that there is a cyclic dependency between the parsers and the awsqlue modules that needs to be solved. When registering a log type we need to generate the `GlueTableMetadata` so we need the `awsglue` module. But if the `awsglue` module needs the `parsers.AnyString` to register the custom mapping it's a cyclic dependency. The `RegisterMapping` was introduced to resolve this. This used to not be a problem because the types where registered in the `registry` module and all things that required *both* `awsglue` and `parsers` lived there. If we want to avoid the cyclic dependency a new module that defines the dynamic registry and the log type entry in the registry needs to be added to be the point where the two modules can co-exist without cyclic deps.

internal/log_analysis/log_processor/pantherlog/registry.go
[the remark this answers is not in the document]
alxarch: The PR was refactored to enforce the mutability rules for the log type registry.
I decided to declare the new types at the `parsers` module and keep the `pantherlog` module for all code related to a single log event. The lineage of code is clearer this way as things that supersede the old parser interface are kept in the same package.

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
alxarch: Yep, the way GoLand does the refactor is a bit weird. Will try to be more careful in the future.

internal/log_analysis/awsglue/schema.go
alxarch: Added panic in the event of duplicate 'from'.

internal/log_analysis/log_processor/parsers/parsers.go
alxarch: Fixed

rleighton:
This isn't really a Draft PR, right? Shouldn't it be promoted to a real PR?

internal/log_analysis/log_processor/parsers/registry.go
alxarch: We can't have a concrete type because every parser would be defining their own params. I was thinking of `map[string]interface{}` but this is not better either. If we want to restrict the format for parser parameters I would suggest to use `url.Values` and the UI should encode the parameters as a URL query or using the stricter `map[string]string` that would not allow for multiple values for a parameter.

alxarch: It is important to have the parameter passing resolved with this PR even if we just pass `nil` for now as it would be a PITA to retrofit this in the registry.

alxarch:
> This isn't really a Draft PR, right? Shouldn't it be promoted to a real PR?
I just did a last change in the organization of the PR based on feedback. I undraft this latest version.

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -167,10 +171,14 @@ type Processor struct {
 	operation  *oplog.Operation
 }
 
-func NewProcessor(input *common.DataStream) *Processor {
+func NewProcessor(input *common.DataStream, logTypes ...pantherlog.LogType) *Processor {
+	if logTypes == nil {
+		logTypes = pantherlog.AvailableLogTypes()
+	}
+	cls := classification.NewClassifier(logTypes...).(*classification.Classifier)
```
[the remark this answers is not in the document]
alxarch: changed, was left over from a different approach

internal/log_analysis/log_processor/parsers/pantherlog.go
rleighton: In my experience, if you are trying to work around cyclic dependencies, that means you should refactor (rather than fight it).  I was actually forced to make such a package for the same reasons recently. It is called `gluetables` in `log_analysis` and the important deps are:
```
        "github.com/panther-labs/panther/internal/log_analysis/awsglue"
	"github.com/panther-labs/panther/internal/log_analysis/log_processor/registry"
```
Perhaps you put the code here. If you want to change the name to something else besides `gluetables`, feel free.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -128,14 +126,15 @@ type S3Destination struct {
 	// thresholds for ejection
 	maxBufferedMemBytes uint64 // max will hold in buffers before ejection
 	maxDuration         time.Duration
+	logTypes            *pantherlog.Registry
```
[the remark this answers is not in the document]
alxarch: renamed to `registry`

internal/log_analysis/awsglue/schema.go
alxarch: Removed the registering all-together by defining the log type registry in a separate module and avoiding the cyclic dependency while limiting the changes in this PR

internal/log_analysis/log_processor/destinations/s3.go
alxarch: renamed to `registry`

internal/log_analysis/log_processor/pantherlog/pantherlog.go
alxarch: No, it's not needed

internal/log_analysis/log_processor/destinations/s3.go
alxarch: done

internal/log_analysis/log_processor/parsers/registry.go
alxarch: This code would always fail if a field was renamed on output since it wouldn't be able to unmarshal the JSON. The intention of the code is to check that the value provided serialized properly to a JSON object so it can be used to concatenate obejects. The field rewritting JSON instance would be useful if we could somehow verify that the schema produces a proper glue table without issues.

internal/log_analysis/log_processor/parsers/timestamp/timestamp.go
alxarch: The custom mapping was removed altogether and a new module was used to avoid the cyclic dependency

rleighton:
This looks very close. I noticed a few comments still unresolved. It would be very good to get this merged soon so we have the rest of the sprint to burn-in.

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
rleighton: Still needs fixing

internal/log_analysis/awsglue/schema.go
rleighton: Where in the code?

internal/log_analysis/log_processor/parsers/timestamp/timestamp.go
rleighton: Where in the code?

internal/log_analysis/log_processor/parsers/pantherlog.go
alxarch: I followed your advice about cyclic dependencies and used a different module name for the registry and log type entry types so that awsglue and parsers can co-exist without cyclic deps

internal/log_analysis/log_processor/parsers/gcplogs/auditlog.go
alxarch: Fixed

internal/log_analysis/awsglue/schema.go
alxarch: The GlueMappings is left untouched in the last version of the PR. I resolved the cyclic dependency issue by introducing a separate module for the registry and log type entries.

alxarch: The new module is under `github.com/panther-labs/panther/internal/log_analysis/logtypes`

internal/log_analysis/log_processor/parsers/timestamp/timestamp.go
alxarch: The new module lives at `github.com/panther-labs/panther/internal/log_analysis/logtypes`

internal/log_analysis/log_processor/parsers/registry.go
alxarch: Any resolution on this? should we use `url.Values` (ie `map[string][]string`) or `map[string]string`?

kostaspap: I imagine that if we go with `interface{}` we would define the config as a struct, while if we go with `map[string]string` or similar, everything has to be serialized. Do you have an example of what information will be contained in a parser config?

rleighton: One example is a timestamp format.

internal/log_analysis/log_processor/classification/classifier_test.go
```diff
@@ -20,91 +20,56 @@ package classification
...
+	}.Parser()
+	parserFail1 := testutil.ParserConfig{
+		logLine: errors.New("fail1"),
+	}.Parser()
+	parserFail2 := testutil.ParserConfig{
+		logLine: errors.New("fail2"),
+	}.Parser()
+	//logTypeSuccess := mockLogType("success", parserSuccess)
+	//logTypeFail1 := mockLogType("failure1", parserFail1)
+	//logTypeFail2 := mockLogType("failure2", parserFail2)
```
kostaspap, line 54: Can remove

internal/log_analysis/log_processor/parsers/parsers.go
kostaspap: Still here ^^

internal/log_analysis/logtypes/logtypes.go
```diff
@@ -0,0 +1,45 @@
+package logtypes
```
kostaspap, line 1: We generally follow the structure : `/internal/<product: log_analysis>/<lambda name>/packages`. Any reason you moved this one out? Could be placed under the log_processor folder.

internal/log_analysis/logtypes/logtypes.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+	}
+	return nil
+}
+
+// Register registers log type entries to the package wide registry panicking if an error occurs
+func MustRegister(entries ...Config) {
+	for _, entry := range entries {
+		// nolint:errcheck
+		DefaultRegistry().MustRegister(entry)
```
kostaspap, line 43: Seems you should be checking for errors and panic if one occurs.

internal/log_analysis/logtypes/registry.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func (r *Registry) Del(entry Entry) bool {
+	if entry == nil {
+		return false
+	}
+	name := entry.Describe().Name
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if e, ok := r.entries[name]; ok && e == entry {
```
kostaspap, line 92: In what scenario can it be that the  `e== entry` check fails?

internal/log_analysis/logtypes/registry.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	newEntry := newLogEventType(config.Describe(), config.Schema, config.NewParser)
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if oldEntry, duplicate := r.entries[newEntry.Name]; duplicate {
+		return oldEntry, errors.Errorf("duplicate log type config %q", newEntry.Name)
+	}
+	if r.entries == nil {
+		r.entries = make(map[string]Entry)
+	}
```
kostaspap, line 111: This should be above the previous statement.

internal/log_analysis/logtypes/registry.go
alxarch: The user tries to remove an entry from another Registry instance. If we want to remove by LogType  I can change the signature to `Del(logType string)`

internal/log_analysis/logtypes/registry.go
alxarch: Reads on nil map return the zero value without panics. 
The order doesn't matter. I put it second so it's closer to the assignment (which would panic if entries was nil).

internal/log_analysis/logtypes/logtypes.go
alxarch: I will move it under log_processor.
My reasoning was that the `logtypes` module contains stuff that is used by sibling modules in the `log_analysis` module. Concerning package organization in general I think that the current structure splits a feature in multiple places. Grouping by lambda is not reasonable since types defined in a module might be used by other parts of the code. I would suggest using a topic/feature hierarchy under `internal` (ie `log_analysis/[logtypes, awsglue, log_processor]/...` and use a separate hierarchy at the top level of the project under `lambdas/<lambda-name>/*.go` that uses  `package main` and defines each lambda importing the appropriate modules. Global state for lambdas would be contained to these lambda modules and not pollute the code base with hard to test implementations of features.  That leads to more 'library' style of coding for a feature. So for example `boxlogs` can define both the Parser for box events and the API client for the puller. It would help with a clear separation of enterprise/oss features as well.

kostaspap: I see. The redesign you suggest is an interesting idea. I think doing a quick write up of the idea and distributing to the team would help take a decision on this topic.

internal/log_analysis/logtypes/registry.go
kostaspap: Imo, removing by log type makes sense since the log type should be uniquely identifying an entry.

internal/log_analysis/log_processor/parsers/registry.go
alxarch: After some consideration the `interface{}` seems to be the best option. When we implement the actual parameter parsing (need to add a new epic fot that) we can work out how this `interface{}` value will be read from the source config to be passed to instances of parsers.

internal/log_analysis/log_processor/parsers/parsers.go
alxarch: Fixed

internal/log_analysis/logtypes/registry.go
alxarch: Changed

internal/log_analysis/logtypes/logtypes.go
alxarch: Moved the logtypes module under log_processor

internal/log_analysis/log_processor/classification/classifier_test.go
alxarch: removed

internal/log_analysis/log_processor/parsers/pantherlog.go
```diff
@@ -244,3 +246,54 @@ func AppendAnyString(any *PantherAnyString, values ...string) {
 		any.set[v] = struct{}{} // new
 	}
 }
+
+// Result converts a PantherLog to Result
+// NOTE: Currently in this file to help with review
```
rleighton, line 251: Should this be moved?

rleighton:
When this merges, don't forget to do a `sync` with enterprise ASAP to get all the code there conformant with this change.

internal/log_analysis/log_processor/parsers/pantherlog.go
alxarch: I will remove the comment. The notion was that passing through `parsers.PantherLog` will not be needed after all parsers are ported to `parsers.Interface` without using an adapter.

PR #992, opened 2020-06-04
fix: web development environment for node v12

s0l0ist:
## Background

Running the local development web server using `npm start` was failing when running node **v12**.

Steps to reproduce:
1. `rm -rf node_modules`
2. `npm install`
3. `npm start`

image

## Changes

- Locked **pounce** until it is ready to update
- Locked **typescript** to the current minor version. Latest causes some breaking changes.
- Updated **package-lock.json** by removing and allowing npm to re-initialize it.
- Updated **go.mod**, **go.sum** as they were updated from a fresh setup.

## Testing

- Instal node v10, `rm -rf node_modules`, `npm install`, `npm start`
- Instal node v12, `rm -rf node_modules`, `npm install`, `npm start`
- Sanity check: `mage clean setup`, `mage test:ci`

From a clean environment, both node versions now load the local site as expected and there are no conflicts after subsequent building.


package.json
```diff
@@ -79,7 +79,7 @@
     "svgo-loader": "^2.2.1",
     "terser-webpack-plugin": "^2.3.4",
     "ts-loader": "^6.2.1",
-    "typescript": "^3.7.5",
+    "typescript": "~3.7.5",
```
3nvi, line 82: What issues did you face with  `3.9.x.`? We will eventually need to upgrade since it includes a lot of speed benefits and it compatible with our code.

s0l0ist: @3nvi  It will complain about incompatible types/signatures on the output destination (all). 

A simple test is to just remove package-lock.json, replace with `^` and `npm install`, `npm run validate` to see all the errors.

I was not prepared to take on that big of a change for this PR. Perhaps we can fix it in a separate one.

PR #994, opened 2020-06-05
Prefix version in web app with 'v'

austinbyers:
## Background

In #985 , the documentation link in the wev app doesn't work. I didn't notice this until now because I didn't push the docs branch until *after* finishing the release. We can either make a new tiny release or just bump the v1.4.0 tag forward with this change.

When deploying the pre-packaged template, the version passed to the web stack is '1.4.0' But that version is used for the docs link, which expects `v1.4.0`

## Changes

- Prefix Panther version with 'v' when passing to web stack

## Testing

- Deployed via master template, link works correctly

### DIFF
--- a/deployments/master.yml
+++ b/deployments/master.yml
@@ -352,7 +352,9 @@ Resources:
               - !Sub 349240696275.dkr.ecr.${AWS::Region}.amazonaws.com/panther-community
             tag: !FindInMap [Constants, Panther, Version]
         GraphQLApiEndpoint: !GetAtt Bootstrap.Outputs.GraphQLApiEndpoint
-        PantherVersion: !FindInMap [Constants, Panther, Version]
+        PantherVersion: !Sub
+          - v${version} # has to be prefixed with 'v' for docs link in web app to work
+          - version: !FindInMap [Constants, Panther, Version]
         SecurityGroup: !GetAtt Bootstrap.Outputs.WebSecurityGroup
         SubnetOneId: !GetAtt Bootstrap.Outputs.SubnetOneId
         SubnetTwoId: !GetAtt Bootstrap.Outputs.SubnetTwoId


3nvi:
I can also modify the code that creates the doc link. If everywhere else the version is `1.4.0` instead of `v.1.4.0`, I'd rather modify the JS code

nhakmiller:
> We can either make a new tiny release or just bump the v1.4.0 tag forward with this change.

I'd prefer to just bump the tag unless there is a specific concern with that approach.

kostaspap:
> I can also modify the code that creates the doc link. If everywhere else the version is `1.4.0` instead of `v.1.4.0`, I'd rather modify the JS code

Agreed, that seems like a better approach.

kostaspap:
> > We can either make a new tiny release or just bump the v1.4.0 tag forward with this change.
> 
> I'd prefer to just bump the tag unless there is a specific concern with that approach.

I think @3nvi suggestion was to do the concatenation `v{PantherVersion}` in the JS code, instead of doing it in the CFN template.

austinbyers:
> I think @3nvi suggestion was to do the concatenation v{PantherVersion} in the JS code, instead of doing it in the CFN template. Both would require a bump in the tag I believe.

But when deploying from source, the `v` is already there. So the frontend code would have to conditionally check for the prefix, and add it if it isn't there. That would be in some sense the most robust solution, but I'm not sure if we can merge that in time, I'll check with him

austinbyers:
In retrospect, I should have perhaps saved the version in the master template with the "v" prefix to match the GitHub tags. Ah well, we have a deploy later today so it's too late to change now

austinbyers:
The `v1.4.0` tag has been updated and the new release assets pushed

PR #998, opened 2020-06-05
Add glue:CreatePartition permission

austinbyers:
## Background

One of our deployment upgrades failed because the custom-resources Lambda was missing a `glue` permission

### DIFF
--- a/deployments/bootstrap_gateway.yml
+++ b/deployments/bootstrap_gateway.yml
@@ -161,6 +161,7 @@ Resources:
                 - glue:CreateTable
                 - glue:GetTable
                 - glue:UpdateTable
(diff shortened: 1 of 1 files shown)


austinbyers:
We don't have to move the release tag again, existing deployments can get around it by dropping the `panther-glue` stack and running `mage glue:sync` offline, I've updated the release notes

kostaspap:
> We don't have to move the release tag again, existing deployments can get around it by dropping the `panther-glue` stack and running `mage glue:sync` offline, I've updated the release notes

If users drop `panther-glue` stack, they will lose all their data. This cannot be a migration strategy @austinbyers @rleighton

PR #1000, opened 2020-06-06
Updating build commands

kostaspap:
## Background

`mage test:ci` is a pretty resource intensive command. Trying to make it a bit more lightweight (without sacrificing speed!)
* The `mage test:ci` command runs several commands in parallel. It seems though that some of these subcommands (`build:lambda`, `build:tools`) were themselves using up multiple processors. Updated these commands not spawn their own go routines. 
* The `mage build:api` command builds Go Clients for calling Swagger APIs and configures typescript. Now running commands in parallel.

## Changes

- Modified `mage build:lambda` so that it now compiles the Go binaries sequentially. I ran several tests locally, didn't see the this change affect the time it took for `mage test:ci` to complete. `mage build:lambda` now takes around ~9 seconds to complete in my laptop. 
- Modified `mage build:tools` to compile them sequentially. Also this change didn't seem to impact negatively the overall time it takes for the `mage test:ci ` to complete. The command takes ~5 seconds to finish in my laptop
- Code for both `mage build:lambda` and `mage build:tools` is much simpler since we don't have to deal with channels/go routines. 
- Split `api()` method to one command that generates Go SDK from Swagger files and another that deals with typescript. 
- Specifying `-p` parameter for `go build` and `go test` to value `1`. From `go help build`: 
```
-p n
		the number of programs, such as build commands or
		test binaries, that can be run in parallel.
		The default is the number of CPUs available.
```
- Go linting now is configured to run on 1 processor. 

## Testing

- `mage test:ci`
- Timed it by running `time (for i in {1..10}; do mage test:ci ; done)`

For master results were: `2511.40s user 1056.40s system 502% cpu 11:49.76 total` -> avg 71 seconds
For this  branch: `2536.54s user 987.71s system 526% cpu 11:09.33 total` -> avg 67 seconds (also using less system time)


austinbyers:
@kostaspap go unit tests are now taking *much* longer than the rest of the tests:

```
13:16:41	INFO	    √ terraform validate finished (10/14)
13:16:53	INFO	build:tools: compiling cmd/opstools/requeue/requeue/main.go for 5 os/arch combinations
13:16:57	INFO	    √ npm run eslint finished (11/14)
13:16:58	INFO	build:tools: compiling cmd/opstools/s3queue/s3queue/main.go for 5 os/arch combinations
13:17:03	INFO	    √ build:tools finished (12/14)
13:17:19	INFO	    √ golangci-lint finished (13/14)
13:19:47	INFO	    √ go unit tests finished (14/14)
```

Your timing tests were probably running with cached unit tests; the best way to test the timing is by clearing the go cache first. Same goes for building the lambda source; that may take quite a bit longer for a fresh clone with no go cache.

I don't think we need to force all unit tests to run sequentially, that's not going to scale well at all. Instead, we can run them in parallel by themselves before running the other tests in parallel. E.g. change `mage test:ci` to 

1. format source
1. build lambda source in parallel (optional, I don't have a strong preference here)
1. run go unit tests in parallel
1. trigger remaining tests in parallel

austinbyers:
The average CircleCI time has also increased by 30-90 seconds since this PR merged

PR #1003, opened 2020-06-07
Improve deployment glue sync to avoid timeouts and failed deployments

rleighton:
## Background
After the change away from using CF for table definitions, the custom resource added that is responsible to manage glue tables would sync the partitions of tables when there were schema changes. 

Unfortunately, this can take some time if there are many partitions and could result in a timeout and a failed deployment.

This PR changes the deployment glue sync process to use a non-blocking recursive lambda to avoid blocking the deployment and to avoid timeouts.

## Changes

- Add deadline to sync function
- Add event to datacatlog_updater lambda to run syncs

## Testing

- mage test:ci
- mage deploy


internal/log_analysis/datacatalog_updater/process/integration_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+func TestMain(m *testing.M) {
+	integrationTest = strings.ToLower(os.Getenv("INTEGRATION_TEST")) == "true"
+	if integrationTest {
+		Setup()
+	}
+	os.Exit(m.Run())
+}
+
+func TestIntegrationSynPartitions(t *testing.T) {
```
austinbyers, line 42: typo: `SyncPartitions`

PR #1004, opened 2020-06-08
Updated prettier version

kostaspap:
## Background

Updated prettier version. Seeing speed improvements. 

## Changes

- Version bump for prettier
- Seeing some changes in `*.tsx` files, apparently it is [expected](https://prettier.io/blog/2020/03/21/2.0.0.html#improved-method-chain-breaking-heuristic-6685httpsgithubcomprettierprettierpull6685-by-mmkalhttpsgithubcommmkal). 
- Prettier now runs much faster. `mage fmt` used to take ~17 seconds on average, now takes ~8 seconds

## Testing

- `mage fmt`
- `mage test:ci`

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,4 @@
 {
-  "trailingComma": "es5",
   "printWidth": 100,
   "singleQuote": true,
   "arrowParens": "avoid"
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
   "trailingComma": "es5",
   "printWidth": 100,
-  "singleQuote": true
+  "singleQuote": true,
+  "arrowParens": "avoid"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -12574,9 +12574,9 @@
       "dev": true
     },
     "prettier": {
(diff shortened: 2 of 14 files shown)


.prettierrc
```diff
@@ -1,5 +1,6 @@
 {
   "trailingComma": "es5",
   "printWidth": 100,
-  "singleQuote": true
+  "singleQuote": true,
+  "arrowParens": "avoid"
```
kostaspap, line 5: Had to add this to maintain old behavior. See: https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev

.prettierrc
```diff
@@ -1,5 +1,5 @@
 {
-  "trailingComma": "es5",
```
kostaspap, line 2: This is now the default: https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-trailingcomma-to-es5-6963httpsgithubcomprettierprettierpull6963-by-fiskerhttpsgithubcomfisker

PR #1019, opened 2020-06-09
Small build/deploy fixes

austinbyers:
## Background

A few small improvements revealed after our first few v1.4.0 deployments

## Changes

- Change the data replication IAM role name to be regional (otherwise, Panther deployments in different regions will fail with replication enabled). Ideally, the IAM role shouldn't be named at all, but there's a circular dependency between the S3 bucket resource and the IAM role resource - one of them must have a deterministic name
- More human-readable master stack parameter descriptions
- Remove timestamp check when compiling Lambda functions. It turns out `sam package` is much smarter than `aws cloudformation package` and we no longer need this
- Change output of `mage master:publish` to show the actual S3 URL instead of a made-up version. This way, the URL can be used to directly verify the uploaded template

## Testing

- `mage test:ci deploy`
- `rm -r out`
- `mage deploy` - nothing new was packaged, even though all binaries had been removed


deployments/master.yml
```diff
@@ -66,44 +66,44 @@ Parameters:
...
     Default: ''
   InitialAnalysisPackUrls:
     Type: CommaDelimitedList
-    Description: Initial Python analysis packs
+    Description: Comma-separated list of Python analysis pack URLs installed on the first deployment
     Default: https://github.com/panther-labs/panther-analysis/releases/latest/download/panther-analysis-all.zip
   LayerVersionArns:
     Type: CommaDelimitedList
-    Description: List of LayerVersion ARNs to attach to each Lambda function
+    Description: Comma-separated list of LayerVersion ARNs to attach to each Lambda function (e.g. if you have a serverless monitoring service)
```
nhakmiller, line 92: Not sure if it's worth mentioning, but users can specify at most three layers here. There is a cap of 5 layers per lambda, and we use two of them already.

deployments/master.yml
```diff
@@ -66,44 +66,44 @@ Parameters:
...
     Default: ''
   LogProcessorLambdaMemorySize:
     Type: Number
-    Description: Log processor Lambda memory allocation
+    Description: Log processor Lambda memory allocation. Higher values reduce processing time, but come at a higher cost
     MinValue: 256 # any smaller and we risk OOMs
```
nhakmiller, line 97: Perhaps move this warning up to the description, out of memory errors are important to be wary of.

PR #1020, opened 2020-06-10
Run go tests in multiple processors

kostaspap:
## Background

Speeding up `mage test:ci`. 

## Changes

- Running `go test` in multiple processors
- @alxarch mentioned me that some unit tests have significant delay since they are testing retry logic. Many of them can run in parallel and save us some more time. 

## Testing

- `mage test:ci`
- TODO: proper timing tests with cache enabled/disabled. `CircleCI` build times seem to vary greatly, local testing (by enabling/disable the cache) will be more appropriate.


tools/mage/test_namespace.go
```diff
@@ -57,17 +58,20 @@ var (
 func (Test) CI() {
 	// Formatting modifies files (and may generate new ones), so we need to run this first
 	fmtErr := testFmtAndGeneratedFiles()
+	// Run it serially since it runs itself in multiple processors
```
austinbyers, line 61: 👍

PR #1027, opened 2020-06-10
Support single-stack updates in mage

austinbyers:
## Background

@3nvi and the frontend team need a way to rapidly deploy just the appsync stack. Their workflow looks like this:

1. Deploy appsync change
1. Play with web app locally
1. Repeat several dozen times

Using `mage deploy` every time takes at least 75 seconds with a fast Internet connection, which many of our developers do not have. Single-stack updates are not just useful for the frontend team, backend devs are typically only modifying one stack at a time, they too could benefit from being able to deploy only their changes.

This PR makes it possible to update one stack at a time: `STACK=appsync mage deploy` takes only ~5 seconds (if nothing changed). There are also several small documentation and mage bug fixes

## Changes

- `mage deploy` now supports single-stack updates with the `STACK` env variable
- All `mage` code now shares a global `awsSession` variable. This greatly simplified 90% of function signatures - we don't need to pass it to every function if it's the same everywhere
- Remove `cfnparse.init()` path hack
- Hide Python unit test stderr - new developers/users (e.g. @s0l0ist) see "WARNING" and "ERROR" in the logs and often assume the tests were failing (but it was expected log output). Now 'mage test:ci' has much cleaner output
- bugfix: include capabilities in quick-start documentation

## Testing

- `mage deploy`
- `rm -r out && STACK=appsync mage deploy` (tested with all 9 Panther stacks)
- `mage test:ci` - output is cleaner


3nvi:
Thanks so much for that!

tools/mage/deploy.go
```diff
@@ -96,29 +95,28 @@ var supportedRegions = map[string]bool{
...
-	setFirstUser(awsSession, settings)
 
-	identity, err := sts.New(awsSession).GetCallerIdentity(&sts.GetCallerIdentityInput{})
-	if err != nil {
-		logger.Fatalf("failed to get caller identity: %v", err)
+	if stack := os.Getenv("STACK"); stack != "" {
+		stack = strings.ToLower(strings.TrimSpace(stack))
+		if !strings.HasPrefix(stack, "panther-") {
+			stack = "panther-" + stack
+		}
```
rleighton, line 105: else error?

austinbyers: No error, this is just supporting either `STACK=panther-appsync` or `STACK=appsync`. `mage` is adding the panther prefix if the user didn't already

Issue #1024, opened 2020-06-10
ListAlerts filtering


s0l0ist:
### Describe the ideal solution

The current backend API needs to support more advanced **filtering** queries on data returned from the db.

The focus is only on **filtering** for this issue.

### Describe your use cases

These changes facilitate a frontend request for **filtering** alerts to provide better usability for the `Log Analysis` -> `Alerts` page.

### How frequently would you use such feature

Weekly

### Describe alternatives you have considered

- Duplicate tables with the pk as the filtered ID
- Filtering on frontend

### References

N/A

### Additional context

N/A

Issue #1015, opened 2020-06-09
ListAlerts page changes


3nvi:
### Background
The current `Log Analysis` -> `Alerts` page shows the alerts that have been triggered, but does not provide the same usability found elsewhere in the application such as some of the sort/filtering options the `Log Analysis` -> `Rules` page contains.

**Note**: filtering additions should be implemented first as sorting has its own challenges.

Filtering capability:
- [x] Severity Level (single level)
- [ ] Title (name contains)
- [ ] Alert ID (multiple IDs)
- [ ] Events (count range)
- [ ] Created At (date range)
- [ ] Last Matched At (date range)

Sorting capability:
- [ ] Title
- [ ] Severity Level
- [ ] Alert ID
- [ ] Events count
- [ ] Created At
- [ ] Last Matched At

PR #1029, opened 2020-06-11
ListAlerts filtering, sorting

s0l0ist:
## Background

closes: #1024 #1029 #1015 

Adding in both backend and frontend changes to allow filtering options and sorting for the `Log Analysis` -> `Alerts` page.

Sorting (and direction) will only occur on the `Created At` field, see the below RFC for more details.

## RFC

Displaying time will always be in the user's local time, but the URL parameters will be in UTC for cross-collaboration.

Sorting is relevant for only the `Created At` time field for due to the nature of displaying time-series events. Therefore, we should only allow sorting by the `Created At` field which would result in two benefits:

1. Presents data in the context the user is expecting.
2. No additional `GSI` or complex sorting logic in the backend.

Ex:
image

## Changes

- [x] Filter by `Severity` level
- [x] Filter by `Title` contains
- [x] Filter by `Event` count range (min, max)
- [x] Filter by a `RuleID`
- [x] Filter by a `AlertID`
- [x] Filter by `CreatedAt` date-range (before, after)
- [x] Sort direction on `CreatedAt`

## Testing

- [x] Filter by `Severity` level
- [x] Filter by `Title` contains
- [x] Filter by `Event` count range (min, max)
- [x] Filter by a `RuleID`
- [x] Filter by a `AlertID`
- [x] Filter by `CreatedAt` date-range (before, after)
- [x] Sort direction on `CreatedAt`


api/graphql/schema.graphql
```diff
@@ -82,8 +82,16 @@ input GetS3LogIntegrationTemplateInput {
 
 input ListAlertsInput {
   ruleId: ID
-  pageSize: Int
+  pageSize: Int # defaults to `25`
   exclusiveStartKey: String
+  severity: SeverityEnum
+  contains: String # matches against title
```
austinbyers, line 88: Then let's rename to `titleContains`

api/lambda/alerts/models/api.go
```diff
@@ -55,13 +55,33 @@ type GetAlertOutput = Alert
...
+
+	// Infinite scroll/pagination query key
 	ExclusiveStartKey *string `json:"exclusiveStartKey,omitempty"`
+
+	// Filtering
+	Severity        *string    `json:"severity,omitempty" validate:"omitempty,oneof=INFO LOW MEDIUM HIGH CRITICAL"`
+	Contains        *string    `json:"contains,omitempty"`
+	EventCountMin   *int       `json:"eventCountMin,omitempty" validate:"omitempty,min=0"`
+	EventCountMax   *int       `json:"eventCountMax,omitempty" validate:"omitempty,min=0"`
+	CreatedAtBefore *time.Time `json:"createdAtBefore,omitempty" validate:"omitempty"`
```
austinbyers, line 80: `validate:"omitempty"` is the same as not having a validate tag at all, I think you can remove it

api/lambda/alerts/models/api.go
```diff
@@ -55,13 +55,33 @@ type GetAlertOutput = Alert
...
+//         "eventCountMax": "500",
+//         "createdAtAfter": "2020-06-17T15:49:40Z",
+//         "createdAtBefore": "2020-06-17T15:49:40Z",
+//         "sortDir": "ascending",
 //     }
 // }
 type ListAlertsInput struct {
-	RuleID            *string `json:"ruleId,omitempty"`
-	PageSize          *int    `json:"pageSize,omitempty"  validate:"omitempty,min=1,max=50"`
+	RuleID   *string `json:"ruleId,omitempty"`
```
austinbyers, line 69: The `json:"omitempty"` tags are only used for serialization (it tells `json` not to marshal the field if it has a default value). Since this is only an input struct, it has no effect.

Not sure why @kostaspap added them but I'd vote to remove them to avoid confusion

`validate:"omitempty"` is completely separate and those can remain

api/lambda/alerts/models/api.go
```diff
@@ -55,13 +55,33 @@ type GetAlertOutput = Alert
...
+	PageSize *int    `json:"pageSize,omitempty" validate:"omitempty,min=1,max=50"`
+
+	// Infinite scroll/pagination query key
 	ExclusiveStartKey *string `json:"exclusiveStartKey,omitempty"`
+
+	// Filtering
+	Severity        *string    `json:"severity,omitempty" validate:"omitempty,oneof=INFO LOW MEDIUM HIGH CRITICAL"`
+	Contains        *string    `json:"contains,omitempty"`
+	EventCountMin   *int       `json:"eventCountMin,omitempty" validate:"omitempty,min=0"`
+	EventCountMax   *int       `json:"eventCountMax,omitempty" validate:"omitempty,min=0"`
```
austinbyers, line 79: Nit: Having `min=0` for a max filter doesn't make sense because it would never match anything. Maybe `min=1`

internal/log_analysis/alerts_api/api/list_alerts.go
```diff
@@ -35,10 +35,12 @@ func (API) ListAlerts(input *models.ListAlertsInput) (result *models.ListAlertsO
...
 	var alertItems []*table.AlertItem
-	if input.RuleID != nil { // list per specific ruleId
-		alertItems, result.LastEvaluatedKey, err = alertsDB.ListByRule(*input.RuleID, input.ExclusiveStartKey, input.PageSize)
-	} else { // list all alerts time desc order
-		alertItems, result.LastEvaluatedKey, err = alertsDB.ListAll(input.ExclusiveStartKey, input.PageSize)
+	if input.RuleID != nil {
+		// list per specific ruleId
+		alertItems, result.LastEvaluatedKey, err = alertsDB.ListByRule(input)
+	} else {
+		// list all alerts time desc order
```
austinbyers, line 42: This comment is no longer accurate, correct? It's in time order, but could be ascending or descending

internal/log_analysis/alerts_api/api/list_alerts.go
```diff
@@ -35,10 +35,12 @@ func (API) ListAlerts(input *models.ListAlertsInput) (result *models.ListAlertsO
...
+		alertItems, result.LastEvaluatedKey, err = alertsDB.ListAll(input)
```
austinbyers, line 43: I would do some validation / standardizing on the input parameters before passing them to the Dynamo handler.

For example, do you want the `titleContains` field to be case-sensitive? Usually we do not, in which case you can `strings.ToLower` the filter before continuing.

Similarly, if the time range is missing or invalid, raise a `genericapi.InvalidInputError` here before calling `alertsDB`

The idea with the package separation is that the `api` package handles input processing and validation, then invokes the "real" handler which talks to AWS APIs

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -26,84 +26,254 @@ import (
...
+		keyBuilder = expression.Key(RuleIDKey)
+		return keyBuilder
+	}
+	keyBuilder = expression.Key(TimePartitionKey)
+	return keyBuilder
+}
+
+// getKeyCondition - gets the key condition for a query
+func (table *AlertsTable) getKeyCondition(keyBuilder *expression.KeyBuilder,
+	input *models.ListAlertsInput) (keyCondition expression.KeyConditionBuilder) {
```
austinbyers, line 72: Nit: You don't need a named return

```suggestion
	input *models.ListAlertsInput) expression.KeyConditionBuilder
```

Then you can simply `return keyBuilder.Equal(...)` in each case

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -26,84 +26,254 @@ import (
...
+			expression.GreaterThanEqual(expression.Name("eventCount"), expression.Value(*input.EventCountMin)),
+		)
+	}
+}
 
-	// queries require and = condition on primary key
-	keyCondition := expression.Key(ddbKey).Equal(expression.Value(&ddbValue))
+// applyFilters - adds filters onto an expression
+func (table *AlertsTable) applyFilters(builder *expression.Builder, input *models.ListAlertsInput) {
+	// Start with an empty filter for a known attribute
```
austinbyers, line 176: 👍  Nice! Yeah, this is one of the quirks with the expression builders, they don't work on empty variables but have to be initialized with something

internal/log_analysis/alerts_api/api/list_alerts.go
austinbyers: I just realized that we can't just lowercase the search filter because Dynamo's filtering is case-sensitive.

Our other APIs handle this by either doing the filtering after scanning Dynamo or adding a new column to the table, e.g. `lowercaseTitle`, and applying the filter to that. The second option is harder because you have to change all of the code that writes to the table and we'd have to add a migration to update existing deployments in the next release.

I'd recommend:

* We should implement case-*in*sensitive filtering (imagine how frustrated you would be if you had to search for an alert by "My HostName" instead of "my hostname")
* To implement, apply the title filter *after* scanning Dynamo. We have to pay for the same number of requests either way, the extra network/processing overhead to filter in the Lambda is not going to make much of a difference

api/graphql/schema.graphql
kostaspap: +1. Then the comment is not needed :)

api/lambda/alerts/models/api.go
kostaspap: Sure, you can remove if you wish.

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -26,84 +26,254 @@ import (
...
+		*filter = filter.And(
+			expression.Contains(expression.Name("title"), *input.Contains),
+		)
 	}
+}
+
+// filterByEventCount - fiters by an eventCount defined by a range of two numbers
+func filterByEventCount(filter *expression.ConditionBuilder, input *models.ListAlertsInput) {
+	// Ensure we are checking for valid inputs that are within an acceptable range
+	if input.EventCountMax != nil && input.EventCountMin != nil && *input.EventCountMin >= 0 && *input.EventCountMax >= *input.EventCountMin {
```
kostaspap, line 158: * The `input.EventCountMin>=0` shouldn't be needed since this check is performed already at the api level
*  Similar for `*input.EventCountMax >= *input.EventCountMin`. This check should be performed on API level. You can do it using this: https://godoc.org/gopkg.in/go-playground/validator.v9#hdr-Cross_Field_Validation . That way if someone specified invalid parameters (Min > Max) the operation should just fail. 
* The `*filter = filter...` assignment seems like a no-op? If this code is working it should be that `filter` is stateful, so you shouldn't need this assignment. Otherwise, I think this function should be returning the expression.ConditionBuilder? 
* You should define `eventCount` as a constant in `table.go` and just reference it here
* I believe the conditions can be simplified like:
```
if input.EventCountMax != nil {
	filter.And(expression.LessThanEqual(expression.Name("eventCount"), expression.Value(*input.EventCountMax)))
}
if input.EventCountMin != nil {
	filter.And(expression.GreaterThanEqual(expression.Name("eventCount"), expression.Value(*input.EventCountMin)))
}
```

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -26,84 +26,254 @@ import (
...
+		)
+	}
+}
 
-	// queries require and = condition on primary key
-	keyCondition := expression.Key(ddbKey).Equal(expression.Value(&ddbValue))
+// applyFilters - adds filters onto an expression
+func (table *AlertsTable) applyFilters(builder *expression.Builder, input *models.ListAlertsInput) {
+	// Start with an empty filter for a known attribute
+	filter := expression.AttributeExists(expression.Name("id"))
```
kostaspap, line 177: You can replace `"id"` with `AlertIDKey`

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -26,84 +26,254 @@ import (
...
+}
+
+// getSortDirection - gets the direction to sort the data
+func (table *AlertsTable) getSortDirection(input *models.ListAlertsInput) bool {
+	// By default, sort descending
+	if input.SortDir == nil {
+		return false
+	}
+	return *input.SortDir == "ascending"
+}
```
kostaspap, line 195: Since this returns a bool, method should be renamed to something like "isAscendingOrder" or something.

api/graphql/schema.graphql
s0l0ist: I'll put it as `nameContains` for uniformity.

internal/log_analysis/alerts_api/api/list_alerts.go
s0l0ist: Yeah, the proper way would be to store all items in ddb as either upper or lower case which would make the filter request a lot faster/cheaper.

Problems with the second approach:

1. Instead of post-processing on the ddb side where the filter runs, the query will essentially ignore the `title` and return a ton more records to the lambda for our custom post-processing.
2. The data received by the lambda will on average need to spend more time on each request.
3. The user will spend a lot more time waiting for data to come in during infinite scroll because there could be queries that could return empty results, but also have contain the `LastEvaluatedKey` signaling we need to continuously query. This will happen more often than if we ask ddb to filter the results.

Ex:
1. Alerts page renders, user sets the `nameContains` field over 1M records.
2. Run the query and paginate by 25 due to the viewport from the user's browser
3. For the next 10x queries, there are records returned to the lamba, but are subsequently filtered out and we return an empty set to the user.
4. The browser still gets the `LastEvaluatedKey` and continues from #2.
5. No more `LastEvaluatedKey` is present in the response, so the browser stops rendering results.

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -26,84 +26,254 @@ import (
...
 
+	// Limit the returned results to the specified page size or max default
 	var queryResultsLimit *int64
-	if pageSize != nil {
-		queryResultsLimit = aws.Int64(int64(*pageSize))
+	if input.PageSize != nil {
+		queryResultsLimit = aws.Int64(int64(*input.PageSize))
+	} else {
+		queryResultsLimit = aws.Int64(int64(25))
 	}
```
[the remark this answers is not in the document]
s0l0ist: I think we should always sanitize a user input field and not rely on the frontend to specify this value. Otherwise, it is possible it could be abused and return large `1MB` results.

I'll move this to the other layer, however.

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -26,84 +26,254 @@ import (
...
+	keyBuilder = expression.Key(TimePartitionKey)
+	return keyBuilder
+}
+
+// getKeyCondition - gets the key condition for a query
+func (table *AlertsTable) getKeyCondition(keyBuilder *expression.KeyBuilder,
+	input *models.ListAlertsInput) (keyCondition expression.KeyConditionBuilder) {
+
+	// If we have a ruleId, set the primary key and allow for filtering by createdAt
+	if input.RuleID != nil {
```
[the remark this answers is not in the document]
s0l0ist: Great catch, I built this in haste and forgot to optimize once it was working.

internal/log_analysis/alerts_api/table/list.go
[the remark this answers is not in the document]
kostaspap: > I think we should always sanitize a user input field and not rely on the frontend to specify this value. Otherwise, it is possible it could be abused and return large `1MB` results.

Sorry, I was probably misunderstood. I was not suggesting to not sanitize it, rather depend on the `validate` tags on the `ListAlertsInput` to do the sanitization. So my thought was that you could mark it as `validate:"required,..."`

internal/log_analysis/alerts_api/table/list.go
s0l0ist: Makes perfect sense. Also thanks for the cross field validation link, this was exactly what I was looking for.

web/src/pages/ListAlerts/ListAlertsActions/ListAlertsActions.tsx
```diff
@@ -0,0 +1,166 @@
...
+      label: 'Severity',
+      items: ['', ...severityOptions],
+      itemToString: (severity: SeverityEnum | '') =>
+        severity === '' ? 'All' : capitalize(severity.toLowerCase()),
+      inputProps: {
+        placeholder: 'Choose a severity...',
+      },
+    },
+  },
+  ruleId: {
```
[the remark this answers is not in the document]
s0l0ist: Yes, I would like this as well. Could we add this to the next sprint?

web/src/pages/ListAlerts/ListAlertsActions/ListAlertsActions.tsx
```diff
@@ -0,0 +1,166 @@
...
+    component: FormikTextInput,
+    props: {
+      label: 'Event count (max)',
+      placeholder: 'Enter number...',
+      type: 'number',
+      min: 0,
+    },
+  },
+  createdAtAfter: {
+    component: FormikTextInput,
```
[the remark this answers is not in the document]
s0l0ist: Unfortunately, we also need the time. The granularity it offers is huge when you have a ton of alerts in a short amount of time.

internal/log_analysis/alerts_api/api/list_alerts.go
kostaspap: I don't think the second approach is worth it right now. As you mention @austinbyers , the second approach would require a migration. If we were to do a migration, I would prefer to migrate our existing  DDB tables to store the rule/policy IDs lowercase and fix this issue for good. Doesn't worth a migration to half-fix the issue.

s0l0ist: This still applies to the alert's title

api/graphql/schema.graphql
```diff
@@ -805,6 +813,10 @@ enum ListRulesSortFieldsEnum {
   severity
 }
 
+enum ListAlertsSortFieldsEnum {
+  createdAt
```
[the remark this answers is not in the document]
s0l0ist: Unfortunately, this is really time-series data that we are presenting on this page. In my opinion, the context should always be sorted by time. Please correct me if this is not the case here.

Sorting by the number of events would also be a good feature, but it will require a lot more changes in the backend schema to make it performant. By keeping it out of this release, we are not performing any schema related updates and keeping the context in time-series form.

api/graphql/schema.graphql
```diff
@@ -82,8 +82,16 @@ input GetS3LogIntegrationTemplateInput {
 
 input ListAlertsInput {
   ruleId: ID
-  pageSize: Int
+  pageSize: Int # defaults to `25`
   exclusiveStartKey: String
+  severity: SeverityEnum
+  contains: String # matches against title
+  eventCountMin: Int
+  eventCountMax: Int
```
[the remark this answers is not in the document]
s0l0ist: Good point, but I like the finer controls as it allows me to look for alerts that have rarely triggered.

web/src/pages/ListAlerts/ListAlertsActions/ListAlertsActions.tsx
```diff
@@ -0,0 +1,166 @@
...
+    },
+  },
+  ruleId: {
+    component: FormikTextInput,
+    props: {
+      label: 'Rule ID',
+      placeholder: 'Enter a rule ID...',
+    },
+  },
+  eventCountMin: {
```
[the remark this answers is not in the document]
s0l0ist: Honestly, I think it shouldn't be a slider for an input where the upper bound could be really high. Unless you know of a way that can allow the slider to be really granular when the max input could be say 1M? It really depends on the number of alerts one can generate in 15 mins (smallest dedup period) for a given rule.

jacknagz:
> My question to @jacknagz is whether we are ok with:

> Adding datetime manually (instead of a picker)

Picker would be better, otherwise, it might get confusing on the format?

> Writing a RuleID manually (instead of autocompele)

Autocomplete would be nice, but a string match could work for v1 maybe? For example, if I typed CloudTrail or something, it would find all AWS.CloudTrail.ExampleRule. We should also be able to search by log type because that's how logs are organized in the data lake.

> Adding minEvents and maxEvents filters manually through an input (instead of a range/slider component)

A slider would be cool, but not a blocker

web/src/hooks/useRequestParamsWithInfiniteScroll.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
```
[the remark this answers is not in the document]
s0l0ist: Sounds good. Will update.

api/graphql/schema.graphql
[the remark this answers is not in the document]
austinbyers: Right, time is the only field we can efficiently sort today. Our other APIs allow sorting by any field by scanning all of them. This works because we assume there aren't that many total items (e.g. analysis-api). But here, there could be many thousands of alerts

internal/log_analysis/alerts_api/api/list_alerts.go
austinbyers: Right, I understand the lambda-filtering is less efficient. My point was simply that if we wanted case-insensitive filtering today, that's an easy way to do it. It makes us squirm as engineers because we know there's a better way, but in all likelihood it would work just fine. Dynamo and Go Lambdas are very fast, my guess is it wouldn't add more than a few tens of ms.

For now, we can keep the case-sensitive filtering as-is and then implement either lowercasing approach in a future PR. Just keep in mind developer time is incredibly valuable - the most efficient approach isn't always the best

internal/log_analysis/alerts_api/table/list.go
s0l0ist: Regarding `*filter = filter...` the assignment is needed because `filter.And` is returning a `ConditionBuilder`

Second, unless we implement custom validators we will still need some conditional checks. Ex: `eventCountMin` should not be greater than `eventCountMax`, but only if it is defined.

kostaspap: > Second, unless we implement custom validators we will still need some conditional checks. Ex: `eventCountMin` should not be greater than `eventCountMax`, but only if `eventCountMax` is defined.

I think validator should be able to handle that case btw as it can be instructed to ignore validations if a field is empty. I *think* this should work: 

```
EventCountMax   *int       `json:"eventCountMax" validate:"omitempty,min=1,gte=eventCountMin"`
```
(Note that it is important that the `omitempty` goes first). 
Ideally such checks would already be performed at the API level, If this doesn't work and it gets too complicated, you can leave it as is and we can take another stab at it later.

internal/log_analysis/alerts_api/api/list_alerts.go
s0l0ist: @austinbyers I was more talking about the time spent waiting for the higher number of queries to come back to the browser. 

The lambda will indeed spend a tiny bit more time, but the bottleneck is really the increased number of queries e2e as they are initiated by the browser. It will feel like a really slow pagination when there are tons of data to filter through. This is still a problem today, but it would be exacerbated if doing lambda filtering. 

Resolving for now.

s0l0ist:
By default the form will allow text input string to be accepted. The format is highlighted by the placeholder text in the event the browser doesn't support the `datetime-local` feature.

Ideally, we will replace this with a datetime picker which will solve this problem.

internal/log_analysis/alerts_api/api/list_alerts.go
austinbyers: If we implemented filtering in the lambda, the lambda should be responsible for filling each page, not the browser. The frontend would work exactly the same as before, you're right it would be far too slow to make the frontend handle empty pages.

So instead of scanning Dynamo once, the Lambda would keep scanning Dynamo over and over and filtering items until you get a full page, then return the result to the frontend. You'd have to be careful to keep the paging keys aligned correctly, since they would no longer match 1:1 with the Dynamo pages. But this is *almost* as fast as filtering in Dynamo itself, because the latency between Lambda and Dynamo in the same region is very low.

I know we're not doing it right now, but I wanted to clear up that misunderstanding. When we do implement, we may have time to implement the more ideal way in Dynamo itself, to be determined.

api/graphql/schema.graphql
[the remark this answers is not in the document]
3nvi: Then why do we have sorting as an option altogether? If that's the case, I think we should always fetch alerts sorted by `createdAt` and not expose "sorting" neither in the web UI nor in `schema.graphql`. It can be passively added (and hardcoded) to the related  Appsync resolve.

Thoughts?

web/src/pages/ListAlerts/ListAlertsActions/ListAlertsActions.tsx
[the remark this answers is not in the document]
3nvi: Sure, can you create p1 issue about it and  self-assign it?

web/src/pages/ListAlerts/ListAlertsActions/ListAlertsActions.tsx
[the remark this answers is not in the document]
3nvi: Valid point.

We should ideally then present the  min-max input fields "closer" to one another (since they work in tandem), but I think that's a bit subjective.

web/src/pages/ListAlerts/ListAlertsActions/ListAlertsActions.tsx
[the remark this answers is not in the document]
3nvi: we will need a datetime component for pounce, we just haven't gotten around to doing that. Can we create a  ticket with a p2 prio in pounce to add it?

api/graphql/schema.graphql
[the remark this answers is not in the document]
s0l0ist: We need a directional flip to sort in either **ascending** or **descending** order for the `createdAt` column. I believe users will appreciate being able to see what is happening 'now' (descending), but also helps to see what alerts they may have missed by looking 'backwards' (ascending) if there were a lot in a short amount of time.

web/src/helpers/utils.tsx
```diff
@@ -253,3 +253,27 @@ export const generateDocUrl = (baseUrl: string, version: string) => {
 export const isNumber = (value: string) => /^-{0,1}\d+$/.test(value);
 
 export const toStackNameFormat = (val: string) => val.replace(/ /g, '-').toLowerCase();
+
+// Adds a 'Z' suffix to a date string
+// Used to sanitize date string going to backend
+export const sanitizeDates = (parms: Partial<any>) =>
```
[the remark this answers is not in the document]
s0l0ist: Good point. I'm working on an alternative solution and will resolve this once it is finished.

web/src/pages/ListAlerts/ListAlertsTable/ListAlertsTable.tsx
```diff
@@ -53,15 +73,15 @@ const ListAlertsTable: React.FC<ListAlertsTableProps> = ({ items }) => {
...
             <Table.Cell>
               <Box my={-1}>
                 {alert.severity ? <SeverityBadge severity={alert.severity} /> : 'Not available'}
               </Box>
             </Table.Cell>
-            <Table.Cell>{shortenId(alert.alertId)}</Table.Cell>
             <Table.Cell align="right">{alert.eventsMatched}</Table.Cell>
-            <Table.Cell>{formatDatetime(alert.updateTime)}</Table.Cell>
+            <Table.Cell>{alert.ruleId}</Table.Cell>
+            <Table.Cell>{new Date(alert.creationTime).toUTCString()}</Table.Cell>
```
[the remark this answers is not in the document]
s0l0ist: Yep, this was my naive attempt to show UTC time. Will be showing local time.

web/src/helpers/utils.tsx
```diff
@@ -38,6 +39,9 @@ import sum from 'lodash-es/sum';
 import { ErrorResponse } from 'apollo-link-error';
 import { ApolloError } from '@apollo/client';
 
+// Add the plugin
+dayjs.extend(utc);
```
3nvi, line 43: we shoould add this at  `src/web/config.ts`.  That's where we do it in our other projects.  I.e.

```ts
import relativeTimePlugin from 'dayjs/plugin/relativeTime';

// Extend dayjs to allow to display relative times (i.e. "5 seconds ago")
dayjs.extend(relativeTimePlugin);
```

web/src/hooks/useRequestParamsWithPagination.tsx
```diff
@@ -36,7 +36,7 @@ function useRequestParamsWithPagination<AvailableParams extends { page?: number
   // new parameters, it clears all the parameters and just sets the parameters passed as an argument
   const setRequestParamsAndResetPaging = React.useCallback(
     (newParams: Partial<AvailableParams>) => {
-      updateUrlParams({ ...newParams, page: 1 });
+      setUrlParams({ ...newParams, page: 1 });
```
3nvi, line 39: Awesome! Thanks for noticing that.

One small thing, previously we only had `updateUrlParams` so we were forced to do `setRequestParamsAndResetPaging({ status: undefined });` to "clear out" a parameter.

Now that we have `setUrlParams`, could we update **all usages** of `setRequestParamsAndResetPaging` and remove every param that was previously set to  `undefined`.

In short,  this:
`setRequestParamsAndResetPaging({ status: undefined, page: 1 });`

should now be this:
`setRequestParamsAndResetPaging({ page: 1 });`

Makes sense?

web/src/hooks/useRequestParamsWithoutPagination.tsx
```diff
@@ -0,0 +1,51 @@
+/**
```
3nvi, line 1: I don't think this is needed at all. What this does code does, is exactly what  `useUrlParams` already does. Am I missing something?

web/src/hooks/useUrlParams.tsx
```diff
@@ -56,11 +56,25 @@ function useUrlParams<T extends { [key: string]: any }>() {
     });
   };
 
+  /**
+   * stringifies an object and replaces the query params of a URL
```
3nvi, line 60: Optional: we can re-use some of the  logic  from the function above instead of duplicating it,  but I'm cool with both since they are co-located in the same file

web/src/pages/ListAlerts/ListAlerts.tsx
```diff
@@ -73,30 +86,34 @@ const ListAlerts = () => {
...
-    );
-  }
-
-  if (!alertItems.length) {
+  if (!alertItems.length && isEmpty(requestParams)) {
     return <ListAlertsPageEmptyDataFallback />;
   }
 
-  //  Check how many active filters exist by checking how many columns keys exist in the URL
+  const showIfError = Boolean(error);
```
3nvi, line 93: Optional: naming is a bit weird here, since I'd expect  `showIfError`  to be a function, but that's ok

web/src/hooks/useRequestParamsWithoutPagination.tsx
s0l0ist: wraps it with useCallback and provides almost the exact same api as the component with paging.

web/src/hooks/useRequestParamsWithPagination.tsx
s0l0ist: yep, fixed where applicable.

web/src/hooks/useRequestParamsWithoutPagination.tsx
3nvi: My  reasoning for not needing this hook, is that I want   to keep URL-manipulating React hooks to a  minimum. 

Currently,  `useUrlParams` handles most cases, with `useRequestParamsWithPagination` composing the former to provide a cleaner API. The current one just adds some syntactic sugar for a sole-case, so I feel that a re-usable hook won't be needed (i.e. `useUrlParams` can be utilized directly).

Just because I'm not feeling super strongly on this, I'm letting you ultimately decide what's better here

web/src/pages/PolicyDetails/PolicyDetails.tsx
```diff
@@ -111,9 +111,7 @@ const PolicyDetailsPage = () => {
                 count={getComplianceItemsTotalCount(totalCounts)}
                 text="All"
                 isActive={!requestParams.status && !requestParams.suppressed}
-                onClick={() =>
-                  setRequestParamsAndResetPaging({ status: undefined, suppressed: undefined })
-                }
+                onClick={() => setRequestParamsAndResetPaging()}
```
3nvi, line 114: I'd rather have a param here set to  `{}` to explicitly  show the reader that I'm clearing up those params. IMO param should be  mandatory on the hook as  well

s0l0ist: Okay no problem, I'll remove the optional.

api/lambda/alerts/models/api.go
```diff
@@ -55,13 +55,38 @@ type GetAlertOutput = Alert
...
+	// Infinite scroll/pagination query key
+	ExclusiveStartKey *string `json:"exclusiveStartKey"`
+
+	// Filtering
+	Severity        []*string  `json:"severity" validate:"omitempty,dive,oneof=INFO LOW MEDIUM HIGH CRITICAL"`
+	NameContains    *string    `json:"nameContains"`
+	CreatedAtBefore *time.Time `json:"createdAtBefore"`
+	CreatedAtAfter  *time.Time `json:"createdAtAfter"`
+	RuleContains    *string    `json:"ruleContains"`
+	AlertIDContains *string    `json:"alertIDContains"`
```
austinbyers, line 84: The json here should probably be `alertIdContains` to be consistent with `ruleId` and our other APIs

internal/log_analysis/alerts_api/api/list_alerts.go
```diff
@@ -35,11 +36,19 @@ func (API) ListAlerts(input *models.ListAlertsInput) (result *models.ListAlertsO
...
-	} else { // list all alerts time desc order
-		alertItems, result.LastEvaluatedKey, err = alertsDB.ListAll(input.ExclusiveStartKey, input.PageSize)
+
+	// Perform some validation here for items that do not have custom validators implemented
+	if input.CreatedAtAfter != nil && input.CreatedAtBefore != nil && input.CreatedAtBefore.Before(*input.CreatedAtAfter) {
+		return nil, &genericapi.InternalError{Message: "Invalid range, created at 'before' must be greater than 'after'"}
+	}
+
+	if input.EventCountMax != nil && input.EventCountMin != nil && *input.EventCountMax < *input.EventCountMin {
+		return nil, &genericapi.InternalError{Message: "Invalid range, event count 'max' must be greather or equal to 'min'"}
```
austinbyers, line 46: Typo: "greather" => "greater"

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -19,93 +19,124 @@ package table
...
+	// Continuously query until we have enough results for the requested page size
+	err = table.Client.QueryPages(queryInput, func(page *dynamodb.QueryOutput, isLast bool) bool {
+		for _, item := range page.Items {
+			// Define temp container
+			var alert *AlertItem
+			// Unmarshal each item to an alert
+			err = dynamodbattribute.UnmarshalMap(item, &alert)
+			if err != nil {
+				return false
+			}
```
austinbyers, line 102: No need for a double-pointer, and you also need to handle that error condition instead of just ignoring it. One way to do that is to define a `var errMarshal error` outside the function closure (e.g. line 93). Then:

```suggestion
			var alert AlertItem
			// Unmarshal each item to an alert
			if err = dynamodbattribute.UnmarshalMap(item, &alert); err != nil {
				// Something is wrong with this Dynamo item
				marshalErr = err
				return false
			}
```

The way you have it now, the `err` variable would be overwritten with `nil` on line 94 when `QueryPages` returns successfully, even if there were an error inside the function. Error handling can sometimes get awkward inside a function closure.

Then you would check if `err != nil` or `marshalErr != nil`

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -114,3 +145,153 @@ func (table *AlertsTable) list(ddbKey, ddbValue string, exclusiveStartKey *strin
...
+			RuleIDKey:    item[RuleIDKey],
+			CreatedAtKey: item[CreatedAtKey],
+			AlertIDKey:   item[AlertIDKey],
+		}
+	}
+	return DynamoItem{
+		TimePartitionKey: item[TimePartitionKey],
+		CreatedAtKey:     item[CreatedAtKey],
+		AlertIDKey:       item[AlertIDKey],
+	}
```
austinbyers, line 164: Nice! 👍

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -114,3 +145,153 @@ func (table *AlertsTable) list(ddbKey, ddbValue string, exclusiveStartKey *strin
...
+	// We are allowing either Before -or- After to work together or independently
+	if input.CreatedAtAfter != nil && input.CreatedAtBefore != nil {
+		keyCondition = keyCondition.And(
+			expression.Key(CreatedAtKey).Between(
+				expression.Value(*input.CreatedAtAfter),
+				expression.Value(*input.CreatedAtBefore),
+			),
+		)
+	}
+	if input.CreatedAtAfter != nil && input.CreatedAtBefore == nil {
```
austinbyers, line 195: Nit: since only one of these 3 cases is true, a `switch` or `if-else` might be more appropriate:

```go
switch {
    case input.CreatedAtAfter != nil && input.CreatedAtBefore != nil:
        keyCondition = ...
    case input.CraetedAtAfter != nil && input.CreatedAtBefore == nil:
        keyCondition = ...
    case input.CreatedAtAfter == nil && input.CreatedAtBefore != nil:
        keyCondition = ...
}
```

This way, it will stop checking as soon as a match is found

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -114,3 +145,153 @@ func (table *AlertsTable) list(ddbKey, ddbValue string, exclusiveStartKey *strin
...
+		multiFilter := expression.Name(SeverityKey).Equal(expression.Value(*input.Severity[0]))
+
+		// Then add or conditions starting at a new slice from the second index
+		for _, severityLevel := range input.Severity[1:] {
+			multiFilter = multiFilter.Or(expression.Name(SeverityKey).Equal(expression.Value(*severityLevel)))
+		}
+
+		*filter = filter.And(
+			multiFilter,
+		)
```
austinbyers, line 233:
```suggestion
		*filter = filter.And(multiFilter)
```

api/lambda/alerts/models/api.go
s0l0ist: Good catch, I'll also rename `RuleContains` -> `RuleIDContains`

internal/log_analysis/alerts_api/table/list.go
s0l0ist: Good catch. I've now captured the error outside the closure.

austinbyers:
🎉  congrats @s0l0ist on your first major contribution! May it be the first of many more

PR #1030, opened 2020-06-11
Add embedded metric wrapper

nhakmiller:
## Background

Iteration on what the embedded metric logging could look like. I put it in `pkg` with our other loggers.

Now we have a `Logger` and a `MonoLogger` (maybe `FastLogger` is better?).

`Logger` requires you to specify dimension values and metrics with each call to `Log`, and allows you to specify any number of metrics (up to the AWS limit of 100) per log call. Dimension values must be provided for each dimension in each `DimensionSet` specified when the logger was instantiated. Specifying `DimensionSet`s during logger instantiation is optional.

`MonoLogger` is much more restrictive, requiring you to specify a single metric and (optionally) dimension set up front, but then relieving you of the burden of specifying those with each subsequent call to `Log`. It also does far less error/input checking on each  `Log` as it verifies more up front. This should be faster/simpler/more ideal for performance critical parts of the code.

Finally, we can still call `LogEmbedded` directly. This is the function used internally by the `Logger`. Calling it directly allows you to specify the namespace, metrics, dimensions, and dimension sets all at once.

I made `fastLogEmbedded` (the function used by the `MonoLogger`) internal so that it could not be used directly. This is because it does not do any error checking on user input, instead assuming that was already performed when the `MonoLogger` was instantiated. We could export this function, but I feel like that might lead to poorly constructed metrics that don't do what they're supposed to do.

In this PR the logger is only used in the `log_processor`, alongside the existing opslogger. Part of the full rollout of embedded metric logging will likely replace all usage of the opslogger but for now I left them side by side. My goal is to get a full end to end workflow for at least one metric (including being queried with the future metrics API) so that the frontend team has something to work with, then go wide with it to cover all the other cases.

## Changes

- Created the `metrics` package for logging in the AWS embedded metric format
  - Created the `Logger` struct for general metric logging
  - Created the `MonoLogger` struct for stream lined metric logging
- Updated the `log_processor` lambda to log the number of bytes processed per log type in the embedded metrics format

## Testing

- Manually verified that the metrics are being generated as intended
- Updated unit tests to verify the metrics are being logged as intended


internal/log_analysis/log_processor/processor/processor_test.go
```diff
@@ -327,10 +338,25 @@ func TestProcessClassifyFailure(t *testing.T) {
...
+				zap.Any("bytesProcessed", 0),
+				zap.Any("_aws", embeddedMetric),
+			},
+		},
 	}
 	require.Equal(t, len(expected), len(actual))
 	for i := range expected {
+		if i == len(expected)-1 {
+			continue
+		}
```
nhakmiller, line 357: When I try to compare these with the `assertLogEqual` function it panics saying `unknown log type ...`. When I try to compare them with the normal `assertEqual` I get the following:

```
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -12,3 +12,3 @@
                                    Key: (string) (len=7) "logType",
                                -   Type: (zapcore.FieldType) 15,
                                +   Type: (zapcore.FieldType) 0,
                                    Integer: (int64) 0,
                                @@ -19,6 +19,6 @@
                                    Key: (string) (len=14) "bytesProcessed",
                                -   Type: (zapcore.FieldType) 11,
                                +   Type: (zapcore.FieldType) 0,
...
```
The FieldType 0 is for `UnknownType` logs, and the other field types are strings, int64s, etc. Even though I'm setting up the expected fields with `zap.Any(...)`, it still appears to be specifying a type which is failing in the comparison. Still looking into it but I plan to get these passing before merging.

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -158,6 +170,13 @@ func (p *Processor) logStats(err error) {
 	p.operation.Log(err, zap.Any(statsKey, *p.classifier.Stats()))
 	for _, parserStats := range p.classifier.ParserStats() {
 		p.operation.Log(err, zap.Any(statsKey, *parserStats))
+		values := map[logger.Metric]interface{}{
+			processorMetric: parserStats.BytesProcessedCount,
+		}
+		err = eLogger.Log(values, map[string]string{"logType": parserStats.LogType})
+		if err != nil {
```
austinbyers, line 177: Having to check for 'err != nil' on every log call is too burdensome for the caller I'd say.

Move this error check inside so that `.Log` can always be a simple function call

pkg/embeddedlogger/logger.go
```diff
@@ -0,0 +1,128 @@
...
+
+type Logger struct {
+	namespace  string
+	dimensions [][]string
+}
+
+// Create a new logger for a given namespace and set of dimensions
+func NewLogger(namespace string, dimensions [][]string) (*Logger, error) {
+	if namespace == "" || dimensions == nil {
+		return nil, errors.New("namespace and dimension cannot be empty")
```
austinbyers, line 52: Dimensions are required? I'm surprised, I didn't think that was true of CloudWatch metrics in general

In that case, you could add an additional `MustLogger` wrapper function which calls `NewLogger` and then panics for an error (similar to `session.Must`). This way, loggers can be initialized at the package level:

```go
package mylambda

var metricLogger = embeddedlogger.MustLogger(...)
```

pkg/embeddedlogger/logger.go
```diff
@@ -0,0 +1,128 @@
+package embeddedlogger
```
austinbyers, line 1: Would `package metriclogger` be more appropriate or no?

pkg/embeddedlogger/logger.go
```diff
@@ -0,0 +1,128 @@
...
+	dimensions map[string]string, dimensionSets [][]string, timestamp int64) error {
+	// Validate input
+	if namespace == "" {
+		return errors.New("namespace is required")
+	}
+	if len(values) == 0 || len(dimensions) == 0 || len(dimensionSets) == 0 {
+		return errors.New("values, dimensions, and dimensionSets cannot be empty")
+	}
+
+	// Set timestamp to current time if one is not provided
```
austinbyers, line 76: Can you think of any example where the backend service would want to override the metric timestamp?

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -47,7 +49,17 @@ var (
 	// temporarily busy. The writer goroutines will block writing but only when the buffer has been full - something we need
 	// to avoid using up lot of memory.
 	// see also: https://golang.org/doc/effective_go.html#channels
-	ParsedEventBufferSize = 1000
+	ParsedEventBufferSize  = 1000
+	processorLogDimensions = [][]string{
+		{
+			"logType",
+		},
+	}
```
austinbyers, line 57: The fact that this is a 2D array is confusing to me - could this be flattened and the library handles expanding to the AWS expected format? What's the difference between

```go
	processorLogDimensions = [][]string{
		{
			"logType",
                        "sourceId",
		},
	}
```

and

```go
	processorLogDimensions = [][]string{
		{
			"logType",
                },
                {
                        "sourceId",
		},
	}

File: internal/log_analysis/log_processor/processor/processor_test.go (hunk 1 above)
Comment (austinbyers) [reply internal/log_analysis/log_processor/processor/processor_test.go:357]: Strange, I'm not sure the problem either

File: pkg/embeddedlogger/logger.go (hunk 4 above)
Comment (nhakmiller) [reply pkg/embeddedlogger/logger.go:1]: I think so, being an embedded metric format logger is more of an implementation detail. I'll switch this over.

File: pkg/embeddedlogger/logger.go (hunk 5 above)
Comment (nhakmiller) [reply pkg/embeddedlogger/logger.go:76]: My main thought was for a service that does a bunch of work and wants to log the time when the work starts, but only send the log after the work is done. Probably only talking about milliseconds, but because the embedded metric format allows (and in fact requires) an explicit timestamp I thought it was reasonable to make this behavior available to callers. I'd be fine with hiding it if you think every single call will always pass in `0` here.

File: internal/log_analysis/log_processor/processor/processor.go (hunk 6 above)
Comment (nhakmiller) [reply internal/log_analysis/log_processor/processor/processor.go:57]: From my understanding, each dimensionSet passed in corresponds to exactly one queryable metric. So this:
```
	processorLogDimensions = [][]string{
		{
			"logType",
                        "sourceId",
		},
	}
```
means that the dimension is logType + sourceId. You cannot request metrics for just the logType or just the sourceId. So if you want to be able to request metrics based on the logType OR the sourceId OR the logType + sourceId you would need the following:
```
	processorLogDimensions = [][]string{
		{
			"logType",
                        "sourceId",
		},
		{
                        "sourceId",
		},
		{
			"logType",
		},
	}
```

If you just wanted metrics based on the logType OR the logType + sourceId, this would work:

```
	processorLogDimensions = [][]string{
		{
			"logType",
                        "sourceId",
		},
		{
			"logType",
		},
	}
```

If we were ok with each and every metric being logged to only exactly one dimension set we could make this a normal array, but that loses a lot of flexibility I think.

Note that I haven't tested this too much (just played around in the console a bit), but my small testing seems to agree with this and I'm basing it off the AWS docs here: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Dimension (see the section "Dimension Combinations").

File: internal/log_analysis/log_processor/processor/processor.go (hunk 2 above)
Comment (nhakmiller) [reply internal/log_analysis/log_processor/processor/processor.go:177]: I like this, will do. Should it panic if there's an error, or just log an error? I suppose you wouldn't want production code panic'ing on account of failed metric generation, probably logging an error would be preferred.

File: pkg/embeddedlogger/logger.go (hunk 3 above)
Comment (nhakmiller) [reply pkg/embeddedlogger/logger.go:52]: For some reason I thought dimensions were required based on my reading of the docs, but going back to re-read I see that may not be the case after all. I'll do some checking.

**Comment by rleighton** [comment]
Added @alxarch . This is a very important change. Needs to be squinted at by many eyes.

File: internal/log_analysis/log_processor/processor/processor.go (hunk 2 above)
Comment (austinbyers) [reply internal/log_analysis/log_processor/processor/processor.go:177]: Agreed, just have the `.Log` function log any errors and return. (You could wrap your existing function with another one that just catches the error and logs it)

File: pkg/embeddedlogger/logger.go (hunk 5 above)
Comment (austinbyers) [reply pkg/embeddedlogger/logger.go:76]: Don't have a strong preference, but to keep things simple my thought it let's remove it. Just like `zap.L()` doesn't let you configure the timestamp or line number of the log message

File: internal/log_analysis/log_processor/processor/processor.go (hunk 6 above)
Comment (austinbyers) [reply internal/log_analysis/log_processor/processor/processor.go:57]: Got it, makes sense. Maybe adding some of that context to the logger functions will help clarify

File: pkg/embeddedlogger/logger.go (hunk 5 above)
Comment (nhakmiller) [reply pkg/embeddedlogger/logger.go:76]: I feel like a nice compromise is not allowing you to specify when calling `Log` with a `Logger`, but requiring you to specify if you want to do the long form `LogEmbedded` which requires you to specify a lot more. This way it's there for odd cases where you need to specify something, but for the general case its hidden.

File: internal/log_analysis/log_processor/processor/processor.go [hunk 7]
```diff
@@ -47,7 +49,17 @@ var (
...
+	processorMetric = logger.Metric{
+		Name: "bytesProcessed",
+		Unit: "Bytes",
+	}
+	eLogger = logger.MustLogger(processorLogNamespace, processorLogDimensions)
```
kostaspap, line 62: 1. I think it's better that the metrics stuff should have an appropriate namespace e.g. `metrics`. We are using a zap logger underneath, but we could be using as well CW Metrics API directly. This will allow use to define stuff like:
```
package metrics

const PantherNamespace = "Panther"

const (
  Byte = iota
  KiloByte
  Millisecond
....
)

type Dimension struct{
Name string
Value string
}
```

2. The metrics namespace shouldn't have to used in every invocation, it would be better if it was fixed for every given app. We could have a metrics manager that we use like we use e.g. AWS Clients (instantiate once, share in the code) 

```
var MetricsMgr metrics.Manager = metrics.New(<default dimensions>)
....

MetricsMgr.Count("metricName", 100)
MetricsMgr.Bytes("metricName", 1000, metrics.Dimension{Name:"New", Value:"Dimension"})
....
```

or using the pattern followed by zap.ReplaceGlobals, this becomes a bit nicer when using it but has potentially other issues e.g. not easy to test. 

```
metrics.Initialize(<default dimensions>) // use like zap.ReplaceGlobals
......
metrics.Count("metricName", 100)
metrics.Bytes("metricName", 1000, metrics.Dimension{Name:"New", Value:"Dimension"})
metrics.Milliseconds("metricName", 100)
metrics.Datum(metrics.Kilobytes, "metricName", 1000)
....

```

3. Better just keep one namespace e.g. Panther. It makes it easier to group the metrics in the CW Dashboard. The specific component can be just a dimension.

internal/log_analysis/log_processor/processor/processor.go
kostaspap: Better define a `Dimension` struct tbh. Something like:
```
package metrics

type Dimension struct {
Name string
Value string
}
```

nhakmiller: I'll add a dimension struct, it goes well with the metric and directive structs already defined for this. We'll still need a 2D array for defining the dimensions initially though.

internal/log_analysis/log_processor/processor/processor.go
nhakmiller: (I will address these comments one at a time so there can be different threads)

> I think it's better that the metrics stuff should have an appropriate namespace e.g. metrics. We are using a zap logger underneath, but we could be using as well CW Metrics API directly.

I am not sure I understand this comment, are you saying I should rename the package from `metriclogger` to `metrics`? I am ok with that.

nhakmiller: > We could have a metrics manager that we use like we use e.g. AWS Clients (instantiate once, share in the code)

I agree, I think that is essentially what the `Logger` struct I've implemented here is. You specify a namespace and expected dimensions one time and receive back an object with those things configured that you can pass around as you like. Or am I misunderstanding your feedback?

nhakmiller: > or using the pattern followed by zap.ReplaceGlobals, this becomes a bit nicer when using it but has potentially other issues e.g. not easy to test.

This is what I meant with this comment in the description:

```
Instead of creating a Logger and returning it to the caller to be passed around, should we have an Init that sort of creates an internal only logger and then the callers can do an embededlogger.Log(xyz)? This way the logger doesn't have to be passed around which is nice, but it would make logging different types of metrics in the same code more difficult.
```

I think we could make it work, but it would make logging metrics with different dimensions (not just different dimension values) in the same service much more difficult. For example I am thinking that the log processor will want to create the `bytesProcessed` metric with the dimensions `logType` and possibly `logSource`, but may also want to create the metric `logTypesProcessed` with just the dimension `logSource`. This would be very awkward if it was initialized once. Of course with careful dimension & metric selection this may be a non-issue, just a question of if the ease of use is worth the flexibility tradeoff.

nhakmiller: > Better just keep one namespace e.g. Panther. It makes it easier to group the metrics in the CW Dashboard. The specific component can be just a dimension.

I don't have a strong preference on just one namespace or multiple, @austinbyers and I were discussing it might be nice to have some level of separation though e.g. by area. So maybe `panther/log_analysis`, `panther/core`, `panther/cloud_security`. What do you think? I'm sort of assuming that there is some performance or cost benefit from having more granular namespaces when querying metrics, but that could be totally wrong in which case one namespace is probably fine.

austinbyers: Namespace is just that - the name of a metric. Performance/cost shouldn't be affected in the slightest

nhakmiller: > Namespace is just that - the name of a metric. Performance/cost shouldn't be affected in the slightest

Ah ok, I assumed that by adding more namespaces you could reduce your search space when requesting data for a given namespace and therefore reduce the search time.

kostaspap: > > We could have a metrics manager that we use like we use e.g. AWS Clients (instantiate once, share in the code)
> 
> I agree, I think that is essentially what the `Logger` struct I've implemented here is. You specify a namespace and expected dimensions one time and receive back an object with those things configured that you can pass around as you like. Or am I misunderstanding your feedback?

Nope, I think we are aligned. The only difference is that I see that you define the (metrics)logger inside the `processor` package,  while imo it should be a singleton and should be initialized when on the Lambda initialization code, on the same place we for example initialize the loggers

kostaspap: > (I will address these comments one at a time so there can be different threads)
> 
> > I think it's better that the metrics stuff should have an appropriate namespace e.g. metrics. We are using a zap logger underneath, but we could be using as well CW Metrics API directly.
> 
> I am not sure I understand this comment, are you saying I should rename the package from `metriclogger` to `metrics`? I am ok with that.

Yup. Current namespace and method names suggest to me the usage of a logging framework rather than a metrics framework (e.g. `MustLogger`)

kostaspap: > > or using the pattern followed by zap.ReplaceGlobals, this becomes a bit nicer when using it but has potentially other issues e.g. not easy to test.
> 
> This is what I meant with this comment in the description:
> 
> > Instead of creating a Logger and returning it to the caller to be passed around, should we have an Init that sort of creates an internal only logger and then the callers can do an embededlogger.Log(xyz)? This way the logger doesn't have to be passed around which is nice, but it would make logging different types of metrics in the same code more difficult.
> 
> I think we could make it work, but it would make logging metrics with different dimensions (not just different dimension values) in the same service much more difficult. For example I am thinking that the log processor will want to create the `bytesProcessed` metric with the dimensions `logType` and possibly `logSource`, but may also want to create the metric `logTypesProcessed` with just the dimension `logSource`. This would be very awkward if it was initialized once. Of course with careful dimension & metric selection this may be a non-issue, just a question of if the ease of use is worth the flexibility tradeoff.

Yeah, I agree, we should be able to add custom dimensions to a metric as we wish. The initialization I was referring to was to allow us to define custom dimensions that we are sure will not change for the whole application. E.g. name of the Lambda could be one dimension like that.

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -158,6 +170,10 @@ func (p *Processor) logStats(err error) {
 	p.operation.Log(err, zap.Any(statsKey, *p.classifier.Stats()))
 	for _, parserStats := range p.classifier.ParserStats() {
 		p.operation.Log(err, zap.Any(statsKey, *parserStats))
+		values := map[logger.Metric]interface{}{
+			processorMetric: parserStats.BytesProcessedCount,
```
rleighton, line 174: This is expensive, lots of allocs.

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -158,6 +170,10 @@ func (p *Processor) logStats(err error) {
 	p.operation.Log(err, zap.Any(statsKey, *p.classifier.Stats()))
 	for _, parserStats := range p.classifier.ParserStats() {
 		p.operation.Log(err, zap.Any(statsKey, *parserStats))
```
rleighton, line 172: Shouldn't we remove this in favor of whatever new code we write?

rleighton:
This is a big feature, suggest it gets moved to 1.6 .

alexmylonas:
Hey guys, i see this is an evolving conversation! In order for us to plan the frontend counterpart, do you a draft date estimate for a stable version of this? 😄

internal/log_analysis/log_processor/processor/processor_test.go
nhakmiller: I got it eventually.

pkg/embeddedlogger/logger.go
nhakmiller: I haven't actually been able to find anything that says one way or the other, so I guess I'll just assume they're not required for now. I still think we should support them though.

pkg/embeddedlogger/logger.go
nhakmiller: Removed, we can always re-add if it's needed.

internal/log_analysis/log_processor/processor/processor.go
nhakmiller: Tried to reduce the allocs as much as possible for the `MonoLogger`, take a look now.

internal/log_analysis/log_processor/processor/processor.go
nhakmiller: I was going to let them sit side by side for now, unless you'd prefer I just rip it out now?

pkg/metrics/logger.go
```diff
@@ -0,0 +1,290 @@
...
+	Metric
+	Value int
+}
+
+const (
+	UnitBytes = "Bytes"
+	// UnitSeconds      = "Seconds"
+	// UnitMicroseconds = "Microseconds"
+	// UnitMilliseconds = "Milliseconds"
+)
```
nhakmiller, line 90: I considered doing an enumerated type + a toString function (since it's gotta be a string at before we log it) so we could have strong typing here, but it was waaay more boilerplate and didn't feel worth it after I got halfway through so I just switched to these constants.

nhakmiller:
Screen Shot 2020-07-06 at 1 22 15 PM

This graph was generated automatically for me by CloudWatch. I did not create any metric filters.

internal/log_analysis/log_processor/common/metrics.go
```diff
@@ -0,0 +1,37 @@
+package common
```
nhakmiller, line 1: I moved the definition of the logger to the `common` package for the `log_processor` per a suggestion by @kostaspap. I wasn't sure if this was the correct location, but it was where the `opslogger` was defined so I figured it would be a good fit.

pkg/metrics/logger.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	return &Logger{
+		namespace:     namespace,
+		dimensionSets: dimensionSets,
+	}, nil
+}
+
+// Log sends a log formatted in the CloudWatch embedded metric format
+func (l *Logger) Log(values map[Metric]interface{}, dimensions map[string]string) {
```
nhakmiller, line 135: I considered creating a new type, `MetricValue` that contains a `Metric` and a value for the metric. That would let us replace the `map[Metric]interface{}` with a `[]MetricValue`. I left it as a map because:

1. I didn't want to pollute the `metrics` namespace with too many similarly named types, as it might be confusing to maintainers
2. Leaving it as a map allows the caller to just update one metric and re-call, which may be valuable in the case where multiple metrics are being logged at each call but only some of them are actually being updated.

I'm not actually sure if reason 2 has any real use cases, but I thought reason 1 was good enough, and a mapping of a metric to a value makes a certain amount of sense to me.

pkg/metrics/logger.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	timestamp := time.Now().UnixNano() / 1000000 // Nanosecond -> Millisecond conversion
+
+	// Verify that each dimension key required by a dimension set is present. This is mandated by
+	// the AWS specification.
+	//
+	// The inverse is not checked, but if a caller specifies a dimension that is not present
+	// in any dimensionSet it will be logged but ignored by AWS.
+	for _, dimensionSet := range dimensionSets {
```
nhakmiller, line 164: We could speed this up by calculating a super set of all dimension keys when the logger is initialized and just using that. But the dimension sets have at most 9 keys, and there are generally only going to be 1-3 dimension sets. Iterating over them is going to be pretty fast already.

internal/log_analysis/log_processor/processor/processor.go
nhakmiller: I created the dimension struct and use that where appropriate. Additionally, the new `MonoLogger` supports only a single dimension set. For basic many use cases, this should be sufficient and simplify the interactions.

internal/log_analysis/log_processor/processor/processor.go
nhakmiller: I addressed your original points 1 & 2. I still have a more specific namespace in place as that seems still up in the air, but I'm going to resolve this chain just because it is rather long and disjointed and over half of it is no longer relevant.

internal/log_analysis/log_processor/common/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import (
+	"github.com/panther-labs/panther/pkg/metrics"
+)
+
+const processorLogNamespace string = "panther/log_processor"
```
nhakmiller, line 25: We still haven't settled on whether we should have a single namespace and separate systems by dimension, or have multiple namespaces. I've left it how it was before but open to any strong arguments one way or the other.

kostaspap: 1. Better capitalize the first letter e.g. `Panther`. See also how AWS capitalizes the first letter of their namespace e.g. `ApiGateway`, `AppSync`, etc.

1. My vote goes for single namespace. I think sticking to one namespace will be "easier" for any of the users deploying Panther. We have internally the separation of log analysis, cloud security, core but for most of the users the system is just ... Panther.
An example is AWS S3. S3 is a far more complicated system than Panther that is build from 100s microservices, still it is all available under one namespace. 
The Namespace doesn't really matter from a purely engineering perspective, one or multiple namespaces we can all make it work with our APIs. The decision is mostly how we want these metrics to be displayed if anyone accesses the CW Metrics console, and I think a single namespace is easier for the users.

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -164,6 +164,8 @@ func (p *Processor) logStats(err error) {
...
+		common.BytesProcessedDimension.Value = parserStats.LogType
```
kostaspap, line 167: Global mutable variables are... scary. An in this case it wouldn't work, since multiple goroutines run this code simultaneously so we can end up emitting the wrong metrics. 
Better just create the `BytesProcessedLogger` just with the namespace. You can create the appropriate dimension in here.

internal/log_analysis/log_processor/common/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+
+var (
+	BytesProcessedLogger = metrics.MustMonoLogger(processorLogNamespace, metrics.DimensionSet{
```
kostaspap, line 28: We shouldn't need a different logger per metric... this seems overkill. Not sure why would would like to have such a global state. We should make it easy for any piece of the code to emit a metric without requiring to create a new logger, specify a namespace, etc. 
I think it would be better to have a `MetricLogger` which accepts a namespace and some dimensions (that will be included by default for everyone using the metric logger). Then someone can emit a metric in the code by 
```
MetricLogger.Log(metric)
//
MetricLogger.Log(metric, dimensions)
```

To do that we can just change the Metric to include also the value:
```
type Metric struct{
  Name string
  Unit string
  Value interface{}
}
```

internal/log_analysis/log_processor/common/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+		"logType",
+	}, metrics.Metric{
+		Name: "bytesProcessed",
```
kostaspap, line 31: Better capitalize first letter e.g. `LogType`, `BytesProcessed`. Seems to be more common for metrics

pkg/metrics/logger.go
```diff
@@ -0,0 +1,285 @@
...
+// supports one dimension set and one metric which must be set at initialization.
+//
+// These limitations still allow for 90% of use cases, and are more suitable for performance
+// critical parts of the code than the Logger.
+type MonoLogger struct {
+	directive []MetricDirectiveObject
+}
+
+// MustMonoLogger creates a new MonoLogger based on the given input, and panics if the input is invalid
+func MustMonoLogger(namespace string, dimensionSet DimensionSet, metric Metric) *MonoLogger {
```
kostaspap, line 215: I don't think we want to create a different logger per metric.. One logger per application/lambda should be enough. This introduces some state that we don't really need (as I see here in the code) and will make this library more cumbersome to use.

pkg/metrics/logger.go
kostaspap: Having a `MetricValue` would be overkill, but seems that the `Metric` can have a value itself (see comment above).

internal/log_analysis/log_processor/processor/processor.go
nhakmiller: @rleighton did not like when I put it inside the for loop because he said it was too many allocations, so I just moved it to where there would be the least number of allocations 😄 I did not realize this code was running on multiple go routines. I can move it to just before the for loop, there will still be lots of allocations but slightly less.

internal/log_analysis/log_processor/common/metrics.go
nhakmiller: To you first point, that is essentially what the `Logger` is already. Here I am just using the `MonoLogger` because the feedback that I got on the `Logger` was that it was too expensive to do so many operations on every log call.

We need a `Metric` struct that does not have a value to correctly construct the `_aws` node to be in conformance with the AWS embedded metric format. We could create a `Metric` struct with a `Value` field with json tags that indicate to never serialize it, but I was worried that would be confusing because in some cases (i.e. the initialization) you don't need the values at all. I will take another stab at it, but I'm not sure it will be much better.

It is not generally enough to pass just a dimension. You need the dimension sets and the dimension values, and those things have different meanings. They cannot be represented by the same struct unless (as above) you do not serialize the value field, and in that case half the time you're not using half the struct. Even then, the dimension sets need to be represented by a two dimensional array (or slice or w/e), whereas the dimension values themselves are just a mapping of names to values.

Would it be sufficient to allow the initialization of the logger to be passed any number of dimension sets and corresponding dimension values, but then subsequent calls to `Log` can pass only exactly one dimension set and the dimension values of only that dimension set? In which case they could be represented the same way. It would limit the flexibility of the logger, you could not add two different dimension sets when calling `Log`, but it would simplify the interaction.

The problem is we want it to be simple to interact with while also maintaining the full flexibility of the format, and honestly we just can't have both in all cases. It is a complex format, interacting with it fully is complex. Interacting with a subset of it can be simple, but then we can't do everything.

internal/log_analysis/log_processor/common/metrics.go
nhakmiller: Updated.

internal/log_analysis/log_processor/processor/processor.go
nhakmiller: Moved to just before the for loop.

internal/log_analysis/log_processor/processor/processor_test.go
```diff
@@ -267,6 +268,22 @@ func TestProcessClassifyFailure(t *testing.T) {
...
+				Dimensions: []metrics.DimensionSet{{"logType"}},
+				Metrics: []metrics.Metric{
+					{
+						Name: "bytesProcessed",
+						Unit: metrics.UnitBytes,
+					},
+				},
+			},
+		},
+		Timestamp: p.operation.EndTime.UnixNano() / 1000000,
```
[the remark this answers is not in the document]
nhakmiller: Good suggestion! Updated.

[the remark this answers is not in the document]
nhakmiller: Good suggestion! Implemented something similar to this for these purposes.

internal/log_analysis/log_processor/common/metrics.go
nhakmiller: After our conversation on the phone, I decided to look into how adding a `Value` field to the metric struct would look. Unfortunately, with the current implementation, it would add some complications because the reason the `MonoLogger` is faster is that I can pre-create the `MetricDirectiveObject` struct when I initialize the logger by knowing all the metric and dimension names up front. Then I can just re-use this "immutable" struct over and over again for every call to Log. Allowing them to be specified at every call to log means that I can't re-use that struct, so I'll have to generate it each time. It would be pretty easy to do, and might be a nice QOL improvement for the normal logger, but sort of bypasses the point of the `MonoLogger` which is that by specifying everything up front you can re-use the directive object on each subsequent call to `Log`.

I will see about adding it to the normal `Logger` just to make it easier to interact with. Then I will probably merge this PR so we can keep moving forward.

pkg/metrics/logger.go
nhakmiller: Per a suggestion from Kostas I went ahead and added the `Value` field to the `Metric` struct, and just used JSON tags to indicate it was not to be marshaled. This solved the namespace problem neatly, while also allowing communicating with the logger to be a little easier.

pkg/metrics/logger.go
nhakmiller: Went ahead and just added the superset calculation. I decided it was a good pattern to follow, and added dimension checking to `fastLogEmbedded` since it is more straightforward now.

PR #1033, opened 2020-06-12
Redact destination secrets

nhakmiller:
## Background

Previously, configuration secrets such as API keys were displayed in plain text to users when viewing the destination configurations. This change both redacts those secrets before returning them to the frontend, and also allows the frontend to make partial updates to destinations.

Notes to the frontend team: All API calls that return an `AlertOutput` object should now have all secret fields redacted. I redacted them with asterisks `********` but if it would be easier to just return `nil` and have you guys put in a place holder on the frontend let me know. I also changed the signature for `UpdateOutput` API calls, the `displayName`, `defaultForSeverity` and `outputConfig` fields are now optional. Feel free to still send them if you want, but at least for `outputConfig` you need to not send anything if the user didn't change it because you won't know what it's configuration is.

## Changes

- Redact destination configuration secrets before sending them to the frontend
- Allow for partial updates of destinations

## Testing

- unit testing
- deployed to dev account and created a new destination
  - Observed that upon reloading the destinations screen, I was no longer able to see the configuration's API token
  - Went to the AWS lambda console, and manually tested a partial update with the following payload:
```
{
  "UpdateOutput": {
    "userId": "8cf47b32-9e03-4a9f-9c11-704dc55e7287",
    "displayName": "New Name",
    "outputId": "f199b5e4-bfbc-42c6-82f7-a42b1b6b438c",
    "outputConfig": null,
    "defaultForSeverity": null
  }
}
```

### DIFF
--- a/api/lambda/outputs/models/api.go
+++ b/api/lambda/outputs/models/api.go
@@ -81,10 +81,11 @@ type DeleteOutputInput struct {
 // }
 type UpdateOutputInput struct {
 	UserID             *string       `json:"userId" validate:"required,uuid4"`
-	DisplayName        *string       `json:"displayName" validate:"required,min=1,excludesall='<>&\""`
+	DisplayName        *string       `json:"displayName" validate:"min=1,excludesall='<>&\""`
 	OutputID           *string       `json:"outputId" validate:"required,uuid4"`
-	OutputConfig       *OutputConfig `json:"outputConfig" validate:"required"`
+	OutputConfig       *OutputConfig `json:"outputConfig"`
 	DefaultForSeverity []*string     `json:"defaultForSeverity"`
+	Type               *string       `json:"type" validate:"required"`
 }
 
 // UpdateOutputOutput returns the new updated output
--- a/api/lambda/outputs/models/api.go
+++ b/api/lambda/outputs/models/api.go
@@ -85,7 +85,6 @@ type UpdateOutputInput struct {
 	OutputID           *string       `json:"outputId" validate:"required,uuid4"`
 	OutputConfig       *OutputConfig `json:"outputConfig"`
 	DefaultForSeverity []*string     `json:"defaultForSeverity"`
-	Type               *string       `json:"type" validate:"required"`
 }
 
 // UpdateOutputOutput returns the new updated output
--- a/api/lambda/outputs/models/api.go
+++ b/api/lambda/outputs/models/api.go
@@ -22,12 +22,12 @@ package models
 //
 // Exactly one action must be specified.
 type LambdaInput struct {
-	AddOutput            *AddOutputInput            `json:"addOutput"`
-	UpdateOutput         *UpdateOutputInput         `json:"updateOutput"`
-	GetOutput            *GetOutputInput            `json:"getOutput"`
-	DeleteOutput         *DeleteOutputInput         `json:"deleteOutput"`
-	GetOutputs           *GetOutputsInput           `json:"getOutputs"`
-	GetOutputsUnredacted *GetOutputsUnredactedInput `json:"getOutputsUnredacted"`
+	AddOutput             *AddOutputInput             `json:"addOutput"`
+	UpdateOutput          *UpdateOutputInput          `json:"updateOutput"`
+	GetOutput             *GetOutputInput             `json:"getOutput"`
+	DeleteOutput          *DeleteOutputInput          `json:"deleteOutput"`
+	GetOutputs            *GetOutputsInput            `json:"getOutputs"`
+	GetOutputsWithSecrets *GetOutputsWithSecretsInput `json:"getOutputsWithSecrets"`
 }
 
 // AddOutputInput adds a new encrypted alert output to DynamoDB.
@@ -106,7 +106,7 @@ type GetOutputInput struct {
 // GetOutputOutput contains the configuration for an alert
 type GetOutputOutput = AlertOutput
 
-// GetOrganizationOutputsInput fetches all alert output configuration for one organization
+// GetOutputsInput fetches all alert output configuration for one organization
 //
 // Example:
 // {
@@ -116,12 +116,12 @@ type GetOutputOutput = AlertOutput
 type GetOutputsInput struct {
 }
 
-// GetOrganizationOutputsInput fetches all alert output configuration for one organization
+// GetOutputsWithSecretsInput fetches all alert output configuration for one organization
 // without redacting their secrets
-type GetOutputsUnredactedInput struct {
+type GetOutputsWithSecretsInput struct {
 }
 
-// GetOrganizationOutputsOutput returns all the alert outputs for one organization
+// GetOutputsOutput returns all the alert outputs for one organization
 //
 // Example:
 // {
--- a/api/lambda/outputs/models/api.go
+++ b/api/lambda/outputs/models/api.go
@@ -22,11 +22,12 @@ package models
 //
 // Exactly one action must be specified.
 type LambdaInput struct {
-	AddOutput    *AddOutputInput    `json:"addOutput"`
-	UpdateOutput *UpdateOutputInput `json:"updateOutput"`
-	GetOutput    *GetOutputInput    `json:"getOutput"`
-	DeleteOutput *DeleteOutputInput `json:"deleteOutput"`
-	GetOutputs   *GetOutputsInput   `json:"getOutputs"`
+	AddOutput            *AddOutputInput            `json:"addOutput"`
+	UpdateOutput         *UpdateOutputInput         `json:"updateOutput"`
+	GetOutput            *GetOutputInput            `json:"getOutput"`
+	DeleteOutput         *DeleteOutputInput         `json:"deleteOutput"`
+	GetOutputs           *GetOutputsInput           `json:"getOutputs"`
+	GetOutputsUnredacted *GetOutputsUnredactedInput `json:"getOutputsUnredacted"`
 }
 
 // AddOutputInput adds a new encrypted alert output to DynamoDB.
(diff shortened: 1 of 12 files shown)


internal/core/outputs_api/api/update_output.go
```diff
@@ -40,17 +40,11 @@ func (API) UpdateOutput(input *models.UpdateOutputInput) (*models.UpdateOutputOu
 			Message: "A destination with the name" + *input.DisplayName + " already exists, please choose another display name"}
 	}
 
-	outputType, err := getOutputType(input.OutputConfig)
```
nhakmiller, line 43: I no longer specify the type when doing an update, the assumption is you cannot update a destination and change its type. Is that a fair assumption? It doesn't seem to be possible via the UI currently.

internal/core/outputs_api/api/utils.go
```diff
@@ -27,6 +27,8 @@ import (
 	"github.com/panther-labs/panther/internal/core/outputs_api/table"
 )
 
+var redacted = aws.String("********")
```
nhakmiller, line 30: Just putting this here so it's explicit, we could instead return nil for this field if that's easier for the frontend to interpret.

internal/core/outputs_api/api/utils.go
```diff
@@ -64,16 +66,45 @@ func ItemToAlertOutput(input *table.AlertOutputItem) (alertOutput *models.AlertO
...
 	if err != nil {
 		return nil, err
 	}
 
+	// Redact the output before returning to the caller
+	redactOutput(alertOutput.OutputConfig)
 	return alertOutput, nil
 }
 
+func redactOutput(outputConfig *models.OutputConfig) {
```
nhakmiller, line 81: I considered using JSON tags on the `OutputConfig` struct so that it would just always omit these fields, but we use the same struct for going from table item -> object we're sending to the frontend as we use for object the frontend has sent us -> table item. And we wouldn't want to omit these fields of course when saving to dynamo.

nhakmiller:
For some reason I thought that the `alert_delivery` lambda communicated to the outputs dynamo table directly, checking now I see that is not the case. Update incoming so that this doesn't break the `alert_delivery` mechanism.

nhakmiller:
> For some reason I thought that the `alert_delivery` lambda communicated to the outputs dynamo table directly, checking now I see that is not the case. Update incoming so that this doesn't break the `alert_delivery` mechanism.

Ok the new version provides a `GetOutputsUnredacted` endpoint for the backend to use, which returns the unredacted configuration. Talked with @austinbyers and we think it should be impossible for a user to get here from the web UI>

internal/core/outputs_api/api/update_output.go
kostaspap: Yeah, it's fair to assume that the type doesn't get updated.

nhakmiller:
> Have you tried updating an output through the UI e.g. in order to update the severity? I think this will break alert-delivery since it will store `********` as credentials.

The frontend won't work but we are tracking that separately, the frontend team is aware of the changes coming.

internal/core/outputs_api/table/update.go
```diff
@@ -30,13 +30,20 @@ import (
...
 	updateExpression := expression.
-		Set(expression.Name("displayName"), expression.Value(alertOutput.DisplayName)).
 		Set(expression.Name("lastModifiedBy"), expression.Value(alertOutput.LastModifiedBy)).
-		Set(expression.Name("lastModifiedTime"), expression.Value(alertOutput.LastModifiedTime)).
-		Set(expression.Name("outputType"), expression.Value(alertOutput.OutputType)).
-		Set(expression.Name("encryptedConfig"), expression.Value(alertOutput.EncryptedConfig)).
-		Set(expression.Name("defaultForSeverity"), expression.Value(alertOutput.DefaultForSeverity))
+		Set(expression.Name("lastModifiedTime"), expression.Value(alertOutput.LastModifiedTime))
+
+	if alertOutput.DisplayName != nil {
```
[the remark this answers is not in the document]
nhakmiller: Currently the display name may never be null, but I figured since we are making parameters optional we should just make all parameters except the ID optional. That way the frontend only has to send the fields that changed if they want to, as opposed to having special logic to send all fields except the config unless the config changed.

Good callout on the empty list,  @3nvi or @alexmylonas can you confirm that the frontend would prefer to send an empty list instead of a null value for a field like `DefaultForSeverity`?

internal/core/outputs_api/api/utils.go
nhakmiller: Updated.

internal/core/outputs_api/api/add_output.go
```diff
@@ -62,6 +62,7 @@ func (API) AddOutput(input *models.AddOutputInput) (*models.AddOutputOutput, err
 	if err != nil {
 		return nil, err
 	}
+	redactOutput(alertOutput.OutputConfig)
```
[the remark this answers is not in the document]
nhakmiller: Updated.

api/lambda/outputs/models/api.go
```diff
@@ -22,11 +22,12 @@ package models
...
-	UpdateOutput *UpdateOutputInput `json:"updateOutput"`
-	GetOutput    *GetOutputInput    `json:"getOutput"`
-	DeleteOutput *DeleteOutputInput `json:"deleteOutput"`
-	GetOutputs   *GetOutputsInput   `json:"getOutputs"`
+	AddOutput            *AddOutputInput            `json:"addOutput"`
+	UpdateOutput         *UpdateOutputInput         `json:"updateOutput"`
+	GetOutput            *GetOutputInput            `json:"getOutput"`
+	DeleteOutput         *DeleteOutputInput         `json:"deleteOutput"`
+	GetOutputs           *GetOutputsInput           `json:"getOutputs"`
+	GetOutputsUnredacted *GetOutputsUnredactedInput `json:"getOutputsUnredacted"`
```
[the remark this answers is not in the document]
nhakmiller: Interesting, I've never heard of this maxim before. I will make this change if you think it would increase code clarity.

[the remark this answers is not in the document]
kostaspap: It's more important to avoid negation when naming variables (like boolean variables) but I think it adds clarity here as well. 
I think having `GetOutputs` and `GetOutputsRedacted` might somehow be ever more clear, but I am not sure it worth the refactoring effort now. Just renaming it should be fairly easy I think.

internal/core/outputs_api/api/utils.go
```diff
@@ -21,12 +21,13 @@ package api
...
 	"errors"
 
-	"github.com/aws/aws-sdk-go/aws"
-
 	"github.com/panther-labs/panther/api/lambda/outputs/models"
 	"github.com/panther-labs/panther/internal/core/outputs_api/table"
+	"github.com/panther-labs/panther/pkg/box"
 )
 
+var redacted = box.String("********")
```
austinbyers, line 29: @kostaspap why use both this library *and* AWS? My understanding was that we only use our box/unbox library for services that don't otherwise depend on the AWS SDK to avoid introducing a new dependency.

But this service (and virtually all of our Lambda functions) already have to import the AWS SDK, so I would argue we should just keep `aws.String` and not introduce a new custom package.

nhakmiller: Hmm good point Austin. I removed all calls to `aws.String(...)` in this particular file so it no longer needs to import the AWS SDK, but there are other places in the `api` package that still import the AWS SDK and certainly the related `table` package is going to have to import the AWS SDK to talk to dynamo no matter what. I'll revert these changes for now, and if at some point in the future we decide to remove all usages of the AWS SDK from the `api` package here we can do so in a more concerted effort.

I will keep my eye out though for other services that don't need to import the AWS SDK for anything except string wrapping/unwrapping and convert those to using the `box` package as appropriate.

internal/core/outputs_api/api/utils.go
alxarch: One solution to this problem would be to define a 'special' string type (ie `RedactedString`) and have a custom jsoniter config that encodes all `RedactedString` as empty/padded. You would then use the 'normal' jsoniter config to serialize JSON for the DB and use the 'redacted' one for serializing values for the API.

nhakmiller: Ah I like that very smart! For just this one secret I think that may be overkill, but if we find ourselves repeatedly needing to redact secrets then perhaps we could add some support for this to `pkg`. I plan to add secret redaction to one other place in enterprise currently, so maybe I can look into it for that and backport the changes if it makes sense.

internal/core/outputs_api/api/utils.go
kostaspap: The idea behind having the `box`/`unbox` libraries is that for doing simple tasks like e.g. conversion of a `string` to `*string`, we shouldn't need a dependency to the AWS SDK. 

AWS SDK is client library to talk to AWS services, so the parts the code that interact with such AWS services will have a dependency on it by definition and they may use the AWS SDK libraries as they wish. Big part of our code, however,  just implements some business logic and doesn't deal with AWS at all. That code, we can use the `box`/`unbox` libraries. The `box`/`unbox` library is really tiny library anyway. 

We have been discussing for a while that Panther might need to run in a different cloud in the future. This will be a big effort of course (if we ever need to do this), but it's anyway good practice to isolate parts of the code that deal with AWS but either putting appropriate layers (e.g. DDB table abstractions) and just avoiding having AWS SDK dependencies in parts of the code that don't really need it.

austinbyers: I don't think it's helpful right now to make a distinction between which parts of the same Lambda function do or do not talk to AWS. That just depends on the package structure of the particular function and could easily change.

Using the `box` library in one pkg and the `aws` unboxing in another package in the same Lambda function adds zero value in my opinion. It will make future refactors more difficult.

In other words, in my opinion, we should be using it everywhere or nowhere in a function.

kostaspap: > Using the `box` library in one pkg and the `aws` unboxing in another package in the same Lambda function adds zero value in my opinion. It will make future refactors more difficult.
> 
> In other words, in my opinion, we should be using it everywhere or nowhere in a function.

This is fair. I would personally prefer to use `box`/`unbox` for the reasons stated above .If you feel otherwise we can just bring this up for discussion with the team.

PR #1034, opened 2020-06-14
Improve log processor assume role usage

rleighton:
## Before
When processing an s3 object stscreds.NewCredentials() was called each time.

## After
Now  stscreds.NewCredentials() is called only when needed.

## Testing

- mage deploy, checked for any errors (were none)

### DIFF
--- a/internal/log_analysis/log_processor/sources/s3_client_cache.go
+++ b/internal/log_analysis/log_processor/sources/s3_client_cache.go
@@ -121,31 +121,26 @@ func getS3Client(s3Object *S3ObjectInfo) (s3iface.S3API, string, error) {
 			awsRegion: bucketRegion.(string),
 		}
 
-		var client interface{}
-		client, ok := s3ClientCache.Get(cacheKey)
-		if !ok {
+		if client, ok := s3ClientCache.Get(cacheKey); !ok {
 			zap.L().Debug("s3 client was not cached, creating it")
 			client = newS3ClientFunc(box.String(cacheKey.awsRegion), awsCreds)
 			s3ClientCache.Add(cacheKey, client)
+		} else {
+			return client.(s3iface.S3API), *sourceInfo.IntegrationType, nil
 		}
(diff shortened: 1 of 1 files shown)


internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -156,6 +163,7 @@ func getAwsCredentials(roleArn string) *credentials.Credentials {
 	zap.L().Debug("fetching new credentials from assumed role", zap.String("roleArn", roleArn))
 	return newCredentialsFunc(common.Session, roleArn, func(p *stscreds.AssumeRoleProvider) {
 		p.Duration = time.Duration(sessionDurationSeconds) * time.Second
+		p.ExpiryWindow = p.Duration / 2 // give plenty of time to refresh
```
kostaspap, line 166: This is too much, current duration is 60 minutes so this will cause the credentials to be renewed every 30', essentially beating the purpose of having 60' credential duration.  
Usually the default works well (have we seen`ExpiredTokenException` exceptions?), but if we have seen issue it would be fine to pump it few seconds.

rleighton: Yeah, I added that because when I made the first change I would see exceptions (a few every few hours).  How about instead of 1/2 way, I set it to expire 90% of the way?

kostaspap: This one doesn't have to be percentile since it's there to prevent a race conditions (when we invoke an API before the credentials have been updated)... It's not really related to the duration e.g. if we increased the duration to 2 hours, it wouldn't have to double . I think a value like 30seconds or 1' or 5'  or something should be enough  (refreshing the credentials is just an API call that should normally complete within few seconds max).

rleighton: I'll set it to 1min and see if there are no errors.

PR #1037, opened 2020-06-15
Bump Eslint & Backport Enterprise changes

3nvi:
## Background

This PR fixes issues that  we've had in our local IDE caused by the bumping of prettier to v2. Specifically, Eslint didn't correctly understand the new default configuration  of prettier v2, giving false positives in the IDE. In addition, eslint checks will now be quicker.

While there, I also backported some changes from Enterprise to smooth out future integrations

## Changes

- Bump all Eslint related packages
- Update Eslint rules & configuration
- Backport Enterprise code changes to allow for lesser conflicts

## Testing

- CI
- Visually everything looks (and should look) the same


package.json
```diff
@@ -105,7 +105,7 @@
     "history": "^4.10.1",
     "linkifyjs": "^2.1.9",
     "lodash-es": "^4.17.15",
-    "pouncejs": "0.19.0",
+    "pouncejs": "^0.19.0",
```
alexmylonas, line 108: we are ok with this right?

3nvi: oh wow, nice catch! No we are certainly  not. Thanks for noticing!

PR #1040, opened 2020-06-15
Standardize CFN parameters

austinbyers:
## Background

There are some inconsistencies in how we handle CloudFormation parameters that have almost led to very painful bugs a few times. We have to maintain 2 deployment methods for 2 different Panther editions across 9 stacks - it's too easy to make a mistake like forgetting a parameter.

Besides maintaining my own sanity, these changes make testing releases easier by catching some kinds of bugs sooner.

## Changes
Think of this as essentially enforcing a new style guide specifically for CloudFormation parameters.

1. **Do not include default values for stack parameters except for the top-level stack.** 

Defaults are specified either in the config file (when deploying from source) or in the master template (when using a pre-packaged deployment). Putting defaults in the nested stacks is confusing and also makes it possible to add a parameter but forget to pass that parameter when deploying the template. This kind of bug is very hard to catch because the deploy will succeed but with an incorrect parameter value.

This way, all parameters are required on every deploy. If you forget to pass a parameter, the deploy will fail loud and early. This is particularly helpful for the master template because developers typically are not interacting with it - release testing will now immediately catch missing parameters.

2. **Use mappings instead of parameters for constants**

This is related to the first issue. A few stacks had "parameters" with default values, but no support for changing those parameters from the settings file / master stack:

* `log-analysis` - alert retry duration, min/max retry delay, retention
* `web` - ECS cpu, memory, port

Changing the default value of a parameter does *not* update that value in an existing stack. You have to explicitly pass the new value. A "parameter" should be user-configurable. Either make it configurable at the top level, or make it a "constant" (mapping).

I chose to move these settings into mappings, copying the work we've already done with Lambda memory/timeout.

3. **Always alphabetize parameter lists**

Some stacks have as many as 30 parameters, they need to be in the same order in the master template, the nested stacks, and mage so their code can be read side-by-side. The easiest consistent ordering is lexicographic.

This also matches the ordering in the CloudFormation console and will make enterprise merging easier. Enterprise-specific parameters will be alphabetized in a separate section below the open-source parameter sets

## Testing

- `mage deploy test:integration`
- `mage master:deploy`


deployments/appsync.yml
```diff
@@ -21,19 +21,15 @@ Description: AppSync data sources and resolvers
 # This template adds all the configuration: schema, data sources, resolvers
 
 Parameters:
```
rleighton, line 23: How about a comment line in every file saying something "All parameters must be in lexicographic order with no defaults values"

deployments/bootstrap.yml
```diff
@@ -30,49 +30,47 @@ Description: The very first Panther stack - static resources which don't require
...
-  LogSubscriptionPrincipals:
-    Type: CommaDelimitedList
-    Description: The list of Principal ARNs to allow read access to the ProcessedDataBucket and subscribe access to ProcessedDataTopicArn
-    Default: ''
-  EnableS3AccessLogs:
-    Type: String
-    Description: Enable S3 access logging for all Panther buckets.
-    AllowedValues: [true, false]
-    Default: true
+  # Alphabetize parameters so it's easy to compare side-by-side with other code that passes values
```
rleighton, line 33: Add to all files.

deployments/core.yml
```diff
@@ -19,102 +19,77 @@ Transform: AWS::Serverless-2016-10-31
...
-    Type: Number
-    Description: Number of seconds SQS will retain a message in the alerts queue
-    Default: 259200 # 3 days
-    MinValue: 60
-    MaxValue: 1209600
+  UserPoolId:
+    Type: String
+    Description: Cognito user pool ID
 
 Mappings:
```
rleighton, line 82: Add comment to all files "All constants should be defined here"

deployments/appsync.yml
austinbyers: I actually realized I could add a CI check to enforce the "no defaults", so that's what I did.

Lexicographic order is a personal preference, doesn't have to be enforced (and won't be true in the enterprise, either). I care more about grouping by OSS/enterprise than their exact ordering. But I will go ahead and copy the "alphabetize" text in each template

deployments/core.yml
austinbyers: Since CI enforces no defaults now, this won't be necessary (the error message will explain to use a mapping)

PR #1041, opened 2020-06-15
Fix outputs-api integration tests

austinbyers:
## Background

#1033 broke the `outputs-api` integration tests

## Changes

- Fix `outputs-api` integration tests
- Change `mage test:integration` so that errors are not blocking

It takes several minutes to run all integration tests, a failure halfway through is frustrating because `mage` will stop immediately and you'll have no way to know if the remaining tests would have passed.

Now, all errors are logged so you can address all integration test failures at once

## Testing

- `mage test:integration` 3 times plus `PKG=./internal/core/outputs_api/main mage test:integration` a few times

### DIFF
--- a/internal/core/outputs_api/main/integration_test.go
+++ b/internal/core/outputs_api/main/integration_test.go
@@ -254,11 +254,6 @@ func updateSlack(t *testing.T) {
 	var output models.UpdateOutputOutput
 	require.NoError(t, genericapi.Invoke(lambdaClient, outputsAPI, &input, &output))
 
-	assert.Equal(t, aws.StringSlice([]string{"CRITICAL"}), output.DefaultForSeverity)
-	assert.Equal(t, slackOutputID, output.OutputID)
-	assert.Equal(t, aws.String("alert-channel-new"), output.DisplayName)
-	assert.Equal(t, aws.String("slack"), output.OutputType)
-
 	expected := models.UpdateOutputOutput{
 		CreatedBy:          userID,
 		CreationTime:       output.CreationTime,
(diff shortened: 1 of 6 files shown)


internal/core/outputs_api/main/integration_test.go
```diff
@@ -240,29 +239,6 @@ func updateInvalid(t *testing.T) {
 	assert.Equal(t, expected, err)
 }
 
-func updateSlack(t *testing.T) {
-	t.Parallel()
-	slack.WebhookURL = aws.String("https://hooks.slack.com/services/DDDDDDDDD/EEEEEEEEE/" +
```
austinbyers, line 245: My understanding is that this field can no longer be realistically changed by the `update` endpoint. Should this integration test expect a specific failure? @nhakmiller 

We still have an `updateSns` integration test, we don't need one for every output type

nhakmiller: This field can still be changed by the `update` endpoint, but the response will redact it. Was there an issue in doing the update?

nhakmiller: In fact a good test here could be doing a partial update, specifying everything except the secret field, doing an update, then doing a `getOutputsWithSecrets` API call to verify that the partial update didn't delete the secrets.

austinbyers: Ok I'll change the update to just change the display name and add tests for `getOutputsWithSecrets`

austinbyers:
> Very sorry for this, I never think about integration tests. I will remind myself to be more thoughtful about them in the future.

Haha it's fine, none of the reviewers caught it either (myself included). I was actually working on the non-blocking errors first and just happened to notice the failing tests

PR #1043, opened 2020-06-16
Secrets manager permissions

nhakmiller:
## Background

Give the policy and rules engines permissions to communicate to AWS secrets manager.

I gave the two engines access to different 'paths' in secrets manager, because we separate permissions for policy/rule writing so technically one could have a secret that the other shouldn't have access to. But generally any rule can access any secret put in by any other rule.

They also have a shared path for shared secrets.

We're considering adding something to the `panther_analysis_tool` to assist in uploading these secrets, so you can securely store the initial secret. But currently you'll have to load the secret in from the AWS console.

## Changes

- Updated permissions of policy/rules engines

## Testing

- Doing a deployment to dev account currently to test this

### DIFF
--- a/deployments/cloud_security.yml
+++ b/deployments/cloud_security.yml
@@ -1238,6 +1238,12 @@ Resources:
             - Effect: Allow
               Action: dynamodb:*Item
               Resource: !Sub arn:${AWS::Partition}:dynamodb:${AWS::Region}:${AWS::AccountId}:table/panther-kv-store
+        - Id: GetPantherPolicySecrets
+          Version: 2012-10-17
+          Statement:
+            - Effect: Allow
+              Action: secretsmanager:GetSecretValue
+              Resource: !Sub arn:aws:secretsmanager:*:${AWS::AccountId}:secret:panther-policies/*
 
   PolicyEngineLogGroup:
     Type: AWS::Logs::LogGroup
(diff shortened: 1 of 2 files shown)


deployments/cloud_security.yml
```diff
@@ -1238,6 +1238,18 @@ Resources:
             - Effect: Allow
               Action: dynamodb:*Item
               Resource: !Sub arn:${AWS::Partition}:dynamodb:${AWS::Region}:${AWS::AccountId}:table/panther-kv-store
+        - Id: GetPantherPolicySecrets
+          Version: 2012-10-17
+          Statement:
+            - Effect: Allow
+              Action: secretsmanager:GetSecretValue
+              Resource: !Sub arn:aws:secretsmanager:*:${AWS::AccountId}:secret:panther-policies/*
```
austinbyers, line 1246: I don't think you can access secrets cross-region, in which case there's no reason not to interpolate the region here. Also we should add the partition to be standard with our other code.

```suggestion
              Resource: !Sub arn:${AWS::Partition}:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:panther-policies/*
```

nhakmiller:
Verified this is working by deploying to my dev account and testing the following policy (after manually creating the secret key/pair named `panther-policies/secretKey` with contents `{'asdf': 'abcXYZ'}` via the console):

```
import boto3
import json

def policy(resource):
	name = "panther-policies/secretKey"
	region = 'us-west-2'
	session = boto3.session.Session()
	client = session.client(service_name='secretsmanager', region_name=region)
	secret_response = client.get_secret_value(SecretId=name)
	return json.loads(secret_response['SecretString'])['asdf'] == 'abcXYZ'
```

Testing it returned `True`, then changing the last line to something different returned `False`.

PR #1045, opened 2020-06-17
Hiding secrets and sensitive information for Destinations in UI

### LINKED ISSUE #1009: Hide secrets in the web app

When viewing existing sources or destinations, secrets such as the API token or web-hook URL are visible from within the web UI. They should not visible from within the UI, and should in fact not even be sent to the UI to display.

### Describe your use cases

Not displaying the secrets in the web UI will prevent shoulder surfing attacks on the secret keys. Not returning the API token at all will prevent administrators from recovering the key, which is the preferred way to treat secrets like API tokens.

### Additional context

Screen Shot 2020-06-08 at 5 56 28 PM

- This token should be hidden

alexmylonas:
## Background

As described in more detail on issue #1009: 
1. We should hide sensitive information on forms when creating a destination to prevent shoulder surfing attacks
2. We should avoid sharing sensitive information between UI and the backend whenever it's feasible -- part of this covered in the backend by replacing secrets with empty strings.

This PR handles the first point and converts all sensitive input fields to password so they aren't visible on input. For the second point, in order to avoid sharing back and forth info with the backend, we show empty fields with proper placeholder for sensitive fields but we allows users to input a new valid one when updating a destination.

Closes #1009 

## Changes

- Updated schema to include different types for `AddDestination` & `EditDestination`
- Updated:
1. `AsanaDestinationForm`
2. `CustomWebhookDestinationForm`
3. `GithubDestinationForm`
4. `JiraDestinationForm`
5. `MicrosoftDestinationForm`
6. `OpsgenieDestinationForm`
7. `PagedutyDestinationForm`
8. `SlackDestinationForm`
- Minor updates to other forms to handle new types
- Includes some backend fixes by @nhakmiller 

## Testing

QA' ed all Destinations that creation and editing works as intented
- Locally
- Deployed Env


nhakmiller:
> If for some reason user needs to update API keys, tokens etc, they are required to delete the destination and create a new one.

This should not be required, you can just send the updated key back to us. Did you run into issues with this?

api/lambda/outputs/models/api.go
```diff
@@ -196,26 +196,72 @@ type OutputConfig struct {
 	CustomWebhook *CustomWebhookConfig `json:"customWebhook,omitempty"`
 }
 
+// OutputConfigUpdate contains the configuration for the output
+type OutputConfigUpdate struct {
```
nhakmiller, line 200: Discussed with @austinbyers , going to refactor this out so it's backed to a shared struct again.

internal/core/outputs_api/api/utils.go
```diff
@@ -137,3 +137,79 @@ func getOutputType(outputConfig *models.OutputConfig) (*string, error) {
 
 	return nil, errors.New("no valid output configuration specified for alert output")
 }
+
+// mergeOuutputConfigs takes two outputConfigs, and condenses them into one based on updating
+func mergeOutputConfigs(old *models.OutputConfig, new *models.OutputConfigUpdate) *models.OutputConfig {
```
nhakmiller, line 142: Going to refactor this to cast everything into a byte slice then a map so we can just iterate through regardless of type. This will allow us to to forward compatible with new fields or types.

nhakmiller:
I fairly extensively re-factored the backend changes from this morning, but I also extended the integration tests and did quite a bit of manual testing. I noticed that you're blocking users from updating the secrets in the frontend, that is not necessary. If a user does not update a secret, then  the frontend shouldn't send its contents during an update (because the frontend does not know the contents). But if the user does decide to update the secret, that's totally fine.

nhakmiller:
Tagging Kostas and Austin as well because what started out as a minor tweak to your branch to fix a bug turned into some moderate backend re-factorings. Spent a lot of time testing the changes this time though, through the UI and the AWS console and for several different types of destinations.

api/lambda/outputs/models/api.go
```diff
@@ -198,59 +198,59 @@ type OutputConfig struct {
 
 // SlackConfig defines options for each Slack output.
 type SlackConfig struct {
-	WebhookURL *string `json:"webhookURL" validate:"required,url"` // https://hooks.slack.com/services/...
+	WebhookURL *string `json:"webhookURL" validate:"omitempty,url"` // https://hooks.slack.com/services/...
```
nhakmiller, line 201: We're changing these from `required` to `omitempty` because they are only required on create operations, not on update operations (due to hiding some fields from the frontend). There is some supplementary validation done in the `AddOutput` endpoint direcltly.

internal/core/outputs_api/main/integration_test.go
```diff
@@ -110,6 +117,15 @@ func TestIntegrationAPI(t *testing.T) {
 		return
 	}
 
+	// Perform a partial update and then verify that secrets aren't changed
+	t.Run("PartialUpdate", func(t *testing.T) {
+		t.Run("PartialUpdateDisplayName", partialUpdateDisplayName)
+		t.Run("PartialUpdateConfig", partialUpdateConfig)
+	})
```
nhakmiller, line 124: @austinbyers these are for you :)

alexmylonas:
> This should not be required, you can just send the updated key back to us. Did you run into issues with this?

It's a matter of ambiguity, its a bit strange to allow users to edit a field that has "wrong" data even if we are display them. We also need to remove them from the request and forms requirements conditionally. Therefore we disable editing for those fields. 
I am pretty ok with this approach, @3nvi what do you think?

api/lambda/outputs/models/api.go
3nvi: It may have been better  to split  it into `AddSlackConfig` and  `UpdateSlackConfig`, but I  may be missing context

web/src/components/sidesheets/UpdateDestinationSidesheet/UpdateDestinationSidesheet.tsx
```diff
@@ -69,9 +69,18 @@ export const UpdateDestinationSidesheet: React.FC<UpdateDestinationSidesheetProp
   });
 
   const handleSubmit = React.useCallback(
-    async (values: BaseDestinationFormValues<Partial<DestinationConfigInput>>) => {
+    async (values: BaseDestinationFormValues<Partial<AddDestinationConfigInput>>) => {
       const { displayName, defaultForSeverity, outputConfig } = values;
 
+      delete outputConfig.slack;
+      delete outputConfig.pagerDuty;
```
3nvi, line 76: I'm personally not really fond of that.

@nhakmiller can we just "ignore" the sensitive fields on the update? I  want to isolate business logic on dedacted data to a single unit (i.e. the lambda)

alexmylonas:
> With regards to form validation, a - potentially clever - way of doing that is to not only allow the proper shape of the secrets, but also consider as "valid" the string of `****`. This way you don't have to create separate validation schemas for adding & editing.

There is a slight difference between form validation and condition of stripping from the request.
i.e. For Pagerduty token must be 32 chars, the form must allow 32 chars & '****' (or only 32 chars and ignore field validation on update as its currently implemented) but the updateHandler should strip the token if it matches '*****'.

3nvi:
> > With regards to form validation, a - potentially clever - way of doing that is to not only allow the proper shape of the secrets, but also consider as "valid" the string of `****`. This way you don't have to create separate validation schemas for adding & editing.
> 
> @3nvi There is a slight difference between form validation and condition of stripping from the request.
> i.e. For Pagerduty token must be 32 chars, the form must allow 32 chars & '****' (or only 32 chars and ignore field validation on update as its currently implemented) but the updateHandler should strip the token if it matches '*****'.

@alexmylonas gotcha! In my mind the backend should return data **in the same shape as it was** but with the content replaced with `*` charactes (apart from some  at the end, which should remain pristine for visibility reasons)

web/src/components/forms/BaseDestinationForm/BaseDestinationForm.tsx
```diff
@@ -27,7 +27,7 @@ import FormikCheckbox from 'Components/fields/Checkbox';
 import SeverityBadge from 'Components/SeverityBadge';
 
 export interface BaseDestinationFormValues<
-  AdditionalValues extends Partial<DestinationConfigInput>
+  AdditionalValues extends Partial<AddDestinationConfigInput>
```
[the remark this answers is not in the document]
alexmylonas: Updated!

web/src/components/forms/AsanaDestinationForm/AsanaDestinationForm.tsx
```diff
@@ -29,36 +29,48 @@ import { isNumber } from 'Helpers/utils';
...
-// https://github.com/jquense/yup/issues/522
-const mergedValidationSchema = defaultValidationSchema.concat(asanaFieldsValidationSchema);
+const editGithubFieldsValidationSchema = Yup.object().shape({
+  outputConfig: Yup.object().shape({
+    asana: Yup.object().shape(baseAsanaShapeObject),
+  }),
+});
 
 const AsanaDestinationForm: React.FC<AsanaDestinationFormProps> = ({ onSubmit, initialValues }) => {
+  const existing = initialValues.displayName.length;
```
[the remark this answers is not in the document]
alexmylonas: Correct, i will be using outputId!

web/src/components/forms/AsanaDestinationForm/AsanaDestinationForm.tsx
```diff
@@ -29,36 +29,48 @@ import { isNumber } from 'Helpers/utils';
...
-      projectGids: Yup.array().of(Yup.number()).required(),
     }),
   }),
 });
 
-// We merge the two schemas together: the one deriving from the common fields, plus the custom
-// ones that change for each destination.
-// https://github.com/jquense/yup/issues/522
-const mergedValidationSchema = defaultValidationSchema.concat(asanaFieldsValidationSchema);
+const editGithubFieldsValidationSchema = Yup.object().shape({
```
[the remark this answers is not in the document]
alexmylonas: nice catch!

api/lambda/outputs/models/api.go
nhakmiller: That was my first implementation, but it was a lot of boilerplate and this seemed easier to maintain.

alexmylonas:
@3nvi PR is updated based on our last discussion with @nhakmiller!

We agreed that the backend returns empty string for sensitive info and also accepts empty string as null and does'nt update secrets.

Code looks much cleaner, more maintainable and in a better direction, good job everyone!

web/src/components/forms/AsanaDestinationForm/AsanaDestinationForm.tsx
```diff
@@ -59,9 +58,10 @@ const AsanaDestinationForm: React.FC<AsanaDestinationFormProps> = ({ onSubmit, i
     >
       <Field
         as={FormikTextInput}
+        type="password"
         name="outputConfig.asana.personalAccessToken"
         label="Access Token"
-        placeholder="Your personal Asana access token"
+        placeholder={existing ? '<hidden information>' : 'Your personal Asana access token'}
         mb={6}
         aria-required
```
3nvi, line 66: Let's modify the `aria-required` depending on the the value of `existing`

web/src/helpers/utils.tsx
```diff
@@ -65,6 +65,7 @@ export const integrationLabelValidation = () =>
     .matches(SOURCE_LABEL_REGEX, 'Can only include alphanumeric characters, dashes and spaces')
     .max(32, 'Must be at most 32 characters');
 
+export const webhookValidation = () => Yup.string().url('Must be a valid webhook URL');
```
3nvi, line 68: I don't think we need that in `utils`. We can just inline the related code IMO

web/src/components/forms/AsanaDestinationForm/AsanaDestinationForm.tsx
3nvi: (in all related forms)

web/src/helpers/utils.tsx
3nvi: Update: Just saw that we use  it in multiple places. Thought it was only for the custom webhook. Ignore my comment :)

web/src/components/forms/AsanaDestinationForm/AsanaDestinationForm.tsx
alexmylonas: Updated!

PR #1046, opened 2020-06-17
Log processor performance tweaks

rleighton:
## Background
This PR is the result of stress testing the log processor.

There are 2 important changes in this PR:
 - bound the number of files processed per call
 - under high memory usage slow/stop reading more files

Under _extreme_ high load of many very small files the log processor could timeout on deleting processed files from the queue. I set the file limit per call to 5000, but the important change is not the number (perhaps it could be higher), just that this is bounded.

Under _extreme_ high load of os query data I was able to OOM the log processor. I added a test for memory usage and slowed/stopped reading new files until heap usage went below 80%.

NOTE: it appears that the 1000 limit on concurrent lambda executions is by account+region and not account+region+function because I was able to generate throttles on OTHER lambdas by overloading the log processor lambda. Good to know.

## Changes

- Add limit to number of files processed by log processor per call 
- High memory back pressure
- Adjust RulesEngine lambda timeout to max
- Adjust catalog updater lambda mem to 512 from 128 to improve speed
- Fix some comments in s3.go
- Add comment to panther_config.yml about raising lambda limit.

## Testing

- mage test:ci
- did load testing in demo account and also snowflake saas account


alxarch:
The fix seems to address the issue described correctly. How does this affect the 'low volume' scenario where we need to force denser packing of s3 files by 'polling' for more events? The `0.5` deadline allocation for processing seems a bit excessive. Does this affect only the 'polling' period for batching log event buffers?  
Just noting that if we proceed with using context.Context to handle timing and cancelation of goroutines in the future (instead of passing deadlines around), blocking channels could help with rate limiting the event poller to automatically handle backpressure when the last event is not yet processed.

rleighton:
> The fix seems to address the issue described correctly. How does this affect the 'low volume' scenario where we need to force denser packing of s3 files by 'polling' for more events? The `0.5` deadline allocation for processing seems a bit excessive. Does this affect only the 'polling' period for batching log event buffers?
> Just noting that if we proceed with using context.Context to handle timing and cancelation of goroutines in the future (instead of passing deadlines around), blocking channels could help with rate limiting the event poller to automatically handle backpressure when the last event is not yet processed.

These changes have no effect on the low volume case.

rleighton:
> The fix seems to address the issue described correctly. How does this affect the 'low volume' scenario where we need to force denser packing of s3 files by 'polling' for more events? The `0.5` deadline allocation for processing seems a bit excessive. Does this affect only the 'polling' period for batching log event buffers?
> Just noting that if we proceed with using context.Context to handle timing and cancelation of goroutines in the future (instead of passing deadlines around), blocking channels could help with rate limiting the event poller to automatically handle backpressure when the last event is not yet processed.

These changes have no effect on the low volume case.

rleighton:
I'm going to wait to push this until the release is done

PR #1047, opened 2020-06-17
Standardizing the fields sent for PagerDuty/Sns/Sqs

### LINKED ISSUE #1044: SNS/SQS Alert Outputs Should Have the Generated Alert Titles

SNS/SQS Alert Outputs Should Have the Generated Alert Titles

### Describe your use cases

Forwarding to automation

Daily

kostaspap:
## Background

Fixes #1044 . 
I verified that we include link to Panther UI in all destinations. 

## Changes

- Fixing a long standing issue in the AlertDelivery API. The `Alert` model was referring to `PolicyName`, `PolicyID`, `PolicyDescription` while the field could actually be containing the Rule name or Rule ID. @s0l0ist ran into it recently.
I renamed it to `AnalysisID`, `AnalysisName`, `AnalysisDescription`. @nhakmiller introduced the term, I was thinking to stick with it. 
- Further updated the `Alert` so that it doesn't contain `*string`, `[]*string` etc but rather `string`, `[]string`, etc for required fields. This simplifies production and testing code A LOT. 
- @s0l0ist introduced a `CustomWebhookOutputMessage` as part of his recent work. I "store" that and renamed it to `outputs.Notification` (feel free to suggestion different name). `outputs.Notification` is sent as payload to a Webhook destination, to SNS, SQS and PagerDuty. 
As a result some of the SQS/SNS messages now have slightly different structure. I am thinking it would be sufficient to notify the users in the release notes about that. @austinbyers what do you think?

## Testing

- [x] Unit tests
- [x] SQS test
```
{"analysisId":"TestRule","createdAt":"2020-06-17T16:51:37Z","severity":"INFO","type":"RULE","link":"https://web-1351808492.eu-west-1.elb.amazonaws.com/log-analysis/alerts/41f2eae42bec7978f18621ef750c9eb5","title":"New Alert: My cool dynamically generated title","analysisName":null,"alertId":"41f2eae42bec7978f18621ef750c9eb5","description":"","runbook":"","tags":[],"version":"OnXavV1r0gnQncfDuJpqCNIB46_3FFRy"}
```
- [x] SNS test
```
{
   "Type":"Notification",
   "MessageId":"1484ed7c-45c8-5bae-9dfa-ba180992100c",
   "TopicArn":"arn:aws:sns:eu-west-1:1234567890:test",
   "Subject":"New Alert: My cool dynamically generated title",
   "Message":"{\"analysisId\":\"TestRule\",\"createdAt\":\"2020-06-17T16:51:37Z\",\"severity\":\"INFO\",\"type\":\"RULE\",\"link\":\"https://web-XXXXXXX.eu-west-1.elb.amazonaws.com/log-analysis/alerts/41f2eae42bec7978f18621ef750c9eb5\",\"title\":\"New Alert: My cool dynamically generated title\",\"analysisName\":null,\"alertId\":\"41f2eae42bec7978f18621ef750c9eb5\",\"description\":\"\",\"runbook\":\"\",\"tags\":[],\"version\":\"OnXavV1r0gnQncfDuJpqCNIB46_3FFRy\"}",
   "Timestamp":"2020-06-17T16:51:38.245Z",
   "SignatureVersion":"1",
   "Signature":"xxxxxx"
   "SigningCertURL":"XXXXX",
   "UnsubscribeURL":"XXXXXX"
}
```
- [x] PagerDuty test
Screen Shot 2020-06-17 at 6 52 36 PM


austinbyers:
> As a result some of the SQS/SNS messages now have slightly different structure. I am thinking it would be sufficient to notify the users in the release notes about that. @austinbyers what do you think?

You mean the alert outputs? I don't think they're used widely enough yet to be much of an issue.

Or is this referring to internal message structures? In which case is a migration necessary? E.g. draining the queues before upgrading?

kostaspap:
> > As a result some of the SQS/SNS messages now have slightly different structure. I am thinking it would be sufficient to notify the users in the release notes about that. @austinbyers what do you think?
> 
> You mean the alert outputs? I don't think they're used widely enough yet to be much of an issue.
> 
> Or is this referring to internal message structures? In which case is a migration necessary? E.g. draining the queues before upgrading?

Yeah, I was not clear. In this statement I was referring to alert outputs. I know some people use it, but I am assuming notifying them about these changes in the release notes might be sufficient? I would keep the old format though if we thinking it might cause troubles. Maybe @rleighton has an opinion on this as well. 

But yeah, the internal messages is another issue. In case there are some alerts in flight (alerts by definition are low volume so I don't expect a bunch of them), these alerts might be dropped. I am assuming that it is expected that Panther might be non operational during an upgrade.

kostaspap:
> > > As a result some of the SQS/SNS messages now have slightly different structure. I am thinking it would be sufficient to notify the users in the release notes about that. @austinbyers what do you think?
> > 
> > 
> > You mean the alert outputs? I don't think they're used widely enough yet to be much of an issue.
> > Or is this referring to internal message structures? In which case is a migration necessary? E.g. draining the queues before upgrading?
> 
> Yeah, I was not clear. In this statement I was referring to alert outputs. I know some people use it, but I am assuming notifying them about these changes in the release notes might be sufficient? I would keep the old format though if we thinking it might cause trouble. Maybe @rleighton has an opinion on this as well.
> 
> But yeah, the internal messages is another issue. In case there are some alerts in flight (alerts by definition are low volume so I don't expect a bunch of them), these alerts might be dropped. I am assuming that it is expected that Panther might be non operational during an upgrade.

Decided to stick with existing naming. I am aware of a few people using existing SQS notifications. Since there is no real reason to update the output message (other than... aesthetics :P ) and given that this PR already deals with the main goal of sending some missing fields to the destinations, better to avoid any migrations for now until we need to do so.

PR #1049, opened 2020-06-18
Removing 4XX alarms for ALB

kostaspap:
## Background

~Configuration ALB alarms to trigger only if we see persistent issues. AWS can occasionally encounter some "glitches", with this update we make sure we trigger an alarm only if the issues persist.~
After some consideration, thinking the best way to move forward is to get rid of the 4XX alarms. 
Some reasons why I think this is the best way of action: 

1. There is a very high chance for a false positive due to "expected" user activity. 
2. If the increased 4XX is due to some of our Lambdas experiencing issues, we will identify such issues from Lambda alarms.
3. Elevated 4XX might be a sign of a security issue. We should rely on Panther log (ALB) processing for this instead of CW Alarms. 

## Testing
* `mage test:ci`
* Deployed

### DIFF
--- a/deployments/web_server.yml
+++ b/deployments/web_server.yml
@@ -94,7 +94,6 @@ Resources:
     Type: Custom::ElbAlarms
     Properties:
       AlarmTopicArn: !Ref AlarmTopicArn
-      ClientErrorThreshold: 10 # within 5 minutes
       LatencyThresholdSeconds: 0.5
       LoadBalancerFriendlyName: web
       LoadBalancerFullName: !Ref ElbFullName
--- a/internal/core/custom_resources/resources/alarms_elb.go
+++ b/internal/core/custom_resources/resources/alarms_elb.go
@@ -41,7 +41,6 @@ type ElbAlarmProperties struct {
 	LoadBalancerFriendlyName string `validate:"required"`
 	LoadBalancerFullName     string `validate:"required"`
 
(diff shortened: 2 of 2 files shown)


austinbyers:
This is a problem with custom resources in general - updates aren't triggered unless a parameter changed. @rleighton encountered the same problem fixing some metric filters.

I'm going to see if I can add a "Version" dummy parameter to all custom resources to force all of them to be re-evaluated after every upgrade.

But in this case, I still think you should expose that parameter to CloudFormation. If we needed to change it once, we're likely to want to tweak it again.

rleighton:
[This](https://github.com/panther-labs/panther/pull/1051) should ensure the update.

kostaspap:
> This is a problem with custom resources in general - updates aren't triggered unless a parameter changed. @rleighton encountered the same problem fixing some metric filters.
> 
> I'm going to see if I can add a "Version" dummy parameter to all custom resources to force all of them to be re-evaluated after every upgrade.

Yeah, i saw that PR! Nice fix. 

I went ahead and removed the 4XX alarms btw, we had been discussing it for a while and while investigating some recent alarms seems it would be the best way forward (so don't really need to expose the new property neither :) )

austinbyers:
This will not work to update existing deployments now that you've removed it, even with #1051

It will force an update of the custom resource, but that will just re-create the existing alarms - it will not delete the client errors alarm. 

I guess the fix is to do the "migration" in the update code - we need to delete the old alarm on every update

PR #1051, opened 2020-06-18
Add CustomResourceVersion

austinbyers:
## Background

Like all CloudFormation resources, custom resources are not updated until their input parameters change.

This is not necessarily intuitive - for example #1049 tried to update an alarm group by changing the Go code in the custom resources Lambda function. That will work for new deployments, but will not be applied to existing deployments. Similarly, @rleighton is working on correcting the Python metric filters.

In both cases, the input parameters must change in some way to update existing deployments. To that end, this PR introduces a "CustomResourceVersion" parameter to most stacks which isn't used for anything except triggering custom resource updates.

For pre-packaged deployments, this will be the Panther release version, which will guarantee bug fixes like #1049 are applied. For `mage deploy`, this is the major release version by default. The developer can also force custom resource updates with `FORCE_CUSTOM_RESOURCE_UPDATE=1 dev -- mage deploy`

## Changes

- Add `CustomResourceVersion` parameter to most CloudFormation stacks and most custom resources

## Testing

- `mage deploy`
- `CUSTOM_RESOURCE_VERSION=v1.5.0 mage deploy`

### DIFF
--- a/deployments/appsync.yml
+++ b/deployments/appsync.yml
@@ -37,7 +37,7 @@ Parameters:
     Description: API Gateway HTTPS endpoint for panther-compliance-api
   CustomResourceVersion:
     Type: String
-    Description: Dummy parameter which forces updates to custom resources when changed
+    Description: Forces updates to custom resources when changed
   RemediationApi:
     Type: String
     Description: API Gateway HTTPS endpoint for panther-remediation-api
--- a/deployments/appsync.yml
+++ b/deployments/appsync.yml
@@ -35,6 +35,9 @@ Parameters:
   ComplianceApi:
     Type: String
     Description: API Gateway HTTPS endpoint for panther-compliance-api
+  CustomResourceVersion:
+    Type: String
+    Description: Dummy parameter which forces updates to custom resources when changed
   RemediationApi:
     Type: String
     Description: API Gateway HTTPS endpoint for panther-remediation-api
@@ -53,6 +56,7 @@ Resources:
       ApiName: panther-graphql-api
       AlarmTopicArn: !Ref AlarmTopicArn
       ClientErrorThreshold: 20 # per 5-minute window
+      CustomResourceVersion: !Ref CustomResourceVersion
       ServiceToken: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-cfn-custom-resources
 
   GraphQLSchema:
--- a/tools/mage/master_namespace.go
+++ b/tools/mage/master_namespace.go
@@ -60,7 +60,7 @@ func (Master) Deploy() {
 	err := sh.RunV(filepath.Join(pythonVirtualEnvPath, "bin", "sam"), "deploy",
 		"--capabilities", "CAPABILITY_IAM", "CAPABILITY_NAMED_IAM", "CAPABILITY_AUTO_EXPAND",
 		"--region", region,
-		"--stack-name", "panther-master",
+		"--stack-name", "panther",
 		"-t", pkg,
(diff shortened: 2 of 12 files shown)


deployments/appsync.yml
```diff
@@ -35,6 +35,9 @@ Parameters:
   ComplianceApi:
     Type: String
     Description: API Gateway HTTPS endpoint for panther-compliance-api
+  CustomResourceVersion:
+    Type: String
+    Description: Dummy parameter which forces updates to custom resources when changed
```
nhakmiller, line 40: A think calling it a dummy parameter is a bit misleading, it's a parameter used to indicate to custom resources that the version has changed. That could have meaning, especially with migrations (which will see the old property with say v1.4.0 and the new property v1.5.0). Just a thought.

tools/mage/master_namespace.go
```diff
@@ -60,7 +60,7 @@ func (Master) Deploy() {
 	err := sh.RunV(filepath.Join(pythonVirtualEnvPath, "bin", "sam"), "deploy",
 		"--capabilities", "CAPABILITY_IAM", "CAPABILITY_NAMED_IAM", "CAPABILITY_AUTO_EXPAND",
 		"--region", region,
-		"--stack-name", "panther-master",
+		"--stack-name", "panther",
```
rleighton, line 63: Why was this changed?

deployments/appsync.yml
austinbyers: It's an interesting idea, but I was deliberately trying to convey that the value of this parameter should carry no meaning. It might be the Panther version, or it could be a completely random string (see `FORCE_CUSTOM_RESOURCE_UPDATE=1 mage deploy`). 

If we want to think of this as the actual panther version, then how will developer migrations work? Suppose a resource is checking explicitly whether we're going from "v1.4.0" to "v1.5.0", and that migration is required to fix a breaking change. Every developer will need that migration, but the version in their repo will still be something like `v1.4.0-g2789sa`

We could change the environment variable override to define the Panther version: `PANTHER_VERSION=v1.5.0 mage deploy`
, which devs could use to force the migration. But then, in the next deploy, without that override, the custom resource will downgrade from "v1.5.0" to "v1.4.0"

In summary, I'm open to the idea of forcing a Panther version, since it does allow resources to explicitly compare their versions. But taking advantage of that in a way that works for all deployment methods adds unexpected complexity. It is simpler to just assume "update if this value every changes" without caring what the value is

tools/mage/master_namespace.go
austinbyers: So far, this is only used when testing the pre-packaged deployment, its value doesn't really matter. A few reasons I removed it, but I can revert if you prefer:

* I kept trying to do `STACK=panther mage teardown` during testing. I always assumed it had that name, and it's the name we suggest when using the pre-packaged deployment
* Really, the "master" template just refers to the single top-level template. Right now the nested stack names are `panther-master-ASDFJWER-core`, which doesn't really make sense
* The community has started avoiding use of the term "master" in the context of code. While I'm not going to take the time now to change all references to the master template, I can at least ensure it's removed from all stack names (since they are persistent)

rleighton: Ok, just curious and wanted to keep sure this did not create the need for a migration.

deployments/appsync.yml
austinbyers: After some more thought, I went ahead with your suggestion. The `CustomResourceVersion` is always the major Panther version, but developers can explicitly choose its value with `CUSTOM_RESOURCE_VERSION=v1.5.0 mage deploy`

It's an easy thing to change later if we want, and since it's almost always the version anyway, might as well make it official. Good idea!

PR #1052, opened 2020-06-19
Add layer attachment resource

nhakmiller:
## Background

We found a bug where if you update the base python layer then do an upgrade of Panther, the global layer gets detached. This is because AWS Lambda does not allow for partial layer updates, so when CloudFormation goes to re-attach the new lambda layer it detaches all other layers. This change adds a custom resource for managing the layers attached to our engine functions.

Note that after deploying this change, you will need to trigger a re-build of the lambda layer. This can be done by re-uploading a policy/rule pack via the bulk upload functionality that contains at least one global in it, communicating to the `layer-manager` lambda directly, or updating the a global via the web UI. I considered making a migration step for this in the migration function in mage but I wasn't actually sure how to add that to pre-packaged deployments, which will also need this migration step. Since it's a relatively easy one time migration, I figured a migration note is adequate.

## Changes

- Create a custom resource for managing the attachment of layers to a lambda function

## Testing

- Deployed to my dev account and did several deployments, rebuilding the base layer each time. Between each deployment I verified that the global layer was still attached.


internal/core/custom_resources/resources/layer_attachment.go
```diff
@@ -0,0 +1,223 @@
...
+		FunctionName: newProps.FunctionArn,
+	})
+	if err != nil {
+		return "", nil, err
+	}
+
+	// Insert/update/remove layers as appropriate
+
+	// First determine if there are any existing layers that need to be removed
+	var layersToRemove []*Layer
```
rleighton, line 82: Could this me a map and the test below simplified as a test if the arn is in the map?

nhakmiller:
Ok after me and Austin's discussion we decided a different approach would be best. Instead of updating the resource directly (which would then conflict with each update), we just construct the list of parameters to pass to the lambda function. The code is much cleaner now I think!

internal/core/custom_resources/resources/layer_attachment.go
nhakmiller: This is totally different now!

internal/core/custom_resources/resources/layer_attachment.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func customLayerAttachment(_ context.Context, event cfn.Event) (string, map[string]interface{}, error) {
+	switch event.RequestType {
+	case cfn.RequestCreate, cfn.RequestUpdate:
+		return handleCreateUpdateRequests(event)
+	case cfn.RequestDelete:
+		return event.PhysicalResourceID, nil, nil
+	default:
+		return "", nil, fmt.Errorf("unknown request type %s", event.RequestType)
```
austinbyers, line 46: Since the error is essentially impossible, you can just do

```suggestion
	default:
		// ignore deletes
		return event.PhysicalResourceID, nil, nil
```

Either way is fine, I don't think I've always been consistent here

internal/core/custom_resources/resources/layer_attachment.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"github.com/aws/aws-lambda-go/cfn"
+	"github.com/aws/aws-sdk-go/service/lambda"
+	"go.uber.org/zap"
+)
+
+const latest = "LATEST"
+
+type LayerAttachmentProperties struct {
+	FunctionArn *string
```
austinbyers, line 35: You're not using this for anything except the custom resource ID, let's remove it. The custom resource ID can be `custom:layerSet` or `custom:layers:singleton` or really anything, doesn't matter.

This resource no longer cares where the layers are used, it should have no concept of relation to a Lambda function. It just takes an input list and returns a modified list

internal/core/custom_resources/resources/layer_attachment.go
```diff
@@ -0,0 +1,103 @@
...
+
+func getLatestLayerVersion(layerArn string) (string, error) {
+	response, err := lambdaClient.ListLayerVersions(&lambda.ListLayerVersionsInput{
+		LayerName: &layerArn,
+	})
+	if err != nil {
+		return "", err
+	}
+	greatest := int64(-1)
+	for _, layer := range response.LayerVersions {
```
austinbyers, line 94: Is the response not sorted already by version number? Since this response is a list, I'd be surprised if it's in a random order

internal/core/custom_resources/resources/map.go
```diff
@@ -159,6 +159,18 @@ var CustomResources = map[string]cfn.CustomResourceFunction{
...
 
+	// Attach layers to a lambda function without removing existing ones.
+	//
+	// Parameters:
+	//     FunctionArn:           string
+	//     Layers:
+	//         - FunctionArn      string
+	//           FunctionVersion  int
+	//
+	// Outputs: None
```
austinbyers, line 170: No longer accurate - there is a `LayerArns` output

deployments/log_analysis.yml
```diff
@@ -871,3 +863,24 @@ Resources:
...
+            - '${base},${pip},${global}'
+            - base: !Join [',', !Ref LayerVersionArns]
+              pip: !Ref PythonLayerVersionArn
+              global: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:layer:panther-engine-globals:LATEST
+        - !Split # CFN doesn't have list append, so convert to/from CSV string to prepend base layers
+          - ','
+          - !Sub
+            - '${pip},${global}'
+            - pip: !Ref PythonLayerVersionArn
+              global: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:layer:panther-engine-globals:LATEST
```
austinbyers, line 885: Damn, CloudFormation makes this ugly. We should eventually change the "LayerVersionArns" parameter to just accept a single layer, users can't and won't realistically add multiple anyway. Then we don't have to do this split + join hack.

In the meantime, lets move the conditional deeper to avoid having to duplicate the entire list:

```suggestion
      # CFN doesn't have list append, so convert to/from CSV string to prepend base layers
      LayerArns: !Split
        - ','
        - !Sub
          - '${base}${comma}${pip},${global}'
          - base: !Join [',', !Ref LayerVersionArns]
            comma: !If [AttachLayers, ',', ''] # add comma after base layers if applicable
            pip: !Ref PythonLayerVersionArn
            global: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:layer:panther-engine-globals:LATEST
```

same for the policy engine. I tested this (without the global) and it worked

internal/core/custom_resources/resources/layer_attachment.go
nhakmiller: Does the custom resource ID not need to be unique for resources of the same type? For some reason I thought that was a requirement. I can remove it.

internal/core/custom_resources/resources/layer_attachment.go
nhakmiller: None of the AWS docs make this promise, if it is the case it is an implementation detail that is subject to change whenever. I'd prefer the safer approach of checking each version returned, for the global layer at least we clean up old versions so there should only ever be one or two responses at most anyways.

internal/core/custom_resources/resources/map.go
nhakmiller: I just updated this whole comment as it is no longer very accurate.

internal/core/custom_resources/resources/layer_attachment.go
austinbyers: During an UPDATE request, CloudFormation compares the old and new PhysicalID returned by the custom resource. If the PhysicalID changes, it will automatically issue a DELETE for the old PhysicalID. That's really all it is used for

I've never tried two custom resources in the same stack with the same physical ID, I don't know if that would cause problems or not. But we have many "singleton" resources with hardcoded IDs.

In any case, if uniqueness were required, you could just append a random number or the hash of the input, for example.

nhakmiller: I also just realized we can create this custom resource once in the bootstrap stack and pass it's output through to both the cloud security and log analysis stacks. Do you think that's worthwhile or too confusing?

nhakmiller: Since we still have to pass in the `LayerVersionArns` anyways for the rest of the lambdas, I'll just leave it as is for now. But if we start following this pattern of "read only" custom resources, we could front load them in the bootstrap stack since there are a lot of other slow running operations there we can do in parallel and there are no dependencies (hopefully).

austinbyers: > Sounds good. I also just realized we can create this custom resource once in the bootstrap stack and pass it's output through to both the cloud security and log analysis stacks.

I see, the layer set is exactly the same in both cases? Let's think about the order of operations - is the global layer always defined in a fresh deployment *before* you build these layer lists right now? If so, that would no longer be true when building in the bootstrap. If not, that dependency ordering may be a bug we need to fix

PR #1054, opened 2020-06-19
Log processor update source status with the time it received the last event

kostaspap:
## Background

Current `Sources` page contains little information the current status of a source. 
While we are looking to provide more rich monitoring around a source status, adding a small piece of information as a stop-gap solution. 

With this solution, the log processor updates the "status" of the source with the timestamp of the most recently received event. The status will have 1 minute accuracy. 

## Changes

- Added new API method "UpdateStatus" to the panther-source-api Lambda.
- Updated log processor to invoke the source-api
- Shortening `ddb.IntegrationItem` to just `ddb.Integration`. The namespace `ddb` adds context that this is a resource related to DynamoDB. 
- Added "omitempty" to several fields returned by the UI. While we do this for most of our other APIs, it seemed we were not following this practice for `source-api`.

## Testing

- mage test:ci
- Deployed. ListIntegration operation now returns this: 

```
 {
    "awsAccountId": "415773754570",
    "createdAtTime": "2020-06-17T16:10:19.240641152Z",
    "createdBy": "00000000-0000-4000-8000-000000000000",
    "integrationId": "3734447d-5b26-444b-921b-acb5dbbf3d30",
    "integrationLabel": "panther-account-eu-west-1",
    "integrationType": "aws-s3",
    "s3Bucket": "panther-bootstrap-auditlogs-XXXXXXXX",
    "logTypes": [
      "AWS.ALB",
      "AWS.S3ServerAccess",
      "AWS.VPCFlow"
    ],
    "logProcessingRole": "arn:aws:iam::123456789012:role/PantherLogProcessingRole-panther-account-eu-west-1",
    "stackName": "panther-log-analysis-setup-panther-account-eu-west-1",
    "lastEventReceived": "2020-06-19T14:14:26.710855261Z"
  },
```

### DIFF
--- a/internal/log_analysis/log_processor/sources/s3_client_cache.go
+++ b/internal/log_analysis/log_processor/sources/s3_client_cache.go
@@ -201,7 +201,7 @@ func getSourceInfo(s3Object *S3ObjectInfo) (result *models.SourceIntegration, er
 		}
 	}
 
-	//
+	// If the incoming notification maps to a known source, update the source information
 	if result != nil {
 		previousTime := lastEventReceived[*result.IntegrationID]
 		if previousTime.Add(statusUpdateFrequency).Before(now) {
--- a/internal/log_analysis/log_processor/sources/s3_client_cache.go
+++ b/internal/log_analysis/log_processor/sources/s3_client_cache.go
@@ -73,6 +73,11 @@ var (
 	//used to simplify mocking during testing
 	newCredentialsFunc = stscreds.NewCredentials
 	newS3ClientFunc    = getNewS3Client
+
+	// Map from integrationId -> last time an event was received
+	lastEventReceived = make(map[string]time.Time)
+	// How frequently to update the status
+	statusUpdateFrequency = 1 * time.Minute
 )
 
 func init() {
@@ -170,7 +175,7 @@ func getAwsCredentials(roleArn string) *credentials.Credentials {
 // Returns the source configuration for this S3 object.
 // It will return error if it encountered an issue retrieving the role.
 // It will return nil result if no source exists for this object.
-func getSourceInfo(s3Object *S3ObjectInfo) (*models.SourceIntegration, error) {
+func getSourceInfo(s3Object *S3ObjectInfo) (result *models.SourceIntegration, err error) {
 	now := time.Now() // No need to be UTC. We care about relative time
 	if sourceCache.cacheUpdateTime.Add(sourceCacheDuration).Before(now) {
 		// we need to update the cache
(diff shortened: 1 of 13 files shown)


rleighton:
I like this!

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -190,11 +195,37 @@ func getSourceInfo(s3Object *S3ObjectInfo) (*models.SourceIntegration, error) {
...
+				break
 			}
 		}
 	}
-	return nil, nil
+
+	// If the incoming notification maps to a known source, update the source information
+	if result != nil {
+		previousTime := lastEventReceived[*result.IntegrationID]
+		if previousTime.Add(statusUpdateFrequency).Before(now) {
```
austinbyers, line 207:
```suggestion
		if time.Since(previousTime) > statusUpdateFrequency
```

Similarly, use `time.Since` on line 180. It reads much more clearly

kostaspap:
> Be sure to run integration tests as well, and add new integration tests for the new endpoint

Integration tests do not work for this API unfortunately, this is a known issue since few months -> https://github.com/panther-labs/panther/issues/394

internal/log_analysis/log_processor/sources/s3_client_cache.go
kostaspap: > Similarly, use `time.Since` on line 180. It reads much more clearly

Hmmm, I don't see a way to do this while using the `now` variable.

austinbyers: Right, you don't need the `now` variable at all then

kostaspap: I was trying to reuse the `now` value to avoid calling `time.Now` for a second time inside the same function (`time.Since` calls `time.Now` internally).
Apparently Go doesn't make a syscall every time `time.Now` is invoked (it seems it has a type of thread-local cached value) but it seems it is considered a good practice to avoid invoking many times in "hot paths". This path is accessed frequently, but not THAT much (max a few 10s per second)

I did some small refactoring and I think the code is now more readable while reusing the time value. If you think there are still readability issues, I will switch it.

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -177,7 +177,7 @@ func getAwsCredentials(roleArn string) *credentials.Credentials {
 // It will return nil result if no source exists for this object.
 func getSourceInfo(s3Object *S3ObjectInfo) (result *models.SourceIntegration, err error) {
 	currentTime := time.Now() // No need to be UTC. We care about relative time
-	if sourceCache.cacheUpdateTime.Add(sourceCacheDuration).Before(currentTime) {
+	if currentTime.Sub(sourceCache.cacheUpdateTime) > sourceCacheDuration {
```
alxarch, line 180: The previous form is usually preferred as [it only compares the monotonic clock](https://golang.org/src/time/time.go
)

> // If Times t and u both contain monotonic clock readings, the operations
> // t.After(u), t.Before(u), t.Equal(u), and t.Sub(u) are carried out
> // using the monotonic clock readings alone, ignoring the wall clock
> // readings. If either t or u contains no monotonic clock reading, these
> // operations fall back to using the wall clock readings.

PR #1058, opened 2020-06-19
Migration: remove deleted alarm

austinbyers:
## Background

Follow-up to #1049 which removes the alarm for existing deployments, not just new ones

The `CustomResourceVersion` parameter ensures the custom resource Go code will run, but right now that justs reinstalls the existing alarms, it will never remove the deprecated alarm until the resource is removed.

## Changes

- Delete deprecated alarm when updating the ELBAlarm resource type

We can potentially remove this migration after a few release cycles, when all of our customers have migrated off of it.

## Testing

- `mage test:ci`


internal/core/custom_resources/resources/alarms_elb.go
```diff
@@ -55,7 +56,18 @@ func customElbAlarms(_ context.Context, event cfn.Event) (string, map[string]int
...
-		return "custom:alarms:elb:" + props.LoadBalancerFriendlyName, nil, putElbAlarmGroup(props)
+
+		physicalID := "custom:alarms:elb:" + props.LoadBalancerFriendlyName
+		if err := putElbAlarmGroup(props); err != nil {
+			return physicalID, nil, err
+		}
+
+		// Migration: In v1.5.0, the elbTargetClientErrorAlarm was removed
+		if err := deleteMetricAlarms(event.PhysicalResourceID, elbClientErrorAlarm); err != nil {
+			zap.L().Error("failed to remove deprecated alarm", zap.Error(err))
```
austinbyers, line 67: There is no error if the alarm does not exist

internal/core/custom_resources/resources/alarms_elb.go
```diff
@@ -55,7 +56,18 @@ func customElbAlarms(_ context.Context, event cfn.Event) (string, map[string]int
...
 		}
-		return "custom:alarms:elb:" + props.LoadBalancerFriendlyName, nil, putElbAlarmGroup(props)
+
+		physicalID := "custom:alarms:elb:" + props.LoadBalancerFriendlyName
+		if err := putElbAlarmGroup(props); err != nil {
+			return physicalID, nil, err
+		}
+
+		// Migration: In v1.5.0, the elbTargetClientErrorAlarm was removed
+		if err := deleteMetricAlarms(event.PhysicalResourceID, elbClientErrorAlarm); err != nil {
```
kostaspap, line 66: Also need to delete `elbTargetClientErrorAlarm`.

PR #1061, opened 2020-06-20
Fix python lambda metric filter expressions

rleighton:
## Background
The python lambda metric filters for error and warns were not correct. Checking older releases showed this was not a regression and was a long standing (and unnoticed) issue.

## Changes

- Corrected the expressions

## Testing

- mage test:ci
- CUSTOM_RESOURCE_VERSION=1234 mage deploy , then run a rule that always threw an exception and watched for alarms


kostaspap:
If I understand it correctly @austinbyers , this code change will keep the old metric filters and also introduce new ones?

PR #1065, opened 2020-06-22
Bug fixing/improvements in log processor

kostaspap:
## Background

This PR is fixing a couple of issues with the log processor
- If the log processor attempts to process a file of unsupported file type (we only support `text/plain` and `gzip` files), it will log an error and skip the file
- We avoid trying the CSV parsing if incoming event is not CSV file. 

## Testing

- `mage test:ci`
- Deployed to dev account . Uploaded `jpg` file to be processed by log processor. 
It logged an error and just skipped the file.


internal/log_analysis/log_processor/parsers/json_utils.go
```diff
@@ -0,0 +1,28 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+// Returns true if log looks like a JSON log.
+func LooksLikeJSON(log string) bool {
+	length := len(log)
+	if length < 2 {
+		return false
+	}
+	return log[0] == '{' && log[length-1] == '}'
```
rleighton, line 27: Hmmm, not sure we want the closing brace, since the JSON might be on more than 1 line.

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -106,6 +106,11 @@ func handleNotificationMessage(notification *SnsNotification) (result []*common.
 		var dataStream *common.DataStream
 		dataStream, err = readS3Object(s3Object)
 		if err != nil {
+			if _, ok := err.(*ErrUnsupportedFileType); ok {
+				// If the incoming message is not of a supported type, just skip it
```
rleighton, line 110: I think a Warn() should be logged that includes the bucket and path to track the misconfiguration.

internal/log_analysis/log_processor/parsers/json_utils.go
kostaspap: Even if the JSON was multiline (currently we don't support that), for the log to be a valid JSON it would have to start with `{` and finish with `}`. If there are `\n` in between it shouldn't matter.

internal/log_analysis/log_processor/sources/s3.go
kostaspap: Actually the current code will log an Err including the bucket and S3 key - see line 122.

internal/log_analysis/log_processor/parsers/json_utils.go
rleighton: That's not the reason ... if a BIG json blob that will mess up the CSV parsers is submitted, this only catches it if the last char is a `}`. This is not a json detector really, it is a "not csv" detector really. Shouldn't we be conservative?

internal/log_analysis/log_processor/sources/s3.go
rleighton: Yup. Missed that in all the code folding.

internal/log_analysis/log_processor/parsers/json_utils.go
kostaspap: Should I make a `looksLikeCsv` method in csv_utils.go ? It will return false if first character is `{`.

rleighton: Oh, like that much better. Makes the csv code easier to read too.

kostaspap: Done! Let me know what you think now.

internal/log_analysis/log_processor/parsers/csv_utils.go
```diff
@@ -23,6 +23,17 @@ import (
 	"strings"
 )
 
+// Returns true if log looks like a CSV log.
+// It can be used to fail fast for logs that are no CSV
```
rleighton, line 27: s/no/not

kostaspap: Done

PR #1067, opened 2020-06-22
Update all backend dependencies

austinbyers:
## Background

We try to upgrade dependencies as part of every release. In particular, upgrading `golangci-lint` proved to be really useful as it identified several previously undetected issues:

```
tools/cfnparse/parse.go:68:9: G306: Expect WriteFile permissions to be 0600 or less (gosec)
	return ioutil.WriteFile(path, contents, 0644)
	       ^
tools/mage/fmt.go:138:9: G306: Expect WriteFile permissions to be 0600 or less (gosec)
	return ioutil.WriteFile(path, bytes.Join(newLines, []byte("\n")), 0644)
	       ^
tools/mage/util.go:142:12: G306: Expect WriteFile permissions to be 0600 or less (gosec)
	if err := ioutil.WriteFile(path, data, 0644); err != nil {
	          ^
internal/core/custom_resources/resources/layer_attachment.go:77:16: S1039: unnecessary use of fmt.Sprintf (gosimple)
	resourceID := fmt.Sprintf("custom:lambda:layerattachment")
	              ^
internal/log_analysis/athenaviews/athenaviews.go:126:32: S1039: unnecessary use of fmt.Sprintf (gosimple)
			sqlLines = append(sqlLines, fmt.Sprintf("\tunion all"))
			                            ^
internal/log_analysis/athenaviews/athenaviews.go:130:30: S1039: unnecessary use of fmt.Sprintf (gosimple)
	sqlLines = append(sqlLines, fmt.Sprintf(";\n"))
	                            ^
internal/compliance/snapshot_poller/pollers/aws/kms_key.go:189:49: SA5011: possible nil pointer dereference (staticcheck)
			TimeCreated:  utils.DateTimeFormat(*metadata.CreationDate),
			                                             ^
internal/compliance/snapshot_poller/pollers/aws/kms_key.go:195:31: SA5011: possible nil pointer dereference (staticcheck)
		CloudHsmClusterId: metadata.CloudHsmClusterId,
		                            ^
internal/compliance/snapshot_poller/pollers/aws/kms_key.go:196:31: SA5011: possible nil pointer dereference (staticcheck)
		CustomKeyStoreId:  metadata.CustomKeyStoreId,
		                            ^
internal/compliance/remediation_api/handlers/post.go:30:2: ST1019: package "github.com/panther-labs/panther/api/gateway/remediation/models" is being imported more than once (stylecheck)
	"github.com/panther-labs/panther/api/gateway/remediation/models"
	^
internal/log_analysis/awsglue/integration_test.go:41:2: ST1003: const testDb should be testDB (stylecheck)
	testDb           = "panther_glue_test_db"
	^
internal/log_analysis/log_processor/parsers/zeeklogs/dns.go:39:2: ST1003: struct field Ts should be TS (stylecheck)
	Ts         *timestamp.UnixFloat `json:"ts,omitempty" validate:"required" description:"The earliest time at which a DNS protocol message over the associated connection is observed."`
```

## Changes

- Bump Go library versions installed directly from `setup.go`
- Reset Go modules: `rm go.mod go.sum && mage clean setup test:ci`
- Reset Python requirements following instructions in `requirements-top-level.txt`

## Testing

- `mage test:ci`


alxarch:
We should be careful when updating packages especially when they have no support for go module versioning. I think that removing the go.mod altogether is not a good idea, we cannot pin package versions that way. We just upgrade to latest and hope our unit tests catch breakage. The safest is to remove sum and use `go get -u=patch`for dependencies

austinbyers:
> We should be careful when updating packages especially when they have no support for go module versioning.

I understand your concern, but the only entry in `go.mod` that does not support proper modules is "golang.org/x/tools" whose version we already pin in `tools/mage/setup.go` - we explicitly fetch the commit that matches the latest tagged release. (And that lib is just for formatting/CI, not the Lambda binaries).

Since there is no other module whose version we care about, it's safe to rebuild `go.mod` each time. It's what we've done in every release since the beginning.

You're right that if one day we had to explicitly pin a specific version in `go.mod`, that approach would no longer work. So far, that's never happened, and I hope it never will. If it does, we can add comments in the file about why we have to pin the version and potentially add protections to `mage` to ensure it's not overwritten.

But in general, I would rather upgrade constantly, find a breaking change in release testing and rollback, then never upgrade major versions because we're being too conservative.

PR #1069, opened 2020-06-22
Fixed error handling case

nhakmiller:
## Background

We observed rate limiting during scanning of IAM users. This is because if an IAM credential report does not exist, we have to create and then wait for it to exist. While waiting for the report to be created, we sometimes get rate limited on our `GenerateCredentialReport` API call. In these cases, we just want to re-attempt the scan in a few minutes once the report has finished being created. But in order to do that, we have to know that the issue was related to rate limiting and not some other issue. To do this we attempt to explicitly catch the rate limiting error and when we see it re-queue the scan for some point in the future (hopefully after the rate limit has passed and the credential report has finished being generated).

Previously, we were attempting to catch the wrong type of error to handle these situations. The unfortunate thing is this can only be tested once every 4 hours in an account, which makes verifying the issue and the fix difficult even with a handful of accounts. I'm still trying to replicate the issue exactly, but based on the logs I believe this change will be the solution.

## Changes

- Change the error being caught for retrys

## Testing

- Working on reliably reproducing the issue

### DIFF
--- a/internal/compliance/snapshot_poller/pollers/aws/iam_user.go
+++ b/internal/compliance/snapshot_poller/pollers/aws/iam_user.go
@@ -47,6 +47,7 @@ const (
 	maxCredReportBackoff                = 1 * time.Minute
 	rootAccountNameCredReport           = "<root_account>"
 	rootDeviceSerialSuffix              = ":mfa/root-account-mfa-device"
+	throttlingErrorCode                 = "Throttling"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


nhakmiller:
Still can't reliably reproduce this in Panther, but this little test snippet seemed to do what I wanted:

```
        const throttlingErrorCode                 = "Throttling"
	awsSession := session.Must(session.NewSession(&aws.Config{}))
	client := iam.New(awsSession)

	for i:= 0; i <10; i++ {
		_, err := client.GenerateCredentialReport(&iam.GenerateCredentialReportInput{})
		if err != nil {
			if awsErr, ok := err.(awserr.Error); ok {
				if awsErr.Code() == throttlingErrorCode {
					fmt.Println("AWS throttling error")
					fmt.Println(err)
					return
				}
			}
			fmt.Println("other error")
			fmt.Println(err)
			return
		}
		fmt.Println("no error")
	}
}
```
yields
```
% go run scratch.go
testing iam report generation
no error
AWS throttling error
Throttling: Rate exceeded
        status code: 400, request id: 8a834b7
```

So I think we are correctly grabbing the right error now.

PR #1072, opened 2020-06-23
Move VPC flow logs to bootstrap stack

austinbyers:
## Background

VPC flow logs moved from the bootstrap stack to the onboard stack when preparing the pre-packaged deployment because of a SAR compatibility issue. Our hosted deployments generally do not enable self-onboarding, meaning the entire onboard stack, including VPC flow logs, does not exist in those environments.

VPC flow logs are similar to S3 access logs - they are enabled on each resource, not account-wide like CloudTrail/GuardDuty. Like S3 access logs, VPC flow logs should always be enabled if possible, even if they aren't being ingested by the Panther deployment in the same account.

## Changes

- Move FlowLogs resource from onboard into bootstrap stack (SAR compatibility is no longer a concern)

The configuration can't just be copied exactly without a migration:

```
11:43:31	INFO	deploy: update CloudFormation stack panther-bootstrap
11:44:08	ERROR	stack panther-bootstrap: AWS::EC2::FlowLog FlowLogs CREATE_FAILED: Error. There is an existing Flow Log with the same configuration and log destination. (Service: Ec2, Status Code: 400, Request ID: cbfb08c1-ca97-4daa-9ec3-a604e8428c41, Extended Request ID: null)
```

Rather than writing a migration (which would require a new custom resource), I just changed the S3 target to include the VPC id as a top level folder when writing the logs. This is how S3 access logs work as well, and I'd argue it's a better configuration than writing logs to the root of the bucket anyway.

During the v1.4.0 => v1.5.0 migration, then, there will briefly be two different flow logs configured on the same Panther VPC - one which writes to the bucket root, one which writes under a folder.

## Testing

- Upgraded existing deployment

### DIFF
--- a/deployments/bootstrap.yml
+++ b/deployments/bootstrap.yml
@@ -615,6 +615,16 @@ Resources:
       EnableDnsHostnames: true
       CidrBlock: !FindInMap [SubnetConfig, VPC, CIDR]
 
+  FlowLogs:
+    DependsOn: AuditLogsBucketPolicy
+    Type: AWS::EC2::FlowLog
+    Properties:
+      LogDestination: !Sub ${AuditLogs.Arn}/${VPC}/
+      LogDestinationType: s3
+      ResourceId: !Ref VPC
+      ResourceType: VPC
+      TrafficType: ALL
+
   # We define a public subnet so that we can access our web server from a public IP. The empty
   # string in !GetAZs is equivalent to AWS::Region
   PublicSubnetOne:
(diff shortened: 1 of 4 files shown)


deployments/bootstrap.yml
```diff
@@ -932,9 +942,6 @@ Outputs:
     Value: !Ref Source
 
   # Networking + elb
-  VpcId:
```
austinbyers, line 935: Output is no longer used anywhere

PR #1076, opened 2020-06-25
Log Source Table to display "lastEventReceived" for each source

### LINKED ISSUE #1055: Log sources page should display the time a source received an event
### Description

We should display in the Log Sources UI the time a source received an event last time. 

This information is returned by the backend as part of the "ListIntegrations" operation and is implemented by this PR: https://github.com/panther-labs/panther/pull/1054

### Acceptance Criteria

- Log Sources page displays for every source the time it received last an event
- In case a log source hasn't received an event, display "N/A", "Unknown", etc

alexmylonas:
## Background

We want to show to users when was the latest event received for each integration. The purpose of that is to primarily understand if an integration is active (or recently active) and not just "healthy". The backend functionality for this is implemented on #1054. This PR introduces the changes needed to display that field in UI.

Note: Ignore backend changes on this PR, they should be handle on #1054 

Closes #1055 

## Changes

- Added `lastEventReceived` field to `S3LogIntegration`
- Updated `LogSourceTable` to display field

## Testing

- Locally
- Deployed env

Screenshot 2020-06-25 at 9 09 38 AM

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -368,6 +368,7 @@ type S3LogIntegration {
   integrationId: ID!
   integrationType: String!
   integrationLabel: String!
+  lastEventReceived: AWSDateTime
   s3Bucket: String!
   s3Prefix: String
   kmsKey: String
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -368,7 +368,7 @@ type S3LogIntegration {
   integrationId: ID!
   integrationType: String!
   integrationLabel: String!
-  lastEventReceived: AWSTimestamp
+  lastEventReceived: AWSDateTime
   s3Bucket: String!
   s3Prefix: String
   kmsKey: String
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -368,7 +368,7 @@ type S3LogIntegration {
   integrationId: ID!
   integrationType: String!
   integrationLabel: String!
-  lastEventReceived: AWSDateTime
+  lastEventReceived: AWSTimestamp
   s3Bucket: String!
   s3Prefix: String
   kmsKey: String
--- a/web/__generated__/schema.tsx
+++ b/web/__generated__/schema.tsx
@@ -29,8 +29,8 @@ export type Scalars = {
   Float: number;
   AWSDateTime: string;
   AWSJSON: string;
-  AWSTimestamp: number;
   AWSEmail: string;
+  AWSTimestamp: number;
 };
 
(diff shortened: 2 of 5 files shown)


3nvi:
Hey @alexmylonas, code is fine. Before we merge it though, I want to see if we indeed want to showcase it as a table cell  and not include it in the health check at all. Last time I spoke with @kostaspap I think we agreed that "last event received" can't be correlated to the health of a  source.

kostaspap:
> Hey @alexmylonas, code is fine. Before we merge it though, I want to see if we indeed want to showcase it as a table cell and not include it in the health check at all. Last time I spoke with @kostaspap I think we agreed that "last event received" can't be correlated to the health of a source.

The "last event received" does not have an impact to what we display currently as "health" (`healthy` yes/no) of a source. They are related but it is possible that a source is marked as "healthy" while the last event received has a None value.

alexmylonas:
@kostaspap Should we add something of an internal type of Alert for this? Meaning if a source doesn't receive events for a day(or hour whatever makes sense) that something probably is failing?

kostaspap:
> @kostaspap Should we add something of an internal type of Alert for this? Meaning if a source doesn't receive events for a day(or hour whatever makes sense) that something probably is failing?

There is no way we (Panther) can set an appropriate alert threshold for this. Some sources get thousands events/second, others can have 1 per week (e.g. GuardDuty). There is no way for us to know this before hand, it really depends on the monitored system. 

We could at some point expose a functionality that would allow the users to configure the thresholds themselves, but this is a more ambitious effort :)

api/graphql/schema.graphql
```diff
@@ -368,6 +368,7 @@ type S3LogIntegration {
   integrationId: ID!
   integrationType: String!
   integrationLabel: String!
+  lastEventReceived: AWSDateTime
```
3nvi, line 371: Can this be `null`? (@kostaspap )

alexmylonas: @kostaspap Can you confirm this, so we can merge it?

kostaspap: So, when we have not received any event, we return the default `time` value of `"0001-01-01T00:00:00Z"`. Maybe better that I return `null`? What do you think?

alexmylonas: Better `null` that date seems "valid" which turns into the below screenshot 
Screenshot 2020-07-01 at 12 13 45 PM

kostaspap: Updated in the backend to return null if it is not set.

PR #1079, opened 2020-06-25
Updating API to remove pointers

kostaspap:
## Background

After using the source-api for a while, it seems that not all fields should necessarily be pointers. I went ahead and made some changes that I think simplify the code. 
I don't think we should be "force" other API to remove pointers or anything like that, this works on a case by case basis. I just considered that for the particular APIs/fields this change was preferable. 

. Open to suggestions if you think we shouldn't change some of the APIs or fields in the APIs. I know it's a big PR (many other of our services depend on source-api) but don't be hesitant to express your preference, if you think some of the changes shouldn't happen. 

## Changes

- Updated API and callers

## Testing

- mage test:ci
- integration tests
- Deployed and tried basic functionality of the app (clicking around int he sources, adding source, etc)


api/lambda/source/models/api.go
```diff
@@ -65,22 +65,21 @@ type CheckIntegrationInput struct {
...
-	ScanIntervalMins   *int      `json:"scanIntervalMins,omitempty" validate:"omitempty,oneof=60 180 360 720 1440"`
-	UserID             *string   `json:"userId" validate:"required,uuid4"`
-	S3Bucket           *string   `json:"s3Bucket,omitempty"`
-	S3Prefix           *string   `json:"s3Prefix,omitempty" validate:"omitempty,min=1"`
-	KmsKey             *string   `json:"kmsKey,omitempty" validate:"omitempty,kmsKeyArn"`
-	LogTypes           []*string `json:"logTypes,omitempty" validate:"omitempty,min=1"`
+	IntegrationLabel   string   `json:"integrationLabel" validate:"required,integrationLabel,excludesall='<>&\""`
+	IntegrationType    string   `json:"integrationType" validate:"required,oneof=aws-scan aws-s3"`
+	UserID             string   `json:"userId" validate:"required,uuid4"`
+	AWSAccountID       string   `genericapi:"redact" json:"awsAccountId" validate:"omitempty,len=12,numeric"`
```
kostaspap, line 75: Some of the integrations no longer require AWS Account Id -> it's no longer marked as required.

internal/core/source_api/api/check_integration.go
```diff
@@ -171,31 +169,31 @@ func evaluateIntegration(api API, integration *models.CheckIntegrationInput) (st
...
 
-		if !aws.BoolValue(status.S3BucketStatus.Healthy) {
+		if !status.S3BucketStatus.Healthy {
 			return "log processing role cannot access s3 bucket", false, nil
 		}
 
-		if integration.KmsKey != nil {
-			return "log processing role cannot access kms key", aws.BoolValue(status.KMSKeyStatus.Healthy), nil
+		if !status.KMSKeyStatus.Healthy {
+			return "log processing role cannot access kms key", false, nil
```
kostaspap, line 196: This was actually a bug -> we would return an error message even if the resource was healthy.

api/lambda/source/models/api.go
```diff
@@ -44,18 +44,18 @@ type LambdaInput struct {
 
 // CheckIntegrationInput is used to check the health of a potential configuration.
 type CheckIntegrationInput struct {
-	AWSAccountID     *string `genericapi:"redact" json:"awsAccountId" validate:"required,len=12,numeric"`
-	IntegrationType  *string `json:"integrationType" validate:"required,oneof=aws-scan aws-s3"`
-	IntegrationLabel *string `json:"integrationLabel" validate:"required,integrationLabel"`
+	AWSAccountID     string `genericapi:"redact" json:"awsAccountId" validate:"required,len=12,numeric"`
```
austinbyers, line 47: Why I originally decided that all input/output should be pointers:

1. You can always distinguish a missing field from a default value
2. Consistent with the AWS SDK - often, many fields are passed directly to AWS APIs, which always expect pointers

The first one is the real concern. For example, you have two boolean variables in this input struct: `EnableCWESetup` and `EnableRemediation`. Before, the frontend could omit those entries from the request. Now, they will always be defined (defaulting to false). Perhaps in this case, you always want them to default to false, but it may sometimes be useful to distinguish a missing boolean from a 'false' boolean. It is now impossible to make that distinction. Same goes for integers - you may not always want them to default to 0 if they are undefined.

For that reason, I would still prefer to have all input/output be pointers like the AWS SDK and rely on `validate` tags alone to enforce required fields, but it's not necessarily a strong preference. It is more work for the backend to deal with the pointers, but I would argue that's the job of an API handler - to generally not trust the input and handle a variety of error conditions (i.e., missing fields)

If, in general, we don't ever need to make distinctions between missing and default fields, I agree getting rid of the pointers is safer and simpler and we could move all APIs in that direction. Each service could handle these things differently, but I would prefer to have a consistent story about when we should and should not use pointers. Perhaps we flip the logic - all fields are *not* pointers unless we need to distinguish empty vs missing fields? Or they are only pointers if they are optional fields?

I'd love some input from @rleighton and @alxarch as well, because this will become more broadly relevant if we overhaul our APIs next quarter.

austinbyers:
I'd love to get input from @nhakmiller and @rleighton and @alxarch as well on this philosophical discussion, which will become even more relevant if we decide to unify and consolidate our APIs next quarter

alxarch:
> I'd love to get input from @nhakmiller and @rleighton and @alxarch as well on this philosophical discussion, which will become even more relevant if we decide to unify and consolidate our APIs next quarter

I understand the logic behind using `*string`, `*bool` etc. The default go behavior works ...most of the time but hits some walls when interacting with languages/systems that treat missing fields differently. OTOH using `*string` et al everywhere has the downside of requiring one extra allocation *for each* field that uses it and requiring to always box/unbox the values in code (ie `aws.String`, `aws.StringValue`). For struct types that will be used in 'hot paths' I think we should use specialized types that behave like `*string` but avoid allocations. One such example is fields in a struct defining a log type event. The struct needs to be unmarshaled/marshaled *for every* line in an input log file and *on each* pointer field in this struct we impose one extra allocation. This adds up quickly.My suggestion for those fields is to use the `strictnull.String` type that is defined in [this module](https://github.com/panther-labs/panther-enterprise/blob/master/internal/log_analysis/log_processor/common/strictnull/string.go). The behavior of this type is as follows:
- If the field is omitted in the input JSON it is considered `null`
- If the field is `null` in the input JSON it is considered `null`
- If the field has a value (even `""`) it is considered non null
- When encoded it will omit `null` and write all non-null values from the input (even `""`)

I have a local branch that I haven't yet pushed to make a PR that adds support for `Int64` and `Float64`. I was expecting to expand the types to cover all scalar values to make the PR (this will happen over the course of the customlogs-backend PR).

Now, in relation to this PR I would not consider the API endpoints as `hot paths`. The performance benefit of avoiding the allocations would be minimal. OTOH if the behavior of the field in question is clearly defined for empty values I think we can use the non-pointer types for those fields. Taking into consideration the update issues I would suggest to either:

1. Keep using pointer types since most API endpoints already include the `aws` package. 
    Pros: no changes! 
    Cons: This would probably leave other parts of the code that use the `models` with the requirement
    to box/unbox/nullcheck all over the place.
2. Revisit this after all scalar types are supported by the `strictnull` package so that we don't mix coding styles
    Pros: users of `models` can handle the zero value as they see fit without forcing null checks, less allocations
    Cons: If this happens in one swoop it will be a big PR, if not we will have 2 coding styles causing confusion as to the 'right' 
    way to do it to new developers.

api/lambda/source/models/api.go
```diff
@@ -94,20 +93,20 @@ type ListIntegrationsInput struct {
...
-	LogTypes           []*string `json:"logTypes,omitempty" validate:"omitempty,min=1"`
+	IntegrationID      string   `json:"integrationId" validate:"required,uuid4"`
+	IntegrationLabel   string   `json:"integrationLabel" validate:"required,integrationLabel,excludesall='<>&\""`
+	CWEEnabled         bool     `json:"cweEnabled"`
+	RemediationEnabled bool     `json:"remediationEnabled"`
+	ScanIntervalMins   int      `json:"scanIntervalMins" validate:"omitempty,oneof=60 180 360 720 1440"`
+	S3Bucket           string   `json:"s3Bucket" validate:"omitempty,min=1"`
+	S3Prefix           string   `json:"s3Prefix" validate:"omitempty,min=1"`
+	KmsKey             string   `json:"kmsKey" validate:"omitempty,kmsKeyArn"`
+	LogTypes           []string `json:"logTypes" validate:"omitempty,min=1"`
```
alxarch, line 104: We should keep the string slices change `[]*string` -> `[]string`  even if we decide to not switch the other fields. `[]string` behaves correctly on updates and is much nicer to use in other parts of the code. `[]*string` makes no sence really.

api/lambda/source/models/api.go
```diff
@@ -141,18 +140,18 @@ type GetIntegrationTemplateInput struct {
 
 // UpdateIntegrationLastScanStartInput is used to update scan information at the beginning of a scan.
 type UpdateIntegrationLastScanStartInput struct {
-	IntegrationID     *string    `json:"integrationId" validate:"required,uuid4"`
-	LastScanStartTime *time.Time `json:"lastScanStartTime" validate:"required"`
-	ScanStatus        *string    `json:"scanStatus" validate:"required,oneof=ok error scanning"`
+	IntegrationID     string    `json:"integrationId" validate:"required,uuid4"`
+	LastScanStartTime time.Time `json:"lastScanStartTime" validate:"required"`
```
alxarch, line 144: `time.Time` is another case that needs special attention. The 'go way' is to use `time.Time.IsZero()` to check for an empty timestamp and the empty timestamp is not Unix zero but `0 AD` a highly unlikely value to have been used on purpose by the user on out systems. A lot of packages use `time.Time` instead of `*time.Time` to set deadlines etc and have explicit checks with `IsZero()` in their code so we would need to always nil check and dereference to use the timestamp anywhere in code.

rleighton:
@alxarch basically said it. We had a long discussion about this issue w.r.t. log processing and plan to make the changes Alex talked about to avoid pointers (allocs) when we can.

I personally, don't think using ptrs when there is no performance issue to deal with nulls is an issue and is consistent with the aws sdk.  It just means that we will have some code doing it differently. Perhaps we leave this issue for now, since a Q3 task is to produce an external api and likely we will refactor much of the internal code I would expect.

kostaspap:
Cool, thanks for chiming in everyone! 

I think the `strictnull` struct is going to be very useful in the log processor where very allocation counts, but I think forcing all of our codebase to use such a package is going to be an overkill. Golang scalars & pointers should be fine for the vast majority of the code; for anything that is performance sensitive, we can make use of `strictnull`. 

I don't think forcing all of our APIs to only use pointers is a good idea. We have been doing that for a while but to me it has shown some weaknesses: For many variables declared as pointers, we end up calling `aws.*Value` in every place they are accessed  to avoid nil dereferences. I have seen that in some recent PRs, others have modified some of the APIs to using scalars since it was just simpler to interact the variables. 

This is the reason that I made initially this PR: I have been working quite some time with `source-api` lately and there are some fields that *we know* will be populated (required fields) and we still have to dereference them. It makes the code more difficult to work with. 

If there are no objections, I am thinking to trim this PR down a bit so that: 
1. I am changing `[]*string` -> `[]string` (I think we all agree having `[]*string` is nonsensical)
2. Change required fields to be scalars instead of pointers. Leave some ambiguous ones (e.g. the `CWEEnabled` that @austinbyers  mentioned) to be as is (pointers). 

We can keep discussing this as we plan for the new API.

nhakmiller:
Recently I made a similar change in the destinations API, which might be what @kostaspap is referring to when he says other APIs have been modified in the same way recently. I made this change because the frontend did not want to get nulls for these fields, even if they were simple scalars, because apparently the forms they use cannot send null back to us very easily.  In this case, where there was a clear and present need to switch away from pointers, I thought the change made sense.

In general though, I don't have a strong preference one way or the other. I could see an approach where we don't use pointers anywhere, except where the difference between a zero value and null value matters. In those places we use pointers. This has the advantage for developers where anytime you see a pointer value being required, they know that the API is treating a null value differently than a zero value and can act accordingly. With IDE's/autocomplete and go's strong typing I don't there is much risk of having a mix of pointer and non-pointer values causing confusion or anything like that. We could still use the strict types in performance critical paths, but I wouldn't want to use them everywhere. They are non-standard and I think having normal values where zero vs nil doesn't matter, pointers where zero vs nil does matter, and strictnull's for where zero vs nil does matter & performance matters is pretty straightforward. In each case, we are using the bare minimum we need for the situation.

As for when/if we should invest the effort to cut everything over, I think we can just do it organically as in the case of this API. Whenever an API is being worked on, if an engineer determines some fields don't need to be pointers they can switch it over. Code reviewers can perhaps help remind by asking questions about the necessity of pointers.

api/lambda/source/models/api.go
austinbyers: Sure, I like that one. Although again, AWS SDK will always expect `[]*string`

PR #1080, opened 2020-06-25
Add firehose permissions to deployment role

austinbyers:
## Background

When testing panther-enterprise v1.5.0, I found the deployment role was missing `firehose:*` permissions for the event bridge delivery stream.

We added that permission in the enterprise repo, but not here in the community edition. We only have one deployment role template, and the one that we had published to S3 does not work for enterprise.

## Changes

- Add `firehose:*` permission for `panther-*` delivery streams to the deployment role

## Testing

- Enterprise deployment with this role
- Copied template from community into enterprise to ensure there were no other differences

I'm not going to move the v1.5.0 git tag, but I did overwrite the v1.5.0 S3 template in the public bucket.

### DIFF
--- a/deployments/auxiliary/cloudformation/panther-deployment-role.yml
+++ b/deployments/auxiliary/cloudformation/panther-deployment-role.yml
@@ -151,8 +151,7 @@ Resources:
             Resource: !Sub arn:${AWS::Partition}:execute-api:*:${AWS::AccountId}:*
           - Effect: Allow
             Action: firehose:*
-            Resource:
-              - !Sub arn:${AWS::Partition}:firehose:*:${AWS::AccountId}:deliverystream/panther-*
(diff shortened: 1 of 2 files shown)


deployments/auxiliary/cloudformation/panther-deployment-role.yml
```diff
@@ -149,6 +149,10 @@ Resources:
           - Effect: Allow
             Action: execute-api:Invoke
             Resource: !Sub arn:${AWS::Partition}:execute-api:*:${AWS::AccountId}:*
+          - Effect: Allow
+            Action: firehose:*
+            Resource:
+              - !Sub arn:${AWS::Partition}:firehose:*:${AWS::AccountId}:deliverystream/panther-*
```
kostaspap, line 155: Can just be in one line?

austinbyers: It could, I just copied exactly what we had in the enterprise repo. I'll change

PR #1089, opened 2020-06-28
Make generic api composable

rleighton:
## Background
While working on the Snowflake database API I needed to extend our database api model with some snowflake specific entry points (routes). Rather than clutter the general model I wanted to just use Go struct composition to add just what was needed.

Before this change, nested structs were not supported.

If we decide to continue using the generic api, this capability might prove useful by allowing common models to be shared easily.

## Changes

- recursively search structs for routes

## Testing

- mage test:ci
- with sf api


pkg/genericapi/router.go
```diff
@@ -135,31 +135,48 @@ func findRequest(lambdaInput interface{}) (*request, error) {
 	structValue := reflect.Indirect(reflect.ValueOf(lambdaInput))
 
 	// Check the name and value of each field in the input struct - only one should be non-nil.
-	var result *request
+	requests := findNonNullPtrs(structValue)
+	switch len(requests) {
+	case 1:
+		break
```
austinbyers, line 141: Nit: `return &requests[0], nil` since every other case also returns.

Then you can delete line 158

rleighton: Done

PR #1090, opened 2020-06-29
Lessen differences between dev & prod front-end environments

3nvi:
## Background

Currently, we use `express` to serve our production app, while we utilize `webpack-dev-server` in order to do local development. This has created a few issues where:

-  express' EJS parser handles template parsing differently than the webpack-related loader (causing exceptions)
- Security headers present in production "block" some network calls (while they don't exist in webpack-dev-server)

This PR aims to force  both dev & prod to "go through" the express server and essentially make sure that the difference between dev & prod  is as little as  possible  (the only difference now is the presence of dynamic file compression on prod)

## Changes

- Deprecated `webpack-dev-server` in favor of `webpack-dev-middleware`
- Add `webpack-hot-middleware` to express and configure it to work in tandem with `react-fast-refresh`
- Remove any template-parameter replacement logic from webpack and move all template-parameter replacements on express
- Conditionally enable express middlewares:
-  - `webpack-***-middlewares` on `development` only
-  - file-compression middlewares on `production` only

## Testing

- Local development worked
- Local "production" mocking  worked
- Docker image was successfully  built

## Notes

The related ticket mentioned "identical" but in reality, making them identical makes the development experience really bad (i.e. adding compression, build-time optimizations, etc.)

This PR makes sure that the **logic** is identical, while minor global & safe optimizations (like file compression) are still present


web/scripts/serve.js
```diff
@@ -16,28 +16,21 @@
...
-
-  return 'no-cache';
-};
+// Set the rendering engine and the location where its views (rendering templates) will exist
+app.set('view engine', 'ejs');
+app.set('views', path.resolve(__dirname, '../dist'));
 
-const addHeaders = (req, res, next) => {
+// Add Security headers to all responses
+app.use('*', (req, res, next) => {
```
s0l0ist, line 33: Maybe out of scope here, but couldn't we just replace our custom logic for security headers to instead use [helmet](https://github.com/helmetjs/helmet)?

3nvi: Actually that is a really good idea. I forgot about `helmet` when the related ticket was created, but we still need to check if `helmet` supports  all the headers we  need

3nvi: Related ticket: https://github.com/panther-labs/panther/issues/1120

PR #1093, opened 2020-06-29
Remove mage migrations

austinbyers:
## Background

There are currently several hundred lines of code in `mage` for upgrading deployments to v1.3.0 and v1.4.0. I want to remove this code now because:

1. Some of our deployments still leverage the original StackSet for self-onboarding - every time we deploy to these accounts, `mage` tries to delete the StackSet and logs a warning when it fails to do so. At best, this causes confusion, at worst this may inadvertently remove incoming data if it should succeed.
2. I do not want developers to see `migrations.go` and think they can add more migrations in mage. Because of pre-packaged deployments, *all* future migrations must be in a custom resource. By removing them from mage, I'm removing the temptation to introduce a bug which wouldn't be caught until release time, if we're lucky.

These migrations are safe to remove because:

- All of our customers and partners are on v1.4 or greater
- Very old deployments can still upgrade to v1.5.1, then upgrade again to v1.6.0+

## Changes

- Fail CI if custom resources are added to the bootstrap stack. This can be easily missed because developers will update their existing deployment, not realizing it wouldn't work on a fresh deployment.
- Remove migrations in `mage` code for 1.3.0 and 1.4.0

## Testing

- `mage test:ci`


tools/mage/test_namespace.go
```diff
@@ -172,7 +172,16 @@ func testCfnLint() error {
 		}
 
 		if template == bootstrapTemplate {
-			// Custom resources can't be in the bootstrap stack, skip remaining checks
+			// Custom resources can't be in the bootstrap stack
+			for logicalID, resource := range body["Resources"].(map[string]interface{}) {
+				t := resource.(map[string]interface{})["Type"].(string)
+				if strings.HasPrefix(t, "Custom::") {
+					return fmt.Errorf("%s: %s: custom resources will not work in this stack - use bootstrap-gateway instead",
```
nhakmiller, line 179: Interesting, I wasn't aware of this. Could you give the quick explanation why this is the case?

austinbyers: The `bootstrap` stack is the very first Panther stack which is deployed. The Lambda function which handles custom resources is created in `bootstrap-gateway`, right after `bootstrap`. Therefore, if you tried to add a custom resource to the very first stack, a fresh deployment would fail because the function doesn't exist.

Why do we put the custom resource Lambda in `gateway` and not `bootstrap`? Because when deploying from source, the Lambda source must be uploaded to S3, but the source S3 bucket is created in the bootstrap stack

austinbyers:
> For people using pre-packaged deployments this code never runs anyways right?

Correct

> Do you think it would be a disproportionate amount of effort to add a little warning message though for people doing upgrades from a pre 1.5.0 to a post 1.5.0 version?

To clarify, migrations would only be affected from v1.3 or earlier to v1.6+ If you're on v1.4, you've already passed all of the migrations here.

We could add a quick check for the `PantherVersion` tag on the bootstrap stack and log a warning, that seems reasonable

PR #1106, opened 2020-07-01
Changing API for sources-api to return null when a timestamp is not configured

kostaspap:
## Background

Changing API for sources-api to return null when a timestamp is not configured. Current API was returning `0001-01-01T00:00:00Z` when a timestamp was not configured (this is zero time for Golang's `time.Time`). 
However the above default value doesn't make sense for other languages, it is Go specific. 
Given that the APIs are also used by the frontend, it makes sense that optional `time.Time` fields are defined as pointers, so that they are omitted when they are serialized. 

## Changes

- Changed optional time fields from source-api to be `*time.Time`

## Testing

- `mage test:ci`
- Deployed and verified behavior.

### DIFF
--- a/internal/core/source_api/ddb/items.go
+++ b/internal/core/source_api/ddb/items.go
@@ -22,32 +22,32 @@ import "time"
 
 // Integration represents an integration item as it is stored in DynamoDB.
 type Integration struct {
-	CreatedAtTime    time.Time `json:"createdAtTime"`
-	CreatedBy        string    `json:"createdBy"`
-	IntegrationID    string    `json:"integrationId"`
-	IntegrationLabel string    `json:"integrationLabel"`
-	IntegrationType  string    `json:"integrationType"`
+	CreatedAtTime    time.Time `json:"createdAtTime,omitempty"`
+	CreatedBy        string    `json:"createdBy,omitempty"`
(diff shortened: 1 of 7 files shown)


internal/core/source_api/ddb/items.go
```diff
@@ -22,32 +22,32 @@ import "time"
...
+	LogProcessingRole string   `json:"logProcessingRole,omitempty"`
 }
 
 type IntegrationStatus struct {
-	ScanStatus        string    `json:"scanStatus"`
-	EventStatus       string    `json:"eventStatus"`
-	LastEventReceived time.Time `json:"lastEventReceived"`
+	ScanStatus        string     `json:"scanStatus,omitempty"`
+	EventStatus       string     `json:"eventStatus,omitempty"`
+	LastEventReceived *time.Time `json:"lastEventReceived,omitempty"`
```
kostaspap, line 52: I marked all DDB fields as `omitempty`. While some of these fields may always be present, i thought it might be better to define a "standard" approach for all the fields that makes sense for DDB case, instead of cherry-picking which fields are optional and which required. Which field is required and which is optional is defined at API level, by defining everything as `omitempty`, the DDB layer doesn't need to know about the context of each field.

PR #1112, opened 2020-07-02
Fix for Jira Destination form

### LINKED ISSUE #1063: JIRA Destination not Working
### Describe the bug

A JIRA destination (properly configured as far as I can tell) is failing to send messages, logging WARN's in the `alert_delivery` lambda until eventually exceeding the max retries and erroring out.

### Steps to reproduce

1. Go to destinations
2. Configure a JIRA destination with all the required fields
3. Trigger a policy/rule at the severity of the JIRA destination
4. Observe WARN's in the `alert_delivery` lambda until the max retries are exceeded, at which point it errors out

### Expected behavior

A properly configured JIRA destination will deliver messages.

### Environment

- Tagged release 1.4

### Additional context

This was observed in another user's account, I don't have access to the original JIRA instance where this was observed so I am mostly troubleshooting from the Panther side.

When attempting to send to a JIRA destination with the issue type `Bug` I see this error:
```
"error": "request failed: 400 Bad Request: {\"errorMessages\":[],\"errors\":{\"issuetype\":\"issue type is required\"}}"
```

When attempting to send to a JIRA destination with the issue type `Task` I see this error:
```
    "error": "request failed: 400 Bad Request: {\"errorMessages\":[],\"errors\":{\"summary\":\"Field 'summary' cannot be set. It is not on the appropriate screen, or unknown.\",\"description\":\"Field 'description' cannot be set. It is not on the appropriate screen, or unknown.\"}}"
```

alexmylonas:
## Background

We were previously having an `issueType` field on the Jira destination form as a dropdown (Bug,Task etc). This was causing issues since jira their's users to customize those as desired. With this PR, I changed dropdown to text input field

Closes #1063 

## Changes

- Changed `issueType` field to text input
- Made username a required form field

## Testing

- Locally

Screenshot 2020-07-02 at 3 34 07 PM


nhakmiller:
> LGTM.
> 
> We just have to make sure the backend doesn't have a validation requirement on the previous enum fields. I'm assuming it doesn't since you've tested this change.

I remember checking the backend for this and didn't see anything, which I was surprised by on account of the frontend validation. Should be good I think.

web/src/components/forms/JiraDestinationForm/JiraDestination.tsx
```diff
@@ -63,15 +60,15 @@ const JiraDestinationForm: React.FC<JiraDestinationFormProps> = ({ onSubmit, ini
       <Field
         as={FormikTextInput}
         name="outputConfig.jira.orgDomain"
-        label="Organization Domain"
+        label="Organization Domain*"
```
3nvi, line 63: we normally to `* {LABEL}`. 

For example, in  the RuleForm  we have the following prp `label="* ID"`

web/src/components/forms/JiraDestinationForm/JiraDestination.tsx
```diff
@@ -105,13 +102,12 @@ const JiraDestinationForm: React.FC<JiraDestinationFormProps> = ({ onSubmit, ini
...
-        as={FormikCombobox}
+        as={FormikTextInput}
         name="outputConfig.jira.issueType"
-        label="Issue Type"
+        label="Issue Type*"
         mb={6}
         aria-required
-        items={Object.keys(JiraIssueTypesEnum)}
-        inputProps={{ placeholder: 'Select a type of issue' }}
+        placeholder="What type of issue you want us to create? i.e. Bug"
```
3nvi, line 110:
```suggestion
        placeholder="What type of issue you want us to create?"
```

3nvi: Can we add a helper text  below (in the new pounce there is a component named `FormHelperText`) to  say something  along the lines of:

`Typical values include Bug, Task, etc.` or  something like that

PR #1116, opened 2020-07-02
Rebuild global layer when modified from web UI

nhakmiller:
## Background

When we added support for multiple globals we moved the `rebuild global layer` API call out of the `WriteItem` helper that all the analysis-api endpoints rely on to write to dynamo and into the API endpoints directly. But we forgot to do this for the `ModifyGlobal` endpoint. This meant that the global layer was only rebuilt when new globals were created, globals were deleted, or a global was uploaded via the bulk upload endpoint. Modifying a single global (e.g. via the web UI) did not actually rebuild the global layer. This fixes that.

## Changes

- Rebuild the global layer after a global is modified via the web UI

## Testing

- Deploying into dev account now to verify

### DIFF
--- a/internal/core/analysis_api/handlers/sqs.go
+++ b/internal/core/analysis_api/handlers/sqs.go
@@ -48,7 +48,7 @@ func queuePolicy(policy *tableItem) error {
 //
 // Currently only the global type is supported, but in the future we may support other types as
 // well.
-func updateLayer(_ string) error {
+func updateLayer() error {
 	_, err := sqsClient.SendMessage(&sqs.SendMessageInput{
 		MessageBody: aws.String(typeGlobal),
 		QueueUrl:    aws.String(env.LayerManagerQueueURL),
--- a/internal/core/analysis_api/handlers/sqs.go
+++ b/internal/core/analysis_api/handlers/sqs.go
@@ -43,10 +43,14 @@ func queuePolicy(policy *tableItem) error {
 	return err
 }
 
-// updateLayer sends a message to the layer manager lambda indicating a layer of a certain type needs to be re-built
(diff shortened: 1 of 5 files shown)


internal/core/analysis_api/handlers/sqs.go
```diff
@@ -43,10 +43,14 @@ func queuePolicy(policy *tableItem) error {
...
 }
 
-// updateLayer sends a message to the layer manager lambda indicating a layer of a certain type needs to be re-built
-func updateLayer(analysisType string) error {
+// updateLayer sends a message to the layer manager lambda indicating a layer of a certain type
+// needs to be re-built
+//
+// Currently only the global type is supported, but in the future we may support other types as
+// well.
+func updateLayer(_ string) error {
```
austinbyers, line 51: Let's either remove the parameter (and hardcode `typeGlobal` below)

Or keep the argument, and pass it to `MessageBody`

Here, it's confusing because you're both passing `typeGlobal` above, but then ignoring its value here and hardcoding it instead

nhakmiller: When I leave the argument in I get a linting error that fails `test:ci`:

```
internal/core/analysis_api/handlers/sqs.go:51:18: `updateLayer` - `analysisType` always receives `typeGlobal` (`"GLOBAL"`) (unparam)
func updateLayer(analysisType string) error {
                 ^
14:03:44        ERROR       X golangci-lint failed (9/14): running ".setup/golangci-lint run --timeout 10m -j 1" failed with exit code 1
```

I think the linter must have been updated recently or something because my original change didn't touch this file at all but `test:ci` failed which is why I made these changes. I'll just remove the parameter for now, not too hard.

PR #1117, opened 2020-07-02
Prevent persistence after user deletion

austinbyers:
## Background

When a Panther user is deleted, we remove them from the Cognito pool immediately, but their access token is still valid for up to 1 hour. That is a common limitation in many auth systems, Cognito included - we don't have any way to invalidate that token nor is the expiration configurable.

But we can take some steps to limit the damage a deleted account can do. In particular, today it is possible for a deleted account to continue to create new user accounts every hour to maintain indefinite access to the system. To prevent that, all user management write operations now require proof of the requester identity, which is validated against the Cognito pool.

We will address this more fundamentally in the next quarter with some kind of unified API where a single point of entry can handle extra validation of this kind for *all* requests, not just extra-sensitive ones like user management.

## Changes

- Invoke `AdminGlobalSignOutUser` before deleting a user - this will invalidate their *refresh* token
- All `users-api` modify operations (Invite, Delete, Update, ResetPassword) now require a `requesterId`. This ID is set by AppSync from the login token, so it cannot be spoofed from the user's browser. The `users-api` verifies the requester is still a valid account in Cognito before continuing.

## Testing

- Fresh deploy
- New integration test to ensure `InviteUser` fails if the requester no longer exists in Cognito
- Tested user invite/update/delete/passwordReset from the Panther web app
- Using a deleted user session, tried to invite another user:

Screen Shot 2020-07-02 at 1 32 32 PM

### DIFF
--- a/api/lambda/users/models/api.go
+++ b/api/lambda/users/models/api.go
@@ -67,6 +67,9 @@ type GetUserOutput = User
 
 // InviteUserInput creates a new user with minimal permissions and sends them an invite.
 type InviteUserInput struct {
+	// Which Panther user is making this request?
+	RequesterID *string `json:"requesterId" validate:"required,uuid4"`
+
 	GivenName  *string `json:"givenName" validate:"required,min=1,excludesall='<>&\""`
 	FamilyName *string `json:"familyName" validate:"required,min=1,excludesall='<>&\""`
 	Email      *string `json:"email" validate:"required,email"`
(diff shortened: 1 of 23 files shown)


api/lambda/users/models/api.go
```diff
@@ -115,13 +118,16 @@ type ListUsersInput struct {
 //    ]
 // }
 type ListUsersOutput struct {
-	Users []*User `json:"users"`
+	Users []User `json:"users"`
```
austinbyers, line 121: I didn't want to take the time now to remove *all* pointers from the api, but we've agreed that `[]*struct` is the most egregious, so I went ahead and made that change

internal/core/users_api/cognito/delete_user.go
```diff
@@ -26,16 +26,23 @@ import (
 	"github.com/panther-labs/panther/pkg/genericapi"
 )
 
-// DeleteUser calls cognito api delete user from a user pool
 func (g *UsersGateway) DeleteUser(id *string) error {
```
nhakmiller, line 29: Should we validate requester here? Wouldn't want a bad admin to go delete all the other admins.

austinbyers: Yes, this happens one step up in the call stack (the `api` package, not the `cognito` package)

austinbyers: https://github.com/panther-labs/panther/pull/1117/files/f8652d5b6797d88a46d75858ad001f4e1da42e84#diff-24c1c2c99953b0651b93f94b1d7ab21aR28

PR #1123, opened 2020-07-06
Updated Totp form to showcase code instead of only showing QR

### LINKED ISSUE #1077: MFA Visual Enhancements
### Description

Based on feedback from users, we need to improve some minor aspects in our MFA screen. Specifically, the issues are that:

- MFA TOTP is just named "Organization"
- Only QR code is presented and not the textual representation

To combat that  we need  to investigate how can we rename aspects of the TOTP parameters to rename the word `Organization` to something that makes more sense. We also need to show the related `otpauth` URL without the need for QR code scan

### Acceptance Criteria

- Remove the word `Organization` from the OTP URL parameter (see `formatSecretCode` under `helpers/utils`)
- Show the secret code under the QR code, so users that can't (or don't want ) to scan the QR code, can see the  OTP URL directly

alexmylonas:
## Background

We are currently only showing the QR code to allow users to save their MFA configuration. With this PR, we also showing the code generated by cognito that's used to generate MFA's. In addition, we limit input length to 6 digits. 

Closes #1077 
## Changes

- Display secret code for MFA configuration
- Limit MFA input to 6 chars

## Testing

- Locally 

Screenshot 2020-07-06 at 1 02 25 PM


web/src/helpers/utils.tsx
```diff
@@ -44,8 +44,7 @@ export const isMobile = /Mobi|Android/i.test(navigator.userAgent);
 
 // Generate a new secret code that contains metadata of issuer and user email
 export const formatSecretCode = (code: string, email: string): string => {
-  const issuer = 'Panther';
-  return `otpauth://totp/${email}?secret=${code}&issuer=${issuer}`;
+  return `otpauth://totp/${email}?secret=${code}`;
```
3nvi, line 47: Need some context as to why we are removing `issuer`? What was the problem with that?

web/src/components/forms/TotpForm/TotpForm.tsx
```diff
@@ -70,10 +70,18 @@ export const TotpForm: React.FC = () => {
             bgColor={theme.colors['navyblue-800']}
           />
         </Flex>
+        <Box color="gray-200" textAlign="center" my={8}>
+          Or enter manually the secret code:
+          <Tooltip content={'Click to copy'}>
+            <Box mt={2} cursor="pointer" onClick={() => copyTextToClipboard(code)}>
+              {code}
```
3nvi, line 77: This seems to pollute the UI a bit. Can we add an inline button (looking like a link) to **either**:

- trigger "copy to clipboard"
- trigger "show code" where the code will be shown after this click.

Thoughts @alexmylonas @vorillaz

vorillaz: My 0.2£:
Personally, I would implement a separate component with a link and button look and feel, block element should not inherit or act as triggering elements. Semantically and A11Y wise this can cause problems regarding an A11Y audit, semantics parsed through the browser and screen readers. 😖

3nvi: Hmm perhaps I didn't express myself that well. I was envisioning something like:

```jsx
<p>
   Or enter the code manually
   <button onClick={handleClick}>show code</button>
</p>
```

I  was just saying that the button should look like a "link" with CSS (kinda like the way  the "small text" is on the screenshot of the PR). 

If I understand correctly you are ok with what I wrote, but you want the button to actually **look** like a button, correct? Am I missing something else?

3nvi:
@alexmylonas @vorillaz let's help close this

web/src/components/forms/TotpForm/TotpForm.tsx
```diff
@@ -70,10 +70,18 @@ export const TotpForm: React.FC = () => {
             bgColor={theme.colors['navyblue-800']}
           />
         </Flex>
+        <Box color="gray-200" textAlign="center" my={8}>
+          Or enter manually the secret code:
+          <Tooltip content={'Click to copy'}>
+            <Box mt={2} cursor="pointer" onClick={() => copyTextToClipboard(code)}>
```
3nvi, line 76: @alexmylonas can we use `AbtractButton` for it? Clickable element should be a button and not a `<div>`

web/src/helpers/utils.tsx
alexmylonas: Wasn't that part of the AC1 on #1077? If I not i will revert this.

3nvi: They reported  that something said the  word "Organization"

> MFA TOTP is just named "Organization"

I don't exactly know what they mean. Can we try adding the MFA to DUO and  see if we see the MFA entry named "Organization" instead of "Panther"? I  think that's what they meant. Most likely  we need some additional URL params to the code, but not 100% sure

web/src/components/forms/TotpForm/TotpForm.tsx
alexmylonas: Made some changes, added a show code button, removed tooltip, removed copyToClipboard. Check screenshots:

Screenshot 2020-07-09 at 1 33 45 PM

Screenshot 2020-07-09 at 1 33 51 PM

Let's nail down UX and implementation before making any other change

web/src/components/forms/TotpForm/TotpForm.tsx
alexmylonas: Yeap, replaced `Box` with `AbstractButton`!

web/src/components/forms/TotpForm/TotpForm.tsx
3nvi: UI is fully  subjective here. I personally prefer:

- an inline button that looks like a link (cause we don't want to have 2 primary action buttons getting the user's attention)
- The "hide  code" to dissapear once the code is shown (to reduce on-screen elements, since the button is useless after you click it)

alexmylonas: Screenshot 2020-07-10 at 2 26 49 PM

@3nvi Latest update, let me if you are ok with this so i can merge this!

Show Code is hidden on click

3nvi: This is perfect now. Resolving.

web/src/helpers/utils.tsx
3nvi: @alexmylonas can we get an update on this?

alexmylonas: This seems to the case. DUO ignores this info but MS authenticator shows this as kind of label. 

"Organization" feels generic, i think our options is either leave it as "Panther", leave it empty or try to get the name of the company

3nvi: we always want to name it `Panther`. My question is why was this shown as `Organization` since it was clearly set by us to `Panther`?

alexmylonas: I have added also label in the URL link as described [here](https://github.com/google/google-authenticator/wiki/Key-Uri-Format). 

I think that's the best we can do at this point. MS and Google Authenticator picks "Panther" as expected but DUO ignores both label and issuer at this point

Issue #610, opened 2020-03-30
Rules don't verify LogType for unit tests


nhakmiller:
### Describe the bug

When creating and editing rule tests, you can set a log type that is not one of the supported log types for that rule and the tests will still run and pass.

### Steps to reproduce

1. Go to the rule editor
2. Create a new test case
3. Set the log type of the test case to be a log type not supported by the rule
4. Run the test
5. The rule is evaluated as normal and will return passing if the event defined in the test case passes

### Expected behavior

A test case for a log type that a rule is not applied to doesn't pass.

### Environment

How are you deploying or using Panther?

- Panther version 1.0

### Additional context

I believe we can actual remove this field both here and in the policy pages, and instead have this be implied based on the rule/policy being tested if necessary.

nhakmiller:
Additional context from internal discussions:

```
Nick
Looks like the rules engine does not verify log type during testing

Whereas the policy engine will return this if the resource type is specified incorrectly for a test:
ERROR
<Policy Name>
test resource type AWS.CloudTrail is not applicable to this policy

Kostas
I don’t thing we need the log type?
I don’t see why the engine should fail

Nick
I was going to make the same comment on the github issue if one got created, but I think the field is totally extraneous at this point

Kostas
I mean, it’s just an event in the end
It could pass or fail depending on the python code
But not sure it makes sense to verify anything else
I mean, could be a CloudTrail event but have the label “alb” :P

Austin
With policies, I believe the resource type is part of the log itself (we add it) - you could write a generic policy which applies to all resources and then switch on the type in the python code
I’m not sure if the same is true for logs
But I think that’d be reasonable to do the same thing - you could imagine a rule which is almost identical across two different log types except for one edge case, in which case having two different rules completely would be overkill

Nick
It exists because in the policy engine we take a list of resources and a list of policies, and apply each policy to the resource that have the right type. So for testing purposes, we need the test cases to specify a resource type that the policy applies to. We mimicked this UI into the rules section. The rules engine though actually has explicit logic for handling tests, which don't consider the type at all.
That's true Austin, but I just double checked the code to verify and we're no longer using the resource object in the policy engine so those fields would just be normal fields that could be defined in the unit test case like any other I think
```

kostaspap:
I don't see this is a bug. As I see it, it is the expected behavior. The rule could be used for the additional log type and it would work properly. Why would we throw an exception for something that actually passes successfully the Python rule?
[...]

PR #1124, opened 2020-07-06
Remove policy test resourcetype fields

giorgosp:
## Background

The Resource Type (and Log Type) field in Policy/Rule unit tests are not needed and don't offer any extra feature
to the user. See #610 for more context.

## Changes

These fields are removed both from the web and backend/api code.

## Testing

With `mage test:integration` and manually.

@3nvi the frontend change (cleaning up the fields) seemed trivial and went ahead and implemented it in the 3rd commit but I can split it to a different issue/PR if you wish.

### DIFF
--- a/internal/core/analysis_api/handlers/test_policy.go
+++ b/internal/core/analysis_api/handlers/test_policy.go
@@ -116,12 +116,9 @@ func TestPolicy(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyR
 			testResults.TestsPassed = append(testResults.TestsPassed, string(test.Name))
 
 		default:
-			// The test case had a resource type that the policy did not apply to, consider an error for now.
-			testResults.TestsErrored = append(testResults.TestsErrored, &models.TestErrorResult{
-				ErrorMessage: "test resource type " + string(test.ResourceType) + " is not applicable to this policy",
-				Name:         string(test.Name),
-			})
-			testResults.TestSummary = false
+			// This test didn't run (result.{Errored, Failed, Passed} are all empty). This must not happen absent a bug.
+			zap.L().Error("unable to run test for resourceID", zap.String("resourceID", result.ID))
+			return &events.APIGatewayProxyResponse{StatusCode: http.StatusInternalServerError}
 		}
 	}
 
@@ -143,7 +140,7 @@ func getRuleResults(input *models.TestPolicy) (*enginemodels.RulesEngineOutput, 
 		inputEvents[i] = enginemodels.Event{
 			Data: attrs,
 			ID:   testResourceID + strconv.Itoa(i),
-			Type: string(test.ResourceType),
+			Type: input.ResourceTypes[0],
 		}
 	}
 
@@ -196,7 +193,7 @@ func getPolicyResults(input *models.TestPolicy) (*enginemodels.PolicyEngineOutpu
 		resources[i] = enginemodels.Resource{
 			Attributes: attrs,
 			ID:         testResourceID + strconv.Itoa(i),
-			Type:       string(test.ResourceType),
+			Type:       input.ResourceTypes[0],
 		}
 	}
 
--- a/internal/core/analysis_api/handlers/test_policy.go
+++ b/internal/core/analysis_api/handlers/test_policy.go
@@ -140,7 +140,7 @@ func getRuleResults(input *models.TestPolicy) (*enginemodels.RulesEngineOutput, 
 		inputEvents[i] = enginemodels.Event{
 			Data: attrs,
 			ID:   testResourceID + strconv.Itoa(i),
-			Type: input.ResourceTypes[0],
+			Type: input.ResourceTypes[0], // TODO(giorgosp): Seems not used by the rules engine
 		}
 	}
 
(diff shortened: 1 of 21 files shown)


internal/core/analysis_api/handlers/test_policy.go
```diff
@@ -143,7 +140,7 @@ func getRuleResults(input *models.TestPolicy) (*enginemodels.RulesEngineOutput,
 		inputEvents[i] = enginemodels.Event{
 			Data: attrs,
 			ID:   testResourceID + strconv.Itoa(i),
-			Type: string(test.ResourceType),
+			Type: input.ResourceTypes[0],
```
kostaspap, line 143: * Why do we always pick the first one? 
* Is the type even needed?

internal/core/analysis_api/main/lambda.go
```diff
@@ -51,6 +53,16 @@ var methodHandlers = map[string]gatewayapi.RequestHandler{
...
+	"GET /local": local,
+}
+
+func local(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyResponse {
+	fmt.Println(request)
+	return &events.APIGatewayProxyResponse{
+		Body:       `{"hello":"json"}`,
+		StatusCode: 200,
+	}
 }
```
kostaspap, line 66: leftover?

internal/core/analysis_api/handlers/test_policy.go
giorgosp: It doesn't matter which one we choose. It's actually a hack in order to reuse the policy engine code. We "trick" it to run our policy by giving a resource type that matches one of the policy's resource types.

internal/core/analysis_api/main/lambda.go
giorgosp: Yeap :) Good catch

internal/core/analysis_api/handlers/test_policy.go
giorgosp: Added a comment inside the code as well

internal/core/analysis_api/handlers/test_policy.go
```diff
@@ -143,7 +140,7 @@ func getRuleResults(input *models.TestPolicy) (*enginemodels.RulesEngineOutput,
 		inputEvents[i] = enginemodels.Event{
 			Data: attrs,
 			ID:   testResourceID + strconv.Itoa(i),
-			Type: string(test.ResourceType),
+			Type: input.ResourceTypes[0], // TODO(giorgosp): Seems not used by the rules engine
```
kostaspap, line 143: If it is not used, why not just remove it?

giorgosp: Need to check it a bit more cause integration tests failed.

internal/core/analysis_api/handlers/test_policy.go
```diff
@@ -196,7 +193,9 @@ func getPolicyResults(input *models.TestPolicy) (*enginemodels.PolicyEngineOutpu
 		resources[i] = enginemodels.Resource{
 			Attributes: attrs,
 			ID:         testResourceID + strconv.Itoa(i),
-			Type:       string(test.ResourceType),
+			// The engine picks a policy to run based on the input resource type. To make the engine run the
+			// input policy, we just pass one of its resource types in the input resource.
+			Type: input.ResourceTypes[0],
```
nhakmiller, line 198: If no resource type is set for a policy, then it is considered to be applied to all resource types. I know this is a bit confusing because we don't support this functionality easily in the `panther_analysis_tool` (should work if you set `ResourceTypes` to `nil` but its unintuitive). In that case though I think this would be an index out of bounds error.

giorgosp: Ah, I see. Adding a default test resource type in case the policy is applicable to all resources. 
Would be nice if the catch-all resource type was an explicit value (sth like `__ALL__`) but I guess it will complicate things for users and existing polices if we change it now.

nhakmiller: I agree, I had also suggested we use `ALL` instead of leaving it blank back a long time ago when this was implemented but the consensus was to go with leaving it blank. A default resource type in case none is specified in the policy should do the trick.

internal/core/analysis_api/handlers/test_policy.go
kostaspap: Sure, let's check and remove it if not needed (as suggested by the TODO)

web/src/components/forms/BaseRuleForm/BaseRuleFormTestSection/BaseRuleFormTestSection.tsx
```diff
@@ -174,15 +171,6 @@ const BaseRuleFormTestSection: React.FC = () => {
                     placeholder="The name of your test"
                     label="Name"
                   />
-                  <Field
```
3nvi, line 177: Please change

```
<Grid columnGap={5} templateColumns="1fr 1fr 2fr" mt={2} mb={6}>
```
to
```
<Grid columnGap={5} templateColumns="1fr 2fr" mt={2} mb={6}>
```

since a grid-entry got removed

giorgosp: Done!

internal/core/analysis_api/handlers/test_policy.go
giorgosp: Removed and added a couple of more tests for testing rules. I think we are ready to go on this one.

PR #1126, opened 2020-07-07
Updating Gitlab production logs

kostaspap:
## Background

Marking the "format" field as optional - seen in HTTP 302 response

## Changes

- Updated parser
- Updated test

## Testing

- unit tested with sample log


internal/log_analysis/log_processor/parsers/gitlablogs/testdata/productionlog_samples.jsonl
```diff
@@ -780,3 +780,4 @@
...
+{"method":"GET","path":"/index.php","format":null,"controller":"ApplicationController","action":"route_not_found","status":302,"location":"http://34.222.254.254/users/sign_in","time":"2020-07-05T19:35:32.337Z","params":[{"key":"s","value":"/Index/\\think\\app/invokefunction"},{"key":"function","value":"call_user_func_array"},{"key":"vars","value":{"0":"md5","1":["HelloThinkPHP"]}},{"key":"unmatched_route","value":"index"}],"remote_ip":"195.54.254.254","user_id":null,"username":null,"ua":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36","queue_duration_s":0.007208,"correlation_id":"DScuWzVUYA5","meta.caller_id":"ApplicationController#route_not_found","redis_calls":1,"redis_duration_s":0.000292,"cpu_s":0.02,"db_duration_s":0.0,"view_duration_s":0.0,"duration_s":0.00341,"tag":"test"}
```
alxarch, line 783: From the sample it appears there is an extra string field for redirect responses called `location` that's a URL. It might be wise to add this to the event as well.

kostaspap: Good catch. Yes, makes a lot of sense to add this field, I will do so and migrate this sample to a unit test.

internal/log_analysis/log_processor/parsers/gitlablogs/rails.go
```diff
@@ -52,6 +52,8 @@ type Production struct {
 	DBDurationSeconds     *float32           `json:"db_duration_s,omitempty" description:"Total time to retrieve data from PostgreSQL"`
 	ViewDurationSeconds   *float32           `json:"view_duration_s,omitempty" description:" Total time taken inside the Rails views"`
 	DurationSeconds       *float32           `json:"duration_s" validate:"required" description:"Total time taken to retrieve the request"`
+	MetaCallerID          *string            `json:"meta.caller_id,omitempty" description:"Caller ID"`
+	Location              *string            `json:"location" description:"(Applies only to redirects) The redirect URL"`
```
kostaspap, line 56: @alxarch it seems that the new log had two new fields, added both. I am seeing that the biggest difference between GitLab.Production and Gitlab.API logs is the API logs have a "severity" field.

internal/log_analysis/log_processor/parsers/testutil/testutil.go
```diff
@@ -177,3 +177,8 @@ func CheckPantherMultiline(t *testing.T, logs string, parser parsers.LogParser,
 		EqualPantherLog(t, expect, []*parsers.PantherLog{result}, nil)
 	}
 }
+
+func NewRawMessage(jsonString string) *jsoniter.RawMessage {
+	rawMsg := (jsoniter.RawMessage)(jsonString)
+	return &rawMsg
+}
```
kostaspap, line 184: This function used to be present in `internal/log_analysis/log_processor/parsers/awslogs/utils_test.go`, I just moved it here. 
I considered changing its signature to make it return `jsoniter.RawMessage`, but this required significant changes in other parts of the codebase (AWS log schemas & tests) that use this method. I decided to refrain to avoid scope creep.

alxarch: I would put this in `box` package as `RawMessageJSON(jsoniter.RawMessage) *jsoniter.RawMessage`
and use `box.RawMessage(jsoniter.RawMessage(str))` in the tests

alxarch: or just `box.JSON` since we use `box` package as a utility and short names matter.

internal/log_analysis/log_processor/parsers/gitlablogs/rails.go
alxarch: They are based on the same Rails logger internally so we should expect breakage if they decide to add `severity` based logging to the production logs as well. I assume that they use a different logger config for API currently. On their side it should be considered 'backwards compatible' to do that.

internal/log_analysis/log_processor/parsers/testutil/testutil.go
kostaspap: Looking at the code again I realized that the `jsoniter.RawMessage` is just `[]byte`. So in the new Gitlab tests I can just initialize the values with `[]byte("...")`. The suggestion was good, but seems the change was not really needed.

PR #1128, opened 2020-07-07
Invalidate sessions on Password Change

3nvi:
## Background

Currently, when a user changes their password they don't get (instantly) signed out of all existing devices. This PR remediates 
that by forcing a `globalSignOut` on behalf of the user.

Closes [#325](https://github.com/panther-labs/panther-enterprise/issues/325)

## Changes

- Call `globalSignOut` on self-deletion & password change
- Properly handle pervasive "sign out" by closing the related Drawer

## Testing

- Successfully changed password and got other sessions to invalidate

## Notes

A refresh action needs to be performed  for this to take effect


web/src/components/forms/EditProfileForm/EditProfileForm.tsx
```diff
@@ -56,11 +56,13 @@ const EditProfileForm: React.FC<EditProfileFormProps> = ({ onSuccess }) => {
       }),
   });
 
+  /* eslint-disable camelcase */
   const initialValues = {
-    email: userInfo.email || '',
-    familyName: userInfo.family_name || '',
-    givenName: userInfo.given_name || '',
+    email: userInfo?.email || '',
```
3nvi, line 61: I think signout causes a quick re-render of the current componentst. If I didn't add `?`, the component would cause an exception, cause  it attempted to  re-render for a  single second before the app went to the sign-in screen

This small "hack" handles an edge case scenario where having  the "edit account drawer" open while signing out (through password change), would cause `EditProfileForm`  to re-render with `userInfo` being `null`

PR #1129, opened 2020-07-07
Sync zap logger on `lambdalogger.ConfigureGlobal`

alxarch:
## Background

The current state of `lambdalogger` uses `zap.ReplaceGlobals()` on each lambda invocation.
Apart from the overhead of creating a new zap logger on each lambda invocation the implementation does not call `Sync()`
on the previous global logger. This has the potential of losing buffered log lines from the last Lambda invocation.
This patch forces the `Sync` on the logger that gets replaced to avoid this. It would be desired to avoid replacing the global logger on each lambda invocation but such a change would be quite big and should be considered in a separate PR.

## Changes

- Call `zap.Logger.Sync()` on the logger that gets replaced by `lambdalogger.ConfigureGlobal()` to ensure all log lines are written to stdout

## Testing

- mage test:ci

### DIFF
--- a/pkg/lambdalogger/logger.go
+++ b/pkg/lambdalogger/logger.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	"github.com/aws/aws-lambda-go/lambdacontext"
+	"github.com/pkg/errors"
 	"go.uber.org/zap"
 )
 
@@ -88,6 +89,10 @@ func ConfigureGlobal(
 		log.Panic("failed to build zap logger: " + err.Error())
 	}
 
+	// Sync the current global logger so we don't miss any log lines
+	if err := zap.L().Sync(); err != nil {
+		panic(errors.Wrap(err, "failed to sync logger"))
+	}
 	zap.ReplaceGlobals(logger)
 	return lc, logger
 }


internal/log_analysis/log_processor/parsers/juniperlogs/audit.go
```diff
@@ -32,11 +32,6 @@ import (
 
 const TypeAudit = `Juniper.Audit`
 
-// nolint:lll
-const DescAudit = TypeAudit + ` The audit log contains log entries that indicate non-idempotent (state changing) actions performed on WebApp Secure.
-
-Reference: https://www.juniper.net/documentation/en_US/webapp5.6/topics/reference/w-a-s-incident-log-format.html`
-
```
kostaspap, line 39: Leftover from local testing? (PR is above syncing logger so it should just be about pkg/lambdalogger/logger.go)

pkg/lambdalogger/logger.go
```diff
@@ -88,6 +88,10 @@ func ConfigureGlobal(
 		log.Panic("failed to build zap logger: " + err.Error())
 	}
 
+	// Sync the current global logger so we don't miss any log lines
+	if err := zap.L().Sync(); err != nil {
+		log.Panic("failed to flush zap logger: " + err.Error())
```
kostaspap, line 93: Should this be best effort ->  if syncing failed... move on?
Is panic helping here?

alxarch: If syncing failed it probably means we cannot log anymore. The `zap` global logger as constructed here writes to `stdout`. Something must have gone completely wrong in the process state if we fail to `Sync()`. In addition this function does not return an error so we cannot handle this in any meaningful way.

alxarch: In general I consider missing log lines a fatal error for our lambdas. Especially since we will be basing metrics on the `zap.Logger` as was decided.

alxarch: TBH this should be changed to just `panic(err)` since we are probably missing the error again by feeding it to the logger that just failed to `Sync()` :)

alxarch:
This appears to hit a 'wontfix' bug on `zap` itself. The issue is complex and involves the following paramters:
- The default configuration for the `zap.Logger` uses `stderr` and `stdout` directly without buffering 
- `go` actually buffers `stdout` and `stderr` on new lines `\n`
- The underlying `syscall` seems to return errors when syncing on these 'special' files on Linux and OSX
The result of this is that *when using the default `zap` config* the OS handles buffering and fails to actually 'sync' the loggers.
The major implication is that as we stand now *we have no control over logger buffering* and could be affecting performance if the logger is overused. Usage of the logger should be closely monitored on 'hot paths' in lambdas logging at any level other than `Debug` in production environment. In addition care must be taken by the metrics implementation to have some unique identifier on each log it needs to output so that metrics are not omitted by the [sampling behavior](https://github.com/uber-go/zap/blob/46939fcb20fb0e823f3e82ef197a9ee84be79bc3/zapcore/sampler.go#L142) of the default logger.

https://github.com/uber-go/zap/issues/637
https://github.com/uber-go/zap/issues/328

austinbyers:
We used to `defer logger.Sync()` at the end of every Lambda invocation - but we stopped once we realized it was always returning an error. I didn't dive deep into why, thanks for the info!

It would certainly be nice to have more efficient logging - for example, we always access the global logger which is a function call + mutex lock. It just hasn't been a high priority to address so far

PR #1130, opened 2020-07-07
Added CloudWatch Events parser

### LINKED ISSUE #921: Support CloudWatch Events log type
### Description

Support CloudWatch Events

### Acceptance Criteria

- Parsers are in place
- Unit tests for parsers
- Log type selectable in UI
- Glue tables created

kostaspap:
## Background

Closes #921 
## Changes

- Added CloudWatch Events parser

## Testing

- Unit tests
- Deployed to my account, verified end to end. 
Screen Shot 2020-07-07 at 3 16 48 PM

### DIFF
--- a/internal/log_analysis/log_processor/parsers/awslogs/cloudwatch_events.go
+++ b/internal/log_analysis/log_processor/parsers/awslogs/cloudwatch_events.go
@@ -70,7 +70,7 @@ func (p *CloudWatchEventParser) Parse(log string) ([]*parsers.PantherLog, error)
 
 // LogType returns the log type supported by this parser
 func (p *CloudWatchEventParser) LogType() string {
-	return TypeCloudwatchEvents
+	return TypeCloudWatchEvents
 }
 
 func (event *CloudWatchEvent) updatePantherFields(p *CloudWatchEventParser) {
--- a/internal/log_analysis/log_processor/parsers/awslogs/cloudwatch_events.go
+++ b/internal/log_analysis/log_processor/parsers/awslogs/cloudwatch_events.go
@@ -19,7 +19,6 @@ package awslogs
  */
 
 import (
-
 	jsoniter "github.com/json-iterator/go"
 	"github.com/pkg/errors"
 
--- a/internal/log_analysis/log_processor/parsers/awslogs/cloudwatch_events.go
+++ b/internal/log_analysis/log_processor/parsers/awslogs/cloudwatch_events.go
@@ -1,0 +1,84 @@
+package awslogs
+
+/**
+ * Panther is a Cloud-Native SIEM for the Modern Security Team.
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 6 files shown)


jacknagz:
Looks good! Is there a way we can elegantly handle CloudTrail data sent via CloudWatch Events? Otherwise, it will be deeply nested in the `details` key, which makes writing rules very difficult. This is a common workflow for most security teams who need low latency analysis for this data.

CC @rleighton

internal/log_analysis/log_processor/parsers/testutil/testutil.go
```diff
@@ -35,7 +35,8 @@ import (
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/parsers"
 )
 
-// used by log parsers to validate records
+// DEPRECATED - use `CheckPantherParser` instead
```
alxarch, line 38: It's not deprecated, it's used in `CheckPantherParser` and it's the only util we have to compare two pantherlogs

internal/log_analysis/log_processor/parsers/testutil/testutil.go
```diff
@@ -171,6 +172,7 @@ func CheckPantherMultiline(t *testing.T, logs string, parser parsers.LogParser,
 		require.NoError(t, err)
 		actual = append(actual, results...)
 	}
```
alxarch, line 174: The error check is not really needed as long as we read from a `string.Reader`.

kostaspap:
> Looks good! Is there a way we can elegantly handle CloudTrail data sent via CloudWatch Events? Otherwise, it will be deeply nested in the `details` key, which makes writing rules very difficult. This is a common workflow for most security teams who need low latency analysis for this data.
> 
> CC @rleighton

All CT events sent through CW have the same `detail-type` field value: `AWS API Call via CloudTrail`

This is documented here: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/CloudWatchEventsandEventPatterns.html

We actually make use of this field in Cloud Security: https://github.com/panther-labs/panther/blob/master/deployments/auxiliary/cloudformation/panther-cloudwatch-events.yml#L73-L74

If a security engineer wants to write a rule just for CT events delivered through CWE, they can implement logic like:
```
def rule(event):
   if event.get('detail-type') != 'AWS API Call via CloudTrail`
      return False
  ....
```

internal/log_analysis/log_processor/parsers/awslogs/cloudwatch_events.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (p *CloudWatchEventParser) New() parsers.LogParser {
+	return &CloudWatchEventParser{}
+}
+
+// Parse returns the parsed events or nil if parsing failed
+func (p *CloudWatchEventParser) Parse(log string) ([]*parsers.PantherLog, error) {
+	var event CloudWatchEvent
+	if err := jsoniter.UnmarshalFromString(log, &event); err != nil {
+		return nil, errors.New("failed to parse event")
```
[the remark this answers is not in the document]
kostaspap: You are right, missed. Updating.

internal/log_analysis/log_processor/parsers/testutil/testutil.go
kostaspap: Yes, the reason I put this note is because while writing the tests for the parser I was not sure which method would be more appropriate to use. I could structure my tests to use either `CheckPantherParser` or `EqualPantherLog`, but `EqualPantherLog` seemed more appropriate. 

I wanted to make clear to anyone writing tests in the future that they should be using this method, but I agree that writing `DEPRECATED` might be misleading. I updated the description, let me know what you think.

internal/log_analysis/log_processor/parsers/testutil/testutil.go
kostaspap: Yeah, just added it for completeness (similar to how we check in some tests that e.g. `jsoniter.Marshal(...)` doesn't return an error even though it should never return such).

internal/log_analysis/log_processor/parsers/testutil/testutil.go
alxarch: :+1:

PR #1133, opened 2020-07-07
Remove session storage entries upon form submission

### LINKED ISSUE #1053: Form session storage data persist after successfully submitting a form
### Describe the bug

We use session storage to save input data in several forms including Rules, Policies etc. 
But there is a minor bug when you actually complete the form, the input data remain on session storage so when you create a subsequent Rule/Policy, it has initial values as the last created form.

This was initial reported by @rleighton and we talked with @3nvi to reproduce this issue

### Steps to reproduce

1. Click on Create Rule
2. Input your desired data
3. Successfully create the rule
4. Click on Create Rule again
5. You will see your last input data prefilled

### Expected behavior

As a user, I expect form session storage to work as before but when submitting a form successfully I expect for this data to not be displayed on the subsequent creation.

### Environment

- Panther version or commit: bdddc3b

vorillaz:
## Background

This was a weird one, `session-id` was used from the `useFormSessionRestoration` hook as a temporary draft storage engine for different forms across the app. For persisted forms, everything worked as expected as a unique id was provided upon submission. For new items created through form submissions, the session storage cache was not properly invalidated.
Using the `useFormikContext` hook I was able to detect form submissions and remove any relevant entries, furthermore, a debounce method was added as minor performance improvement. Upon changing the form's stage the storage will get triggered again.

Close #1053 

## Changes

- Added proper sanitation and cleanup within the React lifecycle

## Testing

- Successfully retrieved a persisted rule upon editing.
- Successfully retrieved a new rule upon closing the tab.
- Successfully cleaned session storage upon new rule submission.

### DIFF
--- a/web/src/hooks/useFormSessionRestoration.tsx
+++ b/web/src/hooks/useFormSessionRestoration.tsx
@@ -17,6 +17,7 @@
  */
 import React from 'react';
 import { useFormikContext } from 'formik';
+import debounce from 'lodash-es/debounce';
 import storage from 'Helpers/storage';
 
 export interface UseFormSessionRestorationProps {
@@ -29,7 +30,7 @@ export interface UseFormSessionRestorationProps {
  * changes
  */
 function useFormSessionRestoration<FormValues>({ sessionId }: UseFormSessionRestorationProps) {
-  const { values, dirty, setValues } = useFormikContext<FormValues>();
+  const { values, dirty, setValues, isSubmitting } = useFormikContext<FormValues>();
 
   React.useEffect(() => {
     const previousSessionValues = storage.session.read<FormValues>(sessionId);
@@ -38,13 +39,22 @@ function useFormSessionRestoration<FormValues>({ sessionId }: UseFormSessionRest
     }
   }, [sessionId, setValues]);
 
+  const syncStorage = React.useCallback(
+    debounce(() => {
+      if (dirty && !isSubmitting) {
+        storage.session.write(sessionId, values);
+      }
+    }, 200),
+    [dirty, isSubmitting, values]
+  );
+
+  React.useEffect(syncStorage, [dirty, isSubmitting, values]);
+
   React.useEffect(() => {
-    if (dirty) {
-      storage.session.write(sessionId, values);
-    } else {
+    if (isSubmitting) {
       storage.session.delete(sessionId);
     }
-  }, [values, dirty]);
+  }, [isSubmitting]);
 }
 
 export default useFormSessionRestoration;


web/src/hooks/useFormSessionRestoration.tsx
```diff
@@ -38,13 +39,22 @@ function useFormSessionRestoration<FormValues>({ sessionId }: UseFormSessionRest
     }
   }, [sessionId, setValues]);
 
+  const syncStorage = React.useCallback(
+    debounce(() => {
```
3nvi, line 43: Although  not bad to have, `storage.session.write` take `0.3ms` on average. We can keep the debouncing, but we'd ideally want a lower debounce time (around 200-300ms), since writing to the storage is a cheap operation

web/src/hooks/useFormSessionRestoration.tsx
```diff
@@ -38,13 +39,22 @@ function useFormSessionRestoration<FormValues>({ sessionId }: UseFormSessionRest
...
+    [values, dirty]
+  );
+
+  React.useEffect(syncStorage, [values, dirty]);
+
   React.useEffect(() => {
-    if (dirty) {
-      storage.session.write(sessionId, values);
-    } else {
+    if (isSubmitting) {
```
3nvi, line 54: I like it. Have we validated that the hook works well with form submission errors? Take the following scenario:

- User presses "Create Single Policy"
- User starts filling in the form
- User submits form
- Form can't be "saved" since API returned errors
- User  leaves the page (or refreshes)

Will he see the same state he had right before the failed submission or will he start from the start?

P.S.  To mock that,  try adding  an `ID` value that already exists (i.e. `AWS.RDS.Instance.SnapshotPublicAccess`)

vorillaz: @3nvi Yeap you are right on this one, values are not properly restored upon the API throwing an error. 
This a bit tricky though as there are no errors returned from the `useFormikContext` hook when there is something wrong with the API, it looks like the `useFormik` returns just the front end validation errors of the form.

```tsx
const { values, dirty, setValues, isSubmitting, errors } = useFormikContext<FormValues>();
console.log(errors); // => {}
```
[This one](https://github.com/panther-labs/panther/blob/33ca44cd6220cd028ff465242fcee9701ae1196c/web/src/pages/CreateRule/CreateRule.tsx#L53) is responsible for matching the appropriate errors from the API.

Any ideas on this one?

vorillaz: Nevermind, I found out a solution combining the `dirty` and `isSubmitting` along with values and everything works as expected. For reference reasons, dirty gets triggered even when the API is throwing an error :)

PR #1139, opened 2020-07-08
Add an illustrated blank slate when no alerts are in place

### LINKED ISSUE #931: Show Image When No Alerts Present on Rule Overview

After creating a rule, I'm taken to the rule overview page, where no alerts are present. It would be nice to show a little image similar to the default list alerts page when there are no results.

### Describe your use cases

UX improvement

Upon new creation of a rule

None

vorillaz:
## Background
When no alerts are yet generated we shall show an illustrated blank slate instead of an empty data table. 

Closes #931 .

## Before
Screenshot 2020-07-08 at 6 36 26 PM

## After
Screenshot 2020-07-08 at 6 36 39 PM

## Changes

- Check for generated alerts and add the appropriate blank slate.

## Testing

- Manually and visually.


web/src/pages/RuleDetails/RuleDetails.tsx
```diff
@@ -124,8 +128,11 @@ const RuleDetailsPage = () => {
       <Box mt={5} mb={6}>
         <Panel title="Alerts">
           <ErrorBoundary>
-            <RuleDetailsAlertsTable alerts={listAlertsData.alerts.alertSummaries} />
-            {hasMoreAlerts && (
+            {hasAnyAlerts && (
```
3nvi, line 131: optional: we can use a ternary for `hasAlerts ? ... : ...` instead of two boolean expressions. This might make it more understandable by a person not related to FE.

That's 100% optional

PR #1143, opened 2020-07-09
Metrics API

nhakmiller:
## Background

This new API will be used by the frontend to request single data points and time series data to be displayed in charts and graphs on the various overview pages. I am particularly interested in feedback about how we would like to request this data and how we would like to serve this data. With that in mind I tagged everyone that I believe has had any input on the metrics so far.

After some initial discussion with @alexmylonas , we agreed on a few points. This is what we agreed on:

- The frontend would only ever need to hit one endpoint to get any metric data it wanted
- Metric data would be in the form of either a single value OR time series data
- The frontend would send the following pieces of information for reach request:
  - The start and end date of the period to be considered
  - The length of periods to be considered, for time series data
  - The names of the metrics to be generated

With these in mind I've sketched out a basic API.

You can request metrics by sending queries of the following format to the `panther-metrics-api`:

```
{
  "getMetrics": {
    "metricNames": [
      "eventsProcessed"
    ],
    "fromDate": "2020-07-09T12:00:00Z",
    "toDate": "2020-07-09T23:00:00Z",
    "intervalHours": 1
  }
}
```

The `fromDate` is the start of the period to be considered. The `toDate` is the end of the period to be considered. The `intervalHours` is the length in hours of each interval in the period to be considered (so in the above example, the given timeframe will be broken into 1 hour intervals because of the value 1 for `intervalHours`). The `metricNames` parameter is a list of metrics you are requesting data for. All metrics will be evaluated with the same to/from date and period. This will allow the frontend to request all of the metrics on say the new log analysis overview page with just one API call.

The response is in the following format:

```
{
  "metricNames": [
    {
      "MetricName": "EventsProcessed",
      "seriesData": [
        {
          "Label": "AWS.ALB",
          "Timestamps": ["2020-07-09T22:00:00Z", "2020-07-09T21:00:00Z", "2020-07-09T20:00:00Z"],
          "Values": [6, 2, 2]
        },
        {
          "Label": "AWS.VPCFlow",
          "Timestamps": ["2020-07-09T22:00:00Z", "2020-07-09T21:00:00Z", "2020-07-09T20:00:00Z"],
          "Values": [1485, 1583, 1534]
        }]}
  ],
  "fromDate": "2020-07-09T12:00:00Z",
  "toDate": "2020-07-09T23:00:00Z",
  "intervalHours": 1
}
```

First observe that I send back the `fromDate`, `toDate`, and `intervalHours` requested. I just do this for convenience, I could easily drop it if preferred. But I figured at some point we may be saving/caching/sharing these results somehow so it might be useful.

The other top level key in the response is the `metricNames` key. This is a list of objects, where each object in the list corresponds to one of the metric names that was requested in the `GetMetrics` request. This is indicated by the `MetricName` value in the object. Each object also contains a list of JSON objects that contains a `label` and either a `singleValue` integer value, or a `seriesData` which contains time series data.

Time series data is represented with two pieces of data, a list of timestamps in `Timestamps` and a list of values in `Values`. The i'th value is the metric value at the i'th timestamp.

## Changes

- Refactored the `MonoLogger` to the `StaticLogger`.
  - Allowed the `StaticLogger` to accept multiple metrics (must still be declared at logger instantiation)
  - Allowed the `StaticLogger` to accept multiple dimension sets (must still be declared at logger instantiation)
  - Added some quick input sanity checking to the `StaticLogger` while logging. The primary benefit of not rebuilding the metrics directive object on each call to log still remains
- Added the `panther-metrics-api` lambda function
  - Added support for requesting the `EventsProcessed` metric
- Updated the `panther-log-processor` to emit the `EventsProcessed` metric in addition to the `BytesProcessed` metric

## Testing

- Unit testing for `StaticLogger` and `panther-log-processor` changes
- Manual testing for the `panther-metrics-api` via the AWS Lambda console
  - Not sure how to add integration tests here without just mocking all the CloudWatch API calls (which might be the best bet), but then that feels the same as unit testing for this lambda? Either way I'll at least be adding some unit testing before merging but wanted to get this PR to the team for feedback on the API portion


api/lambda/metrics/models/api.go
```diff
@@ -0,0 +1,55 @@
...
+type GetMetricsOutput struct {
+	MetricResults map[string]*MetricResult `json:"metricNames" validate:"required"`
+	FromDate      time.Time                `json:"fromDate"`
+	ToDate        time.Time                `json:"toDate"`
+	Period        int64                    `json:"period"`
+}
+
+type MetricResult = struct {
+	SingleValue *int64                         `json:"singleValue,omitempty"`
+	SeriesData  map[string]*TimeSeriesResponse `json:"seriesData"`
```
nhakmiller, line 49: Instead of mapping a string to a `TimeSeriesResponse` we could just add a `Name` field to the `TimeSeriesResponse` struct and make this a slice. No strong preference, I like maps because you can quickly access one data point if that's all you're looking for but presumably in most cases you're only going to request the one data point you need if you only need one.

api/lambda/metrics/models/validator.go
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+// Validator builds a custom struct validator.
+func Validator() (*validator.Validate, error) {
+	result := validator.New()
+	return result, nil
+}
```
nhakmiller, line 27: Couldn't actually tell if I need a custom validator here because there are no custom fields to validate, but all the other APIs I looked at had one so I left this in.

deployments/core.yml
```diff
@@ -105,6 +105,9 @@ Mappings:
     LayerManager:
       Memory: 512
       Timeout: 60
+    MetricsAPI:
+      Memory: 256
+      Timeout: 180
```
nhakmiller, line 110: Slightly bigger memory than most functions because it's only tied to user actions and we want it to be fast. Additionally, users can request up to 100,000 data points (each of which has a timestamp and a value) so we don't want to run out of memory.

internal/core/metrics_api/api/get_metrics.go
```diff
@@ -0,0 +1,130 @@
...
+const (
+	maxSeriesDataPoints   = 100000
+	eventsProcessedMetric = "EventsProcessed"
+)
+
+var (
+	metricsInternalError = &genericapi.InternalError{Message: "Failed to generate requested metrics. Please try again later"}
+	metricResolvers      = map[string]func(input *models.GetMetricsInput) (*models.MetricResult, error){
+		"eventsProcessed": getEventsProcessed,
+	}
```
nhakmiller, line 42: This is basically adding a second, weakly typed, more error prone router. But @alexmylonas said having a single API endpoint to hit would be much better than having to hit 7 different endpoints to build one page, and this gives us the flexibility to do that (and also change what graphs are on a page without backend changes).

api/lambda/metrics/models/api.go
```diff
@@ -0,0 +1,55 @@
...
+//
+// GetMetricsInput: Used by the UI to request a series of data points
+//
+
+// GetMetricsInput is used to request data points for a number of metrics over a given time frame and periods
+type GetMetricsInput struct {
+	MetricNames []string  `json:"metricNames" validate:"required"`
+	FromDate    time.Time `json:"fromDate" validate:"required"`
+	ToDate      time.Time `json:"toDate" validate:"required"`
+	Period      int64     `json:"period" validate:"required"`
```
austinbyers, line 37: You may also want to enforce a positive number here, `validate:"required,gt=0"`

api/lambda/metrics/models/api.go
```diff
@@ -0,0 +1,55 @@
...
+
+type MetricResult = struct {
+	SingleValue *int64                         `json:"singleValue,omitempty"`
+	SeriesData  map[string]*TimeSeriesResponse `json:"seriesData"`
+}
+
+type TimeSeriesResponse struct {
+	Timestamps []*time.Time
+	Values     []*float64
+}
```
austinbyers, line 55: Per @alxarch and @kostaspap suggestion, we've started dropping pointers from API slices - these should be `[]time.Time` and `[]float64`

austinbyers: On a related note, I'd be curious how they feel about using pointers for your maps above as well. I'd be fine removing them

api/lambda/metrics/models/api.go
austinbyers: As far as I know, we *always* send a list rather than a map because it's much easier to define the schema in GraphQL / AppSync. This was so important, it actually made it into our (now-outdated) [style guide](https://panther-labs.atlassian.net/wiki/spaces/EN/pages/1867809/Go+Style+Guide#Models)

The frontend team has the final say here, but unless they have given you a specific exception, **avoid maps with dynamic keys**

api/lambda/metrics/models/validator.go
austinbyers: No, you can (and should) remove this. The `genericapi` router will automatically build a standard validator for you if you don't specify one

deployments/core.yml
austinbyers: Probably worth a comment in the source, then :)

internal/core/metrics_api/api/get_metrics.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/aws/aws-sdk-go/service/cloudwatch"
+	"go.uber.org/zap"
+
+	"github.com/panther-labs/panther/api/lambda/metrics/models"
+	"github.com/panther-labs/panther/pkg/genericapi"
+	"github.com/panther-labs/panther/pkg/metrics"
+)
+
+const (
+	maxSeriesDataPoints   = 100000
```
austinbyers, line 34: Is this an AWS limitation? If so, please notate it as such

api/lambda/metrics/models/api.go
```diff
@@ -0,0 +1,55 @@
...
+type GetMetricsInput struct {
+	MetricNames []string  `json:"metricNames" validate:"required"`
+	FromDate    time.Time `json:"fromDate" validate:"required"`
+	ToDate      time.Time `json:"toDate" validate:"required"`
+	Period      int64     `json:"period" validate:"required"`
+}
+
+type GetMetricsOutput struct {
+	MetricResults map[string]*MetricResult `json:"metricNames" validate:"required"`
+	FromDate      time.Time                `json:"fromDate"`
```
austinbyers, line 42: You can use [cross-field validation](https://godoc.org/gopkg.in/go-playground/validator.v9#hdr-Cross_Field_Validation) to ensure `toDate` > `fromDate`

internal/core/metrics_api/api/get_metrics.go
```diff
@@ -0,0 +1,133 @@
...
+	duration := input.ToDate.Sub(input.FromDate)
+	samples := int64(duration.Hours()) / input.Period
+	if samples*int64(len(listMetricsResponse.Metrics)) > maxSeriesDataPoints {
+		return nil, &genericapi.InvalidInputError{Message: "too many data points requested please narrow query scope"}
+	}
+
+	// Build the query based on the applicable metric dimensions
+	//
+	// This will fail if there are more than 100 log types. If we get to the point where we expect
+	// users will have over 100 log types, this will have to be batched then merged.
```
austinbyers, line 94: The SDK provides [ListMetricsPages](https://docs.aws.amazon.com/sdk-for-go/api/service/cloudwatch/#CloudWatch.ListMetricsPages) - we almost always use the paginated version from the beginning, just to save us any trouble down the line. I'd recommend that we do the same here - 100 log types is not at all unreasonable.

```go
input := &cloudwatch.ListMetricsInput{...}
err := cloudwatchClient.ListMetricsPages(input, func (page *cloudwatch.ListMetricsOutput, isLast bool) bool {
   for _, metric := range page.Metrics {
        // ... add metric to result set ...

        if len(result.metrics) > maxSeriesDataPoints {
            return false // stop paging - too many entries 
        }
    }
    return true // keep paging over metrics
})

if len(result.metrics) > maxSeriesDataPoints {
    // return "too many data points" error
    
   // note that you can't return the error from the paginator above, which is a function closure.
}
```

Happy to help with the paginator if needed

internal/core/metrics_api/main/lambda.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+var router *genericapi.Router
+
+func init() {
+	validator, err := models.Validator()
+	if err != nil {
+		panic(err)
+	}
+	router = genericapi.NewRouter("core", "metrics_api", validator, api.API{})
```
austinbyers, line 39: You can pass `nil` for the validator

internal/core/metrics_api/api/get_metrics.go
nhakmiller: I'm familiar with the paginated calls, we use them in the snapshot poller everywhere. I had actually meant to come back and change this to a paginated call once it was working, just forgot to do so so thanks for the reminder!

The 100 log types thing is actually a separate issue, you can only have up to 100 `MetricDataQuery`s per `GetMetricData` API call, regardless of if you use the builtin pagination or paginate yourself. So if we want to be protected against 100+ different log types causing an issue here, we have to make (# of log types)/100 `GetMetricData` API calls. I can add the logic for that, I was just skipping it since I think we're only at like 20 or 30 log types right now so I didn't think it would be an issue for a while.

internal/core/metrics_api/main/lambda.go
nhakmiller: Perfect, thanks.

api/lambda/metrics/models/api.go
nhakmiller: Ok, I wasn't aware of that limitation. With that in mind I will probably need to change several things.

api/lambda/metrics/models/api.go
nhakmiller: I was using pointers here because I'm directly ripping these fields out of the API response, which is all slices of pointers. Making them not pointers means I need to iterate through each of the slices that AWS returns and dereference every element into a new slice.

My understanding of the intention behind removing slices of pointers is that it's to make it more straightforward to use (without needing to do aws.String() and aws.StringValue() all the time), do you think changing these to `[]time.Time` and `[]float64` still accomplishes that goal? Happy to make the change if you think so, I just want to make sure we're following standards because they make the code better not just because they're the standard.

internal/core/metrics_api/api/get_metrics.go
nhakmiller: The AWS limitation is actually 108,000 I think, I just rounded to 100,000. I'll update with a comment and the exact value.

austinbyers:
> Not sure how to add integration tests here without just mocking all the CloudWatch API calls (which might be the best bet), but then that feels the same as unit testing for this lambda?

Right, unit tests would require mocking, and generally aren't too helpful for API handlers like this one (in my opinion). All you're doing is taking input, invoking AWS, and then returning a slightly modified result.

An integration test means querying the real service, no mocking. We should absolutely add an integration test, though it doesn't have to be in this PR - do so once the API model is defined. Here's how that could work:

* Call cloudwatch `PutMetricData` directly with varied timestamps
* Then run some basic queries (using your Lambda API) to make sure the data is returned

The challenge with this is the delay in the metrics service. From [the docs](https://docs.aws.amazon.com/sdk-for-go/api/service/cloudwatch/#CloudWatch.PutMetricData):

> Data points with time stamps from 24 hours ago or longer can take at least 48 hours to become available for GetMetricData or GetMetricStatistics from the time they are submitted. Data points with time stamps between 3 and 24 hours ago can take as much as 2 hours to become available for for GetMetricData or GetMetricStatistics.

So an integration test should be able to seed metrics between now and 2 hours ago, which is a wide enough window to test the range abilities of your query. It may take some trial and error, but we can certainly add *something* for an integration test. Even if it's literally just put one metric value, wait a minute, and then read it back - that already helps immensely, because you've validated the permissions are correct, the AWS API call is correct, and the Lambda function is responding with the structure you expect.

api/lambda/metrics/models/api.go
austinbyers: > I was using pointers here because I'm directly ripping these fields out of the API response, which is all slices of pointers. Making them not pointers means I need to iterate through each of the slices that AWS returns and dereference every element into a new slice.

A great point - this is why I chose to make everything pointers in the first place - it makes it much easier to copy responses from the AWS SDK.

I would agree, then, that having to dereference the entire response and build new slices is not at all worth it. Perhaps we just let each service decide what standard makes the most sense for them

api/lambda/metrics/models/api.go
nhakmiller: Ok I got rid of both maps, and replaced them with lists of objects. I took the key and inserted that as a field in the corresponding object.

api/lambda/metrics/models/api.go
nhakmiller: I did go back through and un-pointerfy any field that wasn't coming directly from AWS, but I left the slices as is.

internal/core/metrics_api/api/get_metrics.go
nhakmiller: Limit is 100,800, I updated with a comment.

internal/core/metrics_api/api/get_metrics.go
nhakmiller: I updated `ListMetrics` and `GetMetricData` to both use the SDK's builtin pagination functions. Additionally, I moved the data point validation and batching for `GetMetricData` to a helper function so that it could be re-used by future metric handlers.

api/lambda/metrics/models/api.go
```diff
@@ -0,0 +1,66 @@
...
+//
+// GetMetricsInput: Used by the UI to request a series of data points
+//
+
+// GetMetricsInput is used to request data points for a number of metrics over a given time frame
+type GetMetricsInput struct {
+	MetricNames   []string  `json:"metricNames" validate:"required"`
+	FromDate      time.Time `json:"fromDate" validate:"required"`
+	ToDate        time.Time `json:"toDate" validate:"required,gtfield=FromDate"`
+	IntervalHours int64     `json:"intervalHours" validate:"required,gt=0"`
```
austinbyers, line 37: Well there never be smaller intervals? I could imagine use cases where it would be helpful to see metrics within an hour - for example, right after onboarding data, it'd be nice to the graph tick up after a few minutes

api/lambda/metrics/models/api.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// GetMetricsOutput contains data points for a number of metrics over the specified time frame
+type GetMetricsOutput struct {
+	MetricResults []MetricResult `json:"metricNames" validate:"required"`
```
austinbyers, line 42: The json should be `metricResults` to match the field name. Also, since this is an output struct, you can omit the `validate` tags (we don't run them)

internal/core/metrics_api/api/get_metrics.go
```diff
@@ -0,0 +1,164 @@
...
+// getEventsProcessed returns the count of events processed by the log processor per log type
+//
+// This is a time series metric.
+func getEventsProcessed(input *models.GetMetricsInput) (*models.MetricResult, error) {
+	// First determine applicable metric dimensions
+	var listMetricsResponse []*cloudwatch.Metric
+	err := cloudwatchClient.ListMetricsPages(&cloudwatch.ListMetricsInput{
+		MetricName: aws.String(eventsProcessedMetric),
+		Namespace:  aws.String(metrics.Namespace),
+	}, func(page *cloudwatch.ListMetricsOutput, keepPaging bool) bool {
```
austinbyers, line 80: Nit: the second parameter to this function should be named `isLast` or `_` - it's how AWS tells you this is the final page. We've never needed it, but let's avoid confusion.

The *return value* determines whether to keep paging

internal/core/metrics_api/api/get_metrics.go
```diff
@@ -0,0 +1,164 @@
...
+	for start := 0; start < queryCount; start += maxMetricsPerRequest {
+		end := start + maxMetricsPerRequest
+		if end > queryCount {
+			end = queryCount
+		}
+		err := cloudwatchClient.GetMetricDataPages(&cloudwatch.GetMetricDataInput{
+			EndTime:           aws.Time(input.ToDate),
+			MaxDatapoints:     aws.Int64(maxSeriesDataPoints),
+			MetricDataQueries: queries[start:end],
+			StartTime:         aws.Time(input.FromDate),
```
austinbyers, line 153: Most of these fields are the same on every request, so you can build the input outside the loop. You can also use `&` to get pointers to skip some function calls

```go
getInput := &cloudwatch.GetMetricDataInput{
    EndTime: &input.ToDate,
    MaxDatapoints: aws.Int64(maxSeriesDataPoints),  // can't take the address of a constant
    StartTime: &input.FromDate,
}
for start := 0;  ... {
    // build start, end
    getInput.MetricDataQueries = queries[start:end]
}
```

pkg/metrics/logger.go
```diff
@@ -93,7 +93,7 @@ type Metric struct {
 	// This value is not marshalled to JSON as it is not part of the AWS embedded metric format. We
 	// simply include it here for convenience when calling the loggers defined here, so that it is
 	// not required to consider the value of a metric separate from its Name and Unit.
-	Value *interface{} `json:"-"`
+	Value interface{} `json:"-"`
```
austinbyers, line 96: 👍  very rarely would you want a `*interface{}`

api/lambda/metrics/models/api.go
nhakmiller: Good catch, lazy copy/paste on my behalf.

Issue #975, opened 2020-06-03
Enhachements for Cloud Security dashboards


kostaspap:
### Description

The dashboard has currently the following diagrams: 

![Screen Shot 2020-06-03 at 10.17.12 AM.png](https://images.zenhubusercontent.com/5e4fc81c2009ef65e12fafd8/0131f090-32bd-41ba-bdd0-f796609600b3)

Some notes: 
* Top left chart is named "Policy Failure" but also includes passing policies. I am thinking we should just remove the passing policies and just display the failing policies. Maybe rename to "Failing Policies"
* "Policy Severity" maybe rename to "Enabled Policies"? It took me a bit to understand that this was not displaying the failing policies but just the policies that were enabled (failing or not)
* In the existing bottom left diagram, the "Failing, Passing" in the legend is in reverse order than it is displayed in the diagram. Should show first "Passing" then "Failing" in the legend. 
* I think it makes sense to remove the existing diagram that just displays "AWS"  resources and just have a "Policy Health" diagram which shows two bars, passing and failing policies. 

I was thinking maybe the charts would be organized like this:

"Policy Health" | "Failing Policies"
"Resource Health" | "Enabled Policies"

### Acceptance Criteria

- Above changes are implemented

rleighton:
+1

jacknagz:
Also - the totals don't seem to be dynamically updating (as they did in the past) if you were to de-select a category:

![cloudsec-new-overview](https://user-images.githubusercontent.com/11466941/83699890-1d29cc00-a5ba-11ea-9095-6adf2aeb681a.gif)

PR #1145, opened 2020-07-10
Cloud security dashboard enhachements

vorillaz:
## Background

This is a WIP pull request with the changes suggested in #975. Eventually, the order, the titles and the data displayed have changed.

## TBD
As @kostaspap suggested I added the "Policy health" graph with passing and failing policies and also added the total policy number, unfortunately, the Total policies are also projected within the "Enabled policies" box. Any suggestions on this one would be highly appreciated.

## Before 
Screenshot 2020-07-15 at 11 11 03 AM

## After
Screenshot 2020-07-15 at 11 18 13 AM

## Changes

- Change the order of the boxes
- Fix the legend order.
- Introduce a new block with 
- Remove any stale chart block.
- Updated the skeleton component

## Testing

- Manually and visually


3nvi:
I'll be waiting for a validation from @kostaspap that the behavior is indeed what we want, prior to any reviewing

kostaspap:
Change looks pretty awesome, thanks for doing this!
One issue I found is that the "Policy Health" box seems to be reporting wrong values for failing policies. This is also viewable in the Screenshot you attached in this PR. You can see that "Policy Health" reports 16 failing policies, but the "Failing policies" box reports 15 failing policies overall. 

Apart from that, LGTM

vorillaz:
Nice catch @kostaspap , mea culpa.
I fixed the inconsistent data and also updated the screenshots.
The skeleton component was updated as well :)

PR #1147, opened 2020-07-10
Remove build:lambda and build:tools from CI

austinbyers:
## Background

This may be a controversial proposal, but I argue we don't need `build:lambda` and `build:tools` in CI. They add very little value, certainly not enough to justify such a huge proportion of the CI runtime.

The go linter and unit tests will already fail if the Go code can't compile (even if there aren't associated unit tests). So the only possible reason `build:lambda` or `build:tools` could fail on their own is if we are invoking the `go` binary with the wrong arguments. Such an issue would immediately be caught by every developer when they try to deploy.

Ran some timings on my laptop: `go clean -cache && time mage test:ci`

* `build:lambda` and `build:tools` are among the first parallel tasks to start and are the very last to finish
* `mage test:ci` takes 193s today, but only 119s without these two checks! A 38% reduction

In other words, we're adding more than 70 seconds of wall clock time to test ~10 lines of mage code, which is just invoking `go`. Not worth it

## Changes

- Remove `build:lambda` and `build:tools` from CI

## Testing

- `mage test:ci`


austinbyers:
> The scenario we'd like to avoid is that build:lambda or build:tools fail while test:ci succeeds. This would be problematic as it would e.g. allow bad commits to make it to master.

Again, that's theoretically possible, but the same is true for all of the mage commands, e.g. `mage deploy`. We don't have any CI testing for `mage deploy` except just to verify the Go code compiles. But any issues with build/deploy would very quickly be caught by every developer

kostaspap:
> > The scenario we'd like to avoid is that build:lambda or build:tools fail while test:ci succeeds. This would be problematic as it would e.g. allow bad commits to make it to master.
> 
> Yes, that's possible, but the same is true for all of the mage commands, e.g. `mage deploy`. We don't have any CI testing for `mage deploy` except just to verify the Go code compiles. But any issues with build/deploy would very quickly be caught by every developer

So your point is that the only benefit of running `mage build:lambda` and `mage build:tools` as part of `mage test:ci` is validating that the corresponding mage code is correct - there are other checks that we perform as part of `mage test:ci`  that verify that the "lambda" and "tool" Go code  is compiling. Correct? 
If that's the case, I agree with the changes in the PR. Doesn't seem to worth running them as part of `mage test:ci`

austinbyers:
> So your point is that the only benefit of running mage build:lambda and mage build:tools as part of mage test:ci is validating that the corresponding mage code is correct - there are other checks that we perform as part of mage test:ci that verify that the "lambda" and "tool" Go code is compiling. Correct?

Yes, exactly. The Go code is already validated by `go test` and the go linter. So they only possible issue would be with the ~10 lines of mage code that invoke Go with the correct flags. A problem here would be immediately identified by developers and isn't worth the enormous CI overhead - that is my argument

austinbyers:
Seeing no other objections after a few days, going to go ahead and merge

PR #1148, opened 2020-07-12
Setup FE integration testing

3nvi:
## Background

This PR adds a lot of testing helpers & sets up CI in such a way, so that all FE tests run as part of it.

Specifically, this PR adds a helpful `npm run test` command that  will run all FE tests in the app, while creating modules that allow:

- rendering a component for testing
- generating fake data
- selecting random items & generating random arrays
- querying for expectations using a lot of different methods

## Changes

This PR creates:

- a helpful React component renderer that accepts
  -  whether the user should be considered authenticated
  - the URL that the app's router should be under
  - a list of API requests along with their mocked responses (in a simple list)
- helpers for:
  - selecting random items from an array of items
  - creating an array of random length, using an item-creating function
- queries for selecting items by text, label, placeholder, aria-attributes, etc.
- a function for generating fake data
- enhanced expectation matchers for assertions
- a simple test file for demo purposes

## Testing

- Introduced a base example testfile for validating PR work. This is run by the CI and the checks pass :)

## Additional Notes

Planning to extend this PR with a complimentary second PR which adds utility functions for generating fake data for our GraphQL queries. This PR will also add type safety on our  mock API request implementation.


api/graphql/schema.graphql
```diff
@@ -506,7 +506,7 @@ input DestinationInput {
 input DestinationConfigInput {
   slack: SlackConfigInput
   sns: SnsConfigInput
-  sqs: SQSConfigInput
```
3nvi, line 509: fixed typo on casing which caused some codegen plugins to fail

package.json
```diff
@@ -38,7 +38,10 @@
     "@graphql-codegen/typescript-resolvers": "^1.13.1",
     "@pmmmwh/react-refresh-webpack-plugin": "^0.3.3",
     "@sentry/cli": "^1.49.0",
+    "@testing-library/jest-dom": "^5.11.0",
```
3nvi, line 41: `@testing-library/jest-dom` -> extra jest assertions for DOM related stuff
`@testing-library/react` -> our test rendering library

package.json
```diff
@@ -60,15 +63,21 @@
     "eslint": "^7.2.0",
     "eslint-config-airbnb-base": "^14.2.0",
     "eslint-config-prettier": "^6.11.0",
+    "eslint-import-resolver-alias": "^1.1.2",
```
3nvi, line 66: lets ESLint know of additional import aliases for tests

package.json
```diff
@@ -60,15 +63,21 @@
...
     "eslint-import-resolver-webpack": "^0.12.1",
     "eslint-plugin-import": "^2.21.2",
     "eslint-plugin-prettier": "^3.1.4",
     "eslint-plugin-react": "^7.20.0",
+    "faker": "^4.1.0",
```
3nvi, line 71: fake data generator

package.json
```diff
@@ -60,15 +63,21 @@
...
     "file-loader": "^5.0.2",
     "fork-ts-checker-webpack-plugin": "^4.0.3",
     "html-loader": "^1.1.0",
     "html-webpack-plugin": "^4.0.0-beta.5",
     "husky": "^3.1.0",
+    "jest": "^26.1.0",
```
3nvi, line 77: `jest` -> test runner
`jest-extended` -> additional (non-DOM related) assertions to make life even easier
`jest-localstorage-mock` -> mocks localstorage & sessionstorage in JS Dom
`jest-watch-typeahead` -> allows dynamic filtering (directly  on the CLI) of which tests do you want to run

tools/mage/test_namespace.go
```diff
@@ -78,6 +78,7 @@ func (Test) CI() {
 		{"mypy (python type checking)", testPythonMypy},
 		{"npm run eslint", testWebEslint},
 		{"npm run tsc", testWebTsc},
+		{"npm run test", testWebIntegration},
```
3nvi, line 81: adds `npm run test` to CI

web/.eslintrc.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
     'plugin:prettier/recommended',
   ],
   env: {
+    jest: true,
```
3nvi, line 30: lets ESLint be aware of the global vars that `jest` has

web/.eslintrc.js
```diff
@@ -70,6 +71,10 @@ module.exports = {
       webpack: {
         config: path.resolve(__dirname, 'webpack.config.js'),
       },
+      alias: {
+        map: [['test-utils', path.resolve(__dirname, '__tests__/utils')]],
```
3nvi, line 75: tells ESlint that `import { Whatever } from 'test-utils'` is valid can be resolved through the `__tests__/utils/index.ts` file

web/__tests__/__mocks__/file.ts
```diff
@@ -0,0 +1,19 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+export default 'test-file-stub';
```
3nvi, line 19: A mock for all statics (images, fonts, SVGs).

In general, instead of returning the `base64` code of an image, just return a hardcoded string for testing purposes

web/__tests__/__mocks__/react-spring.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import type { UseTransitionProps, TransitionKeyProps, UseTransitionResult } from 'react-spring';
+
+const realModule = jest.requireActual('react-spring');
+
+module.exports = {
```
3nvi, line 23: mocks `react-spring` package by "pruning" intermediate transitions, makings things "instant" on testing environments

web/__tests__/example.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
```
3nvi, line 19: example test for sanity check in the PR. Helps:

- evaluates if `test-utils` works
- tests renderer
- tests CI integration

web/__tests__/setup.ts
```diff
@@ -0,0 +1,69 @@
...
+
+// additional matchers for jest. Adds the ability to instantly check for `null` or to check
+// whether a mock has been called before another mock
+// https://github.com/jest-community/jest-extended#api
+import 'jest-extended';
+
+window.alert = () => {};
+window.scrollTo = () => {};
+
+if (window.matchMedia === undefined) {
```
3nvi, line 32: mock additional stuff (other than localstorage & sessionstorage)

web/__tests__/utils/auth.ts
```diff
@@ -0,0 +1,65 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import faker from 'faker';
```
3nvi, line 19: mocks the data returned by the `useAuth()` hook, since we don't want to deal with the authentication workflow here (which will be delegated to E2E testing)

web/src/components/utils/UIProviders/UIProviders.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import { SnackbarProvider, ThemeProvider } from 'pouncejs';
+import { SidesheetManager, SidesheetProvider } from 'Components/utils/Sidesheet';
+import { ModalManager, ModalProvider } from 'Components/utils/Modal';
+
+// Helper that allows us to guarantee same core providers in production & testing environments
+const UIProviders: React.FC = ({ children }) => (
```
3nvi, line 25: That was needed so we reduce manual errors, when trying to keep the "test" providers up-to-date with the ones used in dev/prod

web/codegen.yml
```diff
@@ -55,8 +55,8 @@ generates:
     presetConfig:
       baseTypesPath: ../__generated__/schema.tsx
     plugins:
-      - typescript-operations:
```
3nvi, line 58: fixes simple typo. No changes here.

web/__tests__/utils/queries.ts
```diff
@@ -0,0 +1,46 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import { queryHelpers, buildQueries, Matcher, MatcherOptions } from '@testing-library/react';
```
3nvi, line 19: create additional queries for  `react-testing-library`.

Specifically, we create `getByAriaLabel` to allow us to select icon buttons that don't have text. @vorillaz you might like this

web/__tests__/utils/index.ts
```diff
@@ -0,0 +1,21 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+export * from './render';
```
3nvi, line 19: The final thing exposed by `test-utils`

web/src/components/utils/UIProviders/UIProviders.tsx
3nvi: I couldn't place it in  the same file as `app.tsx`, since importing the `app` file within our `__tests__/utils/render.tsx` file caused issues (specifically, `app.tsx` also imported apollo http links, so it creates issues with the fact that `fetch`  is  not implemented in JS Dom)

3nvi: I'm not too happy with this, but I'm not unhappy either. If you guys have any suggestions @alexmylonas @vorillaz, I'm all ears

tools/mage/test_namespace.go
vorillaz: Just a quick note here.
We need to ensure that the CI runs something like [npm ci](https://docs.npmjs.com/cli/ci.html) instead of a full-blown install of the `node_modules`

web/__tests__/setup.ts
```diff
@@ -0,0 +1,69 @@
...
+if (window.URL.revokeObjectURL === undefined) {
+  Object.defineProperty(window.URL, 'revokeObjectURL', { value: noOp });
+}
+
+// Mock the server-side EJS-injected AWS configuration.
+// See web/public/index.ejs
+const scriptTag = document.createElement('script');
+scriptTag.id = '__PANTHER_CONFIG__';
+scriptTag.type = 'application/json';
+scriptTag.innerHTML = JSON.stringify({
```
vorillaz, line 61: That's duplication for internal configuration settings, is there any way we can dynamically include this kind of information from an internal configuration file?

web/__tests__/utils/auth.ts
```diff
@@ -0,0 +1,65 @@
...
+ * or a logged-in user
+ *
+ * @param isAuthenticated Whether we should mock the `AuthContext` value as if the user was
+ * authenticated.
+ */
+export const mockAuthProviderValue = (isAuthenticated: boolean) => {
+  let userInfo = null;
+  if (isAuthenticated) {
+    userInfo = {
+      email: faker.internet.email(),
```
vorillaz, line 42: Faker produces random data on demand.
If by chance you would try and make any assertions against this slice of information you may encourage inconsistencies. Personally I would make these files static or even use something like [chai's factories](https://www.chaijs.com/plugins/chai-factories/)

web/__tests__/utils/queries.ts
```diff
@@ -0,0 +1,46 @@
...
+
+// Builds custom queries based on aria-attribute selectors
+const buildQueryForAriaAttribute = (ariaAttribute: string) => {
```
vorillaz, line 22: That's something I was thinking about when we first discussed the possibilities of Jest.
Personally I would add dead simple but quite specific selectors for testing purposes.
Something like `data-testid` or so.
In that manner, you can explicitly detect any visual assertions as you browse through the codebase.

This is a testing strategy Instagram, Twitter, and other major companies are using with Jest.
Screenshot 2020-07-13 at 11 26 14 AM

web/__tests__/utils/render.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  return {
+    history,
+    userInfo: authProviderValue.userInfo,
+    ...rtlRenderResult,
+  };
+};
+
+export { render };
+export * from '@testing-library/react';
```
vorillaz, line 95: This export can be placed inside `web/__tests__/utils/index.ts`

web/src/components/utils/UIProviders/UIProviders.tsx
vorillaz: @3nvi I am 100% confident about this one. This is the way to go in order to keep everything in sync with the production environment.

tools/mage/test_namespace.go
3nvi: Hmm.. interestingly, I think we are caching our node_modules based on a hash created by the `package-lock.json` content, to ensure that tests are running fast when no npm modules are modified (@austinbyers am I right?).  How does  `npm ci` differ from an `npm i` when we have a valid `package-lock`? Won't  it install the exact same dependencies? The only difference I'm seeing is the "clean slate" where `node_modules` get deleted prior to any npm module installation.

web/__tests__/setup.ts
3nvi: I don't like this either, but the `pantherConfig` just parses the context of a script tag.

My other solution would be to put the `pantherConfig` on its own file (it's currently declared in configuration file among other stuff) and then just "mock" all imports to this file in jest. Didn't  do this since it would add a lot of changes in a single PR, but If you prefer this approach I can easily spin up a different PR to do it, prior to merging this

vorillaz: I don't have strong feelings about this one, I am just thinking that we may miss something are the config is declared in two places.

web/__tests__/utils/auth.ts
3nvi: Right.

in the `render`  function, I'm always returning the `userInfo` content, so you can safely make assertions against the data that  were generated on this particular test run. 

Factories is something I thought for the mocking of the  API related stuff (as  stated in the PR's description, another PR is coming), but I didn't  think  about using it here.

Would you prefer a "factory builder" patteern for this? Just because it's a one-off, I may just implement it manually, since all the API response builders are based on a `graphql-codegen` plugin that I found

web/__tests__/utils/queries.ts
3nvi: Sure, I'm not opposed to that. I just prefer to use them **only**  when there's no alternative. Ideally I want to use stuff visible  to the user (even `aria-label`) and fallback to `data-testid` when there's nothing else to select by

web/__tests__/utils/render.tsx
3nvi: sure

web/__tests__/setup.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+if (window.URL.revokeObjectURL === undefined) {
+  Object.defineProperty(window.URL, 'revokeObjectURL', { value: noOp });
+}
+
+/**
+ * Mock the server-side EJS-injected AWS configuration.
+ * See `web/public/index.ejs`
+ */
+loadDotEnvVars(path.resolve(__dirname, '.env.test'));
```
3nvi, line 62: @vorillaz Let me know if you prefer this approach (see the [related commit](https://github.com/panther-labs/panther/pull/1148/commits/05cbd9b4c72d90fba1e9a377e0c54a0805307598))

Essentially I'm creating a "mock" dotenv file and load the config from there. The `getAppTemplateParams` already has a busting mechanism, where if an ENV var is not set, then an error will be thrown.

Do you feel that this is a better approach?

tools/mage/test_namespace.go
austinbyers: Right, we tell CircleCI to cache all dependencies (Go modules, Go helper binaries, NPM modules, pip packages):

https://github.com/panther-labs/panther/blob/caec9e529006fc94d8a7e3ddbb0c3bd75d2dded1/.circleci/config.yml#L25-L28

and https://github.com/panther-labs/panther/blob/caec9e529006fc94d8a7e3ddbb0c3bd75d2dded1/.circleci/config.yml#L49-L53

web/__tests__/setup.ts
vorillaz: This one looks super clean, kudos!

web/__tests__/utils/auth.ts
vorillaz: I would go with a factory builder with chaining, maybe since the `graphql-codegen` is in place we can tweak the npm script a bit in order to create factory data one demand. For the purpose of this PR we're ready to go!

tools/mage/test_namespace.go
3nvi: @theodore given the conversation on Slack #eng, we agreed to create an RFC about npm package locking. We can continue the conversation there.

Feel free to use Austin's feedback as a reference on how we can apply a better package locking mechanism in our CI as well

3nvi: I'm marking this as resolved for the context of this PR, while waiting for an RFC for the related thread

web/__tests__/utils/auth.ts
3nvi: Take a look at [this](https://github.com/panther-labs/panther/pull/1158/files#diff-3303cdeedcb25308330ef87aba6dd231)

web/__tests__/utils/queries.ts
3nvi: Flagging as resolved :)

PR #1150, opened 2020-07-13
Add helper types and functions to enforce proper JSONL output.

alxarch:
## Background

There is a corner case when using `jsoniter.RawMessage` and writing output to `JSONL` files.

If the input contains formatted `JSON` that includes newline characters, the output will not strip these new lines resulting
in `JSONL` output that is invalid.

This PR adds helpers and methods to enforce proper JSONL formatting regardless of input JSON.
The helpers of this PR are meant to be used by any part of Panther that needs to output JSONL and is unsure about the input JSON format (ie calls to external JSON APIs in log pullers).

## Changes

- Adds `AppendJoinLines` low level helper to strip new lines from input.
- Adds `EncoderJSONL` type that writes proper JSONL output to the underlying `io.Writer`
- Adds `SingleLineRawMessage` type that acts like `jsoniter.RawMessage` but strips all new line characters.

## Testing

- Unit tests

### DIFF
--- a/internal/log_analysis/log_processor/jsonutil/jsonl.go
+++ b/internal/log_analysis/log_processor/jsonutil/jsonl.go
@@ -1,0 +1,126 @@
+package jsonutil
+
+/**
+ * Panther is a Cloud-Native SIEM for the Modern Security Team.
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import (
+	"bytes"
+	"io"
+
+	jsoniter "github.com/json-iterator/go"
+)
+
+// AppendJoinLines appends src to dst omitting new lines ('\n').
+// It can be used for removing new lines from a jsoniter.RawMessage so it can safely be used in JSONL.
+// It is safe to use tha same buffer as dest and src to remove new lines in-place:
+// ```go
+// ```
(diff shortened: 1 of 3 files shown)


internal/log_analysis/log_processor/jsonutil/jsonl.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"bytes"
+	jsoniter "github.com/json-iterator/go"
+	"io"
+	"reflect"
+	"unsafe"
+)
+
+// SingleLineRawMessage is a jsoniter.RawMessage that removes newlines so it can safely be used in JSONL outputs.
+type SingleLineRawMessage []byte
```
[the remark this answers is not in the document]
alxarch: Unfortunately the go JSON interfaces always add overhead to the process. Also by not providing an interface to detect empty values we would end up always having to use `*SingleLineRawMessage` all over the place.

internal/log_analysis/log_processor/jsonutil/jsonl.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"bytes"
+	"io"
+
+	jsoniter "github.com/json-iterator/go"
+)
+
+// AppendJoinLines appends src to dst omitting new lines ('\n').
+// It can be used for removing new lines from a jsoniter.RawMessage so it can safely be used in JSONL.
+// It is safe to use tha same buffer as dest and src to remove new lines in-place:
```
kostaspap, line 30: Well, then we can just change the signature to 
```
func RemoveNewLines(buf []byte)
```

Let chance for misusing it in our `destination` code (and copying data without need to)

internal/log_analysis/log_processor/jsonutil/jsonl.go
```diff
@@ -0,0 +1,126 @@
...
+		if pos := bytes.IndexByte(src, newLine); 0 <= pos && pos < len(src) {
+			dst, src = append(dst, src[:pos]...), src[pos+1:]
+		} else {
+			return append(dst, src...)
+		}
+	}
+	return dst
+}
+
+type EncoderJSONL struct {
```
kostaspap, line 45: What's all this needed for? The bug fix is about eliminating the chance that we write `\n` to the destination output. What's the code below used for?

kostaspap (commented):
Is it possible to submit this bug fix for the S3 destination? 
Can you add to this the fix for the writer in the pullers as well?

jacknagz:
@alxarch can we close this if it's inactive?

alxarch:
Will reopen when we refactor the log processor

PR #1151, opened 2020-07-13
Replaced cache.modify with cache.evict for deletions 

alexmylonas:
## Background

With the latest apollo upgrade, we can substitute `cache.modify` code for deletions with one liners `cache.evict`. This will provide clarity as well with getting us up to date with the latest version. Users shouldn't notice any difference. 

## Changes

- Replace cache.modify in `DeleteComplianceSourceModal`, `DeleteDestinationModal`, `DeleteGlobalPythonModuleModal`, `DeleteLogSourceModal`, `DeletePolicyModal`, `DeleteRuleModal`, `DeleteUserModal`

## Testing

- Verified cached object are removed both from apollo cache and from the UI.

### DIFF
--- a/web/src/components/modals/DeletePolicyModal/DeletePolicyModal.tsx
+++ b/web/src/components/modals/DeletePolicyModal/DeletePolicyModal.tsx
@@ -48,7 +48,6 @@ const DeletePolicyModal: React.FC<DeletePolicyModalProps> = ({ policy, ...rest }
     update: async cache => {
       cache.evict({ id: cache.identify({ ...policy, __typename: 'PolicySummary' }) });
       cache.evict({ id: cache.identify({ ...policy, __typename: 'PolicyDetails' }) });
-      cache.gc();
     },
     onCompleted: () => {
       pushSnackbar({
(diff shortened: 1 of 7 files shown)


web/src/components/modals/DeletePolicyModal/DeletePolicyModal.tsx
```diff
@@ -46,32 +46,8 @@ const DeletePolicyModal: React.FC<DeletePolicyModalProps> = ({ policy, ...rest }
...
-              return data;
-            }
-            return helpers.DELETE;
-          },
-        },
-      });
-
-      cache.gc();
+      cache.evict({ id: cache.identify({ ...policy, __typename: 'PolicySummary' }) });
+      cache.evict({ id: cache.identify({ ...policy, __typename: 'PolicyDetails' }) });
```
3nvi, line 50: @alexmylonas have we tested that this doesn't fail if either the `PolicySummary` or `PolicyDetails` doesn't exist in the cache?

alexmylonas: Yeap, tested it out. 

This specific update runs twice one for optimistic response and one for regular response. No error thrown whatsover but i found a bug when deleting a policy directly from policy details card (also affecting master)

Issue #1025, opened 2020-06-10
Prevent Python saves if tests do not pass


austinbyers:
### Describe the ideal solution

When editing a rule or policy, the save operation should fail if the rule/policy is enabled and unit tests are failing. Unit tests don't have much value unless they are enforced.

If a rule/policy does not have unit tests, we can either allow it ("all" of the 0 tests are passing) or deny it (you can't put a rule in production if it doesn't have tests).

### Describe alternatives you have considered

Today, we assume rule authors will always run tests on their own when developing Python in the web app. This is better than nothing, but imagine the frustration if you don't get an alert you were expecting because you made a typo which the existing unit tests would have prevented.

3nvi:
Just verifying that this is not related to the front-end, correct?

austinbyers:
Right, this is backend only as long as rule/policy saves already gracefully handle error conditions

kostaspap:
> If a rule/policy does not have unit tests, we can either allow it ("all" of the 0 tests are passing) or deny it (you can't put a rule in production if it doesn't have tests).

In my eyes the first option is probably preferable. I am afraid the second can be disruptive for some teams. In the end, we should let the customers decide what's best for their usecases :)

Users can still save a policy/rule with failing tests if it is disabled, right? I think you are implying this, just making sure are thoughts are aligned.

austinbyers:
> Users can still save a policy/rule with failing tests if it is disabled, right?

Yes, that's what I envisioned

kostaspap:
Summing up different scenarios: 
1. Policy/Rule enabled and tests passing: Allow saving
2. Policy/Rule disabled and tests passing: Allow saving
3. Policy/Rule enabled and tests failing: Block saving
4. Policy/Rule disabled and tests failing: Allow saving
5. Policy/Rule enabled and no tests: Allow saving
6. Policy/Rule disabled and no tests: Allow saving.

PR #1152, opened 2020-07-13
Prevent saving an enabled policy/rule if their tests fail

giorgosp:
## Background

See #1025 as Austin describes the background neatly. The idea is to "force" users to create unit tests for their rules and policies to verify their expected system behavior.

## Changes

Saving an enabled policy/rule with failing tests returns an API error. See attached screenshot.
However, the user is still allowed to save a rule/policy if it is disabled or they disable it in the same request.

## Testing

Using the existing integration tests and through manual UI testing.
Testing can also be improved with some more specialized unit or integration tests which I plan to commit next.

Screenshot 2020-07-13 at 5 58 49 PM


internal/core/analysis/doc.go
```diff
@@ -0,0 +1,20 @@
+// package analysis contains proxies to interact with the Panther analysis backends.
+package analysis
```
austinbyers, line 2: Today, each directory in `internal/core`, `internal/compliance`, and `internal/log_analysis` represents a Lambda function.

This `analysis` pkg you've introduced is not a lambda function, but just helper functions for `analysis_api`, right? In that case, move this to a package underneath the `analysis_api` folder.

If you want this code to be shared across multiple functions, then you could move this to `/pkg/analysis` (relative to the repo root).

What is the purpose of this empty `doc` file? You can put the "package" comment in any file in the package

internal/core/analysis/policy_engine.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return engineOutput, errors.Wrap(err, "failed to invoke the policy engine")
+	}
+	if invokeOutput.FunctionError != nil {
+		return engineOutput, errors.Errorf("failed to run policy:\n %s", *invokeOutput.FunctionError)
+	}
+
+	if err := jsoniter.Unmarshal(invokeOutput.Payload, &engineOutput); err != nil {
+		return engineOutput, errors.Wrap(err, "failed to unmarshal engine response into PolicyEngineOutput")
+	}
```
austinbyers, line 129: Our `genericapi` package provides an `Invoke` function which handles these 15 lines: marshaling the input, sending the request, checking for errors, and unmarshaling the response:

```suggestion
	err := genericapi.Invoke(e.lambdaClient, e.lambdaName, &input, &output)
```

internal/core/analysis/policy_engine.go
```diff
@@ -0,0 +1,162 @@
...
+
+	return engineOutput, err
+}
+
+func makeTestResources(policy *models.TestPolicy) ([]enginemodels.Resource, error) {
+	resources := make([]enginemodels.Resource, len(policy.Tests))
+	for i, test := range policy.Tests {
+		// TODO: Can swagger unmarshall this already?
```
austinbyers, line 137: Sadly, no, not to my knowledge. All of our API gateway / swagger - backed APIs have to more or less manually parse the input. One of many reasons I hope to move away from this as soon as possible

internal/core/analysis/rule_engine.go
```diff
@@ -0,0 +1,138 @@
...
+	jsoniter "github.com/json-iterator/go"
+	enginemodels "github.com/panther-labs/panther/api/gateway/analysis"
+	"github.com/panther-labs/panther/api/gateway/analysis/models"
+	"github.com/pkg/errors"
+	"strconv"
+	"strings"
+)
+
+// RuleEngine is a proxy for the rule engine backend (currently another lambda function).
+type RuleEngine struct {
```
austinbyers, line 34: This entire file is nearly identical to `policy_engine`, which makes sense since they should share the same models.

Try to either pull out common functions or just use the same `type PythonEngine struct` with `RunRuleTests` and `RunPythonTests` functions instead of a single `Run` function

internal/core/analysis/policy_engine.go
austinbyers: I know you were just copying the code, but might as well clean it up while we're here :)

austinbyers:
Great work! Looking forward to this change

This closes #1025, right? If so, add "closes: #1025" to the PR description and GitHub will automatically close the issue when the PR merges

internal/core/analysis/policy_engine.go
giorgosp: Of course, fixed

internal/core/analysis/policy_engine.go
giorgosp: We already parse the policy once
```
func parseTestPolicy(request *events.APIGatewayProxyRequest) (*models.TestPolicy, error) {
	var result models.TestPolicy
	if err := jsoniter.UnmarshalFromString(request.Body, &result); err != nil {
		return nil, err
	}

	if err := result.Validate(nil); err != nil {
		return nil, err
	}

	return &result, nil
}
```
and would expect it to unmarshall every policy.Tests[x].Resource as well. At least into a `map[string]interface{}`.
Will give it a try, I think it is just a matter of the swagger definitions.

internal/core/analysis/rule_engine.go
giorgosp: Actually removed the `Run` function in favor of `genericapi.Invoke`

internal/core/analysis_api/handlers/create_policy.go
```diff
@@ -19,12 +19,12 @@ package handlers
  */
 
 import (
+	"errors"
```
giorgosp, line 22: Maybe we should somehow enforce the use of `pkg/errors` instead of std errors package, for consistency

giorgosp:
> This closes #1025, right? If so, add "closes: #1025" to the PR description and GitHub will automatically close the issue when the PR merges

Have already added it inside a commit :)

internal/core/analysis/doc.go
austinbyers: > It could be placed in any file you mentioned but if a package has many files, in which would you put the doc? 

Usually, we do it on the "main" file, the entry point to the package or the one that defines the global variables. "main.go", "api.go", "handler.go", something like that - so far, our packages have always had a clear entry point. I'm fine with `doc.go` if you like

In reality, we don't do this consistently at all and have never once cared about the `go doc` output, lots of room for improvement here.

internal/core/analysis_api/handlers/create_policy.go
austinbyers: Consistent error handling is yet another aspect we'd love to standardize when unifying our APIs, I already have a note about it. Today, we necessarily have to handle errors differently in API gateway handlers like this one vs lambda API handlers like `outputs-api` or `users-api`

The reason for the difference - errors returned from API gateway need to be converted into HTTP response codes, errors returned from a Lambda API will trigger errors and alarms, which isn't often what we want.

Feel free to handle the error however you like, but remember that ultimately the error message is shown to the end user, so it needs to be clear and concise - no stack traces, for example

austinbyers:
> Have already added it inside a commit :)

Will that work when the commits are squashed? I had no idea

internal/core/analysis_api/analysis/doc.go
```diff
@@ -0,0 +1,20 @@
+// package analysis contains proxies to interact with the Panther analysis backends.
```
austinbyers, line 1: Nit: `package` should be capitalized (complete sentence)

`golint` enforces things like this, but our new metalinter does not. I may have disabled the check and forgot about it...

giorgosp: Done.

Issue #1135, opened 2020-07-07
Alert statuses permissions


s0l0ist:
### Description

The user needs to be able to apply a status to an alert which involves write permissions. Consideration on permissions assignment is critical.

### Acceptance Criteria

- Define which roles will receive the ability to write to an alert
- The user should be able to click a button and select a status
- The user should have access to this button in three locations: alerts table, alert details, rule details (associated alerts table)

Issue #1134, opened 2020-07-07
Alert status filtering


s0l0ist:
### Description

Allow the user to filter alerts by their status

### Acceptance Criteria

Once the user is able to apply a `status` to an alert, he should be able to filter the table by multiple status indicators.

- Dropdown for selecting (multiple) statuses
- Frontend should query, process, and return only the alerts with the desired status.

PR #1156, opened 2020-07-13
Alert statuses

### LINKED ISSUE #1098: Alert Statuses
### Description

Alerting is one of the products which provides valuable insights; however, there are times when a rule/policy can get noisy and generate many alerts. Upon inspection, many alerts maybe _false_ positives. In other cases, it is useful for an analyst to be able to mark the statuses of an alert during an investigation to track progress in a multi-user environment. In order to reduce clutter in the interface, a user must be able to mark an alert with a status such as:

- open (default)
- triaged (investigating)
- closed (false positive)
- resolved (remediated)

Adding this feature will allow the user more easily manage, analyze, and resolve their Panther alerts.

### Acceptance Criteria

- Add a field in the alert which stores a status level
- Define a reasonable set of statuses for the user to select
- Be able to click on an alert to view the details and easily change the status
- Be able to filter by the new statuses on the view which lists all the alerts

s0l0ist:
## Background

Closes #1098 #1135 #1134 

## Changes

- Added changes to backend to support alert statuses
- Backend allows filtering based on the selected alert status
- Frontend allows the user to modify the alert's status in three locations (List Alerts, Alert Details, Rule Details)
- Frontend shows tooltip containing the user's name and timestamp from the last user who modified the alert.

## Testing

- mage test:ci
- manual frontend testing with empty/null data in DDB for the newly added fields to simulate old data

## Screenshots
Screen Shot 2020-07-21 at 4 57 13 PM
Screen Shot 2020-07-21 at 4 57 36 PM
Screen Shot 2020-07-21 at 5 04 47 PM
Screen Shot 2020-07-21 at 4 57 42 PM
Screen Shot 2020-07-21 at 4 57 50 PM
Screen Shot 2020-07-21 at 4 57 59 PM


internal/log_analysis/alert_forwarder/forwarder/forwarder_test.go
```diff
@@ -37,7 +37,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	policiesclient "github.com/panther-labs/panther/api/gateway/analysis/client"
-	"github.com/panther-labs/panther/api/gateway/analysis/models"
+	rulesModel "github.com/panther-labs/panther/api/gateway/analysis/models"
```
austinbyers, line 40: Nit: in the previous file, you use `ruleModel`, let's be consistent one way or the other (no preference which)

internal/log_analysis/alerts_api/api/update_alert.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"github.com/panther-labs/panther/api/lambda/alerts/models"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/common"
+	"github.com/panther-labs/panther/pkg/gatewayapi"
+	"github.com/panther-labs/panther/pkg/genericapi"
+)
+
+// UpdateAlert modifies user attributes.
+func (API) UpdateAlert(input *models.UpdateAlertInput) (result *models.UpdateAlertOutput, err error) {
+	operation := common.OpLogManager.Start("updateAlert")
```
austinbyers, line 30: I don't think you need this here, because the `genericapi` router handles it for you

internal/log_analysis/alerts_api/api/update_alert.go
```diff
@@ -0,0 +1,63 @@
...
+	// Perform input sanitization checks
+	if input.AlertID == nil {
+		return nil, &genericapi.InternalError{Message: "An AlerID must be specified"}
+	}
+	if input.Status == nil {
+		return nil, &genericapi.InternalError{Message: "A Status must be specified"}
+	}
+	if input.RequesterID == nil {
+		return nil, &genericapi.InternalError{Message: "A RequesterID must be specified"}
+	}
```
austinbyers, line 45: As long as you have the `validate` tags in the input, you don't need these checks. The `genericapi` router will automatically validate the input and return the appropriate error before invoking this handler

internal/log_analysis/alerts_api/api/update_alert.go
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+import (
+	"github.com/panther-labs/panther/api/lambda/alerts/models"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/common"
+	"github.com/panther-labs/panther/pkg/gatewayapi"
+	"github.com/panther-labs/panther/pkg/genericapi"
+)
+
+// UpdateAlert modifies user attributes.
```
austinbyers, line 28: Copy-pasta comment is a lie!

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -153,6 +149,18 @@ func (table *AlertsTable) ListAll(input *models.ListAlertsInput) (
 	return summaries, lastEvaluatedKey, nil
 }
 
+func genStartKey(input *models.ListAlertsInput) (DynamoItem, error) {
+	// Optionally continue the query from the "primary key of the item where the [previous] operation stopped"
+	var queryExclusiveStartKey DynamoItem
+	if input.ExclusiveStartKey != nil {
+		queryExclusiveStartKey = make(DynamoItem)
+		if err := jsoniter.UnmarshalFromString(*input.ExclusiveStartKey, &queryExclusiveStartKey); err != nil {
```
austinbyers, line 157: Use Dynamo marshalers rather than JSON to convert to/from Dynamo items. In this case, [dynamodbattribute.MarshalMap](https://docs.aws.amazon.com/sdk-for-go/api/service/dynamodb/dynamodbattribute/#MarshalMap) will convert the struct into `map[string]*dynamodb.AttributeValue`, aka `DynamoItem`

internal/log_analysis/alerts_api/table/update_alert.go
```diff
@@ -0,0 +1,128 @@
...
+	return expression.
+		Set(expression.Name(StatusKey), expression.Value(input.Status)).
+		Set(expression.Name(UpdatedByKey), expression.Value(input.RequesterID)).
+		Set(expression.Name(UpdatedByTimeKey), expression.Value(aws.Time(time.Now().UTC())))
+}
+
+// createConditionBuilder - creates a condition builder
+func createConditionBuilder(input *models.UpdateAlertInput) expression.ConditionBuilder {
+	return expression.AttributeExists(expression.Name(AlertIDKey)).
+		And(expression.Equal(expression.Name(AlertIDKey), expression.Value(input.AlertID)))
```
austinbyers, line 91: Don't need the first check, if the attribute equals the target value, then it most certainly exists

internal/log_analysis/alert_forwarder/forwarder/forwarder_test.go
s0l0ist: nice catch, will update

internal/log_analysis/alerts_api/api/update_alert.go
s0l0ist: Yep, this was an artifact while testing. Thank you!

internal/log_analysis/alerts_api/api/update_alert.go
```diff
@@ -0,0 +1,44 @@
...
+	// Run the update alert query
+	alertItem, err := alertsDB.UpdateAlert(input)
+	if err != nil {
+		return nil, err
+	}
+
+	// If there was no item from the DB, we return an empty response
+	if alertItem == nil {
+		return nil, nil
+	}
```
kostaspap, line 37: What will this return? I assume this will return a serialized `models.UpdateAlertOutput` with empty values, right? Should you call `gatewayapi.ReplaceMapSliceNils` to the item returned?

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -238,6 +248,41 @@ func filterBySeverity(filter *expression.ConditionBuilder, input *models.ListAle
...
 }
 
+// filterByStatus - filters by Status(es)
+func filterByStatus(filter *expression.ConditionBuilder, input *models.ListAlertsInput) {
+	if len(input.Status) > 0 {
+		// Start with the first known key
+		var multiFilter expression.ConditionBuilder
+
+		// Alerts that don't have a status or have an empty string status are considered open.
+		if *input.Status[0] == models.OpenStatus {
```
kostaspap, line 258: Wouldn't it be easier to just define the `models.ListAlertsInput.Status` as `[]string`? You could avoid some dereferences here and in lines 265, 271, 278. I'm not sure there is any benefit to defining it as `[]*string`.

internal/log_analysis/alerts_api/table/table.go
```diff
@@ -70,6 +74,9 @@ type AlertItem struct {
 	CreationTime    time.Time `json:"creationTime"`
 	UpdateTime      time.Time `json:"updateTime"`
 	Severity        string    `json:"severity"`
+	Status          string    `json:"status"`
 	EventCount      int       `json:"eventCount"`
 	LogTypes        []string  `json:"logTypes"`
+	UpdatedBy       string    `json:"updatedBy"`
+	UpdatedByTime   time.Time `json:"updatedByTime"`
```
kostaspap, line 81: `updateTime` vs `updateByTime` can be confusing. I am not sure however that we can change `updateTime` now - it is how it is stored in DDB. 
Worth putting a comment on top of these two variables though.

internal/log_analysis/alerts_api/table/update_alert.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// createUpdateBuilder - creates an update builder
+func createUpdateBuilder(input *models.UpdateAlertInput) expression.UpdateBuilder {
+	// When settig an "open" status we actually remove the attribute
+	// for uniformity against previous items in the database
+	// which also do not have a status attribute.
+	if *input.Status == models.OpenStatus {
+		return expression.
+			Remove(expression.Name(StatusKey)).
```
kostaspap, line 77: So the "OPEN" status will be store in DDB as empty/nil? Is this really what we want long term? 
Seems that long term it would be just better to store the "OPEN" string for new fields and we can have some backwards compatible check when querying.

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -238,6 +248,41 @@ func filterBySeverity(filter *expression.ConditionBuilder, input *models.ListAle
 	}
 }
 
+// filterByStatus - filters by Status(es)
+func filterByStatus(filter *expression.ConditionBuilder, input *models.ListAlertsInput) {
```
kostaspap, line 252: Seems like "status" is a good candidate for a Global Secondary Index? It will be much faster than filtering.

api/lambda/alerts/models/api.go
```diff
@@ -92,6 +95,47 @@ type ListAlertsInput struct {
...
+	RequesterID *string `json:"requesterId" validate:"required,uuid4"`
+}
+
+// UpdateAlertOutput the returne alert summary after an update
+type UpdateAlertOutput = AlertSummary
+
+// Constants defined for alert statuses
+const (
+	// Empty is present on items that already exist in the DB and is considered in an "OPEN" state
+	EmptyStatus = ""
```
kostaspap, line 124: Since this package is defining the public API, I think the `EmptyStatus` shouldn't be here. We could/should make sure that the Alerts API always returns a "populated" status. Any empty status checks can be performed on the DDB-related code e.g. you could modify the `alertItemToAlertSummary` to return `OpenStatus` when it encountered an empty value. It would make any caller easier to deal with this API outputs.

api/lambda/alerts/models/api.go
```diff
@@ -92,6 +95,47 @@ type ListAlertsInput struct {
...
+// }
+type UpdateAlertInput struct {
+	// ID of the alert to update
+	AlertID *string `json:"alertId" validate:"required,hexadecimal,len=32"` // AlertID is an MD5 hash
+
+	// Variables that we allow updating:
+	Status *string `json:"status" validate:"required,oneof=OPEN TRIAGED CLOSED RESOLVED"`
+
+	// User who made the change
+	RequesterID *string `json:"requesterId" validate:"required,uuid4"`
```
kostaspap, line 115: 1. nit: You can leave out the `required` from the `validate` tag. The `len=32`, `oneof`, `uuid4` checks will fail if the value is not of len32, one of the given options or valid UUID4. 
2. Would your code be simpler if these fields were defined as `string` instead of `*string`? It's your choice, but don't feel obliged to define them as pointers if it doesn't give you any benefit. We have seen that in other APIs that trying to define everything as a pointer complicated things so decided to use pointers only when we see appropriate. 
3. In other APIs this field is defined as `UserID`, e.g. https://github.com/panther-labs/panther/blob/74511248bd2ff5bdfb911a0de47d29b80fc6dbef/api/lambda/source/models/api.go#L74  https://github.com/panther-labs/panther/blob/master/api/gateway/analysis/api.yml#L928-L929

I think we can stick to that convention

api/graphql/schema.graphql
```diff
@@ -137,6 +139,9 @@ type AlertDetails {
   events: [AWSJSON!]!
   eventsLastEvaluatedKey: String
   dedupString: String!
+  status: String
+  updatedBy: ID
+  updatedByTime: AWSDateTime
```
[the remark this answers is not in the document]
s0l0ist: You're correct, I didn't want to clobber the original `updateTime`. At this time, I'm not entirely sure if we can change the name since this would most likely require a migration.

What about additional comments instead? At the very least, when creating a unified API we could have an adapter that transforms these names.

internal/log_analysis/alerts_api/table/update_alert.go
s0l0ist: @kostaspap This was a problem I debated about for along time. There are pros and cons.

One of the primary reasons for not setting an "OPEN" state was because there would still be older data that doesn't have this value set. In addition, after speaking with others we decided **against** performing a migration for all previous alerts and ones that are in-flight.

Given that we wouldn't be updating older alerts, this introduces a bi-state:
- Empty status (attribute doesn't exist)
- Valid status enum (OPEN, TRIAGED, CLOSED, RESOLVED)

Another issue is now normalizing alerts coming from DDB that didn't have the attribute (empty) and marshaling to the "OPEN" state. This needs to be done in all places where we retrieve an alert from DDB (get, list, update) and of course any new APIs that are added which can lead to bugs in the future including other APIs which may read from this table.

Rather than have to deal with _two_ "OPEN" states, we can simply remove the "OPEN" attribute from enum and treat all empty as "OPEN" to simplify all APIs and guarantee consistency.

I'm okay with writing "OPEN" states for new alerts and allowing the user to write (via update) them as well. I just wanted to mention the above issue.

api/graphql/schema.graphql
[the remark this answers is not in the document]
s0l0ist: instead of changing `updateTime` I've changed `updatedBy*` to lastUpdatedBy*`

internal/log_analysis/alerts_api/table/list.go
s0l0ist: I initially thought that as well, but then how would we be able to query data for a previous record that doesn't contain this the `status` field?  Would this also require a migration/backfilling? I'm not too knowledgeable with DDB...

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -47,12 +46,6 @@ const (
 
 // GetAlert retrieves details for a given alert
 func (API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOutput, err error) {
-	operation := common.OpLogManager.Start("getAlert")
-	defer func() {
-		operation.Stop()
-		operation.Log(err)
-	}()
```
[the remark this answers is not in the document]
austinbyers: I wasn't sure the purpose of these when we already have them in the `genericapi` router, what's the difference?

https://github.com/panther-labs/panther/blob/9f99a89bdc036b2b51b924a66b4fb2e74bf9f877/pkg/genericapi/router.go#L76-L79

internal/log_analysis/alerts_api/api/update_alert.go
austinbyers: Good catch - yes, we shouldn't be returning `nil` here because it has the wrong type for the frontend.

The generic router already handles the replacement, but the result needs to be non-nil for it to work:

https://github.com/panther-labs/panther/blob/9f99a89bdc036b2b51b924a66b4fb2e74bf9f877/pkg/genericapi/router.go#L117-L121

So this could be

```go
return nil, &models.UpdateAlertOutput{}
```

internal/log_analysis/alerts_api/table/list.go
austinbyers: I think a GSI would work with an undefined field, (Dynamo / noSQL is all about schema flexibility) but that's a good point, I'm not sure how it would handle that, either

api/graphql/schema.graphql
```diff
@@ -27,6 +27,7 @@ type Mutation {
   resetUserPassword(id: ID!): User!
   suppressPolicies(input: SuppressPoliciesInput!): Boolean
   testPolicy(input: TestPolicyInput): TestPolicyResponse
+  updateAlert(input: UpdateAlertInput!): AlertSummary
```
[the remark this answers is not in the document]
3nvi: Although counter-intuitive, in GraphQL we welcome duplication. We want to create mutations that are as specific to the thing that  we are modifying as  we can.

I understand that we haven't committed to that 100%, but wherever we haven't it's just old code that we never had the chance to update

web/src/components/dropdowns/UpdateAlertDropdown/graphql/updateAlert.graphql
```diff
@@ -0,0 +1,5 @@
+mutation UpdateAlert($input: UpdateAlertInput!) {
+  updateAlert(input: $input) {
```
3nvi, line 2: ideally we should create a [`merge` policy](https://www.apollographql.com/docs/react/caching/cache-field-behavior/#the-merge-function) where whenever `AlertSummary` status is updated, the corresponding `AlertDetails` also gets updated in the cache (and vice versa).

Currently, I saw that the Alert Details page has a `cache-and-network` fetch policy which makes sure to "refetch" data on each visit, but it  would be nice if they were already in sync within the cache

web/src/pages/AlertDetails/AlertDetailsInfo/AlertDetailsInfo.tsx
```diff
@@ -81,36 +83,55 @@ const AlertDetailsInfo: React.FC<AlertDetailsInfoProps> = ({ alert, rule }) => {
...
-        >
-          <Icon type="info" />
-        </Tooltip>
-      </Flex>
-      <Flex spacing={4} as="ul" mb={6}>
-        <Box as="li">
-          <SeverityBadge severity={rule.severity} />
-        </Box>
-      </Flex>
+      <SimpleGrid columns={2} spacing={5}>
```
3nvi, line 86: subjective: Since we opted for a `<Badge />`, we normally  "group" all badges next to one another, so I would put it right next  to the existing `<SeverityBadge />` component, to follow the convention of other pages

web/src/pages/AlertDetails/AlertDetailsInfo/AlertDetailsInfo.tsx
```diff
@@ -81,36 +83,55 @@ const AlertDetailsInfo: React.FC<AlertDetailsInfoProps> = ({ alert, rule }) => {
...
+            </Box>
+          </Flex>
+        </Flex>
+        <Flex direction="row" justify="flex-end" align="top" spacing={2}>
+          <Flex direction="column" align="center" spacing={4}>
+            <Box
+              color="gray-100"
+              fontSize="medium"
+              fontWeight="bold"
+              aria-describedby="alert-status-description"
```
3nvi, line 125: nice :)

web/src/components/AlertStatusBadge/AlertStatusBadge.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import { Badge, BadgeProps, PseudoBox } from 'pouncejs';
+import { AlertStatusesEnum } from 'Generated/schema';
+
+const STATUS_COLOR_MAP: {
+  [key in StatusBadgeProps['status']]: BadgeProps['color'];
```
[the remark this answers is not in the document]
3nvi: Fully subjective:

I'm not  100% sure on the `open` and `closed` colors. I think `open` might be "red" to indicate a threat. Closed on  the  other side should be something more pale like a gray (since from my understanding this is not something that should be examined)

api/graphql/schema.graphql
```diff
@@ -93,6 +94,7 @@ input ListAlertsInput {
   createdAtAfter: AWSDateTime
   ruleIdContains: String
   alertIdContains: String
+  status: [AlertStatusesEnum]
```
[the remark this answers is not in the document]
s0l0ist: Not exactly. This schema is used for filtering and the user may not put in a filter for status so it is optional.

api/graphql/schema.graphql
```diff
@@ -132,11 +134,17 @@ type AlertDetails {
...
   updateTime: AWSDateTime!
   eventsMatched: Int!
   events: [AWSJSON!]!
   eventsLastEvaluatedKey: String
   dedupString: String!
+  status: String
+  # lastUpdatedBy - stores the UserID of the last person who modified the Alert
+  lastUpdatedBy: ID
+  # lastUpdatedByTime - stores the timestamp of the last person who modified the Alert
+  lastUpdatedByTime: AWSDateTime
```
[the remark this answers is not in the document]
s0l0ist: This one is tricky because we already have `updateTime` so the idea was to create an additional field `lastUpdatedBy` and its corresponding timestamp, `lastUpdatedByTime`.

api/graphql/schema.graphql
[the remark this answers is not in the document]
s0l0ist: Makes sense, I was just conforming to what was there already.

web/src/pages/AlertDetails/AlertDetailsInfo/AlertDetailsInfo.tsx
s0l0ist: We're also waiting for final designs to come out so I expect this to change again.

web/src/components/dropdowns/UpdateAlertDropdown/graphql/updateAlert.graphql
s0l0ist: I think I've already done this defined in the `useUpdateAlertStatus` hook: https://github.com/panther-labs/panther/commit/7a1621f00adfcac14db8bd2de4c81a16166f222f#diff-c7c980e0af1f5908099ee5de965feba1R52

Also, since we get an `AlertSummary` from the update API instead of `AlertDetails` I didn't want to write logic to find, merge and return a new array. Instead I just find the specific `AlertDetails` by its `alertId` and update the settings value. The alert status component itself holds the timestamps and everything else necessary for its operation.

Do you have in mind a cleaner way?

api/lambda/alerts/models/api.go
s0l0ist: Not to nitpick, but in that reference line I see `required,uuid4` is this nomenclature redundant as well? I'm just not that familiar with golang validators and am not sure if they account for `null `_and_ empty values.

internal/log_analysis/alerts_api/api/get_alert.go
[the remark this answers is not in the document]
s0l0ist: Waiting for @kostaspap final word to keep or remove and will update accordingly.

api/graphql/schema.graphql
```diff
@@ -132,11 +134,17 @@ type AlertDetails {
...
   creationTime: AWSDateTime!
+  # updateTime - stores the timestamp from an update from a dedup event
   updateTime: AWSDateTime!
   eventsMatched: Int!
   events: [AWSJSON!]!
   eventsLastEvaluatedKey: String
   dedupString: String!
+  status: String
+  # lastUpdatedBy - stores the UserID of the last person who modified the Alert
+  lastUpdatedBy: ID
```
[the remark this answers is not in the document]
s0l0ist: @3nvi I have a resolver `UpdateAlertStatusResolver` already in `deployments/appsync.yml`. Is this what you mean?

api/graphql/schema.graphql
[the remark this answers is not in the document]
3nvi: oh right! Sorry  didn't  notice it was an `input` type. You are absolutely right 👍

api/graphql/schema.graphql
```diff
@@ -149,9 +157,20 @@ type AlertSummary {
   creationTime: AWSDateTime!
   eventsMatched: Int!
   title: String!
+  # updateTime - stores the timestamp from an update from a dedup event
   updateTime: AWSDateTime!
   ruleId: String
   severity: SeverityEnum
+  status: String
```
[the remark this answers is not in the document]
3nvi: there was a comment  that  got deleted about the naming of  `lastUpdatedByTime`. Just want to make sure we get an answer  to  that

api/graphql/schema.graphql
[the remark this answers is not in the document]
3nvi: Not exactly.

Take a look at [this](https://github.com/panther-labs/panther-enterprise/blob/e85536c2ba94442ec5f1039ae19cc233d8b5f4fe/deployments/appsync.yml#L287-L306)

You can see that we have  a "custom field" named `role` in the `User` object which gets resolved by a separate Appsync resolver. Now in your original response `lastUpdatedBy` was an `ID`, but in the linked resolver you  will see that  I'm using `$ctx.source.roleId` (in your case it  would be `$ctx.source.lastUpdatedBy`) to actually fetch a `Role` instance (in your case a `User` instance).

With this technique, the `lastUpdatedBy` in your `schema.graphql` won't be  an `ID` anymore, but an actual `User` type. This will allow you to do:

```graphql
alert {
  lastUpdatedBy {
    name
  }
}
```
since the separate field resolver for this type will handle converting the `ID` -> `User` before it reaches the FE. Makes sense?

web/src/pages/AlertDetails/AlertDetailsInfo/AlertDetailsInfo.tsx
3nvi: Most likely so.

But until then I'd **personally** prefer to follow the UI of other stuff (which with regards to this particular piece of the UI, it's considered finalized).

If you want to do a single final-pass when the designs are out, then I'm fine with this as well 😃

api/graphql/schema.graphql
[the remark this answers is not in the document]
3nvi: @kostaspap I think that "modification" as a word suits better in this context, since it implies that someone made the change (while the word "update" doesn't). Also, I feel that the english is a bit off on `lastUpdatedByTime` since it doesn't clearly  convey what this field is.

I'm not feeling strongly about renaming it, but I just wanna say that this is a name I haven't encountered ever before 😅

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
```diff
@@ -0,0 +1,173 @@
...
+    variables: {
+      input: {
+        status: status as AlertStatusesEnum,
+        alertId: alert.alertId,
+      },
+    },
+
+    // This hook ensures we update the AlertDetails's `status`
+    // field in the cache instead of refetching from the network.
+    update: (cache, { data }) => {
```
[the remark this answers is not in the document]
3nvi: Update: I now see that you wanted to sync updated in another type (namely `AlertDetails`), but that's  a bit optimistic as well, since perhaps you are making the update within the alert details page (so the `AlertSummaries` need to be updated)

web/src/components/dropdowns/UpdateAlertDropdown/graphql/updateAlert.graphql
3nvi: oh gotcha.

Yeah didn't see that you were changing another type. I think the "merge policy"-way linked above is cleaner and we have a plan to gradually move more logic in cache policies like this.

The reason is that we don't have to think about cascading updates any more, since we added the logic in a centralized place. For example, with 2 merge policies you can say "Whenever the AlertSummary `status` updated, also update the same field in `AlertDetails` in the cache" & vice versa.

In the same fashion, you can say "Whenever a `RuleSummary` gets deleted, also evict  from the cache the corresponding `RuleDetails` if you have it"

Things like  that make it easier  to "not have to worry" about remembering to update stuff in other types and in general this is the direction forward (Apollo 3.0 that features this just got released yesterday and they added this functionality out of a need for it)

api/graphql/schema.graphql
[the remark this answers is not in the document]
s0l0ist: yes, very cool. I will implement this instead.

web/src/pages/AlertDetails/AlertDetailsInfo/AlertDetailsInfo.tsx
s0l0ist: Sounds good!

internal/log_analysis/alerts_api/table/list.go
s0l0ist: After some more investigation and discussions with @kostaspap, it looks like a GSI **requires** the primary partition key to be populated with a non-null/non-empty value for an Alert and will not index existing records that don't have an attribute _or_ are empty/null values.

Ultimately, this means we would need to backfill previous records aka a migration in order to support a GSI on an alert's status. However, we've decided a migration can happen in the future if performance becomes an issue. We will move forward with not putting in a GSI for the alert status and performing filtering the traditional way.

internal/log_analysis/alerts_api/table/update_alert.go
s0l0ist: Closing as we will not be using a GSI due to limitations requiring a full migration.

api/lambda/alerts/models/api.go
s0l0ist: Will remove the `required` field moving forward to reduce clutter

internal/log_analysis/alerts_api/api/get_alert.go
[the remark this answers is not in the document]
s0l0ist: Decided to remove.

internal/core/alert_delivery/models/api.go
```diff
@@ -57,6 +57,9 @@ type Alert struct {
 	// Runbook is the user-provided triage information.
 	Runbook *string `json:"runbook,omitempty"`
 
+	// Status is the user-provided status level. An empty value is equivalent to OPEN.
+	Status string `json:"status" validate:"omitempty,oneof=TRIAGED CLOSED RESOLVED"`
```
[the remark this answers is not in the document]
s0l0ist: Removed!

internal/log_analysis/alerts_api/api/update_alert.go
s0l0ist: nice catch, I've updated it in the `GetAlert` API as well.

api/graphql/schema.graphql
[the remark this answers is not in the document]
s0l0ist: Before, the return value from an `AlertSummary` was either an empty string or one of the `enum` values.

After a few revisions, we have coerced any empty strings, `""`, to be a valid status and therefore we can make this a **required** **type**.

web/src/components/AlertStatusBadge/AlertStatusBadge.tsx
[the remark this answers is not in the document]
s0l0ist: This is definitely up in the air, I just put it together randomly with zero thought and will wait for final design.

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
```diff
@@ -0,0 +1,173 @@
...
+  ]);
+
+  // Create our dropdown button
+  const dropdownButton = React.useMemo(
+    () => (
+      <DropdownButton as={AbstractButton} outline="none" aria-label="Status Options">
+        <AlertStatusBadge status={(status as AlertStatusesEnum) || AlertStatusesEnum.Open} />
+      </DropdownButton>
+    ),
+    [alert, status]
```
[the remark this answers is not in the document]
s0l0ist: Previously, it would not because it was a string that could be empty. Now that is not the case and will update to your suggestion.

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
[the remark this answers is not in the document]
s0l0ist: This update field performs an update on the status and returns the `AlertSummaries` following a successful outcome, I update `AlertDetails` regardless of what page I'm on.

I think the `optimisticResponse` handler that I've defined updates this specific alert summary automatically (even during failure) so this solves the case where the user would see a 'flash' between two different 'states' of the alert state when navigating between the two pages after an update has complied (or failed).

Moreover in the particular case of updating on the alert details page, the cached value is already set but a request to refetch this list is made (due to filtering) and overwrites the data here anyway.

web/src/components/AlertStatusBadge/AlertStatusBadge.tsx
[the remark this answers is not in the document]
s0l0ist: Also pale gray looks pretty bad now that I'm playing with it. I've put in `transparent` for now. definitely waiting for designs to come in.

web/src/components/dropdowns/UpdateAlertDropdown/graphql/updateAlert.graphql
s0l0ist: Tabling this raised issue until we migrate to a later release of Apollo.

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
```diff
@@ -0,0 +1,173 @@
...
+  alert: AlertSummaryFull;
+}
+
+const UpdateAlertDropdown: React.FC<UpdateAlertDropdownProps> = ({ alert }) => {
+  const { status = AlertStatusesEnum.Open } = alert;
+
+  const { pushSnackbar } = useSnackbar();
+
+  // Retrieve a list of users so we can match attribution
+  const { data: listUsersData } = useListUsers({
```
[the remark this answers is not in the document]
s0l0ist: Leaving in for performance reasons

web/src/apollo/typePolicies.ts
```diff
@@ -50,27 +50,64 @@ export type TypePolicies = Partial<
...
         );
       },
-      getS3LogIntegration(existingData, { args, toReference }) {
-        return (
-          existingData || toReference({ __typename: 'S3LogIntegration', integrationId: args.id })
-        );
+      getS3LogIntegration(existing, { args, toReference }) {
+        return existing || toReference({ __typename: 'S3LogIntegration', integrationId: args.id });
       },
+      // TODO: when apollo client is updated to 3.0.0-rc.12+, use this code
```
3nvi, line 61: nice thanks!

api/graphql/schema.graphql
```diff
@@ -127,31 +129,57 @@ type S3LogIntegrationHealth {
   kmsKeyStatus: IntegrationItemHealthStatus!
 }
 
-type AlertDetails {
+interface Alert {
```
3nvi, line 132: super nice. Thanks Nick

web/src/apollo/typePolicies.ts
3nvi: I  think the commented-out code for the `alert` is not needed (In my mind we only  needed this policy for the 'fetchmore' part, meaning `alerts` and  `events` but I might be mistaken)

3nvi: In any case we can review when necessary :D

web/src/components/AlertStatusBadge/AlertStatusBadge.tsx
```diff
@@ -0,0 +1,65 @@
...
+  [AlertStatusesEnum.Closed]: 'navyblue-550' as const,
+  [AlertStatusesEnum.Resolved]: 'navyblue-500' as const,
+};
+
+interface StatusBadgeProps {
+  status: AlertStatusesEnum;
+}
+
+const AlertStatusBadge: React.FC<StatusBadgeProps> = ({ status }) => (
+  <PseudoBox role="group">
```
3nvi, line 37: This can be simplified  I think.

- No need for `PseudoBox` for the `Badge`. Just use a direct `Badge` and don't change the padding since it should be part of the design system (i.e. I don't allow you to change it  from the `<Badge>` intentionally  :P)
- Cursor should be the default one, since the Badge itself is not clickable
- The bottom component should be an `<IconButton>` component with the variant of `ghost`. It won't have a border, but we can add it manually through a wrapping `<Box>` (this "button style" was not originally  specced for pounce)

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
```diff
@@ -0,0 +1,210 @@
...
+    () => (
+      <DropdownButton as={AbstractButton} outline="none" aria-label="Alert Status Options">
+        <AlertStatusBadge status={alert.status} />
+      </DropdownButton>
+    ),
+    [alert]
+  );
+
+  // Create a wrapped dropdown button with a tooltip.
+  const wrappedDropdownButton = React.useMemo(
```
3nvi, line 158: I think I commented about not having multi-purpose triggers (dropdowns & tooltips), but I can't find your answer anymore

web/src/components/AlertStatusBadge/AlertStatusBadge.tsx
3nvi: Also, I don't think it syncs with what's here https://github.com/panther-labs/panther/issues/1162

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
3nvi: We can add the "tooltip" on the badge itself, but the dropdown trigger needs to be an external button from what I'm seeing https://github.com/panther-labs/panther/issues/1162

web/src/pages/ListAlerts/ListAlerts.tsx
```diff
@@ -65,16 +65,26 @@ const ListAlerts = () => {
           },
         },
         updateQuery: (previousResult, { fetchMoreResult }) => {
-          // FIXME: Centralize this behavior for alert pagination, when apollo fixes a bug which
```
3nvi, line 68: We still need the FIXME for when we update to the latest Apollo. Can we re-instate it?

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
s0l0ist: fixed

web/src/pages/ListAlerts/ListAlerts.tsx
s0l0ist: Sure, but I didn't find this to be an issue with the latest Apollo with the code changes I now have there.

web/src/components/AlertStatusBadge/AlertStatusBadge.tsx
s0l0ist: - Unfortunately, I've tried using `Badge`, but I cannot add the prop `display="inline-flex"` because `Badge` doesn't support it. Otherwise, you cannot center the component inside the a table.
- Fixed the cursor
- I originally tried `IconButton` but it doesn't work as expected. I can't adjust the size of the icon properly and it is still too large even when set to `size="small"`, I couldn't add custom props needed to resize it without compiler complaints, and a couple of others that I'm forgetting now.

web/src/apollo/typePolicies.ts
s0l0ist: I found on the latest Apollo this was needed to ensure we sync any merges in the attributes for `AlertDetails` such as when the user modifies the Alert `status` (on an `AlertSummary`) and this change is merged to `AlertDetails`, but then a paginated query is executed on the AlertDetails-Page and it was complaining on not being able to merge the `events` properly.

web/src/components/AlertStatusBadge/AlertStatusBadge.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <Box aria-labelledby="updated-by-timestamp-label">{lastUpdatedByTime}</Box>
+        </Flex>
+      </Flex>
+    ),
+    [lastUpdatedBy, lastUpdatedByTime]
+  );
+
+  const statusBadge = React.useMemo(
+    () => (
+      <PseudoBox
```
3nvi, line 61: let's find a better way to write this

- position-related props should be added through wrapper wherever the component is used
- `backgroundColor`  can be replaced with `color` directly on `Badge`
- `padding` should be replaced with a `width`

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
```diff
@@ -0,0 +1,206 @@
...
+    if (user.givenName && user.familyName) {
+      return `${user.givenName} ${user.familyName}`;
+    }
+    if (!user.givenName && user.familyName) {
+      return user.familyName;
+    }
+    if (user.givenName && !user.familyName) {
+      return user.givenName;
+    }
+    return user.email;
```
3nvi, line 135: This should be extracted in a util function `getUserDisplayName` and re-use it accross the app.

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
```diff
@@ -0,0 +1,206 @@
...
+      pushSnackbar({
+        variant: 'error',
+        title: `Failed to set the alert status`,
+        description: extractErrorMessage(error),
+      });
+    },
+  });
+
+  // Extracts a user from a list of users by its ID
+  const getUser = React.useCallback((listUsers, lastUpdatedBy) => {
```
3nvi, line 111: why create a getter? I feel we can have a simpler solution like:

```tsx
const user = listUsers?.users.find(({ id }) => id === lastUpdatedBy)
const userDisplayName = user ? getUserDisplayName(user)  :  ''
```

In general, don't wrap everything in a `React.useMemo`. This is for **expensive** operations. Sometimes, not using a `React.useMemo` is faster since the hashlookup, comparison & memory increase are more expensive than iterating through a list of 5 users
and be done with it

- no `React.useCallback` or  `React.useMemo`
- no additional helper functions that only get used once

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
```diff
@@ -0,0 +1,206 @@
...
+  }, []);
+
+  // Create the display name
+  const lastUpdatedBy = React.useMemo(() => getDisplayName(alert.lastUpdatedBy, listUsersData), [
+    alert,
+    listUsersData,
+  ]);
+
+  // Create a formatted timestamp
+  const lastUpdatedByTime = React.useMemo(() => formatDatetime(alert.lastUpdatedByTime), [alert]);
```
3nvi, line 145: no need for `React.useMemo`

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
```diff
@@ -0,0 +1,206 @@
...
+    () => (
+      <Flex spacing={1} justify="center" align="center">
+        <AlertStatusBadge
+          status={alert.status}
+          lastUpdatedBy={lastUpdatedBy}
+          lastUpdatedByTime={lastUpdatedByTime}
+        />
+        <DropdownButton
+          as={AbstractButton}
+          display="inline-flex"
```
3nvi, line 158: Again I feel that there can be a simpler solution to that

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
```diff
@@ -0,0 +1,206 @@
...
+            key={index}
+            disabled={alert.status === statusVal}
+            onSelect={() =>
+              updateAlertStatus({
+                variables: { input: { status: statusVal, alertId: alert.alertId } },
+              })
+            }
+          >
+            <Flex minWidth={85} spacing={2} justify="space-between" align="center">
+              <Box aria-labelledby="status-item">{statusKey}</Box>
```
3nvi, line 196: there isn't any item with the `status-item` id. Thus `aria-labelledby`  is invalid here

web/src/pages/ListAlerts/ListAlerts.tsx
3nvi: We eventually need  to add it in the `typePolicies`. It just serves as an extra reminder for me (cause I'll do a global search for `FIXME`)

PR #1158, opened 2020-07-13
FE: Add autogen GraphQL buillders & autogen API mock-functions

3nvi:
# Important!

Please read the description thoroughly

## Background

This PR compliments #1148 by autogenerating mock data & strongly typed mock queries that will facilitate API testing.

In summary, it creates 2 custom `graphql-codegen` plugins that:

- [create `faker` builders for our GraphQL types](https://github.com/panther-labs/panther/pull/1158/files#diff-3303cdeedcb25308330ef87aba6dd231)
- [create strongly-typed mock functions for our GraphQL operations (queries & mutations)](https://github.com/panther-labs/panther/pull/1158/files#diff-9691e320b7f5eaff38f18680ead79060)

Those plugins are written in TS, but `graphql-codegen` only understands JS. Thus, we have a separate `tsconfig.json` and a "building" step to translate the source code to JS outputs. Those JS outputs are created as part of the `npm run transpile-graphql-codegen-plugins` command and get used inside `codegen.yml`.

The final artifacts of those plugins are:

- `builders.generated.ts` which exposes builders for **every** GraphQL type (i.e. `buildUser`, `buildAlertDetails`, etc.)
- functions that mock **all of our GraphQL operations** that get added too the `xxx.generated.ts` files (i.e. `mockListUsers`, `mockGetAlert`, etc.)

These artifacts should be imported and used in tests. `example.test.tsx` has been updated with a showcase of that.

To sum up:

PR contains 2 custom TS plugins that get compiled into JS and used in `codegen.yml`. They produce additional functions on every `xxx.generated.ts` file and a new single `builders.generated.ts` file.

## Benefits

The benefits of this approach are:

-  type-checking on mock API operations (making sure you correctly mock what you have to mock)
- automatic generation of mock functions so you can import the corresponding mock function for each query you want to mock
- an easy way to create dynamic fixtures (through our GraphQL builders)
- every mock data is type-safe and stays up-to-date with out schema automatically (i.e. no misaligned fixtures). This is perhaps the most important benefit.

## Changes

**⚠️⚠️ Most of the changes are just updates to our `xxx.generated.ts` files. Don't be scared, since less than 15 files are up for review**

- Create 2 TS files under `__tests__/codegen` with the code for 2 custom plugins
- Create a `tsconfig` that can transpile them into JS
- Make ESLint ignore the outputted JS since it's not source code
- Creaate  `npm run transpile-graphql-codegen` to transpile plugins as JS code
- Generation of test-utility artifacts as mentioned above. This includes mocking API payloads for **each of our queries & mutations** as part of `graphql-codegen`, making sure that each React Apollo hook, also exports a corresponding mock function that is type-safe and accepts a payload as a param

## Testing

- Added custom test case for making sure builders & mock functions work

## Context

These 2 custom GraphQL codegen plugins were created because no solution out there was robust enough to serve us.
Initially, this PR contained "fixed" data (instead of dynamic resource builders), but that didn't add enough "randomness" to guarantee that our tests can cover us. Also it could potentially be mis-used by a team-member, by checking for specific values in `expect` statements, since the values would never change (i.e. a  dev might make hardcoded assertions on the values of autogenerated fixtures).

These 2 custom GraphQL plugins are based  on:

- https://github.com/ardeois/graphql-codegen-typescript-mock-data

- https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/react-apollo

and although I know the implementation is not ideal or clean, I think that the code for these plugins is irrelevant as long as the mock data can be properly produced. Generally speaking, they are based on the official GraphQL spec, so the only way they can have breaking changes is if the spec completely changes. I do want your opinion on that.


package.json
```diff
@@ -112,6 +113,7 @@
     "express-static-gzip": "~2.0.6",
     "formik": "^2.1.3",
     "graphql": "^14.6.0",
+    "graphql-codegen-typescript-mock-data": "github:3nvi/graphql-codegen-typescript-mock-data#master",
```
3nvi, line 116: Implemented bug fixes & minor features for this 3rd-party GraphQL-fake-data-generation library and ported them into a fork.

I added 2 corresponding PRs in the [original repo](https://github.com/ardeois/graphql-codegen-typescript-mock-data) and will update the dependency as soon as they are merged

web/__tests__/codegen/mockGraphqlOperationsPlugin.js
```diff
@@ -0,0 +1,78 @@
...
+class MockGraphqlOperationsVisitor extends ClientSideBaseVisitor {
+  // eslint-disable-next-line class-methods-use-this
+  buildOperation(
+    node,
+    documentVariableName,
+    operationType,
+    operationResultType,
+    operationVariablesTypes
+  ) {
+    return `export function mock${node.name.value}({ data, variables, error}: {
```
3nvi, line 36: for each operation (query or mutation), generate a type-safe string with the related query or mutation document pre-filled.

Users already import `useListUsers` for dev/prod. Now they will import `mockListUsers` for testing purposes. To use it just call:

```tsx
mockListUsers({ users: [] });
```

web/codegen.yml
```diff
@@ -49,6 +49,21 @@ generates:
         AWSTimestamp: number
         AWSDateTime: string
         AWSJSON: string
+
+  web/__tests__/__mocks__/resources.generated.ts:
```
3nvi, line 53: creates mocks for all resources in our schema. Also make sure to replace custom scalars with proper "fake" values (that's what  I submitted a PR for)

web/codegen.yml
```diff
@@ -57,6 +72,7 @@ generates:
     plugins:
       - typescript-operations
       - typescript-react-apollo
+      - web/__tests__/codegen/mockGraphqlOperationsPlugin.js
```
3nvi, line 75: make sure to generate mockAPI functions for all GraphQL operations (queries & mutations)

web/src/components/forms/BaseRuleForm/BaseRuleFormTestSection/graphql/testPolicy.generated.ts
```diff
@@ -87,3 +87,18 @@ export type TestPolicyMutationOptions = ApolloReactCommon.BaseMutationOptions<
   TestPolicy,
   TestPolicyVariables
 >;
+export function mockTestPolicy({
```
3nvi, line 90: the rest are just autogen files. Nothing to review here.

package.json
```diff
@@ -79,6 +80,7 @@
     "jest-localstorage-mock": "^2.4.2",
     "jest-watch-typeahead": "^0.6.0",
     "lint-staged": "^10.0.7",
+    "pascal-case": "^3.1.1",
```
3nvi, line 83: needed for some GraphQL -> TS manipulations. Couldn't implement it easily myself, since it had some edge cases

package.json
```diff
@@ -134,7 +136,7 @@
     "serve": "node web/scripts/serve.js",
     "test": "jest --config web/__tests__/jest.config.js",
     "test:watch": "npm run test -- --watch",
-    "graphql-codegen": "graphql-codegen --config web/codegen.yml",
+    "graphql-codegen": "tsc -p web/__tests__/codegen && graphql-codegen --config web/codegen.yml",
```
3nvi, line 139: The custom plugins are written in TS for type-safety. They need to be compiled into JS before `graphql-codegen` can use them

web/.eslintrc.js
```diff
@@ -31,7 +31,13 @@ module.exports = {
     browser: true,
     node: true,
   },
-  ignorePatterns: ['npm-debug.log*', 'dist', '__generated__', '*.generated.*'],
+  ignorePatterns: [
+    'npm-debug.log*',
+    'dist',
+    '__generated__',
+    '*.generated.*',
+    '__tests__/codegen/*.js',
```
3nvi, line 39: the custom plugins' JS code shouldn't be type checked since it's an artifact, not source code

web/__tests__/codegen/mockGraphqlOperationsPlugin.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
```
3nvi, line 1: This is an artifact. Ignore it.

web/__tests__/codegen/mockGraphqlOperationsPlugin.ts
```diff
@@ -0,0 +1,77 @@
+/**
+ * Panther is a Cloud-Native SIEM for the Modern Security Team.
```
3nvi, line 2: for each operation (query or mutation), generate a type-safe string with the related query or mutation document pre-filled.

Users already import `useListUsers` for dev/prod. Now they will import `mockListUsers` for testing purposes. To use it just call:

```tsx
mockListUsers({ data: { users: [] } });
```

Its implementation extends the basic visitor implemented through `@graphql-codegen/visitor-plugin-common`

web/__tests__/codegen/mockGraphqlResourcesPlugin.js
```diff
@@ -0,0 +1,190 @@
+'use strict';
```
3nvi, line 1: This is an artifact. Ignore it.

web/__tests__/codegen/mockGraphqlResourcesPlugin.ts
```diff
@@ -0,0 +1,238 @@
+// Based on work from @ardeois and @3nvi on MIT licensed  `graphql-codegen-typescript-mock-data`
```
3nvi, line 1: This is a bit of a "weird" looking code and it's a modified version of [graphql-codegen-typescript-mock-data](https://github.com/ardeois/graphql-codegen-typescript-mock-data)

web/__tests__/codegen/mockGraphqlResourcesPlugin.ts
```diff
@@ -0,0 +1,238 @@
...
+  schema,
+  documents,
+  config
+) => {
+  const printedSchema = printSchema(schema); // Returns a string representation of the schema
+  const astNode = parse(printedSchema); // Transforms the string into ASTNode
+
+  // List of types that are enums
+  const types: TypeItem[] = [];
+  const visitor: VisitorType = {
```
3nvi, line 141: This is a visitor where each key is a "visiting function" for the proper GraphQL entity while traversing the GraphQL tree

web/__tests__/codegen/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
```
3nvi, line 1: a TS configuration that transpiles TS -> JS. We have it JUST for the plugins, since we need to "emit" JS files that `graphql-codegen` can understand.

Our typical project-wide `tsconfig.json` has `emit: false` which means no JS files are created from the typescript compiler  running

web/__tests__/example.test.tsx
```diff
@@ -17,13 +17,34 @@
...
+import ListUsersPage, { mockListUsers } from 'Pages/Users';
 
 test('renders correct  breadcrumbs', async () => {
   const { container } = render(<Breadcrumbs />, { initialRoute: urls.compliance.policies.list() });
 
   expect(container).toHaveTextContent('Home');
   expect(container).toHaveTextContent('Policies');
 });
+
+test('renders a list of users in the users page', async () => {
```
3nvi, line 32: example use case of the PR's work

web/codegen.yml
```diff
@@ -49,6 +49,12 @@ generates:
         AWSTimestamp: number
         AWSDateTime: string
         AWSJSON: string
+
+  web/__tests__/__mocks__/builders.generated.ts:
+    plugins:
+      - web/__tests__/codegen/mockGraphqlResourcesPlugin.js:
```
3nvi, line 55: generates the `builders.generated.ts` using our `schema.tsx` for type-validation

vorillaz:
That's amazing @3nvi, the approach is neat and the mocking part looks modular and handy.
I am just wondering if `faker` is the correct approach in order to create predictable data. 
Since assertions in integration testing should be static I am convinced that we should provide static values as defaults, otherwise the assertions we are introducing side-effects. 
e.g

```tsx
export const usersDefaults = {
  givenName: 'Aragorn',
  familyName: 'Elrond',
  id: '1da08f5a-90e8-4f5a-8f81-9f077b3109a0',
  email: 'aragorn@lotr.com',
  createdAt: '2019-10-18T20:09:53.633Z',
  status: 'fighter',
};

export const buildUser = (overrides?: Partial<User>): User => {
  return {
    ...usersDefaults,
    ...overrides,
    __typename: 'User',
  };
};
```

One the call site, the consumers can either import the `usersDefaults` or they may use faker in order to produce their own mock data.
Once more kudos for that piece of poetry.

kostaspap:
> Solid work cant wait to write the first test!

... said no engineer ever :P

3nvi:
> That's amazing @3nvi, the approach is neat and the mocking part looks modular and handy.
> I am just wondering if `faker` is the correct approach in order to create predictable data.
> Since assertions in integration testing should be static I am convinced that we should provide static values as defaults, otherwise the assertions are introducing side-effects.
> e.g
> 
> ```tsx
> export const usersDefaults = {
>   givenName: 'Aragorn',
>   familyName: 'Elrond',
>   id: '1da08f5a-90e8-4f5a-8f81-9f077b3109a0',
>   email: 'aragorn@lotr.com',
>   createdAt: '2019-10-18T20:09:53.633Z',
>   status: 'fighter',
> };
> 
> export const buildUser = (overrides?: Partial<User>): User => {
>   return {
>     ...usersDefaults,
>     ...overrides,
>     __typename: 'User',
>   };
> };
> ```
> 
> On the call site, the consumers can either import the `usersDefaults` or they may use faker in order to produce their own mock data.
> Once more kudos for that piece of poetry.

Thanks a ton for the kind words.

You know what the  funny thing is? I initially had it exactly as you said, but I wanted to  have  dynamic data every time (since once in another app I had discovered an issue with `'`). Is there a benefit  on static data that  I'm not seeing?

3nvi:
> > That's amazing @3nvi, the approach is neat and the mocking part looks modular and handy.
> > I am just wondering if `faker` is the correct approach in order to create predictable data.
> > Since assertions in integration testing should be static I am convinced that we should provide static values as defaults, otherwise the assertions are introducing side-effects.
> > e.g
> > ```tsx
> > export const usersDefaults = {
> >   givenName: 'Aragorn',
> >   familyName: 'Elrond',
> >   id: '1da08f5a-90e8-4f5a-8f81-9f077b3109a0',
> >   email: 'aragorn@lotr.com',
> >   createdAt: '2019-10-18T20:09:53.633Z',
> >   status: 'fighter',
> > };
> > 
> > export const buildUser = (overrides?: Partial<User>): User => {
> >   return {
> >     ...usersDefaults,
> >     ...overrides,
> >     __typename: 'User',
> >   };
> > };
> > ```
> > 
> > 
> > On the call site, the consumers can either import the `usersDefaults` or they may use faker in order to produce their own mock data.
> > Once more kudos for that piece of poetry.
> 
> Thanks a ton for the kind words.
> 
> You know what the funny thing is? I initially had it exactly as you said, but I wanted to have dynamic data every time (since once in another app I had discovered an issue with `'`). Is there a benefit on static data that I'm not seeing?

After discussing offline, @vorillaz raised the concern of debugging. When something fails, it's much more harder  to replicate the test if the data changes on  every run (i.e. dynamic).

Having fixed fixtures helps us identify the test & the problem quicker. In addition it makes us more certain about failures, since if something fails it's never a false positive  (as it sometimes is with dynamic random data, creating an edge case value that shoouldn't matter).

For these reasons, I agreed to switch back into static generation. The PR got simplified  as well.

3nvi:
@austinbyers `mage fmt` produces no diffs locally, but CI says there is an issue with it. Any ideas?

austinbyers:
> @austinbyers mage fmt produces no diffs locally, but CI says there is an issue with it. Any ideas?

14:47:36	ERROR	1 file diffs after build:api + fmt:
 
  ~ web/__tests__/__mocks__/builders.generated.ts

Is this file not checked into the repo?

austinbyers:
You're right, I checked out this branch, `mage clean setup fmt`, no pending changes. Very strange

austinbyers:
Ah, because `mage test:ci` includes `mage build:api`, which calls `npm run graphql-codegen` which changes some dates in your file.

So I just ran `mage test:ci` locally and pushed the result

3nvi:
> Ah, because `mage test:ci` includes `mage build:api`, which calls `npm run graphql-codegen` which changes some dates in your file.
> 
> So I just ran `mage test:ci` locally and pushed the result

@austinbyers It still doesn't look fixed :D

austinbyers:
Yes, it's still failing on the same file. When I run `mage test:ci` locally, it works fine for me. Can you try checking out the branch now and running `mage test:ci`? Do you see any pending changes?

Did you change the graphql code generation? That file appears to be changing when mage runs `npm run graphql-codegen` I don't have enough context about that command to troubleshoot it

austinbyers:
In other words, as far as I can tell, this is a legitimate issue with the web code in this PR and not a random issue with mage

austinbyers:
Here's the full diff after running `npm run graphql-codegen` in CI:

```
 ~ web/__tests__/__mocks__/builders.generated.ts

<         : '1978-03-29',
---
>         : '1978-03-30',
291c273
<       overrides && overrides.hasOwnProperty('updateTime') ? overrides.updateTime! : '2009-11-02',
---
>       overrides && overrides.hasOwnProperty('updateTime') ? overrides.updateTime! : '2009-11-03',
801c783
<       overrides && overrides.hasOwnProperty('createdAt') ? overrides.createdAt! : '2011-09-26',
---
>       overrides && overrides.hasOwnProperty('createdAt') ? overrides.createdAt! : '2011-09-27',
805c787
<         : '2013-02-24',
---
>         : '2013-02-25',
1259c1241
<       overrides && overrides.hasOwnProperty('createdAt') ? overrides.createdAt! : '1970-10-15',
---
>       overrides && overrides.hasOwnProperty('createdAt') ? overrides.createdAt! : '1970-10-16',
1556c1538
<         : '1998-01-03',
---
>         : '1998-01-04',
```

You'll notice all of the dates are off by one, it looks like the generated file depends on what timezone you're in. This explains why you and I and CI all had slightly different versions of this file; we're all in different timezones

Again, I'm not sure how you're generating this file, but make sure the dates are fixed and not relative

austinbyers:
Notice my commit https://github.com/panther-labs/panther/pull/1158/commits/06500330acea820927f3b38162427a1e09054d65 above lowered the dates by one. We need to figure out a way to generate the same dates for everyone.

I added the diffing logic to the CI config, feel free to remove it once you've resolved :

```
       - run:
          name: Copy original problem file
          command: cp web/__tests__/__mocks__/builders.generated.ts builders.orig

      - run:
          name: npm graphql-codegen
          command: npm run graphql-codegen

      - run:
          name: check diff
          command: diff builders.orig web/__tests__/__mocks__/builders.generated.ts
```

The diff check will always fail because of the missing license header, don't worry about that. If the rest of the file is the same, you can remove that check and everything else should pass

3nvi:
> Notice my commit [0650033](https://github.com/panther-labs/panther/commit/06500330acea820927f3b38162427a1e09054d65) above lowered the dates by one. We need to figure out a way to generate the same dates for everyone.
> 
> I added the diffing logic to the CI config, feel free to remove it once you've resolved :
> 
> ```
>        - run:
>           name: Copy original problem file
>           command: cp web/__tests__/__mocks__/builders.generated.ts builders.orig
> 
>       - run:
>           name: npm graphql-codegen
>           command: npm run graphql-codegen
> 
>       - run:
>           name: check diff
>           command: diff builders.orig web/__tests__/__mocks__/builders.generated.ts
> ```
> 
> The diff check will always fail because of the missing license header, don't worry about that. If the rest of the file is the same, you can remove that check and everything else should pass

Holy shit. Couldn't see the diff. It's  because the seed for the random generator is TZ based. Sick bug.

PR #1159, opened 2020-07-14
Add extra validation for string parameters in master stack

austinbyers:
## Background

When using the pre-packaged master stack, the "FirstUserEmail" is the only required parameter. Our [quick start guide](https://docs.runpanther.io/quick-start#deployment) has a screenshot showing this parameter needs to be set, and I assumed CFN would fail if the parameter was not defined, which is how it behaves in the SDK.

I was wrong - if the parameter is not defined, the CloudFormation console defaults it to an empty string. If I had a dollar for every bug caused by empty strings vs non-existent ones, I'd now have at least $1, probably more! Anyway, in that case, the deployment continues until the very last step, when inviting the first user fails with a blank email. Then all 9 stacks have to rollback, a total of 20 minutes deployed and rolled back.

It's easy to forget this parameter, and in fact it happened today with a customer. We should fail much faster in this case. Turns out, there is an easy way to add protection - by enforcing a minimum string length.

## Changes

- Enforce a minimum string length for applicable parameters in the master template

## Testing

- Tried to deploy master template from CloudFormation console without changing any default parameters:

Screen Shot 2020-07-13 at 5 44 22 PM


deployments/master.yml
```diff
@@ -73,14 +74,17 @@ Parameters:
   FirstUserEmail:
     Type: String
     Description: Initial Panther user - email address
+    MinLength: 5 # a@b.co
```
jacknagz, line 77: You can actually specify a regular expression, which would be better here:

`AllowedPattern`

jacknagz: This regex should work: `^([a-zA-Z0-9_\-\.]+)@([a-zA-Z0-9_\-\.]+)\.([a-zA-Z]{2,5})$`

austinbyers: Ah, sure enough! Ok, will do.

Fun fact - it's impossible to have a generic email regex, but [this one](https://emailregex.com/) is the closest to the RFC:


```
(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])
```

We'll go with something simpler, I think :)

austinbyers: > This regex should work: ^([a-zA-Z0-9_\-\.]+)@([a-zA-Z0-9_\-\.]+)\.([a-zA-Z]{2,5})$

Haha nope - I commonly use `user+test@runpanther.io`, so plus signs are definitely allowed before the `@` sign. And there are domains longer than 5 characters.

So you begin to see the complexities involved. We'll get something good enough, probably

`^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}$` from https://www.regular-expressions.info/email.html

austinbyers:
Tested again with the email pattern:

Screen Shot 2020-07-14 at 11 07 17 AM

Standard emails, including with `+`, worked as expected

PR #1164, opened 2020-07-14
Remove npm audit from CI

austinbyers:
## Background

A potentially controversial suggestion - let's remove `npm audit` from CI entirely.

`npm audit` will start failing as soon as any vulnerable package is discovered. While we do want to know about that as soon as possible, there's usually nothing we can do about it until days or weeks later when a patch is released. The failure is essentially never related to a pull request, yet we run the check on every PR.

The main downside, which @jacknagz identified again, is that all of our builds show as failing and it looks bad. The README and CircleCI show Panther as failing all builds:

Screen Shot 2020-07-14 at 12 19 10 PM

And every PR will forever have the red X:

Screen Shot 2020-07-14 at 12 22 57 PM

A check like this makes make more sense to run periodically, for example as part of every release cycle

## Changes

- Remove `npm audit` from CI config

## Testing

- CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,17 +68,8 @@ jobs:
           name: Run test suite
           command: mage test:ci
 
-  npm_audit:
-    executor: panther-buildpack
-    steps:
-      - checkout
-      - run:
-          name: Audit NPM packages
-          command: npm audit --production # does not require installing packages first
-
 workflows:
   version: 2
   pipeline:
     jobs:
       - mage_test_ci
-      - npm_audit


austinbyers:
Put another way: having a check which is always failing and always ignored is about as helpful as not running the check at all

austinbyers:
It's also worth noting some vulnerabilities may not be applicable to us, and can be safely ignored. No need to show the whole application as failing to build for that reason

3nvi:
@austinbyers I'm not in favor of this. I think @kostaspap [disagrees as well](https://panther-labs.slack.com/archives/C013XKQU3BR/p1594712211018600?thread_ts=1594682024.012200&cid=C013XKQU3BR)

Let's please discuss/agree internally before moving on to a formal PR

3nvi:
> Put another way: having a check which is always failing and always ignored is about as helpful as not running the check at all

As stated it's not ignored, it has [an internal deadline](https://panther-labs.slack.com/archives/C013XKQU3BR/p1594704882018200?thread_ts=1594682024.012200&cid=C013XKQU3BR) for switching the vulnerable package

3nvi:
> It's also worth noting some vulnerabilities may not be applicable to us, and can be safely ignored. No need to show the whole application as failing to build for that reason

If you view it as an actual security vulnerability, I feel that users should be able to know which packages have issues and which not. Hiding it while "internally" knowing that  we have vulnerabilities is not the level of transparency/visibility that I'd personally like as a user.

In general, `npm audit` being part of the CI is a pretty typical step on **a lot** of products, both open-sourced and internal.

kostaspap:
If this check is removed, how are we going to be notified for vulnerable packages? 
The current check was never ignored, there has been a lengthy discussion with a deadline to fix it.

austinbyers:
I missed some of that internal discussion, I just wanted to have something public we could refer to when people ask why our builds are failing.

We don't do automatic checks for vulnerabilities in other languages (python / go) on every PR - Github itself has vulnerability detection and will notify us / put up PRs to fix it. We also review python/go library versions as part of every release - I was merely suggesting we could do the same with the npm libraries.

But since it it's the web community standard to run `npm audit` in CI, then we can do the same

PR #1165, opened 2020-07-14
Configurable ALB ingress security group

austinbyers:
## Background

The ALB security group which allows HTTP(S) access to the web app (ALB) is currently 0.0.0.0/0

Not everyone wants an open security group like this, we need to provide a configuration option for the ingress cidr block.

## Changes

- Add `LoadBalancerSecurityGroupCidr` parameter to `bootstrap` and `master` stacks

## Testing

- `mage test:ci`
- `STACK=bootstrap mage deploy`

### DIFF
--- a/deployments/bootstrap.yml
+++ b/deployments/bootstrap.yml
@@ -59,6 +59,11 @@ Parameters:
     Type: String
     Description: Enable S3 access logging for all Panther buckets.
     AllowedValues: [true, false]
+  LoadBalancerSecurityGroupCidr:
+    Type: String
+    Description: Allow HTTP(S) ingress access to the web app (ALB) security group from this IP block. Use 0.0.0.0/0 to allow unrestricted access
+    # cfn-lint suggested this regex pattern:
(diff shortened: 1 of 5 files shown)


austinbyers:
> Definitely an improvement, for larger organizations it would be nice to specify a list of IP blocks. Is this simple to add here (with e.g. a CSV split), or would that add a lot of complexity? If simple I'd say it's worthwhile, if not then its not a big deal.

Good question - it's not simple, unfortunately. The [ingress resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-ingress.html) requires each ip to be a separate element in a list. You can't just pass a list of cidr blocks, for example (I tried).

And [CloudFormation parameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html) only allow a CSV list of strings or a list of numbers, you can't pass a list of objects.

Finally, there is no [intrinsic function](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html) which can loop over a list to expand it.

So, there is no way to go from a list of IP cidrs to a list of objects attached to the ingress rule without using a custom resource or a CloudFormation macro. Neither solution would work in the `bootstrap` stack, so we'd need to move some networking resources around to make it happen.

Certainly possible for the future, though

PR #1176, opened 2020-07-15
Feat: Bundle Size CI check

3nvi:
## Background

Recently,  we saw that our web app size of JS went up, because something escaped in one of the PRs which caused us to import the entire library instead of only what we wanted. This is now fixed, but it could easily happen again (we are only humans). In the spirit of reducing manual effort, as well as our company value:

>**Find and Improve**
Leave the world better than how you found it.

I added a CI check to automatically check the size of our bundles for us. Specifically it will make sure we set budgets on:

- JS bundle sizes
- SVG Illustration size
- General Image elements 

## Changes

- Add `bundlesize` CI check
- Configure fixed budgets for assets (we can tweat them as necessary depending on project needs)
- add `npm run build:analysis` to allow FE devs to debug locally any size-related issues

## Testing

- CI check

## Screenshots

To prove its functionality, this is the output before the buggy import got fixed

Screen Shot 2020-07-15 at 12 07 20 PM

and that's the output after it got fixed

Screen Shot 2020-07-15 at 12 08 18 PM

## Notes

- This check compares the built bundles (including our custom/source code) and not just the external dependencies

- This check will run on **every PR** regardless of whether JS was added or not, since it's almost impossible to make a CircleCI job run only when certain files change


.circleci/config.yml
```diff
@@ -76,9 +76,27 @@ jobs:
           name: Audit NPM packages
           command: npm audit --production # does not require installing packages first
 
+  bundlesize:
+    executor: panther-buildpack
+    steps:
+      - checkout
+      - restore_cache:
```
3nvi, line 83: @austinbyers not calling `save_cache`, since the `mage_test_ci` will already do it for us. No need to do it twice

.circleci/config.yml
```diff
@@ -76,9 +76,27 @@ jobs:
...
+          keys:
+            - npm-{{ checksum "package-lock.json" }}
+
+      - run:
+          name: Install NPM dependencies
+          command: npm i
+
+      - run:
+          name: Check bundle size
+          command: npm run bundlesize
```
austinbyers, line 94: You feel this should be separate and not part of `mage test:ci`? PRs can merge even if this fails unless we add it as a new required check

Generally, we want all tests to be in `mage test:ci` so that developers can verify locally the exact same tests running in CI. `npm audit` is only an exception because we don't require it to pass, will this be the same?

3nvi: Yeah it won't be a required check. But we should aim for it to be passing

3nvi: I don't want it to be part of mage CI since I want it to be:

* optional
* explicit  about what part of CI is failing without going into CircleCI

PR #1178, opened 2020-07-15
Move log group deletion to custom teardown resource

austinbyers:
## Background

Every time a user deletes their master Panther stack, there will be leftover log groups which break all future deployments. Developers also encounter this issue fairly often, it's quite frustrating. You can run `mage teardown` to fix it, or delete the log groups manually, but it'd be nice if every rollback / stack delete just took care of this.

One solution could have been to create the log groups with a custom resource which doesn't fail if the group already exists. Unfortunately, we can't easily make that migration now because you can't change the type of an existing CloudFormation resource and we don't want to delete the existing log groups.

Why are there leftover log groups? CloudFormation deletes them, but Lambda often recreates them from still-running functions or buffered logs. We can have the `PantherTeardown` custom resource automatically remove leftover groups, but it is particularly difficult to delete the `custom-resources` log group since our teardown code is running in the function whose very logs we're trying to delete!

To get around this, the custom resources Lambda function modifies its own IAM role to block future logs right before it stops running and CloudFormation deletes the log group.

## Changes

- Delete leftover `/aws/lambda/panther-*` log groups as the very last step of the custom `PantherTeardown` resource
- Fixes lambda alarm deletion (closes: #1208 )
- Fixes step function alarm deletion (closes: #1209 )

## Testing

- Lots of deploys and teardowns - no log groups nor alarms left over


deployments/bootstrap_gateway.yml
```diff
@@ -183,6 +184,12 @@ Resources:
             - Effect: Allow
               Action: iam:PassRole
               Resource: !Sub arn:aws:iam::${AWS::AccountId}:role/Panther*
+            - Effect: Allow
+              # This function needs to be able to modify its own role policy for a clean teardown.
+              # We don't know the role name, which is randomly generated, but we know it will contain
+              # at least part of the CloudFormation resource name.
+              Action: iam:PutRolePolicy
+              Resource: !Sub arn:aws:iam::${AWS::AccountId}:role/*CustomResourceFunction*
```
nhakmiller, line 192: This permission scares me a bit. `PutRolePolicy` is sort of the quintessential privilege escalation permission when it comes to IAM, being able to modify your own permissions it especially scary.

Would we instead be able to use IAM permissions boundaries here? Much less chance for abuse and still should cover the same needs. Instead of a `PutRolePolicy` API call that denies CloudWatch permission, you would call `PutRolePermissionsBoundary` that denies CloudWatch.

The primary difference being that permissions boundaries cannot grant permissions, only make it "legal" for other things (such as the role policy) to grant permissions. I feel like this would be much safer, especially considering how much the custom resource function is doing on every update.

rleighton:
Q: Would it be simpler to _force_ (via dependencies) a custom resource to run last, which lists the log groups and delete all that start with `panther-`? ... tags would be better but I don't see in the console that log groups can be tagged.

nhakmiller:
> Q: Would it be simpler to _force_ (via dependencies) a custom resource to run last, which lists the log groups and deletes all that start with `panther-`? ... tags would be better but I don't see in the console that log groups can be tagged.

From my understanding we already do that, but custom resources are actually just lambda functions so while that works for the majority of log groups, it doesn't work specifically for the custom resource's own log group (if it deletes its own log group, the log group will be re-created after it exits to log the final information etc). The workaround here is to still do that exactly that, except also prevent itself from logging after it exits which I think is quite clever. @austinbyers can correct me if I'm wrong here.

deployments/bootstrap_gateway.yml
austinbyers: Good idea, I'll look into the permissions boundary. As long as it is attached to the role and automatically removed during a teardown, then we should be good to use it.

Yes, `UpdateRole` allows theoretical privilege escalation

austinbyers:
>> Q: Would it be simpler to force (via dependencies) a custom resource to run last, which lists the log groups and deletes all that start with panther-? ... tags would be better but I don't see in the console that log groups can be tagged.

> From my understanding we already do that, but custom resources are actually just lambda functions so while that works for the majority of log groups, it doesn't work specifically for the custom resource's own log group (if it deletes its own log group, the log group will be re-created after it exits to log the final information etc). The workaround here is to still do that exactly that, except also prevent just the custom resource from logging after it exits which I think is quite clever. @austinbyers can correct me if I'm wrong here.

@nhakmiller nailed it, as always. This is the very last custom resource, and it does list all panther log groups and delete them, as you suggest. But if it deleted its *own* log group without extra protections, it would be recreated as soon as the custom resource function finished running, because Lambda would write the "END request" CloudWatch log.

You're also right CloudWatch logs do not support group tagging, so we have to either list them by prefix or pre-compile the list of all functions, which would of course be a struggle to keep up to date

rleighton:
> > > Q: Would it be simpler to force (via dependencies) a custom resource to run last, which lists the log groups and deletes all that start with panther-? ... tags would be better but I don't see in the console that log groups can be tagged.
> 
> > From my understanding we already do that, but custom resources are actually just lambda functions so while that works for the majority of log groups, it doesn't work specifically for the custom resource's own log group (if it deletes its own log group, the log group will be re-created after it exits to log the final information etc). The workaround here is to still do that exactly that, except also prevent just the custom resource from logging after it exits which I think is quite clever. @austinbyers can correct me if I'm wrong here.
> 
> @nhakmiller nailed it, as always. This is the very last custom resource, and it does list all panther log groups and delete them, as you suggest. But if it deleted its _own_ log group without extra protections, it would be recreated as soon as the custom resource function finished running, because Lambda would write the "END request" CloudWatch log.
> 
> You're also right CloudWatch logs do not support group tagging, so we have to either list them by prefix or pre-compile the list of all functions, which would of course be a struggle to keep up to date

Ugh. Infinite regress. I hate changing permissions like this but it is a clever solution in a little bit of a scary way.

deployments/bootstrap_gateway.yml
austinbyers: Great suggestion, I was able to change to `PutRolePermissionsBoundary`

austinbyers:
@nhakmiller @rleighton ready for re-review - I used a permission boundary instead of changing the actual permissions and also fixed a few bugs with cloudwatch alarm deletion

internal/core/custom_resources/resources/panther_teardown.go
```diff
@@ -96,3 +105,65 @@ func destroyLambdaLayers() error {
...
+
+	zap.L().Info("blocking future log writes")
+	_, err = iamClient.PutRolePermissionsBoundary(&iam.PutRolePermissionsBoundaryInput{
+		PermissionsBoundary: aws.String("arn:aws:iam::aws:policy/AWSDenyAll"),
+		RoleName:            &roleName,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to add iam permission boundary to self: %v", err)
+	}
+	return nil
```
nhakmiller, line 154: I'm guessing you don't have to explicitly delete the last log group because CloudFormation does that for you after this function exits?

austinbyers: Right, there is a comment to that effect in the CloudFormation, but I can put one here as well

austinbyers:
Did a fresh deploy + teardown, the `cfn-custom-resources` log group was still left behind. The logs show other custom alarms being deleted - we need to make sure `CustomTeardown` is the very last custom resource in the stack, working on that now

austinbyers:
Actually, further testing shows the permissions boundary is not working as expected, I'll keep playing with it for a bit

austinbyers:
There we go! Works now, just needed to sleep for a few seconds at the very end

nhakmiller:
> just needed to sleep for a few seconds at the very end

I think we can all relate to that.

PR #1180, opened 2020-07-16
Metrics API Rework

nhakmiller:
## Background

After the initial metrics API PR merged, @alexmylonas had some time to play around with it and provided some feedback on how the API should work. We discussed this feedback and agreed on the following changes:

- Fill in zero values. When CloudWatch is queried for data in a given time period, it will break the time period into intervals and return a value for each interval that had metrics in it. However, if an interval has no values, the interval is omitted entirely. So if a given time frame has three intervals based on the start time, end time, and period (say `[t1, t2, t3]`), but there is only data available in two of those time periods (say `t1` and `t3`), then CloudWatch (and the previous version of this API) would only return those values. We would return two slices, `Timestamps` with `[t1, t3]` and `Values` with `[v1, v3]`. If you try to graph these data points you get a straight line from point `t1,v1` to point `t3,v3`, when in reality there was an interval starting at `t2` where the value was 0. Imagine if the `eventsProcessed` metric just never showed a zero value, you would think you were constantly processing logs when in reality there may be quite large gaps where no logs are processed. To resolve this, we calculate what timestamps SHOULD be present, then for any missing timestamps fill in a zero value.
- Encode timestamps only once. This is related to the above point. Previously, each metric dimension might only return some subset of the total possible number of values/timestamps because CloudWatch omitted the empty values. To represent this, we returned a list of values and a corresponding list of timestamps for each `SeriesData` object within a metric. Now, since we know all the metric dimensions will have the exact same number of timestamps, we store one single list of timestamps with the metric object and then only the values for each dimension.
- Return metrics as named objects. Previously we returned a list of metric objects,  where each object had a metric name and then the values. Now each metric that can be requested will be returned as a top level key with the same name as the metric and containing the values of the metric.

In addition to these API wide changes, we added support for two new metrics.

- `alertsBySeverity`, which returns time series data representing how many log analysis alerts were fired for each severity during a given time
- `totalAlertsDelta`, which returns two single value metrics, `Current Period` which indicates how many alerts happened in the specified time frame, and `Previous Period` which indicates how many alerts happened in the timeframe of identical length to the specified time frame but immediately before it.

To support these two new metrics, I added some metric generation to the `panther-log-alert-forwarder` lambda function.

## Changes

- Added support to the metrics API for `alertsBySeverity` and `totalAlertsDelta` metrics
  - Added metric generation to the `panther-log-alert-forwarder` to support these metrics
- Before returning time series data, the values are "normalized" so that all metric dimensions for a given time period have the exact same number of values, filling in zeros for missing values.
- Metric results are returned as top level named objects, as opposed to members of a list
- Timestamps are only encoded once per metric, not once per dimension per metric
- Made some unrelated changes to the `panther-alert-deliery` lambda to avoid shadowing package names
- Changed `IntervalHours`  to `IntervalMinutes` to allow for finer granularity in  metric generation
- Removed the `Component` dimension from all metrics

## Testing

- Added new unit tests for all the timestamp processing work that needs to be done now
- Tested the old API and both new APIs in the AWS Lambda console (see below for queries)
- Updated the integration test to expect the new format

## AWS Lambda Console Tests

Input:
```
{
  "getMetrics": {
    "metricNames": [
      "eventsProcessed"
    ],
    "fromDate": "2020-07-10T01:45:00Z",
    "toDate": "2020-07-10T04:00:00Z",
    "intervalMinutes": 60
  }
}
```

Response:
```
{
  "eventsProcessed": {
    "seriesData": {
      "timestamps": ["2020-07-10T03:45:00Z", "2020-07-10T02:45:00Z", "2020-07-10T01:45:00Z"],
      "series": [
        {
          "label": "AWS.VPCFlow",
          "values": [1404, 1433, 1441]
        },
        {
          "label": "AWS.S3ServerAccess",
          "values": [10445, 10030, 10658]
        },
        {
          "label": "AWS.ALB",
          "values": [0, 4, 21]
        }]}
  },
  "fromDate": "2020-07-10T01:45:00Z",
  "toDate": "2020-07-10T04:00:00Z",
  "intervalMinutes": 60
}
```

Input:
```
{
  "getMetrics": {
    "metricNames": [
      "totalAlertsDelta",
      "alertsBySeverity"
    ],
    "fromDate": "2020-07-16T15:00:00Z",
    "toDate": "2020-07-16T16:20:00Z",
    "intervalMinutes": 60
  }
}
```

Output:
```
{
  "totalAlertsDelta": {
    "singleValue": [
      {
        "label": "Current Period",
        "value": 199
      },
      {
        "label": "Previous Period",
        "value": 179
      }
    ],
    "seriesData": {
      "timestamps": [],
      "series": []
    }
  },
  "alertsBySeverity": {
    "seriesData": {
      "timestamps": ["2020-07-16T16:00:00Z", "2020-07-16T15:00:00Z"],
      "series": [
        {
          "label": "CRITICAL",
          "values": [3, 2]
        },
        {
          "label": "HIGH",
          "values": [100, 100]
        },
        {
          "label": "MEDIUM",
          "values": [88, 92]
        },
        {
          "label": "LOW",
          "values": [0, 0]
        },
        {
          "label": "INFO",
          "values": [0, 0]
        }]}
  },
  "fromDate": "2020-07-16T15:00:00Z",
  "toDate": "2020-07-16T16:20:00Z",
  "intervalMinutes": 60
}
```


api/lambda/metrics/models/api.go
```diff
@@ -40,28 +40,33 @@ type GetMetricsInput struct {
 
 // GetMetricsOutput contains data points for a number of metrics over the specified time frame
 type GetMetricsOutput struct {
-	MetricResults []MetricResult `json:"metricResults"`
-	FromDate      time.Time      `json:"fromDate"`
-	ToDate        time.Time      `json:"toDate"`
-	IntervalHours int64          `json:"intervalHours"`
+	EventsProcessed  *MetricResult `json:"eventsProcessed,omitempty"`
+	TotalAlertsDelta *MetricResult `json:"totalAlertsDelta,omitempty"`
+	AlertsBySeverity *MetricResult `json:"alertsBySeverity,omitempty"`
```
nhakmiller, line 45: Per feedback from @alexmylonas, this list of results was removed and now each possible result is a separate top level struct.

internal/core/alert_delivery/delivery/dispatch.go
```diff
@@ -97,7 +97,7 @@ func send(alert *alertmodels.Alert, output *outputmodels.AlertOutput, statusChan
 //
 // Returns true if the alert was sent successfully, false if it needs to be retried.
 func dispatch(alert *alertmodels.Alert) bool {
-	outputs, err := getAlertOutputs(alert)
+	alertOutputs, err := getAlertOutputs(alert)
```
nhakmiller, line 100: While I was looking in this function to see if this was the appropriate place to generate metrics (I determined it was not), my IDE warned me that the variable `outputs` was shadowing the package `outputs`.

internal/core/alert_delivery/main/lambda.go
```diff
@@ -41,7 +41,7 @@ func lambdaHandler(ctx context.Context, event events.SQSEvent) (err error) {
 	var alerts []*models.Alert
 
 	lc, _ := lambdalogger.ConfigureGlobal(ctx, nil)
-	operation := oplog.NewManager("cloudsec", "alert_delivery").Start(lc.InvokedFunctionArn).WithMemUsed(lambdacontext.MemoryLimitInMB)
+	operation := oplog.NewManager("core", "alert_delivery").Start(lc.InvokedFunctionArn).WithMemUsed(lambdacontext.MemoryLimitInMB)
```
nhakmiller, line 44: This is function is created in the `core` stack,  and should therefore be logging as such.

internal/core/metrics_api/api/alert_metrics.go
```diff
@@ -0,0 +1,152 @@
+package api
```
nhakmiller, line 1: I moved what I've been calling the "metric resolvers" to separate files as the `get_metrics.go` file was getting quite crowded. In `get_metrics.go` I left only the generally applicable functions.

internal/core/metrics_api/api/get_metrics.go
```diff
@@ -59,90 +63,130 @@ func (API) GetMetrics(input *models.GetMetricsInput) (*models.GetMetricsOutput,
...
 	}
 
-	for i, metricName := range input.MetricNames {
+	for _, metricName := range input.MetricNames {
 		resolver, ok := metricResolvers[metricName]
 		if !ok {
 			return nil, &genericapi.InvalidInputError{Message: "unexpected metric [" + metricName + "] requested"}
 		}
-		metricData, err := resolver(input)
+		err := resolver(input, response)
```
nhakmiller, line 71: Since we're no longer building a list of metric results, we cannot just append each result somewhere. Instead of adding a big switch statement here,  I just pass in the response as a pointer and it is the responsibility of the resolver to update the appropriate field.

If we decide to multithread the metric requests (which is probably a good idea), this should still be safe I think since each resolver will be updating a different field in the response.

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -112,8 +113,14 @@ func storeNewAlert(rule *models.Rule, alertDedup *AlertDedupEvent) error {
...
 	_, err = ddbClient.PutItem(putItemRequest)
 	if err != nil {
-		return errors.Wrap(err, "failed to update store alert")
+		return errors.Wrap(err, "failed to store alert")
 	}
+
+	staticLogger.LogSingle(1,
+		metrics.Dimension{Name: "Severity", Value: string(rule.Severity)},
+		componentDimension,
+	)
```
nhakmiller, line 122: This is a case where some of the benefits @alxarch has referenced for aggregating metrics via a proper metrics library would be beneficial. Currently this function does not do any aggregation of results, so I just log a metric with a value of 1 for every alert created. The application logic could have been updated to aggregate some amount of metrics before logging, but updating application logic was something we specifically wanted to avoid while implementing the metric generation.

This should be fine here as this function is generally under fairly light load. If we see this pattern arising often, it may be a good motivation to incorporate a more mature metric aggregation library as we've previously discussed.

internal/core/metrics_api/api/alert_metrics.go
```diff
@@ -0,0 +1,152 @@
...
+		FromDate:      dStart,
+		ToDate:        input.ToDate,
+		IntervalHours: int64(math.Abs(timeFrame.Hours())),
+	}
+	metricData, err := getMetricData(dInput, queries)
+	if err != nil {
+		return err
+	}
+
+	values, _ := normalizeTimeStamps(dInput, metricData)
```
nhakmiller, line 137: Instead of generating & normalizing time series data, I could have created two separate queries above with different start & end dates and avoided having to normalize here. But a lot of the logic needed to do that is sort of already covered in the timestamp normalization, so I figured it would be just as easy like this,

internal/core/metrics_api/api/alert_metrics.go
```diff
@@ -0,0 +1,151 @@
...
+	"go.uber.org/zap"
+
+	analysismodels "github.com/panther-labs/panther/api/gateway/analysis/models"
+	"github.com/panther-labs/panther/api/lambda/metrics/models"
+)
+
+// getAlertsBySeverity returns the count of log analysis alerts generated by severity
+//
+// This is a time series metric.
+func getAlertsBySeverity(input *models.GetMetricsInput, output *models.GetMetricsOutput) error {
```
kostaspap, line 36: This is a bit... scary. Do we have to create a new method for every new metric that we need to publish to the frontend? It would be much better and easy to extend if the `metrics_api` was a thin layer between frontend and Cloudwatch.
The only thing we should need to add a new metric would be: Update backend code to emit it, FE makes the appropriate call to `panther-metrics-api`. Having to build custom logic for every metric we are planning to emit will be cumbersome.

internal/core/metrics_api/api/alert_metrics.go
```diff
@@ -0,0 +1,151 @@
...
+	for i, severity := range analysisSeverities {
+		queries = append(queries, &cloudwatch.MetricDataQuery{
+			Id:    aws.String("query" + strconv.Itoa(i)),
+			Label: severity,
+			MetricStat: &cloudwatch.MetricStat{
+				Metric: &cloudwatch.Metric{
+					Dimensions: []*cloudwatch.Dimension{
+						{
+							Name:  aws.String("Component"),
+							Value: aws.String("AlertForwarder"),
```
kostaspap, line 56: Would it make sense to just remove the "Component" dimension from our metrics? It seems to add more trouble than actually help. 
Just having a metric named "alerts" with dimension "policy"/"rule" I think makes more sense: 
we could get all of the alerts, or if we provided a dimension we could get the alerts for policies or rules. 
Having a dimension per "component" doesn't seem that useful - we could have it, but it could be optional and not used by our `metric-api`. 
Backend components might be renamed, shift and change. 

In other words - I think the metrics-api should just care about the "business logic" of the metrics and care less about the backend architecture.

internal/core/metrics_api/api/get_metrics.go
```diff
@@ -59,90 +63,130 @@ func (API) GetMetrics(input *models.GetMetricsInput) (*models.GetMetricsOutput,
...
+				Label:  metricData.Label,
+				Values: metricData.Values,
+			}
+			continue
+		}
+
+		// In some cases, an interval will have no value. AWS just omits these intervals from the
+		// results, but most systems will not implicitly understand an omitted interval to mean zero
+		// activity, so we fill in a zero value.
+		fullValues := make([]*float64, len(times))
```
kostaspap, line 123: Does this have to be `[]*float64` or would it be easier if it was `[]float64`?

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
kostaspap, line 122: First we store, then we send the notification. I think it's better to emit the metric when both have completed successfully -> I think you should log the metric after line 62, because if e.g. sending the notification fails, this code will be retried and we will emit metric for another time.

internal/core/metrics_api/api/alert_metrics.go
nhakmiller: I don't feel strongly either way on the `component` dimension. Currently it's not being used for anything so I have no problem removing it. Good call on the `AnalysisType` dimension, I'll add that.

internal/core/metrics_api/api/get_metrics.go
nhakmiller: I think it's easier to have `[]*float64` since that is the format AWS returns it in. We'd have to go through and de-reference each value otherwise.

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
nhakmiller: Makes sense.

internal/core/metrics_api/api/alert_metrics.go
kostaspap: My suggestion is basically not doing what is mentioned here: https://github.com/panther-labs/panther/pull/1180#discussion_r456050368

The development effort of adding and display a new metric is with this PR the following: 
1. BE emits the metric
2. BE adds a new method and translation layer that converts the metric data to specialized structs (e.g. `AlertsBySeverity`)
3. FE needs to query and display the data

I believe what we had prior (and what I believe simplifies things) was:
1. BE emits the metric
2. FE needs to query and display the data. 

But probably i am missing some background here as why we needed to have the specialized structs. As I see it, the FE makes a request with a list like `[]{start, end time, metric name, dimensions,...}` and the backend will just return a response that contains a list of results. Each element in the list of results maps one to one to the requested metrics, so it can render them appropriately. What do we need another translation layer? Can't the FE just map the requested metric to the metric values in the response?

kostaspap: I guess my suggestion would work if the input was something like: 
```
type GetMetricsInput struct {
	Requests []GetMetricRequest   
}

type GetMetricRequest struct{
	MetricName string
	Dimensions []string
	FromDate time.Time
	ToDate time.Time
	IntervalHours int64
}
```

nhakmiller: I am also not a big fan of the specialized struct responses (although if you notice they all have the same type). Originally I wanted to use a `map[string]MetricResult`, where the string was the name of the metric. I thought this would be very natural to use, but Austin informed me that maps are very difficult to return with AppSync so I should avoid them if possible.

Then I switched to a `[]MetricResult` as you suggest to address the map concern, but Alex said that it was strange to work with and having the values be named would be better.

Since a map doesn't appear to be an option, I went with the structs. I would actually be interested in hearing from @3nvi or @alexmylonas about whether returning a map would cause that much issue, it seems like it would be much nicer for the backend than having to do custom structs and much nicer for the frontend than having a list.

nhakmiller: That being said, the custom functions (e.g. `getAlertsBySeverity`) that build the actual queries are necessary. They could live in either the frontend or the backend, but they are necessary somewhere in order to properly query the data.

kostaspap: I think we shouldn't really need it...
The FE can just make a request with start/end time, interval, dimensions it wants. The BE will just returns those results. If the FE wants metrics for all the dimensions, it could make a request for this...

internal/core/metrics_api/api/get_metrics.go
```diff
@@ -59,90 +63,130 @@ func (API) GetMetrics(input *models.GetMetricsInput) (*models.GetMetricsOutput,
...
+		for j, k := 0, 0; j < len(times); j++ {
+			// If the k'th value occurred during the j'th time, keep it and increment k.
+			if k < len(metricData.Values) && *times[j] == *metricData.Timestamps[k] {
```
rleighton, line 126: Can you explain why you are sure that `times[j] == *metricData.Timestamps[k]` ? The reason I ask is that `times` is created by this code by chopping up a time interval by hour. How do you know this will align with   `metricData.Timestamps[k]` ? Is it that you always ask CW for at  hour size bin? If so, can you add that as a comment?

rleighton: Also, if hour is only bin size, can you explain why that was picked? Seems limiting, certainly in the API (could we have the API support different intervals but only _initially_ implement hour?).

nhakmiller: Good question, this part of the code is really important and sadly GitHub sort of butchered the formatting above where we generate `times` so it's hard to read in this PR.

`times` is created by chopping up a time interval based on the request's parameters. When someone requests a metric, they must specify a start date, end date, and an interval length. We use that interval length above to construct the list  of timestamps. We then pass the exact same start, stop,  and interval time to CloudWatch when making the metrics request.

So we don't always ask CW for  an hour size bin, but we do always ask CW for a size bin of the same size as the one we use to construct `times`. I can add some more comments to this function since some of the logic might be unintuitive, and is based around how CW does things.

rleighton: Sorry, I'm confused. I see `intervals := int(math.Ceil(delta.Hours() / float64(input.IntervalHours)))` which looks like an hour bin size, right?

nhakmiller: > Also, if hour is only bin size, can you explain why that was picked? Seems limiting, certainly in the API (could we have the API support different intervals but only _initially_ implement hour?).

Hourly is the minimum bin size we support, with no maximum bin size. @austinbyers had asked if we should support smaller bin sizes as well and I forgot to answer, but basically Alex and I discussed it and figured the minimum length would usually be about a half a day or so, so hourly time bins made sense as a minimum. We didn't want people to ask for minute time bins on a years worth of data, which is why we put the limitation in at all.

rleighton: Can I ask for an hour time range with 5min bins?

nhakmiller: > Sorry, I'm confused. I see `intervals := int(math.Ceil(delta.Hours() / float64(input.IntervalHours)))` which looks like an _multiple_ of hour bin size, right?

Correct, each bin size is 1 or more hours based on the what the requester specified for the `IntervalHours` parameter.

nhakmiller: > Can I ask for an hour time range with 5min bins?

Right now we don't, the minimum bin size is 1 hour. But it's entirely an artificial limitation imposed by us, if we wanted to we could support any size bins (although sub minute sized bins get weird and aren't supported on older data, which we'd have to account for). Do you think users would want five minute time bins? I can look into it, shouldn't be too hard.

rleighton: Most troubleshooting will be "today" (<24 hour/data points). Hourly is not good at all.

For trending over a week or sth then hourly is fine.

I think we need to change the API to allow (in the future) smaller than hour bins. This just mean making the units for the interval minutes (for example) but only allowing multiples of hours OR using symbols the interval. I would vote for discreet symbols. I would change `"intervalHours": <int>` to `"interval": <"hour","day">` THEN if we want to add new intervals there are very few changes to the UI. The UI would have a drop down of the intervals.

nhakmiller: The main complication I was trying to avoid was that for older metrics CW starts aggregating the metrics so the minimum bucket size becomes larger. For data older than 63 days, the minimum bucket size is one hour. For data older than 15 days, the minimum bucket size is 5 minutes. For data less than 15 days old, the minimum bucket size is 1 minute. So to avoid having to handle those situations separately I just enforced a flat 1 hour minimum interval across all data.

But we have to address those time intervals anyways for calculating the starting time point, so might as well add this in. I'll change the interval to minutes for now since that will be a simple change, then in the  future we can consider the symbols change maybe if it makes more sense for the UI.

rleighton: Cool. Just wanted to make it a little easer when we get asked to make those changes, because we know we will.

nhakmiller:
Updated integration tests so they are passing now and did some more manual testing. Going to merge this now so the frontend team can wrap up their changes early next week.

internal/core/metrics_api/api/alert_metrics.go
nhakmiller: I will merge for now for the 1.6 release, but I think it bears more discussion. Perhaps we can discuss on Monday with Alex as well.

Fundamentally though someone has to do it. When you say this:

> The FE can just make a request with start/end time, interval, dimensions ("Severity" + value) it wants

You're saying  "we can take what this backend code is currently doing and move it into the frontend to make it their responsibility", and I agree we can do that. My question is, should we do that? I'm not convinced that we should, but I'm open to hear arguments either way.

PR #1181, opened 2020-07-16
Use a composite alarm over all the primitive lambda alarms to reduce pages

rleighton:
# Background
The lambda alarms are often correlated. For example, a lambda can fail before logging is enabled, so we alarm on that.
We also alarm when a lambda runs, logs an informative error and fails. If paging is tied to both alarms and a lambda runs, logs and error and fails the 2 pages are generated (not good). Both alarms are needed because the lambda may fail before logging is running. There are also other alarms that can be generated for different reasons.

To avoid multiple pages to oncall, this PR creates a composite alarm over all the primitive alarms and is associated with the SNS topic to generate pages. The supporting alarms exists so that root cause can be easily identified but these alarms **do not generate pages.**

The composite alarm will show the oncall which supporting alarm is the root cause:
image

In the future we may want to make ALL the alarms composite and depend on primitive alarms + a "deployment" alarm. We can have the deployment custom resources manage the "deployment" alarm and in this way suppress pages during deployments.

# Testing

 - mage test:ci
 - manually caused alarms


deployments/bootstrap_gateway.yml
```diff
@@ -146,6 +146,11 @@ Resources:
             - Effect: Allow
               Action:
                 - cloudwatch:PutMetricAlarm
+                - cloudwatch:PutCompositeAlarm
+                - cloudwatch:TagResource
+              Resource: !Sub arn:${AWS::Partition}:cloudwatch:${AWS::Region}:${AWS::AccountId}:alarm:*
+            - Effect: Allow
```
kostaspap, line 152: Do these permissions need to be added in the deployment role as well?

rleighton: I checked, the deployment role has `cloudwatch*`

rleighton:
> Much much better. Thanks for this cleanup!
> 
> Is there some special migration needed, or just deploying over an existing Panther deployment will update the old alarms (to stop sending to SNS)?

For a new deployment or release the changes will be forced. For an existing dev deployment you need to do 
```
export CUSTOM_RESOURCE_VERSION=`date`
```
to force the change.

PR #1199, opened 2020-07-20
Fixing nil pointer dereference

kostaspap:
## Background

Fixing a  nil-pointer dereference

## Testing

- mage test:ci

### DIFF
--- a/internal/compliance/snapshot_poller/pollers/aws/cloudwatchlogs_log_group.go
+++ b/internal/compliance/snapshot_poller/pollers/aws/cloudwatchlogs_log_group.go
@@ -84,8 +84,8 @@ func PollCloudWatchLogsLogGroup(
 	if snapshot == nil {
 		return nil, nil
 	}
-	snapshot.Region = aws.String(resourceARN.Region)
-	snapshot.AccountID = aws.String(resourceARN.AccountID)
+	snapshot.Region = &resourceARN.Region
+	snapshot.AccountID = &resourceARN.AccountID
 	scanRequest.ResourceID = snapshot.ARN
 	return snapshot, nil
 }
--- a/internal/compliance/snapshot_poller/pollers/aws/cloudwatchlogs_log_group.go
+++ b/internal/compliance/snapshot_poller/pollers/aws/cloudwatchlogs_log_group.go
@@ -72,7 +72,14 @@ func PollCloudWatchLogsLogGroup(
 
 	// Split out the log group name from any additional modifiers
 	lgName := strings.Split(lgResource, ":")[0]
-	logGroup := getLogGroup(cwClient, aws.String(lgName))
+	logGroup, err := getLogGroup(cwClient, lgName)
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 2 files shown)


internal/compliance/snapshot_poller/pollers/aws/cloudwatchlogs_log_group.go
```diff
@@ -72,37 +72,44 @@ func PollCloudWatchLogsLogGroup(
...
+		return nil, nil
+	}
 	snapshot := buildCloudWatchLogsLogGroupSnapshot(cwClient, logGroup)
 	if snapshot == nil {
 		return nil, nil
 	}
-	snapshot.Region = aws.String(resourceARN.Region)
-	snapshot.AccountID = aws.String(resourceARN.AccountID)
+	snapshot.Region = &resourceARN.Region
+	snapshot.AccountID = &resourceARN.AccountID
```
kostaspap, line 88: Simpler to get the pointer with `&` than making a function call.

internal/compliance/snapshot_poller/pollers/aws/cloudwatchlogs_log_group.go
```diff
@@ -72,37 +72,44 @@ func PollCloudWatchLogsLogGroup(
...
-	logGroup := getLogGroup(cwClient, aws.String(lgName))
+	logGroup, err := getLogGroup(cwClient, lgName)
+	if err != nil {
+		return nil, err
+	}
+	if logGroup == nil {
+		// this can happen in case we didn't find the requested log group - it might have been deleted
+		return nil, nil
+	}
 	snapshot := buildCloudWatchLogsLogGroupSnapshot(cwClient, logGroup)
```
kostaspap, line 83: The `logGroup` would be null, and this could result in a nil pointer dereference in line 152

internal/compliance/snapshot_poller/pollers/aws/cloudwatchlogs_log_group.go
```diff
@@ -72,37 +72,44 @@ func PollCloudWatchLogsLogGroup(
 
 	// Split out the log group name from any additional modifiers
 	lgName := strings.Split(lgResource, ":")[0]
-	logGroup := getLogGroup(cwClient, aws.String(lgName))
+	logGroup, err := getLogGroup(cwClient, lgName)
+	if err != nil {
+		return nil, err
+	}
```
nhakmiller, line 78: I'd recommend not returning an error here, it's just going to get logged a second time in the `Handle`  function then dropped.

Probably easiest to not return an error from `getLogGroup` and instead just return up `nil, nil` if `logGroup` is `nil`.

kostaspap: Ah interesting. So in case we had trouble scanning the resource due to some issue we give up then, we don't retrying using SQS queue retries? Interesting. I guess this is one of the items that is meant to be fixed with the planned refactoring, right?

PR #1202, opened 2020-07-21
Add `tcodec` package to handle timestamps during JSON decoding/encoding

alxarch:
## Background

Currently to read/write timestamps for log events we need to define new types for each timestamp format encountered.
Each such type must use pointer fields to pass validation and re-implement the `encoding/json` interfaces and register itself with `awsglue` to work properly. This PR introduces a new package called `tcodec` that provides a `jsoniter.Extension` to manage all timestamp requirements without the need for all the wiring.

### Usage
```go

// Use UTC and format timestamps with a specific layout 
outputCodec :=  tcodec.LayoutCodec(`2006-01-02 15:04:05.000000000`)

ext := tcodec.NewExtension(tcodec.Config{
  // Will output all timestamps using this codec (tag only defines which codec to use for decode) 
  DecorateCodec: func (codec tcodec.Codec) tcodec.Codec {
    dec, _ := Split(codec)
    enc := EncodeIn(time.UTC, outputCodec)
    return tcodec.Join(dec, enc)
  },
})

// register the extenstion with a default jsoniter API (can be registered to individual APIs as well)
jsoniter.ConfigDefault.RegisterExtension(ext)

type Event struct {
  // Note we use simple time.Time value here (zero time will be omitted as null)
  Timestamp time.Time `json:"ts,omitempty" tcodec:"rfc3339" description:"The event timestamp"`
  Name string `json:"name" description:"The event name"`
}
event := Event{}
input := `{
  "ts": "2020-07-21T15:16:17.9999Z",
  "name": "Foo"
}`
_ = jsoniter.UnmarshalFromString(input, &event)
// event.Timestamp is a proper `time.Time` value at "2020-07-21T15:16:17.9999Z"

output, _ := jsoniter.MarshalToString(&event)
// output is `{"ts":"2020-07-21 15:16:17.999900","name":"Foo"}`
```

The way the extension works is by using a struct tag to define a `TimeCodec` to use for encoding/decoding a timestamp.
Normal `time.Time` fields can be used to define timestamps and the extension will handle proper JSON handling.
Configuration options allow to modify the extension behavior in detail to cover all use cases. 
 

This PR originates in ongoing work to support custom log parsers and can be used by both custom and 'normal' go log parsers. The code in this PR is isolated and does not affect the current log processor in any way.

## Changes

- Adds `tcodec` module under `pantherlog`

## Testing

- mage test:ci
- A similar approach was used on a separate PoC branch with proper E2E results.


internal/log_analysis/log_processor/pantherlog/tcodec/extension.go
```diff
@@ -0,0 +1,275 @@
...
+	return (*time.Time)(ptr).IsZero()
+}
+func (enc *jsonTimeEncoder) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
+	tm := *((*time.Time)(ptr))
+	enc.encode(tm, stream)
+}
+
+type jsonTimePtrEncoder struct {
+	encode TimeEncoderFunc
+}
```
kostaspap, line 194: Do we need a `ptr` encoder/decoder? Or maybe it would be enough for our usecases to just use `time.Time`? Could save us from some code, especially if we thing `time.Time` can cover all our cases (I don't see why not)

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec.go
```diff
@@ -0,0 +1,318 @@
...
+	fn(tm, stream)
+}
+
+// Split is a helper to split a TimeCodec into a decoder and an encoder.
+func Split(codec TimeCodec) (TimeDecoder, TimeEncoder) {
+	return codec, codec
+}
+
+// Join is a helper to compose a TimeCodec from a decoder and an encoder.
+func Join(decode TimeDecoder, encode TimeEncoder) TimeCodec {
```
kostaspap, line 69: Why do we need both `Join` and `JoinfFunc`?

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec.go
```diff
@@ -0,0 +1,318 @@
...
+	return EncodeIn(time.UTC, encoder)
+}
+
+// EncodeIn forces a `time.Location` on all encoded timestamps
+func EncodeIn(loc *time.Location, encoder TimeEncoder) TimeEncoder {
+	return &locEncoder{
+		encode: resolveEncodeFunc(encoder),
+		loc:    loc,
+	}
+}
```
kostaspap, line 277: Where will this be used?

alxarch: To enforce UTC on all encoded timestamps

kostaspap: Right, `EncodeUTC` should be enough then?

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec.go
alxarch: The `Func` variant saves typing when defining a codec 'inline'.

We write 
```
codec := tcodec.Join(func (tm time.Time, stream *jsoniter.Stream) {
// encode time
}, func (iter *jsoniter.Iter) time.Time {
// decode time
})
```

InsteadOf
```
codec := tcodec.Join(tcodec.TimeEncoderFunc(func (tm time.Time, stream *jsoniter.Stream) {
// encode time
}), tcodec.TimeDecoderFunc(func (iter *jsoniter.Iter) time.Time {
// decode time
}))
```

internal/log_analysis/log_processor/pantherlog/tcodec/extension.go
alxarch: We have a lot of pointer timestamps all over the codebase. We will need this to port older parsers without modifying the field types if other parts of the code rely on pointers.

alxarch:
> Nice work! My main question: Do we need all of this functionality or not? If not, I'd prefer to not include it. We should just leave in the things relevant to us, the smaller the code the easier to maintain. Better start small and add functionality as we need it.

We need *most* of the functinality for the builtin parsers and we will use the full extend of the package for customlogs. This excludes some trivial helper methods that help compose timecodecs with less typing.

internal/log_analysis/log_processor/pantherlog/tcodec/registry.go
```diff
@@ -0,0 +1,104 @@
...
+	defaultRegistry = &Registry{
+		codecs: map[string]TimeCodec{
+			"unix":    UnixSecondsCodec(),
+			"unix_ms": UnixMillisecondsCodec(),
+			"rfc3339": LayoutCodec(time.RFC3339Nano),
+		},
+	}
+)
+
+type Registry struct {
```
kostaspap, line 36: Could you add some comments/documentation explaining the purpose of this, why it was created and how it is meant to be used?

internal/log_analysis/log_processor/pantherlog/tcodec/registry.go
```diff
@@ -0,0 +1,104 @@
...
+		if other == nil {
+			continue
+		}
+		for name, codec := range other.codecs {
+			r.set(name, codec)
+		}
+	}
+}
+
+func Register(name string, codec TimeCodec) error {
```
kostaspap, line 88: How is this meant to be used? Will our code every interact with the `defaultRegistry` like this? Wouldn't we use the `(r *Registry) Register` functionality instead?

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec.go
```diff
@@ -0,0 +1,317 @@
...
+			iter.ReportError("ReadUnixMilliseconds", err.Error())
+			return
+		}
+		return UnixMilliseconds(msec)
+	default:
+		iter.Skip()
+		iter.ReportError("ReadUnixMilliseconds", `invalid JSON value`)
+		return
+	}
+}
```
kostaspap, line 215: Might make sense to split each codec to its own file. We will add eventually 5-10 more codecs (maybe some will be external contributions!), which could lead to this file growing a lot and making it harder for people to navigate through it. It could also help isolate the tests better.

internal/log_analysis/log_processor/pantherlog/tcodec/registry.go
alxarch: We can register timestamp formats that are used more frequently with the global registry so they can be available to custom logs and our own parsers as well.

kostaspap: Ok. And when are we going to use the `(r *Registry) Register`? Trying to understand why we need both.

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec.go
alxarch: Fixed by using only `EncodeIn`

PR #1203, opened 2020-07-21
Call the rule engine (not policy) when saving a rule

giorgosp:
## Background

Bug fix where rule saving calls the policy engine to test the rule instead of the rule engine.

## Changes

The fix

## Testing

Added integration tests for successfully saving a rule and a policy with passing tests.


internal/core/analysis_api/handlers/create_rule.go
```diff
@@ -118,7 +118,7 @@ func enabledRuleTestsPass(rule *models.UpdateRule) (bool, error) {
 		ResourceTypes: rule.LogTypes,
 		Tests:         rule.Tests,
 	}
-	testResults, err := policyEngine.TestPolicy(tp)
+	testResults, err := ruleEngine.TestRule(tp)
```
giorgosp, line 121: This is the fix

PR #1204, opened 2020-07-21
Add jsoniter extension to force `omitempty` on all fields

alxarch:
## Background

The current log parsers require adding `omitempty` to all fields. Forgetting to set omitempty could result in `null` values being written on output JSON. We do not like that. This PR adds a `jsoniter.Extension` that forces `omitempty` behavior on all struct fields.  

## Changes

- Adds a `jsoniter.Extension` to force omitempty.

## Testing

- mage test:ci

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/aws/aws-lambda-go v1.17.0
 	github.com/aws/aws-sdk-go v1.32.7
 	github.com/cenkalti/backoff/v4 v4.0.2
+	github.com/fatih/structtag v1.2.0
 	github.com/go-openapi/errors v0.19.6
 	github.com/go-openapi/runtime v0.19.19
 	github.com/go-openapi/strfmt v0.19.5
@@ -20,6 +21,7 @@ require (
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/leodido/go-urn v1.2.0 // indirect
 	github.com/magefile/mage v1.9.0
+	github.com/modern-go/reflect2 v0.0.0-20180701023420-4b7aa43c6742
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.6.1
 	github.com/tidwall/gjson v1.6.0
--- a/internal/log_analysis/log_processor/pantherlog/omitempty/omitempty.go
+++ b/internal/log_analysis/log_processor/pantherlog/omitempty/omitempty.go
@@ -18,15 +18,6 @@ package omitempty
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-/**
- * Copyright (C) 2020 Panther Labs Inc
- *
- * Panther Enterprise is licensed under the terms of a commercial license available from
- * Panther Labs Inc ("Panther Commercial License") by contacting contact@runpanther.com.
- * All use, distribution, and/or modification of this software, whether commercial or non-commercial,
- * falls under the Panther Commercial License to the extent it is permitted.
- */
-
 import (
 	"reflect"
 
--- a/internal/log_analysis/log_processor/pantherlog/omitempty/omitempty.go
+++ b/internal/log_analysis/log_processor/pantherlog/omitempty/omitempty.go
@@ -97,7 +97,7 @@ func injectOmitempty(original reflect.StructTag, key string) reflect.StructTag {
 	}
 	// Assert jsoniter omits fields witg `-` name
 	if tag.Name == "-" {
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -20,6 +21,7 @@ require (
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/leodido/go-urn v1.2.0 // indirect
 	github.com/magefile/mage v1.9.0
+	github.com/modern-go/reflect2 v0.0.0-20180701023420-4b7aa43c6742
```
kostaspap, line 24: This project has 334 stars and last activity 2 years ago. Can we do it without this library? Same for `https://github.com/fatih/structtag`

internal/log_analysis/log_processor/pantherlog/omitempty/omitempty.go
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+/**
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * Panther Enterprise is licensed under the terms of a commercial license available from
+ * Panther Labs Inc ("Panther Commercial License") by contacting contact@runpanther.com.
+ * All use, distribution, and/or modification of this software, whether commercial or non-commercial,
+ * falls under the Panther Commercial License to the extent it is permitted.
+ */
```
kostaspap, line 28: Don't need this one :)

internal/log_analysis/log_processor/pantherlog/omitempty/omitempty_test.go
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+/**
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * Panther Enterprise is licensed under the terms of a commercial license available from
+ * Panther Labs Inc ("Panther Commercial License") by contacting contact@runpanther.com.
+ * All use, distribution, and/or modification of this software, whether commercial or non-commercial,
+ * falls under the Panther Commercial License to the extent it is permitted.
+ */
```
kostaspap, line 28: Can remove this.

go.mod
austinbyers: `jsoniter` relies on `reflect2`, I believe. So we're already using it

alxarch: Yes, it is a requirement because `jsoniter` uses it internally and some extensions expose `reflect2.Type`

internal/log_analysis/log_processor/pantherlog/omitempty/omitempty.go
alxarch: fixed

internal/log_analysis/log_processor/pantherlog/omitempty/omitempty_test.go
alxarch: fixed

alxarch:
> If my understanding is correct, all structs that are encoded/decoded with `jsoniter` will have `omitempty` to all fields by default?

Omitempty only affects the encoding behavior. The 'injection' of `omitempty` happens only if this `jsoniter.Extension` is registered with the `jsoniter.API` that handles the encoding. We will be using specific 'extended' `jsoniter.API` instances for log event decoding/encoding.

internal/log_analysis/log_processor/pantherlog/omitempty/omitempty_test.go
```diff
@@ -0,0 +1,70 @@
...
+	jsoniter "github.com/json-iterator/go"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNew(t *testing.T) {
+	api := jsoniter.Config{}.Froze()
+	api.RegisterExtension(New("json"))
+
+	type A struct {
+		Foo    string `json:"foo"`
```
giorgosp, line 33: My only concern is the implicitness it adds. It may be tiring for the developer to annotate many fields as "omitempty" but very  confusing for the reader who hadn't stumbled upon or forgotten the code that registers this extension to jsoniter.
Maybe I don't really understand the motivation here, but I just wanted to make sure we don't make the code writer's life easier and the reader's harder.

alxarch: This will only be registered with the jsoniter api that will be used to produce JSON from log parsers. It is not something we will globally register. It will only affect log event struct definitions. The motivation not only avoid writing. It is that nullable fields which are not marked omitempty (by mistake or assumption that they are always not null) will output null values when encoded. The behavior we want for log parsers is to always be omitempty. We use validate required to verify the required fields.

PR #1206, opened 2020-07-21
Decouple `awsglue` from other modules defining mappings

alxarch:
## Background

To avoid circular dependencies on the pantherlog refactor we need to inverse the package relation for defining awsglue type mappings. Instead of importing all packages *into* awsglue we use a new `awsglue.MustRegisterMapping()` helper to define mappings *from* the package that defines the type in the first place.

## Changes

- Adds `awsglue.MustRegisterMapping()` helper and defines mappings in respective packages.
- Moves `parsers.RewriteField` to `awsglue` package
- Adds the `awsglue.TimestampLayout` constant to be used for serializing timestamps for glue
- Checks that new logtypes can generate a glue schema on registration

## Testing

- `mage test:ci`


kostaspap:
Where do we hit circular dependencies? Would it be possible to ... break the circle in a different way? 

As I see it, It is fine for AWS Glue to have a dependency on parsers package - it is the other way around that is problematic. 
The parsers shouldn't really depend on the system we use to index/store the data. In time we will want to add mappings from our data types to Snowflake types, GCP types, etc. It would be much better to have the Glue-specific logic in one package, then the Snowflake-specific logic in a Snowflake package and so on. 
(I believe this is something we had discussed in a prior PR)

alxarch:
> Where do we hit circular dependencies? Would it be possible to ... break the circle in a different way?
> 

I hit the circular dependency when trying to validate that the generated event structs from pantherlog and customlogs are valid for awsglue.

> As I see it, It is fine for AWS Glue to have a dependency on parsers package - it is the other way around that is problematic. The parsers and their datatypes are "root" package for us and should not depend on the system we use to index/store the data.

This is the other way around really. Currently we add specific overrides (RewriteFieldNames) in the parsers package to accomodate glue specifically. If we want to accomodate different storage requirements these 'quirks' should live in awsglue and the parsers would use the desired quirk mode for their results depending on where they are used. Dependency graph needs to be reverse to achieve that. 

> In time we will want to add mappings from our data types to Snowflake types, GCP types, etc. It would be much better to have the Glue-specific logic e.g. schema generation, in one package, then the Snowflake-specific logic in a Snowflake package and so on.

Exactly, so quirks for each specific system should be defined in the respective package (in this case awsglue)

> (I believe this is something we had discussed in a prior PR)

Yes and we avoided this by not checking validity of defined structs on registration. This might have worked OK then since we e2e tested the table generation but for customlogs this should be checked on parser creation. Thus the requirement.

rleighton:
> > Where do we hit circular dependencies? Would it be possible to ... break the circle in a different way?
> 
> I hit the circular dependency when trying to validate that the generated event structs from pantherlog and customlogs are valid for awsglue.
> 
> > As I see it, It is fine for AWS Glue to have a dependency on parsers package - it is the other way around that is problematic. The parsers and their datatypes are "root" package for us and should not depend on the system we use to index/store the data.
> 
> This is the other way around really. Currently we add specific overrides (RewriteFieldNames, timestamp format) in the parsers package to accomodate glue specifically. If we want to accomodate different storage requirements these 'quirks' should live in awsglue and the parsers would use the desired quirk mode for their results depending on where they are used. Dependency graph needs to be reverse to achieve that.
> 
> > In time we will want to add mappings from our data types to Snowflake types, GCP types, etc. It would be much better to have the Glue-specific logic e.g. schema generation, in one package, then the Snowflake-specific logic in a Snowflake package and so on.
> 
> Exactly, so quirks for each specific system should be defined in the respective package (in this case awsglue)
> 
> > (I believe this is something we had discussed in a prior PR)
> 
> Yes and we avoided this by not checking validity of defined structs on registration. This might have worked OK then since we e2e tested the table generation but for customlogs this should be checked on parser creation (ie at runtime). Thus the requirement.


Would thinking about this with a different label help? Perhaps we should change 'awsglue' as the package name to 'logschema' (where we happen to use Hive-based format and that is what glue uses). When I do the snowflake schema I use translations from this base hive schema we use.

rleighton:
Where are we with this?

alxarch:
> Where are we with this?

After our call yesterday I updated the PR to separate `parsers.PantherAnyString` into it's own module. This will be needed by the `pantherlog` PR as well to minimize the overhead of adapting the parsers to the new `Result` interface. After the `pantherlog` PR is merged we will explore transitioning to a struct-based schema generation based on a `logschema` module to facilitate defining schemas for all different storage backends.

internal/log_analysis/log_processor/logtypes/registry.go
```diff
@@ -238,5 +238,29 @@ func checkLogEntrySchema(logType string, schema interface{}) error {
 	if err := jsoniter.Unmarshal(data, &fields); err != nil {
 		return errors.Errorf("invalid schema struct for log type %q: %s", logType, err)
 	}
+	// Verify we can generate glue schema from the provided struct
```
rleighton, line 241: Nice

PR #1207, opened 2020-07-21
Expand null types to cover all scalar values

alxarch:
## Background

We need null types for all scalar go values (`string`, `intX`, `uintX`, `floatX`, `bool`) that don't require pointers to detect null input in JSON. Numeric types in this PR are the successors to the `numerics` types and cover all sized numeric values.

> NOTE: All numeric types handle both string and number JSON input to avoid having to duplicate types and to handle common issues when parsing the input logs transparently.

## Changes
- Moves `null` package under `pantherlog`
- Alters `null.String` to use the 'strict' behavior that preserves empty strings in the input JSON.
- Adds `null.NonEmpty` for strings that should be omitted if empty but handle `null` as JSON input
- Adds `null.Uint64`
- Adds `null.Uint32`
- Adds `null.Uint16`
- Adds `null.Uint8`
- Adds `null.Int64`
- Adds `null.Int32`
- Adds `null.Int16`
- Adds `null.Int8`
- Adds `null.Boolean`
- Adds proper `awsglue` mappings for all `null` types
- Adds a helper method to register custom type validation handlers to a `validator.Validate` instance.

## Testing

- unit tests to cover all input cases 
- mage test:ci

## Benchmarks

To test the different behavior benchmarks were written to compare JSON decoding on `AWS.CloudTrail` and `Gitlab.Production` logs using both the pointer and the null approach.
The results after multiple runs consistently show a ~10% speed improvement and ~30% allocation reductions in these complex log types.

```
goos: linux
goarch: amd64
pkg: github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog/null
BenchmarkNullTypes
BenchmarkNullTypes/CloudTrail_-_pointers
BenchmarkNullTypes/CloudTrail_-_pointers-4         	  187176	      6467 ns/op	    1816 B/op	      70 allocs/op
BenchmarkNullTypes/CloudTrail_-_null
BenchmarkNullTypes/CloudTrail_-_null-4             	  206578	      5548 ns/op	    1704 B/op	      52 allocs/op
BenchmarkNullTypes/GitLab_-_pointers
BenchmarkNullTypes/GitLab_-_pointers-4             	   94364	     12828 ns/op	    6100 B/op	      80 allocs/op
BenchmarkNullTypes/GitLab_-_null
BenchmarkNullTypes/GitLab_-_null-4                 	   89343	     11321 ns/op	    5591 B/op	      55 allocs/op
PASS
```


internal/log_analysis/log_processor/pantherlog/null/boolean.go
```diff
@@ -0,0 +1,156 @@
...
+		case 0:
+			*((*Boolean)(ptr)) = Boolean{
+				Value:  false,
+				Exists: true,
+			}
+		case 1:
+			*((*Boolean)(ptr)) = Boolean{
+				Value:  true,
+				Exists: true,
+			}
```
kostaspap, line 124: Interesting. So if a field has a value of `0`, if we mark it to be type `null.Boolean`, it will be unmarshaled to `false`? Neat. 

I think it worth adding the cases it handles (`"true"/"false"`, `true/false`, `0/1`) in the documentation of the `Boolean` struct, so users can easily see in which cases they could use it. Same goes for the rest of the types with similar capabilities.

internal/log_analysis/log_processor/pantherlog/null/boolean.go
```diff
@@ -0,0 +1,156 @@
...
+import (
+	"strconv"
+	"unsafe"
+
+	jsoniter "github.com/json-iterator/go"
+
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/jsonutil"
+)
+
+type Boolean struct {
```
kostaspap, line 31: nitpick: In the other "null" types we use Go's name for the data type e.g. Int8, UInt8 etc, Should we just name this `Bool` since the data type in Go is `bool`?

internal/log_analysis/log_processor/pantherlog/null/boolean.go
```diff
@@ -0,0 +1,156 @@
...
+
+// MarshalJSON implements json.Marshaler interface.
+func (b *Boolean) MarshalJSON() ([]byte, error) {
+	if !b.Exists {
+		return []byte(`null`), nil
+	}
+	return strconv.AppendBool(nil, b.Value), nil
+}
+
+// int8Codec is a jsoniter encoder/decoder for int8 values
```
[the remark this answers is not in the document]
alxarch: fixed

internal/log_analysis/log_processor/pantherlog/null/boolean.go
alxarch: fixed

internal/log_analysis/log_processor/pantherlog/null/boolean.go
alxarch: fixed

PR #1217, opened 2020-07-22
Updating lacework schema to use timestamp for time field

kostaspap:
## Background

Updating lacework schema to use timestamp for time field

## Changes

- Using timestamp format for time field
- Added missing IP addresses

## Testing

- mage test:ci using existing tests


kostaspap:
> Can you add entry here for Lacework: https://github.com/panther-labs/panther/blob/master/docs/gitbook/SUMMARY.md

Good catch

PR #1218, opened 2020-07-22
Fix issue with wrong test payload 

### LINKED ISSUE #1214: Test sends wrong payload to backend
### Describe the bug

It seems that the UI sends wrong data to backend when a user is running a test from a Rule

### Steps to reproduce

1. Go to Log Analysis -> Rule
2. Find an existing rule that has some tests. This is really important! It seems the issue appears only when there are more than one tests for a rule. 
3. Create a new test. Populate same values e.g. `{"test":"test"}`
4. Open dev console in your browser
5. Click Run. You will see that FE sends the wrong `resource` field (it will always send `{}`)

The above blocks users from creating new tests in the UI. 

### Expected behavior

FE would send the correct resource to backend. 

### Environment

- Panther version or commit: fb6cab1a83e73b38f1010a0fa634dd9b2ecd705d

3nvi:
## Background

We currently have an issue where new tests being created are being sent with an empty object `"{}"` as the `resource`. This fixes that

Closes #1214 

## Changes

- Allow re-assigning of referentially equal function on `useFastField`
- Add more spacing to test error text

## Testing

- Manually


3nvi:
> Deployed and verified correct behavior!

Thanks @kostaspap !

PR #1220, opened 2020-07-22
Front-end: Don't restore form values if Cancel is clicked

### LINKED ISSUE #1213: Create Rule operation populates rule fields with older values
### Describe the bug

Creating a rule operation seems to use data from older operation

### Steps to reproduce

1. Go to Log Analysis -> Rules
2. Click on Create -> Single
3. Populate some fields, e.g. ID
4. Click "Cancel"
5. Click again on Create -> Single

You will see that the Rule has kept the older Rule id. 
### Expected behavior

I expect that clicking to create a single rule, displays me a fresh page. 

### Environment

- fb6cab1a83e73b38f1010a0fa634dd9b2ecd705d
- OS: MacOS 10.15.5
- Browser: Chrome 84.0.4147.89

3nvi:
## Background

Currently we persist values for our compex forms and restore them when  the  user navigates away within **the same session**. This is because they might accidentally navigate  away and we don't want  them to lose their existing work.

Some people thought this was a bug, cause sometimes you **do want** to navigate away without restoring anything (again in the same session). This PR makes sure to **not restore**  your values if you click the "cancel" button on a form.

Closes #1213 

## Changes

- Add session value clearance on "cancel" button click

## Testing

- Tested that restoration & clearance  works. `mage test:ci`


web/src/components/utils/FormSessionRestoration/FormSessionRestoration.test.tsx
```diff
@@ -0,0 +1,86 @@
+import { render, fireEvent, waitMs } from 'test-utils';
```
3nvi, line 1: @alexmylonas @s0l0ist

web/src/components/utils/FormSessionRestoration/FormSessionRestoration.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+          <Button onClick={clearFormSession}>Cancel</Button>
+        </Box>
+      )}
+    </FormSessionRestoration>
+  </Formik>
+);
+
+// values stored in `sessionStorage` will also be available in other tests unless we clear them
+beforeEach(() => {
+  sessionStorage.clear();
```
vorillaz, line 32: I guess this should be placed somewhere globally, maybe as a `beforeEach` hook for all tests.

web/src/components/utils/FormSessionRestoration/FormSessionRestoration.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const testValue = 'test-value';
+
+  const { getByLabelText, unmount } = render(<TestForm />);
+
+  const textInput = getByLabelText('Text');
+  fireEvent.change(textInput, { target: { value: testValue } });
+  expect(textInput).toHaveValue(testValue);
+
+  // wait for debounce to kick in
+  await waitMs(300);
```
vorillaz, line 45: There is no actual assertion about using the session storage.
You can use the actual mocking assertions provided by jest
```
  expect(sessionStorage.setItem).toHaveBeenWith(KEY, VALUE);
```

web/src/components/utils/FormSessionRestoration/FormSessionRestoration.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  // Click the cancel
+  fireEvent.click(getByText('Cancel'));
+  expect(sessionStorage.__STORE__).toEqual({});
```
vorillaz, line 72: That's actually highly dependent on the `jest-localstorage-mock`. If we decide to use a headless browser instead of JSDom this one will be invalid.
Mocks are working great for cross-browser compatibility.

web/src/components/utils/FormSessionRestoration/FormSessionRestoration.test.tsx
3nvi: I think it only  affects single files  (since Jest isolates each test), but not 100%  sure

web/src/components/utils/FormSessionRestoration/FormSessionRestoration.test.tsx
3nvi: I'll try and see if I can  use the native API,  but I'm not 100%  sure

web/src/components/utils/FormSessionRestoration/FormSessionRestoration.test.tsx
3nvi: Added it globally

web/src/components/utils/FormSessionRestoration/FormSessionRestoration.test.tsx
3nvi: Sure!

PR #1221, opened 2020-07-22
Add FOSSA directly to CI

austinbyers:
## Background

A few small CI fixes which improve security and lower the runtime.

The main motivation is to remove FOSSA as a Github-level integration if possible, to minimize the number of third-party integrations with direct access to (enterprise) source code

Once this change is merged, I'll remove the FOSSA github app and existing project, replacing it with the project created automatically by this PR

## Changes

- Add FOSSA directly to CI instead of relying on a webhook (following https://docs.fossa.com/docs/circleci)
    - This way, we publish metadata from CI instead of asking FOSSA to download the code and analyze it for us
    - I tried having fossa wait for results and block CI (`fossa test`), but it could never finish in under 10 minutes, so we'll just rely on the README badge and Slack notifications to learn about failures
    - Remove the `.fossa.yml` config file - it was out of date and FOSSA can generate it automatically on every build
- Remove leftover debugging from #1158 
- Remove `dupl` from the Go metalinter - we already disable it in many places, there are many valid reasons to have similar-looking code, especially because Go does not have generics (@alxarch )

## Testing

- CircleCI


austinbyers:
We were told FOSSA might be faster when running from CircleCI, but here it still took more than 6 minutes and didn't seem to work.

I'll play around with it some more

austinbyers:
Ok I updated the PR description and pushed the final configuration. After this PR is merged, the "License Compliance" check you see above will disappear

(I should have made this a draft PR to test first, my apologies.)

jacknagz:
> Remove the .fossa.yml config file - it was out of date and FOSSA can generate it automatically on every build

I'm not seeing it scan our react or python code now. Can we generate the config to make sure those modules are added?

.circleci/config.yml
```diff
@@ -105,10 +94,22 @@ jobs:
           name: Check bundle size
           command: npm run bundlesize
 
+  # Upload dependency metadata to FOSSA, analyze offline
+  fossa_upload:
+    executor: panther-buildpack
+    resource_class: small
+    steps:
+      - run: |
+          curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh | bash
```
3nvi, line 103: normally we start with `checkout`, but I understand  it makes no difference

austinbyers: Doing this first is the official suggestion in [their docs](https://docs.fossa.com/docs/circleci#add-fossa-steps-to-configyml), presumably because there's no point checking out the code if you can't download the tool to analyze it

austinbyers:
> I'm not seeing it scan our react or python code now. Can we generate the config to make sure those modules are added?

I'm guessing it hasn't fully scanned `master` yet, since this branch hasn't merged. But in the scan for this branch, I can see `pip`, `npm` and Go modules, with deep dependencies as well. `fossa init` for me locally generates the following config file:

```yaml
version: 2
cli:
  server: https://app.fossa.com
  fetcher: custom
  project: git@github.com:panther-labs/panther
analyze:
  modules:
  - name: github.com/panther-labs/panther/cmd/devtools/logprocessor/logprocessor
    type: go
    target: github.com/panther-labs/panther/cmd/devtools/logprocessor/logprocessor
    path: cmd/devtools/logprocessor/logprocessor
  # ... other go modules elided ...
  - name: .
    type: npm
    target: .
    path: .
  - name: .
    type: npm
    target: .
    path: .
  - name: .
    type: pip
    target: .
    path: .
```

Which is essentially the same as the config we have now (don't know why it added `npm` twice though).

README.md
```diff
@@ -139,9 +139,9 @@ Panther source code is licensed under [AGPLv3](https://github.com/panther-labs/p
 
 ### FOSSA Status
 
-[![FOSSA Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Fpanther-labs%2Fpanther.svg?type=large)](https://app.fossa.com/projects/git%2Bgithub.com%2Fpanther-labs%2Fpanther?ref=badge_large)
+[![FOSSA Status](https://app.fossa.com/api/projects/custom%2B15287%2Fgit%40github.com%3Apanther-labs%2Fpanther.git.svg?type=large)](https://app.fossa.com/projects/custom%2B15287%2Fgit%40github.com%3Apanther-labs%2Fpanther.git?ref=badge_large)
 
-## Contributors ✨
+## Contributors
```
austinbyers, line 144: With the icon after the name, the anchor link for this header is `#contributors-`, but the badge at the top is expecting `#contributors`, so I removed the star to fix the link

PR #1222, opened 2020-07-22
Add snowflake api configuration documentation

rleighton:
## Background

This PR adds instructions on how to configure Panther to use Snowflake as the "database", instead of Athena in DataExplorer.

## Changes

- updated docs, added screen shots

## Testing

- mage test:ci


docs/gitbook/enterprise/snowflake.md
```diff
@@ -56,9 +66,12 @@ Edit your `deployments/panther_config.yml` to add `arn:aws:iam::34318291XXXX:use
       - arn:aws:iam::34318291XXXX:user/k7m2-s-v2st0722
 ```
 
-Next, run `mage deploy snowflake:snowpipe`
+If deploying from `master yml` also update `DestinationClusterARNs` as above.
 
-When the deployment is done there should be a `snowpipe.sql` file created in `./out/snowflake/snowpipe.sql`
+Next, run `mage deploy` if deploying from source or deploy via `master.yml`.
```
nhakmiller, line 71: I think we generally refer to deploying from `master.yml` as prepackaged deployments. For pre-packaged deployments they just need to update the CloudFormation stack with the same template but different parameters. To avoid confusing customers we might want to use the same "pre-packaged" terminology here.

docs/gitbook/enterprise/snowflake.md
```diff
@@ -56,9 +66,12 @@ Edit your `deployments/panther_config.yml` to add `arn:aws:iam::34318291XXXX:use
...
+
+When the deployment is done, run `mage snowflake:snowpipe`. When finished there should be a `snowpipe.sql` file 
+created in `./out/snowflake/snowpipe.sql`
```
nhakmiller, line 74: For customers using pre-packaged deployments (which includes all self-hosted enterprise customers), they likely do not have the repo checked out or any of the development dependencies setup. If running `mage snowflake:snowpipe` is the only way to configure this, we might want to drop a link to the development workflow docs.

docs/gitbook/enterprise/snowflake.md
```diff
@@ -83,3 +96,72 @@ You can quickly test if the data ingestion is working by running simple queries,
...
+        }
+    ]
+}
+```
+Then click the `Save` button.
+
+Update the `SecretsManagerARN` attribute with the ARN of the secret in the `master.yml` CloudFormation template or
+ the `panther_config.yml` if deploying from source.
+
+Next deploy Panther, either from the `master.yml` CloudFormation template or doing `mage deploy` if from source.
```
nhakmiller, line 158: Same comment about pre-packaged deployments vs. `master.yml`.

docs/gitbook/enterprise/snowflake.md
rleighton: Actually, I'm going to create an ops tool to replace the `mage` target. How do pre-package customs get the ops tools like `requeue`?

docs/gitbook/enterprise/snowflake.md
rleighton: Done

docs/gitbook/enterprise/snowflake.md
rleighton: Done

docs/gitbook/enterprise/snowflake.md
rleighton: Austin told me ... we do not package those today. Will need to in the future.

PR #1226, opened 2020-07-23
feat: PR syncing automation

3nvi:
## Background

We spend some time manually syncing changes to other downstream repos. This PR makes sure to automate this procedure. Specifically, when a PR gets **merged** (not closed) in the OSS repo, a PR is created in our downstream listeners that:

- Creates a branch in the downstream repository (that matches the name of the branch in the origin repo with a prefix) as the PR branch.
- Adds the changes of the merge commit from the original PR (that was already merged)
- Opens a new PR from the newly created branch 
- Assigns the author of the original PR as the actual reviewer
- Adds the title, milestones, labels & body of the original PR

There is a bailout technique where you can use the label named `enterprise-backport` if you **don't want**  your PR synced/cloned to listeners.

## Structure

This describes how the automation is organized:

- We create an action by defining a contract (`action.yml`) and its source code (`index.js`)
- We use it in a workflow in the same way we use a CircleCI job

## Config for use in other repos

The action could be potentially re-used across different repos. It's inputs & outputs are the following:

- Input:`destRepo`, `destBranch`,  `ignoreLabel` & `token`
- Output: `url` (of the new PR) & `message` (status of the process)

## Changes

- Add github action to sync PRs from `src` to `dest`
- Add README to explain
- Add a github workflow to trigger it

## Testing

- Managed to create enterprise PRs from a fork of Panther


.github/workflows/main.yml
```diff
@@ -0,0 +1,27 @@
+on:
+  pull_request:
+    branches:
+      - master
+    types:
+      - closed
```
3nvi, line 6: only monitor closed  (merged or not) PRs

3nvi: (We can't isolate just the `merged` ones )

.github/workflows/main.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  sync_prs_from_community_to_enterprise:
+    runs-on: ubuntu-latest
+    name: Syncs merged PRs from Panther Community to Panther Enterprise
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.PR_SYNC_TOKEN }}
+          fetch-depth: 3
```
3nvi, line 17: `fetch-depth: 2` could work as well (1 won't, since downstream won't be able to compare histories correctly)

.github/workflows/main.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      - name: Monitor PRs on a Source Repo & Create PRs on a dest one
+        uses: ./build/actions/pr_sync
+        id: pr_sync
+        env:
+          ACTIONS_STEP_DEBUG: ${{ secrets.ACTIONS_STEP_DEBUG }}
```
3nvi, line 23: when that's enabled (through github secrets), the debug log will be verbose. By default the debug log is hidden

build/actions/pr_sync/action.yml
```diff
@@ -0,0 +1,25 @@
+name: Sync merged PRs
+description: Automatically creates new PRs in a repo from the merged PRs of another repo
```
3nvi, line 2: kinda like CFN. Don't need to explain much

build/actions/pr_sync/package.json
```diff
@@ -0,0 +1,21 @@
+{
```
3nvi, line 1: each github action is an NPM project, so it needs a `package.json`

3nvi: Because of  that, there is a `node_modules` folder in this folder. @austinbyers I think `mage fmt` will need to ignore the `/node_modules/` regex instead of just ignoring the top-level project-wide `node_modules`

austinbyers: For web code, `mage fmt` is just running prettier, which you setup. So you can configure `.prettierrc` or `.prettierignore`, for example

.github/workflows/main.yml
austinbyers: How does this work when multiple OSS PRs merge in rapid succession?

They may not merge downstream until much later, in which case each enterprise PR may have nontrivial commit histories

build/actions/pr_sync/package.json
3nvi: I meant for the addition of the license comments. They gett added in node_modules stuff :D

austinbyers: Sure, license paths are controlled here: https://github.com/panther-labs/panther/blob/48486c99e5c88c067ceb4126092874d2e404cf06/tools/mage/fmt_license.go#L29

Feel free to change as you need

.github/workflows/main.yml
3nvi: Yeah I was thinking about the exact same thing during vacation. This won't create an issue on the OSS side (since the automation gets a clone of master the moment the PR was merged), but it might create issues if the enterprise ones (the "OSS Sync" ones), stay unmerged for a long time.

With  `fetch-depth: 3`, up to 2 Sync PRs can be open at any given moment before disaster happens 😄 

I'll increase that to `10` for safety purposes. This will make the syncing proceess a few seconds slower (~5 seconds), but I feel it's worth it

build/actions/pr_sync/package.json
3nvi: Thanks. Added an escape hatch with a comment

rleighton:
In the description you say `Clones branch name (with prefix)`. Does this mean PR branches get cloned automatically (why would we want that)? Disclaimer, I've not read the code yet. For me at least, it would help to have a lot more elaboration in the description to help understand the code.

rleighton:
> In the description you say `Clones branch name (with prefix)`. Does this mean PR branches get cloned automatically (why would we want that)? Disclaimer, I've not read the code yet. For me at least, it would help to have a lot more elaboration in the description to help understand the code.

I would suggest explaining how this changes the current process and include some examples of end to end from OSS PR to final merge in Enterprise.

3nvi:
> > In the description you say `Clones branch name (with prefix)`. Does this mean PR branches get cloned automatically (why would we want that)? Disclaimer, I've not read the code yet. For me at least, it would help to have a lot more elaboration in the description to help understand the code.
> 
> I would suggest explaining how this changes the current process and include some examples of end to end from OSS PR to final merge in Enterprise.

I updated the description to make it more verbose. We won't change **anything** in the way we work. Whatever wee did before, we will still be doing it now. This automation just "picks up" the changes of a PR merge and opens a new PR on the enterprise with the recently merged changes.

To illustrate, suppose you have a PR with a single change on an `.md` file. The PR is named **Change MD file** and its changes live in the branch `russ-fix-md`. As soon as your PR is merged, the automation will:

- Get the merge commit of the merged PR
- Create a branch in the enterprise named `oss-sync-russ-fix-md` from the enterprise master and cherry pick the changes of the merge commit in the OSS repo
- Open a PR against the enterprise with the title **[OSS Sync] Change MD file** and tag you as the reviewer
- Copy over any labels or milestones that the original PR had on it

I think a similar thing is explained in the description, although I understand that an actual example might be more helpful. Did I answer your questions?

build/actions/pr_sync/index.js
```diff
@@ -0,0 +1,117 @@
...
+      core.setOutput('message', 'PR contained an ignore label. Terminating...');
+      return;
+    }
+
+    core.debug('An ignore label was not found. Starting sync process...');
+    const destPullRequestBranchName = getPrBranch(srcPullRequest.head.ref);
+
+    core.debug('Creating a branch from the merge commit...');
+    execSync(`git checkout -b ${destPullRequestBranchName}`);
+    execSync(`git remote add target https://github.com/${destRepo}.git`); // prettier-ignore
```
alxarch, line 67: Does this added remote get automatically removed when the action ends? Won't this fail the second time if the remote is already added in a previous run?

3nvi: Yup, it's a new **clean** container everytime. That's how most CIs work 😃

.github/workflows/main.yml
nhakmiller: By increasing it to 10, we can have 9 sync PRs open "before disaster happens"? What does "disaster" mean in this scenario? Does the merge PR just fail to open automatically I'm guessing? Or does it block merging somehow?

3nvi: > we can have 9 sync PRs open "before disaster happens"?
yes

> Does the merge PR just fail to open automatically I'm guessing
Yeah the automation doesn't work :D

tools/mage/fmt_license.go
```diff
@@ -65,6 +65,13 @@ func fmtLicense(paths ...string) {
 }
 
 func addFileLicense(path, asteriskLicense, hashtagLicense string) {
+	// some monorepo dirs have nested node_modules. Until NPM ships their "workspaces" feature,
+	// we can't hoist all node modules in single folder at the project root. This prevents us
+	// from wrongly adding headers to 3rd party modules
+	if strings.Contains(path, "node_modules") {
```
austinbyers, line 71: Ideally we could skip traversing the entire directory rather than testing every filepath like this. In other words, ideally we could move the check higher up the call stack

What's the `node_modules` directory that gets added?

3nvi: it can be anywhere. Right now it's at `build/action/pr_sync/node_modules/*`,  but in the future more nestted `node_modules` dirs can exist

austinbyers: Then the better way to do this is to stop the traversal early if a `node_modules `directory is discovered - otherwise this will be checking thousands of filepaths.

This would happen on line 59 and we'd need to replace the `walk` function with `filepath.Walk` - I'll push a change to this branch

austinbyers: Change pushed and verified

PR #1244, opened 2020-07-27
Move initial analysis set and add CFN parameter validation

austinbyers:
## Background

The initial bulk upload of Python detections currently happens in the `core` stack. However, in Panther Enterprise this creates a cyclic dependency (https://github.com/panther-labs/panther-enterprise/issues/427), so I'm moving the problematic resource to the final stack (`web`).

This approach is cleaner anyway - now the bulk upload custom resource does not need to spin and wait for the python engines to finish. The only downside is that the `web` stack no longer honors its name - it now has several non-web resources that just need to run at the very end. This is similar to `bootstrap-gateway`, which is not just API gateways but any bootstrap resource that requires S3.

The bug linked above went for weeks without detection because `mage deploy` was passing an empty string for a CloudFormation parameter, which CFN accepts. To prevent this from happening again in the future, I'm also adding comprehensive validation for every CFN parameter in every stack (essentially an extension of #1159). This will help users and developers catch mistakes before any resources are actually created

## Changes

- Move the `Custom::AnalysisSet` resource from `panther-core` into `panther-web`
    - To avoid requiring a migration, this resource will always exit early if the `panther-analysis` Dynamo table has any entries. This resource is only needed for the very first deployment, it should never modify a deployment after that
    - Errors with this resource no longer fail deployment - we have encountered sporadic 500 errors when doing the first BulkUpload. If this fails, the entire Panther deployment rolls back today, but that's not necessary - the user can always login and re-upload the pack manually
- Bump version in deployment role and master template to 1.6.0
- Add parameter validation to every CloudFormation stack, using regex patterns when possible
- Clarify when a bucket *name* is expected vs an *arn*

## Testing

(in progress)

- validated regex patterns against an existing deployment with https://regex101.com
- `mage deploy test:integration teardown`
- `mage master:deploy teardown`
- Upgrade from v1.5.1 to this branch


deployments/bootstrap_gateway.yml
```diff
@@ -33,40 +33,52 @@ Parameters:
...
     Description: List of LayerVersion ARNs to attach to each function
   ProcessedDataBucket:
     Type: String
-    Description: S3 bucket which stores processed logs
+    Description: Name of the S3 bucket which stores processed logs
+    AllowedPattern: '^[a-z0-9.-]{3,63}$'
   PythonLayerVersionArn:
     Type: String
-    Description: Custom Python layer for analysis and remediation
+    Description: Optional custom Python layer for analysis and remediation
```
nhakmiller, line 70: Maybe `Optional versioned custom Python layer for analysis and remediation`, it's implicit from the example below but a callout for the less well versed in AWS lambda layers might be helpful.

deployments/bootstrap_gateway.yml
```diff
@@ -33,40 +33,52 @@ Parameters:
...
+    # Example: "arn:aws:lambda:us-west-2:111122223333:layer:panther-analysis:143"
+    AllowedPattern: '^(arn:(aws|aws-cn|aws-us-gov):lambda:[a-z]{2}-[a-z]{4,9}-[1-9]:\d{12}:layer:\S+:\d+)?$'
   TracingMode:
     Type: String
     Description: Enable XRay tracing on Lambda and API Gateway
     AllowedValues: ['', Active, PassThrough]
   UserPoolId:
     Type: String
     Description: Cognito user pool ID
+    # Example: "us-west-2_iHzBYf5SA"
```
nhakmiller, line 80: I'm assuming this ID and others have been anonymized?

deployments/cloud_security.yml
```diff
@@ -46,21 +53,29 @@ Parameters:
...
     Type: String
     Description: Name of the S3 bucket for storing processed logs
+    AllowedPattern: '^[a-z0-9.-]{3,63}$'
   ProcessedDataTopicArn:
     Type: String
     Description: The ARN of the processed data SNS topic
+    AllowedPattern: '^arn:(aws|aws-cn|aws-us-gov):sns:[a-z]{2}-[a-z]{4,9}-[1-9]:\d{12}:\S+$'
   PythonLayerVersionArn:
     Type: String
     Description: Pip libraries for python analysis and remediation
```
nhakmiller, line 63: Maybe specify that the pip libraries must be in the form of a lambda layer? Again implicit from the example but just based on the parameter name and description I'm not seeing lambda being called out explicitly.

deployments/bootstrap_gateway.yml
austinbyers: I changed my account ID, but the other IDs were real. They don't exist anymore, since I've already torn down the deployment

austinbyers: They're randomly generated each time, so I don't see any issue leaving them as is

nhakmiller: Yeah makes sense.

austinbyers:
All tests passed - upgrading from 1.5.1 left changes to policies and rules in place

PR #1249, opened 2020-07-28
Fix Sign In message on error 

alexmylonas:
## Background

There is a minor issue on https://github.com/aws-amplify/amplify-js/pull/4427 that seems to affects us. This PR handles this and displays a valid error.

## Testing

- Locally

### DIFF
--- a/web/src/components/forms/SigninForm/SigninForm.tsx
+++ b/web/src/components/forms/SigninForm/SigninForm.tsx
@@ -50,11 +50,14 @@ const SignInForm: React.FC = () => {
         signIn({
           email: username,
           password,
-          onError: () => {
+          onError: ({ message }) => {
             // FIXME: There is weird issue returning wrong error message on submit
             // correlated heavily on this https://github.com/aws-amplify/amplify-js/pull/4427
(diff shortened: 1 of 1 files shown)


web/src/components/forms/SigninForm/SigninForm.tsx
```diff
@@ -50,10 +50,13 @@ const SignInForm: React.FC = () => {
...
           password,
-          onError: ({ message }) =>
-            setErrors({
-              password: message,
-            }),
+          onError: () => {
+            // FIXME: There is weird issue returning wrong error message on submit
+            // correlated heavily on this https://github.com/aws-amplify/amplify-js/pull/4427
+            return setErrors({
+              password: 'Incorrect username or password.',
```
austinbyers, line 57: Good, this is exactly the same as the standard error message

austinbyers: Except here you're actually catching every possible error, can we limit it to just the "radix" error? There could be an internal error with Cognito that we would not want to hide like this

PR #1250, opened 2020-07-28
Feature: Threshold for Rules

### LINKED ISSUE #1109: UI should let users configure threshold alarms for rule
@kostaspap commented on [Thu Apr 30 2020](https://github.com/panther-labs/panther-enterprise/issues/115)

### Description

UI should let users configure threshold alarms for rule

### Acceptance Criteria

- TBD

---

@3nvi commented on [Mon May 04 2020](https://github.com/panther-labs/panther-enterprise/issues/115#issuecomment-623340517)

@kostaspap would this apply for policies as well or only rules?

My guess is that it should be added in the RuleForm by utilizing a combination of 2 fields:

* Number input for number of rules
* Dropdown with preselected time windows (or free number input) 

---

@kostaspap commented on [Tue May 05 2020](https://github.com/panther-labs/panther-enterprise/issues/115#issuecomment-623917276)

> @kostaspap would this apply for policies as well or only rules?

For the time being we are just looking to add this for rules. I don't know if we have usecases and plan in general to add it for policies, I will let @jacknagz comment on it.

> My guess is that it should be added in the RuleForm by utilizing a combination of 2 fields:
> 
> * Number input for number of rules
> * Dropdown with preselected time windows (or free number input)

Sounds good to me. 

---

@3nvi commented on [Tue May 19 2020](https://github.com/panther-labs/panther-enterprise/issues/115#issuecomment-630898504)

@kostaspap this is not ready from your side right?

---

@kostaspap commented on [Tue May 19 2020](https://github.com/panther-labs/panther-enterprise/issues/115#issuecomment-630899460)

> @kostaspap this is not ready from your side right?

Nope, not yet

vorillaz:
## Background

This one implements the first iteration of the UI part regarding the threshold field for rules. Since the UI is tightened with the redesign process I placed the threshold field as it should be inside the rule form. 
Closes #1109.

## Changes

- Added the threshold field for rule details and submission schema.
- Changed the form structure in order to fit the `NumberInput` field.
- Created a utility helper for number formatting.

## Screenshots
Screenshot 2020-07-28 at 3 36 40 PM

image


api/graphql/schema.graphql
```diff
@@ -216,6 +216,7 @@ input AddRuleInput {
 input UpdateRuleInput {
   body: String
   dedupPeriodMinutes: Int
+  threshold: Int
```
alexmylonas, line 219: Should this also be required?

web/src/pages/CreateRule/CreateRule.tsx
```diff
@@ -35,6 +35,7 @@ import { useCreateRule } from './graphql/createRule.generated';
 const initialValues: Required<AddRuleInput> = {
   body: `${DEFAULT_RULE_FUNCTION}\n\n${DEFAULT_TITLE_FUNCTION}\n\n${DEFAULT_DEDUP_FUNCTION}`,
   dedupPeriodMinutes: 60,
+  threshold: 1,
```
alexmylonas, line 38: Since we default old rules to `0`, shouldn't this be also zero?

vorillaz: Zero actually makes no sense since the threshold is set to 1 in the backend, conceptually we need to provide a sensible value for our users

austinbyers: Ah, then I think it's just a question of what the threshold means - the minimum number of matching events which will trigger an alert? In that case, makes sense

api/graphql/schema.graphql
kostaspap: The backend always returns a value, I think it could be required.

web/src/pages/CreateRule/CreateRule.tsx
kostaspap: The backend now returns `0` for old rules that didn't have the `threshold` field configured but I could update it (in a separate PR) to return 1 to avoid confusion. 

> Ah, then I think it's just a question of what the threshold means - the minimum number of matching events which will trigger an alert?

Yes, this is what we mean essentially. If the number of matches are greater or equal to the threshold, we fire an alert. So, threshold of 10 for a rule it means that a rule will fire if we match 10 events. Threshold of 1, means that the rule will fire on the first match. Does it make sense?

PR #1262, opened 2020-07-29
Streaming results for AWS.CloudTrail logs

alxarch:
## Background

There is currenty an issue with `AWS.CloudTrail` logs that require a lot of memory to process. All log events in each file are in a single line. The current way log parsers work require all events for a log line to be parsed and returned in a single call to `ParseLog`. This PR allows to return a special `error` value that signals the log processor to read results in a streaming function.

## Changes

- Adds `parsers.ResultStream` and `parsers.StreamResultsError` interfaces.
- Adds special handling for `parsers.StreamResultsError` errors in the classifier
- Adds `Stream` field to the `ClassificationResult`
- Handles `Stream` classification results in the log processor
- Adds `awslogs.CloudTrailStreamingParser` to return a stream of `AWS.CloudTrail` results.

## Testing

- mage test:ci
- TODO: unit tests for proper stream handling and parser stats
- TODO: deploy for e2e

### DIFF
--- a/internal/log_analysis/log_processor/classification/classifier.go
+++ b/internal/log_analysis/log_processor/classification/classifier.go
@@ -47,6 +47,7 @@ type ClassifierResult struct {
 	Events []*parsers.Result
 	// LogType is the identified type of the log
 	LogType *string
+	Stream  parsers.ResultStream
 }
 
 // NewClassifier returns a new instance of a ClassifierAPI implementation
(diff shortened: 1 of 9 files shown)


internal/log_analysis/log_processor/classification/classifier.go
```diff
@@ -47,6 +47,7 @@ type ClassifierResult struct {
 	Events []*parsers.Result
 	// LogType is the identified type of the log
 	LogType *string
+	Stream  parsers.ResultStream
```
rleighton, line 50: A comment here explain what this is would be nice.

internal/log_analysis/log_processor/classification/classifier.go
```diff
@@ -129,40 +130,51 @@ func (c *Classifier) Classify(log string) *ClassifierResult {
...
-		result.Events = parsedEvents
-
-		// update per-parser stats
-		var parserStat *ParserStats
-		var parserStatExists bool
-		// lazy create
-		if parserStat, parserStatExists = c.parserStats[logType]; !parserStatExists {
-			parserStat = &ParserStats{
-				LogType: logType,
+		// Parser error signifies streaming results
```
rleighton, line 149:
```
// StreamResultsError signifies streaming  results from a JSON document (rather than JSONL)
```

internal/log_analysis/log_processor/parsers/timestamp/timestamp.go
```diff
@@ -69,7 +70,30 @@ func (ts *RFC3339) MarshalJSON() ([]byte, error) {
...
+	if err = (*time.Time)(ts).UnmarshalJSON(jsonBytes); err == nil {
+		return
+	}
+	if isGlueTimestampJSON(jsonBytes) && unmarshalGlueJSON(jsonBytes, (*time.Time)(ts)) == nil {
+		return nil
+	}
+	return
+}
+
+// In order to be able to re-parse panther log events to log event structs we need to detect if the
```
rleighton, line 82: Why do we want this?

internal/log_analysis/log_processor/classification/classifier.go
alxarch: The interface is for not restricted to JSON logs in any way. It just signifies that results for this log line must be streamed. God forbid we encounter single line XML logs for example :)

internal/log_analysis/log_processor/parsers/timestamp/timestamp.go
alxarch: I required this for testing in this PR. This is needed to be able to re-parse an event written to S3 back into a go struct. The timestamp formats translate the JSON value to a string. Log events produced by the log processor cannot currently be re-processed as all the timestamp parsing would fail. This is a usefull thing to have for many scenarios (ie data migrations)

alxarch:
The memory overhead when using this method is multiple times less. Yes we read the while line in memory but the subsequent allocations for all events are a single event each time. Until we completely change the way the current pipeline works we cannot have something that handles this at the io Reader level. This will limit the effect drastically. Note that using a strings. Reader is also important since unmarshal from string augments the problem by converting string to byte slice

alxarch:
I will close this PR for now as it is not high priority and I am investigating some changes to improve the code.

PR #1271, opened 2020-07-30
Declarative log parser definition

### LINKED ISSUE #1182: Standard Field: RequestID

Security events are often spread into multiple log lines and to link these events together, a request ID is generally set to provide context. We should create a standardized data field for request IDs to join across data sets.

### Describe your use cases

Searching the data lake to understand behavior for a given request.

Daily

None

### References

You can see this field in multiple AWS logs today

alxarch:
## Background

This PR refactors the log processor to allow defining parsers in a declarative way.
Parts of the code introduced here were developed on the way to support user-defined logtypes and parsers.
It also changes the point in code that JSON serialization of log event happens. This allows the log processor to handle different requirements posed by different storage backends in the future. Parsers already defined and using `parsers.PantherLog` are kept compatible with the use of `parsers.AdapterFactory` without modification to their code.

## Usage

For the common case of one JSON log per line, definining a new log parser is greatly simplified.

First we define a log event struct

```go

type SomeEvent struct {
  // Note the `panther:"event_time"` tag that instructs Panther to use the timestamp as the event time
  // Also here we use `tcodec` to specify the tag format in the struct without need for extra code.
  Timestamp time.Time `json:"ts" tcodec:"rfc3339" validate:"required" panther:"event_time" description:"The event timestamp"`
  // Note the `panther:"ip"` tag that instructs panther to append the value to `p_any_ip_addresses`
  RemoteIP null.String `json:"remote_ip" panther:"ip" description:"Request remote ip"`
  // Note that there's no `omitempty` option in the `json` tag. Panther knows to always treat fields as omitempty.
  // `null.String` preserves empty strings in the input JSON without the need for pointers
  UserName null.String `json:"user_name" description:"The user name"`
  // More fields ...
}
```

And then we register it with `logtypes`

```go
func init() {
  logtypes.MustRegister(logtypes.Config{
    Name: "Vendor.SomeEvent",
    Description: "Some log event",
    ReferenceURL: "https://example.com/docs/api/some-event",
    Schema: pantherlog.EventStruct(SomeEvent{}),
    NewParser: &parsers.JSONParserFactory{
      LogType: "Vendor.SomeEvent",
      NewEvent: func () interface{} { return &SomeEvent{} },
    },
  })
}
```

Add the appropriate `import` entries to `registry.go` and `constants.ts` and the new log type is ready to be used!

For more complex cases the struct definitition can handle most of the complexity and in combination with `pantherlog.ResultBuilder` and `pantherlog.ValueWriterTo` most parsers should be straight forward to implement/port.

## Changes

- Moves `parsers.Result` into `pantherlog.Result` and adds more fields to fully describe the result.
- Adds `pantherlog.FieldID` enum type that identifies panther-added fields.
- Adds `pantherlog.Result.Meta` to define which panther-added fields are required on a per-logtype basis.
- Adds a `jsoniter` extension to
  - Scan string values with a `panther` tag during JSON encoding to fill panther 'any' fields.
  - Extract event time from timestamps with a `panther:"event_time"` tag
  - Customize encoders for types that implement `pantherlog.ValueWriterTo` interface to add their extracted values to panther 'any' fields during JSON encoding.
- Adds new testing method that tests a parser on input/output JSON directly and allows test-driven development of new parsers.
- Adds `pantherlog.KindTraceID` and the matching `p_any_trace_ids` field to resolve #1182 and demonstrate how to add new 'meta' fields
- Adds the `pantherlog` cli tool to test the log processor pipeline locally

## Testing

- mage test:ci
- the logic was used to produce parsers for a PoC and was tested successfully E2E
- deployed with mage:deploy and E2E tested on the default AWS logs and by uploading gitlab production log samples

The code in this PR is self-contained for the most part. The only change that affects the log processor is the new adapter logic that writes events to the output. This will be tested on deployments to verify it does not affect the writing of results to S3.


internal/log_analysis/log_processor/pantherlog/pantherlog.go
```diff
@@ -0,0 +1,68 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+// EventStruct composes a struct that extends `event` with panther fields
+func EventStruct(event interface{}) interface{} {
+	return defaultMetaFields.EventStruct(event)
+}
+
+// JSON returns a customized jsoniter.API to be used for serializing panthe log events.
```
rleighton, line 36: s/panthe/panther

internal/log_analysis/log_processor/pantherlog/pantherlog.go
```diff
@@ -0,0 +1,68 @@
...
+	EscapeHTML: true,
+	// Validate raw JSON messages to make sure queries work as expected
+	ValidateJsonRawMessage: true,
+	SortMapKeys:            true,
+}.Froze())
+
+// RegisterExtensions registers all pantherlog required extensions on an API
+func RegisterExtensions(api jsoniter.API) jsoniter.API {
+	api.RegisterExtension(jsonutil.NewEncoderNamingStrategy(awsglue.RewriteFieldName))
+	//api.RegisterExtension(jsoniter.EncoderExtension{
```
rleighton, line 51: remove?

internal/log_analysis/log_processor/pantherlog/values.go
```diff
@@ -0,0 +1,174 @@
...
+			continue
+		}
+		c.index[kind] = append([]string(nil), values...)
+	}
+	return c
+}
+
+// Inspect returns a sorted copy snapshot of the value index
+// This is mainly useful for tests.
+func (b *ValueBuffer) Inspect() map[ValueKind][]string {
```
kostaspap, line 88: name it "Sort"/"Sorted"?

internal/log_analysis/log_processor/pantherlog/values.go
```diff
@@ -0,0 +1,174 @@
...
+
+// Get returns field values sorted
+func (b *ValueBuffer) Get(kind ValueKind) []string {
+	switch values := b.index[kind]; len(values) {
+	case 0:
+		return nil
+	case 1:
+		return values
+	default:
+		sort.Strings(values)
```
kostaspap, line 154: I guess we do it here because ValueBuffer it is a write-heavy and thus calling sort.Strings every time we add a value might be slower? Worth writing a comment to this regard.

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec.go
```diff
@@ -315,3 +315,24 @@ func ValidateEmbeddedTimeValue(val reflect.Value) interface{} {
 	}
 	return tm
 }
+
+func NewTimeEncoder(enc TimeEncoder, pointer bool) jsoniter.ValEncoder {
```
kostaspap, line 319: That seems a bit ..weird. Where do we expect to set the `pointer` to true and when to false?

internal/log_analysis/log_processor/pantherlog/meta.go
```diff
@@ -0,0 +1,285 @@
...
+		KindMD5Hash: {
+			FieldName:     "PantherAnyMD5Hashes",
+			FieldNameJSON: "p_any_md5_hashes",
+			Description:   "Panther added field with collection of SHA256 hashes of any algorithm associated with the row",
+			typ:           typStringSlice,
+			kind:          KindMD5Hash,
+		},
+		KindTraceID: {
+			FieldName:     "PantherAnyTraceIDs",
+			FieldNameJSON: "p_any_trace_ids",
```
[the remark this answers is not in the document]
rleighton: Actually, it is common to accumulate ids in tracing microservices. You have the original request id from the frontend, then you get one for the first called service , so you log your request id and the original request id, then you call another service (passing the upstream ids), etc. That said,  I don't have a specific use case we are modeling that is complex like that.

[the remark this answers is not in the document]
rleighton: I did 5min of googling: https://opentracing.io/docs/overview/spans/.

When this came up as an attribute, was it from gitlab? It would be nice to have a specific log(s) to use as examples.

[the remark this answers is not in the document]
kostaspap: Gitlab logs have it (they call it `correlation_id`) and Zeek logs have similar concept (connection UID). What I am afraid of is introducing a new concept (tracing) that hasn't been really investigated/designed as part of a big PR, a concept that we will have to keep around for a while. I'd prefer to have specific usecases/queries/runbook for features like this one.

internal/log_analysis/log_processor/pantherlog/meta_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog/null"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/parsers/timestamp"
+	"github.com/panther-labs/panther/pkg/box"
+)
+
+func TestMetaEventStruct(t *testing.T) {
+	meta := pantherlog.BuildMeta(pantherlog.KindIPAddress)
+	eventStruct := meta.EventStruct(&testEventMeta{})
+
+	columns, names := awsglue.InferJSONColumns(eventStruct, awsglue.GlueMappings...)
```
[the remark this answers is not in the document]
alxarch: We need to make sure that the composed struct produces proper Glue columns. That was one of the main reasons of decoupling awsglue in the previous PR.

internal/log_analysis/log_processor/pantherlog/pantherlog.go
```diff
@@ -0,0 +1,68 @@
...
+	return defaultMetaFields.EventStruct(event)
+}
+
+// JSON returns a customized jsoniter.API to be used for serializing panthe log events.
+func JSON() jsoniter.API {
+	return jsonAPI
+}
+
+var jsonAPI = RegisterExtensions(jsoniter.Config{
+	EscapeHTML: true,
```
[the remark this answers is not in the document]
alxarch: While this adds some overhead in the encoding it avoids a lot of possible JS injection issues when event log JSON is rendered in an HTML template. This is the default behavior for both `jsoniter` and the std `encoding/json` packages.

go.sum
```diff
@@ -27,6 +27,8 @@ github.com/fatih/structtag v1.2.0 h1:/OdNE99OxoI/PqaW/SuSK9uxxT3f/tcSZgon/ssNSx4
 github.com/fatih/structtag v1.2.0/go.mod h1:mBJUNpUnHmRKrKlQQlmCrh5PuhftFbNv8Ys4/aAZl94=
 github.com/globalsign/mgo v0.0.0-20180905125535-1ca0a4f7cbcb/go.mod h1:xkRDCp4j0OGD1HRkm4kmhM+pmpv3AKq5SU7GMg4oO/Q=
 github.com/globalsign/mgo v0.0.0-20181015135952-eeefdecb41b8/go.mod h1:xkRDCp4j0OGD1HRkm4kmhM+pmpv3AKq5SU7GMg4oO/Q=
+github.com/fatih/structtag v1.2.0 h1:/OdNE99OxoI/PqaW/SuSK9uxxT3f/tcSZgon/ssNSx4=
```
austinbyers, line 30: We're not using this library directly, it looks like, it just updated, maybe? I was going to suggest we should just do the reflection ourselves, but I think you already are

internal/log_analysis/log_processor/pantherlog/scanners.go
```diff
@@ -0,0 +1,155 @@
...
+	ScanHostname(dest, u.Hostname())
+}
+
+// ScanHostname scans `input` for either an ip address or a domain name value.
+func ScanHostname(w ValueWriter, input string) {
+	if checkIPAddress(input) {
+		w.WriteValues(KindIPAddress, input)
+	} else {
+		w.WriteValues(KindDomainName, input)
+	}
```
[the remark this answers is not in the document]
alxarch: This needs to distinguish between ip address and domain names. We have a separate scanner for `hostname` that can either be an IP or a domain. We already do this in multiple logs but without a helper function.

internal/log_analysis/log_processor/pantherlog/values.go
alxarch: This is meant to be used for debugging and testing thus the `Inspect` name

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec.go
alxarch: This returns a `jsoniter` encoder for a type or field. The behavior of the encoder/decoder is completely different in each case. We already use this in the tcodec code. Outside of tcodec this would be useful mainly for other `jsoniter.Extension` implementations, specifically in the `UpdateStructDescriptor` method.

go.sum
alxarch: We use the `structtag` library in the extension that injects `omitempty` on each field. This seemed like the safest/cleanest way to inject a tag option.

internal/log_analysis/log_processor/pantherlog/meta.go
```diff
@@ -0,0 +1,285 @@
...
+			FieldName:     "PantherAnyIPAddresses",
+			FieldNameJSON: "p_any_ip_addresses",
+			Description:   "Panther added field with collection of ip addresses associated with the row",
+			typ:           typStringSlice,
+			kind:          KindIPAddress,
+		},
+		KindDomainName: {
+			FieldName:     "PantherAnyDomainNames",
+			FieldNameJSON: "p_any_domain_names",
+			Description:   "Panther added field with collection of domain names associated with the row",
```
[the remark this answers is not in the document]
alxarch: I simplified the meta logic. It was overly complex after many iterations for no good reason

internal/log_analysis/log_processor/pantherlog/meta.go
```diff
@@ -0,0 +1,285 @@
...
+		tag = fmt.Sprintf(`json:"%s,omitempty" description:"%s"`, m.FieldNameJSON, m.Description)
+	}
+	return reflect.StructTag(tag)
+}
+
+type Meta map[ValueKind]*MetaField
+
+func (m *MetaField) IsCore() bool {
+	return m.kind == KindNone
+}
```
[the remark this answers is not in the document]
alxarch: I refactored meta to be simpler. The distinction was not apparent before. Core fields are the 'standard' fields that Panther adds to logs (`p_log_type`, `p_row_id` etc). These are all defined in `meta.go` under `type CoreFields struct`. Other modules can only register 'non-core' fields. These are the optional `p_any_X` fields that are used by scanners.

internal/log_analysis/log_processor/pantherlog/meta.go
```diff
@@ -0,0 +1,285 @@
...
+			Index:     []int{0},
+		},
+		{
+			Anonymous: true,
+			Name:      "PantherLog",
+			Type:      meta.StructType(),
+			Index:     []int{1},
+		},
+	})
+}
```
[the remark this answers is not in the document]
alxarch: This was confusting and was removed in the latest version. Now we build up the struct fields in a single `reflect.StructOf`.

internal/log_analysis/log_processor/pantherlog/result_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+func newBuilder(id string, now time.Time) *pantherlog.ResultBuilder {
+	return &pantherlog.ResultBuilder{
+		LogType:   "TestEvent",
+		NextRowID: pantherlog.StaticRowID(id),
+		Now:       pantherlog.StaticNow(now),
+	}
+}
+
+func TestNewResultBuilder(t *testing.T) {
```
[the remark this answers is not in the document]
alxarch: GoLand locally reports 89% coverage.

internal/log_analysis/log_processor/pantherlog/values.go
alxarch: Yes we need to have the values sorted for JSON. This is used by the custom `jsoniter.ValEncoder` we use for `pantherlog.Result`.

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec.go
alxarch: Updated to pass reflect.Type so the function can handle all compatible time values.

internal/log_analysis/log_processor/pantherlog/meta.go
```diff
@@ -0,0 +1,285 @@
...
+
+var defaultMetaFields = DefaultMetaFields()
+
+type MetaField struct {
+	FieldName     string
+	FieldNameJSON string
+	Description   string
+	typ           reflect.Type
+	kind          ValueKind
+}
```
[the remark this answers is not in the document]
alxarch: Added comment and simplified struct

internal/log_analysis/log_processor/pantherlog/meta.go
```diff
@@ -0,0 +1,338 @@
...
+	tmp := reflect.New(eventType)
+	return tmp.Interface(), nil
+}
+
+func BuildEventTypeSchema(eventType reflect.Type, extras ...FieldID) (reflect.Type, error) {
+	fields, err := extendStructFields(nil, eventType)
+	if err != nil {
+		return nil, err
+	}
+	fields, _ = extendStructFields(fields, reflect.TypeOf(CoreFields{}))
```
kostaspap, line 250: hmmm, why ignore the error? Seems you should be returning it

internal/log_analysis/log_processor/pantherlog/meta.go
```diff
@@ -0,0 +1,338 @@
...
+	sort.Slice(extras, func(i, j int) bool {
+		return extras[i] < extras[j]
+	})
+
+	// Ensure distinct field ids
+	distinct := map[FieldID]bool{}
+	for _, id := range extras {
+		if id <= 0 {
+			return nil, errors.New(`invalid field id`)
```
kostaspap, line 259: I think you can just call `id.IsCore()` here?

internal/log_analysis/log_processor/pantherlog/result.go
```diff
@@ -0,0 +1,107 @@
...
+		},
+		Event: event,
+		Meta:  b.meta(),
+	}, nil
+}
+
+func (b *ResultBuilder) now() time.Time {
+	if b.Now != nil {
+		return b.Now()
+	}
```
kostaspap, line 71: I assume the b.Now() will be not nil only in tests?

internal/log_analysis/log_processor/pantherlog/scanners.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"net"
+	"net/url"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+func init() {
+}
```
kostaspap, line 30: Can remove?

internal/log_analysis/log_processor/pantherlog/scanners.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	u, err := url.Parse(input)
+	if err != nil {
+		return
+	}
+	ScanHostname(dest, u.Hostname())
+}
+
+// ScanHostname scans `input` for either an ip address or a domain name value.
+func ScanHostname(w ValueWriter, input string) {
```
kostaspap, line 127: Maybe just rename to `ScanHostnameIPAddress` to be more obvious.

internal/log_analysis/log_processor/pantherlog/scanners.go
```diff
@@ -0,0 +1,160 @@
...
+
+// ScanValues implements ValueScanner interface
+func (id FieldID) ScanValues(w ValueWriter, input string) {
+	w.WriteValues(id, input)
+}
+
+// IsCore checks if a field id is core
+func (id FieldID) IsCore() bool {
+	return id <= 0
+}
```
kostaspap, line 160: This seems that it should be moved to the same place as `FieldID` -> `internal/log_analysis/log_processor/pantherlog/meta.go`

internal/log_analysis/log_processor/pantherlog/pantherlog.go
alxarch: Fixed

internal/log_analysis/log_processor/pantherlog/pantherlog.go
```diff
@@ -0,0 +1,68 @@
...
+	ValidateJsonRawMessage: true,
+	SortMapKeys:            true,
+}.Froze())
+
+// RegisterExtensions registers all pantherlog required extensions on an API
+func RegisterExtensions(api jsoniter.API) jsoniter.API {
+	api.RegisterExtension(jsonutil.NewEncoderNamingStrategy(awsglue.RewriteFieldName))
+	//api.RegisterExtension(jsoniter.EncoderExtension{
+	//	reflect2.TypeOf(time.Time{}): tcodec.NewTimeEncoder(awsglue.NewTimestampEncoder(), false),
+	//})
```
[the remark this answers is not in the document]
alxarch: Removed

internal/log_analysis/log_processor/pantherlog/jsoniter.go
```diff
@@ -0,0 +1,375 @@
...
+		// values were borrowed
+		result.Values.Recycle()
+	}
+	result.Values = values
+}
+
+func (*resultEncoder) writePantherFields(r *Result, stream *jsoniter.Stream) {
+	if !extendJSON(stream.Buffer()) {
+		stream.WriteObjectStart()
+	}
```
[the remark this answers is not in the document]
alxarch: Well, if the event was `nil` or something that did not resolve to a JSON Object. This is a 'muted bug' TBH. In our codebase we won't encounter this issue unless something is wrong. Did not want to spend the performance overhead here of checking if the `JSON` data is something weird like `"foo"` or `true` or `   {  }  `.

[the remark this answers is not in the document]
alxarch: The ida behind the if here is to be able to write only the 'meta' fields as a proper JSON Object for unit tests by using an empty stream.

internal/log_analysis/log_processor/pantherlog/meta.go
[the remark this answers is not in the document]
kostaspap: Cool.

internal/log_analysis/log_processor/pantherlog/jsoniter.go
```diff
@@ -0,0 +1,375 @@
...
+
+	// Normal result
+	values := result.Values
+	if values == nil {
+		result.Values = BlankValueBuffer()
+	}
+	att := stream.Attachment
+	stream.Attachment = result
+	stream.WriteVal(result.Event)
+	stream.Attachment = att
```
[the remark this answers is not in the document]
alxarch: Because we register the extension globally. Someone externally could have added a different attachment for their purposes. We need ours to render the `Result` as JSON. We put it back once we're done. It's not ideal but it's better than leaving it with the `Result` attached (memory leak, concurrency issues since streams are recycled in a pool) and setting it to `nil` could mess up someone expecting their Attahment back in the end.

internal/log_analysis/log_processor/pantherlog/jsoniter.go
```diff
@@ -0,0 +1,375 @@
...
+		return true
+	}
+	return false
+}
+
+type pantherExt struct {
+	jsoniter.DummyExtension
+}
+
+func (ext *pantherExt) UpdateStructDescriptor(desc *jsoniter.StructDescriptor) {
```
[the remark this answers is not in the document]
alxarch: These are the 'standard' jsoniter extension methods. They are unfortunately undocumented in jsoniter as well :) I will add some description explaining their use by `jsoniter`

internal/log_analysis/log_processor/pantherlog/jsoniter.go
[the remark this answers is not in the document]
kostaspap: Ok makes sense, this was my understanding as well. Maybe you can add a comment to explain that?

internal/log_analysis/log_processor/pantherlog/jsoniter.go
```diff
@@ -0,0 +1,375 @@
...
+const (
+	TagName      = "panther"
+	tagEventTime = "event_time"
+)
+
+var (
+	typValueWriterTo = reflect.TypeOf((*ValueWriterTo)(nil)).Elem()
+	typStringer      = reflect.TypeOf((*fmt.Stringer)(nil)).Elem()
+	typString        = reflect.TypeOf("")
+	typStringPtr     = reflect.TypeOf((*string)(nil))
```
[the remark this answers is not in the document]
alxarch: We don't have a good reason to remove them. We could as well use `string` if we know a field is always there. Maybe removing support for `*string` is fine since it has an impact on allocations and we want to discourage it's use.

internal/log_analysis/log_processor/pantherlog/meta.go
alxarch: Yup, the method was added later. It will make it clearer why we check for negative.

internal/log_analysis/log_processor/pantherlog/jsoniter.go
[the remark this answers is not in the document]
kostaspap: Sounds good.

internal/log_analysis/log_processor/pantherlog/result.go
alxarch: Yes, the purpose of this is testing. Maybe some complicated re-processing of already-processed logs would require sth similar also

internal/log_analysis/log_processor/pantherlog/scanners.go
alxarch: Yup, moved all to a single `init()`

internal/log_analysis/log_processor/pantherlog/scanners.go
kostaspap: Actually I think it is ok as it is. While reading this I had in mind "ScanDomainName" instead of "Hostname". You can leave as is.

internal/log_analysis/log_processor/pantherlog/meta.go
```diff
@@ -0,0 +1,338 @@
...
+		Name:        "PantherAnyMD5Hashes",
+		NameJSON:    "p_any_md5_hashes",
+		Description: "Panther added field with collection of SHA256 hashes of any algorithm associated with the row",
+	})
+	MustRegisterField(FieldTraceID, FieldMeta{
+		Name:        "PantherAnyTraceIDs",
+		NameJSON:    "p_any_trace_ids",
+		Description: "Panther added field with collection of context trace identifiers",
+	})
+	MustRegisterScanner("ip", ScannerFunc(ScanIPAddress), FieldIPAddress)
```
alxarch, line 146: Please review the names registered here

kostaspap: lgtm

internal/log_analysis/log_processor/pantherlog/meta.go
alxarch: Because the only error is if the `reflect.Type` is not a struct and we know it is.

internal/log_analysis/log_processor/pantherlog/result.go
alxarch: Added comments on `ResultBuilder`

internal/log_analysis/log_processor/pantherlog/meta.go
alxarch: Fixed

internal/log_analysis/log_processor/pantherlog/jsoniter.go
```diff
@@ -0,0 +1,375 @@
...
+	e.ValEncoder.Encode(ptr, stream)
+	tm := (*time.Time)(ptr)
+	if tm.IsZero() {
+		return
+	}
+	if result, ok := stream.Attachment.(*Result); ok {
+		result.PantherEventTime = *tm
+	}
+}
+func (*pantherExt) updateStringBinding(b *jsoniter.Binding, tag string, typ reflect.Type) bool {
```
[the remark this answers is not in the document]
alxarch: Added comments

internal/log_analysis/log_processor/pantherlog/jsoniter.go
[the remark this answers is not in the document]
alxarch: Added comments

internal/log_analysis/log_processor/pantherlog/jsoniter.go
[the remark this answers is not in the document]
alxarch: Added comment explaining the swap

internal/log_analysis/log_processor/pantherlog/jsoniter.go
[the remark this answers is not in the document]
alxarch: Added comment

internal/log_analysis/log_processor/pantherlog/scanners.go
alxarch: Moved

internal/log_analysis/log_processor/pantherlog/jsoniter.go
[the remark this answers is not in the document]
giorgosp: If `extendJSON` returns `false`, we will just write an event with just the Panther fields right? Is there any point in this? Should we consider returning an error or panicking if we cannot extend the json?

internal/log_analysis/log_processor/pantherlog/meta.go
giorgosp: What about adding a good comment explaining why we ignore the error or handle it or panic? It is a bit confusing for a new reader, considering as well that we handle the error above.

giorgosp:
Although merged, I have added some small comments for readability should you wish to iterate.

PR #1277, opened 2020-07-31
Fix retries for "read: connection reset by peer"

### LINKED ISSUE #1195: Backend Lambdas should retry network connectivity errors
### Describe the bug

The Log Processor doesn't retry errors of the form  `read: connection reset by peer` 

when publishing messages to SNS Topic. Such errors are not retried by default by the AWS SDK due to this: https://github.com/aws/aws-sdk-go/blob/master/aws/request/connection_reset_error.go#L8-L9

The reason is explained here: https://github.com/aws/aws-sdk-go/issues/3027#issuecomment-567269161

However for our case such issues should be retried. 

This is affecting other Lambdas as well e.g. `panther-snapshot-poller`. 

### Steps to reproduce

It is really hard to reproduce this behavior since this is a transient networking issues.

### Expected behavior

1. Log Processor should retry SNS Publish messages/S3 Put requests if it receives a `read: connection reset by peer` or `use of closed network connection` errors
2. AWS Snapshot Poller should retry requests with `read: connection reset by peer` or `use of closed network connection` error

### Environment

- Panther version or commit: https://github.com/panther-labs/panther/commit/d6638ba264568b2910341357afc72d9dec16f11b

## Additional context

We could use a custom retryer, see example: https://github.com/aws/aws-sdk-go/tree/master/example/aws/request/customRetryer

kostaspap:
## Background

Fixes #1195 

## Changes

- Added a retryer following AWS example: https://github.com/aws/aws-sdk-go/blob/master/example/aws/request/customRetryer/custom_retryer.go#L64-L78

## Testing

- mage test:ci
- Deployed. It probably needs to "burn in" for a while to make sure it doesn't have side effects (e.g. not retrying known errors)

### DIFF
--- a/pkg/awsretry/connection_retryer.go
+++ b/pkg/awsretry/connection_retryer.go
@@ -25,21 +25,24 @@ import (
 	"github.com/aws/aws-sdk-go/aws/request"
 )
 
-// CustomRetryer wraps the SDK's built in DefaultRetryer adding additional
+func NewConnectionErrRetryer() *ConnectionErrRetryer {
+	return &ConnectionErrRetryer{
+		DefaultRetryer: client.DefaultRetryer{},
+	}
+}
+
+// ConnectionErrRetryer wraps the SDK's built in DefaultRetryer adding additional
 // custom features.
 type ConnectionErrRetryer struct {
 	client.DefaultRetryer
 }
 
 // ShouldRetry overrides the SDK's built in DefaultRetryer adding customization
-// to retry `connection reset by peer` and `use of closed network connection` errors
+// to retry `connection reset by peer` errors
 func (r ConnectionErrRetryer) ShouldRetry(req *request.Request) bool {
 	if req.Error != nil {
 		errMsg := req.Error.Error()
 		if strings.Contains(errMsg, "connection reset by peer") {
-			return true
-		}
-		if strings.Contains(errMsg, "use of closed network connection") {
 			return true
 		}
 	}
--- a/deployments/panther_config.yml
+++ b/deployments/panther_config.yml
@@ -114,7 +114,7 @@ Setup:
 
   # Whether or not to create a CloudTrail in the Panther account and monitor.
   # You may want this off if you have org-level CT configured for Panther to avoid ingesting the same data twice.
-  EnableCloudTrail: true
+  EnableCloudTrail: false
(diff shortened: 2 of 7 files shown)


pkg/awsretry/connection_retryer.go
```diff
@@ -0,0 +1,52 @@
...
+type ConnectionErrRetryer struct {
+	client.DefaultRetryer
+}
+
+// ShouldRetry overrides the SDK's built in DefaultRetryer adding customization
+// to retry `connection reset by peer` errors
+func (r ConnectionErrRetryer) ShouldRetry(req *request.Request) bool {
+	if req.Error != nil {
+		errMsg := req.Error.Error()
+		if strings.Contains(errMsg, "connection reset by peer") {
```
rleighton, line 45: nit: this could be:
```
return strings.Contains(req.Error.Error(), "connection reset by peer") 
```

rleighton: ... or you could do it 1 statement, relying on || short cutting
```
return (req.Error != nil && strings.Contains(req.Error.Error(), "connection reset by peer")) || r.DefaultRetryer.ShouldRetry(req)
```

giorgosp: > return (req.Error != nil && strings.Contains(req.Error.Error(), "connection reset by peer")) || r.DefaultRetryer.ShouldRetry(req)


Bikeshedding here but I think this is a bit harder/slower to read

kostaspap:
> Useful change! Would it make sense to add the original reasoning of why such errors are not retried by default as per the [aws/aws-sdk-go#3027 (comment)](https://github.com/aws/aws-sdk-go/issues/3027#issuecomment-567269161)? i.e it is best practice to use this retryer for idempotent operations

Good point! Adding.

pkg/awsretry/connection_retryer.go
kostaspap: Yeah, I was trying different stuff in there, this was a leftover. Switching to `strings.Contains(req.Error.Error(), "connection reset by peer") `. 
Note I am not going to do just `return strings.Contains(req.Error.Error(), "connection reset by peer") ` -> in case the error is different, I default to the DefaultRetryer behavior.

PR #1280, opened 2020-08-03
Factor user prompt code from mage to pkg/prompt

rleighton:
## Background
I will be moving `mage gluesync` into an ops tool so it can be more easily shared with customers.
This change factors our the user prompt code in `mage` as a package in `pkg`

## Changes

- Factor out prompt code
- Bring README.md current

## Testing

- mage test:ci

### DIFF
--- a/pkg/prompt/prompt.go
+++ b/pkg/prompt/prompt.go
@@ -1,0 +1,90 @@
+package prompt
+
+/**
+ * Panther is a Cloud-Native SIEM for the Modern Security Team.
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


pkg/prompt/prompt.go
```diff
@@ -0,0 +1,90 @@
...
+	"os"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+/*
+ Utility package to read input from terminal
```
austinbyers, line 33: Move this to a package-level comment at the top of the file:

```go
// Prompt is a utility package to read input from the terminal
```

pkg/prompt/prompt.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
```
austinbyers, line 29: Do we need to introduce this extra dependency here? I would think the built-in "errors" would suffice, since it's just printing output to the user - there are no stack traces to track, for example

rleighton: Well, this is a library now. Caller might want that but not in any of our use cases. Will switch.

PR #1290, opened 2020-08-04
Refactor cfn stacks and utilities out of mage

rleighton:
## Background

I am moving the functionality of `mage glue:sync` to an ops tool in order to easily enable customers with a binary only deployment to have this tool.

There are no functional changes in this PR.

## Changes

- moved names of stacks and templates to new package `tools/cfnstacks`
- moved most of `tools/mage/util_cfn.go` to new package `pkg/awscfn`

## Testing

- mage
- mage test:ci


pkg/awscfn/awscfn.go
```diff
@@ -0,0 +1,246 @@
+/*
+Helpers that query/manipulate AWS Cloudformation stacks
+*/
```
austinbyers, line 3: Package comments have a specific format:

```go
// Package awscfn provides utilities to query and manipulate AWS CloudFormation stacks
```

tools/cfnstacks/cfnstacks.go
```diff
@@ -0,0 +1,63 @@
+/*
+Declare pubic constants and vars for Panther stacks and templates for use by tools
+*/
```
austinbyers, line 3: Switch to package comment (and be really careful with that second word... 🤣 )

```go
// Package cfnstacks declares public constants...
package cfnstacks
```

tools/mage/teardown.go
```diff
@@ -96,27 +98,27 @@ func destroyCfnStacks(masterStack string) error {
...
 	// The bootstrap stacks have to be last because of the ECS cluster and custom resource Lambda.
 	parallelStacks := []string{
-		appsyncStack,
-		cloudsecStack,
-		coreStack,
-		dashboardStack,
-		frontendStack,
-		logAnalysisStack,
-		onboardStack,
+		cfnstacks.AppsyncStack,
```
austinbyers, line 113: Since `stack` is already in the package name, it's redundant to have it in the variable name:

```
cfnstacks.Appsync, cfnstacks.Cloudsec
```

tools/mage/util_cfn.go
```diff
@@ -71,207 +45,13 @@ func cfnFiles() []string {
...
-}
-
 // Returns the PantherVersion tag for the given stack.
 //
 // Will be blank if the stack or tag does not exist.
 func stackVersion(stack string) (string, error) {
 	response, err := cfn.New(awsSession).DescribeStacks(&cfn.DescribeStacksInput{StackName: &stack})
 	if err != nil {
-		if errStackDoesNotExist(err) {
+		if awscfn.ErrStackDoesNotExist(err) {
```
austinbyers, line 54: You've already moved the rest of this file to another package - might as well move this function, too. Checking the Panther version in a stack is something opstools may want to do someday, right?

If you want to make it slightly more generic, the tag key could be a parameter - in this case `PantherVersion`. Then function would be renamed something like `getStackTag`

pkg/awscfn/awscfn.go
rleighton: Fixed

tools/mage/util_cfn.go
rleighton: Good suggestion.

PR #1294, opened 2020-08-04
Truncate docs link to minor version

austinbyers:
## Background

Our documentation provider only supports syncing 10 branches at a time - for this reason, we will only publish documentation variants for minor versions, not patches. For example, "v1.7" will host the documentation for all "v1.7.X" versions.

To accommodate this change, this PR truncates the docs link from `v/v1.7.0-docs` to `v/v1.7-docs`

## Changes

- Truncate docs link to minor version (instead of patch version)


web/src/constants.ts
```diff
@@ -139,8 +139,9 @@ export const LOG_TYPES = [
 
 export const PANTHER_SCHEMA_DOCS_MASTER_LINK = 'https://docs.runpanther.io';
 
-export const STABLE_PANTHER_VERSION = pantherConfig.PANTHER_VERSION.split('-')[0];
-export const PANTHER_SCHEMA_DOCS_LINK = `${PANTHER_SCHEMA_DOCS_MASTER_LINK}/v/${STABLE_PANTHER_VERSION}-docs`;
+export const STABLE_PANTHER_VERSION = pantherConfig.PANTHER_VERSION.split('-')[0]; // e.g. "v1.7.1"
+export const MINOR_PANTHER_VERSION = STABLE_PANTHER_VERSION.slice(0, 4); // e.g. "v1.7"
```
austinbyers, line 143: I just realized this won't work once we have "v1.10"

Seems the easiest thing to do, then, is to split by the period. Not sure the best way to do this in ts, but I'll push a fix shortly

austinbyers:
Deployed and verified:

Screen Shot 2020-08-04 at 3 53 20 PM

(this link won't work, but it will once v1.7 lands)

austinbyers:
This should also help if we have custom versions for customers - "v1.7.0-special" would now correctly route to "v1.7"

PR #1295, opened 2020-08-05
Fix bug when no permissions where specified for the SQS queue

kostaspap:
## Background

When a user goes through the new SQS source onboarding, Panther will create a new SQS queue for them to send their data to. 
During the onboarding process (and later on during editing of the source) the users can choose which Principal ARNs (IAM roles, users) and which Source ARNS (S3 buckets, SNS Topics) will be allowed to send messages to the SQS queue. 

The user could choose to specify no Principal and source arns - in that case only the principals inside the Panther AWS account could send data to the queue. Current backend code had a bug would result in an error

## Changes

- Renamed `allowedPrincipals` to `allowedPrincipalArns` property to be aligned with `allowedSourceArns` property naming. 
- Fixed bug mentioned above

## Testing

- mage test:ci
- deployed in my account. 
  - Created a new SQS source with no arns specified. Edited the queue to add new permissions. 
  - Created a new SQS source with some arns specified. Edited it to remove the permissions. 
  - Deleted the new SQS source.

### DIFF
--- a/internal/core/source_api/api/sqs_utils.go
+++ b/internal/core/source_api/api/sqs_utils.go
@@ -56,9 +56,9 @@ func SourceSqsQueueArn(integrationID string) string {
 
 // Creates a source SQS queue
 // The new queue will allow the provided AWS principals and Source ARNs to send data to it
-func CreateSourceSqsQueue(integrationID string, allowedPrincipals []string, allowedSourceArns []string) error {
+func CreateSourceSqsQueue(integrationID string, allowedPrincipalArns []string, allowedSourceArns []string) error {
 	queueName := getSourceSqsName(integrationID)
-	policy := createSourceSqsQueuePolicy(allowedPrincipals, allowedSourceArns)
+	policy := createSourceSqsQueuePolicy(allowedPrincipalArns, allowedSourceArns)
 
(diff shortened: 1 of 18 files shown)


internal/core/source_api/api/sqs_utils.go
```diff
@@ -222,6 +227,10 @@ func createSourceSqsQueuePolicy(allowedPrincipals []string, allowedSourceArns []
 		statements = append(statements, statement)
 	}
 
+	if len(statements) == 0 {
```
giorgosp, line 230: super minor: Change to `if len(allowedPrincipalArns) == 0` and exit early by moving it above the for loop

kostaspap: It will have to be `len(allowedPrincipalArns) + len(allowedSourceArns) == 0` (or similar) . I can use that if you prefer, sure.

giorgosp: Ah right, the `allowedSourceArns` part was not visible in the diff. It is super minor anyway, so whatever you feel like.

kostaspap: Sure, I've updated it. Code is slightly more verbose, but a bit easier to read & understand.

Issue #1275, opened 2020-07-31
Move mage glue:sync to an ops tool


rleighton:
### Description
To make it easy to distribute to users of the binary distribution, `mage glue:sync` should become an ops tool called `gluesync`. User documentation needs to be updated accordingly.

PR #1298, opened 2020-08-05
gluesync opstool

rleighton:
## Background
To make it easy to distribute to users of the binary distribution, mage glue:sync should become an ops tool called gluesync. User documentation needs to be updated accordingly.

See issue: https://github.com/panther-labs/panther/issues/1275

## Changes

- moved code from `mage` to the opstool `gluesync`
- added validation to check compiled vs deployed Panther version
- changed regexp match from matching tables to matching log types
- refactored getting the Panther version into a `getGitVersion()` function in `mage`
- fixed the path structure of the devtool `pantherlog` to match build expectations

## Testing

- mage test:ci
- ran tool


docs/gitbook/operations/ops-home.md
```diff
@@ -57,12 +57,17 @@ The steps to view the dashboard:
...
-To build:
+To build from source:
 
 ```yaml
 mage build:tools
 ```
 
+They are also available pre-compiled for each release as assets associated with the github release.
+
 * **compact**: a tool to back fill JSON to Parquet conversion of log data (used when upgrading to Panther Enterprise)
```
austinbyers, line 68: Where is this one? It's not in the release assets - is the code for this only in enterprise?

rleighton: Yes, enterprise

PR #1299, opened 2020-08-05
Make opstools interactive

rleighton:
## Background
This change adds an `-interactive` flag  to `requeue` and `s3queue` to make them easier to use. This was also done in the new opstool `gluesync`.

## Changes

- prompt user for flags unset

## Testing

- mage test:ci
- ran the tools, with and w/out flags

### DIFF
--- a/cmd/opstools/requeue/requeue/main.go
+++ b/cmd/opstools/requeue/requeue/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/panther-labs/panther/cmd/opstools/requeue"
+	"github.com/panther-labs/panther/pkg/prompt"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


cmd/opstools/requeue/requeue/main.go
```diff
@@ -12,16 +12,18 @@ import (
...
 const (
 	banner = "moves messages from one sqs queue to another"
 )
 
 var (
-	REGION = flag.String("region", "", "The AWS region where the queues exists (optional, defaults to session env vars)")
-	FROMQ  = flag.String("from.q", "", "The name of the queue to copy from")
-	TOQ    = flag.String("to.q", "", "The name of the queue to copy to")
+	REGION      = flag.String("region", "", "The AWS region where the queues exists (optional, defaults to session env vars)")
+	FROMQ       = flag.String("from.q", "", "The name of the queue to copy from")
```
kostaspap, line 24: Do you think, in case the FROMQ is a DLQ (`-dlq` suffix), to create as default value for the TOQ the DQL value without the suffix? 
I was thinking that in most cases this tool is going to be used for re-routing messages from a DLQ to an SQS queue, by defaulting to a value (or even enforcing it), we minimize the changes that a user will mistakenly reroute messages from a DLQ to the wrong "primary" queue.

rleighton: Very good idea and simple. I'll do that today.

PR #1292, opened 2020-08-04
Begin CI doc enforcement

austinbyers:
## Background

Gitbooks is primarily designed for a WYSIWYG workflow, but we have a git-first workflow. That means it's easy for us to accidentally add the wrong internal links - every single version of our documentation (including the current one) has several broken links.

I spent several hours in the last release cleaning up broken links, so now I'm automating their detection. This PR is the first half - using regex patterns to extract the links and at least validate their pattern. Subsequent PRs will validate that the linked files and headers actually exist.

Even this simple step already caught a few currently broken links fixed in this PR

Part of #1169 

## Changes

- Add an index for all log types: we need some file to link to which shows our "supported logs", but you can't link to a directory without a README
    - Refactored `mage doc` to be more object oriented to more easily generate this summary page
- Extract all documentation references and validate their patterns - they must be an image, external link, mailto, or an internal reference
    - Added a `mage test:doc` command which also runs as part of `mage test:ci`

## Testing

- Pushed the documentation: https://docs.runpanther.io/v/austin-ci-docs
    - In particular, see the new "supported logs" summary page here: https://docs.runpanther.io/v/austin-ci-docs/log-analysis/supported-logs
- `mage doc test:doc`

Issue #1169, opened 2020-07-14
Check doc links in CI


austinbyers:
### Description

It is very easy to have broken documentation links, especially ones that rely on relative paths. In fact, we have 48 broken links live in our documentation right now: https://github.com/panther-labs/panther/pull/1167

We should add some simple CI checks to block merges with broken links.

### Acceptance Criteria

CI will fail if any of the following are true:

- A link is invalid
    - This could be a relative path, e.g. `[link text](../path/overview.md#Header)`, in which case both the file and the header within that file need to exist.
    - All other paths must be a valid URL. For example, `[https://example.com](example)` is flipped and should fail CI (this happened in our live documentation)
- There are documentation pages not linked in the SUMMARY table of contents
- There are images in the `assets` folder which are not linked anywhere (these should be deleted)

3nvi:
We can also enhance this check by making sure  that the  UI always links to a valid doc (i.e. URL path still exists). We can do that easily with a static analysis, since the code for  the docs  is checked out in  the same repo.

If you want to, the FE team can own the JS related checks :)

austinbyers:
> We can also enhance this check by making sure that the UI always links to a valid doc (i.e. URL path still exists). We can do that easily with a static analysis, since the code for the docs is checked out in the same repo.

This turned out to be fairly complex, but I was able to implement it

3nvi:
> > We can also enhance this check by making sure that the UI always links to a valid doc (i.e. URL path still exists). We can do that easily with a static analysis, since the code for the docs is checked out in the same repo.
> 
> This turned out to be fairly complex, but I was able to implement it

Amazing! Thanks @austinbyers

PR #1302, opened 2020-08-06
Validate links to images and other documentation files

austinbyers:
## Background

#1296 is a perfect example of why we need automated testing for documentation in CI. @kostaspap rearranged some files, which actually broke a navbar link and an embedded image. The only reason I know that is the code in this PR caught the issues.

This PR is part 2/3 (following #1292) to validate documentation links in CI (#1169)

## Changes

- Move document parsing (link + header extraction) into its own function and add clear error messages and unit tests
- Validate that referenced images and documents actually exist at the given path
- Error if there are unused assets
- Bugfix: `mage test:ci` suppresses the output of some noisy tests, even if they failed. For these tests (which print everything to stderr, both passes and failures), we now capture stderr and print it in the event of a failure
- Bugfix: Correct broken links from #1296

## Testing

- `mage fmt test:go test:doc`
- `test:doc` is already running in CI against all of our documentation

## Next PR

- Validate header links
- Ensure all pages are listed in the SUMMARY


docs/gitbook/SUMMARY.md
```diff
@@ -20,16 +20,16 @@
   * [Slack](destinations/slack.md)
   * [SNS](destinations/sns.md)
   * [SQS](destinations/sqs.md)
-* [Analysis]()
+* Analysis
```
austinbyers, line 23: Having empty links is fine, but there's no real reason to have them

docs/gitbook/SUMMARY.md
```diff
@@ -213,7 +213,7 @@
   * [Example Queries](enterprise/data-analytics/example-queries.md)
   * [Data Explorer](enterprise/data-analytics/data-explorer.md)
   * [Automatic Log Compaction](enterprise/data-analytics/log-compaction.md)
-* [SaaS Logs](enterprise/saas-logs.md)
+* [SaaS Logs](enterprise/saas-logs/saas-logs.md)
```
austinbyers, line 216: Should have changed in #1296

docs/gitbook/enterprise/saas-logs/saas-logs.md
```diff
@@ -29,4 +29,4 @@ Panther has direct support for pulling log data from AWS EventBridge, enabling r
 
 To set up in Panther, navigate to `Log Analysis` > `Sources` > `Add Source` and select `Amazon EventBridge`:
 
-![EventBridge](../.gitbook/assets/enterprise/saas-logs/eventbridge.png)
+![EventBridge](../../.gitbook/assets/enterprise/saas-logs/eventbridge.png)
```
austinbyers, line 32: Should have changed in #1296

tools/mage/doc.go
```diff
@@ -131,7 +131,7 @@ func headerAnchor(sectionTitle string) (string, error) {
 	var formattedWords []string
 	for _, word := range strings.Split(sectionTitle, " ") {
 		// Remove/replace special characters
-		//     "AWS.S3ServerAccess" => "aws-s3serveraccess"
+		//     "aws.s3serveraccess" => "aws-s3serveraccess"
```
austinbyers, line 134: The string is already lowercase when it reaches this point (it didn't used to be)

tools/mage/doc_parse.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// web reference - http(s) links
+	webLinkPattern = regexp.MustCompile(`^https?://[A-Za-z0-9.#?&%/:=_-]{5,}$`)
+
+	// email link - based on the same regex we use for emails in CloudFormation parameters
+	emailLinkPattern = regexp.MustCompile(`^mailto:[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}$`)
+
+	// Asset reference - an image in the docs/gitbook/.gitbook/assets folder
+	// For example, "../.gitbook/assets/log-analysis/setup-sns1.png"
+	assetLinkPattern = regexp.MustCompile(`^(?:\.\./)*\.gitbook/assets/[a-z0-9/-]+\.(?:png|jpg)$`)
```
kostaspap, line 66: I see that we have some [pictures](https://github.com/panther-labs/panther/tree/master/docs/img) under `docs/img` folder. Are they there by mistake and thus should be moved/removed? Otherwise we might need to also check those links as well.

tools/mage/doc_parse.go
```diff
@@ -0,0 +1,186 @@
...
+	// find section titles: "## text" at the beginning of a line
+	headerPattern = regexp.MustCompile(`(?:^|\n)#+(.*)`)
+
+	// note: we need the non-greedy variant, hence ".*?"
+	linkPattern = regexp.MustCompile(`!?\[.*?\]\((.*?)\)`) // e.g. "[myfile](path/to/doc.md)"
+
+	// Every link target must be one of the following:
+
+	// web reference - http(s) links
+	webLinkPattern = regexp.MustCompile(`^https?://[A-Za-z0-9.#?&%/:=_-]{5,}$`)
```
kostaspap, line 59: nitpick: I believe you could replace all `A-Za-z0-9` with word character `\w`

tools/mage/doc_parse.go
```diff
@@ -0,0 +1,186 @@
...
+	//
+	// Examples:
+	//    - "#aws-credentials"                       (header in same file)
+	//    - "../enterprise/data-analytics/README.md" (a different document)
+	//    - "development.md#deploying"               (header in another file)
+	//    - ""                                       (file and header are both optional in regex)
+	//
+	// Non-examples:
+	//    - "../log-analysis"           (prevent directory links - link will fail without README)
+	//    - "quick_start.md"            (use "-" instead of "_")
```
kostaspap, line 82: Out of interesting, is there any specific reason to do this or this is just a style we want to enforce?

tools/mage/doc_parse.go
```diff
@@ -0,0 +1,186 @@
...
+					match[0], assetLinkPattern.String()))
+			}
+		} else if webLinkPattern.MatchString(target) {
+			result.WebLinks = append(result.WebLinks, target)
+		} else if emailLinkPattern.MatchString(target) {
+			result.EmailLinks = append(result.EmailLinks, target)
+		} else if docMatch := docLinkPattern.FindStringSubmatch(target); docMatch != nil {
+			if target == "" {
+				// the doc link pattern could match an empty string
+				errs = append(errs, fmt.Sprintf("%s has empty link target", match[0]))
```
kostaspap, line 151: Does it make sense to move this higher up, e.g. after line 134?

3nvi:
I like it 🌮

tools/mage/doc_parse.go
```diff
@@ -0,0 +1,186 @@
...
+				result.DocLinks = append(result.DocLinks, docLink{Path: docMatch[1], Header: docMatch[2]})
+			}
+		} else {
+			// This link couldn't be classified - try to offer a helpful error message for common mistakes
+			if assetLinkPattern.MatchString(target) {
+				errs = append(errs, fmt.Sprintf("%s looks like an image, but is not prefixed with !", match[0]))
+				continue
+			}
+
+			if strings.Contains(target, ".com") || strings.Contains(target, ".io") {
```
rleighton, line 162: Can you add a comment as to why specifically ".com" and ".io"

tools/mage/test_namespace.go
```diff
@@ -130,71 +129,88 @@ func (Test) Doc() {
...
-			}
+			// TODO - validate headers
 		}
 	}
 
-	if errCount > 1 {
-		return fmt.Errorf("test:doc: %d errors", errCount)
+	// Check for unused image assets
+	walk(assetDir, func(path string, info os.FileInfo) {
+		if _, exists := linkedImages[path]; !exists && !info.IsDir() && filepath.Ext(path) != ".DS_Store" {
```
rleighton, line 207: Why would ".DS_Store" end up in `linkedImages` ?

tools/mage/testdata/docs/invalid.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Duplicate
+
+# Invalid Links
+
+![](wrong-image-format.txt)
+
+[empty]() - empty link
+
+[](example.com) - needs to be prefixed with https://
```
rleighton, line 23: or http:// ?

tools/mage/doc_parse.go
austinbyers: `docs/img` are only used for the README - these cannot be used for documentation, `gitbooks` won't be able to find them (I tried once)

tools/mage/doc_parse.go
austinbyers: `\w` is actually `[0-9A-Za-z_]`, which is often more permissive than we need. Here, that would work, but other times we allow only letters and numbers, for example.

`\w` can also change its meaning depending on whether you allow unicode matching or just pure ASCII. So, I usually prefer to be explicit

tools/mage/doc_parse.go
austinbyers: No specific reason, it's just the style we use everywhere today, figured we could keep it that way

tools/mage/doc_parse.go
austinbyers: I'll go ahead and change it where the entire character set overlaps: here and in email link

tools/mage/doc_parse.go
austinbyers: Doesn't really matter - the empty string can only match the doc pattern, not any others. But it'd be a bit easier to read, so sure!

tools/mage/doc_parse.go
austinbyers: Ah, those are just my guess as to the most common links people would add. This whole block is just trying to figure out the most helpful error message we can show.

This particular check may not be very helpful, I'll just let it fall through to the final error message which shows the expected patterns

tools/mage/test_namespace.go
austinbyers: It wouldn't - that conditional says "if the image does *NOT* exist in linkedImages AND it's not a `.DS_Store` file, then complain"

In other words, `mage test:doc` was complaining on my laptop that `.DS_Store` files were not linked, here we're excluding those

austinbyers:
>  I mistakenly thought #1169 was already checking for this type of links ^^

No worries, not blaming you at all, was just pointing out a good example of why I'm doing this work. #1169 just added the regex parsing, it didn't validate that the paths actually existed.

tools/mage/doc_parse.go
austinbyers: On second thought, it would be confusing to fail to match the regex just because an underscore was used. I'll allow it (and then we can also use `\w` like you suggest)

PR #1303, opened 2020-08-06
Simplify pantherlog Result handling and schema generation

alxarch:
## Background

These changes result from observations when using the `pantherlog` module to define parsers and from the requirements of some new parsers regarding timestamp handling. The changes where collected in a single PR to avoid scattering these in multiple PRs lumped together with parser definitions. 

## Changes

- Removes `Meta` field from `Result` since it's use restricted the `ResultBuilder` to a single log type without real benefits.
- Pass log type as an argument to `ResultBuilder` since it was restricting a `ResultBuilder` to a single log type.
- Add `EventTimer` interface for custom event time requirements
- Add `Result.EventIncludesPantherFields` for better workaround visibility
- Add `pantherlog.FieldSet` to handle sets of fields when building the event schema
- Fix bug with embedded structs in `BuildEventSchema` et al.
- Adds `override` option to `panther` struct tag for `event_time` handling.

## Testing

- mage test:ci

### DIFF
--- a/internal/log_analysis/log_processor/pantherlog/jsoniter.go
+++ b/internal/log_analysis/log_processor/pantherlog/jsoniter.go
@@ -24,6 +24,7 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/fatih/structtag"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/modern-go/reflect2"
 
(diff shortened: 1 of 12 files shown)


internal/log_analysis/log_processor/pantherlog/jsoniter.go
```diff
@@ -197,35 +197,44 @@ func (e *customEncoder) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
...
-		if !ok {
+		tag := string(field.Tag())
+		tags, err := structtag.Parse(tag)
+		if err != nil {
+			continue
+		}
+		pantherTag, err := tags.Get(TagName)
+		if err != nil {
 			continue
 		}
```
kostaspap, line 208: These two shouldn't really ever happen at runtime, right? Worth adding some debug/warn logging?

internal/log_analysis/log_processor/pantherlog/meta.go
```diff
@@ -120,63 +121,63 @@ var (
 
 // FieldNameJSON returns the JSON field name of a field id.
 func FieldNameJSON(kind FieldID) string {
-	return fieldNamesJSON[kind]
+	return registeredFieldNamesJSON[kind]
 }
 
 func init() {
-	MustRegisterField(FieldIPAddress, FieldMeta{
+	MustRegisterIndicator(FieldIPAddress, FieldMeta{
```
kostaspap, line 128: I like the new naming scheme. It is more easy to understand 👍

internal/log_analysis/log_processor/pantherlog/jsoniter.go
```diff
@@ -197,35 +197,44 @@ func (e *customEncoder) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
...
 		return false
 	}
-	if tag == tagEventTime {
-		b.Encoder = &eventTimeEncoder{
-			ValEncoder: b.Encoder,
-		}
+
+	b.Encoder = &eventTimeEncoder{
+		ValEncoder: b.Encoder,
+		override:   tag.HasOption(`override`),
```
kostaspap, line 230: 1. `override` could be a constant?
2. Can you add a description (with a usage example ideally) of how this is meant to be used?

internal/log_analysis/log_processor/pantherlog/jsoniter.go
alxarch: The error is just 'tag not found'. It happens at runtime. We just don't care for fields that don't have a `panther` tag.

internal/log_analysis/log_processor/pantherlog/jsoniter.go
alxarch: Added as a constand with example usage

PR #1306, opened 2020-08-06
Adds source id and source label to panther logs

### LINKED ISSUE #1268: Include Source ID and Source Label in parsed events
### Description

Panther log processing engine should include the source ID and source label in the parsed events.

### Acceptance Criteria

- Rules can access source id and the source label
- Source ID and Source label are available to query through Athena
- Documentation has been updated

alxarch:
## Background

Closes #1268

## Changes

- Adds `PantherSourceID` and `PantherSourceLabel` to `parsers.PantherLog`
- Adds `PantherSourceID` and `PantherSourceLabel` to `pantherlog.CoreFields`
- Changes `parsers.Factory` interface to use `parsers.SourceParams` to pass source information to the parser factory.
- Changes the log processor to pass `SourceParams` to the parsers used for classification

## Testing

- mage test:ci
- pantherlog tool produces log events with `p_source_id` and `p_source_label` for all samples in `testdata` dirs
- TODO mage deploy & E2E test

### DIFF
--- a/internal/log_analysis/log_processor/common/common.go
+++ b/internal/log_analysis/log_processor/common/common.go
@@ -76,8 +76,11 @@ func Setup() {
 
 // DataStream represents a data stream that read by the processor
 type DataStream struct {
-	Reader io.Reader
-	Hints  DataStreamHints
+	Reader      io.Reader
+	Hints       DataStreamHints
+	SourceID    string
+	SourceLabel string
+	LogTypes    []string
 	// The log type if known
 	// If it is nil, it means the log type hasn't been identified yet
 	LogType *string
--- a/internal/log_analysis/log_processor/parsers/pantherlog.go
+++ b/internal/log_analysis/log_processor/parsers/pantherlog.go
@@ -56,6 +56,10 @@ type PantherLog struct {
 	PantherEventTime *timestamp.RFC3339 `json:"p_event_time,omitempty" validate:"required" description:"Panther added standardize event time (UTC)"`
 	PantherParseTime *timestamp.RFC3339 `json:"p_parse_time,omitempty" validate:"required" description:"Panther added standardize log parse time (UTC)"`
 
+	// optional (string)
+	PantherSourceID    *string `json:"p_source_id,omitempty" description:"Panther added field with the source id"`
+	PantherSourceLabel *string `json:"p_source_label,omitempty" description:"Panther added field with the source label"`
+
 	// optional (any)
 	PantherAnyIPAddresses  *PantherAnyString `json:"p_any_ip_addresses,omitempty" description:"Panther added field with collection of ip addresses associated with the row"`
 	PantherAnyDomainNames  *PantherAnyString `json:"p_any_domain_names,omitempty" description:"Panther added field with collection of domain names associated with the row"`
@@ -273,10 +277,12 @@ func (pl *PantherLog) Result() *Result {
 		EventIncludesPantherFields: true,
 		Event:                      event,
 		CoreFields: pantherlog.CoreFields{
-			PantherLogType:   unbox.String(pl.PantherLogType),
-			PantherRowID:     unbox.String(pl.PantherRowID),
-			PantherParseTime: ((*time.Time)(parseTime)).UTC(),
-			PantherEventTime: ((*time.Time)(eventTime)).UTC(),
+			PantherLogType:     unbox.String(pl.PantherLogType),
+			PantherRowID:       unbox.String(pl.PantherRowID),
+			PantherParseTime:   ((*time.Time)(parseTime)).UTC(),
+			PantherEventTime:   ((*time.Time)(eventTime)).UTC(),
+			PantherSourceID:    unbox.String(pl.PantherSourceID),
+			PantherSourceLabel: unbox.String(pl.PantherSourceLabel),
 		},
 	}
 }
--- a/internal/log_analysis/log_processor/processor/processor.go
+++ b/internal/log_analysis/log_processor/processor/processor.go
@@ -30,6 +30,7 @@ import (
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/classification"
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/common"
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/destinations"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/logtypes"
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/parsers"
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/registry"
 	"github.com/panther-labs/panther/pkg/metrics"
@@ -56,9 +57,7 @@ var (
 // and forwarding the logs to the appropriate destination. Any errors will cause Lambda invocation to fail
 func Process(dataStreams chan *common.DataStream, destination destinations.Destination) error {
 	factory := func(r *common.DataStream) *Processor {
-		// By initializing the global parsers here we can constrain the proliferation of globals throughout the code.
-		allParsers := registry.AvailableParsers()
-		return NewProcessor(r, allParsers)
+		return NewProcessor(r, registry.Default())
 	}
 	return process(dataStreams, destination, factory)
 }
@@ -182,7 +181,22 @@ type Processor struct {
 	operation  *oplog.Operation
 }
 
-func NewProcessor(input *common.DataStream, parsers map[string]parsers.Interface) *Processor {
+func NewProcessor(input *common.DataStream, registry *logtypes.Registry) *Processor {
+	entries := registry.Entries(input.LogTypes...)
+
+	src := parsers.SourceParams{
+		SourceLabel: input.SourceLabel,
+		SourceID:    input.SourceID,
+	}
+	parsers := make(map[string]parsers.Interface, len(entries))
+	for _, entry := range entries {
+		parser, err := entry.NewParser(src)
+		if err != nil {
+			panic(err)
+		}
+		parsers[entry.Describe().Name] = parser
+	}
+
 	return &Processor{
 		input:      input,
 		classifier: classification.NewClassifier(parsers),
--- a/internal/log_analysis/log_processor/sources/s3.go
+++ b/internal/log_analysis/log_processor/sources/s3.go
@@ -128,12 +128,23 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
 			zap.String("bucket", s3Object.S3Bucket),
 			zap.String("key", s3Object.S3ObjectKey))
 	}()
+	sourceInfo, err := getSourceInfo(s3Object)
+	if err != nil {
+		err = errors.Wrapf(err, "failed to fetch the appropriate role arn to retrieve S3 object %#v", s3Object)
+		return
+	}
 
-	s3Client, sourceType, err := getS3Client(s3Object)
+	if sourceInfo == nil {
+		err = errors.Errorf("there is no source configured for S3 object %#v", s3Object)
+		return
+	}
+
+	sourceType := sourceInfo.IntegrationType
+	s3Client, err := getS3Client(s3Object, sourceInfo)
 	if err != nil {
 		err = errors.Wrapf(err, "failed to get S3 client for s3://%s/%s",
 			s3Object.S3Bucket, s3Object.S3ObjectKey)
-		return nil, err
+		return
 	}
 
 	getObjectInput := &s3.GetObjectInput{
@@ -144,23 +155,17 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
 	if err != nil {
 		err = errors.Wrapf(err, "GetObject() failed for s3://%s/%s",
 			s3Object.S3Bucket, s3Object.S3ObjectKey)
-		return nil, err
+		return
 	}
 
 	bufferedReader := bufio.NewReader(output.Body)
+	contentType, err := detectContentType(bufferedReader)
 
-	// We peek into the file header to identify the content type
-	// http.DetectContentType only uses up to the first 512 bytes
-	headerBytes, err := bufferedReader.Peek(512)
 	if err != nil {
-		if err != bufio.ErrBufferFull && err != io.EOF { // EOF or ErrBufferFull means file is shorter than n
-			err = errors.Wrapf(err, "failed to Peek() in S3 payload for s3://%s/%s",
-				s3Object.S3Bucket, s3Object.S3ObjectKey)
-			return nil, err
-		}
-		err = nil // not really an error
+		err = errors.Wrapf(err, "failed to Peek() in S3 payload for s3://%s/%s",
+			s3Object.S3Bucket, s3Object.S3ObjectKey)
+		return
 	}
-	contentType := http.DetectContentType(headerBytes)
 
 	var streamReader io.Reader
 
@@ -174,12 +179,12 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
 		if err != nil {
 			err = errors.Wrapf(err, "failed to created gzip reader for s3://%s/%s",
 				s3Object.S3Bucket, s3Object.S3ObjectKey)
-			return nil, err
+			return
 		}
 		streamReader = gzipReader
 	} else {
 		err = &ErrUnsupportedFileType{Type: contentType}
-		return nil, err
+		return
 	}
 
 	if sourceType == models.IntegrationTypeSqs {
@@ -187,7 +192,10 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
 	}
 
 	dataStream = &common.DataStream{
-		Reader: streamReader,
+		Reader:      streamReader,
+		SourceID:    sourceInfo.IntegrationID,
+		SourceLabel: sourceInfo.IntegrationLabel,
+		LogTypes:    sourceInfo.LogTypes,
 		Hints: common.DataStreamHints{
 			S3: &common.S3DataStreamHints{
 				Bucket:      s3Object.S3Bucket,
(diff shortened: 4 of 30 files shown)


internal/log_analysis/log_processor/sources/s3_client_cache_test.go
```diff
@@ -62,18 +59,18 @@ func TestGetS3Client(t *testing.T) {
 		return s3Mock
 	}
 
-	marshaledResult, err := jsoniter.Marshal([]*models.SourceIntegration{integration})
-	require.NoError(t, err)
-	lambdaOutput := &lambda.InvokeOutput{
-		Payload: marshaledResult,
-	}
+	//marshaledResult, err := jsoniter.Marshal([]*models.SourceIntegration{integration})
```
rleighton, line 62: why removed? why leave commented?

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
```diff
@@ -86,65 +83,62 @@ func TestGetS3Client(t *testing.T) {
...
-	require.Equal(t, models.IntegrationTypeAWS3, sourceType)
 
-	// verify that we have updated the source with the last time scanned status
-	updateStatusInvokeInput := lambdaMock.Calls[1].Arguments.Get(0).(*lambda.InvokeInput)
-	var updateStatusInput models.LambdaInput
-	require.NoError(t, jsoniter.Unmarshal(updateStatusInvokeInput.Payload, &updateStatusInput))
-	require.Equal(t, "3e4b1734-e678-4581-b291-4b8a176219e9", updateStatusInput.UpdateStatus.IntegrationID)
-	// Verify that the status was updated within the last 1 minute
-	require.True(t, updateStatusInput.UpdateStatus.LastEventReceived.After(time.Now().Add(-1*time.Minute)))
+	//// verify that we have updated the source with the last time scanned status
```
rleighton, line 95: why removed? why leave commented?

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
alxarch: This one thing that I want reviewed by whoever wrote that code in the first place. 
To have the source info available I had to split the function `getS3Client` and introduce `getSourceInfo`.
After the split, these commented out checks were failing since `getS3Client` did not call out to sources lambda.
I think I need to add separate unit tests for `getSourceInfo` and transfer the required mock checks there.
If someone can review the current way that s3_client_cache works and verify the splitting does not affect the intention I can proceed with the tests and remove commented out code.

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
alxarch: see above

internal/log_analysis/log_processor/parsers/pantherlog.go
```diff
@@ -56,6 +56,10 @@ type PantherLog struct {
 	PantherEventTime *timestamp.RFC3339 `json:"p_event_time,omitempty" validate:"required" description:"Panther added standardize event time (UTC)"`
 	PantherParseTime *timestamp.RFC3339 `json:"p_parse_time,omitempty" validate:"required" description:"Panther added standardize log parse time (UTC)"`
 
+	// optional (string)
```
rleighton, line 59: s/optional/required

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
rleighton: Looks fine, you just pulled the first line out of getS3Client "up", right? Go ahead and fix the tests.

internal/log_analysis/log_processor/parsers/pantherlog.go
alxarch: I marked this as optional because it won't be available in old data.

rleighton: True, but that is not what the comments mean ... they are to tell devs this is required in all code, at least that is how I would read it.

alxarch: IIRC the validate required tag changes the table schema generation. Wouldn't that break the queries if a required column was not there? My question is should I update only the comment or add a validate required tag as well?

rleighton: IMO, these new fields should be treated like all the other core fields. The field will always be there in the schema (`gluesync` updates all the schema's and we now do it automatically on deploy) ... it might not be in the data but that is ok.

rleighton: I think you should just remove this comment, like all the other "core" fields above

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
```diff
@@ -86,61 +71,14 @@ func TestGetS3Client(t *testing.T) {
...
-	require.NoError(t, jsoniter.Unmarshal(updateStatusInvokeInput.Payload, &updateStatusInput))
-	require.Equal(t, "3e4b1734-e678-4581-b291-4b8a176219e9", updateStatusInput.UpdateStatus.IntegrationID)
-	// Verify that the status was updated within the last 1 minute
-	require.True(t, updateStatusInput.UpdateStatus.LastEventReceived.After(time.Now().Add(-1*time.Minute)))
-
-	s3Mock.AssertExpectations(t)
-	lambdaMock.AssertExpectations(t)
-}
-
-func TestGetS3ClientUnknownBucket(t *testing.T) {
```
rleighton, line 112: Dropped this test why?

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
```diff
@@ -86,61 +71,14 @@ func TestGetS3Client(t *testing.T) {
...
 	require.NotNil(t, result)
-	require.Equal(t, models.IntegrationTypeAWS3, sourceType)
 
 	// Subsequent calls should use cache
-	result, sourceType, err = getS3Client(s3Object)
+	result, err = getS3Client(s3Object, integration)
 	require.NoError(t, err)
 	require.NotNil(t, result)
-	require.Equal(t, models.IntegrationTypeAWS3, sourceType)
-
```
rleighton, line 99: So all of this validation is now gone ... why?

alxarch: The `getS3Client` signature was changed to take the `integration` argument and return only the s3 client and an error.
In the test we pass the `integration` from a statically defined value in the tests. So there's no point checking the source type.

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
alxarch: The `getS3Client` was changed to get the integration as a parameter. I will should re-add these tests in a separate unit test for `getSourceInfo` as that is where the status is updated. That was the reason I had those in comments in the initial commit. To understand where these checks should be hapenning.

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -144,23 +163,17 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
...
-	// http.DetectContentType only uses up to the first 512 bytes
-	headerBytes, err := bufferedReader.Peek(512)
 	if err != nil {
-		if err != bufio.ErrBufferFull && err != io.EOF { // EOF or ErrBufferFull means file is shorter than n
-			err = errors.Wrapf(err, "failed to Peek() in S3 payload for s3://%s/%s",
-				s3Object.S3Bucket, s3Object.S3ObjectKey)
-			return nil, err
-		}
-		err = nil // not really an error
+		err = errors.Wrapf(err, "failed to Peek() in S3 payload for s3://%s/%s",
```
kostaspap, line 173: nitpick: "failed to detect content type of S3 payload for ...."

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
alxarch: Sorry, confused above/below. Regarding the `TestGetS3ClientUnknownBucket` it won't show up as an error for `getS3Client`. I am adding a separate unit test for `getSourceInfo` to check both the unknown bucket and the update of the status.

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -96,15 +150,7 @@ func init() {
 // getS3Client Fetches
 // 1. S3 client with permissions to read data from the account that contains the event
 // 2. The type of the integration
```
kostaspap, line 152: This description no longer seems accurate.

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -55,8 +57,53 @@ type s3ClientCacheKey struct {
...
+	}
+	return nil
+}
+
+func (c *sourceCacheStruct) reset(sources ...*models.SourceIntegration) {
+	byID := make(map[string]*models.SourceIntegration, len(sources))
+	byBucket := make(map[string][]*models.SourceIntegration, len(sources))
+	for _, src := range sources {
+		byID[src.IntegrationID] = src
+		bucketName := src.S3Bucket
```
kostaspap, line 87: This is not correct - the bucket/prefix field varies depending on the source type (see `getSourceS3Info` that was deleted as part of this PR). 
This needs to be fixed otherwise it will not work for non-S3 integrations.

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
kostaspap: Agreed - we need a unit test to validate this scenario.

kostaspap:
This PR changes the schema of all Glue tables. 
1. I assume we need a script to update the schema of existing tables? If we don't need, has the update been tested?
2. We will also need to update all partitions of existing tables. Any chance we will have problems there @rleighton ? We also need to test if for both JSON and Parquet data as well as Snowflake.

internal/log_analysis/log_processor/common/common.go
```diff
@@ -76,8 +76,11 @@ func Setup() {
...
-	Reader io.Reader
-	Hints  DataStreamHints
+	Reader      io.Reader
+	Hints       DataStreamHints
+	SourceID    string
+	SourceLabel string
+	LogTypes    []string
 	// The log type if known
 	// If it is nil, it means the log type hasn't been identified yet
 	LogType *string
```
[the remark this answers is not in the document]
alxarch: Nope, I can check if it is used anywhere and remove it

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -184,7 +183,22 @@ type Processor struct {
 	operation  *oplog.Operation
 }
 
-func NewProcessor(input *common.DataStream, parsers map[string]parsers.Interface) *Processor {
+func NewProcessor(input *common.DataStream, registry *logtypes.Registry) *Processor {
+	entries := registry.Entries(input.LogTypes...)
```
kostaspap, line 187: 1. Can you add unit test for this. Small change, big difference
2. This has to be tested end to end
3. This line, even small has big implications. Can we just leave it out and make it as part of a different PR?

alxarch:
> This PR changes the schema of all Glue tables.
> 
> 1. I assume we need a script to update the schema of existing tables? If we don't need, has the update been tested?
> 2. We will also need to update all partitions of existing tables. Any chance we will have problems there @rleighton ? We also need to test if for both JSON and Parquet data as well as Snowflake.

I would also like to add that there are various issues that require 'core' schema changes on the TODO pipeline. We should have a call to discuss all the aspects of changing the `core` schema fields either by adding new required fields such as `p_source_id` or by adding/removing indicator fields on a schema.

rleighton:
Let's defer this to early 1.8 so we can burn-in and confirm the schema sync is working and I can test upgrades to snowflake

internal/log_analysis/log_processor/common/common.go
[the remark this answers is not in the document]
alxarch: One of @rleighton 's  tools uses it. Will keep for now.

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
alxarch: Not relevant anymore. Changed the PR to be less intrusive to s3_client_cache. Testing was very hard because of *a lot* of globals everywhere. The previous modifications should be included in a future PR to fix the subpath prefix issue.

internal/log_analysis/log_processor/sources/s3_client_cache.go
alxarch: Not relevant anymore, most of the s3_client_cache modifications were reverted in favor of a less intrusive way to get the source info.

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -184,7 +183,22 @@ type Processor struct {
...
+
+	src := parsers.SourceParams{
+		SourceLabel: input.SourceLabel,
+		SourceID:    input.SourceID,
+	}
+	parsers := make(map[string]parsers.Interface, len(entries))
+	for _, entry := range entries {
+		parser, err := entry.NewParser(src)
+		if err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
alxarch: Fixed

internal/log_analysis/log_processor/sources/s3.go
alxarch: changed

internal/log_analysis/log_processor/sources/s3_client_cache.go
alxarch: Corrected

api/lambda/source/models/integration.go
```diff
@@ -63,6 +63,14 @@ type SourceIntegrationMetadata struct {
 	SqsConfig          *SqsConfig `json:"sqsConfig,omitempty"`
 }
 
+func (info *SourceIntegration) RequiredLogTypes() (logTypes []string) {
```
alxarch, line 66: This should be extended to include log types from pullers on enterprise.
I think it's a good idea to add some similar helper methods to `models.SourceIntegration` and see if we can come up with an interface for it to be used across oss/enterprise. It would avoid spreading logic about how to get stuff out of the source info all over the code base and make changing the model easier.

internal/log_analysis/log_processor/processor/processor.go
alxarch: I suggest we keep it in a comment with a FIXME comment until the 1.7 release is branched out.

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
alxarch: Not relevant anymore as I changed to have less changes on s3 client cache

internal/log_analysis/log_processor/parsers/pantherlog.go
alxarch: Removed

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
alxarch: The tests where complex and I was soon got myself chasing my tail while trying to override globals. I changed the PR to not modify the s3_client_cache so much. It was not relevant to the PR anyway.

internal/log_analysis/log_processor/processor/processor.go
alxarch: Changed

api/lambda/source/models/integration.go
kostaspap: I really like it, very nice design. 

Nitpick: I think you can just name the function `LogTypes()` since we don't distinguish log types between required and optional.

api/lambda/source/models/integration.go
```diff
@@ -63,6 +63,14 @@ type SourceIntegrationMetadata struct {
...
+	logTypes = append(logTypes, info.LogTypes...)
+	if sqsConfig := info.SqsConfig; sqsConfig != nil {
+		logTypes = append(logTypes, sqsConfig.LogTypes...)
+	}
```
kostaspap, line 70: The log types defined for each source are:
* S3 source : `info.LogTypes`
* SQS source: `info.sqsConfig.LogTypes` *only*, so we shouldn't call append.

internal/log_analysis/log_processor/common/common.go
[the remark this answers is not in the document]
kostaspap: I checked the code and it seems that it is easy to remove it and update the tool code to use LogTypes instead. It seems more appropriate, keeping both LogType & LogTypes fields increases the likelihood for bugs.

internal/log_analysis/log_processor/processor/processor.go
kostaspap: After offline discussions, we keeping it in .

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -144,23 +145,17 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
 	if err != nil {
 		err = errors.Wrapf(err, "GetObject() failed for s3://%s/%s",
 			s3Object.S3Bucket, s3Object.S3ObjectKey)
-		return nil, err
+		return
 	}
 
 	bufferedReader := bufio.NewReader(output.Body)
+	contentType, err := detectContentType(bufferedReader)
```
kostaspap, line 152: Nice cleanup

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -57,6 +58,41 @@ type s3ClientCacheKey struct {
...
+}
+
+func (c *sourceCacheStruct) Update(now time.Time, sources []*models.SourceIntegration) {
+	byBucket := make(map[string][]*models.SourceIntegration)
+	for _, source := range sources {
+		bucketName, _ := getSourceS3Info(source)
+		bucketSources := byBucket[bucketName]
+		byBucket[bucketName] = append(bucketSources, source)
+	}
+	// Sort sources for each bucket
```
kostaspap, line 71: 1. Can you expand the comment to explain why the sorting is happening here - to explain the business value. 
2. Can you add a unit test that verifies the expected behavior (picking correct integration by matching S3 keys in priority)

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -57,6 +58,41 @@ type s3ClientCacheKey struct {
 type sourceCacheStruct struct {
 	cacheUpdateTime time.Time
 	sources         []*models.SourceIntegration
```
kostaspap, line 60: Is this field still needed?

PR #1307, opened 2020-08-06
Adds parsers for Teleport.AuditLog

### LINKED ISSUE #1183: Gravitational Teleport Parser

Implement a parser for Gravitiational's [Teleport](https://github.com/gravitational/teleport) [audit log](https://gravitational.com/teleport/docs/admin-guide/#audit-log)

### Describe your use cases

Monitor SSH connections

### References

Listed above

alxarch:
## Background

Closes #1183 

> **Warning**: we need to have some log samples to verify the fields marked as `required` are indeed required otherwise check that we will have no missclassifications by only setting `event` field as `required`  

## Changes

- Adds and registers parsers for Teleport.Auditlog

## Testing

- mage test:ci
- Pending samples from Gravitational to do proper e2e testing


docs/gitbook/log-analysis/log-processing/supported-logs/README.md
```diff
@@ -51,6 +51,8 @@ Panther currently supports 37 security log types across 12 different categories:
  - [Syslog](Syslog.md)
      - [RFC3164](Syslog.md#syslog-rfc3164)
      - [RFC5424](Syslog.md#syslog-rfc5424)
+ - [Teleport](Teleport.md)
```
kostaspap, line 54: The company is Gravitational and we are consuming the audit logs from their Teleport product. 

I think it might  be better to name the logs `Gravitational.TeleportAudit`

internal/log_analysis/log_processor/pantherlog/result.go
```diff
@@ -88,7 +88,7 @@ func (b *ResultBuilder) now() time.Time {
 	if b.Now != nil {
 		return b.Now()
 	}
-	return time.Now()
+	return time.Now().UTC()
```
kostaspap, line 91: In line 80 we call `b.now().UTC()`. I think calling `.UTC()` here is not really needed?

internal/log_analysis/log_processor/parsers/teleportlogs/audit.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+func init() {
+	logtypes.MustRegister(logtypes.Config{
+		Name:         TypeAuditLog,
+		Description:  `Teleport logs events like successful user logins along with the metadata like remote IP address, time and the session ID.`,
+		ReferenceURL: `https://gravitational.com/teleport/docs/admin-guide/#audit-log`,
+		Schema: pantherlog.MustBuildEventSchema(AuditLog{},
+			pantherlog.FieldIPAddress,
+			pantherlog.FieldDomainName,
```
kostaspap, line 42: I believe the schema of the Glue tables needs to always include **all** the indicator columns - I don't think the views will work otherwise. Have you tried querying the views to test this one out (tables inside `panther_views` database)

internal/log_analysis/log_processor/parsers/testutil/testutil.go
```diff
@@ -329,13 +325,13 @@ func UnmarshalResultJSON(data []byte, r *pantherlog.Result, indicators pantherlo
 	if err := jsoniter.Unmarshal(data, &tmp); err != nil {
 		return err
 	}
-	eventTime, err := time.Parse(awsglue.TimestampLayout, tmp.EventTime)
+	eventTime, err := time.Parse(time.RFC3339, tmp.EventTime)
```
kostaspap, line 328: In line 313 you use `time.RFC3339Nano`. Should you use the same here?

internal/log_analysis/log_processor/parsers/teleportlogs/audit.go
rleighton: No, we allow the tables to vary and then view creation will fill in the empty columns to that the columns of the view are the union of all user `p_` columns.

kostaspap: Interesting, thanks for the clarification @rleighton . 

A question for @alxarch then: Would it be possible to have the `MustBuildEventSchema` extract the indicators that are defined in the event schema and avoid having to list them one by one? I guess it would be doable by looking up the `panther` tags, right?  Current approach seems more error-prone

docs/gitbook/log-analysis/log-processing/supported-logs/README.md
```diff
@@ -2,7 +2,7 @@
 <!-- This document is generated by "mage doc". DO NOT EDIT! -->
 
 # Supported Logs
-Panther currently supports 37 security log types across 12 different categories:
+Panther currently supports 38 security log types across 13 different categories:
```
rleighton, line 5: This sentence looks hard to maintain. I'd recommend removing.

internal/log_analysis/log_processor/parsers/teleportlogs/audit.go
```diff
@@ -0,0 +1,93 @@
...
+// >   - session IDs
+// nolint:lll
+type AuditLog struct {
+	// Validates the event type matches the enum described in the docs
+	Event         null.String `json:"event" validate:"required,oneof=auth session.start session.end session.join session.leave session.disk session.network session.command exec scp resize user.login" description:"Event type"`
+	Login         null.String `json:"login" description:"OS login"`
+	Namespace     null.String `json:"namespace" description:"Server namespace. This field is reserved for future use."`
+	ServerID      null.String `json:"server_id" validate:"required" description:"Unique server ID."`
+	SessionID     null.String `json:"sid" validate:"required" panther:"trace_id" description:"Session ID. Can be used to replay the session."`
+	AddressLocal  null.String `json:"addr.local" validate:"required" panther:"net_addr" description:"Address of the SSH node"`
```
rleighton, line 71: If the `.` dot comes out in the column name, we have an issue since that is not allowed. Perhaps you can add another tag called `column`, which is used to override the json tag on marshaling if present:
```
AddressLocal  null.String `json:"addr.local" validate:"required" panther:"net_addr" column:"addr_local", description:"Address of the SSH node"`
```
I would also suggest some validation on schema generation that all column names are legal ANSCI SQL and throw a nice error if not. I don't recall if we do that now but we should.

rleighton: Technically you can have `.` if you always quote column names ... nobody wants to do that and lots of sw gets this wrong so we do not want to generated quote required fields if we can avoid it.

kostaspap: > Technically you can have `.` if you always quote column names ... nobody wants to do that and lots of sw gets this wrong so we do not want to generated quote required fields if we can avoid it.

Good point. Probably we can add the dot `.` to the [field name replacer](https://github.com/panther-labs/panther/blob/38312e05e139a7da5e44ff61dbb069d14ea71aea/internal/log_analysis/awsglue/quirks.go#L31-L39).

internal/log_analysis/log_processor/pantherlog/result.go
alxarch: Yup this should happen in `BuildResult()` only

internal/log_analysis/log_processor/parsers/teleportlogs/audit.go
alxarch: Yes, we already have this inplace :). Was implemented in the 'fixes' PR. I will update the registration to handle that.

internal/log_analysis/log_processor/parsers/testutil/testutil.go
alxarch: If you take a look at the tests you will see that it's easier to test using a more standard format. The functionality of converting to glue format is an overlay added as part of the `tcodec` jsoniter extension only in `common.BuildJSON()`. In tests we use a `tcodec` extension that keeps the JSON input time format. This helps with testing at JSON level because otherwise we need to format all timestamps in each tested sample to the glue format instead of copy-pasting the raw JSON string.

alxarch: `time.Parse` behaves differently than `time.Format`. `time.RFC3339` handles fractions of a second properly. This is the same way the default `MarshalJSON` and `UnmarshalJSON` behavior for `time.Time`. [Marshal uses `time.RFC3339`](https://golang.org/src/time/time.go#L1267) and [Unmarshal uses `time.RFC3339Nano`](https://golang.org/src/time/time.go#L1281).

docs/gitbook/log-analysis/log-processing/supported-logs/README.md
alxarch: It is automatically generated by `mage doc`. Should I remove it?

rleighton: Oh, not if it is automatic. I did not know that (forgot). Nevermind!

internal/log_analysis/log_processor/parsers/teleportlogs/audit.go
alxarch: Added `.` to `awsglue.RenameField`

internal/log_analysis/log_processor/parsers/teleportlogs/audit.go
alxarch: Updated registration for simple JSON logs to be as easy as possible with `logtypes.MustRegisterJSON(logtypes.Desc, func() interface{}) logtypes.Entry`

internal/log_analysis/log_processor/pantherlog/result.go
alxarch: Fixed

docs/gitbook/log-analysis/log-processing/supported-logs/README.md
alxarch: Renamed

docs/gitbook/log-analysis/log-processing/supported-logs/Teleport.md
```diff
@@ -0,0 +1,61 @@
+
+<!-- This document is generated by "mage doc". DO NOT EDIT! -->
+# Teleport
```
kostaspap, line 3: Remove this file

PR #1312, opened 2020-08-06
Use sugared logger for requeue debug messages

rleighton:
## Changes

- Use sugared logger for requeue debug messages

## Testing

- mage test:ci


cmd/opstools/requeue/requeue.go
```diff
@@ -32,7 +31,7 @@ func Requeue(sqsClient sqsiface.SQSAPI, region, fromQueueName, toQueueName strin
 		return errors.Wrapf(err, "cannot find destination queue %s in region %s", toQueueName, region)
 	}
 
-	zap.L().Debug(fmt.Sprintf("Moving messages from %s to %s", fromQueueName, toQueueName))
+	zap.S().Debugf("Moving messages from %s to %s", fromQueueName, toQueueName)
```
austinbyers, line 34: Is there no global logger object? Not a big deal, but `zap.S()` adds a function call and mutex check

rleighton: Does not matter for this, not a hot path.

PR #1313, opened 2020-08-06
Move documentation to enterprise

austinbyers:
## Background

Multiple sections of our documentation are auto-generated, which means that the documentation generated in this repository does not always include enterprise context. Enterprise runbooks and supported logs, in particular, would only be visible from the enterprise repo.

For this reason, I've changed gitbooks to pull documentation from our enterprise repo instead of the open-source.

## Changes

- Remove `docs/gitbook` and `.gitbook.yml`
- Remove `mage doc` and `mage test:doc`
- Unrelated: remove unused README images from `docs/img`
- Unrelated: add cloud security screenshot to README - we have it in `docs/img` but never added it to README

## Testing

- `mage test:ci`


rleighton:
When we develop OSS parsers it would be nice to still generate the docs to see generated text to QA. How do we deal with that?

kostaspap:
> When we develop OSS parsers it would be nice to still generate the docs to see generated text to QA. How do we deal with that?

We could probably keep the doc generation logic and a) skip the writing part b) write to a temporary location e.g. `out/`

austinbyers:
> We could probably keep the doc generation logic and a) skip the writing part b) write to a temporary location e.g. out/

Sure, I'll re-add the `mage doc` command and write to `out/docs` so it can be triggered manually, but it will no longer run as part of `mage fmt` nor will CI be able to validate any links

austinbyers:
I restored `mage doc`, which now generates this tree:

```
out/docs
└── gitbook
    ├── log-analysis
    │   └── log-processing
    │       └── supported-logs
    │           ├── AWS.md
    │           ├── Apache.md
    │           ├── Fluentd.md
    │           ├── GitLab.md
    │           ├── Juniper.md
    │           ├── Lacework.md
    │           ├── Nginx.md
    │           ├── OSSEC.md
    │           ├── Osquery.md
    │           ├── Suricata.md
    │           ├── Syslog.md
    │           └── Zeek.md
    └── operations
        └── runbooks.md
```

I left out the summary file generation, which includes all the complex header anchors but generates just a very simple index that need not be validated here.

Issue #1293, opened 2020-08-04
First user should not fail deployment


austinbyers:
### Describe the bug

We have a `FirstUser` CFN resource to bootstrap the first Panther user. It's actually pretty neat that we can theoretically manage all Panther users with CloudFormation, but then you can get conflicts with the Panther web app.

In particular, if you delete the first user in Panther after deploying, future deployments of Panther will fail (@vorillaz encountered this):

```
Custom::PantherUser FirstPantherUser UPDATE_FAILED: Failed to update resource. lambda error returned: panther-users-api: userID=9063f3ea-276c-4c9f-xxxx-xxxx does not exist
```

There's no easy way to recover from this problem, especially if you're using the pre-packaged deployment.

### Steps to reproduce

1. Deploy Panther
2. Invite a new admin user and login as the new user
3. Delete the original first user
4. Update Panther - update fails

### Expected behavior

It seems clear we should ignore updates to this resource, that's easy enough.

But should we actually just ignore all errors? We would still log an error to trigger monitoring alarms, but not fail the CloudFormation. This is the approach we decided to take with the `InitialAnalysisSet` for example.

Today, if there is a problem creating the first user, the entire deployment will fail and rollback. This seems a bit excessive, since you could just hit the `users-api` Lambda function directly to invite a user if needed.

PR #1314, opened 2020-08-06
PantherUser custom resource only allows creation

austinbyers:
## Background

Panther deployments can completely fail if there is a problem inviting the first user. Usually, this will rollback the entire deployment, which is frustrating since you could just hit `users-api` directly to invite the first user manually.

It gets worse during an update - if you delete the first user in Panther and subsequently redeploy, the update will fail because the user no longer exists.

We initially had support for update and delete because then you could use CloudFormation to manage your Panther users if you wanted! It's a neat idea, but we never officially supported it and no one is likely to need it.

So instead the FirstUser custom resource is becoming create-only, just like the AnalysisSet did for the same reason - problems talking to Panther's own API should not fail the deployment. These custom resources are just a bootstrapping convenience; they are not necessary for a successful deployment.

Closes: #1293 

## Changes

- `PantherUser` custom resource now ignores update and delete operations

## Testing

- Fresh deploy + login + teardown


internal/core/custom_resources/resources/panther_user.go
```diff
@@ -38,40 +37,28 @@ type PantherUserProperties struct {
...
-		return "custom:panther-user:" + userID, outputs, nil
-
-	case cfn.RequestDelete:
-		split := strings.Split(event.PhysicalResourceID, ":")
-		if len(split) < 3 {
-			// invalid ID (e.g. create failed)
-			return event.PhysicalResourceID, nil, nil
+			// We want to log an error, but not fail the CloudFormation and trigger a rollback.
+			// The user can manually invoke users-api to invite a user if this fails.
+			zap.L().Error("failed to invite user", zap.Any("user", props), zap.Error(err))
```
kostaspap, line 50: Until this user is created, the users cannot really access Panther, right? What would be the easiest way they could resolve it, redeploy Panther maybe? Maybe we can add this suggestion in the error message e.g. "failed to invite user, please re-deploy Panther to fix this issue"

rleighton, line 50: Can we add instructions on how to remediate in the log message?

nhakmiller: From my understanding, re-deploying Panther will not resolve this issue because the custom resource doesn't fail to create when the user fails to create (in order to prevent a rollback). So CloudFormation thinks everything is fine. The fix is going to be talking to the users API directly, either through the CLI or the AWS console. Might be easiest to dump a docs link here to more detailed remediation instructions.

That being said, this error message is not going to be seen by users unless they look at the lambda logs which they will have no reason to do because the custom resource will have successfully been created.

So the failure path is:
1. User fails to get created
2. Panther deployment completes successfully with no signs of issues
3. User never receives an invite email
4. Eventually user decides theres a problem and starts investigating
5. Hopefully the user finds this error message or a docs link describing the issue?
6. User talks to the Users API lambda directly to invite the first user

Not very nice unfortunately :/ not sure how we can make it better though.

austinbyers: Right - this error message is buried in the cloudwatch logs, not very helpful for the user. Still, I'll add a note about needing to invite the user manually. It is possible to set a custom CloudFormation resource status to display a message in the CFN console, but that would require going around their own SDK wrapper we are using right now.

There is already a troubleshooting section in the documentation about missing the invite email, I'll update it with a command to invite the user manually:

```
aws lambda invoke --function-name panther-users-api --payload "{...}"
```

Whoever is doing the deployment by definition has access to the AWS infra. Whoever is hosting the deployment, they can use the deployment role to hit the API to invite the missing user.

The invite email is sent as part of the deploy process - if the deploy finishes and you don't have an email within 60 seconds, you can be nearly certain it failed. You don't have to wait long

> From my understanding, re-deploying Panther will not resolve this issue because the custom resource doesn't fail to create when the user fails to create (in order to prevent a rollback).

Right, the whole point of this PR is to *not* fail the entire deployment if there's an issue inviting a user. So a redeploy will not fix the issue for most users (`mage deploy`, from source will fix it, because we check for missing users at the beginning of the deploy process)

austinbyers: Nick had a great summary. The user can detect the error by:

1. Not getting the invite email immediately after deployment
2. The quick-start and troubleshooting docs will be updated with resolution steps
3. There will be a CloudWatch alarm triggered for the custom-resources Lambda and an error message in the logs

PR #1315, opened 2020-08-07
Supporting Rule threshold through Bulk Upload

kostaspap:
## Background

Fixed a couple of issues I found: 
1. Supporting Threshold through Bulk upload operation
2. Using default Threshold value of '1' if none is specified. 

## Changes

- Fixed above issues
- Added integration test for bulk uploading of Rules

## Testing

- mage test:ci
- integration tests


internal/core/analysis_api/handlers/util.go
```diff
@@ -101,22 +101,22 @@ func setDifference(first, second []string) (result []string) {
 
 // Returns true if the two string slices have the same unique elements in any order
 func setEquality(first, second []string) bool {
-	firstMap := make(map[string]bool, len(first))
+	firstMap := make(map[string]struct{}, len(first))
```
kostaspap, line 104: I came across this code... decided to switch to using `map[string]struct{}` instead of `map[string]bool` since this is what we have been using lately.

austinbyers: It's marginally more efficient, because an empty struct has no value and so takes less space than a 1-byte bool

It also prevents bugs where an entry might exist in the map with a `false` value vs not existing at all - it's nice that a `map[string]struct{}` guarantees no meaningful map value.

I've debated adding a simple `type StringSet = map[string]struct{}` to `pkg` somewhere so we could label it a "set"

PR #1325, opened 2020-08-08
Log processor latency metric

nhakmiller:
## Background

We want to be able to display to user's the average difference between event time and parsing time, sometimes calling event latency. This change both generates that data and updates the metrics API so that the frontend can query it.

In order to calculate this average latency, I sum the latency of each event and log it alongside the count of events and count of bytes being logged.

Then, in order to calculate the average latency, I use a CloudWatch expression to divide the total sum of latencies in a given time frame by the total sum of events. This average latency is returned in milliseconds.

It is possible that the latency could be negative, if events are consistently reporting themselves as happening in the future. I did not think it made sense to take the absolute value here because if you have a negative latency you probably want to know that all your clocks are wrong and not have that info be masked.

## Changes

- Log the `CombinedLatency` metric in the log processor
- Add the `eventsLatency` metric to the metrics API. This metric returns time series data

## Testing

- Deployed this into my dev account and tested with the following lambda test event

input:
```
{
  "getMetrics": {
    "metricNames": [
      "eventsLatency"
    ],
    "fromDate": "2020-08-07T23:30:00Z",
    "toDate": "2020-08-07T23:50:00Z",
    "intervalMinutes": 5
  }
}
``` 

response:
```
{
  "eventsProcessed": {
    "seriesData": {
      "timestamps": ["2020-08-07T23:45:00Z", "2020-08-07T23:40:00Z", "2020-08-07T23:35:00Z", "2020-08-07T23:30:00Z"],
      "series": [
        {
          "label": "AWS.VPCFlow latency",
          "values": [419148.67741935485, 583029.8031496063, 409445.38095238095, 552233.3985507246]
        },
        {
          "label": "AWS.ALB latency",
          "values": [158676.5, 0, 0, 254182.33333333334]
        },
        {
          "label": "AWS.S3ServerAccess latency",
          "values": [0, 0, 0, 0]
        }
      ]
    }
  },
  "fromDate": "2020-08-07T23:30:00Z",
  "toDate": "2020-08-07T23:50:00Z",
  "intervalMinutes": 5
}
```


internal/core/metrics_api/api/log_processor_metrics.go
```diff
@@ -86,3 +90,89 @@ func getEventsProcessed(input *models.GetMetricsInput, output *models.GetMetrics
 	}
 	return nil
 }
+
+// getEventsProcessed returns the count of events processed by the log processor per log type
```
nhakmiller, line 94: Need to update this comment copy pasta

PR #1338, opened 2020-08-11
Fix bug with log processor destination S3 object key

alxarch:
## Background

There was a bug introduced when writing `pantherlog.Result` in S3 where all results were placed in a zero timestamp S3 object when the event being written specified the event time via a struct tag. 

## Changes

- Serialize the event *before* picking a bucket to put it into
- Pass the `jsoniter.API` to use as an argument in `CreateS3Destination`
- Adds `s3BufferSet.writeEvent(event *pathnerlog.Result) (*s3Buffer, error)` method to handle the writing of events.

## Testing

- mage test:ci
- e2e tests with mage:deploy and verified query results + bucket tree

### DIFF
--- a/internal/log_analysis/log_processor/destinations/s3.go
+++ b/internal/log_analysis/log_processor/destinations/s3.go
@@ -39,7 +39,6 @@ import (
 	"github.com/panther-labs/panther/api/lambda/core/log_analysis/log_processor/models"
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/common"
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/logtypes"
-	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog"
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/parsers"
 )
 
@@ -78,7 +77,13 @@ func init() {
 	memUsedAtStartupMB = (int)(memStats.Sys/(bytesPerMB)) + 1
 }
 
-func CreateS3Destination(registry *logtypes.Registry) Destination {
+func CreateS3Destination(registry *logtypes.Registry, jsonAPI jsoniter.API) Destination {
+	if jsonAPI == nil {
+		jsonAPI = jsoniter.ConfigDefault
+	}
+	if registry == nil {
+		registry = logtypes.DefaultRegistry()
+	}
 	return &S3Destination{
 		s3Uploader:          common.S3Uploader,
 		snsClient:           common.SnsClient,
@@ -87,6 +92,7 @@ func CreateS3Destination(registry *logtypes.Registry) Destination {
 		maxBufferedMemBytes: maxS3BufferMemUsageBytes(common.Config.AwsLambdaFunctionMemorySize),
 		maxDuration:         maxDuration,
 		registry:            registry,
+		jsonAPI:             jsonAPI,
 	}
 }
 
@@ -129,6 +135,7 @@ type S3Destination struct {
 	maxBufferedMemBytes uint64 // max will hold in buffers before ejection
 	maxDuration         time.Duration
 	registry            *logtypes.Registry
+	jsonAPI             jsoniter.API
 }
 
 // SendEvents stores events in S3.
@@ -154,7 +161,7 @@ func (destination *S3Destination) SendEvents(parsedEventChannel chan *parsers.Re
 
 	// accumulate results gzip'd in a buffer
 	failed := false // set to true on error and loop will drain channel
-	bufferSet := newS3EventBufferSet()
+	bufferSet := newS3EventBufferSet(destination.jsonAPI)
 	eventsProcessed := 0
 	zap.L().Debug("starting to read events from channel")
 	for event := range parsedEventChannel {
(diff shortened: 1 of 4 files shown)


internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -144,8 +144,11 @@ func streamEvents(sqsClient sqsiface.SQSAPI, deadlineTime time.Time, event event
 		}
 	}()
 
+	// Use a properly configured JSON API for Athena quirks
+	jsonAPI := common.BuildJSON()
+	registeredLogTypes := registry.Default()
 	// process streamChan until closed (blocks)
```
kostaspap, line 150: You are already initializing these in cmd/devtools/logprocessor/logprocessor/main.go . Can't you just use those ones?

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -175,34 +182,17 @@ func (destination *S3Destination) SendEvents(parsedEventChannel chan *parsers.Re
 			})
 		default: // makes select non-blocking
 		}
-
-		buffer := bufferSet.getBuffer(event)
-
-		err := bufferSet.addEvent(buffer, event)
+		buffer, err := bufferSet.writeEvent(event, maxS3BufferSizeBytes, int(destination.maxBufferedMemBytes))
```
kostaspap, line 185: Inside the `writeEvent` method you are re-casting this to `uint64`. I would recommend just updating the `writeEvent` method signature to accept `uint64`.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -331,14 +321,52 @@ func (destination *S3Destination) getS3ObjectKey(logType string, timestamp time.
...
+		stream: stream,
+		set:    make(map[time.Time]map[string]*s3EventBuffer),
 	}
 }
 
+func (bs *s3EventBufferSet) writeEvent(event *parsers.Result, maxBufferSize, maxTotalSize int) (buf *s3EventBuffer, err error) {
+	stream := bs.stream
+	stream.Reset(nil)
+	stream.WriteVal(event)
+	err, stream.Error = stream.Error, nil
```
kostaspap, line 341: Why do we have to do this assignment?

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -331,14 +321,52 @@ func (destination *S3Destination) getS3ObjectKey(logType string, timestamp time.
...
 	return &s3EventBufferSet{
-		set: make(map[time.Time]map[string]*s3EventBuffer),
+		stream: stream,
+		set:    make(map[time.Time]map[string]*s3EventBuffer),
 	}
 }
 
+func (bs *s3EventBufferSet) writeEvent(event *parsers.Result, maxBufferSize, maxTotalSize int) (buf *s3EventBuffer, err error) {
+	stream := bs.stream
+	stream.Reset(nil)
```
kostaspap, line 339: Is this required?

internal/log_analysis/log_processor/destinations/s3.go
alxarch: stream.Reset does not reset the error. If an error occurred during JSON marshal it would be sticky because we re-use the `stream`

internal/log_analysis/log_processor/destinations/s3.go
alxarch: The uint64 is a mistake in the first place. It should be changed to `int64` to have 'free' casting on x64. I kept the `uint64` since it affects metrics code I did not want to add that change into this PR. Just chose the correct signature so we don't keep using `uint64` for no good reason.

alxarch: To elaborate, using `uint64` can cause the opposite of what it (I suppose) is trying to achieve. The fact that the number will not be negative does not mean we *have* to use uint variant. By using the uint variants we risk silent overflows when serializing to JSON and writing to other platforms and at the same time we get the performance penalty  of bounds checks on each required casting.

internal/log_analysis/log_processor/processor/stream.go
alxarch: No, each place where we need this we make a new local instance. The behavior is the same. That is the reason `BuildJSON` is a function and not a global. Basically the target is to instanciate once on each `main.go`  file and pass as argument to the parts of the code that require it.

alxarch: Currently the only exception is the `parsers.JSONFactory` where we have another module local instance there. It does not affect the program as that instance is used only for decoding.

internal/log_analysis/log_processor/destinations/s3.go
alxarch: The linter was complaining about function size when all the changes were inlined in the outer loop. The refactoring was made so that the logic of deciding which buffer needs flushing is inside a single function.

internal/log_analysis/log_processor/destinations/s3.go
kostaspap: I don't see where we serialize the field to JSON, so I am not sure where lies your concern. My suggestion was to remove the casting all together.

internal/log_analysis/log_processor/processor/stream.go
kostaspap: Nevermind, I thought that the `cmd/devtools/logprocessor/logprocessor/main.go` was actually the "main" method of the log processor, so I thought you were instantiating it a second time instead of reusing an existing instance.

internal/log_analysis/log_processor/destinations/s3.go
alxarch: We do it implicitly in the metrics code where we write to CloudWatch

PR #1342, opened 2020-08-11
Standardize `tcodec` behavior when no struct tags are set on the field

alxarch:
## Background

The `tcodec` extension had some weird behavior when no tag was set on a struct field. This PR enforces the default method for JSON decoding/encoding as used in `encoding/json` module for fields that have no `tcodec` tag. It also limits `tcodec` to only work on `time.Time` and `*time.Time` fields to not have overlap with user-defined timestamps.

## Changes

- Adds `tcodec.StdCodec`
- Only handles `time.Time and `*time.Time` fields
- Add a work-around to allow other extensions that decorate encoders/decoders to use `tcodec` as their base no matter what the order of registration is.

## Testing

- `mage test:ci`


internal/log_analysis/log_processor/pantherlog/tcodec/extension.go
```diff
@@ -118,67 +118,49 @@ func (ext *Extension) UpdateStructDescriptor(desc *jsoniter.StructDescriptor) {
...
+				continue
 			}
+		} else if codec = ext.config.DefaultCodec; codec == nil {
+			// Use std codec for time values without a `tcodec` tag
+			codec = StdCodec()
 		}
+
 		dec, enc := ext.split(codec)
-		if dec != nil {
+		vDec, vEnc := NewTimeDecoder(dec, typ), NewTimeEncoder(enc, typ)
```
rleighton, line 136: Can you add a comment explaining what is happening?

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec.go
```diff
@@ -293,53 +297,94 @@ func (d *locDecoder) DecodeTime(iter *jsoniter.Iterator) time.Time {
...
-	if tm.IsZero() {
-		return nil
-	}
-	return tm
-}
-
 func NewTimeEncoder(enc TimeEncoder, typ reflect.Type) jsoniter.ValEncoder {
-	if !isTimeType(typ) {
+	if enc == nil {
 		return nil
```
rleighton, line 302: What does it mean to a caller to get back a `nil`?

internal/log_analysis/log_processor/pantherlog/tcodec/extension.go
alxarch: Will do

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec.go
alxarch: It would mean that no time codec was resolved.

internal/log_analysis/log_processor/pantherlog/tcodec/extension.go
```diff
@@ -259,3 +241,28 @@ func (dec *jsonTimePtrDecoder) Decode(ptr unsafe.Pointer, iter *jsoniter.Iterato
...
+}
+func (c *errCodec) Encode(_ unsafe.Pointer, stream *jsoniter.Stream) {
+	stream.Error = c.err
+}
+func (c *errCodec) Decode(_ unsafe.Pointer, iter *jsoniter.Iterator) {
+	iter.ReportError(c.operation, c.err.Error())
+}
+
+// Force error reporting when decorating
+func (c *errCodec) DecorateEncoder(_ reflect2.Type, _ jsoniter.ValEncoder) jsoniter.ValEncoder {
```
rleighton, line 261: These functions serve what purpose?

alxarch: These are the workaround described in the PR. We need to have a method to detect decorated encoders regardless of order of extension registration. It is a limitation of `jsoniter`. If for example `pantherlog` extension decorates the `time.Time` encoder after it finds a `panther:"event_time"` tag and *then* `tcodec` tries to enforce the `tcoder:"blah"` tag it could overwrite the `pantherlog` behavior. This adds a way specify a `jsoniter.ValEncoder/Decoder` as a decoration and to be able to change it's underlying encoder. It is the only way I could think of to overcome this limitattion. In the case of `errCodec` we use this in the opposite way to make sure that the `unregistered codec` error is always reported.

alxarch:
This PR's target is to limit the ways in which `tcodec` could have unpredictable behavior so that corner cases do not result in hard to detect bugs in the JSON processing pipeline.

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec.go
alxarch: The purpose of this funtion is to provide `jsoniter` encoders/decoders for other modules to register with `jsoniter`. For example:
```go
func init() {
  typ := reflect.TypeOf(customTimestamp{}) 
  jsoniter.RegisterTypeEncoder(typ.String(), tcodec.TimeEncoder(tcodec.UnixCodec(), typ))
}

PR #1347, opened 2020-08-11
fix: show only open/triaged alerts by default

### LINKED ISSUE #1346: Alert filter OPEN and TRIAGED statuses by default
### Description

When a user visits a page that loads and shows a list (table) of alerts, we currently show all statuses. However, the default should be to hide any CLOSED or RESOLVED alerts from the UI.

### Acceptance Criteria

- Any place where there is a table populated with alerts should automatically filer out any CLOSED or RESOLVED alerts.
- The user should still be able to configure and override the default settings.

s0l0ist:
## Background

Closes #1346 

## Changes

- ~Added default filtering to the Apollo query~
- ~Added default values for the status filter to show in the form~
- Added default URL parameters to two links.

## Testing

- mage test:ci
- Shouldn't be a breaking change regardless of unit tests
- Inspect visually that loading pages, with and without URL parameters works as expected
- Pagination holds the default queries as well as specific filters
- Can override the status filters to use Open/Triaged or none at all which has the same effect


3nvi:
Filters are derived from URL params, so we need to keep them synced at all times. Correct me if I'm wrong, but the solution above doesn't respect URL params, thus there is a state inconsistency between what filters are applied and what URL params are present.

The solution that  I have in my mind is actually much simpler than that (due to the connection that API filters and URL params natively have).

All  you have to do is make sure to change the default URL that we have for alert listing, to include the URL params that control the severity filters. This change could be easily implemented globally by changing the URI that `urls.logAnalysis.alerts.list()` resolves to, to include some additional url params.

@s0l0ist what  do you think?

s0l0ist:
This solution does respect the URL parameters at all times if they're present. By **default** if no parameters are specified, it will internally (in the frontend) filter out CLOSED & RESOLVED from the results without changing the URL. As soon as other  filters are 'Applied' using the button, the additional parameters will be added to the URL (and overwrite the default `status` parameters in the query with the ones provided in the URL).

I initially thought about changing the URL, but what were to happen to old links? How would the frontend translate an empty `status` url (request parameter is not present) into the one that can initially query the backend with the proper default parameters? Is having a URL with no parameters a problem we should think about? If not, then I'd be happy to implement this much more simple change.

s0l0ist:
> Filters are derived from URL params, so we need to keep them synced at all times. Correct me if I'm wrong, but the solution above doesn't respect URL params, thus there is a state inconsistency between what filters are applied and what URL params are present.
> 
> The solution that I have in my mind is actually much simpler than that (due to the connection that API filters and URL params natively have).
> 
> All you have to do is make sure to change the default URL that we have for alert listing, to include the URL params that control the severity filters. This change could be easily implemented globally by changing the URI that `urls.logAnalysis.alerts.list()` resolves to, to include some additional url params.
> 
> @s0l0ist what do you think?

@3nvi I've implemented your suggestion. There's just one caveat - the `Filter Options` card will always be present because of the new URL parameters that trigger it to be shown. It also makes the transition animation from the homepage `View Alerts` link to the `ListAlerts` page not pull to the top of the page all the time (sometimes it will be pulled down a bit and I have to scroll up to the top).

PR #1359, opened 2020-08-13
Add identity account parameter to deployment role

nhakmiller:
## Background

Often, the account brokering access to IAM roles is not the same account as the one that IAM roles exist in. We have this setup internally. This change allows an identity account to be specified that will be used in the assume role policy document.

## Changes

- Added parameter to deployment role

## Testing

- None


deployments/auxiliary/cloudformation/panther-deployment-role.yml
```diff
@@ -18,7 +18,13 @@ AWSTemplateFormatVersion: 2010-09-09
...
 
 Metadata:
-  Version: v1.6.0
+  Version: v1.7.0
+
+Parameters:
+  IdentityAccountId:
+    Type: String
+    Description: The account ID for the account the deployment role will be assumed from
+    Default: ''
```
austinbyers, line 27: An empty value will currently make the template unusable, since the principal below will have an invalid ARN. We don't want the default template to be broken.

But I agree we also don't want to always force users to put a value here, it should default to the current account as it did before. The way to do that is to build the principal conditionally:

```yml
Parameters:
  IdentityAccountId:
    Type: String
    Description: The accountID which can assume the deployment role (defaults to the same account)
    Default: ''

Conditions:
  AccountSpecified: !Not [!Equals [!Ref IdentityAccountId, '']]

Resources:
  DeploymentRole:
    # ....
    Statement:
      - Effect: Allow
         Principal: 
           AWS: !If
             - AccountSpecified
             - !Sub arn:${AWS::Partition}:iam::${IdentityAccountId}:root
             - !Sub arn:${AWS::Partition}:iam::${AWS::AccountId}:root
```

nhakmiller: Sounds good, updated with these changes.

austinbyers:
Per https://github.com/panther-labs/panther-enterprise/pull/541, might as well change the cloudformation permission to `panther*` stacks instead of `panther-*` while we're in the deployment role

PR #1363, opened 2020-08-13
Don't run go tests in isolation in CI

austinbyers:
## Background

On the last [CI master build](https://app.circleci.com/pipelines/github/panther-labs/panther/6406/workflows/1ade9aaa-806a-49cc-813c-e4600726604f/jobs/15375):

* `mage test:go` took 138 seconds (wall clock time)
* `mage test:web` took 78 seconds (wall clock time)

Unfortunately, these run sequentially - `test:go` runs by itself before `test:web` runs along with all the other tests. This means the total CI time is bounded below by the *sum* of our two slowest tests - not ideal.

@3nvi has suggested separating the CI task into separate components, so `test:go` and `test:web` can run in separate CI containers in parallel, for example. A great idea which we will definitely implement at some point! But that's a bit more work than we have time for at the moment, so this PR offers a very quick fix in the meantime:

When running in CI, let all the tests run in parallel. This is how we used to do it and it worked well. We changed to let go tests run by themselves just to avoid locking up the dev laptops. But there's no reason CI can't continue to run everything in parallel.

## Changes

- When running in CI, go tests run in parallel with all the other tests

## Testing

This should reduce the total CI time by about 60 seconds (25%), not bad for 18 lines of code :) I'll update with a comment when CI has finished to compare the time to the last master build

### DIFF
--- a/tools/mage/test_ci.go
+++ b/tools/mage/test_ci.go
@@ -40,15 +40,29 @@ func (Test) CI() {
 	fmtErr := testFmtAndGeneratedFiles()
 
 	// Go unit tests and linting already run in multiple processors
-	goUnitErr := testGoUnit()
-	goLintErr := testGoLint()
+	// When running locally, test these by themselves to avoid locking up dev laptops.
+	var goUnitErr, goLintErr error
+	if !runningInCI() {
+		goUnitErr = testGoUnit()
+		goLintErr = testGoLint()
+	}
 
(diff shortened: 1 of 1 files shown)


austinbyers:
Worked even better than I thought! Shaved off 40% of the runtime:

mage test_ci runtime

(although this was only one measurement, far from scientific)

PR #1372, opened 2020-08-17
CI: Add mage_fmt job

3nvi:
## Background

Everytime we open a PR we - sometimes - forget to run `mage fmt`, so after 5 mins (when the  CI fails) we have to context-switch, run `mage fmt` and then push back the branch only to  wait another 5 mins for the  full CI suite to run.

We can save a lot of our devs time by automating the license headers creation. Currently, the presence of a CLI command is not enough, since this is something that should be “there”  without anyone  asking themselves “have I added licenses?”

With this in mind I have proposed the following options:

- Write a CI job that will run `mage fmt` on every PR opening and - if license headers were missing -  will add another commit to the head

- Re-write `mage fmt` to accept a list of files as arguments  & make it run against the checked out files of every  commit. 

## The problem

The best way to solve  it is through pre-commit hooks, but currently  `mage` doesn't accept any inline arguments. This means that we'd have to run `mage fmt` against the entire codebase instead of **just** the  files that have been checked/added to the commit.

This would bring a significant delay to every commit (~15  secs), making developers commit less and less frequently. The alternative was to isolate & execute the underlying commands that `mage fmt` runs under  the hood. That involved some additional work, while also being prone to errors (i.e. if `mage fmt`  adds something else, we need to remember to add it to a pre-commit hook)

##  The solution for now

While not  perfect, adding a `mage fmt`  step to  the CI can ensure the following:

- no developer has to remember if they ran `mage fmt`  or not
- CI checks will fail less frequently causing faster iterations and merging
- **Synced PRs to Enterprise will have their license headers updated automatically** meaning less time spent on making sure the headers are updated when adding new files to both repos.

Downside is that `mage_test_ci` needs to wait for `mage_fmt` to run (so that it runs against the items that are potentially modified by `mage_fmt`). This delay shouldn't be "huge" and is estimated at around 40-55s

## Changes

- Isolate the internals `mage_test_ci` into separate re-usable CircleCI commands
- Add a `mage_fmt` step
- Make  `mage_fmt` run before `mage_test_ci`

## Testing

- Local CircleCI CLI
- Current CI tests (which run `mage_fmt` if you see the checks below)


3nvi:
The PR is failing because the key used by CircleCI is a read-only key. Before we provide a read-write key,  I want to see if you feel that the work present in this PR is beneficial to the team

.circleci/config.yml
```diff
@@ -64,6 +83,62 @@ jobs:
...
+
+  mage_test_ci:
+    executor: panther-buildpack
+    resource_class: xlarge
+    steps:
+      - checkout
+      - restore_go_modules_from_cache
+      - restore_node_modules_from_cache
+      - restore_python_modules_from_cache
+      - restore_binaries_from_cache
```
austinbyers, line 136: Restoring all the caches takes 20-40 seconds, and now it's running twice - once in `fmt` and again in `test_ci`.

For that reason, combined with the fact that you always run them sequentially anyway, why not just combine them into the same logical task? That way, you only have to do all of the expensive setup once

```yml
mage_test_ci:
  steps:
    - checkout
    - restore modules ... 
    - mage fmt
    - git commit
    - rest of testing
```

We don't really need to worry about CI tasks being "re-usable"

.circleci/config.yml
```diff
@@ -64,6 +83,62 @@ jobs:
...
+          # "mage setup" will be a no-op unless new dependencies have been introduced.
+          name: Install new dependencies
+          command: mage setup
+      - persist_go_modules_to_cache
+      - persist_node_modules_to_cache
+      - persist_python_modules_to_cache
+      - persist_binaries_to_cache
+      - run:
+          name: Run test suite
+          command: mage fmt
```
austinbyers, line 118: FYI: In enterprise, you can make this `mage fmt doc` to also commit updates to the auto-generated documentation

.circleci/config.yml
```diff
@@ -64,6 +83,62 @@ jobs:
...
+      - restore_modules_from_cache
+      - run:
+          # By this point, mage is compiled and all dependencies have been restored from the cache.
+          # "mage setup" will be a no-op unless new dependencies have been introduced.
+          name: Install new dependencies
+          command: mage setup
+
+# The pool of jobs that that our CI will be able to run
+jobs:
+  mage_fmt:
```
austinbyers, line 98: With `master` branch protection, CircleCI will never be able to push to master (which is good, that's what we want).

Can we disable this job completely for the master branch?

austinbyers:
It's worth noting that `mage fmt` is now actually running twice - once in the first task to commit any changes and then again in `mage test:ci` to make sure nothing changed.

Eventually, we could disable formatting in the second step in CI, but not right now. There are some auto-generated files not covered by `fmt` (e.g. `mage build:api`) and it only takes ~10 seconds to format

.circleci/config.yml
3nvi: yeah I wanted some isolation, but at the end of the day, what you propose is better

.circleci/config.yml
3nvi: I don't  think the job will ever run against master. The jobs only run against  branches that  have open PRs.  This is not the default behavior of circleci, but it's what  we have active for our repos.

I'll add it as a fallback though, in case we decide to change it

3nvi:
> It's worth noting that `mage fmt` is now actually running twice - once in the first task to commit any changes and then again in `mage test:ci` to make sure nothing changed.
> 
> Eventually, we could disable formatting in the second step in CI, but not right now. There are some auto-generated files not covered by `fmt` (e.g. `mage build:api`) and it only takes ~10 seconds to format

Yeah I was well aware of that and the reason I didn't remove the duplication is exactly what  you wrote above. Thanks though!

.circleci/config.yml
austinbyers: > The jobs only run against branches that have open PRs

Not true - CircleCI runs against the master branch with every change: https://app.circleci.com/pipelines/github/panther-labs/panther?branch=master

3nvi: you're right! I missed the small letters on this

Screen Shot 2020-08-17 at 10 26 18 PM

alxarch (approved):
This should come in handy

alxarch:
@3nvi Another approach to these automations I have seen in other projects is to respond on comments using a bot account. When someone writes a comment like this:
```
@ci-bot mage:fmt
```
The bot would perform the action. That could also be useful. My only fear about automating this is falling in a loop where the automation makes a mistake somehow, we try to revert the automated commit and the automation picks up again and tries to fix the fix. Having a label guard for this would help.

.circleci/config.yml
3nvi: sure thing

3nvi:
> The bot would perform the action. That could also be useful. My only fear about automating this is falling in a loop where the automation makes a mistake somehow, we try to revert the automated commit and the automation picks up again and tries to fix the fix. Having a label guard for this would help.

I can't see a way of this happening. The automation will only commit changes produced by `mage fmt doc`. If  there are no changes, no commit will be present and/or pushed.

The only way this could happen is if there is a bug in `mage fmt`, in which case, addressing it in a separate PR would automatically fix that. The automation (unlike the sync one), won't commit  conflicts. 

All in all, you are not wrong and what you say is something that can give us safety, but  I've spend way too many of my weekends on things like that, so I don't want to spend more time on it 😆 . If you are willing to have a go at it, I'd be really interested to help with any sort of guidance you need

PR #1374, opened 2020-08-17
Migrate styling from pounce to panther

3nvi:
## Background

Due to the fact that we are going to have multiple "tab" styles present in Panther, I decided to move the default styling for tabbers from pounce to panther.

This means that all the styling logic got removed from pounce and needs to be added to panther through this PR. Visually and UX-wise everything is the same

## Screenshot

UI is the same as before

Screen Shot 2020-08-17 at 12 49 26 PM

## Changes

- Add `BorderedTab` and `BorderedTabDivider` component

## Testing

- `mage test:ci`


3nvi:
> Looks really sleek, i am approving but I have a minor question/suggestion.
> 
> There a margin between tabs that's clickable, should we move this margin outside of the button element?
> 
> Screenshot 2020-08-18 at 2 07 58 PM

Absolutely right! Missed that, thanks for noticing. It's a bit hard to achieve this  (since the outer button can't have styling), but I'll  try and work around it to fix it.

Issue #1388, opened 2020-08-18
gluesync CLI tool does not use `-start` flag


alxarch:
### Describe the bug

The CLI tool `gluesync` has a `-start` flag to set the starting date to sync partitions from.
The tool has a bug that ignores this flag if it is set at the CLI and can only use a start time for the partition sync if it is running interactively and it is left empty by asking the start date with a prompt. Since we will need to use `gluesync` for deployment of 1.7 this bug is critical for the release. If a fix is not merged until the release we should inform those handling the deployments that `gluesync` *must* be run interatively and *without* a `-start` flag to function properly.

### Steps to reproduce

1. mage build:tools
2. `out/bin/../gluesync -start 2020-04-02`
3. The flag is ignored and partitions resort to table creation time for the time range to sync

### Expected behavior

`-start` flag should be setting the start date o sync from

rleighton:
Fixed

PR #1389, opened 2020-08-18
Fix start date flag for gluesync

rleighton:
## Background

Processing the `-start` flag was missing (only interactive worked).

Closes: https://github.com/panther-labs/panther/issues/1388

## Changes

- parse flag
- check for -start in the future
- output value of -start in logging

## Testing

- ran cli


alxarch (approved):
LGTM
We should add a `-end` flag as well at some point and set this to `today` by default. This would allow to handle any use case (ie `eventTime` set in the future resulting in partitions being left out of sync). This is outside the scope of #1338 though.

PR #1394, opened 2020-08-20
fix: visual issues with destinations

3nvi:
## Background

While testing I spotted that:

- Webhook URLs are not returned from the API so they shouldn't be displayed
- The severities don't have the necessary visual margin

## Changes

- Add line breaks where needed
- Fix margin issues
- Remove webhook URL from destination card

## Testing

- Manually


alexmylonas (approved):
LGTM!

PR #1406, opened 2020-08-21
fix: Navigation should highlight alerts page

### LINKED ISSUE #1405: Navigation now shown as active in alerts page
### Description

Currently, when visiting the alerts page, the navigation sidebar is not shown as active. This shouldn't be the case

### Acceptance Criteria

- Make sure that the "alerts" navigation entry is highlighted when viewing the alerts page

3nvi:
## Background

The fact that we add  query parameters to the default alerts page URL, causes navigation page matching to fail. Thus, when visiting the alerts page, the navigation isn't highlighted at all.

This PR fixes that by comparing URL pathnames, stripping hashes and URL params.

Fixes #1405 

## Changes

- Compare pathnames for URL matching

## Testing

- Added test cases for URI matching


alexmylonas (approved):
LGTM! Sorry for the delay!

web/src/components/Navigation/NavLink/NavLink.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    expect(getByText('alerts').closest('a')).toHaveAttribute('aria-current', 'page');
+  });
+
+  it('ignores unrelated URIs', () => {
+    const { getByText } = render(<NavLink to="/something/" label="alerts" icon="list" />, {
+      initialRoute: '/something-else/',
+    });
+
+    expect(getByText('alerts').closest('a')).not.toHaveAttribute('aria-current', 'page');
```
vorillaz, line 53: That selector is weird and it might introduce side effects, how about explicitly querying the DOM using the aria attribute?

vorillaz (approved):

web/src/components/Navigation/NavLink/NavLink.test.tsx
3nvi: addressed!

PR #1430, opened 2020-08-25
Multiple scanners in panther tags and use separate event_time tag

alxarch:
## Background

While writting the docs for new parsers two needs became apparent:
  1. Support for multiple indicator scanners for struct fields (ie `panther:"foo,bar")
      This will be useful taking into consideration all the open issues for new indicator fields
  2. The tags for indicator fields and event time should be distinct to be able to easily extend each functionality.

These changes are best be made sooner rather than later before more parsers are ported to the tag-based method.

## Changes

- Adds support for multiple scanners in `panther` tags
- Separates the tag name for indicators and event time ( `panther` / `event_time` ) 

## Testing

- `mage test:ci`


alxarch:
> Makes sense to have this separation. Why is there only 1 parser updated?

Because on OSS we have only written the `Gravitaional` parser using the struct tag approach

rleighton:
> > Makes sense to have this separation. Why is there only 1 parser updated?
> 
> Because on OSS we have only written the `Gravitaional` parser using the struct tag approach

Ok, so this code has no impact on the other parsers until they get ported.

alxarch:
There's no restriction on number of tags on purpose. There's the common case of json that has either one field or another. If each had a timestamp inside, you wouldn't be able to tag it appropriately. The override tag option can prioritize one value. That could have some restriction but the enum case would be problematic again.

rleighton:
> There's no restriction on number of tags on purpose. There's the common case of json that has either one field or another. If each had a timestamp inside, you wouldn't be able to tag it appropriately. The override tag option can prioritize one value. That could have some restriction but the enum case would be problematic again.

Ok, I guess I was not clear. There can be only one `p_event_time,  how is this enforced?

alxarch:
When no override option is set, _the order of the struct definition_ is used. If override is set on a timestamp it sets the timestamp regardless of order. The only check that would be meaningful would be to verify that only one override option is set for all descendants. That would be a bit complicated but we could add it to the validation happening when building a log type entry.

rleighton:
> When no override option is set, _the order of the struct definition_ is used. If override is set on a timestamp it sets the timestamp regardless of order. The only check that would be meaningful would be to verify that only one override option is set for all descendants. That would be a bit complicated but we could add it to the validation happening when building a log type event schema.

I'm very concerned about cut-n-paste errors, I bet I could easily make one where I copy a line, then edit and forget to take off the tag ... 

I think it is very important that make hard to get wrong, where:

- A parser has 1 and only 1 p_event_time
- All required fields are validated as added. I think this is not done by the parser writer (except for event time), so maybe we do not need a validation.

alxarch:
The resulting JSON always has 1 `p_event_time`. The first field in the order they are defined in the go struct
gets to set the time. Any other fields after that will see a non-zero time and will not set anything. I will remove the `override` option altogether and only rely on order of fields to have a clear rule.

rleighton:
> The resulting JSON always has 1 `p_event_time`. The first field in the order they are defined in the go struct
> gets to set the time. Any other fields after that will see a non-zero time and will not set anything. I will remove the `override` option altogether and only rely on order of fields to have a clear rule.

Can we make it fail registration if there are 2 defined?

kostaspap (changes requested):
This PR will need to be updated:
1. Update docs
2. Changes in the Slack parsers that use these tags

alxarch:
Much of the conversation about this PR revolves around the _behavior_ of the `event_time` tag. This falls outside the scope of this PR which only _separates_ `panther` and `event_time` tags. To accommodate feedback, this PR also removed the `override` option for `event_time` and bases the priority of `event_time` tags on _order of fields_. If we believe the 'priority by order of fields' behavior is not good enough or should be a separate issue to decide on, I can revert the `override` change and only limit the PR to it's _original_ scope.

alxarch:
> 2. Changes in the Slack parsers that use these tags
They are on enterprise branch, so any needed changes on the struct tags will happen on sync.

kostaspap (approved):

Issue #1179, opened 2020-07-16
Disabled resources do not correctly sort by status


austinbyers:
### Describe the bug

Sorting resources by status does not work as expected. "Disabled" is mixed in with "Pass":

![Screen Shot 2020-07-15 at 4 59 03 PM](https://user-images.githubusercontent.com/3608925/87611428-e963bb00-c6bc-11ea-8125-8e7cff12f1bc.png)

This seems more likely a backend issue, so I've assigned it to myself for now

### Steps to reproduce

1. Deploy `master` branch
2. Go to Cloud Security > Policies
3. Sort by Status ascending
4. Note the "disabled" status mixed in with the passing status

### Expected behavior

"disabled" should be a separate status group in the sort

3nvi:
Guessing that some additional param is missing on the payload, since that should be handled by the server.

We should also make sure that this is fixed in **all** rule & policy tables

s0l0ist:
It looks like we have two statuses: `PASS`, `FAIL`, or `ERROR`. Sorting is actually working in the backend and frontend; however, if the policy has been **disabled**, the frontend overwrites the status value with `DISABLED` instead. 

In its current state, it is possible to see `DISABLED` on passing, failed, or errored policies. In the screenshot posted, those two disabled policies have an underlying "passed" value that is being masked.

I think the problem is how we're presenting the data to the user while trying not to add clutter to the table. It also looks like we don't have the ability to sort or filter by enabled/disabled statuses either.

Just throwing around some solutions:
1.  Instead of a grey disabled badge we could have a binary-disabled badge that indicates the underlying pass/fail status (gray-green vs gray-red)
2. Create another column in the table with filtering/sorting capabilities for enabled/disabled

austinbyers:
> In its current state, it is possible to see DISABLED on passing, failed, or errored policies. In the screenshot posted, those two disabled policies have an underlying "passed" value that is being masked.

Ah, then would it be sufficient for the backend to sort by "enabled" status after sorting by pass/fail? Then at least they would be grouped together (pass: enabled, pass: disabled, fail: enabled, fail: disabled, error: enabled, error: disabled)

> Instead of a grey disabled badge we could have a binary-disabled badge that indicates the underlying pass/fail status (gray-green vs gray-red)

[...]

PR #1434, opened 2020-08-26
Group disabled policies together when sorting by status

austinbyers:
## Background

When sorting compliance policies by status, they do not appear to be grouped correctly:

Screen Shot 2020-08-26 at 1 04 21 PM

The "Disabled" status shows up among the "Pass" status - this is because even disabled policies are still evaluated, we just don't yet show its underlying status to the user (we should eventually).

Until then, the backend will first sort by enabled/disabled status so that the column sorts as you would expect

Closes: #1179 

## Changes

- When sorting policies by status, sort first by enabled/disabled (to group all disabled policies together at the end)
- Add some clarifying comments

## Testing

- After applying the change and sorting on the column:

Screen Shot 2020-08-26 at 1 10 57 PM

- `PKG=./internal/core/analysis_api/main dev -- mage test:integration`

### DIFF
--- a/internal/core/analysis_api/handlers/list.go
+++ b/internal/core/analysis_api/handlers/list.go
@@ -400,7 +400,9 @@ func sortByStatus(policies []*models.PolicySummary, ascending bool, policyStatus
 
 		// Group all disabled policies together at the end.
 		// Technically, disabled policies still have a pass/fail status,
-		// we just don't show it yet in the web app. They are all shown with status "disabled"
+		// we just don't show it yet in the web app.
+		// The "disabled" status overrides the "pass/fail" status
+		// TODO - remove this block once enabled/disabled status is shown separately (ETA v1.9)
 		//
 		// So the sort order is essentially:
 		//     PASS < FAIL < ERROR < PASS/DISABLED < FAIL/DISABLED < ERROR/DISABLED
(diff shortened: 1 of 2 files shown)


austinbyers:
If this page were redesigned (e.g. enabled/disabled made into a separate column), then we can re-tweak the sorting as needed. This was just the fastest simplest solution which makes it work much better for the user right now

internal/core/analysis_api/handlers/list.go
```diff
@@ -426,13 +443,13 @@ func sortByEnabled(policies []*models.PolicySummary, ascending bool) {
 	sort.Slice(policies, func(i, j int) bool {
 		left, right := policies[i], policies[j]
 		if left.Enabled && !right.Enabled {
-			// ascending: true > false
-			return !ascending
+			// when ascending (default): enabled < disabled
+			return ascending
```
austinbyers, line 447: Flipped so that enabled comes before disabled by default

3nvi:
Thanks for that @austinbyers . I  didn't communicate the  fact that I was already in discussion with  our designer to make sure that we don't group those 2 statuses in 1 UI component.

The result can be seen here https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/master/commits/d0570e02d9fb58bb4e79266e5e7ff6e3978973a8/files/323d0ca1-81a8-40e7-83ee-0a05b870bb4a/layers/4B95AE99-746B-43F4-8FC0-CF251074C34F

With the new design this change won't be needed, although this design won't land until 1.9 Until then, we can definitely keep this work, but,  if the operations performed in this PR are costly, we'll have to remember to revert it

austinbyers:
> Until then, we can definitely keep this work, but, if the operations performed in this PR are costly, we'll have to remember to revert it

Sounds good, I'll merge for now. It's not costly to do the extra sort, but it won't make sense to keep once it's a separate column, so I'll add a TODO so we remember to remove it

PR #1435, opened 2020-08-26
Add custom app client token expiration

austinbyers:
## Background

Cognito now supports configurable expiration for the 3 kinds of tokens (access, ID, and refresh), but only in the SDK (not yet in CloudFormation). We want to lower the defaults, so I've done so via a custom CloudFormation resource

Closes: https://github.com/panther-labs/panther-enterprise/issues/546

## Changes

- Add new `Custom::CognitoClientTokenExpiration` resource
- Lower access token validity to 15 minutes and refresh/ID token validity to 12 hours
- Update Go module versions (I needed the latest AWS SDK for the new Cognito settings)

We can delete the custom resource if and when CloudFormation adds support for the token validity settings.

## Testing

- Deployed to dev account and verified change took effect in the Cognito web console


internal/log_analysis/awsglue/table_metadata_test.go
```diff
@@ -144,7 +144,7 @@ func TestGlueTableMetadataSignature(t *testing.T) {
 	gm := NewGlueTableMetadata(models.LogData, "My.Logs.Type", "description", GlueTableHourly, partitionTestEvent{})
 	sig, err := gm.Signature()
 	require.NoError(t, err)
-	assert.Equal(t, "5a3ca736985afab5ba83361dcb17ecb4fd1ea5632b11674137e6887148556e67", sig)
+	assert.Equal(t, "53372e1ee5b73d1e73594335e6df94489d0a759106fa2119fca66844f7ee5618", sig)
```
austinbyers, line 147: Not sure exactly why this signature changed (updated libraries), but it's value doesn't really matter - it's just used to trigger changes to the appropriate custom resource.

3nvi (approved):
LGTM

We'll have to try it out and see if the delay of refetching the access token 4 times more than normal affects UX or not 😃

Issue #1421, opened 2020-08-24
mage should check that all buckets disable http access


rleighton:
### Describe the ideal solution

mage would fail the CF check if any bucket allows http access

austinbyers:
Can you add more context? Is this high priority? Did we miss this on a new bucket? Is the check you're referring to this one:

```yml
  AnalysisVersionsBucketPolicy:
    Type: AWS::S3::BucketPolicy
    Properties:
      Bucket: !Ref AnalysisVersions
      PolicyDocument:
        Statement:
          - Sid: ForceSSL
            Effect: Deny
            Principal: '*'
            Action: s3:GetObject
            Resource: !Sub arn:${AWS::Partition}:s3:::${AnalysisVersions}/*
            Condition:
              Bool:
                aws:SecureTransport: false
```

It would be pretty easy to check for a "ForceSSL" statement ID with `mage` (or `cfn-lint`), but actually evaluating the policy and proving there is no http access would be harder. I'm not sure if there is a Golang equivalent of `policyuniverse`, for example

rleighton:
I just meant that simple check as a reminder to devs when creating new buckets. Panther detects this itself, that's how Nick found some we forgot.

austinbyers:
We could also check for encryption, blocked public access, logging, versioning, etc. I was curious why this specific setting came up, that's all.

It's mostly a question of how deep to go with this. A few quick checks of the parsed raw CFN template with `mage`? Or relying on other advanced CFN parsing tools? `cfn-lint`, for example, supports some level of custom specifications, but I don't know if it would be sufficient for this use case.

PR #1437, opened 2020-08-27
Update mage test:cfn with check for S3 bucket ForceSSL statement IDs

austinbyers:
## Background

Some of Panther's buckets were missing the `ForceSSL` bucket policies until the last release, so I've added a check to `mage test:cfn` to flag buckets without the appropriate statement.

It's very simple, just checking for "Sid: ForceSSL" attached to every bucket. It's not actually evaluating the policy, which can get quite complex. But this will catch the most common problem - devs forget to add it to new buckets

Closes: #1421 

## Changes

- `mage test:cfn` fails if an S3 bucket does not have a `ForceSSL` statement attached in the same template
- Collapse the ProcessedData bucket policy into a single resource with conditional statements (rather than two completely separate policy resources). Having two separate resources is harder to check with automation and also more prone to errors when CFN switches between the two - it will create the new bucket policy and then delete the old one. But buckets can only have one policy, so I'm not sure if that would work as we expect. It's simpler and safer to have a single bucket policy at all times - the policy statements can update conditionally, not the resource itself

## Testing

- Deployed - the default processed data bucket still has the correct "ForceSSL" statement only
- Removed ForceSSL statement from a bucket and ran `mage test:cfn`:

```
12:51:22	ERROR	    X cfn-lint failed (1/2): deployments/bootstrap.yml: S3 bucket InputDataBucket needs an associated ForceSSL policy statement ID
```

### DIFF
--- a/deployments/bootstrap.yml
+++ b/deployments/bootstrap.yml
@@ -139,7 +139,6 @@ Mappings:
 
 Conditions:
   ConfigureLogSubscriptions: !Not [!Equals [!Select [0, !Ref LogSubscriptionPrincipals], '']]
-  NotConfigureLogSubscriptions: !Not [Condition: ConfigureLogSubscriptions]
   EnableAccessLogs: !Equals [!Ref EnableS3AccessLogs, true]
   ExternalAccessLogs: !Not [!Equals [!Ref AccessLogsBucket, '']]
   ReplicateData: !Not [!Equals [!Ref DataReplicationBucket, '']]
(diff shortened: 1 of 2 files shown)


deployments/bootstrap.yml
```diff
@@ -520,6 +479,28 @@ Resources:
             Condition:
               Bool:
                 aws:SecureTransport: false
+          - !If
```
rleighton, line 482: To truly test this we will need to do a snowflake deployment (which uses this config). I'll to that next week.

austinbyers: It should result in the exact same logical policy.

I tested setting the LogSubscriptionPrincipals parameter to trigger the condition, and the policy updated as expected. Then I reverted, and the extra statements were removed

PR #1438, opened 2020-08-27
Fix deployment role permissions

nhakmiller:
## Background

Somehow we got a bad permission merged into the deployment role that blocks snowflake integrated deployments from upgrading to v1.7.0+. We had tested the deployment role already, but it looks like somehow a typo slipped in.

We were granting `elasticfilesystem:CreateMountPoint`, which is a permission that does not actually exist. We need to be granting `elasticfilesystem:CreateMountTarget`.

I think we can hopefully just publish this one role, no need to bump everything. Not sure if we should publish it on v1.7.1 (to overwrite the bad existing one) or on v1.7.2.

## Changes

- Changed an incorrect permission

## Testing

- None yet, all our deployments are now using the published templates and it's a bit of a pain to switch back, if this doesn't work we can try a few iterations but it should be close since a similar policy already worked once internally

Screen Shot 2020-08-27 at 12 52 46 PM
Screenshot  showing that the existing `CreateMountPoint` permission is not recognized.


deployments/auxiliary/cloudformation/panther-deployment-role.yml
```diff
@@ -18,7 +18,7 @@ AWSTemplateFormatVersion: 2010-09-09
 Description: IAM role for deploying Panther
 
 Metadata:
-  Version: v1.7.1
+  Version: v1.7.2
```
austinbyers, line 21: I would say it's fine to keep the version the same, as long as CloudFormation will still pick up changes in nested stacks even if the template URL does not change (which I'm 90% sure is the case)

austinbyers: Either way

nhakmiller: I'll leave it on 1.7.1 then, no reason to have a broken template out there.

PR #1439, opened 2020-08-28
Enhance error mocking capabilities of mocked graphql operations

3nvi:
## Background

This PR re-configures the way the interface of the automatic "mock" functions that are created through `graphql-codegen`. It now allows GraphQL errors to be passed as arguments in order to mock an errored response on the server.

Up until now, only a network error could be easily mock, with server error mocking being extremely hard.

## Changes

- Change the interface of the autogenerated mocked function for each graphQL operation
- rename npm script

## Testing

- `npm run test:ci`


web/codegen/mockGraphqlOperationsPlugin.js
```diff
@@ -1,5 +1,5 @@
-'use strict';
```
3nvi, line 1: This file is a transpilation artifact. Please ignore it from your review.

web/src/components/dropdowns/UpdateAlertDropdown/graphql/updateAlertStatus.generated.ts
```diff
@@ -19,6 +19,7 @@
 import * as Types from '../../../../../__generated__/schema';
```
3nvi, line 19: # ATTENTION

All the files from here and down below are the autogenerated by running `npm  run graphql-codegen`. You can check out the changes, but there is no human interference in their creation

web/codegen/mockGraphqlOperationsPlugin.js
```diff
@@ -33,15 +33,14 @@ class MockGraphqlOperationsVisitor extends visitor_plugin_common_1.ClientSideBas
     operationResultType,
```
3nvi, line 33: This is a transpilation artifact. Please exclude it from you review.

Issue #1392, opened 2020-08-20
Design: Destination delivery statuses


s0l0ist:
### Description

This **design** issue is for how the user should interact with a re-send button and how they should visualize this information in the panther UI.

When alerts are sent to a destination, we should provide a way for our uses to understand if the alert was delivered successfully. In the event of a failure, there needs to be a way to re-send the alert to the destination that failed.

### Acceptance Criteria

Users should be able to click on an alert, go to the alert details, and quickly check if the alert was delivered successfully or not. Ideally, we should only have a visual indication if there was a problem.

The failure message should contain some information about the failure if possible containing a history of attempts to send (in the event the alert was re-tried automatically or manually).

The shape of the data will be a list of objects containing data from _each_ attempt to send the alert. It is acceptable to have more than one entry for every destination: 
```javascript
[
  {
    outputId: "aaa" // the outputId of the destination (slack)
    success: true, // if the alert was sent successfully or not
    statusCode: 201, // http status code
    message: "", // optional success or failure message of the response
    deliveryAttemptedAt: 2020-01-01T00:00.000Z // timestamp when the alert was attempted to be delivered
  },
  {
    outputId: "bbb" // the outputId of the destination (asana)
    success: false, // if the alert was sent successfully or not
    statusCode: 500, // http status code
    message: "client timeout", // optional success or failure message of the response
    deliveryAttemptedAt: 2020-01-01T00:00.000Z // timestamp when the alert was attempted to be delivered
  },
  {
    outputId: "bbb" // the outputId of the destination (asana)
    success: true, // if the alert was sent successfully or not
    statusCode: 200, // http status code
    message: "success", // optional success or failure message of the response
    deliveryAttemptedAt: 2020-01-01T00:00.000Z // timestamp when the alert was attempted to be delivered
  },
  ...
]
```

- Alert tables:
  - A visual indicator in both tables (in the list alerts & rules pages) containing alerts that informs the user there is a problem and they should click to view the alert details.

- Alert details:
  - Should not show extra information if there were no issues delivering the alerts
  - Should have an optional toggle somewhere that the user can click to reveal the information about the delivery status (should include all successful and/or failed attempts and the data associated)
  - Should have a visual indicator (kind of a highlight) and automatically expand the area (the user needed to click to expand before) so the user can immediately see where the problem is occurring and why.
[...]

Issue #1253, opened 2020-07-28
Destination Status and Tests


s0l0ist:
### Description

Show links to the alerts in the external system (e.g. link to PagerDuty, Slack) and have the ability to send a sample alert to a destination.

### Acceptance Criteria

- Alert details should show a clear and concise message about the failure (or success) of the status of an alert delivery. The information should be presented in a intrusive way (visual indication) if there was an error, otherwise it should be passive.
- User should be able to easily re-send an alert to the destinations configured in panther. This also acts as a 'test' for destination integrations.
- Visual indication if an error occurred. Should also be present in the table-view. This is so the user can spot issues with delivery based on date-range queries. Viewing the error can be done in the details page.

## Caveats

- Not all integrations may support this type of deep link and therefore we may only be able to log metadata

PR #1440, opened 2020-08-31
Delivery statuses

### LINKED ISSUE #1340: [Backend] Destination Status and Tests (refactor)
### Description

The current delivery code is not easily extensible; therefore, it will need to be refactored to take into consideration a stateless approach. Meaning, instead of sending an alert to _its_ destinations, we need to create functions which take in a map of alerts to a list of outputIds that are not bound to each other.

This will allow for an HTTP API to create to send either an alertID or test payload to any (valid) outputID all while using the same functions the normal SQS events would trigger. This also has the benefit of testing the actual delivery code that will be used internally with the new external API.

### Acceptance Criteria

- Modify the lambda to accept SQS events and an HTTP API
- Ensure all delivery related functions are shareable between HTTP API (and test payloads)
- Retrying alerts needs to be dynamically switched in the case of using the HTTP API (we don't want to retry there)
- Create DDB handlers to `ListAppend` a map of delivery statuses. This means we will have at most O(n) status entries per alert where n is the number of outputIds configured in Panther which is an acceptable cap that doesn't involve DB locks.

### LINKED ISSUE #1341: [Test] Destination Status and Tests
### Description

During the new onboarding flow, the user needs to be able to trigger a test payload. This PR includes a hard-coded test payload in the backend that can be triggered via the frontend.

### Acceptance Criteria

- User starts the flow to onboard a new output destination
- User enters in details for the destination and clicks `Next`
- The user is prompted to (optionally) send a test alert and clicks the button
- User sees the test payload delivered to the destination

s0l0ist:
## Background

closes #1341
closes #1340 

This PR holds the backend changes for epic #1253, but frontend work needs to be finalized for everything else **besides** sending a test alert during destination onboarding. This will likely be in a separate PR.

## Changes

The gist:

- Previously, the delivery lambda only had an SQS trigger to send alerts to destinations and did not record the response
- Now, the delivery lambda has 3 entry points:
  - SQS trigger
  - HTTP API to send a hard-coded test alert to a given destination
  - HTTP API to re-send an alert
- All three handlers share the same core delivery mechanism, but only the SQS trigger and the re-send API update the delivery statuses in DDB
- Changed the lambda name from `panther-alert-delivery` -> `panther-alert-delivery-api` for AppSync permissions to call these new APIs.

Status:

- [x] Test Alert - Adds a backend API to send a test alert to a destination
- [x] Test Alert - frontend changes
- [x] Refactor misc backend components such as the outputs cache, SQS delivery statuses to share the same logic as re-send API.
- [x] Added tests to support new changes
- [x] Re-send Alert - Adds a backend API to re-send a given alertId to the specified outputIds
- [ ] Re-send Alert - frontend changes #1392

## Testing

- mage test:ci


internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -0,0 +1,182 @@
...
+		zap.L().Error("Alert not found", zap.String("AlertID", input.AlertID))
+		return nil, &genericapi.DoesNotExistError{
+			Message: "Unable to find the specified alert!"}
+	}
+	return alertItem, nil
+}
+
+// populateAlertData - queries the rule or policy associated and merges in the details to the alert
+func populateAlertData(alertItem *alertTable.AlertItem) *deliveryModels.Alert {
+	// TODO: Fetch and merge the related fields from the Rule into the alert.
```
s0l0ist, line 90: This is the last remaining task for this API to be 100% completed.

api/graphql/schema.graphql
```diff
@@ -527,6 +528,10 @@ type ListResourcesResponse {
   resources: [ResourceSummary]
 }
 
+input SendTestAlertInput {
```
3nvi, line 531: optional: We are generally trying to use `input` when  the input is  a bit  complex. Now that you have just 1 key, you could just "inline" it to the query field definition

api/graphql/schema.graphql
```diff
@@ -527,6 +528,10 @@ type ListResourcesResponse {
...
+  outputIds: [ID]!
```
3nvi, line 532:
```suggestion
  outputIds: [ID!]!
```

web/src/components/wizards/common/DestinationTestPanel/DestinationTestPanel.tsx
```diff
@@ -18,27 +18,38 @@
...
 
 const DestinationTestPanel: React.FC = () => {
-  const [testStatus, sendTestStatus] = React.useState<TestStatus>(null);
+  const [testStatus, setTestStatus] = React.useState<TestStatus>(null);
   const {
     data: { destination },
     reset,
   } = useWizardContext<CreateWizardData & EditWizardData>();
 
+  const [useSendTest, { loading }] = useSendTestAlertLazyQuery({
```
3nvi, line 40:
```suggestion
  const [sendTestAlert, { loading }] = useSendTestAlertLazyQuery({
```

web/src/components/wizards/common/DestinationTestPanel/DestinationTestPanel.tsx
```diff
@@ -18,27 +18,38 @@
...
+    variables: {
+      input: {
+        outputIds: [destination.outputId],
+      },
+    },
+    onCompleted: () => setTestStatus('PASSED'),
+    onError: () => setTestStatus('FAILED'),
+  });
+
   const handleTestAlertClick = React.useCallback(() => {
```
3nvi, line 51: we can  remove  `handleTestAlertClick` and  just do `onClick={sendTestAlert}` directly on the button element

api/graphql/schema.graphql
```diff
@@ -44,6 +44,7 @@ type Mutation {
 type Query {
   alert(input: GetAlertInput!): AlertDetails
   alerts(input: ListAlertsInput): ListAlertsResponse
+  sendTestAlert(input: SendTestAlertInput!): Boolean
```
3nvi, line 47: optional: in the future we always want to send a payload back.  A `boolean` response doesn't say much. How would you convey a business  logic error?  Throwing an exception is not the right way; the response should contain a key with all the user-related errors (named `errors`) along with a `status` field denoting whether the alert was sent  or not.

Due to us wanting to move quickly, I understand if we postpone that, but all this is (and will be) part of an effort  to clean up the API schema and make it more solid

3nvi (approved):
Approving the web related part 🎉 
Some minor minor comments, but overall the approach is solid. Great work @s0l0ist !

api/graphql/schema.graphql
s0l0ist: > optional: in the future we always want to send a payload back.  A `boolean` response doesn't say much. How would you convey a business  logic error?  Throwing an exception is not the right way; the response should contain a key with all the user-related errors (named `errors`) along with a `status` field denoting whether the alert was sent  or not.

The current design didn't call for returning the error payload for sending a test alert. The error could also be anything including very large payloads due to a mistyped URL collecting an entire 404 page, network issues, or an internal exception. 

The other backend APIs will store this information, but for the test alert I didn't think we should change the design and display the exact error in this case. Maybe this is something we can redesign if you like next time.

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -0,0 +1,182 @@
...
+func getAlert(input *deliveryModels.DeliverAlertInput) (*alertTable.AlertItem, error) {
+	alertItem, err := getAlertsTableClient().GetAlert(&input.AlertID)
+	if err != nil {
+		zap.L().Error("Failed to fetch alert from ddb", zap.Error(err))
+		return nil, err
+	}
+
+	// If the alertId was not found, log and return
+	if alertItem == nil {
+		zap.L().Error("Alert not found", zap.String("AlertID", input.AlertID))
```
rleighton, line 81: Don't log, redundant

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil, &genericapi.DoesNotExistError{
+			Message: "Unable to find the specified alert!"}
```
rleighton, line 83: Why not add AlertID?

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -0,0 +1,182 @@
...
+		RetryCount: 0,
+	}
+}
+
+// getAlertOutputMapping - gets a map for a given alert to it's outputIds
+func getAlertOutputMapping(alert *deliveryModels.Alert, outputIds []string) (AlertOutputMap, error) {
+	// Initialize our Alert -> Output map
+	alertOutputMap := make(AlertOutputMap)
+
+	// Direct API hits should not use the cache. Only SQS events.
```
rleighton, line 119: Explain why this is so.

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -0,0 +1,182 @@
...
+	cache.setExpiry(time.Now().Add(time.Minute * time.Duration(-5)))
+
+	// Fetch outputIds from ddb
+	outputs, err := getOutputs()
+	if err != nil {
+		zap.L().Error("Failed to fetch outputIds", zap.Error(err))
```
rleighton, line 125: Do not log, use `errors.Wrap()`

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -0,0 +1,182 @@
...
+		return alertOutputMap, err
+	}
+
+	// Check the provided the input outputIds and generate a list of valid outputs
+	validOutputIds := intersection(outputIds, outputs)
+	if len(validOutputIds) == 0 {
+		zap.L().Error("Invalid outputIds specified", zap.Strings("OutputIds", outputIds))
```
rleighton, line 132: Do not log

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -0,0 +1,182 @@
...
+		return alertOutputMap, &genericapi.InvalidInputError{
+			Message: "Invalid destination(s) specified!"}
```
rleighton, line 134: Why not add the ids?

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	// Map the outputs
+	alertOutputMap[alert] = validOutputIds
+	return alertOutputMap, nil
+}
+
+// intersection - Finds the intersection O(M + N) of panther outputs and the provided input list of outputIds
+func intersection(inputs []string, outputs []*outputModels.AlertOutput) []*outputModels.AlertOutput {
+	m := make(map[string]bool)
```
rleighton, line 144: use `map[string]struct{}`

internal/core/alert_delivery/api/dispatch_alerts.go
```diff
@@ -0,0 +1,195 @@
...
+
+// Global variables
+var (
+	validate  = validator.New()
+	alertsAPI = os.Getenv("ALERTS_API")
+)
+
+// DispatchAlerts - Sends an alert to sends a specific alert to the specified destinations.
+func (API) DispatchAlerts(input []*deliveryModels.DispatchAlertsInput) (interface{}, error) {
+	zap.L().Info("Dispatching alerts", zap.Int("num_alerts", len(input)))
```
rleighton, line 39: Debug

internal/core/alert_delivery/api/dispatch_alerts.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// Extract alerts from the input payload
+	alerts := getAlerts(input)
+
+	zap.L().Info("Extracted from input", zap.Any("alerts", alerts))
```
rleighton, line 44: Debug

internal/core/alert_delivery/api/dispatch_alerts.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Send alerts to the specified destination(s) and obtain each response status
+	dispatchStatuses := sendAlerts(alertOutputMap)
+
+	// Record the delivery statuses to ddb. Ignore the returned output.
+	updateAlerts(dispatchStatuses)
+	zap.L().Info("Finished updating alert delivery statuses")
```
rleighton, line 57: Debug

internal/core/alert_delivery/api/dispatch_alerts.go
```diff
@@ -0,0 +1,195 @@
...
+
+	success, failed := filterDispatches(dispatchStatuses)
+	zap.L().Info("Deliveries that failed", zap.Int("num_failed", len(failed)))
```
rleighton, line 60: Debug

internal/core/alert_delivery/api/dispatch_alerts.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// getAlerts - extracts the alerts from an DispatchAlertsInput (SQSMessage)
+func getAlerts(input []*deliveryModels.DispatchAlertsInput) []*deliveryModels.Alert {
+	alerts := []*deliveryModels.Alert{}
+	for _, record := range input {
+		alert := &deliveryModels.Alert{}
+		if err := jsoniter.UnmarshalFromString(record.Body, alert); err != nil {
+			zap.L().Error("Failed to unmarshal item", zap.Error(err))
+			continue
```
rleighton, line 79: why not fail the whole api call?

internal/core/alert_delivery/api/dispatch_alerts.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// getAlertOutputMap - maps a list of alerts to their specified override outputs or defaults
+func getAlertOutputMap(alerts []*deliveryModels.Alert) (AlertOutputMap, error) {
+	// Create our Alert -> Output mappings
+	alertOutputMap := make(AlertOutputMap)
+	for _, alert := range alerts {
+		validOutputIds, err := getAlertOutputs(alert)
+		if err != nil {
+			zap.L().Error("Failed to fetch outputIds", zap.Error(err))
```
rleighton, line 97: Don't log, use `errors.Wrap()`

internal/core/alert_delivery/api/outputs.go
```diff
@@ -0,0 +1,92 @@
...
+	"go.uber.org/zap"
+
+	deliveryModels "github.com/panther-labs/panther/api/lambda/delivery/models"
+	outputModels "github.com/panther-labs/panther/api/lambda/outputs/models"
+	"github.com/panther-labs/panther/pkg/genericapi"
+)
+
+// Global variables
+var (
+	outputsAPI = os.Getenv("OUTPUTS_API")
```
rleighton, line 34: Can you centralize this in envconfig? Having lots of os.Getenv() make for brittle code.

internal/core/alert_delivery/api/outputs.go
```diff
@@ -0,0 +1,92 @@
...
+
+// getOutputs - Gets a list of outputs from panther (using a cache)
+func getOutputs() ([]*outputModels.AlertOutput, error) {
+	if cache.isExpired() {
+		outputs, err := fetchOutputs()
+		if err != nil {
+			return nil, err
+		}
+		cache.setOutputs(outputs)
+		cache.setExpiry(time.Now().UTC())
```
rleighton, line 78: Are you really setting expiry? Looks more like you are setting timestamp. Perhaps change the name of method.

internal/core/alert_delivery/api/send_test_alert.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"go.uber.org/zap"
+
+	deliveryModels "github.com/panther-labs/panther/api/lambda/delivery/models"
+)
+
+// SendTestAlert sends a dummy alert to the specified destinations.
+func (API) SendTestAlert(input *deliveryModels.SendTestAlertInput) (*deliveryModels.SendTestAlertOutput, error) {
+	// First, fetch the alert
+	zap.L().Info("Sending test alert")
```
rleighton, line 33: Debug

internal/core/alert_delivery/api/send_test_alert.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	// Send alerts to the specified destination(s) and obtain each response status
+	dispatchStatuses := sendAlerts(alertOutputMap)
+
+	// Log any failures and return
+	if err := returnIfFailed(dispatchStatuses); err != nil {
+		return nil, err
+	}
+	zap.L().Info("Test Succeeded")
```
rleighton, line 51: Debug

internal/core/alert_delivery/api/update_delivery.go
```diff
@@ -0,0 +1,84 @@
...
+			DispatchedAt: status.DispatchedAt,
+		}
+		alertMap[*status.Alert.AlertID] = append(alertMap[*status.Alert.AlertID], deliveryResponse)
+	}
+
+	// Init a channel
+	alertSummaryChannel := make(chan alertModels.AlertSummary)
+
+	// Make a lambda call for each alert in parallel. We dont make a single API call to reduce the failure impact.
+	zap.L().Info("Invoking UpdateAlertDelivery in parallel")
```
rleighton, line 54: Debug

internal/core/alert_delivery/api/update_delivery.go
```diff
@@ -0,0 +1,84 @@
...
+func updateAlert(alertID string, deliveryResponse []*alertModels.DeliveryResponse, alertSummaryChannel chan alertModels.AlertSummary) {
+	input := alertModels.LambdaInput{
+		UpdateAlertDelivery: &alertModels.UpdateAlertDeliveryInput{
+			AlertID:           alertID,
+			DeliveryResponses: deliveryResponse,
+		},
+	}
+	var response alertModels.UpdateAlertDeliveryOutput
+	if err := genericapi.Invoke(lambdaClient, alertsAPI, &input, &response); err != nil {
+		zap.L().Error("Invoking UpdateAlertDelivery failed", zap.Any("error", err))
```
rleighton, line 81: Is logging and not telling the caller the right thing to do? If so pls comment as to why

internal/core/alert_delivery/main/lambda.go
```diff
@@ -20,47 +20,39 @@ package main
...
+		return router.Handle(&apiRequest)
 	}
 
-	delivery.HandleAlerts(alerts)
-	return err
+	// If neither handler captured the request, return nothing
+	return nil, err
 }
 
 func main() {
```
rleighton, line 58: Shouldn't `api.Setup()` be called here 1 time unconditionally?

rleighton:
Can you try to reduce the number of globals used? Going forward we all agreed to minimize global usage in lambdas. Ideally you just a have a few in the `main` and the called code as none or maybe a small number to hold cache references.

api/graphql/schema.graphql
austinbyers: But we should still always be sending an *object* payload. Even if that object is just

```
{
    Result: Boolean
}
```

This way, you can extend the response in the future without breaking the type

api/lambda/delivery/models/api.go
```diff
@@ -28,6 +34,52 @@ const (
...
+// SendTestAlertInput sends a dummy alert to the specified destinations
+//
+// Example:
+// {
+//     "sendTestAlert": {
+//         "outputIds": ["198bdbc5-5d94-4d59-8c93-f2bab86359f5"]
+//     }
+// }
+type SendTestAlertInput struct {
+	OutputIds []string `json:"outputIds" validate:"required,gt=0,dive,uuid4"`
```
austinbyers, line 55: Nit: `required` here would imply `gt=0`, so you can drop the latter

internal/core/alert_delivery/api/api.go
```diff
@@ -0,0 +1,91 @@
...
+			AlertsTableName:                    env.AlertsTableName,
+			Client:                             dynamodb.New(awsSession),
+			RuleIDCreationTimeIndexName:        env.RuleIndexName,
+			TimePartitionCreationTimeIndexName: env.TimeIndexName,
+		}
+	}
+	return alertsTableClient
+}
+
+// getSQSClient - lazy load the SQS client
```
austinbyers, line 81: I did some testing awhile ago, the AWS client creation is on the order of *nanoseconds*. We don't need lazy-loading, just initialize all the clients in an `init()` or `setup()` function

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -0,0 +1,182 @@
...
+	outputModels "github.com/panther-labs/panther/api/lambda/outputs/models"
+	alertTable "github.com/panther-labs/panther/internal/log_analysis/alerts_api/table"
+	"github.com/panther-labs/panther/pkg/gatewayapi"
+	"github.com/panther-labs/panther/pkg/genericapi"
+)
+
+// DeliverAlert sends a specific alert to the specified destinations.
+func (API) DeliverAlert(input *deliveryModels.DeliverAlertInput) (*deliveryModels.DeliverAlertOutput, error) {
+	// First, fetch the alert
+	zap.L().Info("Fetching alert", zap.String("AlertID", input.AlertID))
```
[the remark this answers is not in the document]
austinbyers: For context, CloudWatch logs are becoming one of our highest costs - so by default we should only log very useful and actionable information

austinbyers:
This is great work, though! So now we're going to store the status of delivered alerts indefinitely? Makes sense, it'll be good to have a record of confirmed deliveries

api/lambda/delivery/models/api.go
s0l0ist: I thought I tested this and `required` tests that a slice is not `null` and the `gt=0` says the slice cannot be empty.

> This is great work, though! So now we're going to store the status of delivered alerts indefinitely? Makes sense, it'll be good to have a record of confirmed deliveries

Yes! This PR will also allow for users to re-send those alerts which have failed as well.

s0l0ist:
> This is great work, though! So now we're going to store the status of delivered alerts indefinitely? Makes sense, it'll be good to have a record of confirmed deliveries

Yes! This PR will also allow for users to re-send those alerts which have failed as well.

api/lambda/delivery/models/api.go
austinbyers: Ah, I didn't think of that, distinguishing `nil` from `[]`. Makes sense, thanks for checking

austinbyers: In that case, you probably don't need the `required` check (it would be implied by `gt=0`)

web/src/components/wizards/common/DestinationTestPanel/DestinationTestPanel.tsx
s0l0ist: I tried, but `onClick` complains the options of the function aren't of the right type. It has to be wrapped.

api/graphql/schema.graphql
s0l0ist: It could change here so I've left it.

internal/core/alert_delivery/main/lambda.go
s0l0ist: Yes, but I thought if the handler didn't exist, why call it?

internal/core/alert_delivery/api/api.go
```diff
@@ -0,0 +1,91 @@
...
+type envConfig struct {
+	AlertsTableName string `required:"true" split_words:"true"`
+	RuleIndexName   string `required:"true" split_words:"true"`
+	TimeIndexName   string `required:"true" split_words:"true"`
+}
+
+// Setup - parses the environment and builds the AWS and http clients.
+func Setup() {
+	envconfig.MustProcess("", &env)
+	maxRetryCount = getMaxRetryCount()
```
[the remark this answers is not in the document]
s0l0ist: > Can you try to reduce the number of globals used? Going forward we all agreed to minimize global usage in lambdas. Ideally you just a have a few in the `main` and the called code as none or maybe a small number to hold cache references.

Yep, I'm starting to notice the (bad) pattern I've put together. I'll clean this up.

s0l0ist:
> Can you try to reduce the number of globals used? Going forward we all agreed to minimize global usage in lambdas. Ideally you just a have a few in the `main` and the called code as none or maybe a small number to hold cache references.

Yep, I'm starting to notice the (bad) pattern I've put together. I'll clean this up.

internal/core/alert_delivery/api/cache.go
```diff
@@ -0,0 +1,94 @@
...
+type outputsCache struct {
+	// All cached outputs
+	Outputs   []*outputModels.AlertOutput
+	Timestamp time.Time
+}
+
+// Global variables
+var (
+	cache           *outputsCache
+	once            sync.Once
```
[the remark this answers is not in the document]
s0l0ist: Removed in favor of initializing inside the `Setup()` function.

api/lambda/delivery/models/api.go
```diff
@@ -28,6 +34,52 @@ const (
...
 )
 
+// LambdaInput is the invocation event expected by the Lambda function.
+//
+// Exactly one action must be specified, see comments below for examples.
+type LambdaInput struct {
+	DispatchAlerts []*DispatchAlertsInput `json:"Records"` // SQS Shape
+	DeliverAlert   *DeliverAlertInput     `json:"deliverAlert"`
+	SendTestAlert  *SendTestAlertInput    `json:"sendTestAlert"`
+}
```
austinbyers, line 44: This part took me awhile to understand, but it's actually pretty clever - I would not have thought to do it this way!

The format is the `genericapi` format, but where the first option happens to be the exact existing SQS message format. Then you can just route the request with `genericapi` instead of having to explicitly check the fields in the message.

I like it! Maybe just a more descriptive comment than "SQS Shape", something like:

```go
    // This matches the raw SQS message format (hence 'Records' being the name of the field),
    // but genericapi will route the request to the DispatchAlerts handler. This way all requests can be routed
    // by genericapi without having to inspect the message ahead of time
    DispatchAlerts []*DispatchAlertsInput `json:"Records"`

Comment (s0l0ist) [reply api/lambda/delivery/models/api.go]: I'll also add the main reason for doing so was to allow direct lambda invocations with an SQS payload instead of waiting for the SQS trigger to call the lamdda and log the reponse

File: api/lambda/delivery/models/api.go (hunk 27 above)
Comment (s0l0ist) [reply api/lambda/delivery/models/api.go:61]: Didn't think of that! Will update.

**Comment by s0l0ist** [comment]
> > Changed the lambda name from panther-alert-delivery -> panther-alert-delivery-api for AppSync permissions to call these new APIs.
> 
> Renaming a function can have unintended consequences - for example, if there are alerts in transit when an existing deployment is upgraded, will there be errors because the function no longer exists? Will those errors be retried? I believe the answer to both is yes
> 
> But it's certainly much easier to add permissions for a function instead of renaming a function to fit the existing permissions. In fact, we already have an "alerts-api", you could just add a "TestAlert" endpoint to that API and then invoke the alert-delivery in the backend rather than exposing two different alert api functions to the frontend. The small amount of extra latency shouldn't make a difference.
> 
> In summary:
> 
> 1. Renaming a function in a live deployment is risky - I would just avoid it altogether. We hardcode function names in lots of places - IAM permissions, integration tests, `mage`, etc - it'd be easy to miss something.
> 2. If you really want the rename, be sure you've tested the upgrade and searched the entire codebase for references to the old function.
> 3. Rather than exposing another lambda function to appsync, have you considered just adding a new endpoint to the existing `alerts-api`? That should simplify the permissions - the `alerts-api` can invoke `alert-delivery` directly (or even import the testing code and run it inline)

After discussing with @austinbyers internally, we will go ahead with the rename.

File: internal/core/alert_delivery/api/deliver_alert.go (hunk 1 above)
Comment (s0l0ist) [reply internal/core/alert_delivery/api/deliver_alert.go]: Completed

File: internal/core/alert_delivery/api/deliver_alert.go (hunk 13 above)
Comment (austinbyers) [reply internal/core/alert_delivery/api/deliver_alert.go:171]: Right, it's marginally more efficient but more importantly prevents bugs where a key exists in the map but with a `false` value. Using `map[string]struct{}` intentionally makes the value have no meaning

So then, for example, to add an entry: `m[item] = struct{}{}`

File: internal/core/alert_delivery/api/dispatch_alerts.go (hunk 18 above)
Comment (austinbyers) [reply internal/core/alert_delivery/api/dispatch_alerts.go:73]: Agreed - corrupt data is pretty serious, better to return server error than skip over it

File: internal/core/alert_delivery/api/cache.go [hunk 33]
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
```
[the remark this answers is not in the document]
s0l0ist: It's mostly the same except decorated with a setter/getter functions. 

Correct, refresh interval is now 30 seconds.

internal/core/alert_delivery/api/dispatch_alerts.go
s0l0ist: This piece is tricky. If we return an error the batches request goes back into the queue to be retired where it will keep failing repeatedly until put into the DLQ. One bad apple will affect the entire batch. 

By logging errors and discarding, we allow valid alerts that have been marshaled correctly to be delivered and progress through the pipeline normally.

austinbyers: Oh, this is unmarshaling the alert, not reading from a table. Yes, you're right - that seems reasonable (perhaps with a comment).

I feel like a lot of the code could be vastly simplified if we dropped the batching and let each invocation fetch only one alert at a time lol but that's perhaps a change for another day.

internal/core/alert_delivery/api/outputs.go
s0l0ist: Good catch. The `setExpiry` method is correct, I've updated the struct field name from `Timestamp` to `Expiry` to match.

rleighton:
87 files!

internal/core/alert_delivery/api/api_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"github.com/aws/aws-sdk-go/service/sqs/sqsiface"
+	"github.com/stretchr/testify/mock"
+
+	deliveryModels "github.com/panther-labs/panther/api/lambda/delivery/models"
+	outputModels "github.com/panther-labs/panther/api/lambda/outputs/models"
+	"github.com/panther-labs/panther/internal/core/alert_delivery/outputs"
+)
+
+type mockLambdaClient struct {
```
rleighton, line 38: I _think_ you can replace your mocks by using `pkg/testutils/awsmocks.go` . Use these if you can

internal/core/alert_delivery/api/api_test.go
```diff
@@ -0,0 +1,87 @@
...
+	args := m.Called(input)
+	return args.Get(0).(*dynamodb.GetItemOutput), args.Error(1)
+}
+
+type mockOutputsClient struct {
+	outputs.API
+	mock.Mock
+}
+
+func (m *mockOutputsClient) Slack(alert *deliveryModels.Alert, config *outputModels.SlackConfig) *outputs.AlertDeliveryResponse {
```
rleighton, line 73: Add to `pkg/testutils` ?

rleighton (approved):

internal/core/alert_delivery/api/api_test.go
s0l0ist: I'd say this one is too specific to the outputs. It will never be used outside of this scope.

internal/core/alert_delivery/api/api_test.go
s0l0ist: Yup, changed!

PR #1444, opened 2020-08-31
fix: create a destination without selecting a severity level

### LINKED ISSUE #1443: Destination without default severity

Now that we have destination overrides, we should be able to create destinations that don't have a default severity assigned so they can exist solely for the purpose of being a destination that receives overrides. Basically, if I want a destination that is only used as a target for destination overrides, then I must use destination overrides on every rule.

### Describe your use cases

- If I have a "test" destination that should only receive alerts from rules I'm testing, I don't want it to receive other alerts. Today, I have to assign this destination a default severity and it will receive all the alerts of that severity which I don't want.
- If I have multiple teams and I want certain alerts to go to certain teams and I create a separate destination for each team, those destinations will have a default severity.

Whenever alerts are sent, presumably daily.

Instead, I could set destination overrides for every single rule. This solution requires a lot of manual effort.

Alternatively, I could remove all `INFO` level rules/policies and then reserve the `INFO` level for my custom destinations. This solution feels hacky and removes my ability to use `INFO` alerts.

s0l0ist:
## Background

closes #1443

## Changes

- remove frontend form validation

## Testing

- mage test:ci
- create a destination, check CW logs to see an alert will not be delivered.
- set the destination override in the rule to point to the destination without default severity, see that alerts still come in


web/src/components/forms/BaseDestinationForm/BaseDestinationForm.tsx
```diff
@@ -65,11 +65,7 @@ interface BaseDestinationFormProps<AdditionalValues extends Partial<DestinationC
 // The validation checks that Formik will run
 export const defaultValidationSchema = Yup.object().shape({
   displayName: Yup.string().required(),
-  defaultForSeverity: Yup.object<{ [key in SeverityEnum]: boolean }>().test(
-    'atLeastOneSeverity',
-    'You need to select at least one severity type',
-    val => Object.values(val).some(checked => checked)
-  ),
+  defaultForSeverity: Yup.object<{ [key in SeverityEnum]: boolean }>().notRequired(),
```
3nvi, line 68: you don't need the `.notRequired()`. That's the default value 😄

3nvi (approved):
LGTM, but we can remove the redundant method chain 🎉

PR #1446, opened 2020-09-01
Add more source health messages

### LINKED ISSUE #1381: Add verbosity to source health data
Right now, we sometimes show a source as healthy although it’s failing.

We've had users mentioning that  they don’t understand what healthy means, since we don't give details as to what do we do in order to classify something as healthyy.

This task focuses on making  sure that the health data of  source  always returns a message for every health metric and not only when there is an error.

An example of this would be:

- `Checked KMS and we could successfully read it`
- `Couldn’t list your  KMS keys`

## Acceptance Criteria

- The `errorMessage` gets converted to  `message`
- Each health-item always  has a `{healthy,message}` combo  defined
- The `message` is **always** populated with something
- The content of the `message` is verbose and clear. It should **always** be a sentence that makes grammatical & syntactical  sense (i.e. instead of `key missing` it should be `Your KMS key couldn't be found`)

### LINKED ISSUE #1382: Showcase detailed source health data in FE
Right  now,  when a user hovers  over the health source icon of a  source,  a tooltip is displayed. So far, this tooltip either says "everything is ok on our end"  or displays an `errorMessage` from the BE.

This tooltip now needs  to have discrete sentences all the time,  showcasing the status  of all health-related metrics. For  example, if a source has 3 health metrics  and  1  of them is failing,  the tooltip should show 3 messages in  separate  sentences (with a  "green check" in  front of  the successful ones  and  a "red X" in  front of  the failing ones)

## Acceptance  Criteria

- The tooltip always shows 1 sentence for each health metric in separate rows
- The tooltip adds the correct icon in front of each sentence

### LINKED ISSUE #1383: Improve Sources Health details
### Description

This epic focuses on exposing the checks that we do in order to determine whether a source is healthy or unhealthy.

### Acceptance Criteria

- Users can always see a list of successful and/or failed checks, whenever they hover on the  "healthy" or  "unhealthy" badge of a source

nhakmiller:
## Background

This PR adds messages to all source health checks, not just failed ones. Additionally, it makes the messages more useful and human readable.

Closes #1381 
Closes #1382 
Closes #1383 

## Changes

**Backend**
- All health checks, passing and failing, now return a `message`
- Failing health checks also populate `rawErrorMessage`
- Health check messages are now more explicit about what succeeded or failed

**Frontend**
- Moved all badges to `Components/badges/`
- Created a single `SourceHealthBadge` which expects a list of health  checks and shows data for each one of them
- Replace all existing source health checks with the new component

##  Screenshots

Screen Shot 2020-09-02 at 7 44 55 PM

Screen Shot 2020-09-02 at 7 45 01 PM

## Testing

- Deployed to dev account:
- `npm run test:ci`


api/lambda/source/models/integration.go
```diff
@@ -91,8 +91,8 @@ type SourceIntegrationHealth struct {
 }
 
 type SourceIntegrationItemStatus struct {
-	Healthy      bool   `json:"healthy"`
-	ErrorMessage string `json:"errorMessage,omitempty"`
+	Healthy bool   `json:"healthy"`
+	Message string `json:"message"`
```
rleighton, line 95: Does changing this tag effect the UI?

3nvi: Yeah,  this issue is part of an epic that contains UI changes as well

internal/core/source_api/api/check_integration.go
```diff
@@ -101,28 +101,30 @@ func checkKey(roleCredentials *credentials.Credentials, key string) models.Sourc
...
 	}
 	kmsClient := kms.New(awsSession, &aws.Config{Credentials: roleCredentials})
 
 	info, err := kmsClient.DescribeKey(&kms.DescribeKeyInput{KeyId: &key})
 	if err != nil {
 		return models.SourceIntegrationItemStatus{
-			Healthy:      false,
-			ErrorMessage: err.Error(),
+			Healthy: false,
+			Message: "An error occurred while trying to describe the specified KMS key. Error: " + err.Error(),
```
3nvi, line 113: Ideally, I'd want this specific information in a separate key.

The  reasoning is that this is always the actual non-user-friendly error info which is fully omitted in the case of a healthy message. Thus I feel that it should be included in a separate key like `rawError`  or `error`, since we'd want all the content of the  `message` key to be presentable to the user.

What  do you think?

3nvi (approved):
I'm approving the business side of things (code will still have to be validated by another team member).

internal/core/source_api/api/check_integration.go
nhakmiller: So you're thinking there should always be a `Healthy` boolean, and a `Message` string, and only in the case of an error there should also be an `error` string? Seems reasonable from the backend perspective as long as you have some way to display each of these things on the frontend.

nhakmiller:
> Were any unit or integration test effected by this change?
> 
> I would suggest running `mage test:integration`

Integration tests have been disabled for the sources API for a while because they don't work. I can run them just to be sure though.

internal/core/source_api/api/check_integration.go
3nvi: Yeah, I just want to make sure the "message" doesn't get arbitrarily long with weird error messages. The `ErrorMessage` can be seen if the user chooses to see it

nhakmiller: Sounds good, just updated it to include this. Do you want me to merge to master (which will force the UI component to be done before 1.8.0 or things might break) or would you prefer I leave this branch here and the frontend team can work off of it directly? I could go either way.

3nvi: Latter. Leave it here and I'll add FE changes on the same branch. I don't expect them to be a loot so it's not going to hinder the reviewing in any way

vorillaz (approved):

PR #1450, opened 2020-09-01
Autogenerate deployment TF policy

austinbyers:
## Background

The IAM deployment role/policy we provide is very important - we and customers use it so Panther can be deployed in a least-privilege environment. We don't want nor need Panter to be deployed by an admin role.

We do a decent job of keeping the CloudFormation deployment role up to date because we test it in every release. However, we usually forget to update the Terraform copy. Now we have a customer actively using the TF role, so we really want to make sure it's up to date.

The solution, as always, is automation: 'mage fmt' will now copy the deployment policy from CloudFormation into Terraform so we don't have to do it manually anymore.

## Changes

- Deployment role name is now configurable (I believe @nhakmiller hinted at wanting this)
- Update Terraform deployment template with configurable role name and assume role principal, to have the same features as the CF
- 'mage fmt' will update the TF template with the same policy document present in the CloudFormation (which thanks to @3nvi means CI will automatically update TF whenever someone updates the CF)
- bugfix: Remove duplicate `firehose` permission

## Testing

- `mage clean setup test:ci` (which includes `tf validate` and `cfn-lint`)
- Creating deployment role with TF and CF (in progress)


deployments/auxiliary/cloudformation/panther-deployment-role.yml
```diff
@@ -21,19 +21,24 @@ Metadata:
   Version: v1.7.1
 
 Parameters:
+  DeploymentRoleName:
+    Type: String
+    Description: Name of the deployment IAM role. If an empty string, the name will be autogenerated
+    Default: PantherDeploymentRole
```
austinbyers, line 27: Default is the same value we had before for backwards compatibility

deployments/auxiliary/terraform/panther_deployment_role/main.tf
```diff
@@ -44,204 +42,211 @@ resource "aws_iam_role" "deployment" {
...
-        Resource : [
-          "arn:${var.aws_partition}:states:*:${var.aws_account_id}:activity:panther-*",
-          "arn:${var.aws_partition}:states:*:${var.aws_account_id}:execution:panther-*:*",
-          "arn:${var.aws_partition}:states:*:${var.aws_account_id}:stateMachine:panther-*",
-        ]
-      }
-    ]
-  })
+  # DO NOT EDIT - policy is automatically copied from CloudFormation by 'mage fmt'
+  policy = <<EOT
```
austinbyers, line 46: This is how Terraform multi-line strings work, "EOT" stands for "End Of Text"

So `mage` can just do a regex replace for everything between the two "EOT" markers

deployments/auxiliary/cloudformation/panther-deployment-role.yml
```diff
@@ -192,9 +198,6 @@ Resources:
               - !Sub arn:${AWS::Partition}:iam::${AWS::AccountId}:role/panther-*
               - !Sub arn:${AWS::Partition}:iam::${AWS::AccountId}:role/Panther*
               - !Sub arn:${AWS::Partition}:iam::${AWS::AccountId}:server-certificate/panther/*
-          - Effect: Allow
-            Action: firehose:*
-            Resource: !Sub arn:${AWS::Partition}:firehose:*:${AWS::AccountId}:deliverystream/panther-*
```
austinbyers, line 197: Duplicate of lines 184-186

austinbyers:
Tested role creation with TF, works as expected

rleighton:
Excellent!

Issue #1282, opened 2020-08-04
Frontend should retrieve the list of available log types by querying 


kostaspap:
### Description

The list of log types used by frontend is currently defined as a list of constants. We need to modify FE to retrieve the list of log types by querying the appropriate backend API. 

### Acceptance Criteria

- Frontend queries backend to retrieve the list of **deployed** log types during source onboarding/editing and rule creation/editing.

3nvi:
@kostaspap Is this something that we should have for 1.8? 

CC @rleighton

kostaspap:
> @kostaspap Is this something that we should have for 1.8?
> 
> CC @rleighton

@alxarch is owning the epic. @alxarch , I believe this should be in for 1.8?

rleighton:
It would be nice but this is not currently a big issue.

kostaspap:
> It would be nice but this is not currently a big issue.

It will be required for custom log types since the log types can be created/deleted during runtime.

3nvi:
Since we are going to go this way, I'll need to increase the scope of the related work to also include a breakdown of which log maps to which service.

Given the designs [here](https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/849670d3-62c0-462f-b399-bf6174c4352b/collections/eddb711d-3691-4e4f-ba4d-6eb04ff9a773), you can see that the onboarding has

**Step 1**
Select  a Service

**Step 2**
Select  a Transport method

**Step 3**
Show the service's form with the log types prefilled

Currently, the log types are static, so I know which log types map to which service. Once  they are dynamic I can't  do that, so I'll need an API to return to me a list of log types for a particular service that I query.

Is that ok?

3nvi:
Agreed with Kostas & Alex that for now, this can be implemented  statically in the Front-end, thus the API will just need to provide a list of strings

PR #1452, opened 2020-09-02
FE: Fetch Log Types from API

3nvi:
## Background

This ticket is a pre-cursor to custom log type support and relates to #1282 .

Essentially, we no longer maintain a fixed/hardcoded list of `LOG_TYPES`, but we fetch them via an API call. The reasoning is that we eventually want to allow user-defined log types to be  part of the response.

One problem that this created is that we - so far - matched log types to fixed colors from our palette. This will no longer be possible, so a "string -> color" hash function has been created  and was limited to the lightness/saturation style  that we currently use in Panther

## Changes

- Create GraphQL query that fetches log types
- Replace all `LOG_TYPES` usages with an API call
- Generate random (but same across refreshes) color for each log type, that matches the colors we'd use in Panther
- Replace the hardcoded log-type <-> color mapping with  the autogenerated one

## Testing

- Tested that log types maintain their colors across refreshes


alexmylonas (approved):
Code looks solid and PR seems to work as intended. 

I have a couple of suggestions for consideration: 

1. Keep existing colors and use `stringToPale` only if there is no corresponding color assigned, the generated colors feel a bit pale 😛 
Screenshot 2020-09-03 at 12 35 15 PM

2. Should we add the API call to a top level like in `PromptController` and basically cache the response? This API will be used extensively  and the response will change rarely (and only on some specific user action i think).

3nvi:
@alexmylonas 

> only if there is no corresponding color assigned,

There is never a pre-assigned color now. All of them are random, since any hardcoded mapping has been removed.

> Should we add the API call to a top level like in PromptController and basically cache the response? This API will be used extensively and the response will change rarely (and only on some specific user action i think).

I thought the same thing.  What I did is add it to the log analysis sidebar menu. Every time you "open the menu" to go to a "log analysis page" I prefetch the request. Do you think we need to do it **even more** aggressively than that?

alexmylonas:
> There is never a pre-assigned color now. All of them are random, since any hardcoded mapping has been removed.

Yeap, i was thinking of keeping the hardcoded mapping for the most common logTypes and generate the others with your function. Not a pressing issue we can re-visit this. 

> I thought the same thing. What I did is add it to the log analysis sidebar menu. Every time you "open the menu" to go to a "log analysis page" I prefetch the request. Do you think we need to do it even more aggressively than that?

I think this covers most (if all) of the cases, i think I would prefer it in more upper level for clarity but we can leave it as is.

Ship it 🚢

PR #1454, opened 2020-09-02
Backport changes to apigen and lambdamux

alxarch:
## Background

While implementing more endpoints for logtypes API in enterprise, some changes where made to `apigen` and `lambdamux` packages. These are backported to OSS so the branches are in-sync. 

## Changes

- Backport changes to `pkg/x/apigen` (adds models generation and generates client in the same package as api)
- Backport changes to `pkg/x/lambdamux` (removes batch and demux complexity)
- Update `logtypesapi` to use the new features

## Testing

- mage test:ci
- underway: deploy logtypesapi lambda and test manually through AWS console


api/lambda/logtypes/models_gen.go
```diff
@@ -1,4 +1,6 @@
-package internal
+// Code generated by apigen; DO NOT EDIT.
```
austinbyers, line 1: Curious why the permissions on this file changed? It should probably be 0600 or 0644 rather than 0755

internal/core/logtypesapi/dynamodb.go
```diff
@@ -80,6 +77,13 @@ func mustMarshalMap(val interface{}) map[string]*dynamodb.AttributeValue {
 }
 
 type recordKey struct {
-	RecordID   string
-	RecordKind string
+	RecordID   string `json:"RecordID" validate:"required"`
+	RecordKind string `json:"RecordKind" validate:"required,oneof=native custom"`
```
austinbyers, line 81: Nit: If the raw JSON payload has the same field names as the struct, you don't need the json tags at all

pkg/x/apigen/internal/models.go
```diff
@@ -163,3 +206,10 @@ func flattenStruct(fields []*types.Var, tags []string, s *types.Struct) ([]*type
 	}
 	return fields, tags
 }
+
+func isTime(typ types.Type) bool {
+	if named, ok := typ.(*types.Named); ok {
+		return named.Obj().Pkg().Path() == "time" && named.Obj().Name() == "Time"
+	}
+	return false
+}
```
austinbyers, line 215: How do imports work when models want to import from other packages entirely? E.g. some of our APIs import models from AWS or other services

austinbyers (approved):
More deletions than additions, I always like to see that :)

internal/core/logtypesapi/dynamodb.go
alxarch: This is so that the generated models reflect this, FE devs do not know the quirks of go code

pkg/x/apigen/internal/models.go
alxarch: The models generator reduces all types to basic types and inlines them in the file. The generated file is only there for documentation purposes.

api/lambda/logtypes/models_gen.go
alxarch: Will investigate

alxarch: Fixed

PR #1455, opened 2020-09-02
Adds YAML-based tests for log parsers

alxarch:
## Background

Log parsers should have an easy way to define multiple tests based on samples.
The current test method for logtypes includes a lot of boilerplate code in go test modules that could be avoided to focus on writting as many tests as possible for every log sample we find a test case for. 

## Changes

- Adds a `logtesting` module that focuses on testing log parsers.
- Adds `logtesting.RunTestsFromYAML` that reads a YAML file with test cases and executes tests for log parsers.
- Switches the gravitational logs to use this test engine and expands the tests to include more log samples
 
## Testing

- Verified that the tests case from the YAML file is executed.


internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
```diff
@@ -0,0 +1,254 @@
...
+	}.Froze()
+	api.RegisterExtension(omitempty.New("json"))
+	//api.RegisterExtension(&tcodec.Extension{})
+	return api
+}
+
+// Checks that `actual` is a parser result matching `expect`
+// If expect.RowID is empty it checks if actual has non-empty RowID
+// If expect.EventTime is zero it checks if actual.EventTime equals actual.ParseTime
+// If expect.ParseTime is zero it checks if actual.ParseTime is non-zero
```
kostaspap, line 143: `If expect.ParseTime is zero` - The parse time should never be zero. Why do we need this check?

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
```diff
@@ -0,0 +1,254 @@
...
+func TestResult(t *testing.T, expect string, actual *pantherlog.Result, indicators ...pantherlog.FieldID) {
+	t.Helper()
+	logType := jsoniter.Get([]byte(expect), pantherlog.FieldLogTypeJSON).ToString()
+	require.Equal(t, logType, actual.PantherLogType)
+	expectResult := pantherlog.Result{}
+	if indicators == nil {
+		indicators = pantherlog.FieldSetFromJSON([]byte(expect))
+	}
+	require.NoError(t, unmarshalResultJSON([]byte(expect), &expectResult, indicators))
+	//require.Equal(t, -1, bytes.IndexByte(actual.JSON, '\n'), "Result JSON contains newlines")
```
kostaspap, line 154: Remove unused

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
```diff
@@ -0,0 +1,254 @@
...
+	if expectResult.PantherEventTime.IsZero() {
+		EqualTimestamp(t, actual.PantherParseTime, actual.PantherEventTime, "event time not equal to parse time")
+	} else {
+		EqualTimestamp(t, expectResult.PantherEventTime, actual.PantherEventTime, "invalid event time")
+	}
+	if len(expectResult.PantherRowID) == 0 {
+		require.NotEmpty(t, actual.PantherRowID)
+	} else {
+		require.Equal(t, expectResult.PantherRowID, actual.PantherRowID)
+	}
```
kostaspap, line 175: The PantherRowID is a random string. 
This check can be simplified - just check that the actual.PantherRowID is non empty.

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+func JSON() jsoniter.API {
+	api := jsoniter.Config{
+		EscapeHTML:             true,
+		SortMapKeys:            true,
+		ValidateJsonRawMessage: true,
+	}.Froze()
+	api.RegisterExtension(omitempty.New("json"))
+	//api.RegisterExtension(&tcodec.Extension{})
```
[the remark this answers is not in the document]
alxarch: We switched to registering tcodec globally for ease of use. I will remove this.

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
alxarch: If we provide a Row ID in the `actual` it means we *want* to validate that they match. If we don't we only validate it's non-empty.

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
alxarch: These all follow the logic that if we have some value set in `expect` then we want to test for _equality_ otherwise we want to check for non-empty value. The reasoning being that most registered parsers will fill RowID and ParseEventTime with something that changes on each run. Thus we cannot set a reasonable `expect` value for these. But we _want_ to check that they are set to _something_. For `event_time` the parsers should be setting it to parse time if the event had _no event timestamp_. So if `expect.event_time` is non-zero, we test for `expect.event_time == actual.event_time` otherwise we check for `actual.parse_time == actual.event_time`.

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
```diff
@@ -0,0 +1,254 @@
...
+
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/logtypes"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog/omitempty"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog/tcodec"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/parsers"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/registry"
+)
+
+func RunTestsFromYAML(t *testing.T, resolve *logtypes.Registry, filename string) {
```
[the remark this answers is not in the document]
alxarch: no, we want an argument to be able to run tests in parallel.

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
```diff
@@ -0,0 +1,254 @@
...
+		t.Run(testCase.Name, testCase.Run)
+	}
+}
+
+func RunTests(t *testing.T, tests ...TestCase) {
+	t.Helper()
+	for _, tc := range tests {
+		t.Run(tc.Name, tc.Run)
+	}
+}
```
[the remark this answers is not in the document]
alxarch: We will be using it to replace the testutil equivalent on parsers that already use it, once merged.

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
```diff
@@ -0,0 +1,254 @@
...
+	Input   string             `json:"input" yaml:"input"`
+	Result  string             `json:"result" yaml:"result"`
+	Results []string           `json:"results" yaml:"results"`
+	LogType string             `json:"logType" yaml:"logType"`
+	Resolve *logtypes.Registry `json:"-" yaml:"-"`
+}
+
+func (c *TestCase) Run(t *testing.T) {
+	TestLogType(t, c.Resolve, c.LogType, c.Input, append([]string{c.Result}, c.Results...)...)
+}
```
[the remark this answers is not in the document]
alxarch: See below

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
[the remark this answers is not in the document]
kostaspap: Can you update one of those tests as an example? It will help understand how this is going to be used.

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
alxarch: removed

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
alxarch: removed the checks that are not directly related to RunTestsFromYAML

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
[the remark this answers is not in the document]
alxarch: They are on the enterprise branch currently

kostaspap (approved):

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
alxarch: removed

PR #1456, opened 2020-09-02
Refactor logtypes registration to separate builtin from runtime logtypes

alxarch:
## Background

UPDATE: This PR was stale for some time but recent problems with cyclic dependencies for Crowdstrike logs made it relevant again. Merging this will fix those cyclic dependency issues and provide a better experience for managing log types.

While writing the docs for `how to add a new log type` it became clear that the process of registering a new log type to the system was confusing. In addition, as we proceed to implement user-defined log types there is a need for _clear distinction_ at runtime between the 'native' log types defined in go code and the additional user-defined log types.

In the current registration process we require the developer to both call the `logtypes.Register` function *and* remember to import the appropriate module in the `registry` package. 

With the changes in this PR we introduce the `logtypes.Group` type which is an _immutable_ index of log types. Each module defining log types (ie `awslogs`, `gitlablogs`) 'exports' it's available log types through a `LogTypes() logtypes.Group` function. 

A simple code generator scans the `log_processor/parsers/...` modules to find exported log types and generates a file to auto-import all exported log types in the global `registry` module.
Adding/removing of non-native log types can use the `registry.Register` / `registry.Del` function but we _cannot_ modify the 'built-in' log types at runtime. This guarantees that all modules using the `registry` package cannot the affect system in unpredictable ways.

## Changes

- Adds `LogTypes() *logtypes.Group` function to all modules that export log types.
- Auto-registers all the log types exported by `parsers/...` modules during `registry` module's `init()`.
- Changes the `registry.Register` and `registry.Del` functions to only allow runtime modifications for log types that are not in the 'builtin' group.

## Testing

- mage test:ci
- deployed and added a source with some 'native' log type to see that log processing works as expected.


giorgosp:
Generally all good, although I have some clarifying questions:

- Why do we need to separate "builtin" and "runtime" parsers (log types)? Which characteristics differentiate them?
- >While writing the docs for how to add a new log type it became clear that the process of registering a new log type to the system was confusing

Which part is confusing?

kostaspap (commented):
Is this up for review or not? (it has been a draft since a while)

rleighton:
Is this OBE now?

kostaspap (commented):
Make sure to update the docs!

giorgosp (approved):

internal/log_analysis/log_processor/logtypes/entry.go
```diff
@@ -0,0 +1,279 @@
...
+			switch e := e.(type) {
+			case error:
+				err = e
+			case string:
+				err = errors.New(e)
+			default:
+				err = errors.Errorf(`panic %v`, e)
+			}
+		}
+	}()
```
kostaspap, line 273: Why do we need this handling here?

internal/log_analysis/log_processor/parsers/awslogs/awslogs.go
```diff
@@ -36,71 +36,74 @@ const (
...
-	)
+// LogTypes exports the available log type entries
+func LogTypes() logtypes.Group {
+	return logTypes
 }
+
+// nolint:lll
+var logTypes = logtypes.Must("AWS",
+	logtypes.Config{
+		Name:         TypeALB,
```
kostaspap, line 47: The "group" has the name "AWS". `TypeALB` has value `AWS.ALB`. Is it possible to have a group with name different than the prefix of the log type? Should we just use the name of the group as the prefix and drop the prefix from the log type name? So, change the `TypeALB` to have value `ALB`.

kostaspap (approved):

internal/log_analysis/log_processor/logtypes/entry.go
alxarch: Because glue infer code panics with different types of error instead of returning errors. I don't know if this has been streamlined since I wrote it. Note that this is not new code, it just moved to a different file.

internal/log_analysis/log_processor/parsers/awslogs/awslogs.go
alxarch: The idea is that group names are arbitrary, they are there to provide context. For example you could have source id as group name and have a separate group for the log types required by a single source. Or you have group 'native' for all native log types. It is coincidental that the prefix and group name coincide for the groups of each parser module.

kostaspap: Got it. Makes sense

PR #1457, opened 2020-09-02
Fail when a source has an unknown log type

alxarch:
## Background

The log processor lambda currently silently ignores invalid log types in a source configuration. Such an error would be hard to pinpoint. This PR fails the log processor if a source has an unknown log type in it's configuration

## Changes

- Fails the log processor if the incoming source has an unknown log type

## Testing

- mage test:ci


rleighton (approved):
Can you add a test that panics (asserting the panic)?

PR #1458, opened 2020-09-03
Refactor mage and CI

austinbyers:
## Background

I'm working on an end-to-end release test script (`mage test:e2e`) and was going to include its first draft in this PR. However, it wound up requiring lots of refactoring to the `mage` code, so for ease of review I'm starting with just the refactor to existing code.

Right now, all of the `mage` code is basically in a single package. `mage` automatically creates a target for every exported function with the right signature, and all the code shares lots of global context. But I want to put the test script in a separate package to encourage future expansion without bloating or clobbering the rest of the mage code.

That proved to be challenging - I need access to lots of mage subroutines for the testing, like the lower-level deploy and compilation functions. I could have exported the routines I needed from the main mage pkg, but then `mage` would automatically create targets for them.

So I was essentially forced to break `mage` apart into separate modular packages. This is better design anyway, for a number of reasons:

* Code is now more clearly grouped with its associated `mage` target. For example, parsing the config file only matters for `mage deploy`, so that code moved to `tools/mage/deploy`
* Global context is pretty much eliminated, making it easier to test and compose mage subroutines
* Now we could easily compile individual mage targets if we wanted to in the future

### CI

- By popular demand, there is no longer a single `test_ci` job. Instead, a single `setup` job compiles mage, installs dependencies, and autoformats. Then `test:cfn`, `test:go`, `test:python`, and `test:web` run in separate jobs:

Screen Shot 2020-09-06 at 11 47 52 AM

The total end-to-end time is about the same as before, but this will scale better and devs will more quickly get feedback if a subset of tests fail. After this PR is approved, I'll update the required checks on the repo accordingly.

- Update CI image dockerfile
    - Update Go 1.14.3 => 1.15.1
    - Update Python 3.7.7 => 3.7.9
    - Update Npm 12.17.0 => 12.18.3
- Auto-formatting now includes API/CloudFormation code generation as well as `mage fmt`. In other words, we now guarantee that CI will commit all autogenerated files
    - Prevent a potential feedback loop: if there was a bug in the autoformatting step, panther-bot might make a commit, which re-triggers CI, which triggers a new commit from panther-bot. Now the autoformatting stops early if the last commit was authored by the bot
- Bugfix: #1372 disabled the entire test suite for the master branch, but it should have skipped only the autofmt step
- FOSSA license analysis now only runs on master - we don't get feedback fast enough to check it on every PR. Instead, we rely on FOSSA notifications for failed scans, so there's no reason to scan every branch
- The CI docker image is ~54 MB smaller than the previous version by removing libraries we didn't need

### Mage

- Update the versions of the .setup binaries: swagger, Terraform, and golangci-lint
- Mage targets are now explicitly constructed in `magefile.go` instead of relying on implicit imports
- Most errors are now bubbled up instead of being fatally logged
- AWS session and clients are constructed lazily - I was having trouble keeping track of which commands were using which clients and whether they had already been instantiated
- Removed `build:api` and `build:cfn` from the deploy process. Building them every time dirties the repo state unless they are also formatted but @alxarch recently added code generation which is harder to fmt in this way, so we've just removed these. The autogenerated files do not change very often, so there's no reason to slow down the deploy process building them every time anyway. If a developer forgets, CI will now automatically commit any changes to these files.
    - `mage gen` now handles `build:api build:cfn` - renamed to more accurately reflect its purpose (*gen*erating code)
- Removed formatting / autogenerated files from `mage test:ci` - now that we guarantee CI will commit changes to these files, we don't need to test for it
- Moved all `tools/mage` code into modular packages:

```
tools/mage
├── build
│   ├── lambda.go
│   ├── layer.go
│   ├── swagger.go
│   ├── swagger_test.go
│   ├── testdata
│   │   ├── api
│   │   │   └── first.yml
│   │   ├── no-api.yml
│   │   ├── valid-api-expected-output.yml
│   │   └── valid-api.yml
│   └── tools.go
├── clean
│   └── clean.go
├── clients
│   └── aws.go
├── deploy
│   ├── config.go
│   ├── deploy.go
│   ├── deploy_template.go
│   └── frontend.go
├── doc
│   ├── cfndoc.go
│   ├── cfndoc_test.go
│   ├── doc.go
│   └── doc_test.go
├── gen
│   ├── dashboards
│   │   ├── alerts.go
│   │   ├── dashboards.go
│   │   ├── dashboards_test.go
│   │   ├── infra.go
│   │   ├── log_processing.go
│   │   ├── overview.go
│   │   └── remediation.go
│   └── gen.go
├── logger
│   └── logger.go
├── master
│   ├── build.go
│   ├── deploy.go
│   └── publish.go
├── setup
│   └── setup.go
├── srcfmt
│   ├── fmt.go
│   └── fmt_license.go
├── teardown
│   ├── cfn.go
│   ├── s3.go
│   └── teardown.go
├── test
│   ├── cfn.go
│   ├── ci.go
│   ├── go.go
│   ├── integration.go
│   ├── python.go
│   └── web.go
└── util
    ├── cfnparse.go
    ├── files.go
    ├── run.go
    ├── s3.go
    ├── task.go
    └── version.go
```

### Other
* Update Go module and Pip library versions
* Bugfix: The `internal/log_analysis/message_forwarder/forwarder` pkg was taking 100s to run because of backoff. Reduced to 2s runtime by setting MaxRetries to 1 during testing
* Allow CloudFormation to assume the deployment role (I use the deployment role during end-to-end testing)

## Testing

- Dozens of CI runs to test the docker image, new workflow, and panther-bot interactions
- Every mage command, including `mage deploy test:integration teardown`


.circleci/config.yml
```diff
@@ -4,129 +4,170 @@ version: 2.1
...
-  restore_go_modules_from_cache:
-    description: 'Restores all Golang modules that were installed & cached in previous executions'
+  restore_bin_cache:
     steps:
       - restore_cache:
-          name: 'Restore cache: go modules'
+          name: 'Restore cache: .setup binaries'
           keys:
-            - gomod2-{{ checksum "go.mod" }}
+            - setup8-{{ checksum "tools/mage/setup/setup.go" }}
```
austinbyers, line 16: I had to change the cache key prefixes to reset the cache, I went with 8 (from v1.8), but it doesn't matter what it is as long as it's different from what it was before

tools/mage/build/api.go
```diff
@@ -0,0 +1,116 @@
...
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
```
austinbyers, line 19: Moved from `build_namespace.go` without any significant changes

tools/mage/build/api.go
```diff
@@ -0,0 +1,116 @@
...
+	"strings"
+
+	"github.com/magefile/mage/mg"
+	"github.com/magefile/mage/sh"
+
+	"github.com/panther-labs/panther/tools/mage/logger"
+	"github.com/panther-labs/panther/tools/mage/util"
+)
+
+var log = logger.Get()
```
austinbyers, line 35: Since mage is now several different packages, they need a way to get the logger. I tried building it in magefile.go and passing it to every function, but that's incredibly tedious.

Instead, each package builds their own global logger and logger.Get() will privately cache the logger if it was built already. I'm open to other ideas here, but this works quick and easy

tools/mage/build/dashboards/dashboards.go
```diff
@@ -40,6 +36,34 @@ type SubString struct {
...
+
+// Dashboards returns all the declared dashboards
+func Dashboards() (dashboards []*Dashboard) {
+	dashboards = append(dashboards, NewDashboard("PantherOverview", overviewJSON))
+	dashboards = append(dashboards, NewDashboard("PantherCloudSecurity", infraJSON))
+	dashboards = append(dashboards, NewDashboard("PantherAlertProcessing", alertsJSON))
+	dashboards = append(dashboards, NewDashboard("PantherRemediation", remediationJSON))
+	dashboards = append(dashboards, NewDashboard("PantherLogAnalysis", logProcessingJSON))
+	return dashboards
+}
```
austinbyers, line 65: Moved from another file with no significant changes

tools/mage/build/lambda.go
```diff
@@ -0,0 +1,78 @@
+package build
```
austinbyers, line 1: Moved from `build_namespace.go` with no significant changes

tools/mage/build/layer.go
```diff
@@ -0,0 +1,68 @@
+package build
```
austinbyers, line 1: Moved from `deploy.go` with no significant changes

tools/mage/build/tools.go
```diff
@@ -0,0 +1,79 @@
+package build
```
austinbyers, line 1: Moved with no significant changes

tools/mage/master/build.go
```diff
@@ -0,0 +1,81 @@
+package master
```
austinbyers, line 1: Moved from `master_namespace.go` with no significant changes

tools/mage/master/deploy.go
```diff
@@ -0,0 +1,152 @@
+package master
```
austinbyers, line 1: Moved from `master_namespace.go` with no significant changes

tools/mage/master/publish.go
```diff
@@ -0,0 +1,66 @@
+package master
```
austinbyers, line 1: Moved from `master_namespace.go` with no significant changes

tools/mage/teardown/cfn.go
```diff
@@ -0,0 +1,97 @@
+package teardown
```
austinbyers, line 1: Moved from `teardown` with no significant changes

tools/mage/test/go.go
```diff
@@ -63,14 +65,14 @@ func testGoUnit() error {
 	}
 
 	// unit tests and race detection
-	return runGoTest("test", "-race", "-p", strconv.Itoa(maxWorkers), "-vet", "", "-cover", "./...")
+	return runGoTest("test", "-race", "-p", strconv.Itoa(util.MaxWorkers), "-cover", "./...")
```
austinbyers, line 68: `go test` runs `vet` automatically, we don't need to add `-vet ""`

tools/mage/util/run.go
```diff
@@ -0,0 +1,100 @@
...
+		// We need to print the output so the user can see the error message.
+		fmt.Println(buf.String())
+		return err
+	}
+
+	return nil
+}
+
+// Deploy CloudFormation template directly with SAM.
+func SamDeploy(stack, path string, params ...string) error {
```
austinbyers, line 63: I will need this for the end-to-end testing

tools/mage/util/version.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+
+	return gitVersion
+}
+
+// Find the most recent published version of Panther in S3, e.g. "1.7.1"
+//
+// This provides an alternative to checking the git tags in the repo.
+func LatestPublishedVersion() (string, error) {
```
austinbyers, line 52: I will use this for end-to-end testing: I deploy the previous published panther release, which I determine automatically from here

tools/mage/fmt/fmt_license.go
```diff
@@ -1,4 +1,4 @@
-package mage
+package fmt
```
alxarch, line 1: Can we use a different package name for this? It will be confusing for IDE import resolution to use the same name as the go std lib `fmt` package.

tools/mage/util/files.go
```diff
@@ -0,0 +1,88 @@
...
+	SetupDir = ".setup"
+)
+
+var (
+	PyEnv     = filepath.Join(SetupDir, "venv")
+	GoLinter  = filepath.Join(SetupDir, "golangci-lint")
+	Swagger   = filepath.Join(SetupDir, "swagger")
+	Terraform = filepath.Join(SetupDir, "terraform")
+
+	GoTargets = []string{"api", "internal", "pkg", "tools", "cmd", "magefile.go"}
```
alxarch, line 39: I see this is only used in `test:go`, should we move that there?

tools/mage/util/files.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+const (
+	NpmDir   = "node_modules"
+	SetupDir = ".setup"
+)
+
+var (
+	PyEnv     = filepath.Join(SetupDir, "venv")
+	GoLinter  = filepath.Join(SetupDir, "golangci-lint")
```
alxarch, line 35: I see this is only used in `test:go`, should we move that there?

alxarch (approved):
This brings some structure to the `mage` tools.
Tried some targets locally, all seems good.
Some minor notes on package names to avoid conflicts with go std library names (ie `fmt`)

.circleci/config.yml
```diff
@@ -4,129 +4,170 @@ version: 2.1
...
-  restore_node_modules_from_cache:
-    description: 'Restores all Node modules that were installed & cached in previous executions'
+  restore_go_cache:
     steps:
       - restore_cache:
-          name: 'Restore cache: npm'
+          name: 'Restore cache: go modules'
           keys:
-            - npm-{{ checksum "package-lock.json" }}
+            - gomod8-{{ checksum "go.mod" }}
```
3nvi, line 23: should we perhaps pull out all the cache keys in variables so they can be shared between `restore_cache` and `save_cache` commands? This might avoid typos (i.e. forgetting to  update  the `8` on `gomod8`)

.circleci/config.yml
```diff
@@ -4,129 +4,170 @@ version: 2.1
...
+  test_web:
+    executor: panther-buildpack
+    resource_class: large
+    steps:
+      - checkout
+      - attach_workspace:
+          at: /tmp/panther
+      - restore_npm_cache
+      - run:
+          name: mage test:web
```
3nvi, line 169: There is a  small bug here now  that is run in parallel with `autofmt`. The front-end static testing is:

- Using `prettier` to check for un-prettified  code
- Using `eslint` to  check for stuff like unused variables, properly cased variables, etc.
- Using `tsc` to check that all the strongly typed code is ok  (I know it's  funny since Go won't even compile if it isn't)

Now, `autofmt` runs `prettier` in those files, meaning it would "fix" any exceptions found by `test:web`. Thus, if there is a prettier issue, then `test:web` will fail, but `autofmt` will fix it. Unfortunately, they are ran in parallel, so the CI will fail. If `autofmt` ran before `test:web` then the CI would be passing.

The solution would be to either have `autofmt` run before the rest of the parallel checks `test:XXX` are ran or to reduce `autofmt`'s responsibilities

3nvi (commented):
Amazing work @austinbyers . Things like that are much appreciated, since they help us iterate way quicker. 

As we grow and we add more and more tests, parallelizations like that can be of significant value. I just have a couple of comments about the parallelization of `autofmt` and `test:XXX`, as  well as making sure the proper python version is installed.

Other than that it LGTM 🎉

rleighton:
Nice refactor! Careful merging into Enterprise!

internal/log_analysis/message_forwarder/forwarder/forwarder_test.go
```diff
@@ -184,6 +184,8 @@ func TestShouldFailIfPartialFailureToPutRecord(t *testing.T) {
 	config.LambdaClient = mockLambda
 	mockFirehose := &testutils.FirehoseMock{}
 	config.FirehoseClient = mockFirehose
+	// This test will trigger backoff, make sure it doesn't take too long
+	config.MaxRetries = 1
```
rleighton, line 188: Since this is global, shouldn't you reset after the test? Perhaps add ABOVE:
```
// reset back to defaults
oldMaxRetries := config.MaxRetries 
defer func() {
    config.MaxRetries  = oldMaxRetries
}()
```

internal/log_analysis/message_forwarder/forwarder/forwarder_test.go
```diff
@@ -218,6 +220,8 @@ func TestShouldFailIfFailureToPutRecord(t *testing.T) {
...
```
rleighton, line 224: Same here, reset back to default

rleighton (approved):

tools/mage/build/api.go
austinbyers: Good idea, I'll try it

.circleci/config.yml
```diff
@@ -4,129 +4,170 @@ version: 2.1
...
+          at: /tmp/panther
+      - restore_bin_cache # swagger (build:api) and terraform (fmt)
+      - restore_go_cache # goimports (fmt)
+      - restore_npm_cache # prettier (fmt)
+      - restore_pip_cache # yapf (fmt)
       - run:
-          name: Run test suite
-          command: mage fmt doc
+          name: Generate and format source files
+          command: /tmp/panther/magebin build:api build:cfn fmt
```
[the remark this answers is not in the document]
austinbyers: Good idea, I like that distinction

deployments/auxiliary/cloudformation/panther-deployment-role.yml
```diff
@@ -52,6 +52,13 @@ Resources:
             Condition:
               Bool:
                 aws:SecureTransport: true
+          - Effect: Allow
```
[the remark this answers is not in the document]
austinbyers: Until now, I guess we are always assuming the role directly *before* running CloudFormation, or we attach the deployment policy to other users or roles. In other words, *we* have been assuming the role, not CloudFormation. For example, in my dev acct, I have a separate IAM user I can use directly: `aws-vault exec my-dev-deploy-user -- mage deploy`

But in the upcoming end-to-end tests, I'm passing the deployment role as a flag directly to CloudFormation:

```
aws-vault exec my-dev-admin-user -- sam deploy --role-arn ...PantherDeploymentRole
```

In this case, my IAM user has admin credentials, but CloudFormation will assume a different role to make the change. We should have always allowed this, I guess we've just never tested it this way until now :)

internal/core/alert_delivery/api/retry.go
```diff
@@ -77,7 +77,7 @@ func createEntry(messageBody string, index int, delaySeconds int64) *sqs.SendMes
 
 // Generate a random int between lower (inclusive) and upper (exclusive).
 func randomInt64n(min int, max int) int64 {
-	return rand.Int63n(int64(max)-int64(min)) + int64(min)
+	return rand.Int63n(int64(max)-int64(min)) + int64(min) // #nosec (don't need cryptographic rng)
```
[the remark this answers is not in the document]
austinbyers: Hmm, I assume the effect would be the same (I assume you mean `//nolint:gosec`), but there is probably a subtle difference in meaning because we use a high-level metalinter.

`#nosec` is a directive understood directly by the `gosec` standalone linter, while `//nolint:gosec` is a directive for the `golangci-lint` metalinter to disable `gosec` entirely for that line.

I'll switch to the `//nolint` format to be consistent with other parts of the Go code. I'm guessing it's also fractionally more efficient since the disable happens earlier in the evaluation

requirements-top-level.txt
```diff
@@ -13,11 +13,11 @@
 # deactivate
 aws-sam-cli
 bandit
-boto3==1.12.49  # Matches version in Lambda: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html
+boto3==1.14.17  # Matches version in Lambda: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html
```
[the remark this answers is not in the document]
austinbyers: It doesn't. I'll see if I can pin some versions to make the warnings go away

tools/mage/logger/logger.go
```diff
@@ -46,11 +45,15 @@ func init() {
 	// Always disable error traces and use color-coded log levels and short timestamps
 	config.DisableStacktrace = true
 	config.EncoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
```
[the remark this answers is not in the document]
austinbyers: You can see colored output in circle CI just fine:

Screen Shot 2020-09-04 at 11 00 03 AM

tools/mage/fmt/fmt_license.go
austinbyers: Sure

tools/mage/util/files.go
austinbyers: I just checked in the IDE - it's actually only used in `fmt`, I can move it there

tools/mage/util/files.go
austinbyers: It should have been used in setup also, I'll update `setup.go` as well

The idea is that the installation (setup) and usage (fmt/test) use the same variables to save and access the binaries

build/images/ci/Dockerfile
```diff
@@ -57,33 +64,6 @@ ENV PATH="$GOPATH/bin:/usr/local/go/bin:$PATH"
 # Install mage
 RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
 RUN go get github.com/magefile/mage
-# ****************     END GO     *******************
-
-
-
-# ****************     PYTHON     *******************
-RUN apt-get update && apt-get install -y \
```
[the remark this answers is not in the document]
austinbyers: Most of those libraries were need to compile Python from source, not to actually run the Python

build/images/ci/Dockerfile
```diff
@@ -14,38 +14,45 @@
...
+
+# ****************     APT Packages     *******************
+# gcc and libc6-dev are necessary for 'go test'
 RUN apt-get update && apt-get install -y --no-install-recommends \
-        g++ \
         gcc \
         libc6-dev \
         make \
-        pkg-config \
+        python3 \
```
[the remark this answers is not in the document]
austinbyers: Fair enough, I can see the value of building a very specific, consistent version from source.

I'll revert to building from source and then see if I can uninstall the compilation libraries afterward to reduce the image size

build/images/ci/Dockerfile
```diff
@@ -14,38 +14,45 @@
...
-ENV NODE_VERSION="12.17.0"
+ENV GOLANG_VERSION="1.15.1"
+ENV NODE_VERSION="12.18.3"
 
-# ****************     GO     *******************
+
+# ****************     APT Packages     *******************
+# gcc and libc6-dev are necessary for 'go test'
 RUN apt-get update && apt-get install -y --no-install-recommends \
-        g++ \
```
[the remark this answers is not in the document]
austinbyers: If you need it in the development image, we would add it to `build/images/deployment` rather than `build/images/ci`

I know you had considered combining these images, but I've realized we need the CI image to be as small as possible - every CI task takes 10-50 seconds to download and unpack our custom image. It's the biggest bottleneck for most of the tests.

In fact, I'm going to experiment with using a standard CircleCI image for some of the tests after we setup the environment to see if it speeds up the builds.

build/images/ci/Dockerfile
```diff
@@ -14,38 +14,45 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+# Published version:
+#     pantherlabs/panther-buildpack:1.8.0
```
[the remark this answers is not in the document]
austinbyers: Yeah, I was just pinning it to the Panther version so we would be able to see at a glance when we last updated it. We do the same thing with our CloudFormation templates, for example - when we update them (which isn't necessarily every release), we bump the version to match the current release. I don't think it's worth our time right now to release a new image version every time since customers typically don't need it. But that would be cool with automation!

I didn't think about the fact that this breaks semver. I don't have any strong preference here, I'll probably leave it as-is (and add some comments here and in Dockerhub) unless you feel that's inappropriate

deployments/master.yml
```diff
@@ -373,6 +373,7 @@ Resources:
     DependsOn: [CloudSecurity, Core, LogAnalysis]
     Properties:
       TemplateURL: web_server.yml
+      TimeoutInMinutes: 20 # if something goes wrong, the ECS service will get stuck in a retry loop for 4 hours
```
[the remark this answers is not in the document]
austinbyers: Good to know! But this way, if there is an issue, a customer isn't waiting for hours and hours to realize something went wrong

tools/mage/build/api.go
```diff
@@ -0,0 +1,116 @@
...
+
+			newBody := clientImport.ReplaceAll(body, []byte(correctImport))
+			util.MustWriteFile(path, newBody)
+			return nil
+		})
+	}
+	return nil
+}
+
+func generateWebTypescript() error {
```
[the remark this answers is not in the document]
austinbyers: Good idea, will do

.circleci/config.yml
austinbyers: Good idea, I'll see if I can use YAML anchors for this

.circleci/config.yml
austinbyers: I agree the parallel autofmt is a bit strange - it works for reasons I'll explain below, but it does result in wasted cycles and potentially some confusion. I'll see if I can just move it to the `setup` phase for simplicity - it should be fast enough.

> Using prettier to check for un-prettified code

Since we're auto-formatting, we don't need any of the `tests` to check for formatting (in my opinion). This PR already removes format checks from `test:ci`, I'll remove them from `test:web` as well if it's easy to do so.

-----

> run in parallel with autofmt

Quick note about why I chose to do it this way. There are two possibilities:

1. `autofmt` commits changed code, or
2. there are no changes to commit

In the first case, *CI will immediately stop and restart* because there is a new commit. Once `autofmt` finds changes, the rest of the tests won't matter because they won't even finish.

If there were no changes (case 2), then blocking the rest of the tests to wait for `autofmt` was a waste of time. Because of the expensive container setup, this could be as high as 60s. So the idea was to always run the tests as soon as possible, even if we wind up stopping them early because we found some code changes. However, as long as the autofmt is sufficiently fast, I may be able to move it into setup, like I said. I'll experiment with that

internal/log_analysis/message_forwarder/forwarder/forwarder_test.go
austinbyers: I was just following the style around the code - the tests run sequentially and each one changes the global config state to what it needs (which I know is a problematic pattern which @alxarch has called out before). I don't think the changed global matters for tests in other packages, but I'm not sure.

I'll move at least my change into the `TestMain` function designed for this sort of thing, with the defer, just in case.

tools/mage/build/api.go
austinbyers: I added the names, but I decided to add brackets around the name to clean up the output a bit. Otherwise, zap will sometimes butt up the logger name with the output in a way that's hard to read:

```
19:10:28	INFO	deploy	docker build web server (deployments/Dockerfile)
19:10:29	INFO	deploy	create CloudFormation stack panther-onboard
19:12:21	INFO	deploy	    √ panther-onboard finished (8/9)
19:13:35	INFO	deploy	pushing docker image to remote repo
19:13:49	INFO	deploy	create CloudFormation stack panther-web
```

so I switched to:

```
19:20:55	INFO	[deploy]	docker build web server (deployments/Dockerfile)
19:20:55	INFO	[deploy]	update CloudFormation stack panther-onboard
19:21:05	INFO	[deploy]	pushing docker image to remote repo
19:21:07	INFO	[deploy]	update CloudFormation stack panther-web
```

But feel free to tweak the logger names in the future if you don't like how they look.

.circleci/config.yml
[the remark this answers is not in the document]
austinbyers: Done - removed `build:api` and `build:fmt` and replaced it with just `mage gen`. We can split out separate `gen` targets later if needed

magefile.go
```diff
@@ -21,6 +21,115 @@ package main
...
+	"github.com/panther-labs/panther/tools/mage/clean"
+	"github.com/panther-labs/panther/tools/mage/deploy"
+	"github.com/panther-labs/panther/tools/mage/doc"
+	"github.com/panther-labs/panther/tools/mage/fmt"
+	"github.com/panther-labs/panther/tools/mage/master"
+	"github.com/panther-labs/panther/tools/mage/setup"
+	"github.com/panther-labs/panther/tools/mage/teardown"
+	"github.com/panther-labs/panther/tools/mage/test"
 )
+
```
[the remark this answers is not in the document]
austinbyers: I wasn't sure if you'd like this better (it's a bit tedious/duplicated), or if you'd prefer the more explicit documented form. I think this is better for understanding, I'm glad you agree!

requirements-top-level.txt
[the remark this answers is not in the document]
austinbyers: I was able to pin versions to avoid the pip warnings!

tools/mage/build/api.go
```diff
@@ -0,0 +1,116 @@
...
+var log = logger.Get()
+
+// Generate API source files from GraphQL + Swagger
+func API() {
+	mg.Deps(generateSwaggerClients, generateWebTypescript, goGenerate)
+}
+
+func goGenerate() error {
+	const generatePattern = "./..."
+	log.Info("build:api: generating Go code with go:generate")
```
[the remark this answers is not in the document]
austinbyers: Done! Removed all these log prefixes

tools/mage/fmt/fmt_license.go
austinbyers: Switched to `srcfmt`

build/images/ci/Dockerfile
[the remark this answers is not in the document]
austinbyers: I re-added most of them after restoring the Python build

build/images/ci/Dockerfile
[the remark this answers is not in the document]
austinbyers: I added Python 3.7.9 back, but installed with `pyenv` rather than compiled manually from source, which is the same thing that CircleCI does in their [python convenience image](https://github.com/CircleCI-Public/cimg-python/blob/master/3.7/Dockerfile)

build/images/ci/Dockerfile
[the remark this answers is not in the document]
austinbyers: I restored g++ (based on CircleCI go image here: https://github.com/CircleCI-Public/cimg-go/blob/master/1.15/Dockerfile)

.circleci/config.yml
austinbyers: Done! The anchors worked well

.circleci/config.yml
austinbyers: I moved the autofmt to the beginning, so that should resolve your concern

internal/log_analysis/message_forwarder/forwarder/forwarder_test.go
austinbyers: Added TestMain to unset at the end of the test run

.circleci/config.yml
3nvi: :+1

build/images/ci/Dockerfile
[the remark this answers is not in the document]
3nvi: sure thing. I don't feel strongly at all about this one

build/images/ci/Dockerfile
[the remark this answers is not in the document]
3nvi: sure thing. The initial Python installation commands were done in haste, so any improvements to the installation process are much appreciated

build/images/ci/Dockerfile
```diff
@@ -14,87 +14,83 @@
...
+        # CGO
+        g++ \
+        libc6-dev \
+        # Python
         build-essential \
-        libreadline-gplv2-dev \
+        ca-certificates \
+        libbz2-dev \
+        liblzma-dev \
+        libncurses5-dev \
```
3nvi, line 45: most  likely  you don't need both curse libs

build/images/ci/Dockerfile
```diff
@@ -14,87 +14,83 @@
...
         libncursesw5-dev \
-        libssl-dev \
-        libsqlite3-dev \
-        libgdbm-dev \
+        libreadline-dev \
         libffi-dev \
-        zlib1g-dev \
-        libnss3-dev
+        libsqlite3-dev \
```
3nvi, line 49: most of those deps are un-needed by us. So far I maintained a slim list of the ones needed for development  in Panther. 

I don't mind adding more, but I feel that we might be increasing the image's footprint, meaning slower downloads & builds. If possible, try and see if some of the big ones are un-needed

3nvi (approved):
Great refactor on the Dockerfile.

My only (and last) concern is docker image size. I haven't compared it to the previous one, but I see some extra images installed. I'm pro-actively  approving since I feel that this shouldn't block the PR merging at all, but if possible, I'd like to see if we can remove any un-needed libs from the `Dockerfile` declaration.

Other than that 🚀 🚀 🚀

austinbyers:
> My only (and last) concern is docker image size. I haven't compared it to the previous one, but I see some extra images installed.

As you may have seen, I refactored the Dockerfile by copying the installation steps used by CircleCI in their own Go, Node, and Python images. That's where the extra libraries came from - CircleCI uses them to install Python. Some Python features will be silently disabled unless it has the necessary headers during installation.

panther-buildpack v1.2.0 is 439 MB compressed, while this one is currently 467 MB, so slightly larger. I agree we want the image to be as small as possible, but I also like knowing that our installation has the same libraries that they use.

I will try removing as many of the Python libraries as possible - we probably don't need most of them after Python is installed.

## Other Things I Tried
I considered using CircleCI's official [base image](https://circleci.com/developer/images/image/cimg/base), but the `buildpack-deps:buster-scm` base image we use today is substantially smaller.

I also considered using the CircleCI official Go/Npm/Python images for running those respective test groups. The total image size of the CircleCI images is comparable to what we have today, but because their are commonly used, there is a much higher chance that the image is cached on the host, bringing the setup time for each build from 30s to 2s.

For now, I'm leaving all the tests with our own image, so that we don't have to reason about running in different environments. But one day I may revisit using convenience images to get higher cache hits.

austinbyers:
I was able to make the CI image substantially smaller (54 MB smaller than v1.2.0 buildpack) by:

* Removing the default Python2 installation
* Removing version control systems we don't use: mercurial + subversion
* Removing Python header packages after Python is installed

austinbyers:
I've updated the required checks in the repo and published the new CI/dev image to DockerHub.

PR #1459, opened 2020-09-03
Added eventsLatency output

nhakmiller:
## Background

When adding the `eventsLatency` metric, I forgot to add a new output for it. So it just overwrote the `eventsProcessed` metric. Giving it its own entry in the output struct now.

Feel free to merge on approval if this works for you to unblock any frontend work.

## Changes

- Updated the outputs struct to include a `eventsLatency` field

## Testing

- Manual testing


3nvi (approved):
LGTM. @alexmylonas I think we need to update the graphql schema to include this

PR #1461, opened 2020-09-03
Fix conflict for GitLab logs

### LINKED ISSUE #1460: Classifier fails to distinguish between Gitlab.API logs and Gitlab.Production logs
### Describe the bug

Depending on the state of the classifier queue Gitlab.API logs can end up into the Gitlab.Production table

### Steps to reproduce

Not easy to reproduce with certainty because of the randomness in the classifier queue

### Expected behavior

There should be a distinction between the two

### How to fix

Based on samples we just need to add `validate:"required"` to both `controller` and `action` fields in the `Production` schema struct

alxarch:
## Background

Fixes #1460 

## Changes

- Adds required tag to action and controller fields for Production logs
- Adds unit test to ensure production parser does not match API log sample

## Testing

- mage test:ci


rleighton (approved):

alxarch:
I had to add a new field `etag_rorute` to the `Production` struct. It seems that when the production log includes a redirect the `controller` and `action` fields are missing but the `etag_route` is included. The `etag_route` seems to be have the same value as the `action` field. So I added a `validate:"required_without=EtagRoute"` tag to the `action` field and a `validate:"required_without=Action"` tag to the `etag_route` field. This way one of the two is always required.

PR #1465, opened 2020-09-03
http redirect

### LINKED ISSUE #452: HTTP -> HTTPS Redirect
### Description

Currently the Panther app only listens for HTTPS connections on port 443. This is good because we don't want to be serving insecure connections, but we should re-direct HTTP/80 requests to HTTPS/443 so that you don't get a connection refused page if you don't type out https:// every time (or have an always HTTPS plugin).

### Acceptance Criteria

- When attempting to connect to the Panther app over HTTP, automatically get re-directed to the HTTPS listener instead of getting a connection refused page

lindsey-w:
## Background

Panther only had listener on 443. Trying to connect on port 80 resulted in a connection refused page.  We can automatically redirect 80 -> 443 rather than showing an error page.  Closes #452  

## Changes

- Added new http istener in web_server.yml to redirect 80-> 443 

## Testing

- `mage test:ci` and manual testing


nhakmiller (approved):
Thanks for grabbing this one! I'd maybe add either a comment or even just update the name to indicate this listener is serving as a redirect, something like `PublicHttpRedirectListener` maybe.

PR #1469, opened 2020-09-04
Remove unused model field from the alert

s0l0ist:
## Background

The alert model for DDB contained an unused field `OutputIds` which is not stored in DDB nor used in the application.

This issue did not cause any bugs, it is superfluous and is being removed.

## Changes

- removed `OutputIds` from the model
- updated tests to reflect the change
- updated application code where this field was being set

## Testing

- mage test:go
- mage test:ci

### DIFF
--- a/internal/core/alert_delivery/api/deliver_alert.go
+++ b/internal/core/alert_delivery/api/deliver_alert.go
@@ -125,7 +125,7 @@ func populateAlertData(alertItem *alertTable.AlertItem) (*deliveryModels.Alert, 
 		Type:                deliveryModels.RuleType,
 		CreatedAt:           alertItem.CreationTime,
 		Severity:            alertItem.Severity,
-		OutputIds:           alertItem.OutputIds,
+		OutputIds:           []string{}, // We do not pay attention to this field
 		AnalysisDescription: aws.String(string(rule.Description)),
 		AnalysisName:        aws.String(string(rule.DisplayName)),
 		Version:             &alertItem.RuleVersion,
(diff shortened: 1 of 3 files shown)


internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -125,7 +125,7 @@ func populateAlertData(alertItem *alertTable.AlertItem) (*deliveryModels.Alert,
 		Type:                deliveryModels.RuleType,
 		CreatedAt:           alertItem.CreationTime,
 		Severity:            alertItem.Severity,
-		OutputIds:           alertItem.OutputIds,
+		OutputIds:           []string{}, // We do not pay attention to this field
```
s0l0ist, line 128: Previously, this field was getting set, but the value was never used. Initializing to empty to illustrate that the field is ignored.

internal/log_analysis/alerts_api/table/table.go
```diff
@@ -76,7 +76,6 @@ type AlertItem struct {
 	FirstEventMatchTime time.Time                  `json:"firstEventMatchTime"`
 	CreationTime        time.Time                  `json:"creationTime"`
 	DeliveryResponses   []*models.DeliveryResponse `json:"deliveryResponses"`
-	OutputIds           []string                   `json:"outputIds"`
```
austinbyers, line 79: I'm guessing the idea was that we want a record of all alerts that have been delivered to their destination. If this was blank, that was probably a bug.

Are we storing that somewhere else now? Can you tell me for every alert what destinations it sent to?

internal/log_analysis/alerts_api/table/table.go
```diff
@@ -76,7 +76,6 @@ type AlertItem struct {
...
 	// UpdateTime - stores the timestamp from an update from a dedup event
```
s0l0ist, line 79: This is the offending field that is not needed in the model.

internal/log_analysis/alerts_api/table/table.go
s0l0ist: Yes, all that info is in the field above `DeliveryResponses`. This is an artifact from early development.

austinbyers (approved):

PR #1471, opened 2020-09-04
Sensitive TextInput

vorillaz:
## Description
This one introduces a new `SensitiveInput` component that actually hides the inserted value in order to prevent over-the-shoulder attacks. So instead of using an input with `type="password" we have created an enhanced input with a tooltip that informs the users for the UX.

### Designs:
https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/6c4082e9-dc0b-4e26-b60e-cc96e60306ee/commits/fc0e3669ae46aad9ff03e270ad13f07d58e1b9b9/files/8cc3cb34-8b80-405f-97ad-3adafc365a4f/layers/2795B634-BEA5-4AFF-A5C1-757F29AF007B

### Screenshots:

Sep-04-2020 17-44-59

### Checklist:


web/src/components/forms/AsanaDestinationForm/AsanaDestinationForm.tsx
```diff
@@ -65,8 +66,7 @@ const AsanaDestinationForm: React.FC<AsanaDestinationFormProps> = ({ onSubmit, i
           required
         />
         <Field
-          as={FormikTextInput}
-          type="password"
+          as={SensitiveTextInput}
           name="outputConfig.asana.personalAccessToken"
           label="Access Token"
           placeholder={
```
3nvi, line 72: can we convert all placeholders to their default values? (i.e. remove the "information is hidden...")

web/src/components/fields/SensitiveTextInput/SensitiveTextInput.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <TextInput
+          {...props}
+          onFocus={onFocus}
+          onBlur={onBlur}
+          invalid={isInvalid}
+          aria-describedby={isInvalid ? errorElementId : undefined}
+          value={masked ? DEFAULT_SENSITIVE_VALUE : value}
+          type="password"
+          icon="alert-circle-filled"
+          iconProps={{ color: 'violet-300' }}
```
3nvi, line 52: `TextInput` implements `React.memo` under the hood. Can  we not inline the object, but defined it in a constant outside the component and use it (so it keeps the same reference)?

web/src/components/fields/SensitiveTextInput/SensitiveTextInput.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { Box, FormError, TextInput, TextInputProps, Tooltip } from 'pouncejs';
+import { FieldConfig, useField } from 'formik';
+import { DEFAULT_SENSITIVE_VALUE } from 'Source/constants';
+
+const FormikTextInput: React.FC<TextInputProps & Required<Pick<FieldConfig, 'name'>>> = props => {
+  const [isFocused, setFocused] = React.useState(false);
+  const [, meta] = useField(props.name);
+
+  const { touched, error, value } = meta;
+  const masked = React.useMemo(() => value === '' && !isFocused, [value, isFocused]);
```
3nvi, line 29: optional: The result is a `boolean`, so `React.useMemo` is un-needed. Feel free to keep it though

3nvi (approved):
Awesome work @vorillaz , just some minor comments, but nothing that should block the PR 🎉 

When syncing on the enterprise level, let's check if we have more sensitive fields outside of destinations (i.e. on SaaS log sourcces) that we'll need to replace!

web/src/components/forms/AsanaDestinationForm/AsanaDestinationForm.tsx
vorillaz: @3nvi I am not sure if the placeholder should get a default value or should be different per case (Jira API token, Slack token etc)

3nvi: Yeah yeah that's what  I meant 😆 Basically, remove the "information is hidden..." and keep the other part (whatever it was for each field)

PR #1474, opened 2020-09-05
Fix gluesync and datacatalog updater to properly update and sync all glue partitions.

### LINKED ISSUE #1466: Gluesync tool should search to update partitions at dates before the table creation time
### Describe the bug

> **NOTE**: A quickfix for this should be included in the 1.8 to avoid surprises during upgrades

Currently `gluesync` needs a specific date in the past to sync partitions at dates before the Glue table creation time.
This is problematic because historical data and events from misconfigured devices can put partitions at arbitrary points in the past. Especially in 1.8 release we added the `source_id` and `source_label` fields to *all tables*, which requires all Glue tables to be updated and all partitions synced. For tables that have data at some point in the past of table creation time this **will break queries** after deployment and will require manual intervention to fix. Specifically the person deploying should find out which is the oldest date they have data for in the problematic tables and run `gluesync` for these with the appropriate start date.

### Expected behavior

Ideally the deployment should handle the migration transparently to the user

### Proposed fix

After a short discussion with @rleighton we decided on adding a query to the Glue API for table partitions *before table creation time* and also syncing those during a table sync. This will add some delay to the tool but it will ensure that all partitions of the table will be updated to the latest schema.

alxarch:
## Background

AWS Glue does not propagate table schema changes to all defined partitions. To handle this we had some custom code in the `awsglue` package exposed as a CLI tool (`gluesync`) and as an internal Lambda call (`datacatalog_updater`). That code was required a specific time range to perform the sync. The problem is that for processed logs we *cannot know all possible partition dates* unless we scan all partitions known to Glue. This resulted in 'partial updates' during deployments or when using the `gluesync` tool, which in turn resulted in broken Athena queries due to table/partition inconsistencies. In addition to requiring a date range the `gluesync` tool was trying to recover partitions missing from Glue but with data on S3 at the same time. 
The code in this PR tries to fix the issue by first separating the two tasks. 

The `gluesync` task syncs *all partitions of a table* to match it's schema. It *does not accept any date ranges as input* nor does it make assumptions for the time range based on table creation time, to avoid partial table updates that leave the system with broken queries.

The `gluerecover` task *scans for table partitions on a specific date range* and then tries to recover missing partitions by looking for `S3` objects for missing partitions.

In addition, it refactors the custom resource handler code that delegates syncing table partitions through Lambda calls to `datacatalog_updater` to use the `gluetasks.SyncDatabaseTables` task. This adds a common `traceId` to be able to easily track all lambda invocations that were required to complete a full database sync.

## Changes
- Adds `gluetasks` module to handle sync and recover tasks for AWS Glue partitions. This module fixes the issue with partition times when syncing glue partitions and features fully concurrent processing of requests to minimize the overall time required to complete a sync.
- Refactors `datacatalog_updater` to use `gluetasks`  for partition updates and to have better logging by propagating the original sync request id as a trace ID to all sync tasks spawned by deferred Lambda calls.
- Splits `gluesync` CLI tool into `gluesync` and `gluerecover`. The `gluesync` CLI tool only handles table partition updates for *all partitions known to Glue*. The `gluercover` tries to *recover partitions in a specific time range* by scanning S3 to find unregistered partition data.

## Testing

- Ran the tool with a forced column update and verified it works.
- Triggered a datacatalog sync with 'DryRun' on the lambda and watched the scanning proceed without partition modifications.
- Deployed with CUSTOM_RESOURCE_VERSION to watch the Lambda cascade over the existing tables.
- Added a new source and sent some data to verify the tables get properly managed.

NOTE: I will be performing a final round of the above tests before merging this PR to verify any changes that will be introduced for the provided feedback do not break deployments.


go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.6.1
 	github.com/tidwall/gjson v1.6.0
+	go.uber.org/multierr v1.5.0
```
alxarch, line 29: This little package is invaluable for grouping task sub-errors

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -49,7 +49,8 @@ var (
 
 	// The default max number of resources to scan at once. We will keep paging until we scan this
 	// many resources, then do one additional page worth of resources
-	defaultBatchSize   = 100
+	defaultBatchSize = 100
+	// nolint: gosec
```
alxarch, line 53: Irrelevant to the PR but the new linter version complained

internal/core/alert_delivery/api/retry.go
```diff
@@ -77,6 +77,7 @@ func createEntry(messageBody string, index int, delaySeconds int64) *sqs.SendMes
 
 // Generate a random int between lower (inclusive) and upper (exclusive).
 func randomInt64n(min int, max int) int64 {
+	// nolint: gosec
```
alxarch, line 80: Irrelevant to the PR but the new linter version complained

internal/core/custom_resources/resources/glue.go
```diff
@@ -53,13 +51,14 @@ func customUpdateGlueTables(_ context.Context, event cfn.Event) (string, map[str
...
-			if err != nil {
-				if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == glue.ErrCodeAlreadyExistsException {
-					zap.L().Info("database exists", zap.String("database", pantherDatabase))
-				} else {
-					return "", nil, errors.Wrapf(err, "failed creating database %s", pantherDatabase)
-				}
+			if err == nil {
+				continue
+			}
+			if awsglue.IsAlreadyExistsError(err) {
```
alxarch, line 57: Use the new helper for detecting error type

internal/core/custom_resources/resources/glue.go
```diff
@@ -96,21 +95,20 @@ func customUpdateGlueTables(_ context.Context, event cfn.Event) (string, map[str
...
-			if err != nil {
-				if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == glue.ErrCodeEntityNotFoundException {
-					zap.L().Info("already deleted", zap.String("database", pantherDatabase))
-				} else {
-					return "", nil, errors.Wrapf(err, "failed deleting %s", pantherDatabase)
-				}
+			if err == nil {
+				continue
+			}
+			if awsglue.IsEntityNotFoundError(err) {
```
alxarch, line 105: Use the new helper for detecting error type

internal/log_analysis/awsglue/glue_timebin.go
```diff
@@ -48,24 +50,79 @@ func (tb GlueTableTimebin) Validate() (err error) {
 	return
 }
 
+func (tb GlueTableTimebin) Truncate(t time.Time) time.Time {
```
alxarch, line 53: Use the more standard `time.Date` to handle time normalization

internal/log_analysis/awsglue/glue_timebin.go
```diff
@@ -48,24 +50,79 @@ func (tb GlueTableTimebin) Validate() (err error) {
...
 	case GlueTableDaily:
-		return t.Add(time.Hour * 24).Truncate(time.Hour * 24)
+		return time.Date(t.Year(), t.Month(), t.Day()+1, t.Hour(), 0, 0, 0, t.Location())
 	case GlueTableMonthly:
-		// loop a day at a time until the month changes
-		currentMonth := t.Month()
-		for next = t.Add(time.Hour * 24).Truncate(time.Hour * 24); next.Month() == currentMonth; next = next.Add(time.Hour * 24) {
-		}
-		return next
+		return time.Date(t.Year(), t.Month()+1, t.Day(), t.Hour(), 0, 0, 0, t.Location())
```
alxarch, line 74: Use the more standard `time.Date` to handle time normalization

internal/log_analysis/awsglue/glue_timebin.go
```diff
@@ -48,24 +50,79 @@ func (tb GlueTableTimebin) Validate() (err error) {
...
 	default:
 		panic(fmt.Sprintf("unknown GlueTableMetadata table time bin: %d", tb))
 	}
 }
 
+// PartitionsBefore returns an expression to scan for partitions before tm
```
alxarch, line 80: These helper methods return expression filters for Glue partitions

internal/log_analysis/awsglue/glue_timebin.go
```diff
@@ -82,6 +139,33 @@ func (tb GlueTableTimebin) PartitionValuesFromTime(t time.Time) (values []*strin
 	return
 }
 
+func (tb GlueTableTimebin) PartitionTimeFromValues(values []string) (tm time.Time, ok bool) {
```
alxarch, line 142: This is the inverse of the above function to find the time of a partition based on the partition values

internal/log_analysis/awsglue/sync.go
```diff
@@ -0,0 +1,570 @@
...
+		TableName:          tbl.Name,
+	}
+	_, err := s.GlueClient.UpdatePartitionWithContext(ctx, &input)
+	return err
+}
+
+func needsUpdate(tbl *glue.TableData, p *glue.Partition) bool {
+	want := tbl.StorageDescriptor.Columns
+	have := p.StorageDescriptor.Columns
+	//s.Logger.Debug("diff", zap.Any("colsWant", want), zap.Any("colsHave", have))
```
kostaspap, line 174: Remove?

internal/log_analysis/awsglue/sync.go
```diff
@@ -0,0 +1,570 @@
...
+	return p, true, nil
+}
+
+// IsEntityNotFoundError checks if an AWS error has an EntityNotFoundException code.
+func IsEntityNotFoundError(err error) bool {
+	if awsErr, ok := err.(awserr.Error); ok {
+		return awsErr.Code() == glue.ErrCodeEntityNotFoundException
+	}
+	return false
+}
```
kostaspap, line 259: We have this one you could use: https://github.com/panther-labs/panther-enterprise/blob/5dfc0329eb760b4395caf6d54a9d1a871669c49b/pkg/awsutils/awsutils.go#L11-L27

internal/log_analysis/awsglue/sync.go
```diff
@@ -0,0 +1,570 @@
...
+
+// Recover date tries to recover Glue partitions for a specific date
+func (s *SyncTask) RecoverDate(ctx context.Context, tm time.Time) (*RecoverResult, error) {
+	partitions, err := s.ScanDatePartitions(ctx, tm)
+	if err != nil {
+		return nil, err
+	}
+	if len(partitions) == 24 {
+		return nil, err
+	}
```
kostaspap, line 337: I think this would be more clear if it was 
```
return nil, nil
```

internal/log_analysis/awsglue/sync.go
```diff
@@ -0,0 +1,570 @@
...
+	if s.Start.IsZero() {
+		return tb.PartitionsBefore(s.End.UTC())
+	}
+	if s.End.IsZero() {
+		return tb.PartitionsAfter(s.Start.UTC())
+	}
+	return tb.PartitionsBetween(s.Start.UTC(), s.End.UTC())
+}
+
+func (s *SyncTask) ScanDatePartitions(ctx context.Context, tm time.Time) (map[time.Time]*glue.Partition, error) {
```
kostaspap, line 307: nit: It can be lowercased and moved below the "RecoverDate" function where it is used.

internal/log_analysis/awsglue/sync.go
```diff
@@ -0,0 +1,570 @@
...
+			}
+		}
+		return true
+	}
+	err = s.GlueClient.GetPartitionsPagesWithContext(ctx, &input, scanPage)
+	return
+}
+
+// Sync scans all Glue partitions in a time range and updates their descriptors to match the table descriptor.
+func (s *SyncTask) Sync(ctx context.Context, start, end time.Time) (*SyncSummary, error) {
```
kostaspap, line 97: If I interpret the code correctly, we collect all errors while syncing partitions and we display them at the end to the user .

We might want to display to the user a warning as we encounter errors. They might want to terminate the run, fix any issue, re-run the tool. This matter more considering this tool could be running for several minutes (if not hours)

cmd/opstools/gluefix/gluefix/main.go
```diff
@@ -0,0 +1,183 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import (
```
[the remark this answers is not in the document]
alxarch: No, it doesn't do they need to update their schema as well?

internal/log_analysis/awsglue/sync.go
```diff
@@ -0,0 +1,570 @@
...
+
+// ListLogTables returns a list of available tables in the log processing database
+func ListLogTables(ctx context.Context, client *glue.Glue) ([]*glue.TableData, error) {
+	var tables []*glue.TableData
+	scanPage := func(page *glue.GetTablesOutput, _ bool) bool {
+		tables = append(tables, page.TableList...)
+		return true
+	}
+	input := &glue.GetTablesInput{
+		DatabaseName: aws.String(LogProcessingDatabaseName),
```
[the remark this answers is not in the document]
alxarch: Do we run queries against those tables as well? I thought the ones exposed to data explorer were the relevant ones.

cmd/opstools/gluefix/gluefix/main.go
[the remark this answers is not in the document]
kostaspap: Well, rule matches tables should have the same schema as the original tables + some extra fields. So any core field that gets added to the "log" tables should be present in the "rules" tables as well.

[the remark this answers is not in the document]
alxarch: OK, will update the tool to accept a flag for fixing either `log` or `rules` tables or run the two in sequence for both if no value is provided.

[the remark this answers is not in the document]
kostaspap: Sounds good. By default all users should be updating both, but good to give them options.

austinbyers (commented):
Is this `gluefix` intended to replace the `gluesync` tool? Which binary will we publish for the release? Both of them?

rleighton:
> Is this `gluefix` intended to replace the `gluesync` tool? Which binary will we publish for the release? Both of them?

I don't want to use this for this release but I do want to these improvements for next release (I'll move milestone). @austinbyers I will need to add some release instructions for customers that have backfilled data because this release does have schema changes. Once we get the changes suggested by this PR  coded we won't need the release instructions.

internal/log_analysis/awsglue/glue_timebin.go
rleighton: I did not know `Date()` handle wrapping ! That's much better.

internal/log_analysis/awsglue/glue_timebin.go
```diff
@@ -93,6 +207,31 @@ func (tb GlueTableTimebin) PartitionS3PathFromTime(t time.Time) (s3Path string)
...
+		return "year=2006/month=01"
+	case GlueTableDaily:
+		return "year=2006/month=01/day=02"
+	case GlueTableHourly:
+		return "year=2006/month=01/day=02/hour=15"
+	default:
+		return ""
+	}
+}
+func (tb GlueTableTimebin) TimeFromS3Path(path string) (time.Time, bool) {
```
rleighton, line 222: New line above. It would be good if you added a comment explaining the return values.

internal/log_analysis/awsglue/sync.go
```diff
@@ -0,0 +1,661 @@
...
+		return "", err
+	}
+	bucket, tblPrefix, err := ParseS3URL(*tbl.StorageDescriptor.Location)
+	if err != nil {
+		return "", errors.WithMessagef(err, "failed to parse S3 path for table %q", aws.StringValue(tbl.Name))
+	}
+	objPrefix := path.Join(tblPrefix, bin.PartitionS3PathFromTime(tm)) + "/"
+	listObjectsInput := s3.ListObjectsV2Input{
+		Bucket: aws.String(bucket),
+		Prefix: aws.String(objPrefix),
```
rleighton, line 228: You should set the max returned to 1. No need to return default 1000

internal/log_analysis/awsglue/sync.go
```diff
@@ -0,0 +1,661 @@
...
+	if p == nil || p.ErrorDetail == nil {
+		return nil
+	}
+	code := aws.StringValue(p.ErrorDetail.ErrorCode)
+	if code == glue.ErrCodeAlreadyExistsException {
+		return nil
+	}
+	message := aws.StringValue(p.ErrorDetail.ErrorMessage)
+	tm, _ := GlueTableHourly.PartitionTimeFromValues(aws.StringValueSlice(p.PartitionValues))
+	// Using fmt.Errorf to not add stack
```
rleighton, line 471: Why not add stack?

rleighton (commented):
Can you update: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/datacatalog_updater/process/sync.go to use the new code.

Also, you should remove the old sync code and `gluesync` and update the documentation substituting `gluefix` for `gluesync`

I like where this is going. Let's try to get this well tested for 1.9

internal/log_analysis/awsglue/sync.go
alxarch: This is a helper function to convert the `glue.PartitionError` the error did not occur here. We add the stack where the error occurred. This is similar to how `t.Helper()` skips the stack for that function in testing.

internal/log_analysis/awsglue/sync.go
alxarch: Since this is very glue-specific error handling I wanted to keep these helper functions here. This one is simple enough for the compiler to inline and for us to review. For example in the scope of this code base I am certain that this function is only used *before* an error is wrapped with `errors.Wrapf` so I don't have to use `errors.As` and reflection to check it. I saw a PR for this while on PTO. I am skeptical on the usefulness of utility functions that generalize such short and trivial functions. For example on an IDE with autocomplete it's much easier to use a specific function name than to remember the name of a `const` in an external package.

internal/log_analysis/awsglue/sync.go
alxarch: > Can you update: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/datacatalog_updater/process/sync.go to use the new code.
> 
The issue with datacatalog_updater is that, because of the limited runtime of a Lambda handler, It uses a 'recursive' call to progress through the needed tasks. The code for the tool would require some refactoring to be able to break things up in such 'chunks' of tasks and allow to be used with the datacatalog updater. To split tasks in a 'safe' way we should be using each page of the glue partition scan as the unit of task and perform the next Lambda call with the pagination info. I am not sure this works across aws sessions though (ie if the pagination info of a scan made from one session could be continued by another lambda)

internal/log_analysis/awsglue/sync.go
kostaspap: Check like this happens VERY frequently in the codebase. The reason the utility function was created because we have checks like this in many places in the code.

kostaspap:
Have we tested the performance of this tool? 
Can you run it against one of our deployments that have lots of partitions (e.g. demo account) and report the time it took?

internal/log_analysis/awsglue/sync.go
alxarch: OK, will use the helper

internal/log_analysis/awsglue/sync.go
alxarch: I added error logging inside the `SyncTable` method. The level used is `Error`, do you want to change it to `Warn`?

internal/log_analysis/awsglue/glue_timebin.go
alxarch: Fixed

internal/log_analysis/awsglue/sync.go
alxarch: I opted for not specifying the MaxKeys since we have a check for zero size objects. What is the probability of zero sized S3 objects to exist in a partition? Do we leave such empty objects in the bucket? If there is common to have such objects we should use something higher than `1` to avoid a second round trip if possible.

kostaspap: > I opted for not specifying the MaxKeys since we have a check for zero size objects. What is the probability of zero sized S3 objects to exist in a partition? Do we leave such empty objects in the bucket? If it is common to have such objects we should use something higher than `1` to avoid a second round trip if possible.

We shouldn't have any 0 sized objects.

rleighton:
Keep sure you add this change from `gulesync` to work with master stack: https://github.com/panther-labs/panther/blob/master/cmd/opstools/gluesync/gluesync/main.go#L213

austinbyers:
If this is still intended for `v1.9.0` be sure to add the `sync-to-release` label

rleighton:
> If this is still intended for `v1.9.0`, the easiest way to proceed is to:
> 
> 1. Change the base branch to `master`
> 2. Add the `sync-to-release` label

No, should be 1.10. Changed

austinbyers:
> No, should be 1.10. Changed

Got it. But either way, the base branch should be changed to `master` since we don't directly commit to `release` anymore

rleighton:
> > No, should be 1.10. Changed
> 
> Got it. But either way, the base branch should be changed to `master` since we don't directly commit to `release` anymore

Yup. Done.

alxarch:
Integrating the previous version of this PR with the `datacatalog_updater` proved quite a difficult task to get right.
The major issue is that since each lambda handler has a limited amount of time to complete it's task, we need to continue any 'unfinished business' by invoking the same Lambda to continue from where we left of. What further complicates things is that to shorten the time required to complete a full database sync we run these 'recursive' Lambda tasks *in parallel* for each database/table pair required. This means that a single 'sync db' Lambda invocation results in multiple sub-invocations, each of which might invoke itself just before exiting if it didn't have enough time to complete it's job. We had to add strictly structured logging with a common `traceId` for all lambda requests that were performed as a result of a `sync db` event. We still don't have a way to keep track of the status of a sync 'job' without resorting to CloudWatch but at least we can have some easier filtering to trace all lambda output of a single sync run.

web/src/components/sidesheets/PolicyBulkUploadSidesheet/graphql/uploadPolicies.generated.ts
```diff
@@ -0,0 +1,105 @@
+/**
```
kostaspap, line 1: This shouldn't be part of the PR

internal/log_analysis/datacatalog_updater/process/sync_table.go
```diff
@@ -0,0 +1,96 @@
...
+		zap.Int("numCalls", event.NumCalls),
+	)
+	syncTask := event.SyncTablePartitions
+	syncTask.NumWorkers = runtime.NumCPU() + 1
+	// defer invoking continuation
+	defer func() {
+		if errors.Is(err, context.DeadlineExceeded) && nextToken != "" {
+			numCalls := event.NumCalls + 1
+			// protect against infinite recursion
+			if numCalls > maxNumCalls {
```
kostaspap, line 56: When can the recursion happen? or is it just as a safeguard...in case? 
Is it possible to hit this limit under normal operation and when?

internal/log_analysis/datacatalog_updater/process/sync_table.go
```diff
@@ -0,0 +1,96 @@
...
+					TraceID:             event.TraceID,
+				},
+			})
+		}
+	}()
+
+	// Reserve some time for continuing the task in a new lambda invocation
+	// If the timeout too short we resort to using context.Background to send the request outside of the
+	// lambda handler time slot.
+	if deadline, ok := ctx.Deadline(); ok {
```
kostaspap, line 81: When can it happen that there is no deadline set?

internal/log_analysis/datacatalog_updater/process/sync_table.go
```diff
@@ -0,0 +1,96 @@
...
+		const gracefulExitTimeout = time.Second
```
kostaspap, line 82: Are we meant to complete invoking a Lambda invocation within this 1 second? No reason to try to "squeeze" it in. You can make the deadline even 10,30,60 seconds before the Lambda timeout. We lose very little if we terminate earlier. We can lose a lot though if we don't terminate the operations in time.

internal/log_analysis/datacatalog_updater/process/sync_table.go
```diff
@@ -0,0 +1,96 @@
...
+// nolint: nakedret
+func HandleSyncTable(ctx context.Context, event *SyncTableEvent) (err error) {
+	nextToken := event.NextToken
+	log := lambdalogger.FromContext(ctx)
+	log = log.With(
+		zap.String("traceId", event.TraceID),
+		zap.Int("numCalls", event.NumCalls),
+	)
+	syncTask := event.SyncTablePartitions
+	syncTask.NumWorkers = runtime.NumCPU() + 1
```
kostaspap, line 50: The operations these workers are doing are mostly network I/O. I think you could even set it higher eg `2*runtime.NumCPU()`

cmd/opstools/opstools.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	logger, err := config.Build()
+	if err != nil {
+		log.Fatalf("failed to build logger: %s", err)
+	}
+	return logger.Sugar()
+}
+
+func NewHTTPClient(maxConnections int, timeout time.Duration) *http.Client {
+	transport := cleanhttp.DefaultPooledTransport()
```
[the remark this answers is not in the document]
alxarch: It's not a 'new'  `http` client. It returns a 'normal' go `http.Transport`. The only difference is that it initializes the transport to use a 'clean' connection pool for this specific client instead of the `http.DefaultTransport` so things such as timeouts and max connections can be controlled per-client.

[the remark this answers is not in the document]
alxarch: Otherwise we are forced to 'override' the settings of the `http.DefaultTransport` and we do not know if other parts of the code we use have expectations for default timeouts etc.

internal/log_analysis/awsglue/glue_timebin.go
```diff
@@ -120,3 +253,47 @@ func (tb GlueTableTimebin) PartitionHasData(client s3iface.S3API, t time.Time, t
 
 	return hasData, err
 }
+
+// PartitionTimeFromValues resolves the timebin from a table storage descriptor
+func PartitionTimeFromValues(values []*string) (tm time.Time, err error) {
```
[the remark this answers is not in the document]
alxarch: This is to be used specifically with glue partition values. Converting the args to `[]string` would add the `aws.StringSliceValues(p.values)` on each call.

internal/log_analysis/datacatalog_updater/process/sync_table.go
```diff
@@ -0,0 +1,96 @@
...
+					NumCalls:            numCalls,
+					TraceID:             event.TraceID,
+				},
+			})
+		}
+	}()
+
+	// Reserve some time for continuing the task in a new lambda invocation
+	// If the timeout too short we resort to using context.Background to send the request outside of the
+	// lambda handler time slot.
```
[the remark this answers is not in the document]
alxarch: If the Lambda invocation has sufficient timeout we 'reserve' some time by creating a shorter deadline. We invoke the sync code with this 'shorter' context and if the error is `DeadlineExceeded`, we defer the continuation of the task by re-invoking the Lambda. If for any reason the this invocation fails and starts to retry the call with backoff etc, using `context.Background` ensures that we try our best we can to continue the work. There is no 'real' guarantee the job will manage to continue with lambda invoke. There's always the change the invoke will fail and the retry will happen outside the context. Practically it is the exact same way we do in all our code implicitly by  using `Invoke` instead of  `InvokeWithContext` on any lambda->lambda invocation.

internal/log_analysis/datacatalog_updater/process/sync_table.go
alxarch: Well, we only update the nextToken if *all partitions* on each page returned from glue where succcessfully processed. If for some reason a problematic partition keeps failing (and it wasn't an one-of network glitch for example), this safeguards against infinite recursion. The main issue is that since we *have to not rely on time ranges* for the continuation, we are not in direct control of the task progress. 

IMO the recursive lambda calls is a very ugly hack to handle long running jobs that leads to all these 'clever' tricks that obfuscate code. When life gives you lemons you make a lemonade but it is not guaranteed it will taste good. We should dedicate some time to have both a proper job queue to continue long running tasks and some sort of progress tracking for such tasks. This should be done in a service-agnostic way so that any part of Panther that requires long running tasks, such as scheduled queries etc, can submit and track their progress.

kostaspap: It seems that if instead of invoking directly the Lambdas we would send a message to the SQS queue, we would be protected against: 
1. Infinite loops. We can make sure that we don't hit infinite loops by making use of SQS DQL. We retry an operation only X amount of times. 
2. Failing to invoke the "next Lambda" in a chain due to any issues. If we were using SQS, in case we would fail to send the SyncTablePartitions with an updated NextToken, the initial operation would be retried. This is something we cannot protect against.

So overall, I think using SQS is better approach than invoking Lambda directly, asynchronously. 

Agree that for long-running we could be using AWS Step functions (we actually do it for other long-running processes like Parquet conversion).

kostaspap: This might require lot of rework. I don't think it is so critical to do now (we can always use `gluesync` tool), but I think it's a design which would be a bit more robust, without requiring another bit refactoring.

cmd/opstools/gluesync/gluesync/main.go
```diff
@@ -19,263 +19,87 @@ package main
...
+	if optPrefix := *opts.Prefix; optPrefix != "" {
+		matchPrefix = awsglue.GetTableName(optPrefix)
 	}
 
-	// check the version of Panther deployed against what this as compiled against, they _must_ match!
-	tagResponse, err := s3Client.GetBucketTagging(&s3.GetBucketTaggingInput{Bucket: &dataBucket})
+	sess, err := session.NewSession(&aws.Config{
+		Region:     opts.Region,
+		MaxRetries: opts.MaxRetries,
+		HTTPClient: opstools.NewHTTPClient(*opts.MaxConnections, 0),
```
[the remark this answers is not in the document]
alxarch: It's a way to controll the actual number of concurrent requests to AWS with a single value. The `NumWorkers` flag sets the number of parallel updates *per-table*. To sync the whole database in parallel we spawn one task for each database/table pair. This could result in a few hundred goroutines to complete the task. This is OK since most of them will be blocked on network I/O. That network I/O is controlled by a single point, the pool of http connections the session client has. So having this configurable controls the max number of parallel requests to Glue the command will be able to do. Twisting this knob this could speed things up in deployments with many tables.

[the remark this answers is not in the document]
alxarch: The timeout `0` means no timeout. I wanted to add a flag for timeout if we start encountering such errors. I didn't want to peek some arbitrary duration without the ability to override. Would `1min` be ok as an overrideable default?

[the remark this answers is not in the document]
kostaspap: > That network I/O is controlled by a single point, the pool of http connections the session client has

Each of these parallel Lambda invocations will run in a different container so the HTTP Client connection pool size will not help with that, since each container will have its own connection pool.

cmd/opstools/opstools.go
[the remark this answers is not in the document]
kostaspap: Got it

web/src/components/sidesheets/PolicyBulkUploadSidesheet/graphql/uploadPolicies.generated.ts
alxarch: Fixed

internal/log_analysis/datacatalog_updater/process/sync_table.go
alxarch: I agree, this should be deferred to a separate task. SQS could provide a solution for this. Still no progress tracking though.

internal/log_analysis/datacatalog_updater/process/sync_table.go
alxarch: I would say in tests.

internal/log_analysis/datacatalog_updater/process/sync_table.go
alxarch: I will change this to 1 minute. I don't think we even need to shorten the timeout TBH. I think we're overly afraid of doing stuff outside the lambda handler window. The way internal stuff in go code works (ie connection pools, keep alives, deferred statements etc) means that some level of activity *does* happen outside the handler timeout. They (AWS) cannot protect against this and compensate on this 'loss' by charging 100 *milli*seconds of computation when using a language that can handle simple requests in *micro*seconds. They are already ripping us off :)

internal/log_analysis/datacatalog_updater/process/sync_table.go
alxarch: Yes, I think we could set this much higher. This setting is conservative. We can ditch `runtime.NumCPUs()` and experiment with some env var.

cmd/opstools/gluesync/gluesync/main.go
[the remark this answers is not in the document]
alxarch: This is not the code that the Lambdas are running, this is the CLI executable.

internal/log_analysis/datacatalog_updater/main/lambda.go
```diff
@@ -34,26 +32,26 @@ import (
...
 	}()
 
-	if event.Sync {
-		lambdaDeadline, _ := ctx.Deadline()
-		syncDuration := time.Since(lambdaDeadline) / 2 //  allocate a fraction of total time, this value will be negative!
-		syncDeadline := lambdaDeadline.Add(syncDuration)
-		return process.Sync(&event.SyncEvent, syncDeadline)
+	if event.SyncDatabaseEvent != nil {
+		err = process.HandleSyncDatabase(ctx, event.SyncDatabaseEvent)
+		return
```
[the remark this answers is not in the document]
alxarch: Added

cmd/opstools/gluesync/gluesync/main.go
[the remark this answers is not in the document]
kostaspap: > This is not the code that the Lambdas are running, this is the CLI executable.

Ah correct. 
Have you tried that scenario? How does the AWS SDK behave when it tries to acquire a new connection from the pool but none is available? Does it wait for a while with a timeout (and thus we run the risk of the program crashing/erroring in case of a deployment with a lot of tables) or does it wait "for ever"?

internal/log_analysis/datacatalog_updater/process/sync_table.go
kostaspap: You set the `maxNumCalls` to 1000. Can you share the thought process that you went through to pick this number? E.g. why is it `1000` and not `1000000` or `10`.

cmd/opstools/gluesync/gluesync/main.go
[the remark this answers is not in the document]
kostaspap: > The timeout `0` means no timeout. I wanted to add a flag for timeout if we start encountering such errors. I didn't want to peek some arbitrary duration without the ability to override. Would `1min` be ok as an overrideable default?

I would say just use the default value unless you have experienced any issues with it.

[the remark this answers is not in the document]
alxarch: Yes, zero _is the default value_ in `http.DefaultClient`. `cleanhttp` sets it to `30s` IIRC.

internal/log_analysis/awsglue/glue_timebin.go
```diff
@@ -20,6 +20,8 @@ package awsglue
 
 import (
 	"fmt"
+	"strconv"
```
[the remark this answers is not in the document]
alxarch: Added more unit tests for timebin methods. Resisted the urge to move `timebin` to it's own module.

internal/log_analysis/datacatalog_updater/process/sync_table.go
alxarch: During test runs, I saw that regardless the number of partitions involved or the `MaxResults` setting, the Glue API calls always splits the results in pages. This means that for tables with many partitions we will get multiple pages to process.
Each year we need 24x365=8760 partitions. If the behavior observed is correct, we would be getting ~100 pages per year of data. Each run would process 1-4 pages on each run if all partitions need to be updated. The 1000 is a number that is certain to only happen in problematic cases where a partition consistently does not update. Setting it to something lower could prevent a sync from completing, setting it to something higher could mean more $$ wasted on problematic cases. I didn't go into exploring to check the value against number of partitions already scanned as it seemed complex to get it right and the proper solution for this is to nor use deferred recursive lambda calls for continuation but either SQS or step functions.

alxarch: Another approach would be to fail immediately if we have a Deadline error and `sync.NextToken == event.NextToken`. We would then be relying only on session.Session retry mechanism to retry problematic partitions.

alxarch: The balance between numRetries/numParallelRequests/lambdaTimeout seems hard to get right in a deterministic way.

kostaspap: The `sessions.Session` will retry only in case we fail to invoke the new Lambda asynchronously - it will not fail if that Lambda fails during its execution. 

It is possible to define how many times a Lambda that is invoked asynchronously should be retried : https://aws.amazon.com/about-aws/whats-new/2019/11/aws-lambda-supports-max-retry-attempts-event-age-asynchronous-invocations/
and 
https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html#invocation-async-errors

kostaspap: > Another approach would be to fail immediately if we have a Deadline error and `sync.NextToken == event.NextToken`. We would then be relying only on session.Session retry mechanism to retry problematic partitions.

I kinda like that. 
Note that we can configure the Lambda to retry up to 2 times https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html#invocation-async-errors

rleighton:
Can you update the "Converstation" summary at the top with the current status? Starting in on the code now...

internal/log_analysis/datacatalog_updater/process/sync_table.go
```diff
@@ -0,0 +1,109 @@
...
+// HandleSyncTableEvent starts or continues a gluetasks.SyncTablePartitions task.
+// nolint: nakedret
+func HandleSyncTableEvent(ctx context.Context, event *SyncTableEvent) (err error) {
+	log := lambdalogger.FromContext(ctx)
+	log = log.With(
+		zap.String("traceId", event.TraceID),
+		zap.Int("numCalls", event.NumCalls),
+	)
+	sync := event.SyncTablePartitions
+	sync.NumWorkers = runtime.NumCPU() + 1
```
rleighton, line 59: Add comment as to why this setting

internal/log_analysis/datacatalog_updater/process/sync_table.go
```diff
@@ -0,0 +1,109 @@
...
+		// protect against infinite recursion
+		numCalls := event.NumCalls + 1
+		if numCalls > maxNumCalls {
+			log.Error("max calls exceeded",
+				zap.String("table", event.TableName),
+				zap.String("database", event.DatabaseName),
+				zap.String("token", sync.NextToken),
+				zap.Error(err),
+			)
+			return
```
rleighton, line 77: Should you be returning an error here rather than just logging? If not, add a comment as to why. Seems like this should not log and should `Errorf`

rleighton (commented):
This is really coming along very well. I can't wait to try it.

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -36,7 +36,8 @@ var (
 
 func SQS(event events.SQSEvent) error {
 	for _, record := range event.Records {
-		zap.L().Debug("processing record", zap.String("content", record.Body))
+		// uncomment to see all payloads
```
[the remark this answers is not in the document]
alxarch: This adds too much noise in logs during debug. On the default deployment it's almost impossible to track something. (Now that I have added the `traceId` field to zap, it might be possible to re-enable.

internal/log_analysis/datacatalog_updater/process/sync_table.go
alxarch: I want the `traceId` zap field in the logs. The lambda logger does not include that info.

alxarch: I can also wrap the error for the lambda logging to be somewhat useful

internal/log_analysis/datacatalog_updater/process/sync_database.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"go.uber.org/multierr"
+	"go.uber.org/zap"
+
+	"github.com/panther-labs/panther/internal/log_analysis/awsglue"
+	"github.com/panther-labs/panther/internal/log_analysis/gluetasks"
+)
+
+type SyncEvent struct {
+	TraceID       string
```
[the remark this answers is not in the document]
alxarch: fixed

internal/log_analysis/datacatalog_updater/process/sync_database.go
```diff
@@ -0,0 +1,75 @@
...
+
+func HandleSyncDatabase(ctx context.Context, event *SyncEvent) error {
+	log := zap.L()
+	log = log.With(
+		zap.String("traceId", event.TraceID),
+		zap.Bool("dryRun", event.DryRun),
+	)
+	tableEvents := []*SyncTableEvent{}
+	for _, dbName := range event.DatabaseNames {
+		afterTableCreateTime := dbName != awsglue.LogProcessingDatabaseName
```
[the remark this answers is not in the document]
alxarch: added

internal/log_analysis/datacatalog_updater/process/sync_table.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/panther-labs/panther/pkg/lambdalogger"
+)
+
+type SyncTableEvent struct {
+	TraceID  string
+	NumCalls int
+	gluetasks.SyncTablePartitions
+}
+
+const maxNumCalls = 1000
```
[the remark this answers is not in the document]
alxarch: added

internal/log_analysis/datacatalog_updater/process/sync_table.go
alxarch: Changed to `SyncWorkersPerTable` env var with default `10`

cmd/opstools/gluerecover/gluerecover/main.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/panther-labs/panther/cmd/opstools"
+	"github.com/panther-labs/panther/internal/log_analysis/awsglue"
+	"github.com/panther-labs/panther/internal/log_analysis/gluetasks"
+)
+
+var opts = struct {
```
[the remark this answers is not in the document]
alxarch: added

cmd/opstools/gluerecover/gluerecover/main.go
```diff
@@ -0,0 +1,142 @@
...
+	Debug:          flag.Bool("debug", false, "Enable additional logging"),
+	Region:         flag.String("region", "", "Set the AWS region to run on"),
+	MaxRetries:     flag.Int("max-retries", 12, "Max retries for AWS requests"),
+	MaxConnections: flag.Int("max-connections", 100, "Max number of connections to AWS"),
+	NumWorkers:     flag.Int("workers", 8, "Number of parallel workers for each table"),
+	Prefix:         flag.String("prefix", "", "A prefix to filter log type names"),
+}
+
+func main() {
+	flag.Parse()
```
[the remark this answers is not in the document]
alxarch: added

cmd/opstools/gluesync/gluesync/main.go
```diff
@@ -19,263 +19,87 @@ package main
...
-	"github.com/panther-labs/panther/internal/log_analysis/gluetables"
-	"github.com/panther-labs/panther/pkg/awscfn"
-	"github.com/panther-labs/panther/pkg/genericapi"
-	"github.com/panther-labs/panther/pkg/prompt"
-	"github.com/panther-labs/panther/tools/cfnstacks"
+	"github.com/panther-labs/panther/internal/log_analysis/gluetasks"
 )
 
-const (
-	banner = "updates glue table and partition schemas"
```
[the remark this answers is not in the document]
alxarch: added banner

internal/log_analysis/datacatalog_updater/process/sync_table.go
alxarch: Changed to use env var

internal/log_analysis/datacatalog_updater/process/sync_table.go
alxarch: Added comment and return better error

internal/log_analysis/datacatalog_updater/process/sync_table.go
kostaspap: Fair! I was mostly worried of running out of time for the Network I/O stuff. 
What I wanted to mention, is that when a lambda expires, the following things can happen: 
1. The container "freezes". AWS de-allocates any CPU, memory. The container can be revived when a new event triggers the lambda. Any leftover routines will continue from where they left (running previous invocation!). 
2. The container freezes and is discarded. 

Note that if an invocation times out, this is considered a failure. So if we have retries enabled, the code will re-run. 

Maximum lifetime of a container is 8hrs (AWS recycles their whole Lambda fleet every 8hrs)

internal/log_analysis/datacatalog_updater/process/sync_table.go
alxarch: I changed the continuation logic to explicitly track the number of consecutive retries without progress. Using this info we can control exactly how many times we allow this case with a const (currently 5)

kostaspap (approved):

rleighton (approved):
Can't wait to play with it. Nice work!

PR #1477, opened 2020-09-06
Begin 1.8 release testing

austinbyers:
## Background

Updating the release branch with what is currently in `master`. I tried doing `git checkout --theirs` to resolve conflicts after the `git pull`, but kept running into issues. So I literally just copied the entire source tree from a fresh clone of `master` after the `git pull`

I'm still not sure the best way to manage the `release` branches across both of our repos, open to suggestions. For example, now the `master` branch does not have the expected `git describe --tags` output because the tags are only on the release branch. And git shows the `release` branch as really far behind:

Screen Shot 2020-09-06 at 4 05 53 PM

I'm going to try doing a standard merge (not squash) when updating branches, like we used to do for enterprise syncs. That should help fix the tagging and make things more clear. I'll just need to make sure @panther-bot doesn't freak out when we eventually merge the release branch into master without squashing.

## Changes

- From the release branch, `git pull origin master`, then just copy all master source files

## Testing

- CI


austinbyers:
I temporarily overrode the repo settings so the release branch can be ready on Monday for the European team to test

PR #1478, opened 2020-09-07
hotfix: Fix Label link for source

alexmylonas:
## Background

The link existing on the label of `LogSourceCards` it's wrongly hardcoded as SQS, this PR fixes that by generating the correct link based on source type.

## Changes

- Added and used `sourceType` for each card
## Testing

- Locally
- Deployed env


3nvi (approved):
LGTM 🎉

PR #1479, opened 2020-09-07
Re-fetch alerts on every rule page visit

3nvi:
## Background

When re-visiting a rule  that you have visited  in the past, the page won't attempt to re-fetch the previously fetched alerts. This might cause inconsistencies, since more  alerts might have fired since then.

The solution up until now, was to refresh the page. This fix makes that not required

## Changes

- Add `cache-and-network` to alert-related  queries

## Testing

- Locally


alexmylonas (approved):
LGTM 👍

PR #1492, opened 2020-09-08
fix: return a success response for sqs and sns

### LINKED ISSUE #1486: SNS /SQS Test Destination Says Failed when it succeeded
### Describe the bug

I executes an SNS test for the SNS destination, which worked but I got this:
image

Also, if a test fails the error should be displayed or the user cannot troubleshoot.

Same thing happened for SQS. It actually worked but the UI said fail with no error message.

Slack did not have this issue.

### Steps to reproduce

1. Create SNS topic and create an email subscription to yourself
2. Create the destination
3. Execute the test
4. Get failed message.

### Expected behavior

A success status in the UI.

s0l0ist:
## Background

Closes #1486 

## Changes

- Returns a response status for SQS and SNS destinations
- Return more descriptive error payloads

## Testing

- mage test:ci

### DIFF
--- a/internal/core/alert_delivery/outputs/sns.go
+++ b/internal/core/alert_delivery/outputs/sns.go
@@ -45,7 +45,12 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
 	if err != nil {
 		errorMsg := "Failed to serialize default message"
 		zap.L().Error(errorMsg, zap.Error(errors.WithStack(err)))
-		return &AlertDeliveryResponse{Message: errorMsg, Permanent: true}
+		return &AlertDeliveryResponse{
+			StatusCode: 500,
+			Message:    errorMsg,
+			Permanent:  true,
+			Success:    false,
+		}
 	}
 
 	outputMessage := &snsMessage{
@@ -57,7 +62,12 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
 	if err != nil {
 		errorMsg := "Failed to serialize message"
 		zap.L().Error(errorMsg, zap.Error(errors.WithStack(err)))
-		return &AlertDeliveryResponse{Message: errorMsg, Permanent: true}
+		return &AlertDeliveryResponse{
+			StatusCode: 500,
+			Message:    errorMsg,
+			Permanent:  true,
+			Success:    false,
+		}
 	}
 
 	snsMessageInput := &sns.PublishInput{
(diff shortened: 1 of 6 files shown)


internal/core/alert_delivery/outputs/sns.go
```diff
@@ -72,16 +82,50 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
...
-	_, err = snsClient.Publish(snsMessageInput)
+	response, err := snsClient.Publish(snsMessageInput)
 	if err != nil {
 		errorMsg := "Failed to send message to SNS topic"
 		zap.L().Error(errorMsg, zap.Error(errors.WithStack(err)))
-		return &AlertDeliveryResponse{Message: errorMsg}
+		return &AlertDeliveryResponse{
+			StatusCode: 500,
+			Message:    errorMsg,
+			Permanent:  false,
```
rleighton, line 100: What does Permanent mean?

rleighton: Maybe add comments as to why you are setting the value to the specific value.

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -72,16 +82,50 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
...
 	}
 
-	_, err = snsClient.Publish(snsMessageInput)
+	response, err := snsClient.Publish(snsMessageInput)
 	if err != nil {
 		errorMsg := "Failed to send message to SNS topic"
 		zap.L().Error(errorMsg, zap.Error(errors.WithStack(err)))
-		return &AlertDeliveryResponse{Message: errorMsg}
+		return &AlertDeliveryResponse{
+			StatusCode: 500,
```
rleighton, line 98: We should look at checking for aws errors and then picking the status code, in particular permission errors should be differentiated.

internal/core/alert_delivery/outputs/sns.go
s0l0ist: The parameter meaning is inside the struct which you cannot see here, but there are many other places where this value is set and we're just cluttering the code a bit if we write a comment. 

```go
// AlertDeliveryResponse holds the response (success or failure) of an alert delivery request.
type AlertDeliveryResponse struct {
	// StatusCode is the http response status code
	StatusCode int

	// Message is our wrapped description of the problem: what went wrong.
	Message string

	// Permanent indicates whether the alert output should be retried.
	// For example, outputs which don't exist or errors creating the request are permanent failures.
	// But any error talking to the output itself can be retried by the Lambda function later.
	Permanent bool

	// Success is true if we determine the request executed successfully. False otherwise.
	Success bool
}
```

rleighton (approved):
Next step, getting errors to the UI!

PR #1494, opened 2020-09-08
fix: sns and sqs delivery statuses

s0l0ist:
## Background

Sync changes #1492 merged into `release` back into `master`.

## Changes

- Return a success delivery response from SNS/SQS payloads

## Testing

- mage test:ci


austinbyers (approved):
We will sync all release changes into master after we publish, don't feel like you have to do it for every single change individually (unless you're working on a feature in master that needs it)

PR #1497, opened 2020-09-09
Frontend changes for Delivery Statuses

s0l0ist:
## Background

This PR contains the backend AppSync configuration for the frontend to consume. It will be the main PR for the frontend related work #1392

Also fixes issue #1513 

## Changes

- Added resolver for deliverAlert (re-sending alerts)
- Added graphql attributes/fields/types to support frontend invocation

## Testing

- mage test:ci


api/graphql/schema.graphql
```diff
@@ -25,6 +25,7 @@ type Mutation {
   deleteUser(id: ID!): Boolean
   inviteUser(input: InviteUserInput): User!
   remediateResource(input: RemediateResourceInput!): Boolean
+  deliverAlert(input: DeliverAlertInput!): AlertSummary
```
3nvi, line 28: @s0l0ist can this response ever be `null`? Why so?

api/graphql/schema.graphql
```diff
@@ -182,16 +186,38 @@ type AlertSummary implements Alert {
...
 input UpdateAlertStatusInput {
   alertId: ID!
   status: AlertStatusesEnum!
 }
 
+input SendTestAlertInput {
+  outputIds: [ID!]!
+}
+
+type SendTestAlertResponse {
```
3nvi, line 216: @s0l0ist same here, would `success` or `message` not be defined?

api/graphql/schema.graphql
s0l0ist: We always return an empty array. Slices should never be null, only empty.

api/graphql/schema.graphql
s0l0ist: The shape has changed here and has been resolved.

api/graphql/schema.graphql
s0l0ist: Fixed

3nvi (approved):

3nvi:
@s0l0ist assuming you verified the functionality, so I'll be merging this

PR #1502, opened 2020-09-09
fix: add dynamic width  to profile popup

3nvi:
## Background

When users have a big email, the  profile settings popup cuts the  email out.

This PR allows the profile popup to have a dynamic width based on its content

## Changes

- Change `width` to `maxWidth`

## Testing

- Visually


alexmylonas (approved):
Good catch!

vorillaz (approved):

PR #1480, opened 2020-09-07
Revamp the S3 Log Onboarding UI

3nvi:
## Description

Similar to the Compliance Source updates, this PR changes the way an S3 Source can be onboarded. It follows the **exact** same principles & structure as the Compliance updates and it mainly updates visual aspects.

Relates to the work described in https://github.com/panther-labs/panther-enterprise/issues/643 

A final PR, doing the same thing for SQS will finally wrap this task up.

### Designs

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/6c4082e9-dc0b-4e26-b60e-cc96e60306ee/files/c2bbc1a8-4778-4cd6-93a5-4630d8868201/pages/2E5A63C6-5327-4C4D-8EA4-84899A50282C

## Screenshots

Screen Shot 2020-09-07 at 8 38 04 PM
Screen Shot 2020-09-07 at 8 38 24 PM
Screen Shot 2020-09-07 at 8 38 34 PM

### Checklist:

- [x] Make sure you give an high-level overview of the problem & the solution
- [x] Make sure the related issue is referenced
- [x] Make sure screenshots are added if the UI is changed in any way
- [x] Make sure all of the significant logic is covered by tests
- [x] Make sure your changes are accessible (addition of aria-attributes & correct HTML semantics)
- [x] Make sure you reference the related design document if it exists
- [x] Make sure the changes are responsive (appearing decently on devices)
- [x] ~~Make sure telemetry is added (analytics & errors)~~

PR #1441, opened 2020-08-31
Enhance Compliance Workflow

3nvi:
## Description

This PR enhances the way a Cloud Security Source can be onboarded. It uses the newly redesigned screens and adds a better - more automatic - flow to the way the health check and the source gets persisted in Panther

## Screenshots

Screen Shot 2020-08-31 at 5 38 19 PM
Screen Shot 2020-08-31 at 5 36 55 PM
Screen Shot 2020-08-31 at 5 37 34 PM
Screen Shot 2020-08-31 at 5 37 02 PM
Screen Shot 2020-08-31 at 5 37 34 PM
Screen Shot 2020-08-31 at 5 38 32 PM

Closes https://github.com/panther-labs/panther-enterprise/issues/573

## Designs

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/6c4082e9-dc0b-4e26-b60e-cc96e60306ee/files/323d0ca1-81a8-40e7-83ee-0a05b870bb4a/pages/62558DB2-0EAC-4C4C-9920-74571F38FF66

### Checklist:

- [x] Make sure you give an high-level overview of the problem & the solution
- [x] Make sure the related issue is referenced
- [x] Make sure screenshots are added if the UI is changed in any way
- [x] Make sure all of the significant logic is covered by tests
- [x] Make sure your changes are accessible (addition of aria-attributes & correct HTML semantics)
- [x] Make sure you reference the related design document if it exists
- [x] Make sure the changes are responsive (appearing decently on devices)
- [x] ~~Make sure telemetry is added (analytics & errors)~~

PR #1504, opened 2020-09-09
Aggelos enhance sqs onboading

3nvi:
## Background

This PR mimics the PRs #1480 and #1441 by enhancing & updating the flow for onboarding a logs through an SQS queue. Now the SQS queue follows a more clear flow with the  addition of being able to copy the URL of the created SQS queue.

## Screenshots

Screen Shot 2020-09-09 at 7 13 59 PM
Screen Shot 2020-09-09 at 7 14 09 PM
Screen Shot 2020-09-09 at 7 14 34 PM

## Designs

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/39dcb3e2-1e72-49b2-afab-3a59c8459f66/files/c2bbc1a8-4778-4cd6-93a5-4630d8868201/pages/2E5A63C6-5327-4C4D-8EA4-84899A50282C

## Changes

- Bump `Formik`
- Update SQS Wizard
- Add & Update tests for SQS creation & update

## Testing

- `mage test:ci`


web/src/components/wizards/ComplianceSourceWizard/ValidationPanel/ValidationPanel.tsx
```diff
@@ -68,12 +68,17 @@ const ValidationPanel: React.FC = () => {
...
-                <Link as={AbstractButton} variant="discreet" onClick={reset}>
+                <Link
+                  as={AbstractButton}
+                  variant="discreet"
+                  onClick={() => {
+                    resetForm();
+                    resetWizard();
+                  }}
+                >
                   Add Another
```
alexmylonas, line 82: Hmm maybe minor but isn't those two CTAs confusing? 
Users see a button  "Finish Setup" and "Add Another" button when creating. 

A more valid first button text would be "Go back to Sources" or smt like that.

web/src/components/wizards/S3LogSourceWizard/ValidationPanel/ValidationPanel.tsx
```diff
@@ -87,13 +87,18 @@ const ValidationPanel: React.FC = () => {
...
-                <Link as={AbstractButton} variant="discreet" onClick={reset}>
+                <Link
+                  as={AbstractButton}
+                  variant="discreet"
+                  onClick={() => {
+                    resetForm();
+                    resetWizard();
+                  }}
+                >
                   Add Another
```
alexmylonas, line 102: Same as above

alexmylonas (approved):
Solid work! Only a minor copy issue but doesn't stop me from approving/

web/src/components/wizards/ComplianceSourceWizard/ValidationPanel/ValidationPanel.tsx
3nvi: Hmm... That's what the designs have :/

I think the goal is to let  the user know htat the flow is finished. We engineers know that, but a user (analyst) may not underrstand that all this is done, unless you click "finish".

I understand your concern, but that's something for business to decide, not us  (especially since we have it in multiple places like that)

3nvi: Do you think we should remove "add another"? Perhaps that might help (since rarely will a user want to add multiple SQS)

kostaspap:
Can we also display the SQS queue URL instead of just giving them the option to copy it? I think it's not a bad thing to overcommunicate  this information to the customers :)

web/src/components/wizards/ComplianceSourceWizard/ValidationPanel/ValidationPanel.tsx
alexmylonas: I think my main concern is with the "finish" word as its misleading.  

Come to think about it maybe the "add another" helps implying that the source is actually created 😛

alexmylonas: I think its important but not that big of a deal and we can tackle it down the line, your call!

alexmylonas:
> Can we also display the SQS queue URL instead of just giving them the option to copy it? I think it's not a bad thing to overcommunicate this information to the customers :)

I have gotten this comment a lot by multiple engineers and I am really against displaying a long string on UI outright. 

I would suggest we create a re-usable component that has 2 actions - reveal or copy the URL

@3nvi thoughts?

3nvi:
> > Can we also display the SQS queue URL instead of just giving them the option to copy it? I think it's not a bad thing to overcommunicate this information to the customers :)
> 
> I have gotten this comment a lot by multiple engineers and I am really against displaying a long string on UI outright.
> 
> I would suggest we create a re-usable component that has 2 actions - reveal or copy the URL
> 
> @3nvi thoughts?

Reveal is for stuff that **should** be hidden by default (namely sensitive data). This would make sense if this was an API key, but not if it's a public SQS queue.

I agree with you on the big UI string. @kostaspap we actually tried displaying it but it looked awful because it's a huge string.  Currently you can copy if, but as soon as you click "finish setup" you get redirected to the  sources  page, where you can see it in front of you as part of the related source card.

Closing for now, since this  PR also affects https://github.com/panther-labs/panther/issues/1508. If we feel that it's important to actually **do show** it as part of the wizard, let's create a ticket to discuss further

PR #1505, opened 2020-09-09
Small ops improvements

austinbyers:
## Background

A few small bugfixes based on observed problems in existing deployments.

## Changes

- Bump snapshot-poller memory from 128 to 512 MB. One of our deployments is running out of memory, and this is a critical piece of infra that needs more cpu/network at all times anyway (just like the log processor). The extra monetary cost is typically negligible compared to log processing.
- Add a comment about why we lowercase emails in Cognito (@alxarch )
- Allow `mage master:deploy` with an optional `STACK` env variable - this way, you can deploy the pre-packaged version on top of any stack name. We already support `STACK=name mage teardown`, so this is a natural extension.

## Testing

- CI; will deploy to release env after merged internally

### DIFF
--- a/internal/core/custom_resources/resources/alarms_api_gateway.go
+++ b/internal/core/custom_resources/resources/alarms_api_gateway.go
@@ -40,8 +40,10 @@ type APIGatewayAlarmProperties struct {
 }
 
 func customAPIGatewayAlarms(_ context.Context, event cfn.Event) (string, map[string]interface{}, error) {
-	const defaultGatewayLatencyThresholdMs = 1000
-	const defaultGatewayErrorThreshold = 5
+	const (
+		defaultErrorThreshold = 1
+		defaultLatencyThresholdMs = 1000
+	)
 
 	switch event.RequestType {
 	case cfn.RequestCreate, cfn.RequestUpdate:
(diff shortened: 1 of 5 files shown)


internal/core/custom_resources/resources/alarms_api_gateway.go
```diff
@@ -35,20 +35,28 @@ const (
...
 		if err := parseProperties(event.ResourceProperties, &props); err != nil {
 			return "", nil, err
 		}
 
+		if props.ErrorThreshold == nil {
+			props.ErrorThreshold = aws.Int64(defaultErrorThreshold)
+		}
 		if props.LatencyThresholdMs == 0 {
-			props.LatencyThresholdMs = 1000
+			props.LatencyThresholdMs = defaultLatencyThresholdMs
```
kostaspap, line 59: I think instead of increasing the threshold, might be better to increase the number of datapoints ("periods"). 
AWS can have small "spikes" of errors due to some network hiccup that get resolved automatically. We really need to intervene if the errors continue. We could create an alarm if errors keep occuring for e.g. 3 periods.

kostaspap (approved):

internal/core/custom_resources/resources/alarms_api_gateway.go
rleighton: Agree, I think this is more robust.

internal/core/custom_resources/resources/alarms_lambda.go
```diff
@@ -60,13 +60,13 @@ func customLambdaAlarms(_ context.Context, event cfn.Event) (string, map[string]
...
 		if props.LoggedErrorThreshold == nil {
-			props.LoggedErrorThreshold = aws.Int(0)
+			props.LoggedErrorThreshold = aws.Int(1)
 		}
 		if props.LoggedWarnThreshold == nil {
 			props.LoggedWarnThreshold = aws.Int(25)
 		}
 		if props.ExecutionErrorThreshold == nil {
-			props.ExecutionErrorThreshold = aws.Int(0)
+			props.ExecutionErrorThreshold = aws.Int(1)
```
rleighton, line 69: Will this require 2+ errors to set off a lambda? I don't think this should be the default. Specific lambda's. can override.

rleighton (commented):
I really do not like changing the default for a lambda alarm (if I read this correctly). We should discuss this. We could miss important errors.

internal/core/custom_resources/resources/alarms_lambda.go
austinbyers: Ok, will revert to 0 for the error threshold

PR #1506, opened 2020-09-10
Speed up web testing

### LINKED ISSUE #1490: Investigate slow web testing

3nvi:
## Background

Currently web apps are slow. Especially in a docker container, the tests take an average of 3.5 minutes (at the time of writing) and are by far the slowest check in our CI.

This prompted me to investigate why can this happen and specifically why some are far and others are slow. I isolated the slow tests and tried to see what is causing them to be slower than others. 

Initially I thought it might be `pounce`,  but recently  we made pounce way more lightweight,  so it didn't make any  sense. Then I thought  it might be the animations related to modals & snackbars, but even when disabling them the  tests were still slow. Turns out, that the package  we use for our dialogs, shows a backdrop behind our Modal & Sidesheet which adds a "lock" into the tabbing and scrolling behavior when a dialog is active.  Apparently, this was slow. REALLY slow.

The fix was to disable this "locking" behavior during testing (which so far hasn't affected our tests in any way)

See the before and after tests for a  particular module (local environment):

Screen Shot 2020-09-10 at 1 49 04 PM
Screen Shot 2020-09-10 at 1 49 21 PM

Overall, it brought down the time by **45%** in my local environment:

Screen Shot 2020-09-10 at 1 56 07 PM
Screen Shot 2020-09-10 at 1 53 18 PM

I also investigated an unreasonable timing that was displayed in the terminal during the  `Wizard` tests.

Screen Shot 2020-09-10 at 1 52 15 PM

If you add together the timings of the individual tests above, they are well below the 7 seconds displayed in the screenshot above. 

Turns out that `Wizard` tests are **always** the first tests to run. During the initialization phase, `jest` (our test runner) needs to parse Javascript ES6 to ES5 and also convert TS to JS. **The time it takes for this to happen, always gets added to the first test suite.**

I tried reducing the compilation time, but I wasn't able to achieve significant savings. We just have to swallow the fact that the the testing suite setup  will take between 7 -13 seconds (again, in my local environment).

Closes #1490 
  
## Changes

- Mock `@reach-ui/dialog` overlay component by forcing it do disable tab & scroll locking
- Bump resource class of `mage test:web` to `xlarge`, matching the same size assigned to `mage test:go`
- Bump resource class of `bundlesize` to `large` since it has to build the project which is an intensive process. I saw that it was getting similar timings to `mage test:go`, sometimes even being slower than it. Decided to bump it 1 resource class and not it's stable around ~1m

## Testing

Screen Shot 2020-09-10 at 5 49 29 PM


vorillaz (approved):
Amazing work @3nvi, I adore the digging and debugging talent of yours.
Do you think we can also tryout Jest caching?

3nvi:
> Amazing work @3nvi, I adore the digging and debugging talent of yours.
> Do you think we can also tryout Jest caching?

Thanks! 

I think `jest` is using caching by default. According to the docs:

> --cache
Whether to use the cache. Defaults to true. Disable the cache using --no-cache. Note: the cache should only be disabled if you are experiencing caching related problems. On average, disabling the cache makes Jest at least two times slower.

@vorillaz do you mean something different?

.circleci/config.yml
```diff
@@ -157,7 +157,7 @@ jobs:
 
   test_web:
     executor: panther-buildpack
-    resource_class: large
+    resource_class: xlarge
```
austinbyers, line 160: Do web tests run in parallel? Does the larger image actually help the tests finish faster?

If your tests are not running in parallel, the extra cores from a larger instance won't help

.circleci/config.yml
```diff
@@ -178,6 +178,7 @@ jobs:
 
   bundlesize:
     executor: panther-buildpack
+    resource_class: large
```
austinbyers, line 181: Again, extra cores only help if your tests can take advantage of them - did this significantly improve the test time?

austinbyers (commented):
Thanks for looking into this!

.circleci/config.yml
3nvi: I actually ran the tests in both classes. It's not just cores, it's RAM as well. Here are the results:

Screen Shot 2020-09-10 at 3 05 35 PM
Screen Shot 2020-09-10 at 3 12 54 PM

**TLDR; It's about 20 seconds faster**

.circleci/config.yml
3nvi: This is interesting. Look at the SS above. Both oof these SS are with the default `resource_class` but one finishes in `1:11` and the  other at 1:44`

When adding  a bigger resource class it finished  in  `1:01`. About 10seconds time. There is a big variation here, but because of that, I'd personally play it safe and have a higher resource  class.

If we don't want that I'm ok to revert

Screen Shot 2020-09-10 at 5 49 29 PM

.circleci/config.yml
3nvi: Now because I ran the tests just once, I can't be 100% sure that it was not random. We can revert it back too the original value and monitor it if you want

austinbyers (approved):
I'm fine with the higher resource class since it seems to help (it also tends to make the image setup faster)

PR #1512, opened 2020-09-10
feat: add Formik date range component

vorillaz:
## Background

This is the initial `DateRange` component working along with Formik.

## Changes

- Introduce a new Formik component, which parses and propagates dates in UTC format.

### DIFF
--- a/web/src/components/fields/DateRangeInput/DateRangeInput.tsx
+++ b/web/src/components/fields/DateRangeInput/DateRangeInput.tsx
@@ -1,0 +1,70 @@
+/**
+ * Panther is a Cloud-Native SIEM for the Modern Security Team.
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import { Box, FormError, DateRangeInput, DateRangeInputProps } from 'pouncejs';
+import { formatTime } from 'Helpers/utils';
+import { useField } from 'formik';
+
+export interface FieldDateRangeInputProps {
+  nameStart: string;
+  nameEnd: string;
+}
+
+const dateFormatter = formatTime('YYYY-MM-DDTHH:mm:ss[Z]');
+
+const FormikTextInput: React.FC<Omit<DateRangeInputProps, 'name'> & FieldDateRangeInputProps> = ({
+  nameStart,
+  nameEnd,
+  onChange,
+  ...rest
+}) => {
+  const [, metaStart, helpersStart] = useField(nameStart);
+  const [, metaEnd, helpersEnd] = useField(nameEnd);
+
+  const { touched: touchedStart, error: errorStart, value: valueStart } = metaStart;
+  const { setValue: setValueStart } = helpersStart;
+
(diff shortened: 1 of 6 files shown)


web/src/components/fields/DateRangeInput/DateRangeInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const dateFormatter = formatTime('YYYY-MM-DDTHH:mm:ss[Z]');
+
+const FormikTextInput: React.FC<DateRangeInputProps & FieldDateRangeInputProps> = ({
+  nameStart,
+  nameEnd,
+  onChange,
+  ...rest
+}) => {
+  const [, metaStart, helpersStart] = useField(nameStart);
```
3nvi, line 37: can we try and  see if using the `useFastField`  works here? It would save us a lot of un-necessary re-renders on a relatively heavy DateRangeInput component

web/src/components/fields/DateRangeInput/DateRangeInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { useField } from 'formik';
+
+export interface FieldDateRangeInputProps {
+  nameStart: string;
+  nameEnd: string;
+}
+
+const dateFormatter = formatTime('YYYY-MM-DDTHH:mm:ss[Z]');
+
+const FormikTextInput: React.FC<DateRangeInputProps & FieldDateRangeInputProps> = ({
```
3nvi, line 31: can we  Omit `name` from the props?  We want to make sure the  users  never actually use it :)

web/src/components/fields/DateRangeInput/DateRangeInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+      setValueStart(dateFormatter(end));
+    },
+    [setValueStart, setValueEnd]
+  );
+  return (
+    <Box>
+      <DateRangeInput
+        invalid={isInvalid}
+        {...rest}
+        value={[valueStart, valueEnd]}
```
3nvi, line 62: can we use `React.useMemo`  for this array? If we implement `React.memo` under  the hood of `DateRangeInput` (which I think we  should), it would save some un-necessary re-renders.

web/src/components/fields/DateRangeInput/DateRangeInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const [, metaStart, helpersStart] = useField(nameStart);
+  const [, metaEnd, helpersEnd] = useField(nameEnd);
+
+  const { touched: touchedStart, error: errorStart, value: valueStart } = metaStart;
+  const { setValue: setValueStart } = helpersStart;
+
+  const { touched: touchedEnd, error: errorEnd, value: valueEnd } = metaEnd;
+  const { setValue: setValueEnd } = helpersEnd;
+
+  const isInvalid = (touchedStart || touchedEnd) && !!errorStart && !!errorEnd;
```
3nvi, line 46: shouldn't this be 
```suggestion
  const isInvalid = (touchedStart || touchedEnd) && (!!errorStart || !!errorEnd);
```

since we want at  least 1 error?

3nvi (commented):
Hey @vorillaz.

Changed look good!. If possible, please delete all the `.icloud` files that were  added for some reason.  Feel free to modify `.gitignore` so we don't add them again in the future.

web/src/components/fields/DateRangeInput/DateRangeInput.tsx
vorillaz: Nice catch!

web/src/components/fields/DateRangeInput/DateRangeInput.tsx
vorillaz: Done!

web/src/components/fields/DateRangeInput/DateRangeInput.tsx
vorillaz: Done as well.

web/src/components/fields/DateRangeInput/DateRangeInput.tsx
3nvi: @vorillaz did it fail?

3nvi (approved):
Looks good! One comment about `useFastField`, but other than that g2g.

I'm pre-approving since my comment shouldn't block the PR

PR #1514, opened 2020-09-10
refactor: change "closed" status to "invalid"

### LINKED ISSUE #1410: Distinguishing between 'Closed' vs 'Resolved' alert statuses
### Description

Of the 4 available alert statuses (OPEN, TRIAGED, CLOSED, RESOLVED), the two closure statuses may generate confusion. i.e. A user may not know when to choose CLOSED or RESOLVED because there's no indication of the implied differences.

Suggested solution is to simply change the wording of `CLOSED` to `INVALID` to clearly indicate the difference between `INVALID` and `RESOLVED` issues.

### Acceptance Criteria

- Change `CLOSED` -> `INVALID`

s0l0ist:
## Background

closes #1410 

These changes do not perform a migration on existing alerts. Instead, they account for the deprecated status value and correct it on fetch or query.

## Changes

- updates both frontend and backend renaming of the enum
- ensures all APIs that return alerts use the new status if the old one was present

## Testing

- mage test:ci
- visual inspection and filtering of the list alerts page

### DIFF
--- a/api/lambda/alerts/models/api.go
+++ b/api/lambda/alerts/models/api.go
@@ -84,7 +84,7 @@ type ListAlertsInput struct {
 	// Filtering
 	Severity        []*string  `json:"severity" validate:"omitempty,dive,oneof=INFO LOW MEDIUM HIGH CRITICAL"`
 	NameContains    *string    `json:"nameContains"`
-	Status          []string   `json:"status" validate:"omitempty,dive,oneof=OPEN TRIAGED CLOSED RESOLVED"`
+	Status          []string   `json:"status" validate:"omitempty,dive,oneof=OPEN TRIAGED INVALID RESOLVED"`
 	CreatedAtBefore *time.Time `json:"createdAtBefore"`
 	CreatedAtAfter  *time.Time `json:"createdAtAfter"`
 	RuleIDContains  *string    `json:"ruleIdContains"`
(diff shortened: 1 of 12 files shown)


api/lambda/alerts/models/api.go
```diff
@@ -162,9 +162,12 @@ const (
 	// Triaged sets the alert to actively investigating
 	TriagedStatus = "TRIAGED"
 
-	// Closed is set for false positive or anything other reason than resolved
+	// [deprecated] Closed is set for false positive or anything other reason than resolved
 	ClosedStatus = "CLOSED"
```
kostaspap, line 166: Can we remove this? Is the plan to eventually write a migration script that updates customer tables?

kostaspap (approved):
1. Can't we just change the wording on the FE piece? I think it might be simpler than this one? 
2. If we were to make the change in the BE, maybe create a migration script? Should be fairly easy - scan the table and perform in-place update of the attribute. If there is no time for it now, we should create a Github issue to fix this eventually (might be a good task for a new hire). I am afraid we might forget about this and similar issues. 
3. Maybe "FALSE ALARM", "NO ISSUE", ... Not 100% sure of what's the best terminology here (naming is hard).

3nvi (approved):
LGTM 🎉

s0l0ist:
> 1. Can't we just change the wording on the FE piece? I think it might be simpler than this one?
> 2. If we were to make the change in the BE, maybe create a migration script? Should be fairly easy - scan the table and perform in-place update of the attribute. If there is no time for it now, we should create a Github issue to fix this eventually (might be a good task for a new hire). I am afraid we might forget about this and similar issues.
> 3. Maybe "FALSE ALARM", "NO ISSUE", ... Not 100% sure of what's the best terminology here (naming is hard).


I'll investigate using FE changes only.

api/lambda/alerts/models/api.go
s0l0ist: If we do end up going with BE changes and a migration script, we would remove this. I keep it to use it for checking against the filtering/query rules.

austinbyers (commented):
Agreed with just changing it in the frontend for simplicity if possible

s0l0ist:
Closing in favor of #1528

PR #1516, opened 2020-09-10
add name to ec2 resources

### LINKED ISSUE #1422: Add `Name` tag to AWS EC2 resources

Panther tries to indicate that resources created as a part of the Panther application belong to Panther via two methods:

- For resources that support custom names (e.g. lambdas, queues, dynamo tables) we prefix the name with `panther-`
- For resources that support tagging, we add the tag pair `Application: Panther`

For EC2 resources in particular (e.g. subnets, VPCs, etc.) custom names are not possible and the `Application: Panther` tag is hidden away in the console view. For these resources, we could add the tag `Name: panther-subnet-xyz` and from the AWS console view it would appear as if the resource was named `panther-subnet-xyz`. For ease of use, we should add `Name` tags to all EC2 resources.

### Describe your use cases

Managing Panther (or other systems) from the AWS console, this would let me know immediately that the subnet/vpc/whatever belonged to Panther.

Every other week maybe.

Looking at  the `Application` tag every time I want to figure out if an EC2 resource belongs to Panther.

lindsey-w:
## Background

Add name tag to EC2 resources so it is easy to identify panther-created resources in the aws console.  Prevents user from having to dig into the console to find the "Application: Panther" tag by displaying name field. Closes #1422 

## Changes

- Added name tags to EC2 resources that support tagging.  Reference: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html
- A few of the EC2 resources in use don't support tags: [VPCGatewayAttachment, Route, SubnetRouteTableAssociation, SecurityGroupIngress] 

## Testing

- `mage test:ci` and verifying name shows up in the console.

## Updated Output

image

image

### DIFF
--- a/deployments/bootstrap.yml
+++ b/deployments/bootstrap.yml
@@ -671,6 +671,9 @@ Resources:
       EnableDnsSupport: true
       EnableDnsHostnames: true
       CidrBlock: !FindInMap [SubnetConfig, VPC, CIDR]
+      Tags:
+        - Key: Name
+          Value: panther-vpc
 
   FlowLogs:
     DependsOn: AuditLogsBucketPolicy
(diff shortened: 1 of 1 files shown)


nhakmiller (approved):
Looks good! Would you mind attaching a screenshot showing the names in the AWS console?

Issue #206, opened 2020-02-18
Snapshot Poller Unified Retry Handling


nhakmiller:
### Description

Currently there is not a standard way to handle errors/timeouts in the snapshot poller. Each service scanner decides how it will handle errors, when it will log and continue vs. exiting, when it will request a re-scan vs. fail all the way up so lambda retries the event, etc. This story is to unify this into one standardized method of error handling.

### Acceptance Criteria

All snapshot pollers will follow the same methodology when:

- Deciding when/how to initiate re-scans
- Deciding in general when to return an error, and when to log an error and continue
  - Deciding how to handle situations where enriching API calls for a resource return an error
- Handling timeouts

PR #1517, opened 2020-09-10
Fix snapshot poller errors

austinbyers:
## Background

When deployed to my dev account, the snapshot-poller fails to scan the IAM password policy:

```
IAM.GetAccountPasswordPolicy: NoSuchEntity: The Password Policy with domain name 101802775469 cannot be found.\n\tstatus code: 404
```

But we're specifically checking for that error code and ignoring it:

```go
if awsErr, ok := err.(awserr.Error); ok &&  awsErr.Code() == iam.ErrCodeNoSuchEntityException {
    // this means it does not exist, handle it accordingly...
}
```

So why isn't this triggering? Because the recent snapshot poller refactor (#206 ) made better use of the `errors` package for wrapping errors with context. This unfortunately means that you can no longer use a standard type assertion - a wrapped error is *not* of type `awserr.Error`.

So I've replaced everywhere in the codebase (mostly snapshot poller) `if awsErr, ok := err.(awserr.Error)` with `errors.As(err, &awsErr)`. Using `errors.As` works fine if the err is nil or if it is not wrapped, so it's safer to use this everywhere to prevent the bug from happening again

## Changes

- Use `errors.As` instead of awserr type assertions
- Use aws constants for error codes when possible
- Bump `policy-engine` memory and timeout to match the `rules-engine` (I saw `policy-engine` timeouts in the early stages of some of my test deployments)

## Testing

- Deployed to my dev acct - snapshot errors subsided
- Go playground to ensure I'm using the `errors` pkg correctly - https://play.golang.org/p/g0-iRVZiBl1

### DIFF
--- a/internal/compliance/snapshot_poller/pollers/aws/cloudformation_stack.go
+++ b/internal/compliance/snapshot_poller/pollers/aws/cloudformation_stack.go
@@ -130,13 +130,12 @@ func getStack(svc cloudformationiface.CloudFormationAPI, stackName string) (*clo
 		StackName: aws.String(stackName),
 	})
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok {
-			if awsErr.Message() == "Stack with id "+stackName+" does not exist" {
-				zap.L().Warn("tried to scan non-existent resource",
-					zap.String("resource", stackName),
-					zap.String("resourceType", awsmodels.CloudFormationStackSchema))
-				return nil, nil
-			}
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) && awsErr.Message() == "Stack with id "+stackName+" does not exist" {
+			zap.L().Warn("tried to scan non-existent resource",
+				zap.String("resource", stackName),
+				zap.String("resourceType", awsmodels.CloudFormationStackSchema))
+			return nil, nil
 		}
 		return nil, errors.Wrapf(err, "CloudFormation.DescribeStacks: %s", stackName)
 	}
@@ -175,8 +174,8 @@ func detectStackDrift(cloudformationSvc cloudformationiface.CloudFormationAPI, a
 		return detectionID.StackDriftDetectionId, nil
 	}
 
-	awsErr, ok := err.(awserr.Error)
-	if !ok || awsErr.Code() != "ValidationError" {
+	var awsErr awserr.Error
+	if !errors.As(err, &awsErr) || awsErr.Code() != "ValidationError" {
 		// Run of the mill error, stop scanning this resource
 		return nil, errors.Wrapf(err, "CloudFormation.DetectStackDrift: %s", aws.StringValue(arn))
 	}
--- a/internal/compliance/snapshot_poller/pollers/aws/cloudtrail.go
+++ b/internal/compliance/snapshot_poller/pollers/aws/cloudtrail.go
@@ -123,15 +123,14 @@ func getTrailStatus(svc cloudtrailiface.CloudTrailAPI, trailARN *string) (*cloud
 func listTagsCloudTrail(svc cloudtrailiface.CloudTrailAPI, trailArn *string) ([]*cloudtrail.Tag, error) {
 	out, err := svc.ListTags(&cloudtrail.ListTagsInput{ResourceIdList: []*string{trailArn}})
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok {
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) && awsErr.Code() == cloudtrail.ErrCodeARNInvalidException {
 			// At this point in the scan, we know the ARN is valid. This exception is thrown when you
 			// try to make a cloudtrail:ListTags API call on a trail in another account. The only
 			// trail that we would be scanning in another account is an organization trail, which we
 			// are going to discard at the end of the scan anyways so it doesn't matter what we
 			// return here as long as it's not an error.
-			if awsErr.Code() == cloudtrail.ErrCodeARNInvalidException {
-				return nil, nil
-			}
+			return nil, nil
 		}
 		return nil, errors.Wrapf(err, "CloudTrail.ListTags: %s", aws.StringValue(trailArn))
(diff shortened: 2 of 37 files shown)


internal/compliance/snapshot_poller/pollers/aws/redshift_cluster.go
```diff
@@ -89,13 +89,12 @@ func getRedshiftCluster(svc redshiftiface.RedshiftAPI, clusterID *string) (*reds
...
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok {
-			if awsErr.Code() == "ClusterNotFound" {
-				zap.L().Warn("tried to scan non-existent resource",
-					zap.String("resource", *clusterID),
-					zap.String("resourceType", awsmodels.RedshiftClusterSchema))
-				return nil, nil
-			}
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) && awsErr.Code() == redshift.ErrCodeClusterNotFoundFault {
```
austinbyers, line 93: I confirmed this constant is `ClusterNotFound`, not sure why that package chose to suffix all their names with "Fault", but it's not in the string

internal/compliance/snapshot_poller/pollers/aws/iam_user.go
```diff
@@ -525,7 +524,8 @@ func PollIAMUsers(pollerInput *awsmodels.ResourcePollerInput) ([]*apimodels.AddR
 	// Build the credential report for all users
 	userCredentialReports, err = buildCredentialReport(iamSvc)
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok {
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) && awsErr.Code() == throttlingErrorCode {
 			// Check if we got rate limited, which happens sometimes when the credential report takes a long time to generate
 			if awsErr.Code() == throttlingErrorCode {
```
nhakmiller, line 530: Can you remove the bottom conditional if we're including the condition in the top one?

internal/compliance/snapshot_poller/pollers/aws/s3_bucket.go
```diff
@@ -124,11 +124,10 @@ func getBucketLogging(s3Svc s3iface.S3API, bucketName *string) (*s3.GetBucketLog
 func getObjectLockConfiguration(s3Svc s3iface.S3API, bucketName *string) (*s3.ObjectLockConfiguration, error) {
 	out, err := s3Svc.GetObjectLockConfiguration(&s3.GetObjectLockConfigurationInput{Bucket: bucketName})
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok {
-			if awsErr.Code() == "ObjectLockConfigurationNotFoundError" {
-				zap.L().Debug("no object lock configuration found", zap.String("bucket", *bucketName))
-				return nil, nil
-			}
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) && awsErr.Code() == "ObjectLockConfigurationNotFoundError" {
```
nhakmiller, line 128: Oh this is actually a really big one, glad you got this.

nhakmiller (approved):

internal/compliance/snapshot_poller/pollers/aws/iam_user.go
austinbyers: Yep, good call, must have missed this one

austinbyers:
@nhakmiller and I also observed another common ops alarm - compliance-api warnings on every initial deployment. This is due to compliance status not yet existing when the resource-processor checks for it, and is expected behavior.

The core gateway-api logger was logging "Warn" messages for all 4XX errors, but we have ops alarms on warning messages. I've changed this logging to an "Info" for the same reason we had already removed CloudWatch alarms on API gateway 4XX alarms - 404 is not an error in the infra, it's the correct response from the API. We should not alarm on that behavior; the calling service would log an error if the response isn't what it expects

internal/compliance/snapshot_poller/pollers/aws/cloudformation_stack.go
```diff
@@ -175,8 +174,8 @@ func detectStackDrift(cloudformationSvc cloudformationiface.CloudFormationAPI, a
 		return detectionID.StackDriftDetectionId, nil
 	}
 
-	awsErr, ok := err.(awserr.Error)
-	if !ok || awsErr.Code() != "ValidationError" {
+	var awsErr awserr.Error
+	if !errors.As(err, &awsErr) || awsErr.Code() != "ValidationError" {
```
kostaspap, line 178: Shouldn't this be `&&`? If the casting fails, awsErr will be nil.

internal/compliance/snapshot_poller/pollers/aws/cloudtrail.go
```diff
@@ -123,15 +123,14 @@ func getTrailStatus(svc cloudtrailiface.CloudTrailAPI, trailARN *string) (*cloud
 func listTagsCloudTrail(svc cloudtrailiface.CloudTrailAPI, trailArn *string) ([]*cloudtrail.Tag, error) {
 	out, err := svc.ListTags(&cloudtrail.ListTagsInput{ResourceIdList: []*string{trailArn}})
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok {
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) && awsErr.Code() == cloudtrail.ErrCodeARNInvalidException {
```
kostaspap, line 127: We had a small utility function in Enterprise code that helps check AWS for such errors. I am backporting it here: https://github.com/panther-labs/panther/pull/1522

Do you think it makes sense to update that package with a new function e.g. `ContainsAnyError` in addition to the `IsAnyError`. Could help make this code compact

internal/compliance/snapshot_poller/pollers/aws/cloudformation_stack.go
austinbyers: Right, it's the negative condition, I tripped up on that at first as well

PR #1519, opened 2020-09-11
Fix multi-region publication

austinbyers:
## Background

`mage master:publish` failed with:

```
17:51:42	INFO	[master:publish]	publishing to us-east-1
Error: failed to describe https://panther-community-us-west-2.s3.amazonaws.com/v1.8.0-beta/panther.yml : BadRequest: Bad Request
```

Notice the region mismatch - after the mage refactor, most commands just use the region implicitly associated with the current session. But here, we need to pass in an explicit region as we loop through each region to publish the release

## Changes

- Fix `mage master:publish` by replacing the region in the global `clients` package. This would not work for multiple clients in parallel in different regions, but that's a problem for another day

## Testing

- Publication of v1.8.0-beta succeeded

### DIFF
--- a/tools/mage/clients/aws.go
+++ b/tools/mage/clients/aws.go
@@ -63,13 +63,18 @@ var (
 )
 
 // Build the AWS session from credentials - subsequent calls return the cached result.
-func getSession() *session.Session {
-	if awsSession != nil {
+func getSession(region string) *session.Session {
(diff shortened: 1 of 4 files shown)


nhakmiller (approved):

nhakmiller (approved):
LGTM

PR #1520, opened 2020-09-11
Re-export common field types from the pantherlog module

alxarch:
## Background

I forgot to add this PR a while back and docs were not up-to date with the code base. Thankfully someone noticed it on Slack.

Re-export field types from the pantherlog package so event types only need to import a single package.
This makes explaining the process of adding support for a new log type much easier.
It also allows us to change implementations of a field type in the future without modifying parser code

## Changes

- Re-export types for common fields from the pantherlog using type aliases

## Testing

- mage test:ci

### DIFF
--- a/internal/log_analysis/log_processor/pantherlog/pantherlog.go
+++ b/internal/log_analysis/log_processor/pantherlog/pantherlog.go
@@ -1,12 +1,21 @@
 package pantherlog
 
 /**
+ * Panther is a Cloud-Native SIEM for the Modern Security Team.
  * Copyright (C) 2020 Panther Labs Inc
  *
- * Panther Enterprise is licensed under the terms of a commercial license available from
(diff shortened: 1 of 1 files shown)


internal/log_analysis/log_processor/pantherlog/pantherlog.go
```diff
@@ -0,0 +1,36 @@
+package pantherlog
+
+/**
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * Panther Enterprise is licensed under the terms of a commercial license available from
+ * Panther Labs Inc ("Panther Commercial License") by contacting contact@runpanther.com.
+ * All use, distribution, and/or modification of this software, whether commercial or non-commercial,
+ * falls under the Panther Commercial License to the extent it is permitted.
+ */
```
kostaspap, line 10: Updated licence

kostaspap (approved):

PR #1522, opened 2020-09-11
AWS Error checker utility 

kostaspap:
## Background

Backporting a small function to help check for AWS error codes. 

## Changes

- Backported from Enterprise

## Testing

- `mage test:ci`

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -18,15 +18,6 @@ package awsutils
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-/**
- * Copyright (C) 2020 Panther Labs Inc
- *
- * Panther Enterprise is licensed under the terms of a commercial license available from
- * Panther Labs Inc ("Panther Commercial License") by contacting contact@runpanther.com.
- * All use, distribution, and/or modification of this software, whether commercial or non-commercial,
- * falls under the Panther Commercial License to the extent it is permitted.
- */
-
 import "github.com/aws/aws-sdk-go/aws/awserr"
 
 // Method returns true if the provided error is an AWS error with any
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -1,6 +1,24 @@
 package awsutils
 
 /**
+ * Panther is a Cloud-Native SIEM for the Modern Security Team.
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


tools/mage/clients/aws.go
```diff
@@ -80,7 +80,7 @@ func getSession() *session.Session {
 	// Load and cache credentials now so we can report a meaningful error
 	creds, err := awsSession.Config.Credentials.Get()
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NoCredentialProviders" {
+		if awsutils.IsAnyError(err, "NoCredentialProviders") {
```
kostaspap, line 83: There are more places in the code were we make such checks. Updated some to demonstrate how this function can be used.

giorgosp (approved):

pkg/awsutils/awsutils.go
```diff
@@ -0,0 +1,36 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import "github.com/aws/aws-sdk-go/aws/awserr"
+
+// Method returns true if the provided error is an AWS error with any
+// of the given codes.
+func IsAnyError(err error, codes ...string) bool {
+	awserror, ok := err.(awserr.Error)
```
austinbyers, line 26: Careful: this will *not* work if the error is wrapped with the `errors` package. This led to some serious bugs: #1517 

To be safe, use `errors.As` instead of checking the type explicitly:

```go
    var awserror
    if errors.As(err, &awserror) {
```

This function will work whether the error is wrapped or not, or if it's nil. So it's safer overall

austinbyers (changes requested):

pkg/awsutils/awsutils.go
austinbyers: I know it adds a dependency to this library that isn't technically needed, but we use the `errors` package throughout most of our code - I don't want a helper library that will silently fail for a large class of our own errors

rleighton (approved):

pkg/awsutils/awsutils.go
rleighton: Good suggestion.

kostaspap: I was thinking it might make sense to have two methods : `IsAnyError` and `ContainsAnyError`. The first checks if the `error` provided matches any of the provided codes. The second checks if any of the errors in the chain match the provided codes. I was thinking there might be a case where we want to distinguish between the two scenarios. 
Do you think it makes sense? Or will we end up using the `Contains` all the time? 
I made a comment about this [here](https://github.com/panther-labs/panther/pull/1517#discussion_r486823329)

austinbyers: Most of the time, we're checking for a single AWS error code. But sometimes that's not sufficient and we have to instead check the error message. Very rarely, we check for more than one error code. Personally I think the helper function is overkill for this, but doesn't really matter to me either way. I was just want to make sure this:

```go
if awsErr, ok := err.(awserr.Error)
```

doesn't appear anywhere in the codebase, since it will not work with the `errors` package

austinbyers (approved):
Looks good after fixing the type check

kostaspap:
> Looks good after fixing the type check

Updated check.

PR #1523, opened 2020-09-11
Alert Delivery Status [1/3]: Add tabs in Alert Details Page

3nvi:
## Background

This PR is the [1/3] of PRs that will allow delivery failures to be displayed in the UI. The purpose of this is to limit the change footprint and to allow easier reviews.

This PR specifically modifies the UI of the alert details page, by adding a tabbing functionality to it

## Changes

- Split the subcomponents of the alert page in multiple ones
- Split details & events  into 2 separate  tabs
- Fix UI of tabber to align better with the UI

## Screenshots

Screen Shot 2020-09-11 at 2 57 44 PM
Screen Shot 2020-09-11 at 2 57 59 PM

## Designs

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/39dcb3e2-1e72-49b2-afab-3a59c8459f66/files/c2bbc1a8-4778-4cd6-93a5-4630d8868201/pages/046BD146-3306-45D4-BBF8-1E6FC92C272D

## Testing

- In this  PR we tested that tabbing works correctly  (including lazy loading) and shows the correct data


web/src/pages/AlertDetails/AlertDetails.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+    </Route>,
+    {
+      mocks,
+      initialRoute: `${urls.logAnalysis.alerts.details(alert.alertId)}?section=details`,
+    }
+  );
+
+  // expect to see all the data, but not to see the "Triggered events
+  expect(await findByText('Rule')).toBeInTheDocument();
+  expect(getByText('Rule').closest('[hidden]')).not.toBeInTheDocument();
```
vorillaz, line 65: Using closest is not the best selector, re-assuring that something is in the right place.

web/src/pages/AlertDetails/AlertDetails.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+  expect(getByText('Rule Threshold')).toBeInTheDocument();
+  expect(getByText('Deduplication Period')).toBeInTheDocument();
+  expect(getByText('Deduplication String')).toBeInTheDocument();
+  expect(queryByText('Triggered Events')).not.toBeInTheDocument();
+
+  // unmount everything
+  unmount();
+
+  // remount with the events section
+  const { findByText: _findByText, getByText: _getByText } = render(
```
vorillaz, line 75: I should be considering splitting these two into 2 separate test scenarios.

web/src/pages/AlertDetails/AlertDetails.tsx
```diff
@@ -105,16 +124,49 @@ const AlertDetailsPage = () => {
...
+        <AlertDetailsBanner alert={alertData.alert} rule={ruleData?.rule} />
+        <Card position="relative">
+          <Tabs
+            index={sectionToTabIndex[urlParams.section] || 0}
+            onChange={index => updateUrlParams({ section: tabIndexToSection[index] })}
+          >
+            <Box px={2}>
+              <TabList>
+                <Tab>
+                  {({ isSelected, isFocused }) => (
```
vorillaz, line 137: Lot's of duplication here, how about splitting this one into a new component?

web/src/pages/AlertDetails/AlertDetailsEvents/AlertDetailsEvents.tsx
```diff
@@ -42,18 +42,33 @@ const AlertDetailsEvents: React.FC<AlertDetailsEventsProps> = ({ alert, fetchMor
...
 
   return (
-    <Panel
-      title="Events"
-      actions={
+    <Flex direction="column" spacing={6}>
+      <Flex justify="space-between" align="center">
+        <Flex align="center" spacing={2}>
+          <Heading size="x-small">
+            <b>{alert.eventsMatched}</b> Triggered Event{alert.eventsMatched > 1 ? 's' : ''}
```
vorillaz, line 49: Yikes! A naive pluralization function can be cleaner and reusable 
Refs:
https://github.com/plexis-js/plexis/tree/master/packages/toPlural

web/src/pages/AlertDetails/AlertDetailsEvents/AlertDetailsEvents.tsx
```diff
@@ -42,18 +42,33 @@ const AlertDetailsEvents: React.FC<AlertDetailsEventsProps> = ({ alert, fetchMor
...
           totalPages={alert.eventsMatched}
           onPageChange={setEventDisplayIndex}
         />
-      }
-    >
-      <JsonViewer data={JSON.parse(JSON.parse(alert.events[eventDisplayIndex - 1]))} />
-    </Panel>
+      </Flex>
+      <Card variant="dark" p={6}>
+        <JsonViewer data={JSON.parse(JSON.parse(alert.events[eventDisplayIndex - 1]))} />
```
vorillaz, line 69: Ditto 👭

web/src/pages/AlertDetails/AlertDetails.test.tsx
3nvi: how can you verify that something is not hidden? I want  to check that nothing above  it is "hidden", so I use this selector as a means of verifying that something is 100% visible 😞 

Do you have an alternative?

web/src/pages/AlertDetails/AlertDetailsEvents/AlertDetailsEvents.tsx
3nvi: Sure, gonna copy it though to add Typings (since the package is small enough, but doesn't have typings)

web/src/pages/AlertDetails/AlertDetails.tsx
3nvi: I'm leaving it intentionally like that, since `Tabs, TabList, Tab, TabPanels & TabPanel` are all pounce components.

I figured it's easier  to reason about even with this minor duplication, rather than using `Tabs, TabList, **CustomTabComponent**, TapPanels & TabPanel` 

Do you feel it there are more benefits if I interfered in the chain with a custom component ?

web/src/pages/AlertDetails/AlertDetails.test.tsx
3nvi: yeah was debating on that. Will do

web/src/pages/AlertDetails/AlertDetailsEvents/AlertDetailsEvents.tsx
3nvi: will abstracting, but since this is related to a single use-case, I won't be adding it to utils

web/src/pages/AlertDetails/AlertDetails.tsx
vorillaz: Not at all, just in case we want to make a change the total footprint of the changes gets multiplied.

web/src/pages/AlertDetails/AlertDetailsEvents/AlertDetailsEvents.tsx
vorillaz: Yeap, that should work I guess. We can create something like tiny utils for string manipulation.

web/src/pages/AlertDetails/AlertDetails.test.tsx
vorillaz: How about using a more robust selector (aria, test id or so) and `toBeInTheDocument` assertion or even the [`toBeVisible`](https://github.com/testing-library/jest-dom#tobevisible)

web/src/pages/AlertDetails/AlertDetails.tsx
3nvi: sure. Just because this is  not present only here but in a couple of other places, would creating a  separate PR after this is merged suffice?

I'm proposing that since the changes of the current PR would increase from it and the  scope of the  PR would change as well.

P.S. I won't forget it, it's in my TODO list 😄

web/src/pages/AlertDetails/AlertDetails.test.tsx
3nvi: Done 😄

web/src/pages/AlertDetails/AlertDetailsEvents/AlertDetailsEvents.tsx
3nvi: Due to the smart  way it was implemented, we needed TS interface overloads for this to work. Take  a look at the changes in the utils file 😄

vorillaz (approved):
Pre approving this one and we can jump into moving the tabs logic into composable blocks 📦

PR #1527, opened 2020-09-11
Pull v1.8 release bugfixes back to master

austinbyers:
## Background

Pulling the `v1.8.0-beta` bugfixes back to master. This is a standard `git pull origin release`, and we will do this as a merge (not a squash) to keep the commit history aligned.

This is not really an "enterprise backport", but that label should stop panther-bot from trying to sync these changes directly to enterprise.

Again, I plan to re-evaluate our branch management and hopefully figure out a better way to do this in the future.


nhakmiller (approved):

austinbyers:
Doing this merge also means that the master tags are back to what you would expect: `git describe --tags` on `master` now shows `v1.8.0-beta-16-g74b8f7c4`

austinbyers:
@3nvi panther-bot [tried to sync this change](https://github.com/panther-labs/panther/runs/1103320413) even though it had the label it's supposed to ignore. It's skipping that step:

Screen Shot 2020-09-11 at 12 23 23 PM

PR #1533, opened 2020-09-14
Add proper grid alignment for rows in rule details

vorillaz:
## Description

This one fixes some alignment issues regarding the threshold values. Tags may get quite lengthy breaking the lines between the columns, thus a grid is required in order to create a more structured way for displaying the rule details.

#### Before:
Screenshot 2020-09-14 at 1 43 44 PM

#### After:
Screenshot 2020-09-15 at 12 27 04 PM

### Checklist:

### DIFF
--- a/web/src/pages/AlertDetails/AlertDetailsInfo/AlertDetailsInfo.tsx
+++ b/web/src/pages/AlertDetails/AlertDetailsInfo/AlertDetailsInfo.tsx
@@ -75,68 +75,126 @@ const AlertDetailsInfo: React.FC<AlertDetailsInfoProps> = ({ alert, rule, alertD
       )}
       <Card variant="dark" as="section" p={4}>
         <SimpleGrid columns={2} spacing={5} fontSize="small-medium">
-          {rule ? (
-            <Flex spacing={5}>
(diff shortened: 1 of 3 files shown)


kostaspap:
Thanks for fixing this! I would recommend getting a code review from one of the FE team members before shipping this.

3nvi (commented):
Hey @vorillaz thanks for that! The solution is great, but the spacing is a bit off. 

On the 2nd screenshot  the horizontal distance between the items is 5 times as big compared to the 1st screenshot.

Also if possible, can we apply the same solution to the following components:

- AlertDetailsInfo
- PolicyDetailsInfo

since they all implement  the "same logic"

vorillaz:
@3nvi You were right, updated the screenshots, fixed the column size according to the [new design grid](https://share.goabstract.com/58e615ef-9100-42ad-a85a-b0f21047c64e?mode=build&sha=a16f74b5775582ae7c69893be7e25659deef7678), and finally patched the other pages using the same specs.

3nvi (approved):
LGTM 🎉 

Haven't seen the other components, but SS  for alerts page looks correct  :)

PR #1534, opened 2020-09-14
Store rule errors in S3

### LINKED ISSUE #1270: Backend should store errors that occurred while running rules over incoming events
### Description

Backend should store errors that occurred while running rules over incoming events

### Acceptance Criteria

- Backend should store errors that occurred while running rules over incoming events

kostaspap:
## Background

Closes #1270 

This PR introduces new functionality that allows storing events that cause a rule to fail in S3. 

It is the first in a series of PR to add the new capability. Decided to go with small PRs to make sure the reviewers are not overwhelmed. 

## Changes included

- Store the events that cause a rule to fail in S3 in a new prefix of the processed data bucket. 

## Changes NOT included
- Indexing the events in Glue Data catalog
- APIs to retrieve the events

## Testing

- Unit tests
- E2E in my dev account


internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -37,22 +37,35 @@ def __init__(self, analysis_api: AnalysisAPIClient) -> None:
...
             self._populate_rules()
 
-        matched: List[EventMatch] = []
+        engine_results: List[EngineResult] = []
 
         for rule in self.log_type_to_rules[log_type]:
             self.logger.debug('running rule [%s]', rule.rule_id)
             result = rule.run(event)
             if result.exception:
+                # TODO: remove error logging once error reporting notification system is in place
```
giorgosp, line 52: You may write your TODOs as `TODO(kostaspap)` for easier later lookup and preventing someone else from picking them up

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -37,22 +37,35 @@ def __init__(self, analysis_api: AnalysisAPIClient) -> None:
...
         """
         if datetime.utcnow() - self._last_update > _RULES_CACHE_DURATION:
             self._populate_rules()
 
-        matched: List[EventMatch] = []
+        engine_results: List[EngineResult] = []
 
         for rule in self.log_type_to_rules[log_type]:
             self.logger.debug('running rule [%s]', rule.rule_id)
             result = rule.run(event)
```
giorgosp, line 50: Consider (not necessarily in this PR) returning a custom exception "`RuleException`" from `rule.run()` so that we distinguish it from the engine's exceptions (bugs or whatever).
We only want to write rule exceptions to s3, don't we?

giorgosp (approved):
LGTM. 
👍  👍 For the useful comments in the code and trying to make the PR small enough.

internal/log_analysis/rules_engine/src/engine.py
kostaspap: Hmmm, I am not 100% sure about the recommendation. You mean, maybe I should wrap the exceptions (if any) of the `def rule(event)` method invocation to a `RuleException` and check for that inside the `Rule.run` method? Then populate the `EngineResult.exception` with that? Note that all these will be internal changes to the `Rule` class in any case. I don't think there is any need to include a `RuleException` in the in `EngineResult`. The `EngineResult.exception` is designed to contains exceptions thrown when running the `def rule(event)` method.

rleighton (approved):
LGTM

internal/log_analysis/rules_engine/src/engine.py
kostaspap: I guess this is visible with git blame? In any case, added if you think it adds clarity.

internal/log_analysis/rules_engine/src/engine.py
giorgosp: Sorry, I meant to raise a custom exception in  `rule._run_command()`. And `rule.run()` would be something like this:

```
        try:
            rule_result = self._run_command(self._module.rule, event, bool)
            if rule_result:
                title = self._get_title(event)
                dedup_string = self._get_dedup(event, title)
        except RuleException as e:
              return RuleResult(exception=err)
        except Exception as err:  
            # Engine exception, propagate
            raise
```

It isn't a blocker for this PR but an idea to improve the code there.

kostaspap: Ok, got it. Yes, I think this could be an option yup! I tried to keep all rule "handling" logic inside the Rule class and let the caller just deal with the `RuleResult` object, but what you suggest makes sense as well.

Issue #1366, opened 2020-08-13
GetAlerts API should be more selective listing s3 objects


rleighton:
### Issue
Today each S3 object is searched serially calling the s3 select api, if there are many objects this can be slow, sometimes causing a timeout for users when there are very large numbers of objects.

### Solution
Use the timestamp of the alert to start listing

kostaspap:
We only need to fetch 25 events per API call - 1 go routine should be enough for that. I suspect there is some other issue going on there.

rleighton:
Updated description based on slack call

rleighton:
Moving this to next release due to extra time needed for release 1.7.1 and PTO

PR #1537, opened 2020-09-14
Make get alert more efficient listing S3 files

rleighton:
### Issue
Today each S3 object is searched serially calling the s3 select api, if there are many objects this can be slow, sometimes causing a timeout for users when there are very large numbers of objects.

### Solution
Use the timestamp of the alert to start listing

Closes: https://github.com/panther-labs/panther/issues/1366

## Changes

- Use the `CreationTime` as `StartAfter` prefix

## Testing

- mage test:ci
- deployed to dev account and tested paging thru alerts


kostaspap (approved):
Nice!

PR #1543, opened 2020-09-15
fix: JSON tree overflow

### LINKED ISSUE #1542: UI: Resource Attributes Overflowing the Container
### Describe the bug

When viewing certain resources attributes with long values, they overflow the container

### Steps to reproduce

1. Go to the resource viewer for something like an IAM role policy
2. View the `Attributes` overflow the container

### Expected behavior

The string should be contained within the window

### Environment

Panther Enterprise v1.7.1
Chrome
MacOS

### Screenshots

Screen Shot 2020-09-15 at 8 49 55 AM

3nvi:
## Background

When there's a lot of text in a JSON blob, then the text can overflow outside the container.  This PR makes sure that the text is always restricted within the limits of its bounding box.

For more info, check out the reported issue #1542 

Closes #1542 

## Changes

- Add `wordBreak` on `JsonEditor`

## Testing

- Visually


alexmylonas (approved):
LGTM!

PR #1544, opened 2020-09-15
feat: return logTypes for alerts

s0l0ist:
## Background

FE needs to retrieve the `logTypes` from an Alert so we can filter by them. The record in DDB already has this data so we just need to return it to AppSync.

## Changes

- Add the field to backend structs
- Add the field to graphql schema
- Update the `AlertDetailsFull` and `AlertSummaryFull` fragments

## Testing

- mage test:ci


api/graphql/schema.graphql
```diff
@@ -148,6 +148,7 @@ interface Alert {
   severity: SeverityEnum!
   status: AlertStatusesEnum!
   title: String!
+  logTypes: [String]!
```
3nvi, line 151:
```suggestion
  logTypes: [String!]!
```

api/graphql/schema.graphql
```diff
@@ -161,6 +162,7 @@ type AlertDetails implements Alert {
...
```
3nvi, line 165:
```suggestion
  logTypes: [String!]!
```

api/graphql/schema.graphql
```diff
@@ -177,6 +179,7 @@ type AlertSummary implements Alert {
...
```
3nvi, line 182:
```suggestion
  logTypes: [String!]!
```

3nvi (approved):
Minor TS  fixes. Else LGTM 🎉 

cc @alexmylonas

PR #1545, opened 2020-09-15
Improve log processor aggregation by seeing into the future

rleighton:
## Background

Aggregating output files is very important for the log processor. Having many small files is very inefficient for consumers of the data. Many small files results in many rule processing lambda invocations, expensive Snowpipe bills and can slow Athena queries.

The log processor is architected to do well with respect to aggregation under constant load. After processing an event, it asks the SQS q if there are more events and if so reads them (avoiding another lambda call and aggregating more results into the current execution).

However, the log processor does very poorly aggregating data when the events "trickle" in every few seconds. In this case there is no constant load and we usually write one output file for every input file. This can result in thousands of small files per hour. S3 access logs have this characteristic where we get many small files one after another spaced a few seconds apart.

This could be fixed if we could only see into the future. If the log processor "knew" that in a few seconds there would be more data, then it could read again from the queue (using a wait time for each read in order to ensure we read as many events as possible).

This PR allows the log processor to see a little bit into the future.

A delay is configured into the SQS queue of 5 seconds (configurable). All events are shifted in time this amount. The log processor now ALSO asks the queue if there are any delayed events (this is effectively seeing into the future). If so, the log processor will do a read from SQS (with the 5 second wait time).

The result of this is an almost 30% reduction in the number of files produced for s3 access logs processed in my dev account.

The tradeoff here is that now we are paying for lambda time of ~5 sec per call when in this mode, which can raise lambda costs. That said, this is offset to some degree by fewer lambda executions by the rule processor and partition updater.

I want to try this change in 1.9 and test our internal account that processes many s3 access logs to assess the benefit to the aggregation and the effect on the cost. I will vary the delay every few days to see the effect on aggregation and cost to choose the best default value.  Since this delay is configurable, if for some reason it is causing excessive lambda costs the value can be set to 0 in the CF to disable.

## Changes

- Parameterize SQS delay config option throughout stacks and mage
- Use the count of delayed messages to trigger an sqs read

## Testing

- mage test:ci
- compared before and after aggregation of s3 access logs in dev deployment


internal/log_analysis/log_processor/common/common.go
```diff
@@ -74,6 +75,13 @@ func Setup() {
 	if err != nil {
 		panic(err)
 	}
+
+	// we will use this as the sqs WaitTime, so we want it at least 1 and at most 20
+	if Config.SqsDelaySec < 1 {
+		Config.SqsDelaySec = 1
```
kostaspap, line 81: ~We want to allow 0 right? 0 == disabling it if things don't work out the way we want it.~

Maybe make a new variable named "PollDuration" that gets populated like that:
```
var PollDuration int
if config.SqsDelaySec < 1 {
   PollDuration = 1
} else if confg.SqsDelaySec > 20 {
   PollDuration = 20
} else {
    PollDuration = config.SqsDelaySec
}
```
It will be more clear I think

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -186,29 +186,46 @@ func sqsDataStreams(messages []*sqs.Message,
...
 }
 
-func queueDepth(sqsClient sqsiface.SQSAPI) (numberOfQueuedMessages int, err error) {
+func queueDepth(sqsClient sqsiface.SQSAPI) (totalQueuedMessages int, err error) {
 	getQueueAttributesInput := &sqs.GetQueueAttributesInput{
-		AttributeNames: []*string{aws.String(sqs.QueueAttributeNameApproximateNumberOfMessages)},
-		QueueUrl:       &common.Config.SqsQueueURL,
+		AttributeNames: []*string{
+			aws.String(sqs.QueueAttributeNameApproximateNumberOfMessages),
+			aws.String(sqs.QueueAttributeNameApproximateNumberOfMessagesDelayed), // so we can see into the future!
```
kostaspap, line 193: Can you add a more detailed comment here? Might be confusing for someone reading the code :)

kostaspap (approved):
Love this. Small change, potentially huge benefits in $ and performance.

austinbyers (approved):
Great idea! 🔮

internal/log_analysis/log_processor/common/common.go
rleighton: Done

internal/log_analysis/log_processor/processor/stream.go
rleighton: Done

PR #1546, opened 2020-09-15
Feat: PR syncing between master <-> release

3nvi:
## Background

Based on [our recent discussion](https://panther-labs.atlassian.net/wiki/spaces/EN/pages/666468821/RFC+Branch+Management) regarding branch management, we agreed to explore an idea of three-way syncing.

With that in mind, this PR adds an initial PoC that allows syncing a merged PR on `master` to the `release` branch. Specifically:

-  If a user adds the `sync-to-release` label, the automation will kick in
- The automation will create a new  PR against the `release` branch **of the same** repo, with the changes included in the  original PR
- The author of the original PR will be tagged for review

In general this is a similar behavior to the way we sync OSS <-> Enterprise

## Changes

- Add 2nd Github Job for syncing

## Testing

- TBA


.github/workflows/main.yml
```diff
@@ -84,3 +84,51 @@ jobs:
...
+          token: ${{ secrets.PR_SYNC_TOKEN }}
+          fetch-depth: 1
+
+      - name: Validate that PR was merged and not closed
+        if: ${{ github.event.pull_request.merged == false }}
+        run: exit 1
+
+      - name: Validate that PR should be synced to the release branch
+        if: ${{ !contains(github.event.pull_request.labels, env.SYNC_LABEL) }}
+        run: exit 1
```
austinbyers, line 107: We should `exit 0`, not `exit 1`

Otherwise, we'll get a failed action notification for every non-release merge

.github/workflows/main.yml
```diff
@@ -84,3 +84,51 @@ jobs:
...
+
+      - name: Setup Git Credentials
+        if: ${{ success() }}
```
austinbyers, line 110: What does this do? I assume once `exit` is called above, the rest of the commands will stop and so we don't need this?

.github/workflows/main.yml
```diff
@@ -84,3 +84,51 @@ jobs:
...
+          git config --global user.name "panther-bot"
+
+      - name: Bootstrap a PR branch
+        if: ${{ success() }}
+        working-directory: ${{ env.DEST_REPO_CHECKOUT_PATH }}
+        run: |
+          git checkout -b $HEAD
+          git cherry-pick ${{ github.event.pull_request.merge_commit_sha }} || true
+          git add -A .
+          git commit -m "merge changes from $GITHUB_REPOSITORY" || true
```
austinbyers, line 122: Can we instead put the PR number? E.g. `merge release bugfix from panther-labs/panther#1234`

.github/workflows/main.yml
```diff
@@ -84,3 +84,51 @@ jobs:
...
+          git push origin $HEAD
+
+      - name: Clone original PR
```
austinbyers, line 125: Will the PR be created with the `release` branch as the base? We don't want a PR opened against the master branch

.github/workflows/main.yml
```diff
@@ -84,3 +84,51 @@ jobs:
...
+        if: ${{ success() }}
+        uses: ./src/build/actions/pr_sync # Github actions has a bug and we can't use ENV var here
```
austinbyers, line 127: Have we considered using CircleCI instead of GitHub actions? CircleCI already runs on every PR merge and has access to push to branches. It might simplify the setup, and then we can add more complex steps with `mage` or other tooling in the future.

.github/workflows/main.yml
austinbyers, line 122: If there is a conflict, can we store the output in the PR body? This is a problem with our current syncing setup: you have to open every file to find the merge conflicts. I would like to see a PR body with more helpful information:

```
PR: sync release bugfix

Original PR: panther-labs/panther#1234
Original PR body:
    Changing stuff for the better

Merge conflicts:
    file/this.go: 23
    file/other.js: 45
```

austinbyers (approved):
Thanks for taking the initiative here!

.github/workflows/main.yml
3nvi: Right, currently there is **no way** to perform an early return a job. All steps **will always** be ran, unless we add a conditional on them (`if: ...`). The conditional can be anything that evaluates to a boolean (see L106), but Github actions also offers  built-in checks to see if all previous steps were successful or not (see L110).

Thus, we need a way to say to a step "Do not run if any of your predecessors has failed'. That's what `exit 1`  does.  Contrary to what you think, it doesn't "stop" the job, but marks the particular step as "failed",  so that any check like `if: ${{ success() }}` that comes later can be evaluated to  false (since at least one step has failed). This way you can know whether to execute a particular  step based  on whether **any previous step** has failed  or not.

As you said, this will  bring up an email & a notification, so that's why  this PR was  still a  PoC. I've investigated of alternatives, but right now the "early abort" of a job is not something that's  not properly  supported. What we could do instead is return a "custom success/error" value from each step and check it in the next step to see if we should continue. This  way no notification would be sent, but we still have to add this check to **every single step** in the job, apart from the first one

.github/workflows/main.yml
3nvi: Unfortunately no, this causes the step to be flagged as "failed" but the job continues as normal. Read my comment above.

.github/workflows/main.yml
3nvi: yup, sure thing

3nvi: Is this message automatically provided by git? I think cherry-pick has some variation of it by default, but I'm not fully aware of what it contains. 

If the generated  commit message suffices, then sure. If not, we'll need to find a bash script that does this which it's on the harder  side.

.github/workflows/main.yml
3nvi: This will create a PR against the `release` branch, but you're right that the PR's head ref should use the **same** branch  as a base (currently using `master` as base). Will update.

.github/workflows/main.yml
3nvi: Short answer: No and I'm actually thinking of proposing to move away from CircleCI.

Long Answer: CircleCI has no context attached to it. You don't know when a PR gets merged and you can't acccess  any of the PRs context. That's what's so frustrating about circleCI and that's why we used Github actions for the OSS <-> Enterprise sync in the first place. 

From CircleCI's PoV, everything is a commit push. You cannot have a workflow kick in on a PR event and you certainly don't have any information  about a PR being related to the commit (i.e. you don't get PR name, PR author,  PR labels, etc.). CircleCI operates **after** a certain PR results into a commit being added to a remote ref, without any knowledge of what caused this commit to be added there.

Thus, there is **no way** to use CircleCI for it (did research and saw that this is already a feature request for PR context  support, but no action has been taken since 2018). Due to that, I think we'd be better off to have everything in GH actions eventually.

.github/workflows/main.yml
3nvi: Update: I think we may be able to avoid this with a [conditional on the job level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif)

.github/workflows/main.yml
3nvi: Fixed. Check out L100

.github/workflows/main.yml
3nvi: Updated code to use job conditionals

.github/workflows/main.yml
3nvi: Implemented (haven't ran it yet, but this will be done soon)

.github/workflows/main.yml
3nvi: Updated code to use job conditionals, so I'm resolving this

.github/workflows/main.yml
austinbyers: > As you said, this will bring up an email & a notification, so that's why this PR was still a PoC. I've investigated all alternatives, but right now the "early abort" of a job is not something that's officially supported.

One of the many features CircleCI handles much more cleanly :)

.github/workflows/main.yml
austinbyers: I see. We could use the GitHub API to fetch the PR context explicitly, for example. That would be fairly straightforward (a few lines of code in `mage`), but I agree for now not worth the extra time to implement.

From what I've seen of GitHub actions (not much, you know more than I), it's nowhere near ready to be a replacement for a true CI pipeline. Parallel jobs, cached libraries and containers, shared workspace mounting, we use so many CircleCI features that don't have clear parallels in GitHub.

.github/workflows/main.yml
austinbyers: A `git status` after the cherry-pick would show the conflicts, yes

austinbyers (approved):

.github/workflows/main.yml
3nvi: > Job parallelization

All jobs run parallel by default on Github Actions. There are also [different strategies you can pick from](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobs)

> cached libraries

Caching dependencies is [part of Github Actions from day 1](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows)

> contains

Github actions is  actually better thtan CircleCI on this, since it can [read a Dockerfile](https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions), without requiring you to publish it to dockerhub. The downside is that it always "re-downloads"  the image it builds for you, while CircleCI has it locally (so it's faster)

> shared workspace mounting

In terms  of restoring cached workspace for dependencies, that's supported the same way itt's supported in CircleCI (with `checksum`).  In terms of restoring a workspace with custom artifacts, then Github actions allows you to [upload artifacts](https://github.com/actions/download-artifact) in one job and [download them](https://github.com/actions/download-artifact) in another


In general, it's:


- much faster than CircleCI  (excluding initial docker image download, where CircleCI is faster)
- better documentation
- better community that resolves issues
- better pricing, while staying in Github and allowing you to configure it with terraform
- more freedom to filter a job execution by anything (including files changed, labels, etc.), while CircleCI only  allows  filtering on branch name

The reason I bring all this up, is that I feel that CircleCI is indeed more mature, but Github Actions is more production-ready than you might think 😆

.github/workflows/main.yml
3nvi: Git status is to verbose. It adds a ton of other things apart  from conflicts (files added, files modified, etc.).


I opted for showing the conflicts in the commit body (since only those matter). If there aren't any conflicts, the commit body is empty. I think this is the quickest win here and will give the value we need. It's also perhaps more correct since we attach the cherry-pick conflict data to the commit itself, so you can view the issues from a terminal, not just the Github web page

.github/workflows/main.yml
austinbyers: > much faster than CircleCI (excluding initial docker image download, where CircleCI is faster)

Why is it faster, out of curiosity? The docker image download is unfortunately the majority of the total CI runtime overall, but it's rarely cached on the instance because we use a custom one, so it might not matter either way.

Sure, I'm open to it! Some amount of build minutes are included in our GitHub subscription, so you're right there could be a pricing advantage.

3nvi: @austinbyers I was just going to comment that I realized that only  CircleCI's own images are actually being cached. Our custom ones aren't. 

Holy cow! I wasn't aware of the fact that [customized image caching is not available](https://discuss.circleci.com/t/docker-images-not-cached-between-workflow-jobs/22695/9). Gitthub actions takes 20-30 seconds to download it , so that's again faster than the typical 40 seconds that CircleCI takes :|

PR #1549, opened 2020-09-16
Force `node-fetch` resolution to avoid vulnerabilities

3nvi:
## Background

One of our production dependencies `react-json-view` is unmaintained and has a transitive dependency which is vulnerable. Until we replace this package, we will force-resolve  the vulnerable dependency in a version that is secure

## Changes

- Force resolve `node-fetch: "^2.6.1"`

## Testing

- CI is passing
- Made sure that **every** single other dependency, which `node-fetch` was a transitive dependency of, doesn't have issues with the resolution forcing. In simple words, made sure that everything that was working with the previous version of `node-fetch` still works, since the vulnerable package had its major version upgraded


alexmylonas (approved):

vorillaz (changes requested):
Hey, thanks a lot for this one.
Since `node-fetch` seems the actual problem and since this is a resolved dependency I would suggest removing the "^".
Also, it seems that using a preinstall script along with `npx` could actually break the CI as `npx` and it is not cached.

As far as I can tell the `npm-force-resolutions` is an alternative for `yarn resolutions`, instead of using a workaround beyond the package manager it might be a good idea to staring using `yarn` as the npm package is working along with `package-lock` manifest.

3nvi:
> Also, it seems that using a preinstall script along with `npx` could actually break the CI as `npx` and it is not cached.

Hmm we have indeed a problem  with the CI that  we'll need to fix (most likely by installing the package locally). Thanks forh elping out  with that.

> Since `node-fetch` seems the actual problem and since this is a resolved dependency I would suggest removing the "^".

Sure thing. Either way we don't plan it to be a perma-change

> As far as I can tell the `npm-force-resolutions` is an alternative for `yarn resolutions`, instead of using a workaround beyond the package manager it might be a good idea to staring using `yarn` as the npm package is working along with `package-lock` manifest.

Yup you are right. [npm v7 is planning to support resolutions](https://blog.npmjs.org/post/626173315965468672/npm-v7-series-beta-release-and-semver-major). Because of that, I don't believe it's worth changing to `yarn` at the moment since this is a bigger change than anticipated (we have a monorepo using `npm` in multiple sub-repos, including Docker images, Github actions, etc.).

vorillaz (approved):

tools/mage/setup/setup.go
```diff
@@ -230,8 +230,15 @@ func installNodeModules() error {
 		return nil
 	}
 
-	log.Info("npm install...")
-	args := []string{"install", "--no-progress", "--no-audit"}
+	var args []string
+	if util.IsRunningInCI() {
+		log.Info("npm ci...")
```
3nvi, line 235: @austinbyers `npm ci` is required during CI, else CircleCI doesn't run the pre-build command which "patches" our issue.

Resolution taken from [here](https://github.com/rogeriochaves/npm-force-resolutions/issues/11)
Documentation of `npm ci` found [here](https://docs.npmjs.com/cli/ci.html#description)

s0l0ist (approved):

tools/mage/setup/setup.go
austinbyers: Whenever you have a comment like this, put it in the source code so we have it for the future :)

```go
// 'npm ci' is required in CI because we need a pre-build command...
```

austinbyers: Also, is this just a temporary fix? If so, please note it as such. `// TODO - remove once dependency X is no longer vulnerable`

austinbyers (approved):

tools/mage/setup/setup.go
3nvi: Was mainly checking if you're ok with the code changes style-wise. It's not a temp fix, but I'll add some information in the form of a comment

PR #1550, opened 2020-09-16
Support more than 10 services per ECS cluster

### LINKED ISSUE #1526: ECS clusters can only scan 10 services at a time
### Describe the bug

We're getting snapshot poller errors like this:

`ECS.DescribeServices: arn:aws:ecs:us-west-2:.../staging: InvalidParameterException: service names can have at most 10 items`

We can only request up to 10 services at a time for an ECS cluster, but we just request all of them at once. So for ECS clusters with more than 10 services, we have to batch the requests.

### Steps to reproduce

1. Create an ECS cluster with more than 10 services
2. Deploy / trigger ECS scan
3. See error in the snapshot-poller

### Expected behavior

ECS cluster resource generated correctly

lindsey-w:
## Background

If a cluster has more than 10 services, the snapshot poller will throw an exception. This is caused by the fact the DescribeServices API call only allows querying 10 Service ARNs at a time. Closes #1526 

Reference: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeServices.html 

## Changes

- Batch service arns in groups of 10 for each `DescribeServices` call.  
- Added new Unit Test to verify DescribeServices is called as expected when a cluster has > 10 services.
- Updated mocked `DescribeClusters` call to return an error when more than 10 service arns are passed.

## Testing

- `mage test:ci`
- Manually verified that errors no longer thrown in snapshot poller via CloudWatch
- Manually verified all services are listed in resource entry in dynamoDB


internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
```diff
@@ -223,18 +223,32 @@ func getClusterServices(ecsSvc ecsiface.ECSAPI, clusterArn *string) ([]*awsmodel
...
+			Cluster:  clusterArn,
+			Include:  []*string{aws.String("TAGS")},
+			Services: serviceArns[i:end],
+		})
+		if err != nil {
+			return nil, errors.Wrapf(err, "ECS.DescribeServices: %s", aws.StringValue(clusterArn))
+		}
+		// append each round of results to the overall rawServices var
+		rawServices.Services = append(rawServices.Services, rawServicesPage.Services...)
+		rawServices.Failures = append(rawServices.Failures, rawServicesPage.Failures...)
```
nhakmiller, line 251: What are these failure? Are they failures of us to scan, or failed services? It looks like we just throw this information away, depending on what it means we could do something useful with it.

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
```diff
@@ -223,18 +223,32 @@ func getClusterServices(ecsSvc ecsiface.ECSAPI, clusterArn *string) ([]*awsmodel
...
-
-	if err != nil {
-		return nil, errors.Wrapf(err, "ECS.DescribeServices: %s", aws.StringValue(clusterArn))
+	// The DescribeServices API call does not have a version with builtin paging like the list
+	// API call does. API set a limit of 10 services to describe in a single operation.
+	// Loop through results 10 elements at a time and aggregate the results
+	var batchSize = 10
+	if len(serviceArns) <= batchSize {
+		batchSize = len(serviceArns)
+	}
```
nhakmiller, line 232: Is this conditional necessary? It appears you're already protecting from index out of bounds issues with the conditional on lines 238-240.

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
```diff
@@ -223,18 +223,32 @@ func getClusterServices(ecsSvc ecsiface.ECSAPI, clusterArn *string) ([]*awsmodel
...
-		Include:  []*string{aws.String("TAGS")},
-		Services: serviceArns,
-	})
-
-	if err != nil {
-		return nil, errors.Wrapf(err, "ECS.DescribeServices: %s", aws.StringValue(clusterArn))
+	// The DescribeServices API call does not have a version with builtin paging like the list
+	// API call does. API set a limit of 10 services to describe in a single operation.
+	// Loop through results 10 elements at a time and aggregate the results
+	var batchSize = 10
```
nhakmiller, line 229: If you get rid of the below conditional (see my comment there), this could probably be a constant. Also, because we already have a global `defaultBatchSize` (or something similar) constant, we may wish to name this something like `ecsServiceBatchSize` or something similar.

nhakmiller (commented):
Looks pretty good! A few minor comments

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
lindsey-w: For `DescribeServices`, it looks like the only(?)[ failure ](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/api_failures_messages.html)it would return is that it is `MISSING` - I imagine that between the list and the describe, a service would not be missing.  Should I just go ahead and throw away these failures/not add them to the overall rawServices var? Or do you think that is something we should check for?

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
lindsey-w: Good point - removed the duplicate check here.

internal/compliance/snapshot_poller/pollers/aws/ecs_cluster.go
nhakmiller: I imagine the case is when a service is deleted in between the time we call `ListServices` and `DescribeServices`. In that case I think throwing the info away is the right call, you don't even need to append to `rawServices` I don't think.

nhakmiller (approved):
Looks good to me! Great first backend bug fix, really happy to get this in before 1.9.

PR #1551, opened 2020-09-16
Changing evaluation period for Gateway API errors

kostaspap:
## Background

While investigating some operation issues, I found out the following scenario that we alarmed on: Gateway API had returned 1 5xx error to the caller, but the corresponding AWS Lambda (that was meant to serve the request) didn't have any invocation failures.  See screenshot:

Screen Shot 2020-09-16 at 8 09 27 PM

In this particular case, the error seems to be internal to GatewayAPI. We want to alarm for such errors only if they persist, that's why i have increased the evaluation period for the alarm to 3 (15'). 
Note, that in case that the AWS Lambda returns an error, there is a different set of alarms that are meant to monitor the AWS Lambdas specifically that will fire. This change is just about GatewayAPI monitoring itself. 

## Changes

- Updated evaluation periods for Gateway API alarms

@austinbyers seems similar to the issue you had hit the other time. It seems that the GatewayAPI will report in the metrics for their internal 5xx errors.


rleighton (approved):

austinbyers (approved):
Good to know! Thanks for the fix.

In that case, should we just remove 5XX alarms completely? There is nothing the on-call can do about it - it's completely inactionable, unless the lambda itself is also failing

austinbyers:
Actually, we do need to keep it - an API gateway lambda handler which encounters an internal error *will not error in the lambda function*, it will return the 5XX status code.

So your graph is misleading - the Lambda function *shouldn't* be returning an error, that's how it's designed. Instead, it returns the status code in the response payload. This is how we want all of our functions to work in the future.

So raising this threshold is equivalent to raising the default lambda error threshold, which we decided not to do

austinbyers (changes requested):
See above comment

austinbyers:
To provide more clarity:

The lambda-style APIs (no Gateway) work like this: an actual `error` is returned from the function, which will trigger 2 different alarms: the error will be logged (by the `genericapi` router), and the Lambda function will report the failed execution.

The gateway API handlers work like this: errors are reported in the response code: `{"statusCode": 500}`, which will also trigger 2 different alarms: the error will be logged (by the `gatewayapi` router), and the API gateway will report the failed execution

So to find that error, you'll need to search the CloudWatch logs for `analysis-api`

kostaspap:
> To provide more clarity:
> 
> The lambda-style APIs (no Gateway) work like this: an actual `error` is returned from the function, which will trigger 2 different alarms: the error will be logged (by the `genericapi` router), and the Lambda function will report the failed execution.
> 
> The gateway API handlers work like this: errors are reported in the response code: `{"statusCode": 500}`, which will also trigger 2 different alarms: the error will be logged (by the `gatewayapi` router), and the API gateway will report the failed execution. _But the Lambda function itself executed correctly - the Lambda errors will be 0 in this case_
> 
> So to find that error, you'll need to search the CloudWatch logs for `analysis-api`

Thanks for being so diligent! I checked the code, indeed if the Lambda returned a 5xx we would have logged an error: https://github.com/panther-labs/panther/blob/b4d79f3cea52c9c7bf57e56bbac7b859e5b14b63/pkg/gatewayapi/router.go#L87

I went and checked the logs as well but there was no error or warn recorded. See also the relevant metrics: 

Screen Shot 2020-09-17 at 9 10 25 AM

So, if the Lambda would have returned a 5xx, the Lambda Log Error alarm would have fired, but it didn't in this specific case. 

I think it was a Gateway API issue in this case - the same 5xx alarm fired for both Analysis and Resources API at the same time. In both cases, the Lambda's didn't log an error.

austinbyers (approved):
Huh, strange. Good to know, thanks for digging into it!

PR #1552, opened 2020-09-16
mage master:publish: double-prompt to allow overwriting existing release

austinbyers:
## Background

We use `mage master:publish` to package the release and publish to our public S3 buckets. I'm generally the only one who has used the command so far, but the v1.8 release cycle revealed some problems that I can quickly fix

## Changes

- Don't do a full `mage clean` before every publish, because it takes quite a long time to reinstall all the `npm` packages from scratch. All we really care about is removing the build artifacts (`out/`). We still update dependencies (`mage setup`), we just don't need to erase them ahead of time - `setup` will automatically update any that are out of date.
- Allow overwriting an existing release with a warning and an additional prompt. Sometimes, there is a tiny bug that can be fixed by just republishing the existing version number (especially for release candidates / beta versions). Prompting the dev is better than asking developers to comment out the `mage` check in the source (which I believe @s0l0ist had to do before)
- Format source files after generating them, to avoid dirty repo state

## Testing

- `mage test:go`
- Trying to publish `v1.8.0-beta` when it already exists:

```
14:01:40	INFO	[master:publish]	Publishing panther-community v1.8.0-beta to us-east-1,us-east-2,us-west-2
Are you sure you want to continue? (yes|no) yes
14:01:43	WARN	[master:publish]	https://panther-community-us-east-1.s3.amazonaws.com/v1.8.0-beta/panther.yml already exists
Are you sure you want to overwrite the published release in each region? (yes|no)
```

### DIFF
--- a/tools/mage/master/publish.go
+++ b/tools/mage/master/publish.go
@@ -20,6 +20,7 @@ package master
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/service/s3"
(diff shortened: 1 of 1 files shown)


austinbyers:
It's also worth noting that our S3 buckets are versioned - if a dev overwrites an existing version when they shouldn't, we can always restore the master template to a previous version

s0l0ist (approved):
LGTM

PR #1553, opened 2020-09-17
Added retries for Access Denied errors in Log Processor

kostaspap:
## Background

We have noticed that the Log Processor will occasionally fail to read data from input S3 buckets. The exception will be  `AccessDenied` exceptions.
The issue seems to come and go. In all cases, the data are always processed successfully after retrying. 

While investigating the issue I considered the following:
- Increasing the ExpiryWindow, current set to 1minute. This doesn't seem to be the case in our case - if the STS token had expired while trying to access the data we would have gotten `ExpiredTokenException`. 
- Issue with role configuration. This also doesn't seem to issue. 

## Changes

- Added a new retryer to retry AccessDenied Exceptions. 

## Testing

- mage test:ci
- This is a change that will prove itself after we deploy to accounts and monitor it for a while.


internal/log_analysis/log_processor/common/common.go
```diff
@@ -39,7 +39,7 @@ import (
 )
 
 const (
-	MaxRetries     = 15 // retrying for ~15'
+	MaxRetries     = 12 // ~7'
```
kostaspap, line 42: Previous calculation was wrong. http://backoffcalculator.com/?attempts=12&rate=2&interval=0.1

7' seems suffient for retrying a single operation. The Lambda itself has 15' timeout.

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -48,6 +51,7 @@ const (
 
 	s3BucketLocationCacheSize = 1000
 	s3ClientCacheSize         = 1000
+	s3ClientMaxRetries = 10 // ~1'
```
kostaspap: I am thinking it might be better to retry for less time here. This code is running at the beginning of the "log processing pipeline" - it makes more sense to retry harder at later stages where prematurely deciding to give up could lead to data duplication. Not really strong opinion on this one though.

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -48,6 +51,7 @@ const (
 
 	s3BucketLocationCacheSize = 1000
 	s3ClientCacheSize         = 1000
+	s3ClientMaxRetries        = 10 // ~1'
```
kostaspap, line 54: I am thinking it might be better to retry for less time here. This code is running at the beginning of the "log processing pipeline" - it makes more sense to retry harder at later stages where prematurely deciding to give up could lead to data duplication. Not really strong opinion on this one though.

alxarch (commented):
Just noting that IAM had some outage yesterday. And people were saying this happened before. If this was the reason for the errors, retrying won't help, it will just bill us more for something that's Amazon's fault.

giorgosp (approved):

kostaspap:
> Just noting that IAM had some outage yesterday. And people were saying this happened before. If this was the reason for the errors, retrying won't help, it will just bill us more for something that's Amazon's fault.

I haven't followed the news about the IAM outage, so I am not sure which region or APIs it affected. 

The fix is not for this one time issue, this is a fix for occasional AccessDenied that have been popping up the last weeks. These two might be related since AccessDenied errors have been popping up more and more the last weeks, so might be some AWS internal issue that has been growing. They also might not be - in my past lives I have seen that when assuming STS credentials and trying to access some API call might return AccessDenied exception. This is due to service sync issues. 

I wouldn't be worried about costs that much here - an IAM outage affecting all our S3 access(?) operations shouldn't last that long or be too frequent to be a real cost concern. But retrying for the 1 out 1M requests that fail can avoid some alarms on our side.

internal/log_analysis/log_processor/common/common.go
rleighton: BTW, I came up with the ~15 by actually executing this code, there was no calculation. I think you should put it back or if you want to change it, run the code in a loop and add up the times to confirm it is what you want. Don't rely on the 3rd party app.

rleighton (approved):

internal/log_analysis/log_processor/common/common.go
kostaspap: Fair point. I tested locally and measured the duration of the retries - it seems that 13 retries are needed to hit the 7' retry period.

PR #1566, opened 2020-09-18
fix: remove displaying alert delivery statuses from deleted destinations

s0l0ist:
## Background

Alert statuses are shown for deleted destinations which causes the UI to show invalid logos and calculate delivery status incorrectly.

## Changes

- Filter out only valid destinations

## Testing

- visually


rleighton (approved):
Approving because Nick told me too.

PR #1573, opened 2020-09-21
Telemetry implementation [1/2]

alexmylonas:
## Background

This is PR the [1/2] body of work needed to demonstrate and implent how metrics are going to be tracked for our product.

I decided to split events into three categories
- `Page Views` - those events have the purpose of tracking which pages users visits
- `Errors` - those events track unsuccessful or unexpected behaviors of the product, some may be covered by Sentry 
- `Events` - those include triggers, successful actions, and in general everything not covered on other categories

## Changes

- Added a Mixpanel helper
- Added the mixpanel npm package
- Changed `Content-Security-Policy` to allow mixpanel requests
- Added some initial demo events
- Extra: Added a `curly` linting rule

## Next PR [2/2]

- [ ] Consent Checkbox for Analytics - both on initial popup and general settings page
- [ ] Environment variable for mixpanel
- [ ] Additonal events
- [ ] Attempt to lazy-load mixpanel

## Testing

- Checked mixpanel dashboard that events are tracked as intended


web/src/components/utils/AuthContext/AuthContext.tsx
```diff
@@ -21,6 +21,7 @@ import Auth, { CognitoUser } from '@aws-amplify/auth';
 import { USER_INFO_STORAGE_KEY } from 'Source/constants';
 import { pantherConfig } from 'Source/config';
 import storage from 'Helpers/storage';
+import Mxp from 'Helpers/Mixpanel';
```
3nvi, line 24: can we rename it to something more generic?

web/src/components/utils/AuthContext/AuthContext.tsx
```diff
@@ -273,9 +274,10 @@ const AuthProvider: React.FC<{ children: React.ReactNode }> = ({ children }) =>
         const confirmedUser = await Auth.currentAuthenticatedUser();
         setAuthUser(confirmedUser);
         setAuthenticated(true);
-
+        Mxp.track({ name: 'success-sign-in' });
```
3nvi, line 277: hmm I'd like for the helper function to abstract the fact that we are using `mixpanel`. Something like `logger` or `analytics` or just `log` that accepts a set of top-level arguments (use generic names).

web/src/helpers/Mixpanel.ts
```diff
@@ -0,0 +1,96 @@
...
+// TODO: Pending backend to work
+const mixpanelPublicToken = process.env.MIXPANEL_PUBLIC_TOKEN;
+
+const envCheck =
+  mixpanelPublicToken &&
+  // TODO: Pending backend to work
+  storage.local.read<boolean>(ANALYTICS_CONSENT_STORAGE_KEY) &&
+  process.env.NODE_ENV === 'production';
+
+if (envCheck) mixpanel.init(mixpanelPublicToken);
```
3nvi, line 33: nitpick: braces in the `if` 😆

web/src/helpers/Mixpanel.ts
```diff
@@ -0,0 +1,96 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import mixpanel from 'mixpanel-browser';
```
3nvi, line 19: Mixpanel is a "heavy" library. Make sure we lazyy-load it similar to echarts & sentry

Screen Shot 2020-09-21 at 2 26 54 PM

web/src/helpers/Mixpanel.ts
```diff
@@ -0,0 +1,96 @@
...
+const envCheck =
+  mixpanelPublicToken &&
+  // TODO: Pending backend to work
+  storage.local.read<boolean>(ANALYTICS_CONSENT_STORAGE_KEY) &&
+  process.env.NODE_ENV === 'production';
+
+if (envCheck) mixpanel.init(mixpanelPublicToken);
+
+enum TrackEventEnum {
+  'choose-destination-to-create' = 'Choose Destination to create',
```
3nvi, line 36: Nice! I love the fact that we have type-safety on our actions 😍

web/src/helpers/Mixpanel.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type srcType = 'destinations';
+type ctxType = DestinationTypeEnum;
+
+interface TrackPageViewProps {
+  name: keyof typeof TrackPageViewEnum;
+}
+
+interface TrackEventProps {
```
3nvi, line 59: optional: if We should have a `payload` key that includes all the custom keys that an event might need. This should be tied to the available events above.

Actually, I just realized I think about it exactly like we do when we `dispatch` actions in a `useReducer` hook. Kinda like  we do in `ModalContext`

@vorillaz do you like this approach or is it too strict? I feel the type-safety plays a nice role here.

web/src/helpers/Mixpanel.ts
```diff
@@ -0,0 +1,96 @@
...
+
+enum TrackPageViewEnum {
+  'log-analysis-overview' = 'Log Analysis Overview',
+}
+
+type srcType = 'destinations';
+type ctxType = DestinationTypeEnum;
+
+interface TrackPageViewProps {
+  name: keyof typeof TrackPageViewEnum;
```
3nvi, line 56: optional: seems like `page` is a better name
```suggestion
  page: keyof typeof TrackPageViewEnum;
```

web/src/helpers/Mixpanel.ts
```diff
@@ -0,0 +1,96 @@
...
+if (envCheck) mixpanel.init(mixpanelPublicToken);
+
+enum TrackEventEnum {
+  'choose-destination-to-create' = 'Choose Destination to create',
+  'added-destination' = 'Added Destination',
+  'added-rule' = 'Added Rule',
+  'success-sign-in' = 'Successful Sign in',
+}
+
+enum TrackErrorEnum {
```
3nvi, line 42: I think that all of these are events. Regardless of whether they are a success or a failure event, I feel they all belong in the same enum

web/src/helpers/Mixpanel.ts
```diff
@@ -0,0 +1,96 @@
...
+
+type srcType = 'destinations';
+type ctxType = DestinationTypeEnum;
+
+interface TrackPageViewProps {
+  name: keyof typeof TrackPageViewEnum;
+}
+
+interface TrackEventProps {
+  name: keyof typeof TrackEventEnum;
```
3nvi, line 60: optional: seems like `event` is a better name
```suggestion
  event: keyof typeof TrackEventEnum;
```

web/src/helpers/Mixpanel.ts
```diff
@@ -0,0 +1,96 @@
...
+const actions = {
+  identify: id => {
+    if (envCheck) mixpanel.identify(id);
+  },
+  alias: id => {
+    if (envCheck) mixpanel.alias(id);
+  },
+  people: {
+    set: props => {
+      if (envCheck) mixpanel.people.set(props);
```
3nvi, line 81: instead of doing this check everywhere, I'd recommend something like this:

```tsx
const runInProduction = fn => args => {
const envCheck = ...

  if (envCheck) {
    func(...args);
 }
}
```

so then you can do

```tsx
pageView: ({ name }: TrackPageViewProps) => {
  runInProduction(mixpanel.track(TrackPageViewEnum[name], { type: 'pageview' }));
},
```

web/src/helpers/Mixpanel.ts
```diff
@@ -0,0 +1,96 @@
...
+    },
+  },
+  // TODO: Above functions are not used for the moment, either utilize or remove
+  pageView: ({ name }: TrackPageViewProps) => {
```
3nvi, line 85: nice :)

3nvi (commented):
Really really  good work @alexmylonas . Loving the approach

web/src/helpers/Mixpanel.ts
vorillaz: I don't think this would age well in the long run. Eventually, tags are used for filtering purposes in MixPanel. We should keep the API as strict as possible and the payload in a well defined typeset as metrics would get messed up.

vorillaz (commented):
Nicely done @alexmylonas, just a minor note from my side.
I think we don't think have to manually pass the actual context of each event dispatched.
I guess we shall use the `React.router` API and somehow dispatch a new event based on the current URL, in that manner we can easily group additional events based on the app's routing.

web/src/components/utils/AuthContext/AuthContext.tsx
alexmylonas: Good call, I agree will rename to `Helpers/Analytics`

web/src/components/utils/AuthContext/AuthContext.tsx
alexmylonas: I agree with that but that what exactly do you mean by:
> `log` that accepts a set of top-level arguments (use generic names)

not something `log.info` right?

web/src/helpers/Mixpanel.ts
alexmylonas: done, should we update linter to enforce this?

web/src/helpers/Mixpanel.ts
alexmylonas: I think is necessary to keep these data within sanity levels 😛

web/src/helpers/Mixpanel.ts
alexmylonas: good suggestion, will do

web/src/helpers/Mixpanel.ts
alexmylonas: Idd, i am just separating those to help developers use this more cleanly and viewing data on Mixpanel more effectively. 
```
// Current state
Mxp.pageView({ page: 'A page' }
Mxp.track({ name: 'a non-error event', ...}
Mxp.error({ name: 'an error', ... }

// Generic
Mxp.track({ type: 'pageview, name: 'a page'})
Mxp.track({ type: 'event', name: 'a non-error event', ...})
Mxp.track({ type: 'error', name: 'an error'})
```

Also, with the generic approach, i will harder to safe type names based on type of event

web/src/helpers/Mixpanel.ts
alexmylonas: kk its probably doable, i will try it out.

web/src/helpers/Mixpanel.ts
alexmylonas: As a business user, I think I would like to have seperate view for page view, events and errors that occured

web/src/helpers/Mixpanel.ts
alexmylonas: That was my reasoning, behind this strict typing, analytics can get messed up a lot and I feel the stricter that this can be the better.

web/src/helpers/Mixpanel.ts
alexmylonas: 👍

web/src/helpers/Mixpanel.ts
alexmylonas: I like this, with a tweak, i will lazy load mixpanel in this function as well

alexmylonas (commented):
Thanks both of you for the feedback, really constructive. Replied to all, let me know what you think!

alexmylonas:
> Nicely done @alexmylonas, just a minor note from my side.
> I think we don't think have to manually pass the actual context of each event dispatched.
> I guess we shall use the `React.router` API and somehow dispatch a new event based on the current URL, in that manner we can easily group additional events based on the app's routing.

@vorillaz I like this train of thought, you are mostly referring to  "pageView" events right?

web/src/components/utils/AuthContext/AuthContext.tsx
3nvi: 👍

web/src/components/utils/AuthContext/AuthContext.tsx
3nvi: I mean your response regarding `analytics` - renaming would do. `analytics.track` would be fine for now

web/src/helpers/Mixpanel.ts
3nvi: 👍

web/src/helpers/Mixpanel.ts
3nvi: if it's easy, that'd be perfect!

web/src/helpers/Mixpanel.ts
3nvi: Sure. If @vorillaz is onboard, I'm fine

web/src/helpers/Mixpanel.ts
3nvi: yeah. They should be a strict payload for  each event though. For example a `pageview` event should have  **specific** keys as payload

web/src/helpers/Mixpanel.ts
3nvi: I disagree with that. Functions liek that should **not** have side-effects. They should be pure & expected 😄

web/src/helpers/Mixpanel.ts
alexmylonas: He is onboard! Resolving this

web/src/helpers/analytics.ts
```diff
@@ -0,0 +1,87 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import { DestinationTypeEnum } from 'Generated/schema';
+import mx from 'mixpanel-browser';
```
alexmylonas, line 20: I couldn't manage to lazy import this library
```
import('mixpanel-browser').then(mx => {
  mx.init('token')
  mx.track(...)
})
```

Throws an unexpected [error](https://github.com/mixpanel/mixpanel-js/issues/82)
Digging throw issues found [this](https://github.com/mixpanel/mixpanel-js/issues/133#issuecomment-302482849) referring that mixpanel needs to be imported before bootstraping a page.

The alternative is to load mixpanel library on `<head>`

web/.eslintrc.js
```diff
@@ -37,6 +37,7 @@ module.exports = {
     'max-len': 0,
     'no-unused-vars': 'off',
     'no-underscore-dangle': 'off',
+    curly: ['error', 'all'],
```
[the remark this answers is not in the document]
3nvi: Nice!

web/src/components/modals/DeleteUserModal/DeleteUserModal.tsx
```diff
@@ -54,7 +54,9 @@ const DeleteUserModal: React.FC<DeleteUserModalProps> = ({ user, ...rest }) => {
         title: `Successfully deleted user: ${userDisplayName}`,
       });
       // Checking if user deleted is the same as the user signed in
-      if (userInfo.id === user.id) await signOut({ global: true });
+      if (userInfo.id === user.id) {
+        await signOut({ global: true });
+      }
```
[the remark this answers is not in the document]
3nvi: 👍

web/src/helpers/analytics.ts
```diff
@@ -0,0 +1,87 @@
...
+  process.env.NODE_ENV === 'production';
+
+const evaluateTracking = (...args) => {
+  if (envCheck) {
+    mx.init(mixpanelPublicToken);
+    mx.track(...args);
+  }
+};
+
+enum TrackEventEnum {
```
3nvi, line 40: Correct me if I'm wrong, but I think  both me  and @vorillaz said that i t would be better to associate each event with a particular payload.

This can mean to either:

- have a strong type for `track` events other than `src` and `ctx` which are currently "hardcoded" (For example, `src` can be a fixed set of strings and `ctx` can be a set of fixed keys or an arbitrary `Record<string, unknown.`0

- have a strong type for each "event". Meaning that `added-rule` would have different  payload from `success-sign-in`

web/src/helpers/analytics.ts
```diff
@@ -0,0 +1,87 @@
...
+  'picked-destination-to-create' = 'Picked Destination to create',
+  'added-destination' = 'Added Destination',
+  'added-rule' = 'Added Rule',
```
3nvi, line 43: can we stick to a naming convention? `success-sign-in` doesn't match `added-rule`. It could be either all in past tense (`signed-in-succcessfully`) or all without a noun  (i.e. `rule-addition`).

web/src/helpers/analytics.ts
```diff
@@ -0,0 +1,87 @@
...
+  'failed-to-add-destination' = 'Failed to create destination',
+  'failed-to-create-rule' = 'Failed to create Rule',
+  'failed-mfa' = 'Failed MFA',
+}
+
+enum TrackPageViewEnum {
+  'log-analysis-overview' = 'Log Analysis Overview',
+}
+
+type srcType = 'destinations';
```
3nvi, line 57: these  are hardcoded, see comment above

web/src/pages/LogAnalysisOverview/LogAnalysisOverview.tsx
```diff
@@ -72,6 +73,7 @@ const LogAnalysisOverview: React.FC = () => {
   const { alertsBySeverity, totalAlertsDelta, eventsProcessed, eventsLatency } = data.getLogAnalysisMetrics; // prettier-ignore
   const alertItems = alerts?.alerts.alertSummaries || [];
 
+  trackPageView({ page: 'log-analysis-overview' });
```
3nvi, line 76: I want tto find a way to do tthis automatically for all pages, without explicitly adding it in each page. I  don't know if @vorillaz agrees, but perhaps we can add it in an effect on the `Routes.tsx` page, that would observe the history.

Whether we want to keep url params or not (since some of them will be sensitive information like a sourceId or an AWS account ID), is something for business to decide. Perhaps @willow9886 or @jacknagz can help us make a decision here on what "part" of the page we want to track. Would the URL suffice? Do we need hashes and query params?

web/src/helpers/analytics.ts
3nvi: That's super weird. I don't want to block this PR for this, but can we track this in a separate issue? Can you create & assign it to me?

web/src/pages/LogAnalysisOverview/LogAnalysisOverview.tsx
jacknagz: I don't believe I had a requirement to track the overview page... Missing some context here

web/src/helpers/analytics.ts
alexmylonas: Done added this https://github.com/panther-labs/panther/issues/1613

web/src/helpers/analytics.ts
alexmylonas: Tried to copy logic from `ModalContext`, have a look and let me know

web/src/helpers/analytics.ts
alexmylonas: Keeped past tense, seems more appropriate

web/src/hoc/withSEO.tsx
```diff
@@ -30,6 +31,10 @@ function withSEO<P>({ title }: Options) {
     const ComponentWithSEO: React.FC<P> = props => {
       const routerData = useRouter();
 
+      React.useEffect(() => {
+        trackPageView({ path: routerData.history.location.pathname });
+      }, [routerData.history.location.pathname]);
+
```
alexmylonas, line 37: @3nvi do you think this is a valid place for this? 

I still don't think we should track pageviews. Here is an example: 
Rule details have path like this: `/log-analysis/rules/S3 server access/`

So instead of tracking a general `Rule Details view`, we will get stuck with multiple different events that demonstrate the same action.

web/src/pages/LogAnalysisOverview/LogAnalysisOverview.tsx
alexmylonas: @jacknagz There is a debate about how we should track pageViews 
1. either automatically by tracking the changes on pathname i.e '/log-analysis/rules' to '/log-analysis/overview'
2. or manually add an "event" for each page we want to have data for.

- Using the (1), is much easier to implement but has some cons:
  i. Rule details have path like this: `/log-analysis/rules/S3 server access/`, so instead of tracking a general `Rule Details view` event , we will get stuck with multiple different events that demonstrate the same action.
 ii. Event filtering and displaying on analytics will be less intuitive. 
- Using the (2), we will need to add manually those events on all pages that matter.

web/src/hoc/withSEO.tsx
3nvi: I think we agreed to **not track all pages**, unless explicitly instructed so. Currently, my previous comment about tracking all page view is **not needed**, unless someone from business says otherwise

web/src/pages/LogAnalysisOverview/LogAnalysisOverview.tsx
jacknagz: #2 feels right, then

web/src/hooks/useAnalytics.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { PageViewEnum, trackPageView } from 'Helpers/analytics';
+
+function useAnalytics() {
+  const trackPage = (page: PageViewEnum) =>
+    React.useMemo(() => {
+      trackPageView({ page });
+    }, [page]);
+  return { trackPage };
+}
+
```
alexmylonas, line 30: I needed to add this so that re-renders don't trigger duplicates events!

web/src/helpers/analytics.ts
```diff
@@ -0,0 +1,135 @@
...
+  process.env.NODE_ENV === 'production';
+
+const evaluateTracking = (...args) => {
+  if (envCheck) {
+    mx.init(mixpanelPublicToken);
+    mx.track(...args);
+  }
+};
+
+export enum PageViewEnum {
```
3nvi, line 40: nitpick: can we not name an `enum` as  `xxxEnum`? It's like naming a variable for users: `usersVariable`

web/src/helpers/analytics.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export const trackEvent = (payload: TrackEvent) => {
+  evaluateTracking(payload.event, {
+    type: 'event',
+    src: payload.src,
+    ctx: 'ctx' in payload ? payload.ctx : null,
+  });
+};
+
+export enum TrackErrorEnum {
```
3nvi, line 102: since autocomplete is present, I'd consider merging that to our typical events

web/src/helpers/analytics.ts
3nvi: this applies to all enums in this file. For example `EventEnum` could simply be `Events` or `AnalyticsEvents`

web/src/helpers/analytics.ts
```diff
@@ -0,0 +1,135 @@
...
+  FailedToAddDestination = 'Failed to create Destination',
+  FailedToAddRule = 'Failed to create Rule',
+  FailedMfa = 'Failed MFA',
+}
+
+interface ErrorEvent {
+  data: any;
+}
+interface AddDestinationError extends ErrorEvent {
+  error: TrackErrorEnum.FailedToAddDestination;
```
3nvi, line 112: why not use the typical `event` key here?  I don't  think we  need to treat error events differently.

web/src/helpers/analytics.ts
```diff
@@ -0,0 +1,135 @@
...
+  src: SrcEnum.Rules;
+}
+interface MfaError extends ErrorEvent {
+  error: TrackErrorEnum.FailedMfa;
+  src: SrcEnum.Auth;
+}
+
+type TrackError = AddDestinationError | AddRuleError | MfaError;
+
+export const trackError = (payload: TrackError) => {
```
3nvi, line 128: I'm not sold on this yet. I don't feel strongly, but I do believe that it's better to treat errors as actual events instead of having separate typings and  function for them. The typical event could have an optional `data` parameter, although I doubt we'd use that often due to privacy concerns.

The reasoning is that this `type: error` is not a system exception,  it's a known failure from the user's side. Perhaps  we want to track them differently, perhaps not. It's down to what helps business the most. Perhaps @vorillaz can also share his thoughts

web/src/pages/ComplianceOverview/ComplianceOverview.tsx
```diff
@@ -33,6 +35,9 @@ import TopFailingPoliciesTable from './TopFailingPoliciesTable';
 import TopFailingResourcesTable from './TopFailingResourcesTable';
 
 const ComplianceOverview: React.FC = () => {
+  const { trackPage } = useAnalytics();
+  trackPage(PageViewEnum.ComplianceOverview);
```
3nvi, line 39: This needs to be inside a `React.useEffect`

web/src/hooks/useAnalytics.tsx
3nvi: If you  added the `trackPageView` call inside a React effect, you wouldn't need this hook at all. I feel that this is not needed, since the **default** would be  to call `trackPageView` inside a `React.useEffect`

3nvi: What you could do is to create a hook `usePageView` and just execute the effect there:

```tsx
const usePageView = (page: PageViewEnum) => {
  React.useEffect(() => trackPageView(page), [page]);
}
```

and use it like:

```tsx
const Component = () => {
  usePageView(PageViewEnum.ComplianceOverview)

 return <button>clickme</button>

}
```

web/src/helpers/analytics.ts
```diff
@@ -0,0 +1,135 @@
...
+
+const envCheck =
+  mixpanelPublicToken &&
+  // TODO: Pending backend to work
+  storage.local.read<boolean>(ANALYTICS_CONSENT_STORAGE_KEY) &&
+  process.env.NODE_ENV === 'production';
+
+const evaluateTracking = (...args) => {
+  if (envCheck) {
+    mx.init(mixpanelPublicToken);
```
3nvi, line 35: can we check if `init` was already called somehow? We shouldn't call it every single  time, unless it's idempotent

web/src/helpers/analytics.ts
```diff
@@ -0,0 +1,135 @@
...
+  ListAlerts = 'List Alerts',
+  ListLogSources = 'List Log Sources',
+  Home = 'Home',
+}
+
+interface TrackPageViewProps {
+  page: PageViewEnum;
+}
+
+export const trackPageView = ({ page }: TrackPageViewProps) => {
```
3nvi, line 53: Please  add a note for not using this method directly, but  always  importing the related hook

web/src/pages/CreateRule/CreateRule.tsx
```diff
@@ -53,7 +54,12 @@ const CreateRulePage: React.FC = () => {
   const { history } = useRouter();
   const [createRule, { error }] = useCreateRule({
     refetchQueries: [{ query: ListRulesDocument, variables: { input: {} } }],
-    onCompleted: data => history.push(urls.logAnalysis.rules.details(data.addRule.id)),
+    onCompleted: data => {
+      trackEvent({ event: EventEnum.AddedRule, src: SrcEnum.Rules });
```
3nvi, line 58: I feel it's a pain in  the ass to import all these things  to just  file a single event. Look at the sheer number of imports.

Should we perhaps opt for sttring literals? Meaning, should we  add "fixed string" instead of rely on enums?

web/src/helpers/analytics.ts
alexmylonas: Hmm by doing a quick search every `enum` we have follows that pattern, are you sure?

web/src/helpers/analytics.ts
alexmylonas: I feel that this way will be easier for both business and devs to identify and classify events that stray away from the happy path. Maybe the naming should change to something else i.e. `fails`,  `warnings` but those are specific behaviours we want to track down and aim to minimize. I would also like @vorillaz opinion on this.

web/src/pages/CreateRule/CreateRule.tsx
alexmylonas: Had the same train of thought myself, dont like those imports either. I had them as string literals initially but feels more dev friendly this way as you dont have to go find the string. 
@vorillaz thoughts?

web/src/helpers/analytics.ts
vorillaz: I agree with Alex, one way or another this is the first implementation and we shall try to accommodate our needs later on. The idea of tracking "errors" but from a user perspective not an actual thrown error is vital. Personally, I would prefer to dispatch an "event".

web/src/helpers/analytics.ts
3nvi: It's because they are autogenerated 😆 I'm open to any approach, just prefer the aforementioned

web/src/pages/CreateRule/CreateRule.tsx
vorillaz: I don't like it either but I guess it's not a bottleneck as these events you not radically change over time. Being declarative won't hurt anyone and it's just nit picking.

vorillaz (approved):
Overall, that's really good work @alexmylonas , kudos.
I think we shall draw a line as this PR gets fat on comments and it's a bit hard to follow.
I would suggest shipping this one and keep tracking the events tracked, evaluate the actual results and keep an eye for future improvements.

web/src/hooks/useAnalytics.tsx
alexmylonas: Followed your suggestion and update the hook, have a look and let me know!

web/src/helpers/analytics.ts
alexmylonas: agreed, renamed prop `error` to `event`

web/src/helpers/analytics.ts
alexmylonas: Done

web/src/pages/CreateRule/CreateRule.tsx
alexmylonas: Decided to keep enumerations for the moment can be easily revert in the future

3nvi: Sure

3nvi (approved):
Hey @alexmylonas . I'm approving as well. 

I think the condition that it's currently in is more than fine for merging. Any enhancements are welcome, but there's nothing stopping me from moving on even as is


We'll need to do some testing of the tracking functions. This can happen here or in the [2/2] PR

PR #1574, opened 2020-09-21
Storing rule errors in Athena

kostaspap:
## Background

This PR makes Rule errors searchable in Athena. 
* It creates a new DB called `panther_rule_errors`. The DB contains the same tables as panther_rule_matches and panther_log. Each table contains the same columns as `panther_rule_matches` tables + a new column p_rule_error that contains the error that caused the rule matching engine to fail a specific event
* It creates a new view over all the `p_rule_errors` tables. 

Screen Shot 2020-09-21 at 3 40 04 PM
Screen Shot 2020-09-21 at 3 40 55 PM

## Testing

- Unit tests
- Deployed and verified E2E


internal/log_analysis/athenaviews/athenaviews.go
```diff
@@ -95,6 +100,18 @@ func generateViewAllRuleMatches(tables []*awsglue.GlueTableMetadata) (sql string
 	return generateViewAllHelper("all_rule_matches", ruleTables, awsglue.RuleMatchColumns)
 }
 
+// generateViewAllRuleMatches creates a view over all log sources in rule error db the using "panther" fields
```
rleighton, line 103: s/generateViewAllRuleMatches/generateViewAllRuleErrors

internal/log_analysis/datacatalog_updater/process/sync.go
```diff
@@ -59,12 +59,21 @@ func Sync(event *SyncEvent, deadline time.Time) error {
 		logType := event.Continuation.LogType
 		logTable := registry.Lookup(logType).GlueTableMeta() // get the table description
 
-		if event.Continuation.DataType == models.RuleData { // just finish rule matches, log aleady done
+		switch dataType := event.Continuation.DataType; dataType {
+		case models.RuleData:
+			// just finish rule matches, log already done
```
rleighton, line 64: Change to:
```
// finish rule matches (log already done) then do error table from zeroStartTime
```

giorgosp (approved):
LGTM!

internal/log_analysis/datacatalog_updater/process/sync.go
```diff
@@ -59,12 +59,21 @@ func Sync(event *SyncEvent, deadline time.Time) error {
...
 			deadlineExpired, err := syncTable(logTable.RuleTable(), event, startTime, deadline)
 			if err != nil || deadlineExpired {
 				return err
 			}
```
rleighton, line 68: Add below:
```
deadlineExpired, err = syncTable(logTable.ErrorTable(), event, zeroStartTime, deadline)
if err != nil || deadlineExpired {
	return err
}
```

internal/log_analysis/datacatalog_updater/process/sync.go
```diff
@@ -59,12 +59,21 @@ func Sync(event *SyncEvent, deadline time.Time) error {
...
-		} else { // finish log table, then do rule table from zeroStartTime
+		case models.RuleErrors:
+			// finish the rule errors
```
rleighton, line 70: Change to:
```
// finish the rule errors  (rule and log already done)
```

rleighton (approved):
Nice job!

Can you add a "debug" schema change to some table and deploy to test that all 3 tables are updated?  Such as:

-  deploy this patch
- add a column to a parser
- do `export CUSTOM_RESOURCE_VERSION=123` to force a custom resources to run
- deploy
- use glue or athena to confirm change is in all 3 tables
- undo the column change

internal/log_analysis/datacatalog_updater/process/sync.go
```diff
@@ -74,6 +83,8 @@ func Sync(event *SyncEvent, deadline time.Time) error {
 			if err != nil || deadlineExpired {
 				return err
 			}
+		default:
+			return errors.New("Unknown data type " + dataType.String())
```
rleighton, line 87: Add support for the ErrorTable at line 112 below

rleighton (commented):
syn_test.go in the datacatalog_updater needs updating.

internal/log_analysis/athenaviews/athenaviews.go
kostaspap: Fixed

internal/log_analysis/datacatalog_updater/process/sync.go
kostaspap: Updated

internal/log_analysis/datacatalog_updater/process/sync.go
kostaspap: I actually think this is not needed - the Rule Errors tables follow the same partitioning schema as the Rule Matches tables. The time of the partition is related to the processing time not the event time.

internal/log_analysis/datacatalog_updater/process/sync.go
kostaspap: Updated

internal/log_analysis/datacatalog_updater/process/sync.go
kostaspap: Added. It's not 100% clear to me the logic here - is zeroStartTime the time 0? 
Note that in case of rule match and rule error tables, we don't have to start from time 0 - we can start from table creation time since we use the processing time not event time for the partition columns.

kostaspap:
> * export CUSTOM_RESOURCE_VERSION=123

Done. Added a new field in ALB (the table already had data). 
I then ran 
```
SELECT * FROM panther_logs.aws_alb
SELECT * FROM panther_rule_matches.aws_alb
SELECT * FROM panther_rule_errors.aws_alb
```

Saw data and no errors

internal/log_analysis/datacatalog_updater/process/sync.go
kostaspap: Ok, nvm. I figured out the code. `zeroStartTime` actually makes the underlying code use the table creation time.

internal/log_analysis/datacatalog_updater/process/sync.go
kostaspap: Nvm, I figured out the code

kostaspap:
> syn_test.go in the datacatalog_updater needs updating.

Updated

PR #1580, opened 2020-09-22
[git] story and epic issue templates

jacknagz:
## Background

To better ensure consistency in larger projects, I'm adding an Epic issue template and adjusting our Story template to include new headers for designs and more.

## Changes

- Update the story issue template
- Add a new Epic template

## Testing

- None

### DIFF
--- a/.github/ISSUE_TEMPLATE/epic.md
+++ b/.github/ISSUE_TEMPLATE/epic.md
@@ -1,22 +1,22 @@
 ---
-name: Story
-about: Planned or ongoing development work
+name: Epic
+about: A collection of User Stories
 title: ''
-labels: story
+labels: Epic
 assignees: ''
 ---
 
 ### Description
 
-Explain what you are trying to achieve and why this story needs to be implemented. What is the value that this will provide?
+A couple high-level sentences of the feature
 
-### Related Services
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/epic.md
```diff
@@ -0,0 +1,33 @@
+---
+name: Epic
+about: A collection of User Stories
+title: ''
+labels: Epic
```
austinbyers, line 5: An epic is a special ZenHub type which doesn't map natively to GitHub. We don't want to add a GitHub label here, ZenHub should handle it if you create the epic correctly. Creating a proper epic requires the ZenHub extension, and users will need to click the correct button:

Screen Shot 2020-09-22 at 9 00 38 AM

I'm fine with an epic template, just calling out that this template will not automatically create ZenHub epics unless Panther devs click the correct button. Open-source users will not have a way to create epics, which I think is fine

austinbyers (approved):

.github/ISSUE_TEMPLATE/epic.md
jacknagz: Ok, I'll remove the label

PR #1581, opened 2020-09-22
API that allows retrieving Rule Error alerts

### LINKED ISSUE #1269: Create API that allows retrieving rule failures
### Description

Create API that allows retrieving rule failures

### Acceptance Criteria

- Panther backend offers an API that allows retrieving the error that occurred while running rules over events.

kostaspap:
## Background

Closes #1269 

Updated API that allows retrieving rule errors. 
The API is almost identical to the existing API. The ONLY difference is that in case an alert was generated because of a rule error, the response will contain a new field `errorType` with value `RULE_ERROR` .

## Testing

- mage test:ci
- Deployed to my account. Verified: 
  - I could see the rule errors in the list of alerts
  - I got a notification
  - I could retrieve the events through the API


api/lambda/alerts/models/api.go
```diff
@@ -197,6 +197,7 @@ type AlertSummary struct {
 	Title             *string             `json:"title" validate:"required"`
 	LastUpdatedBy     string              `json:"lastUpdatedBy,omitempty"`
 	LastUpdatedByTime time.Time           `json:"lastUpdatedByTime,omitempty"`
+	ErrorType         string              `json:"errorType,omitempty"`
```
kostaspap, line 200: In case an alert is for a rule error, this will be populated with `RULE_ERROR` value, otherwise it will be empty.

rleighton:
Did a quick scan, LGTM. When it is no longer draft I'll take another pass.

kostaspap:
> Did a quick scan, LGTM. When it is no longer draft I'll take another pass.

Done

api/lambda/alerts/models/api.go
giorgosp: Are we going to add other error types soon or you just made it an enum to ease schema migrations later just in case?

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -124,7 +120,9 @@ func (h *Handler) handleNewAlert(rule *ruleModel.Rule, event *AlertDedupEvent) e
 	}
 
 	err := h.sendAlertNotification(rule, event)
-	if err == nil {
+	// If the event was generated by a rule-engine error
+	isError := len(event.ErrorType) == 0
```
giorgosp, line 124: Could tidy this as an `isError()` method if you think it will be used in other places as well.

giorgosp (approved):

api/lambda/alerts/models/api.go
kostaspap: I was considering the CLASSIFICATION_ERROR usecase actually.

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
kostaspap: I like it

api/lambda/alerts/models/api.go
giorgosp: Aha, nice.

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
giorgosp: Btw is the condition correct? Shouldn't have been  `isError := len(event.ErrorType) > 0` ? In that case, would be nice if we could catch this with a unit test.

kostaspap: You are right! Yeah, it makes sense, I will add a unit test.

rleighton (approved):
LGTM

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -40,33 +40,13 @@ import (
...
-		},
-		{
-			"AnalysisType",
-		},
-	}, []metrics.Metric{
-		{
-			Name: "AlertsCreated",
-			Unit: metrics.UnitCount,
-		},
-	})
```
kostaspap, line 57: 1. Regarding the `staticLogger` removal: I wanted to add unit test for metrics to verify that the code is reporting the appropriate metrics. In order to avoid having global state(variable) like static logger, I moved it inside the `Handler` struct. 
2. Regarding replacing `staticLogger` with normal logger: I think it impacts readability of the code to declare the units and metric name in a different place and log it elsewhere. StaticLogger was created to help with "high volume" logging cases, but this is not the case - we log a metric whenever a new alert is created.

pkg/metrics/logger.go
```diff
@@ -112,6 +112,10 @@ type Dimension struct {
 	Value string
 }
 
+type LoggerAPI interface {
```
kostaspap, line 115: I had to introduce this interface to be able to mock metrics logging in the code.

pkg/metrics/logger.go
```diff
@@ -142,23 +146,15 @@ func NewLogger(dimensionSets []DimensionSet) (*Logger, error) {
...
-	err := l.logEmbedded(values, dimensions)
-	if err != nil {
-		zap.L().Error("metric failed", zap.Error(err))
-	}
-}
-
-// LogSingle sends a log with a single metric value
-func (l *Logger) LogSingle(value Metric, dimensions []Dimension) {
-	err := l.logEmbedded([]Metric{value}, dimensions)
+func (l *Logger) Log(dimensions []Dimension, values ...Metric) {
```
kostaspap, line 149: Seemed a bit redundant to have `Log` and `LogSingle` given this can be expressed in a different way. Decided to merge the two.

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
kostaspap: Added unit test and suggested method. 
I had to do some more refactoring to make it possible to unit test the metrics, but now we are protected against such issues. 
I agree that unit testing metrics logging is very very important, it is something that is really hard to catch in integration tests.

kostaspap: Take a look again and let me know what you think @giorgosp

pkg/metrics/logger.go
giorgosp: - Per the Go convention we should name our interfaces after their behaviour, like Logger, HTTPDoer, Greeter, etc.
- Go also allows/suggests narrow interfaces, defining only the behaviour that is needed at the point of use. For example `LoggerAPI` declaration should be moved to `internal/log_analysis/alert_forwarder/forwarder/forwarder.go`.  This will also enable us to get rid of ugly interface names like `SmthAPI` or `lock.Interface`.

pkg/metrics/logger.go
giorgosp: nice

pkg/metrics/logger.go
kostaspap: * There was already a `Logger` struct that's why I decided to name the interface like this. I can change the `Logger` struct if you want. I was just trying to avoid many changes in the metrics code, I really just wanted to add a unit test, not refactor the whole metrics piece. 
* The `Logger` is meant to be common code (our metrics library) - everyone can use that library to log metrics. It's not specific to the forwarder so I wouldn't move it there.

giorgosp: I'm just suggesting to move the interface declaration and rename it to `Logger` or `MetricsLogger`. The naming problem happens because we declare the interfaces in the same package as their implementation. Go gives us the ability (and is a convention) to declare interfaces with only the necessary behavor and close to where they are used (in this example, in `forwarder.go`), instead of creating fat interfaces like in other languages

kostaspap: Chaatted offline: Following Effective Go [guidelines](https://golang.org/doc/effective_go.html#generality),  
```
If a type exists only to implement an interface and will never have exported methods beyond that interface, there is no need to export the type itself
```
we are only exporting the `Logger`interface and making `logger` local.

PR #1584, opened 2020-09-22
API endpoint for most active rules

### LINKED ISSUE #1402: API for Most Active Rules
The log analysis overview page needs to show a chart with the top rules ranked by the number of alerts they sent in a certain time frame

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/master/commits/d0570e02d9fb58bb4e79266e5e7ff6e3978973a8/files/c2bbc1a8-4778-4cd6-93a5-4630d8868201/layers/49B988DE-947E-4204-AA7B-5B48AE7E50E4

### Acceptance Criteria

- The API accepts a date range  as input
- The API returns a list  of rule information (as designated by the needs of the chart above) along with the number of alerts sent
- If it's helpful, the API can optionally return only a fixed number of results (10 or 20) since the chart will only showcase 5  or 10

lindsey-w:
## Background

Request to display information about the most active rules ranked by the number of alerts they sent in a certain time frame.  This change includes both the data generation and the API update.  Closes #1402 

## Changes

- Log the `AlertsTriggered` metric in the alert forwarder.
- Add `alertsByRuleID` metric to the metrics API.  This metric returns time series data. Where the label is the RuleID and the values represent the number of times the rule fired for the given timeframe.

## Testing

- Added new integration test to verify metrics returned successfully and formatted correctly.

input:
```
{
    "GetMetrics": {
        "fromDate": "2020-09-20T23:55:00Z",
        "toDate": "2020-09-21T23:55:00Z",
        "metricNames": [
            "alertsByRuleID"
        ],
        "intervalMinutes": 1440
    }
}
```

response:
```
{
    "alertsByRuleID": {
        "seriesData": {
            "series": [],
            "timestamps": []
        },
        "singleValue": [
            {
                "label": "AWS.S3.ServiceAccess.UnknownRequester",
                "value": 87
            },
            {
                "label": "AWS.IAM.PolicyModified",
                "value": 86
            },
            {
                "label": "AWS.EC2.RouteTableModified",
                "value": 25
            }
        ]
    },
    "fromDate": "2020-09-19T17:50:00Z",
    "intervalMinutes": 14,
    "toDate": "2020-09-24T17:50:00Z"
}
```

~~The input will by default return the top 10 rules that have triggered alerts.  This can be modified to return more/less using the `limit` option:~~ deferred.


api/lambda/metrics/models/api.go
```diff
@@ -44,6 +45,7 @@ type GetMetricsOutput struct {
 	EventsLatency    *MetricResult `json:"eventsLatency,omitempty"`
 	TotalAlertsDelta *MetricResult `json:"totalAlertsDelta,omitempty"`
 	AlertsBySeverity *MetricResult `json:"alertsBySeverity,omitempty"`
+	AlertsByRuleID   *MetricResult `json:"alertsByRuleID,omitempty"`
```
3nvi, line 48: looks good @lindsey-w 🎉

internal/core/metrics_api/api/alert_metrics.go
```diff
@@ -151,3 +156,73 @@ func getTotalAlertsDelta(input *models.GetMetricsInput, output *models.GetMetric
...
+		Namespace:  aws.String(input.Namespace),
+	}, func(page *cloudwatch.ListMetricsOutput, _ bool) bool {
+		listMetricsResponse = append(listMetricsResponse, page.Metrics...)
+		return true
+	})
+	if err != nil {
+		zap.L().Error("unable to list metrics", zap.String("metric", alertsTriggeredMetric), zap.Error(err))
+		return metricsInternalError
+	}
+	zap.L().Warn("found applicable metrics", zap.Any("metrics", listMetricsResponse))
```
nhakmiller, line 177: I don't think this should be a `Warn` level message. `Info` at best, but possibly even `Debug`.

internal/core/metrics_api/api/alert_metrics.go
```diff
@@ -151,3 +156,73 @@ func getTotalAlertsDelta(input *models.GetMetricsInput, output *models.GetMetric
...
+			Label: aws.String(aws.StringValue(metric.Dimensions[0].Value)),
+			MetricStat: &cloudwatch.MetricStat{
+				Metric: metric,
+				Period: aws.Int64(input.IntervalMinutes * 60), // number of seconds, must be multiple of 60
+				Stat:   aws.String("Sum"),
+				Unit:   aws.String(metrics.UnitCount),
+			},
+		})
+	}
+	zap.L().Warn("prepared metric queries", zap.Any("queries", queries), zap.Any("toDate", input.ToDate), zap.Any("fromDate", input.FromDate))
```
nhakmiller, line 193: Same comment as above, we use `Warn` for messages that indicate something that is not supposed to happen is happening, but it's not bad enough to stop processing (which would be an `Error`).

internal/core/metrics_api/api/alert_metrics.go
```diff
@@ -151,3 +156,73 @@ func getTotalAlertsDelta(input *models.GetMetricsInput, output *models.GetMetric
...
+
+	metricData, err := getMetricData(input, queries)
+	if err != nil {
+		return err
```
nhakmiller, line 197: We probably want to log this error and then return a generic internal server error like on line 175. If you copied this from an existing metric, we should probably update it there as well.

internal/core/metrics_api/api/alert_metrics.go
```diff
@@ -151,3 +156,73 @@ func getTotalAlertsDelta(input *models.GetMetricsInput, output *models.GetMetric
...
+	}
+
+	values, timestamps := normalizeTimeStamps(input, metricData)
+	// By default, return top 10 entries
+	limit := 10
```
nhakmiller, line 202: Each metric that gets generated in a single request shares the input struct, where `limit` is defined. In general, do we think that the limit should default to 10? If so, it may be worthwhile to set this default value higher up in the call stack (for example in `GetMetrics` near where we set the default namespace) that way other metrics that need to support the `limit` functionality don't need to remember to set the default value to 10.

This will also still allow individual metrics to overwrite the default at their discretion, if they waant.

internal/core/metrics_api/api/alert_metrics.go
```diff
@@ -151,3 +156,73 @@ func getTotalAlertsDelta(input *models.GetMetricsInput, output *models.GetMetric
...
+	return nil
+}
+
+func totalValue(values []*float64) float64 {
+	result := 0.0
+	for _, v := range values {
+		result += *v
+	}
+	return result
+}
```
nhakmiller, line 228: This could probably live in a helper file somewhere, in the same package but just some file called `helpers.go` or something.

internal/log_analysis/alert_forwarder/forwarder/metrics.go
```diff
@@ -0,0 +1,59 @@
...
+		},
+	}, []metrics.Metric{ // metric name and unit used
+		{
+			Name: "AlertsTriggered",
+			Unit: metrics.UnitCount,
+		},
+	})
+	RuleIDDimension = metrics.Dimension{
+		Name: "RuleID",
+	}
```
nhakmiller, line 58: So I haven't mucked around with these metrics in a minute, but I _think_ you can re-use the existing metric logger and just add a new dimension, then in the applicable metric generators (in the metrics API) you may just be able to specify which dimension sets you care about for a given query. Could be wrong on that, but worth a look.

nhakmiller (commented):
I see this is still in draft, but thought I'd drop a few early comments in!

internal/core/metrics_api/api/alert_metrics.go
lindsey-w: I did use another method as an example when putting this together - but I think we aren't logging right here because the `getMetricData` does the error logging for us (if there is a need) as well as returning `metricsInternalError` or `metricsNodataError`.  Or should we also log something here / return a generic internal error here?

internal/log_analysis/alert_forwarder/forwarder/metrics.go
lindsey-w: Apparently I just needed to tell you I couldn't get this to work - for it to start working. :) Let me know if my latest update isn't what you were expecting. It did seem to simply things quite a bit.

internal/core/metrics_api/api/alert_metrics.go
nhakmiller: Ah you're right, yes returning the error as is there is good 👍

internal/log_analysis/alert_forwarder/forwarder/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import (
+	"github.com/panther-labs/panther/pkg/metrics"
+)
+
+var (
+	StaticLogger = metrics.MustStaticLogger([]metrics.DimensionSet{
```
nhakmiller, line 26: This looks great! Exactly what I had in mind for this part.

internal/core/metrics_api/api/helpers.go
```diff
@@ -0,0 +1,27 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+func totalValue(values []*float64) float64 {
```
nhakmiller, line 21: I know it probably seems very trivial, but ideally we could add one or two unit tests just around this function since it is a standalone function that is easily testable.

internal/core/metrics_api/api/get_metrics.go
```diff
@@ -46,6 +46,7 @@ var (
 		"eventsLatency":    getEventsLatency,
 		"eventsProcessed":  getEventsProcessed,
 		"totalAlertsDelta": getTotalAlertsDelta,
+		"alertsByRuleID":   getAlertsByRuleID,
```
nhakmiller, line 49: nit: lets alphabetize these.

internal/core/metrics_api/api/alert_metrics.go
```diff
@@ -151,3 +155,66 @@ func getTotalAlertsDelta(input *models.GetMetricsInput, output *models.GetMetric
 
 	return nil
 }
+
+// getAlertsByRuleID returns the total count of alerts in the given time period
+//
+// This is a time series metric.
```
nhakmiller, line 161: Hey @3nvi, @alexmylonas, @vorillaz, can we get confirmation on whether we need this metric to return time series data (e.g. a list of values with corresponding timestamps) or just a single value for each rule?

internal/core/metrics_api/api/alert_metrics.go
```diff
@@ -151,3 +155,66 @@ func getTotalAlertsDelta(input *models.GetMetricsInput, output *models.GetMetric
...
+	}
+
+	values, timestamps := normalizeTimeStamps(input, metricData)
+	// prevent index out of bounds errors
+	if input.Limit > len(values) {
+		input.Limit = len(values)
+	}
+	// Sort the results in order of total number of alerts over the given time period
+	sort.Slice(values[:], func(i, j int) bool {
+		return totalValue(values[i].Values) > totalValue(values[j].Values)
```
nhakmiller, line 211: I don't know if it is worth optimizing this, but won't this re-compute the `totalValue` potentially many times for the same metric? We could remember what was computed and cache it keyed by ruleID.

Realistically even with hundreds of rules & hundreds of data points this will just take a few nanaseconds, but it may be worth benchmarking something. @austinbyers has done some go benchmarking before, not sure if he thinks its worth our time to try to optimize it but if so he could give some pointers on benchmarking.

nhakmiller (commented):
The code looks good and I would normally approve, but I want to get confirmation from the frontend end team about whether this should be time series data or not. @3nvi if we could get just a yes/no on that I would really appreciate it.

nhakmiller (approved):
@alexmylonas confirmed that time series data is preferred.

alexmylonas (changes requested):
Talked with @lindsey-w about how we will need this response to be single value instead of timeseries as it's more appropriate for this metric.

Also discussed of renaming the `limit` field to something more explicit for this metric.

Sorry again for not identifying this earlier!

lindsey-w:
I updated this endpoint to return single value metrics.  

Regarding the name change for the `limit` option.  I was thinking that limiting the number of results would be useful for more than just this `alertsByRuleID` metric.  (e.g. only return top N ‘chatty’ logtypes for the events processed metric).  So it seems there could be a couple of options:

1. It only applies to the `alertsByRuleID` and only requires a simple name change
2. If this applies to more than just the `alertsByRuleID` metric, the re-naming/re-evaluating how it would be added to all the relevant endpoints should be done at the same time (in another issue)
3. Potentially add `options` to the GetMetricsInput where we could define different structs for options on a per-metric basis.  

For example:
```
{
    "GetMetrics": {
        "fromDate": "2020-09-20T23:55:00Z",
        "toDate": "2020-09-21T23:55:00Z",
        "metricNames": [
            "alertsByRuleID"
        ],
        "intervalMinutes": 1440, 
         “options”: { 
              “alertsByRuleIDOptions”: { 
                  “limit”:1,
                  ”order”:”ascending”
               }
          } 
    }
}
```
Or 
```
{
    "GetMetrics": {
        "fromDate": "2020-09-20T23:55:00Z",
        "toDate": "2020-09-21T23:55:00Z",
        "metricNames": [
            "alertsByRuleID"
        ],
        "intervalMinutes": 1440, 
        “alertsByRuleIDOptions”: { 
             “limit”:1,
              ”order”:”ascending”
         } 
    }
}
```

Either way, since this issue originally said "optionally return only a fixed number of results (10 or 20) since the chart will only showcase 5 or 10” - I went ahead and removed the limit option for now. Does that work?

alexmylonas (approved):
Thanks @lindsey-w for the change, solid work! I agree we can re-visit the options field in the future.
Both solution seems acceptable to me.

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -125,14 +103,20 @@ func (h *Handler) handleNewAlert(rule *ruleModel.Rule, event *AlertDedupEvent) e
...
+		logStats(rule)
 	}
 	return err
 }
 
+func logStats(rule *ruleModel.Rule) {
+	// Severity, AnalysisType, and RuleID Information
+	StaticLogger.LogSingle(1,
+		metrics.Dimension{Name: "Severity", Value: string(rule.Severity)},
+		metrics.Dimension{Name: "RuleID", Value: string(rule.ID)},
```
kostaspap, line 115: @lindsey-w can you change this from `RuleID` to `AnalysisID`? 
We use the `AnalysisDimension` to differentiate between rules and policies. 
This will make it simpler to generate metrics for policies in the future. 
Someone can see the number of alerts for a rule by specifying as `AnalysisType`-> `Rule` and `AnalysisID`-> `RuleID`.

lindsey-w: Addressed here: https://github.com/panther-labs/panther/pull/1627

PR #1585, opened 2020-09-22
Add meta data options

nhakmiller:
## Background

Add the `EC2MetaDataOptions` field to the CloudSecurity EC2 Instance resource. Closes https://github.com/panther-labs/panther-analysis/issues/112. 

## Changes

- Add new field to resource model

## Testing

- Dev deployment

### DIFF
--- a/internal/compliance/snapshot_poller/models/aws/ec2_instance.go
+++ b/internal/compliance/snapshot_poller/models/aws/ec2_instance.go
@@ -51,6 +51,7 @@ type Ec2Instance struct {
 	KernelId                                *string
 	KeyName                                 *string
 	Licenses                                []*ec2.LicenseConfiguration
+	MetaDataOptions                         *ec2.InstanceMetadataOptionsResponse
 	Monitoring                              *ec2.Monitoring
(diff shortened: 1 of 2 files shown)


internal/compliance/snapshot_poller/models/aws/ec2_instance.go
```diff
@@ -51,6 +51,7 @@ type Ec2Instance struct {
 	KernelId                                *string
 	KeyName                                 *string
 	Licenses                                []*ec2.LicenseConfiguration
+	MetaDataOptions                         *ec2.InstanceMetadataOptionsResponse
```
maxbfigma, line 54: Might it be useful to use `MetadataOptions` here so it matches the output of `describe-instances` in case?

nhakmiller: Yeah seems reasonable.

austinbyers (approved):

internal/compliance/snapshot_poller/models/aws/ec2_instance.go
maxbfigma: Thanks!

PR #1587, opened 2020-09-22
Ensure master template is published with same version tag as repo

rleighton:
## Background

The Panther version is applied manually the master.yml template before a release.

This PR validates that the applied version tag matches the repo git tag.

I did not test this, but it looks right and we can test with 1.9 release.

### DIFF
--- a/tools/mage/master/publish.go
+++ b/tools/mage/master/publish.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
 	"github.com/panther-labs/panther/pkg/awsutils"
@@ -46,6 +47,13 @@ func Publish() error {
 	version, err := GetVersion()
 	if err != nil {
 		return err
+	}
+
+	// ensure the compiled version matches the declared version in master template
+	repoVersion := util.RepoVersion()
+	if version != repoVersion {
+		return errors.Errorf("master panther version %s does not match repo version %s",
+			version, repoVersion)
 	}
 
 	if err = getPublicationApproval(log, version); err != nil {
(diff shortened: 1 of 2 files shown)


tools/mage/master/publish.go
```diff
@@ -48,6 +49,13 @@ func Publish() error {
 		return err
 	}
 
+	// ensure the compiled version matches the declared version in master template
+	repoVersion := util.RepoVersion()
+	if version != repoVersion {
+		return errors.Errorf("master panther version %s does not match repo version %s",
```
austinbyers, line 55: We actually don't use the `errors` package anywhere in mage today, because errors are just logged / shown to the user directly.

I would just return `fmt.Errorf` for consistency, but if you feel we should always use the `errors` package, that's fine with me too. No strong preference

tools/mage/master/publish.go
```diff
@@ -48,6 +49,13 @@ func Publish() error {
 		return err
 	}
 
+	// ensure the compiled version matches the declared version in master template
```
austinbyers, line 52:
```suggestion
	// ensure the git tag in the repo matches the declared version in master template
```

austinbyers (approved):
As part of this same change, then, can you bump the version in `deployments/master.yml` to `v1.9.0-RC1`? Since we're getting ready to test v1.9?

rleighton:
> As part of this same change, then, can you bump the version in `deployments/master.yml` to `v1.9.0-RC1`? Since we're getting ready to test v1.9?
> 
> EDIT: That will actually require at least one other change (the version passed to the `web` stack would no longer need an explicit `v` prefix), so I can do it later if you prefer

Made both changes.

tools/mage/master/publish.go
rleighton: Done

tools/mage/master/publish.go
rleighton: Done

PR #1599, opened 2020-09-23
Fix error logging for PutIntegration

giorgosp:
## Background

If an error happens when saving a new log source integration, the API responds with a generic error message but also tries to log the original error. The implementation had a bug and the generic error ended up being logged instead of the original one.

## Testing

- Manually observing logs and API response

### DIFF
--- a/internal/core/source_api/api/put_integration.go
+++ b/internal/core/source_api/api/put_integration.go
@@ -43,18 +43,14 @@ var (
 
 // PutIntegration adds a set of new integrations in a batch.
 func (api API) PutIntegration(input *models.PutIntegrationInput) (newIntegration *models.SourceIntegration, err error) {
-	defer func() {
-		if err != nil {
-			zap.L().Error("failed to put integration", zap.Error(err))
-		}
-	}()
-
 	if err := api.validateIntegration(input); err != nil {
+		zap.L().Error("failed to put integration", zap.Error(err))
 		return nil, err
 	}
 
 	// Filter out existing integrations
 	if err := api.integrationAlreadyExists(input); err != nil {
+		zap.L().Error("failed to put integration", zap.Error(err))
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


internal/core/source_api/api/put_integration.go
```diff
@@ -43,18 +43,14 @@ var (
 
 // PutIntegration adds a set of new integrations in a batch.
 func (api API) PutIntegration(input *models.PutIntegrationInput) (newIntegration *models.SourceIntegration, err error) {
-	defer func() {
-		if err != nil {
```
giorgosp, line 47: `err` here is overriden by the values used in the `return` statement.

internal/core/source_api/api/put_integration.go
```diff
@@ -66,26 +62,30 @@ func (api API) PutIntegration(input *models.PutIntegrationInput) (newIntegration
 	// First creating table - this action is idempotent. In case we succeed here and
 	// fail at a later stage, in case of retry this will succeed again.
 	if err = createTables(newIntegration); err != nil {
-		err = errors.Wrap(err, "failed to create Glue tables")
+		err = errors.Wrap(err, "Failed to create Glue tables")
```
kostaspap, line 65: Btw, you don't really need to set the err here. We were doing that for the defer statement to work properly. 
You could just combine lines 65 + 66. 

```
zap.L().Error("failed to create Glue tables", zap.Error(err))
```

internal/core/source_api/api/put_integration.go
```diff
@@ -66,26 +62,30 @@ func (api API) PutIntegration(input *models.PutIntegrationInput) (newIntegration
...
+		zap.L().Error("failed to put integration", zap.Error(err))
 		return nil, putIntegrationInternalError
 	}
 
 	// Try to setupExternalResources
 	if err := setupExternalResources(newIntegration); err != nil {
+		zap.L().Error("failed to put integration", zap.Error(err))
```
kostaspap, line 72: Could be "failed to setup external resource". 
Also, in line 73 we should be returning a `putIntegrationInternalError`

internal/core/source_api/api/put_integration.go
```diff
@@ -66,26 +62,30 @@ func (api API) PutIntegration(input *models.PutIntegrationInput) (newIntegration
...
 		return nil, err
 	}
 
 	// Write to DynamoDB
 	if err = dynamoClient.PutItem(item); err != nil {
 		err = errors.Wrap(err, "Failed to store source integration in DDB")
+		zap.L().Error("failed to put integration", zap.Error(err))
```
kostaspap, line 79: Similar to above. 
You can combine both lines to
```
zap.L().Error("failed to store source integration in DDB", zap.Error(err))
```

internal/core/source_api/api/put_integration.go
```diff
@@ -66,26 +62,30 @@ func (api API) PutIntegration(input *models.PutIntegrationInput) (newIntegration
...
 	}
 
 	if input.IntegrationType == models.IntegrationTypeAWSScan {
 		err = api.FullScan(&models.FullScanInput{Integrations: []*models.SourceIntegrationMetadata{&newIntegration.SourceIntegrationMetadata}})
 		if err != nil {
-			err = errors.Wrap(err, "failed to trigger scanning of resources")
-			return nil, putIntegrationInternalError
+			err := errors.Wrap(err, "failed to trigger scanning of resources")
+			zap.L().Error("failed to put integration", zap.Error(err))
+			return nil, err
```
kostaspap, line 88: 1. No need to set the err
2. You can combine lines 86 & 87
3. The error returned will be displayed to the FE, so should be `putIntegrationInternalError`.

kostaspap (approved):
Nice!

internal/core/source_api/api/put_integration.go
giorgosp: Yeap, good point!

internal/core/source_api/api/put_integration.go
giorgosp: I wasn't sure if this was on purpose or something. Fixed

internal/core/source_api/api/put_integration.go
giorgosp: Fixed

PR #1605, opened 2020-09-24
Fixing few issues with Glue Sync tool

kostaspap:
## Background

I found a couple of issues while testing something locally. The Glue Sync will be replaced with a new tool soon, however the tool will not make it for 1.9.

## Changes

- Fixing issue where gluefix would not fix tables added through SQS source. 

## Testing

- Deployed and verified manually.

### DIFF
--- a/api/lambda/source/models/integration.go
+++ b/api/lambda/source/models/integration.go
@@ -74,7 +74,14 @@ func (info *SourceIntegration) RequiredLogTypes() (logTypes []string) {
 }
 
 func (info *SourceIntegration) IsLogAnalysisIntegration() bool {
-	return info.IntegrationType != IntegrationTypeAWSScan
+	switch integType := info.IntegrationType; integType {
+	case IntegrationTypeAWSScan:
+		return false
+	case IntegrationTypeAWS3, IntegrationTypeSqs:
+		return true
+	default:
+		panic("Unexpected integration type " + integType)
+	}
 }
 
 type SourceIntegrationHealth struct {
--- a/api/lambda/source/models/integration.go
+++ b/api/lambda/source/models/integration.go
@@ -73,6 +73,10 @@ func (info *SourceIntegration) RequiredLogTypes() (logTypes []string) {
 	}
 }
 
+func (info *SourceIntegration) IsLogAnalysisIntegration() bool {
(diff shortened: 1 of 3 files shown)


api/lambda/source/models/integration.go
```diff
@@ -73,6 +73,10 @@ func (info *SourceIntegration) RequiredLogTypes() (logTypes []string) {
 	}
 }
 
+func (info *SourceIntegration) IsLogAnalysisIntegration() bool {
+	return info.IntegrationType != IntegrationTypeAWSScan
```
rleighton, line 77: That looks wrong, shouldn't it be `models.IntegrationTypeAWS3`

kostaspap: Anything that is not `IntegrationTypeAWSScan` is log analysis. I have a `!=` condition, not equality ^^

rleighton: Could that bite us later when we add one that does not apply? I would prefer to be explicit unless you think that can't happen (if so add a comment to that effect).

rleighton (approved):

api/lambda/source/models/integration.go
kostaspap: It might be if we add an integration for Cloud Security that is of a new type. It might happen in the distant future. 
I thought of updating the signature to return (bool, err) and I thought it didn't look appropriate - I don't really want to force the callers to handle the error case. 
Should I make it panic?

rleighton: Why not: 
```
var logAnaysisIntegrationTypes = map[string]struct{} {
  IntegrationTypeAWS3: struct{}{},
  IntegrationTypeSqs: struct{}{},
}

func (info *SourceIntegration) IsLogAnalysisIntegration() (found bool) {
  _, found := logAnaysisIntegrationTypes[info.IntegrationType]
  return found
}
```

kostaspap: Well, this has the same problem as the existing code - the previous might return wrong result if a new cloud sec integration is added. The sample will return wrong result if a new log analysis integration is added.

rleighton: Yeah, then panic

kostaspap: Panicking now.

PR #1611, opened 2020-09-25
Fix warnings in FE tests

3nvi:
## Background

This PR fixes some warnings on the FE tests caused by wrong initial values on some forms. While there, it was discovered that initial values were also wrongly passed to the actual forms during destination creation, so this was fixed as well

## Changes

- Add proper  `outputConfig` initial values in all `XXXDestinationForm` tests
- change all `null` values in `CreateDestination` initial values and replaced them with `''`

## Testing

- `npm run test` passes without any warnings


alexmylonas (approved):
LGTM! Ran tests, no warnings.

PR #1615, opened 2020-09-26
Showcase: GraphQL Server

3nvi:
## Background

This PR adds support for a GraphQL server in go that can eventually replace AppSync. This is a WIP, but it showcases how the general approach **might** be.  

This is purely here for showcasing purposes and as a complementary  item to the related RFC.

More information can be found [here](https://gqlgen.com/getting-started/)

## Changes

- Add `graph.Dockerfile` for the server to be able to run in ECS and/or in local docker (some AWS Envs are missing)
- Add authentication
- Add `@aws_auth` authorization directive
- Add dummy resolvers for all of our schema
- Add a `server.go` that can boot up the  GraphQL server

## Does it actually run?

The project itself relies on **real** Cognito authentication, so to bypass it simply remove the `auth.Middleware` found in `server.go`. If you don't do that, you'll see auth-related errors.

To  run the project having removed the authorization middleware, type `go run server.go`

## Does it work?

It does, but the resolver code hasn't been added intentionally. What we'd need to do is to fill in each resolver with the lambda function call that each AppSync resolver already calls.


internal/core/graph_api/auth/directives.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+
+	"github.com/99designs/gqlgen/graphql"
+	"github.com/juliangruber/go-intersect"
+	"github.com/vektah/gqlparser/v2/gqlerror"
+)
+
+// nolint:golint,stylecheck
+// The implementation of the `@aws_auth` GraphQL directive
+func AwsAuth(ctx context.Context, _ interface{}, next graphql.Resolver, cognito_groups []string) (interface{}, error) {
```
austinbyers, line 31: Awesome! Makes sense

internal/core/graph_api/auth/middleware.go
```diff
@@ -0,0 +1,131 @@
...
+var (
+	// Cognito details
+	awsRegion   = os.Getenv("AWS_REGION")
+	userPoolID  = os.Getenv("WEB_APPLICATION_USER_POOL_ID")
+	appClientID = os.Getenv("WEB_APPLICATION_USER_POOL_CLIENT_ID")
+
+	// A static list of JSON Web Key Sets that contain the public RSA keys for validating the associated JWTs. It should
+	// be read once and stored in memory since the RSA keys never change
+	// FIXME: handle error somehow
+	keySet, _ = jwk.Fetch(fmt.Sprintf("https://cognito-idp.%s.amazonaws.com/%s/.well-known/jwks.json", awsRegion, userPoolID))
```
austinbyers, line 49: I wondered how this part would work! Makes sense

internal/core/graph_api/main/server.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/99designs/gqlgen/graphql/playground"
+
+	"github.com/panther-labs/panther/internal/core/graph_api/auth"
+	"github.com/panther-labs/panther/internal/core/graph_api/generated"
+	"github.com/panther-labs/panther/internal/core/graph_api/resolvers"
+)
+
+const defaultPort = "8000"
+
+func main() {
```
austinbyers, line 36: This file is essentially the one handled by API gateway if we choose to use that instead of exposing a port directly on the load balancer. The rest of the code would be more or less the same

internal/core/graph_api/gqlgen.yml
```diff
@@ -0,0 +1,72 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+# Where are all the schema files located? globs are supported eg  src/**/*.graphqls
+schema:
+  - ../../../api/graphql/*.graphql
+
+# Where should the generated server code go?
+exec:
+  filename: generated/runtime.generated.go
+  package: generated
```
austinbyers, line 24: Wow, this is awesome! I love that the Go structs are generated from the GraphQL, good find 🎉

internal/core/graph_api/resolvers/schema.resolvers.go
```diff
@@ -0,0 +1,275 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/panther-labs/panther/internal/core/graph_api/generated"
+	"github.com/panther-labs/panther/internal/core/graph_api/models"
+)
+
+func (r *mutationResolver) AddDestination(ctx context.Context, input models.DestinationInput) (*models.Destination, error) {
+	panic(fmt.Errorf("not implemented"))
```
austinbyers, line 33: Where are the implementations supposed to be defined? The file header says "this file will be automatically regenerated based on the schema", so I'm guessing we don't put it here?

austinbyers (commented):
Great work! Sorry it's taken so long to look at it. There's an open question about whether we use API gateway or ELB to host the GraphQL "server", they each have their pros and cons.

But either way, the work you've done here to parse the JWT and validate the claims / implement RBAC is excellent and will essentially be the same in either implementation. I'm glad there are enough Go libraries to work with JWTs and GraphQL

austinbyers:
I'm impressed with your Go-fu @3nvi 😎

internal/core/graph_api/gqlgen.yml
3nvi: Yeah, there are currently 2 schools. Either golang structs define the GraphQL schema or the schema itself creates the structs. Exactly the same happens with swagger.

Ultimately, my approach is schema-first since I wanna have control over the API that gets shipped to clients rather than rely on autogeneration.

internal/core/graph_api/main/server.go
3nvi: Exactly.

internal/core/graph_api/resolvers/schema.resolvers.go
3nvi: We do put them here. The generation has the following logic:

1. If a resolver already exists, don't re-generate it
2. If a resolver doesn't exist, generate it.

Thus if we make changes to any resolver, these changes will not be overriden by the autogen. If our schema got updated and we want to force an existing resolver to be re-generated, we have to comment it out & run the generation again.

internal/core/graph_api/gqlgen.yml
austinbyers: Agreed

austinbyers:
Branch has been transferred to our enterprise repo

PR #1619, opened 2020-09-28
Refactor: simplify React code on Alerts page

3nvi:
## Background

Minor PR that simplifies the props passed to the Alert page Banner, to align with our future needs

## Changes

- Stop passing  `rule` to the banner element
- Utilize `alert` for the same information

## Testing

- CI


vorillaz (approved):
That looks nice, just two minor comments:

1. A more concrete description could help us track down bugs, regression, and avoid duplicate issues. In cases like these, I would suggest adding some more context: What would you trying to achieve? Is there any design-related issue connected with this PR? Is PR some kind of prep work towards a new feature?
2. A simple snapshot test can help us create some sort of linear history for design or API changes.

3nvi:
> That looks nice, just two minor comments:
> 
> 1. A more concrete description could help us track down bugs, regression, and avoid duplicate issues. In cases like these, I would suggest adding some more context: What would you trying to achieve? Is there any design-related issue connected with this PR? Is PR some kind of prep work towards a new feature?
> 2. A simple snapshot test can help us create some sort of linear history for design or API changes.

1. Just cleaning up something that was noticed. You' re absolutely right about the description, it's just that it was not related to something in particular. Still tho, I should add a bit more context.

2. 100%  on this. Will add a snapshot test with the previous implementation and compare it against these changes

PR #1621, opened 2020-09-28
Renaming 'errorType' to 'alertType'

kostaspap:
## Background

Every item in the alert table contains an attribute named 'errorType'. This attribute was meant to help us differentiate between rule errors (coming in next release) and log classifications errors (coming in the future). 

However, we are looking at using the existing alert system for more operations in the future e.g. system errors, policy failures etc. 
Thinking that naming it just 'alertType' is more generic and easy to extend. 

## Changes

- Renaming `errorType` to `alertType`
- Using the exception message as the alert title in case of an exception. 

## Testing

- mage test:ci


internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -75,7 +75,7 @@ func shouldIgnoreChange(rule *ruleModel.Rule, alertDedupEvent *AlertDedupEvent)
 	// If the number of matched events hasn't crossed the threshold for the rule, don't create a new alert.
 	hasBreachedThreshold := alertDedupEvent.EventCount < int64(rule.Threshold)
 	// If the alertDedupEvent was generated by a rule-engine error
-	isError := alertDedupEvent.ErrorType != nil
+	isError := aws.StringValue(alertDedupEvent.Type) == RuleErrorType
```
kostaspap, line 78: This logic has been changes slightly in the feature branch. Instead of backporting the change here, i will change update the logic in the feature branch @giorgosp

giorgosp (approved):

PR #1623, opened 2020-09-28
Switch to using "Alert Type" in the alert rather than "error Type"

kostaspap:
## Background

We were using "Type" to define if an alert if because of a Rule or Policy and an "ErrorType" to differentiate between "Rule Error" and "Rule matches". Seems better approach to extend the Alert Type to be one of "Policy"(failure), "Rule"(match), "Rule Error". This could make it easier to use the same structure for other types that are not necessarily errors

## Changes

- Changed Alert API to handle the new type
- Changed Alert Delivery to handle the new type
- Changed rules engine to populate handle the new type

## Testing

- mage test:ci
- Deployed and generated a rule that hits an error. Seeing it in the UI: 

Screen Shot 2020-09-28 at 6 31 03 PM

### DIFF
--- a/internal/core/alert_delivery/outputs/outputs.go
+++ b/internal/core/alert_delivery/outputs/outputs.go
@@ -27,8 +27,8 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 
-	alertmodels "github.com/panther-labs/panther/api/lambda/delivery/models"
-	outputmodels "github.com/panther-labs/panther/api/lambda/outputs/models"
+	alertModels "github.com/panther-labs/panther/api/lambda/delivery/models"
+	outputModels "github.com/panther-labs/panther/api/lambda/outputs/models"
 	"github.com/panther-labs/panther/pkg/gatewayapi"
 )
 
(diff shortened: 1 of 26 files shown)


internal/core/alert_delivery/outputs/outputs.go
```diff
@@ -179,13 +185,19 @@ func generateAlertTitle(alert *alertModels.Alert) string {
 	if alert.IsResent {
 		return "[Re-sent]: " + *alert.Title
 	}
-	if alert.Title != nil {
-		return "New Alert: " + *alert.Title
-	}
-	if alert.Type == alertModels.RuleType {
+	switch alert.Type {
```
kostaspap, line 188: @s0l0ist line 186 seems like a bug? What happens if the title is nil?

s0l0ist: @kostaspap , This could be a bug if we're storing an alert in DDB without a title. I thought the forwarder accounted for this and always stored a valid title field in DDB.

If this is not correct, then this is indeed a bug.

internal/core/alert_delivery/outputs/outputs.go
```diff
@@ -158,10 +158,16 @@ func generateNotificationFromAlert(alert *alertModels.Alert) Notification {
...
+	case alertModels.RuleType:
 		return getDisplayName(alert) + " triggered"
+	case alertModels.RuleErrorType:
+		return getDisplayName(alert) + " encountered an error"
+	case alertModels.PolicyType:
+		return getDisplayName(alert) + " failed on new resources"
+	default:
+		panic("uknown alert type " + alert.Type)
```
giorgosp, line 169: +1 for the panic here since this must/will never be reached in production. Returning an error would just complicate things.

internal/core/alert_delivery/outputs/outputs.go
kostaspap: ok will do. 
Yeah, there is no guarantee that the Title is going to be present - it only gets populated if the users specified a "title" function.

giorgosp (approved):

PR #1625, opened 2020-09-28
Include title() & dedup() output in API response when testing a Rule

### LINKED ISSUE #1449: Rule Testing: Show Dedup and Title Output (Backend)
Please see #983 and #1447 for more context.

giorgosp:
Fixes #1449

## Background

Implement backend changes so that we can display `title()` and `dedup()` output when testing a rule.

## Changes

- Refactored the response of the rule engine to facilitate the inclusion of those fields
- The core api response for testing a rule is now different than testing a policy, so split the http `/test` route into two,
with different responses (but same input).

## Testing

- unit, integration tests, manual QA

The implementation is almost ready, will move the PR out of draft when I complete the testing and fix any bugs.
We can discuss the implementation now though.


internal/core/analysis_api/analysis/rule_engine.go
```diff
@@ -66,33 +65,42 @@ func (e *RuleEngine) TestRule(rule *models.TestPolicy) (models.TestPolicyResult,
...
+		} else {
+			testResult.TestsFailed = append(testResult.TestsFailed, string(test.Name))
+		}
 	}
-	return makeTestSummary(rule, output)
+	testResult.TestSummary = len(testResult.TestsFailed) == 0 && len(testResult.TestsErrored) == 0
+	return testResult, nil
 }
+
+const testRuleID = "RuleAPITestRule"
```
kostaspap, line 106: Better define it closer to where it is called.

internal/core/analysis_api/handlers/test_policy.go
```diff
@@ -29,14 +29,14 @@ import (
 	"github.com/panther-labs/panther/pkg/gatewayapi"
 )
 
-// TestPolicy runs a policy (or rule) against a set of unit tests.
-func TestPolicy(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyResponse {
+// TestPolicyOrRule runs a policy (or rule) against a set of unit tests.
+func TestPolicyOrRule(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyResponse {
```
kostaspap, line 33: Seems we could have two different functions? (`TestPolicy` & `TestRule`)

That would help avoid all these if/else blocks.

giorgosp: We could but the function is very trivial. The only `if` that would be split is this:
```
var testResults interface{}
	if input.AnalysisType == models.AnalysisTypeRULE {
		testResults, err = ruleEngine.TestRule(input)
	} else {
		testResults, err = policyEngine.TestPolicy(input)
	}
```
We have different endpoints though.

internal/core/analysis_api/analysis/rule_engine.go
```diff
@@ -66,33 +65,42 @@ func (e *RuleEngine) TestRule(rule *models.TestPolicy) (models.TestPolicyResult,
...
+				ID:       testRuleID, // doesn't matter as we're only running one rule
 				LogTypes: rule.ResourceTypes,
 			},
 		},
 		Events: inputEvents,
 	}
 
 	// Send the request to the rule-engine
-	var engineOutput enginemodels.RulesEngineOutput
+	var engineOutput enginemodels.RuleEngineOutput
```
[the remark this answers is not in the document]
giorgosp: Aha, I see, we call the python package `rules_engine`. 
Btw, maybe we rename to `rule_engine`? To be consistent with `policy_engine`

internal/core/analysis_api/analysis/rule_engine.go
```diff
@@ -44,16 +45,14 @@ func NewRuleEngine(lambdaClient lambdaiface.LambdaAPI, lambdaName string) RuleEn
...
-
+func (e *RuleEngine) TestRule(rule *models.TestPolicy) (*models.TestRuleResult, error) {
 	// Build the list of events to run the rule against
 	inputEvents := make([]enginemodels.Event, len(rule.Tests))
 	for i, test := range rule.Tests {
 		var attrs map[string]interface{}
 		if err := jsoniter.UnmarshalFromString(string(test.Resource), &attrs); err != nil {
 			//nolint // Error is capitalized because will be returned to the UI
-			return empty, &TestInputError{fmt.Errorf(`Event for test "%s" is not valid json: %w`, test.Name, err)}
+			return nil, &TestInputError{fmt.Errorf(`Event for test "%s" is not valid json: %w`, test.Name, err)}
```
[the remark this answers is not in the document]
giorgosp: No It was just a micro-optimization but returning nil and moving the return variable declaration closer to its usage is a bit cleaner.

internal/core/analysis_api/analysis/rule_engine.go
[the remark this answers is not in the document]
kostaspap: That's a big renaming ^^ This includes renaming the python packages but also the AWS Lambda. I would leave that out for a separate PR

internal/core/analysis_api/analysis/policy_engine.go
```diff
@@ -79,7 +79,11 @@ func (e *PolicyEngine) TestPolicy(policy *models.TestPolicy) (models.TestPolicyR
 }
 
 func makeTestSummary(policy *models.TestPolicy, engineOutput enginemodels.PolicyEngineOutput) (models.TestPolicyResult, error) {
-	var testResults models.TestPolicyResult
+	testResults := models.TestPolicyResult{
+		TestsErrored: models.TestsErrored{},
+		TestsFailed:  models.TestsFailed{},
+		TestsPassed:  models.TestsPassed{},
```
giorgosp, line 85: These initializations seem to just satisfy the integration tests. Preferred it over changing every test.

internal/core/analysis_api/analysis/rule_engine.go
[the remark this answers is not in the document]
giorgosp: Shall we keep`RuleEngineOutput` for now?

internal/core/analysis_api/analysis/rule_engine.go
giorgosp: It is called in the above function. Do you suggest we just make it a local variable?

Also, I didn't put it at the top of the file since it is unexported. When reading a new file, seems cleaner to have the exported things at the top, not a strong opinion though. Just saying.

internal/core/analysis_api/analysis/rule_engine.go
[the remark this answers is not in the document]
kostaspap: It's not a big deal, but pointing out that the service is called `Rules Engine` so seems it is more appropriate to be named `RulesEngineOutput` (what it was named before). 
If it's difficult to change, leave it be, if its not I don't see why not do it. When we rename the `Rules Engine` to `Rule Engine` we can change this one as well.

internal/core/analysis_api/analysis/rule_engine.go
kostaspap: I agree that it doesn't have to be in the top. My suggestion was to make it a local constant (move it around line 63?)

internal/core/analysis_api/handlers/test_policy.go
kostaspap: Sure. A suggestion would be to name it `TestAnalysis` . We use the term "analysis" to refer to rule or policy.

kostaspap (approved):
LGTM. Feel free to merge after the comments have been addressed. 
I assume this breaks existing system - we need to create a feature branch? 

Have you verified that integration tests pass?

internal/core/analysis_api/analysis/rule_engine.go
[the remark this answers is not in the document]
giorgosp: Ok reverted

internal/core/analysis_api/analysis/rule_engine.go
```diff
@@ -66,33 +65,42 @@ func (e *RuleEngine) TestRule(rule *models.TestPolicy) (models.TestPolicyResult,
...
 
 	// Translate rule engine output to test results.
-	testResults, err := makeTestSummaryRule(rule, engineOutput)
-	return testResults, err
-}
+	testResult := &models.TestRuleResult{}
+	for _, result := range engineOutput.Results {
+		// Determine which test case this result corresponds to. We constructed resourceID with the
+		// format Panther:Test:Resource:TestNumber (see testResourceID),
+		testIndex, err := strconv.Atoi(strings.Split(result.ID, ":")[3])
```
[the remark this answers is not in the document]
giorgosp: Done

internal/core/analysis_api/analysis/rule_engine.go
giorgosp: Done

internal/core/analysis_api/handlers/test_policy.go
giorgosp: Done

internal/core/analysis_api/analysis/rule_engine.go
giorgosp: Basically leaving this here so that we can reuse elsewhere (tests mainly)

PR #1627, opened 2020-09-29
most active rule api - metric rename

lindsey-w:
## Background

New PR to address Metric Name, as requested here: https://github.com/panther-labs/panther/pull/1584#discussion_r495733335 

## Changes

- Update to dimensions for alert logging.  
- Each metric will now have assocaited `AnalsysisType` and `AnalysisID` to address being able to select `Rule|Policy` and `RuleID|PolicyID` in the future. 
- Updated testing and api to return appropriate output 

## Testing

- `mage test:integration`
- manual testing
Input / Output remains unchanged when accessing the metrics api:

Input:
```
{
    "GetMetrics": {
        "fromDate": "2020-09-17T17:50:00Z",
        "intervalMinutes": 14,
        "limit": 1,
        "metricNames": [
            "alertsByRuleID"
        ],
        "toDate": "2020-09-29T17:50:00Z"
    }
}
```

Output:
```
{
    "alertsByRuleID": {
        "seriesData": {
            "series": [],
            "timestamps": []
        },
        "singleValue": [
            {
                "label": "AWS.S3.ServiceAccess.UnknownRequester",
                "value": 81
            },
            {
                "label": "AWS.IAM.PolicyModified",
                "value": 81
            },
            {
                "label": "AWS.EC2.RouteTableModified",
                "value": 23
            }
        ]
    },
    "fromDate": "2020-09-17T17:50:00Z",
    "intervalMinutes": 14,
    "toDate": "2020-09-29T17:50:00Z"
}
```


kostaspap (approved):

lindsey-w:
@kostaspap  Do you think the api endpoint should also reflect analysisID? e.g. `getAlertsByAnalsysisID`? The output does not reflect any context around `analysisType` and it currently only returns RuleIDs, but wondering if I should I have updated that as well?

kostaspap:
> @kostaspap Do you think the api endpoint should also reflect analysisID? e.g. `alertsByAnalsysisID` in the input above? The output does not reflect any context around `analysisType` and it currently only returns RuleIDs, but wondering if I should I have updated that as well?

I think it's ok as it is. The FE will likely always want to get metrics for Rules or Policies (whenever that happens) so I feel we will have separate methods there.

nhakmiller (approved):
Just a reminder, we can't rename old metrics and maintain their values. Since this hasn't gone to a stable release yet this is fine, and its good we're doing this change now, but if we ever want to change this metric name again in the future we will lose all history of the previously named metric.

PR #1628, opened 2020-09-29
Support for AWS EKS Scanning

wey-chiang:
Added support for AWS EKS Scanning. 

## Changes

Added eks_cluster.go to 
internal/compliance/snapshot_poller/models/aws/
internal/compliance/snapshot_poller/pollers/aws/

Refactored parts of the ECS code to avoid naming conflicts

## Testing

Ongoing -- this PR will be updated with unit tests.


internal/compliance/snapshot_poller/models/aws/eks_cluster.go
```diff
@@ -0,0 +1,97 @@
...
+	// Additional fields
+	NodeGroup      []*EksNodegroup
+	FargateProfile []*EksFargateProfile
+}
+
+// EksNodegroup contains all the information about an EKS Service, for embedding into the EksCluster resource
+type EksNodegroup struct {
+	// Generic resource fields
+	//
+	// This is not a full resource, but it does have an ARN, Tags, and a name.
```
austinbyers, line 57: Copy-pasta comment? Looks like a full resource to me

austinbyers (commented):
Looks great to me! Deferring final approval to @nhakmiller 

Did you actually deploy and see the resources registered in Panther? Unit tests aren't enough to be confident this works as you expect

api/lambda/source/models/vars.go
```diff
@@ -25,6 +25,8 @@ const (
 	IntegrationTypeAWS3 = "aws-s3"
 	// IntegrationTypeSqs is integration type for pulling data from an SQS queue.
 	IntegrationTypeSqs = "aws-sqs"
+	// IntegrationTypeEks is integration type for importing data from customer EKS clusters
+	IntegrationEks = "aws-eks"
```
nhakmiller, line 29: EKS scanning is a subset of the `aws-scan` integration type, I don't think we need this here.

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,310 @@
...
+	return len(*clusters) < defaultBatchSize
+}
+
+// describeEKSCluster provides detailed information for a given EKS cluster
+func describeEKSCluster(eksSvc eksiface.EKSAPI, clusterName *string) (*eks.Cluster, error) {
+	out, err := eksSvc.DescribeCluster(&eks.DescribeClusterInput{
+		Name: clusterName,
+	})
+	if err != nil {
+		return nil, errors.Wrapf(err, "EKS.DescribeClusters: %s", aws.StringValue(clusterName))
```
nhakmiller, line 104: nit: generally I try to do the name of the API call as the error message, so in this case `EKS.DescribeCluster: %s`.

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,310 @@
...
+	// If there are no fargate profiles stop here
+	if len(fargateProfileNames) == 0 {
+		return nil, nil
+	}
+
+	// Describe fargate profiles
+	fargateProfiles := make([]*awsmodels.EksFargateProfile, 0, len(fargateProfileNames))
+
+	// Slightly different from the ECS Clusters since we have to make a call to describe each FargateProfile
+	// rather than batching them all together.
```
nhakmiller, line 136: I think these comments comparing to ECS cluster scanning are useful for explaining what is going on, but I don't  think the direct comparison to ECS cluster scanning is necessary. If we changed how ECS cluster scanning works, for example, these comments would be outdated unless we remembered to update them as well.

Instead, I think just standalone comments explaining what's going on such as:
```
// We can only describe one profile at a time, so we iterate through each one returned by the ListFargateProfiles API call and describe it
```
would work better.

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,310 @@
...
+	// Enumerate tasks
+	var fargateProfileNames []*string
+	err := eksSvc.ListFargateProfilesPages(&eks.ListFargateProfilesInput{ClusterName: clusterName},
+		func(page *eks.ListFargateProfilesOutput, lastPage bool) bool {
+			fargateProfileNames = append(fargateProfileNames, page.FargateProfileNames...)
+			return true
+		})
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "EKS.ListTasksPages: %s", aws.StringValue(clusterName))
```
nhakmiller, line 124: This should probably be `EKS.ListFargateProfilesPages`.

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	// Unlike describeECSCluster, we can only make an API call to describe a single cluster
+	// Thus, we do not need to check then length of the returned output
+
+	return out.Cluster, nil
+}
+
+// getEKSFargateProfile enumerates and then describes all Fargate Profiles of a cluster
+func getEKSFargateProfile(eksSvc eksiface.EKSAPI, clusterName *string) ([]*awsmodels.EksFargateProfile, error) {
```
nhakmiller, line 114: nit: since we're returning multiple profiles, I would name this `getEKSFargateProfiles`

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+// buildEksClusterSnapshot returns a complete snapshot of an EKS cluster
+func buildEksClusterSnapshot(eksSvc eksiface.EKSAPI, clusterName *string) (*awsmodels.EksCluster, error) {
+	if clusterName == nil {
+		return nil, nil
+	}
+
+	details, err := describeEKSCluster(eksSvc, clusterName)
+	// Can details ever be nil without an error?
```
nhakmiller, line 235: I'm not sure if this comment was copied from an existing poller, but ideally we could try to answer this question for EKS for at least some basic cases (e.g. try making the API calls for an EKS cluster that does not exist and see if an empty value is returned without an error, as an example).

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,310 @@
...
+		return nil, err
+	}
+
+	return eksCluster, nil
+}
+
+// PollEksCluster gathers information on each EKS Cluster for an AWS account.
+func PollEksClusters(pollerInput *awsmodels.ResourcePollerInput) ([]*apimodels.AddResourceEntry, *string, error) {
+	zap.L().Debug("starting EKS Cluster resource poller")
+	eksClusterSnapshots := make(map[string]*awsmodels.EksCluster)
```
nhakmiller, line 278: I'm not sure I see what `eksClusterSnapshots` is being used for here, was this copied over from an old commit? This is sort of following the old way we used to do things, but doesn't have the end part which makes it relevant.

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -107,6 +108,7 @@ var (
 		awsmodels.Ec2VolumeSchema:           {"EC2Volume", PollEc2Volumes},
 		awsmodels.Ec2VpcSchema:              {"EC2VPC", PollEc2Vpcs},
 		awsmodels.EcsClusterSchema:          {"ECSCluster", PollEcsClusters},
+		awsmodels.EksClusterSchema:          {"EKSCluster", PollEcsClusters},
```
nhakmiller, line 111: This should be mapped to `PollEksClusters`.

nhakmiller (commented):
Looks like a pretty good first pass! A few minor comments, mostly around style/typos. Once there are unit tests and you've confirmed that deploying works (and you can see the resources in the web UI), I can give it another review.

internal/compliance/snapshot_poller/pollers/aws/poll.go
wey-chiang: Argh, copy pasta. Good catch!

internal/compliance/snapshot_poller/models/aws/eks_cluster.go
wey-chiang: Doh, yup!

api/lambda/source/models/vars.go
wey-chiang: Yup! Will remove

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
wey-chiang: Fixed

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
wey-chiang: Makes sense, will fix.

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
wey-chiang: Fixed

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
wey-chiang: Fixed

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
wey-chiang: Ran `aws eks describe-cluster --name Test-Cluster` to verify this works as intended
Then tried `aws eks describe-cluster --name test-cluster` (notice diff in case for cluster name), which returns
> An error occurred (ResourceNotFoundException) when calling the DescribeCluster operation: No cluster found for name: test-cluster.

So this should error out

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
wey-chiang: Yup, yeah I don't see any functional difference without the eksClusterSnapshots.
Will remove this line and change the array instantiation to be based on `len(clusters)` which is an array of cluster names returned by listEKSClusters

wey-chiang:
> Did you actually deploy and see the resources registered in Panther? Unit tests aren't enough to be confident this works as you expect

image

wey-chiang:
Tested the changes, and fixed issues with the unit tests.

Verified this works via deployment (see above pic)

This PR is ready to be merged

wey-chiang:
Related PR for updating the IAM permissions for the Panther Audit Role:
https://github.com/panther-labs/panther/pull/1653

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	// This only attempts to describe a single cluster so we do not need checks on length
+
+	return out.Cluster, nil
+}
+
+// getEKSFargateProfiles enumerates and then describes all Fargate Profiles of a cluster
+func getEKSFargateProfiles(eksSvc eksiface.EKSAPI, clusterName *string) ([]*awsmodels.EksFargateProfile, error) {
+	// Enumerate tasks
```
nhakmiller, line 114: nit: should this say `Enumerate profiles`?

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,309 @@
...
+			return nil, errors.Wrapf(err, "EKS.DescribeFargateProfile: %s", aws.StringValue(clusterName))
+		}
+
+		profile := *rawFargateProfile.FargateProfile
+
+		fargateProfiles = append(fargateProfiles, &awsmodels.EksFargateProfile{
+			GenericAWSResource: awsmodels.GenericAWSResource{
+				ARN:  profile.FargateProfileArn,
+				Tags: profile.Tags,
+			},
```
nhakmiller, line 152: In the `EksFargateProfile` struct, you include these two embedded structs:

```
	GenericAWSResource
	GenericResource
```

but here you only define `GenericAWSResource`. We should either fill in `GenericResource`, or remove it from the struct definition.

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,309 @@
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "EKS.DescribeNodegroup: %s", aws.StringValue(nodegroup))
+		}
+
+		curNodegroup := *rawNodegroup.Nodegroup
+		nodegroupResults = append(nodegroupResults, &awsmodels.EksNodegroup{
+			GenericAWSResource: awsmodels.GenericAWSResource{
+				ARN:  curNodegroup.NodegroupArn,
+				Tags: curNodegroup.Tags,
+			},
```
nhakmiller, line 204: Same comment here as above about using both `GenericAWSResource` and `GenericResource`.

internal/compliance/snapshot_poller/pollers/aws/eks_cluster_test.go
```diff
@@ -0,0 +1,165 @@
...
+	mockSvc := awstest.BuildMockEksSvc([]string{"DescribeCluster"})
+
+	out, err := describeEKSCluster(mockSvc, awstest.ExampleEksClusterName)
+	require.NoError(t, err)
+	assert.NotEmpty(t, out)
+}
+
+func TestEksClusterDescribeDoesNotExist(t *testing.T) {
+	mockSvc := &awstest.MockEks{}
+	//mockSvc := awstest.BuildMockEksSvcError([]string{"DescribeCluster"})
```
nhakmiller, line 84: Lets remove this comment.

internal/compliance/snapshot_poller/pollers/aws/eks_cluster_test.go
```diff
@@ -0,0 +1,165 @@
...
+	mockSvc.On("DescribeCluster", mock.Anything).
+		Return(
+			&eks.DescribeClusterOutput{
+				Cluster: nil,
```
nhakmiller, line 88: nit: `nil` is the default value so you don't need to have this line. In this case, I actually don't mind that much since it's a unit test so being explicit that we're testing for the case where `Cluster` is `nil` is fine, but just thought I'd point it out in case you weren't aware. Up to you if you want to keep it explicit like this or let it be implicit.

internal/compliance/snapshot_poller/pollers/aws/awstest/eks.go
```diff
@@ -0,0 +1,356 @@
...
+
+	args := m.Called(in)
+	if args.Error(0) != nil {
+		return args.Error(0)
+	}
+	paginationFunction(ExampleEksListClusters, true)
+	return args.Error(0)
+}
+
+func (m *MockEks) ListFargateProfilesPages(
```
nhakmiller, line 312: I don't see any unit tests for `listFargateProfiles` or `listNodeGroups` in `eks_cluster_test.go`, we should add those.

internal/compliance/snapshot_poller/pollers/aws/awstest/eks.go
```diff
@@ -0,0 +1,356 @@
...
+	if args.Error(0) != nil {
+		return args.Error(0)
+	}
+	// Return appropriate ListFargateProfile output based on input ClusterARN
+	if in.ClusterName == ExampleEksClusterName {
+		paginationFunction(ExampleListFargateProfilesMulti, true)
+		return args.Error(0)
+	}
+	paginationFunction(ExampleListFargateProfile, true)
+	return args.Error(0)
```
nhakmiller, line 327: Not sure if this is needed here unless you will be adding more unit tests where this conditional will not be true.  Right now this function mock isn't being used at all, but when you do use it I'm not sure if you will need to have this conditional.

nhakmiller (commented):
Looking good! A few comments.

internal/compliance/snapshot_poller/pollers/aws/awstest/eks.go
```diff
@@ -0,0 +1,356 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/eks"
+	"github.com/aws/aws-sdk-go/service/eks/eksiface"
+	"github.com/stretchr/testify/mock"
+)
+
+type MockCertificate struct {
+	Data string
+}
```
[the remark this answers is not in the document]
wey-chiang: Nope, I think that was leftover from testing -- not sure why the code inspection didn't catch that.
Removed!

internal/compliance/snapshot_poller/pollers/aws/awstest/eks.go
```diff
@@ -0,0 +1,356 @@
...
+			CertificateAuthority: &eks.Certificate{Data: aws.String("example-certificate-data")},
+			ClientRequestToken:   aws.String("example-client-request-token"),
+			CreatedAt:            ExampleCreatedAt,
+			EncryptionConfig:     ExampleEncryptionConfig,
+			Endpoint:             aws.String("example-endpoint"),
+			Identity:             &eks.Identity{Oidc: &eks.OIDC{Issuer: aws.String("example-oidc-issuer")}},
+			Logging:              &eks.Logging{ClusterLogging: ExampleLogSetup},
+			Name:                 ExampleEksClusterName,
+			PlatformVersion:      aws.String("example-cluster-platform-version"),
+			// ResourcesVpcConfig:   ExampleResourceVpcConfig[0],
```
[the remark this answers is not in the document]
wey-chiang: It's redefined in the line underneath it -- removed

internal/compliance/snapshot_poller/pollers/aws/awstest/eks.go
```diff
@@ -0,0 +1,356 @@
...
+				Return(nil)
+		},
+		"ListFargateProfiles": func(svc *MockEks) {
+			svc.On("ListFargateProfiles", mock.Anything).
+				Return(ExampleListFargateProfilesMulti, nil)
+		},
+		"ListFargateProfilesPages": func(svc *MockEks) {
+			svc.On("ListFargateProfilesPages", mock.Anything).
+				Return(nil)
+		},
```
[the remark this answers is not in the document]
wey-chiang: Removed the entry for `ListFargateProfiles`

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+	snapshot.Region = aws.String(parsedResourceID.Region)
+	snapshot.AccountID = aws.String(parsedResourceID.AccountID)
+
+	return snapshot, nil
+}
+
+// listEKSClusters returns all EKS clusters in the account
+func listEKSClusters(eksSvc eksiface.EKSAPI, nextMarker *string) (clusters []*string, marker *string, err error) {
+	zap.L().Debug("Started listEKSClusters")
```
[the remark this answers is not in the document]
wey-chiang: Removed

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
wey-chiang: Hahah doh, yeah.

internal/compliance/snapshot_poller/pollers/aws/eks_cluster_test.go
wey-chiang: Removed

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
wey-chiang: Removed the GenericResource from the struct definitions

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
wey-chiang: Removed the GenericResource from the struct definitions

internal/compliance/snapshot_poller/pollers/aws/eks_cluster_test.go
wey-chiang: Will keep for now in favor of being more explicit.

internal/compliance/snapshot_poller/pollers/aws/awstest/eks.go
wey-chiang: Removed the conditional

internal/compliance/snapshot_poller/pollers/aws/awstest/eks.go
wey-chiang: We have the `listFargateProfilesPages` and `listNodegroupsPages` in `eks_cluster_test.go` already defined

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,310 @@
...
+		RoleArn:              details.RoleArn,
+		Status:               details.Status,
+		Version:              details.Version,
+	}
+
+	eksCluster.FargateProfile, err = getEKSFargateProfiles(eksSvc, details.Name)
+	if err != nil {
+		zap.L().Warn("Error with IAM Permissions - Use the AWS Console or CloudFormation to update the Panther" +
+			"Audit Role policy to include the eks:ListFargateProfiles and eks:DescribeFargateProfile permissions")
+	}
```
nhakmiller, line 267: Lets only catch the error & continue if it's specifically the auth error. If it's not an auth error, we want to return the error like normal.

This can be done by comparing it to a constant defined by the package (preferred) such as here: https://github.com/panther-labs/panther/blob/master/internal/compliance/snapshot_poller/pollers/aws/iam_user.go#L159-L160. Sadly this relies on the AWS service actually defining the error codes in the package, which is not always the case.

If the error codes are not defined in the package, this can be done by doing string comparisons on the error message such as here: https://github.com/panther-labs/panther/blob/master/internal/compliance/snapshot_poller/pollers/aws/cloudformation_stack.go#L133-L134.

Let me know if you have any questions on this.

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -105,8 +106,9 @@ var (
 		awsmodels.Ec2NetworkAclSchema:       {"EC2NetworkACL", PollEc2NetworkAcls},
 		awsmodels.Ec2SecurityGroupSchema:    {"EC2SecurityGroup", PollEc2SecurityGroups},
 		awsmodels.Ec2VolumeSchema:           {"EC2Volume", PollEc2Volumes},
-		awsmodels.Ec2VpcSchema:              {"EC2VPC", PollEc2Vpcs},
+		awsmodels.Ec2VpcSchema:              {"EC2VasPC", PollEc2Vpcs},
```
nhakmiller, line 109: Why this change? Looks like an accident maybe.

nhakmiller (commented):
I think this is my last round of feedback, after these issues have been addressed I will be good to approve.

internal/compliance/snapshot_poller/pollers/aws/poll.go
wey-chiang: O_O
Good catch, not sure what happened there. Reverting

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -0,0 +1,315 @@
...
+
+	eksCluster.FargateProfile, err = getEKSFargateProfiles(eksSvc, details.Name)
+	if err != nil {
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) && awsErr.Code() == eks.ErrorCodeAccessDenied {
+			zap.L().Warn("Error with IAM Permissions - Use the AWS Console or CloudFormation to update the Panther" +
+				"Audit Role policy to include the eks:ListFargateProfiles and eks:DescribeFargateProfile permissions")
+		} else {
+			return nil, nil
```
nhakmiller, line 270: This should be

```
return nil, err
```

nhakmiller (approved):
One minor comment, otherwise looks good to go! Good job on this PR.

PR #1629, opened 2020-09-30
Fix: color of Policy & Rule"test" handles

3nvi:
## Background

There's currently a  small visual bug

Before

Screen Shot 2020-09-30 at 2 50 24 PM

After

Screen Shot 2020-09-30 at 2 50 12 PM

## Changes

- Replace `bg` with `backgroundColor`

## Testing

- N/A


alexmylonas (approved):
👍

PR #1633, opened 2020-09-30
Make rule engine return the exception if title() or dedup() raises

### LINKED ISSUE #1592: Raising an exception in a Rule title or dedup function does not fail UI tests
### Describe the bug

If the title or dedup function raises an exception the UI tests do not fail.

### Steps to reproduce
image

giorgosp:
Closes #1592

## Background

See https://github.com/panther-labs/panther/issues/1592 for more context.

## Changes

The rule engine now returns any errors raised from `title()` or `dedup()` instead of swallowing them.
Subsequently, rule tests will fail if any of its functions raises an exception.

## Testing

- unit test

### DIFF
--- a/internal/log_analysis/rules_engine/src/rule.py
+++ b/internal/log_analysis/rules_engine/src/rule.py
@@ -111,16 +111,21 @@ class Rule:
 
         self._default_dedup_string = 'defaultDedupString:{}'.format(self.rule_id)
 
-    def run(self, event: Dict[str, Any]) -> RuleResult:
-        """Analyze a log line with this rule and return True, False, or an error."""
+    def run(self, event: Dict[str, Any], raise_title_dedup: bool = False) -> RuleResult:
+        """
+        Analyze a log line with this rule and return True, False, or an error.
+        :param event: The event to run the rule against
+        :param raise_title_dedup: Whether to raise exceptions from title() and dedup() or use default values
+        """
 
         dedup_string: Optional[str] = None
         title: Optional[str] = None
         try:
             rule_result = self._run_command(self._module.rule, event, bool)
             if rule_result:
-                title = self._get_title(event)
-                dedup_string = self._get_dedup(event, title)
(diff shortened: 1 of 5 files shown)


internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -135,46 +135,45 @@ def _get_dedup(self, event: Dict[str, Any], title: Optional[str]) -> str:
...
             # If no dedup function defined, return default dedup string
             return self._default_dedup_string
-        try:
-            dedup_string = self._run_command(self._module.dedup, event, str)
-        except Exception as err:  # pylint: disable=broad-except
-            self.logger.warning('dedup method raised exception. Defaulting dedup string to "%s". Exception: %s', self.rule_id, err)
+
+        dedup_string = self._run_command(self._module.dedup, event, str)
+
+        if not dedup_string:
```
kostaspap, line 141: Previously if `dedup` method was raising an exception, the code would use the default deduplication string. 
Same for title. 

With this change, if the title method fails, the whole run will fail. We don't want that.

kostaspap:
The parent ticket is talking about returning the exception in case of tests. Your changes modify the behavior for real data.

kostaspap (commented):
Has this been verified E2E?

giorgosp:
> Has this been verified E2E?

Deploying now. However it shouldn't matter for this change - There are unit tests in the rule engine that check all the cases

internal/log_analysis/rules_engine/src/rule.py
giorgosp: Done

kostaspap:
> > Has this been verified E2E?
> 
> Deploying now. However it shouldn't be needed for this change - There are unit tests in the rule engine that check all the cases

Sure, but since the original issue is about fixing an issue that was identified in the UI, it would be good to go through this kind of verification 😄

kostaspap (approved):
LGTM, once verified E2E

PR #1635, opened 2020-10-01
Frontend Telemetry [2/2]

alexmylonas:
## Description

This PR concludes the effort needed to implement some basic frontend product analytics. This will enable business users to view product usage on the Mixpanel dashboard.

### Changes in UI
- Added a new modal that will be displayed when someone upgrades past this version (check screenshot below) in order to prompt users to give us consent for gathering product analytics.

- Updated both the initial modal to also ask for product analytics alongside error reporting and general settings page (check screenshots)

### User events
- Updated Alert Status
- Invited User
- Picked a Log source to add
- Added Log Source
- Picked a Destination to add
- Added a Destination
- Created Policy
- Created Rule
- User sign in
- Failed MFA

### Page view events: 
- Log Analysis Overview
- Compliance Overview
- List Rules
- List Log sources
- List Alerts

## Screenshots
Screenshot 2020-10-01 at 1 31 45 PM
Screenshot 2020-10-01 at 1 31 29 PM
Screenshot 2020-10-01 at 1 30 34 PM

## Other changes & fixes
- Added OSS mixpanel public token in a `.env` file
- Updated web PR template to include telemetry
- Changed the way we check for analytics env variables, using `pantherConfig` and `NODE_ENV`
## Checklist:

- [x] Make sure the changes are responsive (appearing decently on devices)
- [x] ~~Make sure telemetry is added (analytics & errors)~~


web/src/components/utils/PromptController/PromptController.tsx
```diff
@@ -28,12 +28,16 @@ const PromptController: React.FC = () => {
   useGetGeneralSettingsConsents({
     onCompleted: data => {
       const shouldShowConsentModal = data.generalSettings.errorReportingConsent === null;
+      const shouldShowProductAnalyticsConsentModal = data.generalSettings.analyticsConsent === null;
       if (shouldShowConsentModal) {
```
3nvi, line 32: I honestly don't think we  need 2 separate modals for this. It gets complicated for no reason. We need to fire the **same modal that we use for analytics consent** and we can some `props` to it to control which switches we should.

For example we could add the props `showErrorConsent` and `showAnalyticsConsent`, both of which are boolean. This way we can simplify the codebase and re-use a lot of our existing code.


It's ok for the form to always "send" both consents to the API, since the error consent is hidden and its value won't change (so we are safe)

3nvi (commented):
Really nice work thinking of the migration plan. That's something I wouldn't have thought of.

My main concern has to do with the additional Modal. I think we can fully skip it, so let's see whether that's possible or not

web/src/components/utils/Modal/ModalManager.tsx
```diff
@@ -29,6 +29,7 @@ import DeleteDestinationModal from 'Components/modals/DeleteDestinationModal';
 import DeleteRuleModal from 'Components/modals/DeleteRuleModal';
 import NetworkErrorModal from 'Components/modals/NetworkErrorModal';
 import AnalyticsConsentModal from 'Components/modals/AnalyticsConsentModal';
+import ProductAnalyticsConsentModal from 'Components/modals/ProductAnalyticsConsentModal';
```
vorillaz, line 32: Personally, I would load these components asynchronously since they are not blocking for the end user.

web/src/helpers/analytics.ts
```diff
@@ -20,21 +20,15 @@ import { DestinationTypeEnum } from 'Generated/schema';
...
-const mixpanelPublicToken = process.env.MIXPANEL_PUBLIC_TOKEN;
-
-const envCheck =
-  mixpanelPublicToken &&
-  // TODO: Pending backend to work
-  storage.local.read<boolean>(ANALYTICS_CONSENT_STORAGE_KEY) &&
-  process.env.NODE_ENV === 'production';
+import { AlertSummaryFull } from 'Source/graphql/fragments/AlertSummaryFull.generated';
 
 const evaluateTracking = (...args) => {
```
vorillaz, line 25: @3nvi @alexmylonas Just a minor note that I just spotted. 
In case `mixpanel` throws an error due to connection issues or any unknown errors this one will bubble up to the `ErrorBoundary`. 
Besides collecting some data for market purposes `trackEvent` is not essential for the end-users, we are always consuming the `trackEvent` along with a user's actions:
```jsx
<Button onClick={() => {
  trackEvent(...);
  onClickHandler();
}}>
``` 
In the example above if the `trackEvent` fails for some reason the user is actually affected.
So I would suggest changing the original implementation in order to ignore errors thrown from mixpanel as:
```ts
const evaluateTracking = (...args) => {
  // pseudo parallelism
  setTimeout(() => {
    try {
      const mixpanelPublicToken = process.env.MIXPANEL_PUBLIC_TOKEN;
      if (!mixpanelPublicToken || storage.local.read<boolean>(ANALYTICS_CONSENT_STORAGE_KEY) !== true) {
        return;
      }
      mx.init(mixpanelPublicToken);
      mx.track(...args);
    } catch(ignore) {
      // ignored
    }
  }, 0)
};
```

3nvi: Instead of setTimeout, you can use `window.requestIdleCallback`

web/src/components/forms/AnalyticsConsentForm/AnalyticsConsentSection.tsx
```diff
@@ -0,0 +1,54 @@
...
+      />
+      <Box mt={4}>
+        <Field
+          mt={4}
+          as={FormikSwitch}
+          name="analyticsConsent"
+          label="Report Product Analytics"
+          aria-describedby="product-reporting-section-helper"
+        />
+        <FormHelperText mt={4} id="error-reporting-section-helper">
```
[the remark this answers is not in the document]
alexmylonas: 1. Doesn't look that good + they share the common link. 
2. Updated aria-labels
3. Spacing looks ok imo (check screenshots)

web/src/components/utils/PromptController/PromptController.tsx
alexmylonas: I thought about this since bit this serves basically as backwards compatibility issue felt more correct to isolate it.

Nonetheless, I have updated the modal to take prop and conditionally display text & components. Have a look!

web/src/helpers/analytics.ts
alexmylonas: Very nice thought, to safeguard this. Implemented!

alexmylonas (commented):
Thanks for

web/src/components/dropdowns/UpdateAlertDropdown/UpdateAlertDropdown.tsx
```diff
@@ -92,13 +93,16 @@ const UpdateAlertDropdown: React.FC<UpdateAlertDropdownProps> = ({ alert }) => {
       },
     }),
     onCompleted: data => {
+      const { status, severity } = data.updateAlertStatus;
+      trackEvent({
+        event: EventEnum.UpdatedAlertStatus,
+        src: SrcEnum.Alerts,
```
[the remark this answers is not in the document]
3nvi: @alexmylonas `src` is definitely not the correct name then. In my mind is "where this got executed". Keep it as is for  this PR, but could we create a new PR where we would rename it to something like `rel`, `ref` or whatever 😆

web/src/components/forms/AnalyticsConsentForm/AnalyticsConsentForm.tsx
```diff
@@ -33,10 +34,12 @@ interface AnalyticsConsentFormProps {
 
 const validationSchema = Yup.object().shape({
   errorReportingConsent: Yup.boolean().required(),
+  analyticsConsent: Yup.boolean().required(),
 });
 
 const initialValues = {
   errorReportingConsent: true,
+  analyticsConsent: true,
```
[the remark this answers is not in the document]
3nvi: If that's ok legally, then sure

web/src/components/forms/AnalyticsConsentForm/AnalyticsConsentSection.tsx
```diff
@@ -22,17 +22,43 @@ import { Box, FormHelperText, Link } from 'pouncejs';
 import { PRIVACY_DOC_URL } from 'Source/constants';
 import FormikSwitch from 'Components/fields/Switch';
 
-const ErrorReportingSection: React.FC = () => {
+interface AnalyticsConsentSectionProps {
+  hideErrorReporting?: boolean;
+}
+const AnalyticsConsentSection: React.FC<AnalyticsConsentSectionProps> = ({
+  hideErrorReporting,
```
3nvi, line 29: Nitpick: I would personally change these props to `showXXX` for all items in this section. This helps us "generalize" this, by not creating "exception props", but "feature props".

Meaning, instead of creating props that serve 1 particular use-case, you create more generic props. That's a nitpick of mine though. What do  you think @alexmylonas ?

web/scripts/utils.js
```diff
@@ -29,6 +29,8 @@ function loadDotEnvVars(path) {
 function getAppTemplateParams() {
   const PANTHER_CONFIG = {
     PANTHER_VERSION: process.env.PANTHER_VERSION,
+    SENTRY_DSN: process.env.SENTRY_DSN,
```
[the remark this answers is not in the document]
alexmylonas: This wasnt needed (and wasn't working for that matter) i am not sure what was the root cause

web/src/components/forms/AnalyticsConsentForm/AnalyticsConsentSection.tsx
```diff
@@ -22,17 +22,43 @@ import { Box, FormHelperText, Link } from 'pouncejs';
...
       <Field
         as={FormikSwitch}
-        name="errorReportingConsent"
-        label="Report Web Application Errors"
-        aria-describedby="error-reporting-section-helper"
+        name="analyticsConsent"
+        label="Send Product Analytics"
+        aria-describedby="product-analytics-section-helper"
       />
-      <FormHelperText mt={2} id="error-reporting-section-helper">
```
3nvi, line 34: I still think we need  2  Helper texts,  one for each switch.

- One will say that we gather analytics to make this product better by understanding how users use i
- One will say that  we report exceptions to fix stuff as quickly as possible and make the product reliable

In my eyes, this is **not** handled by a  **single** helper text

web/src/components/utils/PromptController/PromptController.tsx
3nvi: I know it's not ideal and not super pretty, but I feel that it's easier to maintain and reason about 1 year from now, when we'll have forgotten about this

web/src/components/utils/PromptController/PromptController.tsx
```diff
@@ -28,12 +28,16 @@ const PromptController: React.FC = () => {
   useGetGeneralSettingsConsents({
     onCompleted: data => {
       const shouldShowConsentModal = data.generalSettings.errorReportingConsent === null;
+      const shouldShowProductAnalyticsConsentModal = data.generalSettings.analyticsConsent === null;
```
3nvi, line 31: This logic  needs to be updated. The `shouldShowConsentModal` should include the check for `analyticsConsent === null`.

I'd envisions something like:

```ts
const shouldShowErrorConsent = data.generalSettings.errorReportingConsent === null;
const shouldShowProductAnalyticsConsent = data.generalSettings.analyticsConsent === null;
const shouldShowConsentModal = shouldShowErrorConsent || shouldShowProductAnalyticsConsent;

if (shouldShowConsentModal) {
  showModal({ modal: MODALS.ANALYTICS_CONSENT, props: { hideErrorReporting: !shouldShowErrorConsent  });
}
```

The payload

web/src/helpers/analytics.ts
```diff
@@ -20,20 +20,18 @@ import { DestinationTypeEnum } from 'Generated/schema';
...
-  if (envCheck) {
+  const mixpanelPublicToken = process.env.MIXPANEL_PUBLIC_TOKEN;
+  if (!mixpanelPublicToken || storage.local.read<boolean>(ANALYTICS_CONSENT_STORAGE_KEY) !== true) {
+    return;
+  }
+  try {
     mx.init(mixpanelPublicToken);
     mx.track(...args);
+  } catch (e) {
+    // not many things to do here
```
3nvi, line 34: can we report to sentry 😆

web/src/helpers/analytics.ts
3nvi: perhaps wrap the "try-catch" block in a `window.requestIdleCallback(() => { ... })`. This makes the browser execute this only when it has "idle time".

Unfortunately, Safari doesn't have support for this, so in our utils we have to add

```ts
window.requestIdleCallback = window.requestIdleCallback || (cb => cb());
```

and update `web/src/helpers/fireworks` to use `window.requestIdleCallback` since currently it uses a local version of what I wrote above (it re-defines it)

3nvi (commented):
Really good.

Couple of things to agree on and we can move ahead @alexmylonas

web/src/components/utils/PromptController/PromptController.tsx
alexmylonas: updated both this check

web/src/components/forms/AnalyticsConsentForm/AnalyticsConsentSection.tsx
alexmylonas: Updated to have both `showErrorConsent` & `showProductAnalyticsConsent`

web/src/components/forms/AnalyticsConsentForm/AnalyticsConsentSection.tsx
alexmylonas: Added two helpers text!

web/src/helpers/analytics.ts
alexmylonas: Added!

web/src/components/forms/AnalyticsConsentForm/AnalyticsConsentForm.tsx
```diff
@@ -21,34 +21,45 @@ import { Form, Formik } from 'formik';
...
+  onSubmit,
+}) => {
+  const validationSchema = Yup.object().shape({
+    errorReportingConsent: showErrorConsent ? Yup.boolean().required() : null,
+    analyticsConsent: showProductAnalyticsConsent ? Yup.boolean().required() : null,
+  });
   return (
     <Formik<AnalyticsConsentFormValues>
-      initialValues={initialValues}
+      initialValues={showErrorConsent ? initialValues : { analyticsConsent: true }}
```
3nvi, line 53: I think we can just this line as is. Was there a problem?

The way you do it now takes iinto account only `showErrorConsent` and not `showProductAnalyticsConsent`.

web/src/components/utils/PromptController/PromptController.tsx
```diff
@@ -27,13 +27,24 @@ const PromptController: React.FC = () => {
   // We are intentionally over-fetching, in order to proactively add this data to the cache
   useGetGeneralSettingsConsents({
     onCompleted: data => {
-      const shouldShowConsentModal = data.generalSettings.errorReportingConsent === null;
+      const shouldShowErrorConsent = data.generalSettings.errorReportingConsent === null;
+      const shouldShowProductAnalyticsConsent =
+        true || data.generalSettings.analyticsConsent === null;
```
3nvi, line 32:
```suggestion
        data.generalSettings.analyticsConsent === null;
```

Careful 😆

3nvi (commented):
LGTM @alexmylonas . Just fix those 2 small issues and we are good

web/src/components/utils/PromptController/PromptController.tsx
alexmylonas: Good catch, thank you, fixed

3nvi (approved):
LGTM 🎉

PR #1649, opened 2020-10-02
Add support for strftime and unix microseconds/nanoseconds to tcodec 

### LINKED ISSUE #1511: Add support for Python strftime to tcodec
### Description

We should add support for [Python's strftime](https://strftime.org/) to `tcodec` to allow defining timestamp formats using this layout

### Acceptance Criteria

- We can define timestamp layouts using `strftime` in `tcodec` tags

alxarch:
## Background

Closes #1511

## Changes

- Adds the ability to use `tcodec:"strftime=%Y-%m-%d" tags to parse JSON timestamps using a custom format defined in [strftime](https://strftime.org/)
- Adds missing codecs for microsecond and nanosecond UNIX timestamps (`unix_us`, `unix_ns`)

## Testing

- Unit tests

### DIFF
--- a/internal/log_analysis/log_processor/pantherlog/tcodec/tcodec_test.go
+++ b/internal/log_analysis/log_processor/pantherlog/tcodec/tcodec_test.go
@@ -34,6 +34,18 @@ func TestUnixMilliseconds(t *testing.T) {
 	require.Equal(t, expect.Format(time.RFC3339Nano), actual.Format(time.RFC3339Nano))
 	require.Equal(t, expect, actual)
 }
+func TestUnixMicroseconds(t *testing.T) {
+	expect := time.Date(2020, 05, 24, 23, 50, 07, int(259123*time.Microsecond.Nanoseconds()), time.UTC).Local()
+	actual := UnixMicroseconds(1590364207259123)
+	require.Equal(t, expect.Format(time.RFC3339Nano), actual.Format(time.RFC3339Nano))
+	require.Equal(t, expect, actual)
+}
+func TestUnixNanoseconds(t *testing.T) {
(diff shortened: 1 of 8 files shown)


internal/log_analysis/log_processor/pantherlog/tcodec/tcodec_test.go
```diff
@@ -29,14 +29,16 @@ import (
 )
 
 func TestUnixMilliseconds(t *testing.T) {
```
kostaspap, line 31: Tests for micro & nano?

internal/log_analysis/log_processor/pantherlog/tcodec/unix.go
```diff
@@ -0,0 +1,161 @@
...
+	"strconv"
+	"time"
+
+	jsoniter "github.com/json-iterator/go"
+)
+
+// UnixSeconds reads a timestamp from seconds since UNIX epoch.
+// Fractions of a second can be set using the fractional part of a float.
+// Precision is kept up to microseconds to avoid float64 precision issues.
+func UnixSeconds(sec float64) time.Time {
```
kostaspap, line 31: Unix timestamp in seconds doesn't have fractional points. Not sure why we need to support that?

alxarch: It's very common to have timestamps in float numbers where the integral part is seconds and the fractional parts is fractions of a second. We use a single parser that handles both cases.

internal/log_analysis/log_processor/pantherlog/tcodec/tcodec_test.go
alxarch: Added

kostaspap (approved):

PR #1650, opened 2020-10-02
Refactor: Adjust UI to the new Pounce Button sizes

3nvi:
## Background

In Pounce we have added an additional size for `Button`, `IconButton` and `Icon`, so now Panther needs to update itself to account for these new sizes.

## Relates to

https://github.com/panther-labs/pounce/pull/140

### Refactors

- `<Icon>`: Whatever is `small` became `medium` 
- `<IconButton>`: Whatever is `small` became `medium`
- `<Button>`: Whatever was `small` became`medium`

## Changes

- Increase negative margin on Tables
- Replace any `<Button as={RRLink} />` with `<RRLink><Button as="div" /></RRLink`, since semantics were fully wrong and the new buttons had a weird alignment

## Testing

- Visually one by one. Everything **should look the same**


vorillaz (approved):
✂️

PR #1652, opened 2020-10-02
Support versioned release branches in release sync

austinbyers:
## Background

We are going to try creating a new branch for each release: `release-1.10`, `release-1.11`, etc. This allows us to backport fixes to old releases if necessary and, more importantly, means we no longer have to keep `release` in sync with `master`.

When starting a new release, we just have to `git checkout -b release-1.10` and push the new branch, instead of merging in changes from `master` onto a single shared `release` branch.

This changes the `sync-to-release` automation to use the latest version of the release branch, based on the branch name.

## Changes

- When handling `sync-to-release`, find the `release*` branch with the latest version number instead of using `origin/release`

## Testing

- Created local branches to test the sorting:

```bash
# find all "release" branches (you need -r for remote branches, I'm just testing local branches here)
$ git branch --list "release*"
  release
  release-1.10
  release-1.102
  release-1.11
  release-1.20
  release-1.8
  release-1.9

# sort by semantic version in descending order
$ git branch --list "release*" | sort -V -r
  release-1.102
  release-1.20
  release-1.11
  release-1.10
  release-1.9
  release-1.8
  release

# pull out the top choice
$ git branch --list "release*" | sort -V -r | head -n 1
  release-1.102

# and trim spaces
$ git branch --list "release*" | sort -V -r | head -n 1 | xargs
release-1.102
```

Before I discovered the `-V` flag (so happy that exists), I was able to sort manually by splitting on the `.`:

```bash
$ git branch --list "release*" | sort -k 2 -t "." -n -r
  release-1.102
  release-1.20
  release-1.11
  release-1.10
  release-1.9
  release-1.8
  release
```

This means "split on periods, choose the second column, and sort it numerically in reverse order." Just wanted to put that here for reference in case we need it for something

-------

I already have pushed a `release-1.10` branch; once this merges I'll put up a PR against `master` and test the release sync to make it sure it works against the newly versioned branch name


rleighton (approved):
Well that was simple! Excellent.

PR #1653, opened 2020-10-02
EKS FargateProfile Permissions

wey-chiang:
## Background

The current Panther Audit Role does not include permissions for Fargate Profiles

## Changes

Added permissions to allow eks:ListFargateProfiles and eks:DescribeFargateProfile


nhakmiller (approved):
We will need to be sure to publish this as well, @austinbyers should we publish this now manually or wait until the next release and have it be published as part of that?

wey-chiang:
> We will need to be sure to publish this as well, @austinbyers should we publish this now manually or wait until the next release and have it be published as part of that?

👍 I'll wait for confirmation here to merge

deployments/auxiliary/cloudformation/panther-cloudsec-iam.yml
```diff
@@ -18,7 +18,7 @@ AWSTemplateFormatVersion: 2010-09-09
 Description: IAM roles for an account being scanned by Panther.
 
 Metadata:
-  Version: v1.0.1
+  Version: v1.0.2
```
austinbyers, line 21: Nit: we usually bump the version to the next Panther version, so we know what version it was last changed in. I.e. `v1.10.0`

But no strong preference here

austinbyers (approved):
Publishing the aux templates is part of publishing a release, so yes it will happen.

If this needs to be in 1.9.0, you'll have to manually cherry-pick it to the `release-1.9` branch since we've already mostly finished that one.

If this needs to be in 1.10.0, just apply the `sync-to-release` label and panther-bot will copy it over. 
Then I'll publish all the templates next week when I setup the release environment.

If this needs to be in 1.11.0, you don't need to do anything; it'll be copied to the next release when we start 1.11 in 2 weeks' time

deployments/auxiliary/cloudformation/panther-cloudsec-iam.yml
wey-chiang: Reverted back to `v1.0.1`

nhakmiller: > If this needs to be in 1.11.0, you don't need to do anything; it'll be copied to the next release when we start 1.11 in 2 weeks' time

I think `1.11.0` is fine, no customers are waiting on this. In that case, should we set the `Version` to `v1.11.0`?

austinbyers: Sure, that works. Honestly, I'm not sure that we need to explicitly version these templates at all, everyone typically uses `latest` anyway. But since we have versions, `1.11.0` is fine by me

PR #1658, opened 2020-10-02
Release sync automation: Strip origin prefix from branch name

austinbyers:
## Background

Take 3! #1657 was able to get the branch name passed to the PR creation, but because we are listing remote branches, the branch name was `origin/release-1.10` instead of `release-1.10` and the action failed: https://github.com/panther-labs/panther/runs/1200950117?check_suite_focus=true

So now I strip the `origin/` prefix from the branch name before passing the env variable.

## Testing

Verified syntax locally:

```bash
BRANCH="origin/release-1.10"
echo "LATEST_RELEASE_BRANCH=${BRANCH#origin/}"
# LATEST_RELEASE_BRANCH=release-1.10
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,7 +94,7 @@ jobs:
         run: |
           LATEST_RELEASE_BRANCH=$(git branch -r --list "origin/release*" | sort -V -r | head -n 1 | xargs)
           git checkout -b $HEAD $LATEST_RELEASE_BRANCH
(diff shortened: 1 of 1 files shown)


nhakmiller (approved):
Not familiar with this syntax, but I'll trust you on it!

austinbyers:
> Not familiar with this syntax, but I'll trust you on it!

Me neither, I just googled "bash strip string prefix" lol this whole exercise has certainly improved my shell foo

Issue #1651, opened 2020-10-02
Self monitoring produces too many s3 access logs


rleighton:
### Issue
When Panther is configured to self monitor, it enables s3 access logs as one of the log Types. We also configure by default all s3 buckets log to a single place , which is monitored by Panther.

Unfortunately, this creates a positive feedback loop where:
- panther writes processed s3 access log to the processed data bucket
- this generates an s3 access event, which gets written to the bucket we monitor
- which generates another write to the processed data bucket
- .... forever

This produces un-necessary load and expense.

### Solution

We _do_ want to monitor the Panther buckets though. The solution is to have the processed data bucket log s3 accesses to a different location NOT monitored by Panther but in the account for audit.

I expect this is a small change to the CloudFormation.

I will try to get this done early next week so we can have it in 1.10.

PR #1666, opened 2020-10-05
Add s3 access log bucket for data processing buckets to break feedback loop

rleighton:
### Issue
When Panther is configured to self monitor, it enables s3 access logs as one of the log Types. We also configure by default all s3 buckets log to a single place , which is monitored by Panther.

Unfortunately, this creates a positive feedback loop where:
- panther writes processed s3 access log to the processed data bucket
- this generates an s3 access event, which gets written to the bucket we monitor
- which generates another write to the processed data bucket
- .... forever

This produces un-necessary load and expense.

### Solution

We _do_ want to monitor the Panther buckets though. The solution is to have the processed data bucket log s3 accesses to a different location NOT monitored by Panther but in the account for audit.

Closes: https://github.com/panther-labs/panther/issues/1651

## Changes

- Added bucket `DataProcessingS3AccessLogs` and configured `ProcessedData` to use it.

## Testing

- mage deploy, confirmed access logs were going to correct place


deployments/bootstrap.yml
```diff
@@ -259,6 +260,57 @@ Resources:
               Bool:
                 aws:SecureTransport: false
 
+  # Here we store s3 access logs from buckets related to data processing. The reason we put the logs here
+  # instead of the AuditLogs bucket, is because the AuditLog bucket can be "self" monitored by Panther
+  # and if we have data processing related s3 access logs we create a positive feedback loop where we
+  # wrote to a bucket involved with data processing, which creates an access log, which triggers Panther
+  # log processing, which triggers an s3 access log ... positive feedback loop.
```
austinbyers, line 267: What about the other S3 buckets? Standard deployments will have:

* Audit logs
* Input data (for log processing)
* Processed data (log processing intermediate?)
* Historical data
* Athena results

Do none of those other buckets cause feedback issues? Rather than splitting out these access logs, would it be easier to just disable self-onboarding of S3 access logs entirely? We keep the same access logs in the same bucket, but Panther doesn't ingest them anymore.

I'm wary that this could be a problem again, and splitting out just one bucket worth of access logs adds complexity but doesn't actually solve the problem. For example, suppose we have application audit logging with support for internal communication. When the log processor makes an API call, it generates a log, which gets sent to the input bucket, which might trigger the log processor to make an API call again, and so on.

-----

In other words, splitting access logs out to another bucket adds nontrivial overhead - we have to onboard the new bucket in all of our own internal security monitoring, for example (where the monitoring is in a separate account).

I just want to make sure this solution is sufficiently permanent and not something we'll have to do again when we realize a new feature accidentally created a different kind of feedback loop. Curious what @nhakmiller thinks as well

austinbyers: Perhaps one way to think about it is this: we have 2 access log buckets, one which can be self-ingested (e.g. rule/policy history), and one which cannot (some kinds of log analysis).

Then each new bucket can just choose which audit logging is most appropriate, and we can easily change it later.

So at the very least, I might give this bucket a more generic cloudformation name, something like `AuditLogsWithoutSelfIngestion`

rleighton:
I can change the name to be more descriptive.

I believe the feedback is exclusively from the final writes. So the
processing bucket and the historical bucket (parquet) are the ones that
break the cycle. Since parquet processing uses Athena , I was also going to
configure the results bucket in enterprise also

deployments/bootstrap.yml
nhakmiller: We could also use s3 prefixes, not sure if it's a breaking change for current setups but if we configure s3 access logging to write to `self-monitor/bucket_name` or `no-self-monitor/bucket_name`, then we could configure the source integration with a prefix and keep it in the same bucket but still not overloading. Then if we decide a future bucket needs  to not be self monitored, we update it's  config to log to  the other prefix. This also means an external party that doesn't care about a feedback loop can just monitor the whole bucket  (such  as us in the hosted accounts)

nhakmiller: I don't have a strong preference either way, in our hosted environment we disable self-onboarding which if I'm reading the change correctly means nothing will happen.

rleighton:
I started with prefixes, that’s a disruptive change. This is much simpler.

deployments/bootstrap.yml
austinbyers: >  in our hosted environment we disable self-onboarding which if I'm reading the change correctly means nothing will happen.

Nothing will happen, but you'll have to change how we monitor all of those accounts - you'll need to add the new bucket in each account

rleighton:
Not sure I understand what you mean. The hosted accounts use an external S3
access log bucket, right?

austinbyers:
> I started with prefixes, that’s a disruptive change. This is much simpler.

I kinda like the prefix idea, since like @nhakmiller said, it allows our hosted environment and other aggregators to ingest the entire access logs bucket and not care about the distinction.

Creating a new bucket may be simpler to implement in the code, but requires all existing deployments to add monitoring for a new access logs bucket. So you essentially just pushed the "disruptive change" to the users, which affects us and a few of our customers. It's strange to me to have a single service generate two different access log buckets, and seems like it would make things harder for anyone who wants to consume those logs.

I'll approve this PR since it seems like the best path forward, but I'd be curious to know why the prefix idea is too difficult - does it require UI changes, for example?

austinbyers (approved):
LGTM with more generic name

austinbyers:
> Not sure I understand what you mean. The hosted accounts use an external S3
access log bucket, right?

Oh, you're right - if we override the access log bucket completely, I guess this doesn't matter

rleighton:
The prefix change is disruptive because it requires big changes to the self
on boarding code. It would require changing from a single source
integration for the whole bucket to one for each log type

rleighton:
The big motivation for this is to reduce the snowflake bill. S3 access logs
have terrible aggregation so we get huge numbers of tiny files, which
drives up the snowflake ingest bill

PR #1667, opened 2020-10-05
Added backslash to escape dot in 'github.com'

wey-chiang:
## Background

Code Scanning found an insecure regex containing an unescaped period `.` in `github.com`

## Changes

- Added a backslash to regex to denote a literal `.` i.e. `github.com` => `github\.com`


nhakmiller (approved):

austinbyers:
This was a great find by GitHub's new CodeQL service! Thanks for fixing, I always make this same mistake in regexes

PR #1671, opened 2020-10-05
Describe Large Number of Tasks

lindsey-w:
## Background

If a cluster has more than 100 tasks, the snapshot poller will throw an exception. This is caused by the fact the DescribeTasks API call only allows querying 100 Task ARNs at a time. 

## Changes

- Batch task arns in groups of 100 for each DescribeTasks call.
- Added new Unit Test to verify DescribeTasks is called as expected when a cluster has > 100 tasks.
- Updated mocked DescribeTasks call to return an error when more than 100 task arns are passed.

## Testing

- `mage test:ci`

Reference: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeTasks.html


internal/compliance/snapshot_poller/pollers/aws/awstest/ecs.go
```diff
@@ -246,6 +238,16 @@ var (
 	MockEcsForSetup = &MockEcs{}
 )
 
+// initialize globals
+func init() {
```
nhakmiller, line 242: Smart!

nhakmiller (approved):
Thanks for jumping on this one so quick! I don't suppose you were able to test in your dev account by creating a cluster with 100+ tasks? Also, we should add the `sync-to-release` tag so we can get this in 1.10.

PR #1577, opened 2020-09-22
Least privilege for satellite account IAM roles

phattymcG:
## Background

IAM role principals default to the entire master account.

## Changes

- Updated trust principals for IAM roles and one SNS topic policy sid

## Testing

- N/A

PR #1678, opened 2020-10-06
Least privilege trust principals

glerb:
## Background

Worked with Panther engineering to identify specific roles in Panther master account that need to be trusted by infra in monitored accounts.

## Changes

Identification of specific principals.

## Testing

Applied to a running Panther instance.


deployments/auxiliary/terraform/README.md
```diff
@@ -1,3 +1,67 @@
...
+
+# Least Privilege Trust Principals
+
+The default trust principals for various IAM roles and one SNS topic policy sid are the entire _master_ account. These can be further restricted as follows:
+
+"Cloud Security" template:
+--- for AWS configuration scanning and remediation
+
+"panther_audit" role trust principal:
+- arn:aws:iam::${var.master_account_id}:role/panther-app-SnapshotInfra-1111-PollerFunctionRole-11111111111
```
austinbyers, line 25: 1) Would you be comfortable wildcarding these so that they will work even if the role is re-created with a different name? `arn:aws:iam::${var.master_account_id}:role/panther-app-SnapshotInfra-*-PollerFunctionRole-*`

This allows the same template to work for any deployment, while still restricting access to the specific component that needs it.

2) If so, @nhakmiller does it make sense to just update the principals in the template itself rather than just here in the README?

nhakmiller: I don't think you can use wildcards in IAM trust policies, sadly. We could update the template generation code to specify the components exactly, but that only would work for the CloudFormation templates we generate.

austinbyers:
Is this a duplicate of #1577 ?

nhakmiller:
> Is this a duplicate of #1577 ?

Getting the CLA stuff lined up is all, I'll close the other one.

s0l0ist (approved):

nhakmiller (approved):

PR #1679, opened 2020-10-06
Async glue tables

giorgosp:
## Background

When a new log source with many log types is onboarded, creating 
Glue tables/views for these log types may take longer than we want in
the course of an API request.

Thus, we create the tables asynchronously, by sending a message to the
datacatalog updater queue, handled by the datacatalog updater lambda.

## Testing

- Unit 
- Manual E2E


deployments/log_analysis.yml
```diff
@@ -590,7 +590,9 @@ Resources:
             Resource: '*'
             Condition:
               ArnLike:
-                aws:SourceArn: !Ref ProcessedDataTopicArn
+                aws:SourceArn:
+                  - !Ref ProcessedDataTopicArn
+                  - !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-source-api
```
kostaspap, line 595: You don't need this change. It is enough to update the `panther-sources-api` Lambda permissions

internal/core/source_api/api/put_integration.go
```diff
@@ -271,10 +272,39 @@ func generateNewIntegration(input *models.PutIntegrationInput) *models.SourceInt
...
+}
+
+type CreateTablesMessage struct {
+	LogTypes []string
+	S3Bucket string
+}
+
+func (m CreateTablesMessage) Send(sqsClient sqsiface.SQSAPI) error {
+	fmt.Printf("Sent CreateTablesMessage %+v '%s'\n", m, env.DataCatalogUpdaterQueueURL)
+	marshalled, err := jsoniter.Marshal(m)
```
kostaspap, line 293: You can use `jsoniter.MarshalToString`

internal/core/source_api/api/put_integration.go
```diff
@@ -271,10 +272,39 @@ func generateNewIntegration(input *models.PutIntegrationInput) *models.SourceInt
...
+	if err != nil {
+		return err
+	}
+
+	sqsInput := sqs.SendMessageInput{
+		MessageBody: aws.String(string(marshalled)),
```
kostaspap, line 299: You can just do `&`. Even if it will get inlined, no need for function invocation here

internal/core/source_api/api/put_integration.go
```diff
@@ -271,10 +272,39 @@ func generateNewIntegration(input *models.PutIntegrationInput) *models.SourceInt
...
+		QueueUrl:    aws.String(env.DataCatalogUpdaterQueueURL),
+		MessageAttributes: map[string]*sqs.MessageAttributeValue{
+			"MessageType": {
+				DataType:    aws.String("String"),
+				StringValue: aws.String("CreateTables"),
+			},
```
kostaspap, line 305: You shouldn't need that there. It's a pain working with SQS message attributes and probably not needed since the payload should already contain sufficient information.

internal/core/source_api/api/put_integration.go
```diff
@@ -271,10 +272,39 @@ func generateNewIntegration(input *models.PutIntegrationInput) *models.SourceInt
...
+	m := CreateTablesMessage{
+		LogTypes: integration.RequiredLogTypes(),
+		S3Bucket: env.ProcessedDataBucket,
+	}
+	return m.Send(sqsClient)
+}
+
+type CreateTablesMessage struct {
+	LogTypes []string
+	S3Bucket string
```
kostaspap, line 288: I don't think that it's responsibility of the sources API to pass the S3 bucket. Given that `panther-datacatalog-updater` is responsible for updating the Glue tables and has all the "Glue context", I think it's better that it's the `datacatalog-updater` that deals with "all stuff Glue". 
We should decouple the `panther-sources-api` from the Glue internals so I think this should be removed.

internal/core/source_api/api/update_integration.go
```diff
@@ -161,13 +164,20 @@ func getItem(integrationID string) (*ddb.Integration, error) {
...
 	case models.IntegrationTypeSqs:
-		err = addGlueTables(input.SqsConfig.LogTypes)
+		logtypes = input.SqsConfig.LogTypes
 	}
+
+	m := CreateTablesMessage{
+		LogTypes: logtypes,
+		S3Bucket: env.ProcessedDataBucket,
+	}
+	err := m.Send(sqsClient)
```
kostaspap, line 180: comment for later: We should eventually remove the global variables from the sources-api codebase and just move them inside the `API` struct. Just seeing this code reminded me of it. Not needed for this PR :)

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -38,37 +41,70 @@ func SQS(event events.SQSEvent) error {
...
+		if msgType, ok := record.MessageAttributes["MessageType"]; ok && *msgType.StringValue == "CreateTables" {
+			err := handleCreateTablesMessage(record)
+			if err != nil {
+				return err
+			}
+		} else {
+			err := handleS3Notification(record)
+			if err != nil {
+				return err
+			}
```
kostaspap, line 53: DataCatalog Updater is already receiving different types of events. See https://github.com/panther-labs/panther/blob/5ebfda565709a7350e0d21fc68d889da0656d274/internal/log_analysis/datacatalog_updater/process/events.go#L38-L43  - you can add the message there.

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -38,37 +41,70 @@ func SQS(event events.SQSEvent) error {
 	for _, record := range event.Records {
 		// uncomment to see all payloads
 		//zap.L().Debug("processing record", zap.String("content", record.Body))
-		notification := &models.S3Notification{}
-		if err := jsoniter.UnmarshalFromString(record.Body, notification); err != nil {
-			zap.L().Error("failed to unmarshal record", zap.Error(errors.WithStack(err)))
-			continue
+		if msgType, ok := record.MessageAttributes["MessageType"]; ok && *msgType.StringValue == "CreateTables" {
```
kostaspap, line 44: IMO, it's a bit of a pain dealing with message Attributes. We can just differentiate based on payload it seems?

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -38,37 +41,70 @@ func SQS(event events.SQSEvent) error {
...
+			err := handleS3Notification(record)
+			if err != nil {
+				return err
+			}
 		}
+	}
+	return nil
+}
+
+func handleCreateTablesMessage(record events.SQSMessage) error {
```
kostaspap, line 59: The DataCatalog updater already gets a lot of different invocation payloads. The `SQS` method was actually just intended for new object creation notifications. Since now it can have more than that, I think the following change will make the codebase more aligned.
I would say, remove the `SQS` method and 
* Make the `HandleCreateTablesMessage` and `handleS3Notification` public. 
* Change the method signatures to `HandleCreateTablesMessage(msg api.CreateTablesMessage)` 
* Change the method signatures to `HandleS3Notification(msg models.S3Notification{})` 
* Then move invocation of these methods here: https://github.com/panther-labs/panther/blob/5ebfda565709a7350e0d21fc68d889da0656d274/internal/log_analysis/datacatalog_updater/main/lambda.go#L44-L56

internal/core/source_api/api/put_integration.go
kostaspap, line 288: This is actually the API of the DataCatalog Updater. It would be better to declare it there.

internal/core/source_api/api/put_integration.go
giorgosp: Will fix

internal/core/source_api/api/put_integration.go
giorgosp: My understanding is that SQS message attributes exist for that reason. Because the SQS event is already deserialized when it is passed to our lambda handler, we can use them to see what kind of payload is in the body.
Otherwise we would have to do classification based on unmarshalling like in the log processor. Message attributes is definitely simpler, except if we have some other simpler solution as well.

internal/core/source_api/api/put_integration.go
giorgosp: Hmm not sure about it. Making the bucket a param will allow us to reuse the create tables functionality of the datacatalog updater for other buckets as well.

Actually, the real reason I made it a field in the message was convenience, since the source-api seems to use the `ProcessedDataBucket` env var a lot while in the updater we don't pass it as an env var.
 
No problem hardcoding making it an env var in the updater if you think it won't be an issue going forward.

internal/core/source_api/api/update_integration.go
giorgosp: Absolutely

internal/log_analysis/datacatalog_updater/process/process.go
giorgosp: We could prefix our attributes with `Panther-` to avoid conflicts with system attributes. We won't know the payload if we don't try deserialize it with every possible message type.

internal/log_analysis/datacatalog_updater/process/process.go
giorgosp: Will do

internal/core/source_api/api/put_integration.go
kostaspap: > Making the bucket a param will allow us to reuse the create tables functionality of the datacatalog updater for other buckets as well.

The problem is that we are mixing different abstraction layers in the same API. "LogTypes" is a high level abstraction. S3 Bucket is very low level. 
The abstraction layers are like LogTypes -> Databases/Tables -> Underlying datastore (S3/Athena/Glue/Snowflake). We shouldn't mix them.

rleighton: For now, pass it as env var, like we do for other lambdas. There is only 1 bucket the tables live in for now.

internal/log_analysis/datacatalog_updater/process/process.go
kostaspap: > We won't know the payload if we don't try deserialize it with every possible message type.

Yeah, I agree with that, that's a bit horrible. The receiver of this event should know how to deserialize the event it has received, without having to try all alternatives. 

There are two ways to do that: 
1. Include it in the payload. Receiver extracts the `type` field of the received event and knows how to deserialize it. The 'type' can be extracted without deserializing the full event, using `gjson`.
2. Include it in the SQS Metadata

The reason that I don't like #2 is that it makes the design less portable as it relies on SQS for doing the deserialization. And the SQS API for dealing with MessageAttributes is a bit horrible (as you've seen with the whole `DataType` thingy they have). If we were to send this message through SNS or EventBridge or Kinesis we wouldn't have this capability. Sending events to SNS and let them forward them to SQS is actually a super common pattern since SQS allows only one consumer.

internal/log_analysis/datacatalog_updater/process/create_tables.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"github.com/aws/aws-sdk-go/service/sqs/sqsiface"
+	jsoniter "github.com/json-iterator/go"
+
+	"github.com/panther-labs/panther/internal/log_analysis/athenaviews"
+	"github.com/panther-labs/panther/internal/log_analysis/gluetables"
+)
+
+var processedDataBucket = os.Getenv("PROCESSED_DATA_BUCKET")
```
kostaspap, line 33: We should be using `"github.com/kelseyhightower/envconfig"` even for single properties. Soon we will have a second property and then someone will need to refactor. Better start it the proper way.

internal/log_analysis/datacatalog_updater/process/create_tables.go
```diff
@@ -0,0 +1,73 @@
...
+
+	sqsInput := sqs.SendMessageInput{
+		MessageBody: &marshalled,
+		QueueUrl:    &queueURL,
+		MessageAttributes: map[string]*sqs.MessageAttributeValue{
+			PantherMessageType: &CreateTableMessageAttribute,
+		},
+	}
+	_, err = sqsClient.SendMessage(&sqsInput)
+	return err
```
kostaspap, line 61: Wrap the error: `return errors.Wrapf(err, "failed to send message to SQS queue %s", queueURL)`

internal/log_analysis/datacatalog_updater/process/create_tables.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func HandleCreateTablesMessage(msg CreateTablesMessage) error {
+	for _, logType := range msg.LogTypes {
+		_, err := gluetables.CreateOrUpdateGlueTablesForLogType(glueClient, logType, processedDataBucket)
+		if err != nil {
+			return err
```
kostaspap, line 68: Same as above, you should be wrapping the error here. Will help with debugging

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -19,72 +19,69 @@ package process
...
-		}
-
-		for _, eventRecord := range notification.Records {
-			existsInCache, gluePartition, err := getPartition(eventRecord.S3.Bucket.Name, eventRecord.S3.Object.Key)
+		if msgType, ok := record.MessageAttributes["MessageType"]; ok && *msgType.StringValue == "CreateTables" {
+			msg := CreateTablesMessage{}
+			if err := jsoniter.UnmarshalFromString(record.Body, &msg); err != nil {
+				err = errors.WithStack(err)
+				zap.L().Error("failed to unmarshal record", zap.Error(err))
+				return err
```
kostaspap, line 68: We shouldn't be returning the error here. Just log it and continue. Nothing more we can do.

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -19,72 +19,69 @@ package process
...
-				continue
+				return err
 			}
-			if existsInCache { // already exists, nothing to do
-				continue
+		} else {
+			notification := S3Notification{}
+			if err := jsoniter.UnmarshalFromString(record.Body, &notification); err != nil {
+				zap.L().Error("failed to unmarshal record", zap.Error(errors.WithStack(err)))
+				return nil
```
kostaspap, line 78: If we fail to unmarshal, we should continue (skip this message) not return. We still want to process the rest of the messages.

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -19,72 +19,69 @@ package process
...
+		ctx = lambdalogger.Context(ctx, logger)
+		err = HandleSyncTableEvent(ctx, event.SyncTablePartitions)
+	// 3. An SQS message notifying about new data written and possibly needing a new partition to be added
+	default:
+		err = SQS(event.SQSEvent)
+	}
+	return
+}
 
 func SQS(event events.SQSEvent) error {
```
kostaspap, line 59: This could be now lowercase.

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -19,72 +19,69 @@ package process
...
-		}
-
-		if len(notification.Records) == 0 { // indications of a bug someplace
-			zap.L().Warn("no s3 event notifications in message", zap.String("message", record.Body))
-			continue
-		}
-
-		for _, eventRecord := range notification.Records {
-			existsInCache, gluePartition, err := getPartition(eventRecord.S3.Bucket.Name, eventRecord.S3.Object.Key)
+		if msgType, ok := record.MessageAttributes["MessageType"]; ok && *msgType.StringValue == "CreateTables" {
```
kostaspap, line 63: `"MessageType"` and `"CreateTables"` should be constants.

internal/log_analysis/datacatalog_updater/process/create_tables.go
giorgosp: Done

internal/log_analysis/datacatalog_updater/process/create_tables.go
giorgosp: Done

internal/log_analysis/datacatalog_updater/process/create_tables.go
giorgosp: Done

internal/log_analysis/datacatalog_updater/process/process.go
giorgosp: Done

internal/log_analysis/datacatalog_updater/process/process.go
giorgosp: Done

internal/log_analysis/datacatalog_updater/process/process.go
giorgosp: Good catch, fixed

internal/log_analysis/datacatalog_updater/process/process.go
giorgosp: Yeap, done

internal/core/source_api/api/put_integration_test.go
```diff
@@ -280,23 +273,7 @@ func TestPutLogIntegrationUpdateSqsQueuePermissions(t *testing.T) {
...
-	mockAthena.On("GetQueryExecution", mock.Anything).Return(&athena.GetQueryExecutionOutput{
-		QueryExecution: &athena.QueryExecution{
-			QueryExecutionId: aws.String("test-query-1234"),
-			Status: &athena.QueryExecutionStatus{
-				State: aws.String(athena.QueryExecutionStateSucceeded),
-			},
-		},
-	}, nil).Times(3)
-	mockAthena.On("GetQueryResults", mock.Anything).Return(&athena.GetQueryResultsOutput{}, nil).Times(3)
+	mockSQS.On("SendMessage", mock.Anything).Return(&sqs.SendMessageOutput{}, nil)
```
kostaspap, line 276: At least once of the tests should validate the that we send the proper payload in the SQS. Either here or in some other test replace `mock.Anything` with the expected payload.

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -19,72 +19,69 @@ package process
...
-			zap.L().Warn("no s3 event notifications in message", zap.String("message", record.Body))
-			continue
-		}
-
-		for _, eventRecord := range notification.Records {
-			existsInCache, gluePartition, err := getPartition(eventRecord.S3.Bucket.Name, eventRecord.S3.Object.Key)
-			if err != nil {
-				zap.L().Error("failed to get partition information from notification",
-					zap.Any("notification", notification), zap.Error(err))
+		if msgType, ok := record.MessageAttributes[PantherMessageType]; ok && *msgType.StringValue == *CreateTableMessageAttribute.StringValue {
```
kostaspap, line 63: Better use `aws.StringValue(msgType.StringValue)` to avoid potential nil pointer dereference. Same for the `*CreateTableMessageAttribute.StringValue`.

internal/log_analysis/datacatalog_updater/process/s3notification.go
```diff
@@ -0,0 +1,108 @@
...
+		zap.L().Warn("no s3 event notifications in message", zap.Any("message", notification))
+		return nil
+	}
+
+	for _, eventRecord := range notification.Records {
+		existsInCache, gluePartition, err := getPartition(eventRecord.S3.Bucket.Name, eventRecord.S3.Object.Key)
+		if err != nil {
+			zap.L().Error("failed to get partition information from notification",
+				zap.Any("notification", notification), zap.Error(err))
+			return nil
```
kostaspap, line 76: You should `continue`. We want to skip only one record, not the full batch.

internal/log_analysis/datacatalog_updater/process/s3notification.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		if existsInCache { // already exists, nothing to do
+			return nil
```
kostaspap, line 79: `continue`

kostaspap (commented):
Check the CI, it seems to be failing.

internal/log_analysis/datacatalog_updater/process/process.go
```diff
@@ -19,72 +19,72 @@ package process
...
+			if err := jsoniter.UnmarshalFromString(record.Body, &msg); err != nil {
+				err = errors.WithStack(err)
+				zap.L().Error("failed to unmarshal record", zap.Error(err))
 				continue
 			}
-			if existsInCache { // already exists, nothing to do
+			err := HandleCreateTablesMessage(msg)
+			if err != nil {
+				return err
+			}
```
kostaspap, line 76: nit: it could be simplified to 
```
return HandleCreateTablesMessage(msg)
```

Same below.

kostaspap (approved):
Nice!

PR #1680, opened 2020-10-06
Increase aws sdk retries for rules engine

rleighton:
## Background

We have occasional alarms due to aws api failures in the rule engine. It is using default settings. This PR increase the retries to 10 and uses a more modern (similar to Golang SDK) implementation for retries.

## Changes

- Added config for retries using "standard' retry mechanism
- Refactored AWS client creation

## Testing

- mage test:ci
- deployed and created an `AlwaysTrue` rule, no errors in CloudWatch and UI displayed the alerts and alert events.


nhakmiller (approved):
LGTM

kostaspap (approved):
Nice! Sync to release?

Issue #1669, opened 2020-10-05
CI: Authenticated docker image pulls


austinbyers:
### Description

Received a notice from CircleCI:

> On November 1st, Docker Hub will begin limiting anonymous image pulls. We want to make sure you know how you might be impacted and what you can do to avoid interruptions to your workflow.

> Adding [Docker authentication](https://circleci.com/docs/2.0/private-images/) to your pipeline config is the easiest way to avoid any service disruptions. If you use the Docker executor or pull Docker images when using the machine executor on CircleCI, we encourage you to authenticate. Because the anonymous API rate limits are based on IP addresses, they will impact CircleCI cloud customers. Authenticated users get higher per-user rate limits, regardless of IP.

### Acceptance Criteria

- Use authenticated docker pulls in CircleCI by Nov 1

Alternatively, we could switch to ECS for storing our docker image, which CircleCI also supports. We already use ECS to publish the web image, using it to publish the CI image seems reasonable. Then we shouldn't have to worry much about rate limits, since we're already paying AWS to host the ECS image

PR #1693, opened 2020-10-07
CI: Use authenticated docker pulls

austinbyers:
## Background

> On November 1st, Docker Hub will begin limiting anonymous image pulls. Adding Docker authentication to your pipeline config is the easiest way to avoid any service disruptions.

https://circleci.com/docs/2.0/private-images/

Closes: #1669 

## Changes

- In our DockerHub org, I created a new `pantherci` user with no permissions other than being able to pull the public images
- That username/password has been added to the CircleCI environment variables
- The environment variables are added to the CI template here to allow authenticated pulls

## Testing

- CI


3nvi (approved):
Looks good @austinbyers 🚢

austinbyers:
So far, it looks like this actually makes the `docker pull` faster as well, but we don't have enough data points to say for sure yet

PR #1694, opened 2020-10-07
fix: alert delivery status calculation

### LINKED ISSUE #1684: Delivery marked a successful even though it has failed
### Describe the bug

An alert failed to deliver to a destination. However, the UI was reporting the delivery as successful

### Steps to reproduce

1. Create more than one destinations. Make sure that one destination is misconfigured
2. Trigger an alert
3. We will report that delivery was successful, however a destination has failed

### Expected behavior

We should report that delivery was not successful

### Environment

- Panther version or commit: 1.10 testing

s0l0ist:
## Background

Closes #1684 

## Changes

- Added helper functions to properly calculate for each destination sorted by dispatch time
- Modified tests to reflect these changes
- Added a boolean to disable the user from hitting the re-send button too many times.

## Testing

- Visually by manually updating DDB entries and re-sending alerts
- mage test:web


web/src/pages/AlertDetails/AlertDetailsInfo/AlertDeliverySection/AlertDeliverySection.tsx
```diff
@@ -97,12 +98,43 @@ const AlertDeliverySection: React.FC<AlertDeliverySectionProps> = ({
       </Flex>
     );
   }
+  // Need to determine success for each destination (group by destination).
+  const deliveryStatusByDestination = React.useMemo(() => {
+    return enhancedAndSortedAlertDeliveries.reduce((acc, dest: DeliveryResponseFull) => {
```
3nvi, line 103: a quicker way to do this would be using [`lodash/groupBy`](https://lodash.com/docs/4.17.15#groupBy):

```ts
const deliveryStatusByDestination = groupBy(enhancedAndSortedAlertDeliveries, d =>  d.outputIId)
```

web/src/pages/AlertDetails/AlertDetailsInfo/AlertDeliverySection/AlertDeliverySection.tsx
```diff
@@ -97,12 +98,43 @@ const AlertDeliverySection: React.FC<AlertDeliverySectionProps> = ({
...
+      }
+      acc[dest.outputId] = [...acc[dest.outputId], dest];
+      return acc;
+    }, {});
+  }, [enhancedAndSortedAlertDeliveries]);
+
+  // Next, we sort each status inside each group by dispatchedAt and determine if it was successful
+  // This is all or nothing. The most recent status for ALL destinations should be successful, otherwise
+  // notify the user of a failure.
+  const allDestinationDeliveryStatuesSuccessful = React.useMemo(() => {
```
3nvi, line 116: a quicker way to do this would be using [`lodash/mapValues`](https://lodash.com/docs/4.17.15#mapValues) and [`lodash/sortBy`](https://lodash.com/docs/4.17.15#sortBy):

```ts
const allDestinationDeliveryStatuesSuccessful = mapValues(
  deliveryStatusByDestination, 
  dests => sortBy(deliveryStatusByDestination, 'dispatchedAt')
)
```

3nvi: We could also combine the 2 calls into a single `React.useMemo`

3nvi (commented):
Thanks for the  work @s0l0ist .

Because this logic is complex for the Front-end world, is there a chance to modify our  test cases accordingly so we can make sure that this is  being tested?

Options are:

- Write a separate new test
- Modify an existing test accordingly, so it can validate this passively

web/src/pages/AlertDetails/AlertDetailsInfo/AlertDeliverySection/AlertDeliverySection.tsx
s0l0ist: Not sure this is _quicker_, but definitely easier to read.

3nvi: Even if we have 10,000 deliveries, it will only  be slower by less than half a millisecond 😄

s0l0ist: yeah, we shouldn't ever be manipulating that large amount of data. I'm putting in the easier-to-read code.

3nvi (approved):
Looks  good @s0l0ist 🎉

PR #1695, opened 2020-10-08
Tests for tracking utils

### LINKED ISSUE #1663: Add testing for our analytics tracking utilities
### Description

Currently we use some basic analytics tracking utilities to track basic events if the user allows us to.

We should make sure that those utilities are tested correctly

### Related Services

Web

### Designs

N/A

### Acceptance Criteria

* Add `MIXPANEL_PUBLIC_TOKEN`  to `__tests__/.env.test`
* Globally mock `mixpanel` and `logError` for **all tests** in `__tests__/setup.ts`
* With `localStorage` key `ANALYTICS_CONSENT_STORAGE_KEY` set to `true`
  * `useTrackPageView` calls `mixpanel` with the correct arguments
  * `trackEvent` calls `mixpanel` with the correct arguments
  * `trackError` calls  `mixpanel` with the  correct arguments
* With `localStorage` key `ANALYTICS_CONSENT_STORAGE_KEY` set to `false`
  * `useTrackPageView` doesn't call `mixpanel` at all
  * `trackEvent` doesn't call `mixpanel` at all
  * `trackError` doesn't call `mixpanel` at all
* With `mixpanel` mock overridden to a function that always throws an error
  * `useTrackPageView` calls `logError` 
  * `trackEvent` calls `logError`
  * `trackError` calls `logError`

alexmylonas:
## Background

We should have tests for a couple of cases regarding our tracking utils. This includes enabling/disabling tracking when certain conditions are met, reporting to our error tracking when analytics tracking fails etc.

Closes #1663 

## Changes

- Added global mocks for `mixpanel` & `sentry`
- Updated local storage to have specific values before each test
- Added functionality to fail any test if there is an unexpected console.error

## Testing

- Test suite runs successfully


web/__tests__/setup.ts
```diff
@@ -28,6 +28,45 @@ import '@testing-library/jest-dom';
...
+
+// This mocks mixpanel module for all tests
+jest.mock('mixpanel-browser', () => {
+  const original = jest.requireActual('mixpanel-browser');
+  const { PageViewEnum, EventEnum, TrackErrorEnum } = jest.requireActual('Helpers/analytics');
+  return {
+    ...original,
+    init: jest.fn(),
+    // This is extra check for checking events & types passed are correct
+    track: jest.fn((eventName: string, data: any) => {
```
3nvi, line 53: doesn't TS already prevent that? CI would fail if you tried to track something wrongly since `npm run tsc` would complain. What **additional thing** are we gaining from this?

web/__tests__/setup.ts
```diff
@@ -91,11 +130,25 @@ beforeAll(() => {
...
+  jest.clearAllMocks();
   localStorage.clear();
   sessionStorage.clear();
+
+  // Keys are hardcoded since getting values from constants fails to run the test suite
+  localStorage.setItem('panther.generalSettings.errorReportingConsent', 'true');
+  localStorage.setItem('panther.generalSettings.analyticsConsent', 'true');
+
+  // Any console.error should fail the test
+  jest.spyOn(global.console, 'error').mockImplementation((...args) => {
```
3nvi, line 146: we already mock the `console.error`. I wouldn't want to mock it again. 

IMO you are are approaching errors in a weird way. Just do `throw new Error(...)`. I don't think forcing `console.error` to create an error is the correct approach.

web/src/helpers/__tests__/analytics.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+        trackError({ event: TrackErrorEnum.FailedMfa, src: SrcEnum.Auth, data: error });
+      };
+      const { getByText } = render(<TestComponent onClick={() => onClick()} />);
+
+      const btn = getByText('Button click');
+
+      fireEvent.click(btn);
+      await waitMs(50);
+      expect(localStorage.getItem(ANALYTICS_CONSENT_STORAGE_KEY)).toBeTruthy();
+      expect(mixpanel.init).toHaveBeenCalledTimes(3);
```
3nvi, line 63: please test the payload as well. Expect `toHaveBeenCalledWith(...)` for both sentry & mixpanel

3nvi (commented):
Thanks for that Alex. I feel we should also test the `useTrackPage` hook  (just a single small test).

Other than that, I just want to review again the test setup file and we are g2g

web/__tests__/setup.ts
alexmylonas: Yeap, you are correct i dont see any additional benefit from this removed it.

package.json
```diff
@@ -87,6 +88,7 @@
     "prettier": "^2.0.5",
     "react-dev-utils": "^10.1.0",
     "react-refresh": "^0.8.3",
+    "react-test-renderer": "^16.12.0",
```
alexmylonas, line 91: New testing libraries necessary for testing hooks properly

web/src/helpers/__tests__/analytics.test.tsx
alexmylonas: Added below for mixpanel, sentry was already accounted for.

web/src/hooks/__tests__/useTrackPageView.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+jest.mock('Helpers/Analytics');
+
+describe('useTrackPageView hook tests', () => {
+  it('should call trackPageView only once', async () => {
+    const { rerender } = renderHook(page => useTrackPageView(page), {
+      initialProps: PageViewEnum.LogAnalysisOverview,
+    });
+    rerender(PageViewEnum.LogAnalysisOverview);
+    expect(trackPageView).toHaveBeenCalledTimes(1);
+    expect(trackPageView).toHaveBeenCalledWith({ page: PageViewEnum.LogAnalysisOverview });
```
alexmylonas, line 32: This is checking that, we won't trigger the same page view event on rerendering

web/src/hooks/__tests__/useTrackPageView.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  it('should call trackPageView twice', async () => {
+    const { rerender } = renderHook(page => useTrackPageView(page), {
+      initialProps: PageViewEnum.LogAnalysisOverview,
+    });
+    expect(trackPageView).toHaveBeenCalledWith({ page: PageViewEnum.LogAnalysisOverview });
+    rerender(PageViewEnum.ComplianceOverview);
+    expect(trackPageView).toHaveBeenCalledWith({ page: PageViewEnum.ComplianceOverview });
+    expect(trackPageView).toHaveBeenCalledTimes(2);
```
alexmylonas, line 42: This is checking that, we **will** trigger a page view event on rerendering, if the page changes

alexmylonas (commented):
Addressed your feedback, please have a fresh look!

web/__tests__/setup.ts
```diff
@@ -91,11 +115,17 @@ beforeAll(() => {
...
  */
-afterEach(() => {
+beforeEach(() => {
+  // It important clearAllMocks to happen before updating local storage
+  jest.clearAllMocks();
   localStorage.clear();
   sessionStorage.clear();
+
+  // Keys are hardcoded since getting values from constants fails to run the test suite
+  localStorage.setItem('panther.generalSettings.errorReportingConsent', 'true');
```
3nvi, line 127: can we import  the related CONSTANTS  here instead of hardcoding the string?

package.json
3nvi: You never use this though  :S

web/src/helpers/__tests__/analytics.test.tsx
3nvi: 👍

web/__tests__/setup.ts
3nvi: 👍

web/src/hooks/__tests__/useTrackPageView.test.tsx
3nvi: Nice

web/src/helpers/__tests__/analytics.test.tsx
3nvi: 👍

web/src/hooks/__tests__/useTrackPageView.test.tsx
3nvi: 👍

web/src/hooks/__tests__/useTrackPageView.test.tsx
3nvi: 👍

3nvi (commented):
2 small small comments and we are g2g

package.json
3nvi: Hmm weirdly, this  is a [peer dependency]  of the testing library for hooks. Weird.

3nvi (approved):
Looks good @alexmylonas

PR #1696, opened 2020-10-08
Removed data from trackError for analytics

alexmylonas:
## Background

We shouldn't report error data to our product analytics.

## Changes

- Removed data property from `trackError`

## Testing

- Locally


3nvi (approved):
LGTM

PR #1698, opened 2020-10-08
Sync file with enterprise

giorgosp:
# Changes
- Sync some files to be same as in the Enterprise version to reduce conflicts
- Fix lambda permissions after moving the Glue create tables functionality from source-api to the datacatlog-updater lambda.

# Testing
Just deployed and tested table creation as well.


deployments/core.yml
```diff
@@ -864,18 +864,6 @@ Resources:
...
-          Statement:
-            - Effect: Allow
-              Action:
-                - glue:CreateTable
-                - glue:GetTable
-                - glue:UpdateTable
-              Resource:
-                - !Sub arn:${AWS::Partition}:glue:${AWS::Region}:${AWS::AccountId}:catalog
-                - !Sub arn:${AWS::Partition}:glue:${AWS::Region}:${AWS::AccountId}:database/panther*
-                - !Sub arn:${AWS::Partition}:glue:${AWS::Region}:${AWS::AccountId}:table/panther*
```
kostaspap, line 878: 👍

kostaspap: I think we don't need the Athena permissions anymore.

kostaspap (approved):

PR #1706, opened 2020-10-09
Support custom VPCs

nhakmiller:
## Background

This change adds support for specifying an existing EC2 VPC & Security Group while deploying Panther. We still create the load balancer, load balancer targets, subnets, and the security group the web application exists in.

If users want to specify an existing VPC to deploy into, they must specify the VPC ID, the Security Group ID of a Security Group in the VPC into which the load balancer will be deployed, and two valid (and available) subnets within the VPC for use by Panther. When we create the subnets we use /26's, but I suspect we could use  /28's (will test to see).

The user is responsible for ensuring the following:
- Traffic can be routed to the load balancer on port 443 from wherever you intend to access the load balancer from
- Traffic can be routed out from the web server to the internet (in order to hit public AWS API endpoints)

We would like to have this in for v1.10.0 so I have added the `sync-to-release` tag.

**IMPORTANT** when you specify the VPC, the load balancer is put in `private` scheme mode! This means the _public_ DNS name is resolved to the _private_ IP address. This means you can only access it if you are on a network that can route to the private IP address of the load balancer!

## Changes

- Allow users to specify a VPC ID, Security Group ID,  and two subnet IP ranges

## Testing

- Did a manual deployment into a dev account
- Planning to do an upgrade of an existing deployment (a deployment with an existing Panther created VPC)


deployments/bootstrap.yml
```diff
@@ -803,12 +838,14 @@ Resources:
   # Associate the route table that we have created (with the mapping right above) to our public
   # subnets in order to be used by it
   PublicSubnetOneRouteTableAssociation:
```
nhakmiller, line 840: We could rename all these to not include the word `Public` in the name since they would technically not be public anymore. But then existing deployments would have to re-deploy the whole networking stack which would be messy.

kostaspap: Not a big deal, agreed, but wouldn't CFN handle that during deployment though anyway?

kostaspap: Maybe you can update at least the description to explain that it actually "might not be " public.

rleighton:
Keep sure to test with snowflake

deployments/bootstrap.yml
```diff
@@ -74,10 +74,26 @@ Parameters:
...
+    Type: String
+    Description: Specify a SecurityGroup to deploy Panther into
+    AllowedPattern: '^sg-[0-9a-f]{10,}$'
+  SubnetOneIPRange:
+    Type: String
+    Description: Specify a SecurityGroup to deploy Panther into
+    AllowedPattern: '^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/([0-9]|[1-2][0-9]|3[0-2]))$'
+  SubnetTwoIPRange:
+    Type: String
+    Description: Specify a SecurityGroup to deploy Panther into
```
austinbyers, line 87: Descriptions here are invalid copy-pasta. Explain what the difference is in the two subnets (I'm not sure myself)

deployments/bootstrap.yml
```diff
@@ -74,10 +74,26 @@ Parameters:
...
+    AllowedPattern: '^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/([0-9]|[1-2][0-9]|3[0-2]))$'
   TracingMode:
     Type: String
     Description: Enable XRay tracing on GraphQL queries & mutations
     AllowedValues: ['', Active, PassThrough]
+  VpcID:
+    Type: String
+    Description: Specify a VPC to deploy Panther into
+    AllowedPattern: '^vpc-[0-9a-f]{10,}$'
```
austinbyers, line 96: Does this support both old- and new-style VPCs? I know the default format changed somewhat recently

deployments/bootstrap.yml
austinbyers: It should be ok to rename (CFN will just rebuild it for every deployment), and I could see users panicking when they see a "PublicSubnetOne" in their CFN resources when they deploy in a VPC, but on the other hand I'm not sure anyone would notice or care. I could go either way.

Since you plan to include this in 1.10, I'd say leave the name as is so we don't have to re-test the migration

Adding a comment seems sufficient

deployments/panther_config.yml
```diff
@@ -26,6 +26,14 @@ Infra:
   # Use 0.0.0.0/0 to allow unrestricted access
   LoadBalancerSecurityGroupCidr: 0.0.0.0/0
 
+  # These optional settings will allow you to deploy the panther load balancer into a security group
+  # And VPC of your choice. If you set VpcID, you must also set SecurityGroupID and ensure the
+  # subnet IP ranges are valid & available
+  VpcID: ''
+  SecurityGroupID: ''
+  SubnetOneIPRange: '172.31.250.0/26'
+  SubnetTwoIPRange: '172.31.251.0/26'
```
austinbyers, line 35: Should these be comments rather than actual defaults in the config file?

austinbyers (approved):
API gateway and AppSync are also both Internet facing and I know at least API gateway supports VPCs.

I assume we're just going to worry about that during the API migration? You explain in the config file the setting is only for the load balancer, which is accurate, so I think this is a good start.

deployments/bootstrap.yml
austinbyers: (even more important: check the descriptions in the `master` template)

nhakmiller: I'll update. There is no difference, application load balancers just require two subnets in different availability zones for availability reasons (being a load balancer & all I suppose).

deployments/panther_config.yml
nhakmiller: I picked these because they should be valid and available subnets in the default VPC in a region if nothing else is setup. But a comment is probably more appropriate,  will update.

austinbyers: Ah, well as long as these values don't trigger a change in an existing deployment (it doesn't look like they have any effect unless the VPC is specified), the way you have it now should be fine.

Are these also the default values in the master template, then? I'm sure users would appreciate not having to define an IP range if we already have one that we know will work

rleighton:
Please look at this first in enterprise:
https://github.com/panther-labs/panther-enterprise/blob/master/deployments/bootstrap.yml#L788

Make sure this stuff does not have a conflict. The Snowflake lambda needs subnets set up correctly since it is a VPC lambda.

nhakmiller:
> API gateway and AppSync are also both Internet facing and I know at least API gateway supports VPCs.
> 
> I assume we're just going to worry about that during the API migration? You explain in the config file the setting is only for the load balancer, which is accurate, so I think this is a good start.

Correct, those resources are internet facing but restrict access to only requests signed with IAM credentials from the Panther account if I recall correctly. Regardless, since those are going away soon I'm not worrying about them here.

nhakmiller:
> Please look at this first in enterprise:
> https://github.com/panther-labs/panther-enterprise/blob/master/deployments/bootstrap.yml#L788
> 
> Make sure this stuff does not have a conflict on merge.
> The Snowflake lambda needs private subnets set up correctly since it is a VPC lambda.

I will merge here (because I can't account for merge  issues in enterprise in  the  open source code base), but I won't just merge the  sync  PR. I'll check  it  out, update it to  ensure  the private subnets are  working, test  an  enterprise  deployment, then merge.

PR #1710, opened 2020-10-09
Redesign Rules page 1/3

vorillaz:
## Description
This is the initial implementation of the rules redesign page. Implements a basic card display with pagination. There is a minor issue with `threshold` missing from the `RuleSummary` API, @kostaspap is it possible to expand the API accordingly? 

## Designs:
https://share.goabstract.com/ec1f6697-cb30-4028-b5b4-9d7c89d68efe?mode=build&sha=7de7cfe171c9c39caad6d003179bc4580244610b

## Screenshots:
Screenshot 2020-10-09 at 8 16 45 PM
Screenshot 2020-10-09 at 8 16 50 PM

### Checklist:


austinbyers (commented):
Looks good! Do you think we should hide the threshold when its 0, so that we don't have "Threshold: 0" cluttering every card? I could go either way - it is a nice way to showcase the feature for users who might not know about it

web/src/pages/ListRules/ListRules.tsx
```diff
@@ -79,14 +76,13 @@ const ListRules = () => {
...
-          <ListRulesTable
-            items={ruleItems}
-            onSort={updateRequestParamsAndResetPaging}
-            sortBy={requestParams.sortBy || ListRulesSortFieldsEnum.Id}
-            sortDir={requestParams.sortDir || SortDirEnum.Ascending}
-          />
-        </Card>
+        <Panel title="Rules">
+          <Card as="section" position="relative">
+            {ruleItems.map(rule => (
```
3nvi, line 81: You need a wrapper around the `.map` to give vertical spacing to the cards.

Something like `<Flex direction="column" spacing={4}> ... </Flex>` or smth similar

3nvi (commented):
Looks good. Minor UI issues and we are good

web/src/pages/ListRules/ListRules.tsx
vorillaz: All the card items are consuming the `<GenericItemCard />` which already has everything in place.

web/src/components/cards/RuleCard/RuleCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+                  {rule.logTypes.map(logType => (
+                    <BulletedLogType key={logType} logType={logType} />
+                  ))}
+                </Flex>
+              }
+            />
+          </GenericItemCard.ValuesGroup>
+          <GenericItemCard.ValuesGroup>
+            <Flex ml="auto" mr={0} align="flex-end" spacing={2}>
+              <Box px={2}>
```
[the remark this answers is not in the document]
vorillaz: I am not sure about that or what do you mean, I wanted to add spacing for the outer container.

web/src/components/GenericItemCard/GenericItemCard.tsx
```diff
@@ -95,15 +96,15 @@ const GenericItemCardOptions = React.forwardRef<HTMLButtonElement>(function Gene
   );
 });
 
-const GenericItemCardValue: React.FC<GenericItemCardValueProps> = ({ label, value }) => {
-  const id = slugify(`${label}${value}`);
+const GenericItemCardValue: React.FC<GenericItemCardValueProps> = ({ label, value, id }) => {
+  const cardId = id || slugify(`${label}${value}`);
```
3nvi, line 100: nice touch

3nvi (approved):
Minor fix on card spacing and we are g2g.


Pre-approving 🎉

kostaspap:
> ## Description
> This is the initial implementation of the rules redesign page. Implements a basic card display with pagination. There is a minor issue with `threshold` missing from the `RuleSummary` API, @kostaspap is it possible to expand the API accordingly?

I now saw this. Yes, it would be possible to extend it. Either I can do this or @nhakmiller team since they own that part now. Whatever works.

PR #1713, opened 2020-10-09
Fix VPC input check

nhakmiller:
## Background

Fix a parameter validation bug.

## Changes

- Allow the empty string as an input for some optional fields

## Testing

- Deployed

### DIFF
--- a/deployments/bootstrap.yml
+++ b/deployments/bootstrap.yml
@@ -93,7 +93,7 @@ Parameters:
   VpcID:
     Type: String
     Description: An existing VPC to deploy Panther into
-    AllowedPattern: '(^vpc-[0-9a-f]{10,})?$'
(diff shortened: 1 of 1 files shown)


s0l0ist (approved):

austinbyers (approved):

deployments/bootstrap.yml
```diff
@@ -93,7 +93,7 @@ Parameters:
   VpcID:
     Type: String
     Description: An existing VPC to deploy Panther into
-    AllowedPattern: '^vpc-[0-9a-f]{10,}$'
+    AllowedPattern: '(^vpc-[0-9a-f]{10,})?$'
```
austinbyers, line 96: Isn't the `^` supposed to be outside the paren?

nhakmiller: I suppose it should work either way, I deployed it like this and it worked. I'll change it though.

PR #1720, opened 2020-10-12
Add version to all tracking events

alexmylonas:
## Background

All product analytics events should have an additional field about the system version the are generated

## Changes

- Added `STABLE_PANTHER_VERSION` as version property for mixpanel metadata

## Testing

- Locally


3nvi (approved):
LGTM

vorillaz (approved):
🥳

PR #1723, opened 2020-10-12
Add panther version check to opstools gluesync and gluerecover

rleighton:
## Background

The validation that the compiled gluesync matches the deployed version were lost when the new tools were made.

This is an important guard rail.

## Changes

- Add checks to `gluesync` and `gluerecover` that the compiled version matches deployed version.

## Testing

- mage test:ci
- I'll do a few more tests tomorrow where I test against different versions (have not had time just yet but will do).


alxarch (approved):
I think the check is not really needed for the `gluesync` tool, it might be needed for `gluerecover`. What `gluesync` does is just updating the descriptor tables. If we wanted to also change partition values it would be a _completely different problem_ that might not fit the scope of `gluesync` at all.

Even with the check in place we are not protected from version mismatches completely. It might be the case that user deployed a new (incompatible) version of panther and then some tables would have been created by an older incompatible version. What I would suggest we do is we add the *panther version as metadata to the partition itself* so we can handle this properly going forward (if the metadata is not there we handle the table as 'legacy'). 

I will approve the PR  as I understand the backward-breaking changes would cause problems.

Hopefully we can also add a `--no-version-check` flag or ENV VAR in a subsequent PR for the next release to help with using the tool on a deployment without having to mess with  build flags.

PR #1733, opened 2020-10-13
rule error function prefix

giorgosp:
- Add version to all tracking events (#1720)
- Tests for tracking utils (#1695)
- Redesign Rules page 1/3 (#1710)
- Add panther version check to opstools gluesync and gluerecover (#1723)
- Make StackOutputs() return error (#1721)
- Added warning (#1730)
- Include title() & dedup() output in API response when testing a Rule (#1625)
- Use camelCase for rule engine JSON responses
- Prefix rule result error message with the function that raised the error


internal/log_analysis/rules_engine/src/main.py
```diff
@@ -80,11 +80,10 @@ def direct_analysis(request: Dict[str, Any]) -> Dict[str, Any]:
...
                     'matched': rule_result.matched,
-                    'titleOutput': rule_result.title,
-                    'dedupOutput': rule_result.dedup_string,
-                    'errored': rule_result.exception is not None,
-                    'errorMessage':
-                        '{}: {}'.format(type(rule_result.exception).__name__, rule_result.exception) if rule_result.exception else None,
+                    'titleOutput': rule_result.title_output,
+                    'dedupOutput': rule_result.dedup_output,
+                    'errored': rule_result.errored,
+                    'errorMessage': rule_result.exception_pretty if rule_result.errored else None
```
kostaspap, line 86: Has this been discussed with FE? Are the ok with this? 

In my mind it might make more sense to have a `ruleException`, `titleException`, `dedupException` here? 
That could give us more flexibility and I think improve user experience. Currently, if the `title` & `dedup` methods both throw an exception, we cannot present it to the user at once. We will return 1 error, user fixes the `title` method, then they have to re-run the test to see the error of the `dedup` method.

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -72,10 +72,10 @@ def analyze(self, log_type: str, event: Dict[str, Any]) -> List[EngineResult]:
                     rule_tags=rule.rule_tags,
                     rule_reports=rule.rule_reports,
                     log_type=log_type,
-                    dedup=result.dedup_string,  # type: ignore
+                    dedup=result.dedup_output,  # type: ignore
```
kostaspap, line 75: nit: is the `#type: ignore` still needed?

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -50,7 +50,7 @@ def analyze(self, log_type: str, event: Dict[str, Any]) -> List[EngineResult]:
             result = rule.run(event)
             if result.exception:
                 # TODO(kostaspap): remove error logging once error reporting notification system is in place
-                self.logger.error('failed to run rule %s %s %s', rule.rule_id, type(result).__name__, repr(result.exception))
+                self.logger.error('failed to run rule %s %s %s', rule.rule_id, type(result).__name__, result.exception_pretty)
```
kostaspap, line 53: Line 51 should be changed to `if result. errored`?

internal/log_analysis/rules_engine/src/engine.py
giorgosp: Yes because `result.dedup_output` is `Optional[str]`

internal/log_analysis/rules_engine/src/engine.py
giorgosp: Shouldn't matter but updated it for consistency

internal/core/analysis_api/analysis/rule_engine.go
```diff
@@ -80,38 +80,33 @@ func (e *RuleEngine) TestRule(rule *models.TestPolicy) (*models.TestRuleResult,
 
 	// Translate rule engine output to test results.
 	testResult := &models.TestRuleResult{
-		TestsErrored: models.TestsErrored{},
-		TestsFailed:  models.TestsFailed{},
-		TestsPassed:  []*models.RulePassResult{},
+		TestSummary: true,
```
kostaspap, line 83: What is this field used for?

internal/log_analysis/rules_engine/src/main.py
```diff
@@ -53,40 +53,33 @@ def direct_analysis(request: Dict[str, Any]) -> Dict[str, Any]:
...
+
+        rule_result = test_rule.run(event['data'], raise_title_dedup=True)
+        results.append(
+            {
+                'id': event['id'],
+                'ruleId': raw_rule['id'],
+                'errored': rule_result.errored,
+                'matched': rule_result.matched,
+                'ruleError': format_exception(rule_result.rule_exception),
+                'titleOutput': format_exception(rule_result.title_output),
```
kostaspap, line 77: The output doesn't need the `format_exception` method invocation, right? Same for all similar.

internal/log_analysis/rules_engine/src/main.py
```diff
@@ -53,40 +53,33 @@ def direct_analysis(request: Dict[str, Any]) -> Dict[str, Any]:
...
+                'titleError': format_exception(rule_result.title_exception),
+                'dedupOutput': format_exception(rule_result.dedup_output),
+                'dedupError': format_exception(rule_result.dedup_exception),
```
kostaspap, line 80: Checking the code in `Rules` class, if the title returns an exception, we don't run the dedup function at all. Is this something we want?

internal/core/analysis_api/analysis/rule_engine.go
kostaspap: Maybe `TestsSummary` is more appropriate (summary of all the tests?)

internal/log_analysis/rules_engine/src/main.py
giorgosp: Added the respective fields (also discussed with Aggelos)

internal/core/analysis_api/analysis/rule_engine.go
giorgosp: Helper for the UI to quickly know if all the tests passed or not. No problem about the name, it is just consistent with the respective one in the policy test results.

internal/log_analysis/rules_engine/src/main.py
giorgosp: Good catch!

internal/log_analysis/rules_engine/src/main.py
giorgosp: Good question. It was working this way before but now that we have different fields for each of the functions (rule/title/dedup) there is no need to return early.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -111,24 +121,36 @@ def __init__(self, config: Dict[str, Any]):
...
+        In batch mode, title/dedup functions are not checked if the rule won't trigger an alert and also title()/dedup()
+        won't raise exceptions, so that an alert won't be missed.
         """
+        rule_result = RuleResult()
+        try:
+            rule_result.matched = self._run_command(self._module.rule, event, bool)
+        except Exception as err:  # pylint: disable=broad-except
+            rule_result.rule_exception = err
+
+        if batch_mode and not rule_result.matched:
```
kostaspap, line 138: Should be `if batch_mode or not rule_result.matched`.  Otherwise in the tests you will try run the `title` and `dedup` function when the rule has returned `False`.

kostaspap (approved):

internal/log_analysis/rules_engine/src/rule.py
giorgosp: Yes, this is on purpose. In tests, it is fine to run all the functions. But during analysis, run title and dedup only if the rule matches.

PR #1735, opened 2020-10-13
fix: use semver package to compare versions

### LINKED ISSUE #1722: Mage precheck has bug now that we are at version 1.10
### Describe the bug

Redeploying v1.10-RC1 fails with:
```
Error: trying to upgrade from v1.10.0-RC1 to v1.10.0-RC1 will not work - upgrade to v1.5.1 first
```

A "less than" test is not correct in mage precheck here:
https://github.com/panther-labs/panther/blob/master/tools/mage/deploy/deploy.go#L142

### Steps to reproduce

Deploy version 1.10-RC1, then deploy again.

s0l0ist:
## Background

Closes #1722. We do not need to have this comparison anymore.

## Changes

- Remove check

## Testing

- Manually tested version strings.

### DIFF
--- a/tools/mage/deploy/deploy.go
+++ b/tools/mage/deploy/deploy.go
@@ -157,7 +157,7 @@ func PreCheck(checkForOldVersion bool) error {
 			return fmt.Errorf("failed to parse repo version %s: %v", repoVersion, err)
 		}
 
-		if bVersion.LTE(oldVersion) {
+		if bVersion.LT(oldVersion) {
 			return fmt.Errorf("trying to upgrade from %s to %s will not work - upgrade to v1.5.1 first",
 				bVersion, rVersion)
 		}
--- a/tools/mage/deploy/deploy.go
+++ b/tools/mage/deploy/deploy.go
@@ -27,7 +27,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/service/lambda"
-	"github.com/blang/semver"
 	"github.com/magefile/mage/sh"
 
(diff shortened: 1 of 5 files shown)


tools/mage/deploy/deploy.go
```diff
@@ -139,9 +140,26 @@ func PreCheck(checkForOldVersion bool) error {
 		if err != nil {
 			log.Warnf("failed to describe stack %s: %v", cfnstacks.Bootstrap, err)
 		}
-		if bootstrapVersion != "" && bootstrapVersion < "v1.4.0" {
+		bVersion, err := semver.ParseTolerant(bootstrapVersion)
+		if err != nil {
+			return fmt.Errorf("failed to parse bootstrap version %s: %v", bootstrapVersion, err)
+		}
+
+		oldVersionString := "v1.4.0"
```
austinbyers, line 148: Honestly, this is an old enough version I'd say let's remove this check completely. We don't have any customers before v1.4, and if someone hasn't upgraded since then, support is really only best-effort anyway

austinbyers (changes requested):
Let's remove this check completely - v1.4.0 is so old by this point it's not worth even checking for it in my opinion

tools/mage/deploy/deploy.go
nhakmiller: Seems reasonable. They can still do partial updates, they just aren't warned about it.

nhakmiller (approved):
I'm good either way on Austin's comment. A footnote in the release notes may be worthwhile either way.

austinbyers (approved):
Thanks!

Issue #1711, opened 2020-10-09
Consolidate Panther Versioning


austinbyers:
### Description

We have some kind of Panther version in all of the following places:

* The published master template
* GitHub
* CloudFormation custom resources
* The web app (General Settings)
* Auxiliary templates
* dev/CI Docker images
* Documentation

Problems with our current approach:

* The git commit SHA is not shown when deploying the master template. When troubleshooting a deployment, it would be good to know exactly what commit was deployed, because the release tag sometimes changes. Also, the tag in the master stack may not match the tagged release (see #1587)
* We have to manually increment the version in the master template and the documentation
* Auxiliary templates and docker images have versions which are only incremented when they change - if a dev changes the deployment role but forgets to update the version number, it will overwrite the existing published version

When possible, we would like to load the same version string everywhere in the repo and not have to manually change it in the code.

### Acceptance Criteria

- Remove the hardcoded version from the `master` and auxiliary templates
    - `mage` will inject the version to the templates automatically before they are published
- Remove hardcoded version from documentation: in enterprise, `mage doc` will embed the latest stable release number in the quick-start docs, depending on what branch is being built
- Panther web app will always show the commit in the General Settings, e.g. "v1.9.0-0-g123abc"

PR #1737, opened 2020-10-13
fix: version bump for templates

s0l0ist:
## Background

Update version to match latest v1.10.0-RC2 release

## Changes

- Update master template version

## Testing

- N/A


austinbyers (approved):
Once #1711 is implemented, this sort of PR won't be necessary

PR #1741, opened 2020-10-14
chore: update version

s0l0ist:
## Background

Update Panther version

## Changes

- update master template

## Testing

- N/A


austinbyers (approved):
Not really necessary, since we always publish from a release branch anyway, but thanks! Nice for developers to see the tag at least so they're not confused

PR #1751, opened 2020-10-15
Fix gluesync and gluerecover prefix match

alxarch:
## Background

Prefix matching was not working properly in `gluesync` and `gluerecover`. While the documentation stated that filtering was using regular expressions, the expression uses `glob` pattern syntax. Also adds a `-skip-version-check` flag to both commands to help developers override version checks.

## Changes

- Fix pattern matching for table names in `gluesync`, `gluerecover`
- Adds `-skip-version-check` flag to both commands to help developers override version checks.

## Testing

- run command in dry-run mode


cmd/opstools/gluesync/gluesync/main.go
```diff
@@ -39,24 +39,26 @@ var (
...
-		NumWorkers:     flag.Int("workers", 8, "Number of parallel workers for each table"),
-		Prefix:         flag.String("prefix", "", "A prefix to filter log type names"),
+		DryRun:           flag.Bool("dry-run", false, "Scan for partitions to sync without applying any modifications"),
+		Debug:            flag.Bool("debug", false, "Enable additional logging"),
+		Region:           flag.String("region", "", "Set the AWS region to run on"),
+		MaxRetries:       flag.Int("max-retries", 12, "Max retries for AWS requests"),
+		MaxConnections:   flag.Int("max-connections", 100, "Max number of connections to AWS"),
+		NumWorkers:       flag.Int("workers", 8, "Number of parallel workers for each table"),
+		Prefix:           flag.String("prefix", "", "A prefix to filter log type names"),
+		SkipVersionCheck: flag.Bool("skip-version-check", false, "Skip panther version check"),
```
kostaspap, line 61: I'm not sure this is what we want. What we want is to make sure that the version check works properly always, not skip it.

rleighton: There is no reason I can think of to add a skip version check. Dev's can always build to the version they have deployed.

rleighton:
The issue with Earnin, @kostaspap found was due to the data catalog updater being on an older version than the customer resource. Can you add a dependency here https://github.com/panther-labs/panther/blob/master/deployments/log_analysis.yml#L134 on the data catalog updater pls

cmd/opstools/gluesync/gluesync/main.go
alxarch: When I deploy panther to test e2e, I want to be able to run the tool from my current working tree without having to:
- build for 5 os/arch combinations each time
- having to revert my local changes to the ones currently deployed so the version match
- I frequently deploy 'custom' versions from the ECS instance with ad-hoc branches merged just for the deploy.
- this gets even worse when having 2-3 deployments you want to test this against

That being said since I am the only one who finds the version check annoying, I can just comment it out and use `go run` when it gets in my way.

kostaspap:
> PR looks good.
> We need to address @rleighton comment as well. Not sure if this will be done as part of this PR or can be done in separate.

I'll make a separate PR for @rleighton issue

PR #1758, opened 2020-10-15
Backport required changes for implementing custom logs.

alxarch:
## Background

The custom logs feature was directly merged onto the enterprise release branch for 1.10.
It included a number of changes that need to be backported to OSS.

## Changes

- Adds `logtypes.Resolver` interface for resolving `logtype.Entry` at runtime
- Refactor log processor to use `logtypes.Resolver`
- Decouple S3 destination from `registry` module
- Update data catalog updater to create tables based on resolved log types
- Update custom resource handler to use the updated code when creating Athena views
- Fix validation for null types
- Fix problems with apigen/lambdamux

## Testing

- mage test
- e2e testing (pending)


kostaspap (approved):
Hard to say if these are all the changes that we need to add. 
The changes look good though.

PR #1764, opened 2020-10-15
Aggelos fix overview tables

3nvi:
## Background

In Cloud Security overview page, some data cause the tables to overflow. This PR makes sure to force words to wrap( since we have stuff like ARNs which is a **single big word** that can easily overflow) whenever they exceed the available width

## Changes

- Add `word-break: break-all` to links that have large texts

## Testing

- Visually


alexmylonas (approved):
LGTM!

PR #1766, opened 2020-10-15
General settings redesign

vorillaz:
## Description

Align redesign for General Settings. Also, fixed a bug with a misaligned spacing.

### Visual Bug:
Screenshot 2020-10-15 at 7 52 10 PM

### Screenshots:
Screenshot 2020-10-15 at 7 46 41 PM

### Designs:
https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/39dcb3e2-1e72-49b2-afab-3a59c8459f66/commits/fc0e3669ae46aad9ff03e270ad13f07d58e1b9b9/files/8cc3cb34-8b80-405f-97ad-3adafc365a4f/layers/5A875BE7-2743-49D5-AB60-BCC80722F71D

### Checklist:


web/src/components/Layout/Layout.tsx
```diff
@@ -25,11 +25,13 @@ const Layout: React.FC = ({ children }) => {
...
-      <Box as="main" minHeight={1} flex="1 0 auto" bg="navyblue-600" zIndex={0}>
-        <Box width={1214} mx="auto">
+      <Box as="main" minHeight={1} flex="1 0 auto" bg="navyblue-600" zIndex={0} position="relative">
+        <Flex width={1214} mx="auto" minHeight="100%" direction="column">
           <Header />
           {children}
-        </Box>
+        </Flex>
+        {/* Sticky footer used from some components */}
+        <Box id="footer" position="absolute" width="100%" bottom={0}></Box>
```
3nvi, line 34: if it's sticky, shouldn't it be `position: fixed`?

This would also save you the trouble of adding the `Flex` component above, as well as the `position="relative"` prop

web/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Box, Flex } from 'pouncejs';
+
+const Footer: React.FC = ({ children }) => {
+  return ReactDOM.createPortal(
+    <Box bg="navyblue-500">
+      <Flex width={1214} mx="auto" minHeight="100%" direction="column" py={6}>
```
3nvi, line 26: why don't we remove the `Box` and add the `bg` to the `<Flex>` component?

web/src/components/Layout/Layout.tsx
3nvi, line 34:
```suggestion
        <Box as="footer" id="footer" position="absolute" width="100%" bottom={0}></Box>
```

web/src/pages/GeneralSettings/GeneralSettings.tsx
```diff
@@ -71,66 +73,77 @@ const GeneralSettingsPage: React.FC = () => {
...
                 variant="outline"
               >
                 Change
               </LinkButton>
-            </Flex>
+            </Box>
             <Box as="section">
               <Box color="navyblue-100" fontSize="small" mb={1}>
-                Version
+                AWS Account ID
```
3nvi, line 127: optional: should we add some minor accessibility here?

- `aria-*` attributes to relate labels with values
- Perhaps a better tag on the labels? Like `h4`? We could even use `dd, dt, dl`

web/src/components/Footer/Footer.tsx
3nvi: optional: I think we could omit the hardcoded `1214` since we can just "center" the content with `justify="center"` and be done with it

3nvi (commented):
Nice work @vorillaz 

Can we add a few tests for it? A single snapshot test + another that validates tthat the info is there would do just fine :)

Lastly, I'm not 100% sure about the button placement on the "change". I would perhaps put it underneath the logo, but perhaps @dimitriszoitas can help us a bit more with this

Also I feel tthat the `<Box id="footer">` should also be part of the  `__tests__/render` function

web/src/components/Layout/Layout.tsx
vorillaz: Nope, `fixed` would not respect the `z-index` values. Furthermore, we should "stick" the footer to the bottom of the page and restrict the view inside the view content of the page.

web/src/components/Footer/Footer.tsx
vorillaz: @3nvi The option for 1214 width comes from this part 
https://github.com/panther-labs/panther/blob/a2b0701e07c1bce6cde612b4fdcb72d7469c2ff2/web/src/components/Layout/Layout.tsx#L29-L31

I originally thought of centering the content but according to the mockups the content is expanded to the width of the container 😒

web/src/components/Layout/Layout.tsx
vorillaz: Thanks for this one, fixed.

web/src/pages/GeneralSettings/GeneralSettings.tsx
vorillaz: You are right, fixed.

web/src/pages/GeneralSettings/GeneralSettings.tsx
```diff
@@ -71,66 +73,85 @@ const GeneralSettingsPage: React.FC = () => {
...
+            </Box>
+            <Box as="section">
               <LinkButton
                 external
+                size="medium"
                 to="https://runpanther.io/pricing/"
                 variantColor="navyblue"
                 variant="outline"
               >
                 Change
```
3nvi, line 124: @dimitriszoitas mentioned we should change it
```suggestion
                Get Enterprise
```

3nvi: Can we also entertain the idea of putting it under the label? If it doesn't fit like that, perhaps with a `size="small"`?

3nvi (approved):
Just a couple of comments regarding the tests & the button.

I  don't feel strongly about the button, but I  kind of not like it where it is. Given the fact this is not laid out in the acceptance criteria, I can pre-approve this PR.

PR #1773, opened 2020-10-16
Fix AppSync for ListAvailableLogTypes

alxarch:
## Background

ListAvailableLogTypes was not working after backporting cutomlog changes. (The change was missed in the sea of changed files)

No need to sync to release, already fixed there

## Changes

- Forces empty object (`{}`) payload for `ListAvailableLogtypes` endpoint in AppSync resolver

## Testing

- Deployed and tested that FE can retrieve available log types

### DIFF
--- a/deployments/appsync.yml
+++ b/deployments/appsync.yml
@@ -1902,7 +1902,7 @@ Resources:
           "version" : "2017-02-28",
           "operation": "Invoke",
           "payload": $util.toJson({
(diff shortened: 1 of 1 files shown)


kostaspap (approved):

kostaspap:
Yup, seems about right in the 1.10 release: https://github.com/panther-labs/panther-enterprise/blob/release-1.10/deployments/appsync.yml#L2742

PR #1782, opened 2020-10-16
[CircleCI] Fail without commit in Autofmt  and inform users when a PR is from a fork

alxarch:
## Background

We recently had a PR from an external contributor. Their code was fine but probably had some minor formatting issues. 
This triggered a push in the `Autofmt` step. That cannot succeed for PRs that originate from a fork.
Fortunately for us, CircleCI makes it easy to detect PRs that come from forks using the `CIRCLE_PR_USERNAME` env var.
This PR changes the `Autofmt` step to fail with informative output to the user telling them to run `mage gen fmt`  and update the PR.

## Changes

- Check if a PR is from a fork during `Autofmt` and warn users.

## Testing

- Tested the changes in local bash
- Need help on how to test this on CircleCI


.circleci/config.yml
```diff
@@ -90,7 +90,25 @@ commands:
...
+            fi;
+
+            # CircleCI docs state that when a PR is from a fork this env variable is set
+            # https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
+            if [[ -v CIRCLE_PR_USERNAME ]]; then
+                # We cannot commit and push the differences automatically, inform the user.
+                echo "Hey there $(CIRCLE_PR_USERNAME)!"
+                echo "Thanks for your contribution to Panther."
+                echo "We tried to auto-generate and format your code and it resulted in file differences."
+                echo "Please run 'mage gen fmt' on your local repo, commit and push the changes, so it can be processed by our CI pipeline."
```
austinbyers, line 108: Note that in enterprise, this becomes `mage gen doc fmt` (just FYI for when you sync the change)

austinbyers (approved):
> Need help on how to test this on CircleCI

You can verify the autofmt still works for Panther developers by pushing a malformed file and seeing if panther-bot still fixes it (it looks like that already happened, actually)

As for testing the external PR, the only real way is to merge this and then try it!

.circleci/config.yml
austinbyers: Although I guess enterprise can't have external contributors anyway, so this isn't necessary

PR #1784, opened 2020-10-16
Handle scanning large resources better

### LINKED ISSUE #1771: CloudSecurity scanning fails on large resources
### Describe the bug

CloudSecurity stores resource snapshots in DynamoDB, which has a max resource row size of 400Kb. If a JSON representation of  a resource is larger than 400Kb, then it gets dropped and logs an error.

### Steps to reproduce

Create a very large resource (VPCs with  lots of security  groups and network acls tend to be large).
Scanning will fail when calling the `panther-resources-api` due to the resource's large size.

### Expected behavior

Scanning should handle large resources more gracefully.

### Environment

- Panther version 1.10.1

### Additional context

EC2 VPC resources are the only resources we've observed to breach the threshold, which inline multiple other resources as a convenience. Since we've supported resource lookups in the policy engine since before v1.0 even, we could drop these resource inlines. This would break any existing policies that rely on the embedded resource,  but the fix is a very minor change in the policy.

nhakmiller:
## Background

This is the short term fix which closes #1771. We may consider larger fixes, at the very least it will be an ongoing effort.

This change removes embedded security group and network acl resources from the VPC resource, replacing them instead with just a list of IDs of security groups and network acls. Using the `resource_lookup`  functionality in the policy engine, policies that are evaluating  the contents of a VPCs security groups/network acls can still access this information. Additionally, I pulled up the default security group and default network acl IDs for ease of reference.

To address the problem more generally, I added some utilities to calculate the size of a dynamo item before writing it to dynamo and put a safeguard in the `resources-api` to not attempt to write to dynamo on oversized resources. Instead, a  warning message with some useful identifying information is logged.

I considered adding this protection in the `dynamodbbatch.BatchSend` function as well to protect other functions, but since that is shared across several functions I decided I didn't want to potentially introduce a bug there that could affect many services. I could still go either way on it.

I've applied the `sync-to-release` tag in order to get this bug fix out to affected customers more quickly, but if any reviewers think that's inappropriate let me know and we can discuss it.

## Changes

- Removed embedded network  ACL and security group resources from the VPC resource
- Added protections to the `resources-api` to prevent attempting to write oversized resources to dynamo
- Added utilities to calculate dynamo item size

## Testing

- Deployed to dev account, working on creating a 400Kb+ object to scan
- Manually tested the dynamodb math against an online calculator


pkg/awsbatch/dynamodbbatch/item_size.go
```diff
@@ -0,0 +1,107 @@
...
+	//
+	// "Numbers are variable length, with up to 38 significant digits. Leading and trailing zeroes
+	// are trimmed. The size of a number is approximately
+	// (length of attribute name) + (1 byte per two significant digits) + (1 byte).
+	//
+	// So to estimate that approximation, I just divide the length by 2 (to approximate significant
+	// digits) and add 1.
+	//
+	// Additionally, dynamo rounds up on the significant digits / 2 math, so we add .5 then cast to
+	// an int to round up
```
austinbyers, line 85: Casting to an int does *not* round up: https://play.golang.org/p/DNrddvQBJRe

```go
func main() {
	x := 2.0 + 0.3
	fmt.Println("Hello, playground", int(x))
}
```

> Hello, playground 2

In fact, it will actually always round down (it's a truncation operation). So `int(x + 1)` would round to the next highest integer. But then, `int(x + 1)` would also "round" 2.0 to 3.0

This may be fine for the purposes of your estimate here, but if you actually want a ceiling, consider using `math.Ceil` instead: `int(math.Ceil(x))`

austinbyers (commented):
Great job on the quick fix!

Instead of estimating ahead of time, an alternate strategy would be to ask for forgiveness rather than permission, and simply check the Dynamo error code to see if the item was too large. This will be faster in the normal case (the item fits) because we don't have to calculate anything, but slower in the bad case (the item is too big) because we make an unnecessary API call.

I imagine the size calculation is very fast (it's just arithmetic), so what you have seems like a good solution to me! And estimating Dynamo size could have other applications as well.

But I agree we should not add it by default for *every* Dynamo batch write operation

pkg/awsbatch/dynamodbbatch/item_size.go
nhakmiller: Sorry I did not explain well, this is actually relying on truncation.  What I meant is:

If the value is 7 and you divide by 2 then you get 3.5. When calculating size, Dynamo rounds this up to 4 but casting to an int rounds it down to 3 so we have a mismatch.

If instead of just dividing by two we divide by two then add .5, we get 7/2 = 3.5 + .5 = 4,  which is the same value dynamo will calculate. For even values like 8, we divide by 2 to get 4 add .5  to get 4.5, when we then cast to an int to get 4 which is again the value dynamo will calculate.

I'm just using "add .5 then cast to an int" as shorthand for round up without needing to import `math` for the `math.Ceil` function because I know the only values I will have are whole numbers and whole numbers + .5. If we're worried about floating point precision or "being too clever" I can just import `math.Ceil` instead. At the very least I can clarify the comment.

nhakmiller:
> Instead of estimating ahead of time, an alternate strategy would be to ask for forgiveness rather than permission, and simply check the Dynamo error code to see if the item was too large.

I originally did exactly this, but the error logging happens in the shared `BatchWriteItem` function so even if we added better error handling in the `resources-api` we'd still get paged every time it happens.  And I didn't want to remove error logging from the shared code because I'm not sure what all other systems rely on the `BatchWriteItem` function to handle all error logging.

Even then, I was going to just go track down all the places that `BatchWriteItem` got called and make sure they're doing appropriate error handling themselves, but then I realized we wouldn't be able to know exactly which item failed by the time we returned so making a useful error message would be difficult. It was at that point that I decided to give up and  go with the pre-calculation. Definitely a tricky situation, and I blame AWS for making it super complicated to work with dynamo sizes 🤦

pkg/awsbatch/dynamodbbatch/item_size.go
austinbyers: Ah, got it. Clarifying the comment works for me!

lindsey-w (commented):
Looks good to me!

pkg/awsbatch/dynamodbbatch/item_size.go
```diff
@@ -0,0 +1,107 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+)
+
+// GetDynamoItemSize calculates the size that dynamo considers the item to be
+func GetDynamoItemSize(item map[string]*dynamodb.AttributeValue) int {
```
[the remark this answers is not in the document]
nhakmiller: Added some basic unit tests, it was hard to validate the correct answers for the bigger entries but I compared my results to an online calculator and assumed they were generally correct.

austinbyers (approved):

PR #1785, opened 2020-10-16
Create shared VERSION file

austinbyers:
## Background

Every new release, we have to remember to update a panther version string in several different places: aux templates, master template, docs, etc:

*  `git describe --tags` for the version when deploying from source
* Hardcoded template version for aux templates
* Hardcoded template version for master template
* Hardcoded version for quick start docs (enterprise repo)

Besides requiring an annoying number of small version PRs, the hardcoded master template version means we can't see the deployed commit like we can when building from source.

With this change, all 4 of the above will automatically load the version from the same place - a single top-level VERSION file. This is used to build a short and a long version string:

* Short version: "v1.10.1" - this is the version used for S3 template path, docker tag, doc link, etc, exactly the same as before
* The full version string (shown in CFN stack tags and the general settings page) now looks like `v1.10.1-master-ff2c3abc-dirty`, where:
    * `v1.10.1` is the short base version
    * `master` is the branch name
    * `ff2c3abc` is the 8-character commit ID
    * `-dirty` is added as a suffix if there are local uncommitted changes

----

In summary, these changes mean less work for release engineers and more detailed and consistent version information for all deployments.

The git release tag is no longer relevant for the deployment, which solves a number of its own issues as well - for example, the release tag does not have to be pulled back into `master` for developers to see the new version number

Closes: #1711 

## Changes

- Add VERSION top-level file
- Remove explicit version strings from all CFN templates
- `master:publish` now requires a clean repo state and a `release-1.X` branch

## Testing

- `mage clean setup deploy`
- `mage master:deploy` (in a different region)
- `make upload-all` (aux templates)

The version looks correct in the web app general settings, doc link, and the sidebar:

Screen Shot 2020-10-16 at 12 16 42 PM
Screen Shot 2020-10-16 at 12 16 57 PM


s0l0ist:
Excellent! Would this show something like `v1.10.1-release-123456` for official releases?

tools/mage/master/publish.go
```diff
@@ -43,19 +43,15 @@ func Publish() error {
 		return err
 	}
 
-	version, err := GetVersion()
-	if err != nil {
-		return err
+	version := util.RepoVersion()
+	if strings.HasSuffix(version, "-dirty") {
+		return fmt.Errorf("you have local changes; commit or stash them before publishing")
```
s0l0ist, line 48: nice

rleighton (approved):
LGTM. Big improvement.

austinbyers:
> v1.10.1-release-1.10-123456 for official releases?

Correct. A bit redundant perhaps, but I like being able to see that the official release really was created from the correct release branch

PR #1790, opened 2020-10-19
Alerts API now can filter the alerts by type (Rule matches or Rule errors)

### LINKED ISSUE #1749: Alerts API should allow filtering based on Alert type (rule error or rule match)
### Description

Alerts API should allow filtering based on Alert type (rule error or rule match)

### Related Services

panther-alerts-api

### Acceptance Criteria

A concise list of specific user stories that qualify this story as done.

- Alerts API supports an "Alert type" filter that allows retrieving all alerts that are related to rule errors
- Alerts API supports an "Alert type" filter that allows retrieving all alerts that are related to rule errors for a specific Rule ID

kostaspap:
## Background
Closes #1749 

Alerts API now can filter the alerts by type (Rule matches or Rule errors). We introduce a new "Type" parameter that lets the users select the API to return `RULE` or `RULE_ERROR` (in the future it could add one filter for policy @s0l0ist )

## Changes

- Added the ability to filter alerts by type

## Testing

- mage test:ci
- Tested E2E (tested also backwards compatibility)

### DIFF
--- a/api/lambda/alerts/models/api.go
+++ b/api/lambda/alerts/models/api.go
@@ -57,6 +57,7 @@ type GetAlertOutput = Alert
 // {
 //     "listAlerts": {
 //         "ruleId": "My.Rule",
+//	    "type" : "RULE_ERROR",
 //         "pageSize": 25,
 //         "exclusiveStartKey": "abcdef",
 //         "severity": ["INFO"],
@@ -82,7 +83,8 @@ type ListAlertsInput struct {
 	ExclusiveStartKey *string `json:"exclusiveStartKey"`
 
 	// Filtering
-	Severity        []*string  `json:"severity" validate:"omitempty,dive,oneof=INFO LOW MEDIUM HIGH CRITICAL"`
+	Type            string     `json:"type" validate:"omitempty,oneof=RULE RULE_ERROR"`
+	Severity        []string   `json:"severity" validate:"omitempty,dive,oneof=INFO LOW MEDIUM HIGH CRITICAL"`
 	NameContains    *string    `json:"nameContains"`
 	Status          []string   `json:"status" validate:"omitempty,dive,oneof=OPEN TRIAGED CLOSED RESOLVED"`
 	CreatedAtBefore *time.Time `json:"createdAtBefore"`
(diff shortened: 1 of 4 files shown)


api/lambda/alerts/models/api.go
```diff
@@ -82,7 +83,8 @@ type ListAlertsInput struct {
 	ExclusiveStartKey *string `json:"exclusiveStartKey"`
 
 	// Filtering
-	Severity        []*string  `json:"severity" validate:"omitempty,dive,oneof=INFO LOW MEDIUM HIGH CRITICAL"`
+	Type            string     `json:"type" validate:"omitempty,oneof=RULE RULE_ERROR"`
+	Severity        []string   `json:"severity" validate:"omitempty,dive,oneof=INFO LOW MEDIUM HIGH CRITICAL"`
```
kostaspap, line 87: Changed the type from `[]*string` to `[]string` in order to simplify it

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -299,6 +301,26 @@ func filterByLogType(filter *expression.ConditionBuilder, input *models.ListAler
...
+		case alertdeliverymodels.RuleType:
+			// Alerts for rule matches don't always have the attribute specified
+			multiFilter = expression.
+				Equal(expression.Name(TypeKey), expression.Value(input.Type)).
+				Or(expression.Name(TypeKey).AttributeNotExists())
+		default:
+			panic("Uknown type :" + input.Type)
+		}
+
+		*filter = filter.And(multiFilter)
```
giorgosp, line 320: minor: Would it read better if `filterByType` returned the multifilter and the called would `And` the filters? Although I see that we follow this pattern in the other filter methods.

giorgosp (approved):

internal/log_analysis/alerts_api/table/list.go
kostaspap: That's a good callout, I also don't like that we modify the filter that is passed as an argument. 
I will give it a shot but if it takes too much time to refactor I will leave out for the next time.

giorgosp: Sure. I'm a fun of TODOs as well :D

kostaspap: It ended up being more messy. I will skip for now, but added a todo

PR #1792, opened 2020-10-19
Add maxbuffer limit check to log processor

rleighton:
## Background

Running this [benchmark](https://summitroute.com/blog/2020/10/09/public_dataset_of_cloudtrail_logs_from_flaws_cloud/) data caused the log processor to OOM and/or timeout. 

The issue is that these are not real CloudTrail files in the sense they were produced operationally. They are dense with events that span a wide range of time. One file has over 12k unique hours. While this is not realistic for CloudTrail, agent logs like Crowdstrike often have messy time distributions so fixing this does have value. The fix is to add a missing limit on maximum number of in memory buffers and force rapid flushing to S3 when exceeded. 

After this fix (note: api count does not match, but perhaps they define it differently):

image

## Changes

- add maxbuffer limit and trigger flush of buffers if exceeed
- minor refactoring

## Testing

- mage test:ci
- deployed and ingested benchmark data
- ran with future change in log processor aggregation (PR soon)


internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -320,19 +326,26 @@ type s3EventBufferSet struct {
 	totalBufferedMemBytes uint64 // managed by addEvent() and removeBuffer()
 	set                   map[time.Time]map[string]*s3EventBuffer
 	stream                *jsoniter.Stream
+	maxBuffers            int
+	maxBufferSize         int
+	maxTotalSize          uint64
 }
 
-func newS3EventBufferSet(jsonAPI jsoniter.API) *s3EventBufferSet {
+func newS3EventBufferSet(destination *S3Destination, maxTotalSize int) *s3EventBufferSet {
```
rleighton, line 334: Minor refactor to pass the destination in that has the values needed for `writeEvent()`.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -354,23 +367,34 @@ func (bs *s3EventBufferSet) writeEvent(event *parsers.Result, maxBufferSize, max
...
-		return buf, nil
+		sendBuffers = append(sendBuffers, buf)
 	}
 	// Check if bufferSet is bigger than threshold for total memory usage
-	if bs.totalBufferedMemBytes >= uint64(maxTotalSize) {
-		if buf := bs.largestBuffer(); buf != nil {
-			bs.removeBuffer(buf)
-			return buf, nil
+	if bs.totalBufferedMemBytes >= bs.maxTotalSize {
+		if largestBuffer := bs.largestBuffer(); largestBuffer != nil {
```
rleighton, line 391: Minor refactor, change name of var to avoid shadowd `buf` var.

serbantanasa:
Can the buffer limit exceeded-triggered flush above cause data loss in this scenario?

rleighton:
> Can the buffer limit exceeded-triggered flush above cause data loss in this scenario?

No, what this change does is force the flushes to happen concurrently during processing rather than wait until the end of the processing (because they are small and would not trigger early flushing, but we have tens of thousands of them and waiting until the end of processing would not allow enough time ... or the sum of all of these tiny partitions would cause an OOM).

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -354,23 +367,34 @@ func (bs *s3EventBufferSet) writeEvent(event *parsers.Result, maxBufferSize, max
 	if err != nil {
 		return nil, err
 	}
+	// Check if bufferSet has too many entries
+	if len(bs.set) > bs.maxBuffers {
+		// The hope is most of the flushed buffers were done updating (as events often come roughly in time order)
+		bufferReduction := bs.maxBuffers / 2
+		removeBuffers := func(buffer *s3EventBuffer) error {
+			if len(sendBuffers) >= bufferReduction {
+				return errors.New("stop") // this just makes apply() stop
```
kostaspap, line 376: Why not just `return nil`?

rleighton: The apply() function only stops looping  on an error of the function applied. I could change it to return also a boolean to stop and not rely on an error: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/destinations/s3.go#L428 .

We want to stop after collecting `bufferReduction ` number of buffers.

kostaspap: Yeah, I think that would make more sense. Using error as a control variable doesn't seem right, would lead to unexpected consequences as the code evolves.

kostaspap (approved):

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Done.

Issue #1772, opened 2020-10-15
Allow individual Lambda deploys


austinbyers:
### Description

@kostaspap and other developers have asked for a way to use `mage` to deploy one Lambda at a time to save time deployment/debugging.

It's already possible to deploy individual stacks: `STACK=log-analysis mage deploy`, but deploying individual functions would be faster still.

### Related Services

- `mage`

### Acceptance Criteria

- It should be possible to update the source for an individual Lambda function, e.g. `LAMBDA=panther-resources-api mage deploy`

### Implementation Notes

CloudFormation doesn't support targeted updates like Terraform does. So the simplest approach for this is to avoid CFN completely: `mage` can directly update the Lambda source code

The only question then, is whether to update just the function source or the configuration as well: environment variables, memory/timeout, event triggers, IAM policy, etc. Mage could theoretically parse the CFN and try to apply some or all of these parameters directly, but at that point we're re-implementing SAM/CloudFormation for Lambda functions.

During a rapid development cycle, I assume it's typically only the source code which is being modified, so we can just start with that

PR #1795, opened 2020-10-19
Allow individual Lambda deploys

austinbyers:
## Background

Developers have asked for a way to deploy individual Lambda functions directly for the fastest developer iteration, so now you can:

```
LAMBDA=log-processor dev -- mage deploy
```

Closes: #1772 

## Changes

- `mage deploy` checks for the `LAMBDA` environment variable, in the same way it was already checking for the `STACK` variable

If found, it will update the given Lambda function's source code directly. No CloudFormation, and to my suprise, no S3 required. Note that only the lambda *source* is updated, not the configuration.

## Testing

- Deployed all 28 Lambda functions directly - no errors and the product still functions normally with no alarms

When this merges to enterprise, I'll update the development docs and will handle the function that we have to cross-compile


tools/mage/deploy/deploy.go
```diff
@@ -169,6 +179,111 @@ func setFirstUser(settings *PantherConfig) error {
 	return nil
 }
 
+// Update a single Lambda function for rapid developer iteration.
+//
+// This will only update the Lambda source, not the function configuration.
+func deploySingleLambda(function string) error {
+	switch function {
```
austinbyers, line 186: I'm exploring the possibility of parsing all of the CFN templates to automatically find the function source path and language, which would be a bit harder to implement but would mean newly added functions would be discovered automatically without having to be added to this switch statement.

This was the fastest implementation, if I can quickly switch to the CFN parsing I will, otherwise I'll leave that as a future TODO

austinbyers: It turned out to be pretty easy to read from the templates instead, which simplified the code a bit

alxarch (approved):
LGTM
This will definitely help next time I get stuck debugging some lambda.

PR #1800, opened 2020-10-20
Onboarding Logs validation fix

alexmylonas:
## Background

This PR includes fixes for some bad validation checks for wizards used when onboarding logs.

## Changes

- Correct validation for `integrationLabel`
- Removed misleading button when creating/updating SQS

## Testing

- Locally

### DIFF
--- a/web/src/components/wizards/SqsSourceWizard/SqsSourceConfigurationPanel/SqsSourceConfigurationPanel.tsx
+++ b/web/src/components/wizards/SqsSourceWizard/SqsSourceConfigurationPanel/SqsSourceConfigurationPanel.tsx
@@ -26,6 +26,7 @@ import { WizardPanel } from 'Components/Wizard';
 import { pantherConfig } from 'Source/config';
 import logo from 'Assets/sqs-minimal-logo.svg';
 import { useListAvailableLogTypes } from 'Source/graphql/queries/listAvailableLogTypes.generated';
+import { yupIntegrationLabelValidation } from 'Helpers/utils';
 import { SqsLogSourceWizardValues } from '../SqsSourceWizard';
 
 const emptyArray = [];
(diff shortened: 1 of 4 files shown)


web/src/components/wizards/SqsSourceWizard/ValidationPanel/ValidationPanel.tsx
```diff
@@ -148,11 +148,6 @@ const ValidationPanel: React.FC = () => {
           alt="Validating source health..."
           src={WaitingStatus}
         />
-        <WizardPanel.Actions>
-          <Button variantColor="darkgray" onClick={goToPrevStep}>
-            Cancel
-          </Button>
-        </WizardPanel.Actions>
```
alexmylonas, line 155: This button is misleading users. When a user clicks it, it doesn't actually reverse the creation or update process triggered from the last step. 

There is no way to do this atm.

web/src/components/wizards/SqsSourceWizard/SqsSourceConfigurationPanel/SqsSourceConfigurationPanel.tsx
```diff
@@ -105,7 +106,12 @@ const SqsSourceConfigurationPanel: React.FC = () => {
       </Box>
       <WizardPanel.Actions>
         <WizardPanel.ActionNext
-          disabled={(!values.logTypes.length && !values.integrationLabel) || !isValid || !dirty}
+          disabled={
+            (!values.logTypes.length &&
+              !yupIntegrationLabelValidation.isValidSync(values.integrationLabel)) ||
```
alexmylonas, line 111: `yupIntegrationLabelValidation` is necessary to check `integrationLabel` is valid

web/src/components/wizards/SqsSourceWizard/ValidationPanel/ValidationPanel.tsx
3nvi: cool

web/src/components/wizards/SqsSourceWizard/SqsSourceConfigurationPanel/SqsSourceConfigurationPanel.tsx
3nvi: `yupIntegrationLabelValidation.isValidSync` I personally don't like this since it freezes the main thread to perform a validation that should already have been calculated by Yup+Formik.

What's the problem here? What problems would we be facing if we completely  removed the following line?

```tsx
(!values.logTypes.length && !yupIntegrationLabelValidation.isValidSync(values.integrationLabel))
```

vorillaz (approved):

3nvi (approved):

PR #1802, opened 2020-10-20
Removed error from metrics API

### LINKED ISSUE #1648: Error Reported for Empty Metrics
### Describe the bug

When querying a metric from a newly deployed panther instance (that has not yet created the specified metric), an error is returned rather than an empty array/object.

### Steps to reproduce

Query the metric api for a metric that has not yet been created in the panther instance.  For example, a freshly deployed panther instance might not have the `alertsCreated` metrics yet, and this api call will return an error:

```
aws lambda invoke --function-name panther-metrics-api --payload '{"GetMetrics":{"fromDate":"2020-08-17T17:50:00Z","toDate":"2020-08-29T17:50:00Z","metricNames":["alertsByRuleID"], "intervalMinutes":14}}' response.json
```

```
% cat response.json | python -mjson.tool
{
    "errorMessage": "Could not find data points for the given metric in the selected time period",
    "errorType": "DoesNotExistError"
}
```

### Expected behavior

An empty array/object returned rather than an error

### Environment

- Panther version: v1.9.0

### Screenshots

image

kostaspap:
## Background

Closes #1648 

Sometimes Panther will have no data points to report for a specific metric. This is an expected state for the system (I believe) and can happen if someone hasn't enabled some part of the Panther application (e.g. if they haven't onboarded any logs) 

Stop throwing an error if the metrics api doesn't have datapoints for a given metric for a given period. 

## Changes

- Remove error check

## Testing

- mage test:ci
- Pending E2E test

### DIFF
--- a/api/lambda/metrics/models/api.go
+++ b/api/lambda/metrics/models/api.go
@@ -52,8 +52,8 @@ type GetMetricsOutput struct {
 
 // MetricResult is either a single data point or a series of timestamped data points
 type MetricResult = struct {
-	SingleValue []SingleMetric   `json:"singleValue,omitempty"`
-	SeriesData  TimeSeriesMetric `json:"seriesData,omitempty"`
+	SingleValue []SingleMetric   `json:"singleValue"`
+	SeriesData  TimeSeriesMetric `json:"seriesData"`
 }
 
 type SingleMetric struct {
--- a/internal/core/metrics_api/api/get_metrics.go
+++ b/internal/core/metrics_api/api/get_metrics.go
@@ -230,7 +230,7 @@ func getMetricData(input *models.GetMetricsInput, queries []*cloudwatch.MetricDa
 	}
 
 	if len(responses) == 0 {
-		zap.L().Warn("no metrics returned for query", zap.Any("queries", queries))
+		zap.L().Info("no metrics returned for query", zap.Any("queries", queries))
 	}
 
 	return responses, nil
--- a/internal/core/metrics_api/api/get_metrics.go
+++ b/internal/core/metrics_api/api/get_metrics.go
@@ -39,7 +39,7 @@ const (
 
 var (
 	metricsInternalError = &genericapi.InternalError{Message: "Failed to generate requested metrics. Please try again later"}
-	metricResolvers = map[string]func(input *models.GetMetricsInput, output *models.GetMetricsOutput) error{
+	metricResolvers      = map[string]func(input *models.GetMetricsInput, output *models.GetMetricsOutput) error{
 		"alertsByRuleID":   getAlertsByRuleID,
 		"alertsBySeverity": getAlertsBySeverity,
 		"eventsLatency":    getEventsLatency,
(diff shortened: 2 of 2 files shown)


rleighton:
Will this effect the UI in any way?

kostaspap:
> Will this effect the UI in any way?

Will do an E2E and see how it works, up (Updated testing section). Wanted to verify with @nhakmiller if it's ok to remove it or if it the error was there for a specific reason

lindsey-w:
I defer to @nhakmiller, but I think this #1648 is related / we have talked about fixing this in the past.

internal/core/metrics_api/api/get_metrics.go
```diff
@@ -231,10 +229,5 @@ func getMetricData(input *models.GetMetricsInput, queries []*cloudwatch.MetricDa
 		}
 	}
 
-	if len(responses) == 0 {
-		zap.L().Warn("no metrics returned for query", zap.Any("queries", queries))
-		return nil, metricsNoDataError
-	}
```
nhakmiller, line 237: I'd leave the warning message in and just not return an error. Today, this activity is only driven by user behavior. If something else in the system is screwed up causing this to constantly be empty, having the warning messages helps debug. The original intent was to help troubleshoot issues where metric generation/retrieval was screwed up, by adding a datapoint to show we did indeed attempt to retrieve the metrics but there was nothing to retrieve. A couple WARN messages from a user hitting the page too early won't notify the on-call staff, so it should be fine.

Not a blocker, just a suggestion. If not a WARN, then at the very least an INFO message.

nhakmiller (approved):

nhakmiller:
> I defer to @nhakmiller, but I think this #1648 is related / we have talked about fixing this in the past.

Yep, this closes #1648.

nhakmiller:
> > Will this effect the UI in any way?
> 
> Will do an E2E and see how it works, up (Updated testing section). Wanted to verify with @nhakmiller if it's ok to remove it or if it the error was there for a specific reason

I suspect the frontend will just show some weird empty graphs, but good to double check.

kostaspap:
@rleighton Verified that it works as expected:

Screen Shot 2020-10-22 at 12 39 10 PM

api/lambda/metrics/models/api.go
```diff
@@ -52,8 +52,8 @@ type GetMetricsOutput struct {
 
 // MetricResult is either a single data point or a series of timestamped data points
 type MetricResult = struct {
-	SingleValue []SingleMetric   `json:"singleValue,omitempty"`
-	SeriesData  TimeSeriesMetric `json:"seriesData,omitempty"`
+	SingleValue []SingleMetric   `json:"singleValue"`
+	SeriesData  TimeSeriesMetric `json:"seriesData"`
```
kostaspap, line 56: Had to remove the `omitempty` so that it would return something like:
```
"alertsByRuleID": {
    "singleValue": [],
    "seriesData": {
      "timestamps": [],
      "series": []
    }
  },
```
With the `omitempty`, the `singleValue` field would be null and FE would complain.

internal/core/metrics_api/api/get_metrics.go
kostaspap: Makes sense. I added log message. I am worried that we can get multiple Warn for every metric that has no value - they could easy add up and get lots of false alarms, so I switched to INFO.

PR #1807, opened 2020-10-20
Fixing threshold for Log Processor input queue

kostaspap:
## Background

Adjusting threshold for Log processor input SQS queue


austinbyers (approved):

rleighton (approved):

internal/core/custom_resources/resources/alarms_sqs.go
```diff
@@ -90,7 +90,7 @@ func putSQSAlarmGroup(props SQSAlarmProperties) error {
 			fmt.Sprintf("Panther-%s-%s", sqsAgeAlarm, props.QueueName))
 		input.MetricName = aws.String("ApproximateAgeOfOldestMessage")
 		input.Statistic = aws.String(cloudwatch.StatisticMaximum)
-		var threshold float64 = 900.0
+		threshold := 900.0
```
alxarch, line 93: I would prefer to change this to `(15*time.Minute).Seconds()` for clarity

alxarch: Why does this const differ from the one defined in the YAML file?

kostaspap: This is the default value. The value in the CFN template is override for a specific queue.

kostaspap: Updated.

PR #1820, opened 2020-10-21
Slight increase in source-api retries

kostaspap:
## Background

Small increase in default number of retries for source-api. Now doing 5 retries (compared to default of 3), so we are retrying for ~5 seconds. 
Source-api is queried by BE and FE services, so need to keep the number of retries to something that would make sense for both. 

## Testing

- test:ci

### DIFF
--- a/internal/core/source_api/api/list_integrations.go
+++ b/internal/core/source_api/api/list_integrations.go
@@ -19,6 +19,8 @@ package api
  */
 
 import (
+	"go.uber.org/zap"
+
 	"github.com/panther-labs/panther/api/lambda/source/models"
 	"github.com/panther-labs/panther/pkg/genericapi"
 )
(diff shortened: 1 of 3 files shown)


internal/core/source_api/api/list_integrations.go
```diff
@@ -31,6 +33,7 @@ func (API) ListIntegrations(
 
 	integrationItems, err := dynamoClient.ScanIntegrations(input.IntegrationType)
 	if err != nil {
+		zap.L().Error("failed to list integrations", zap.Error(err))
```
kostaspap, line 36: Logging the error - useful for debugging purposes.

internal/core/source_api/ddb/ddb.go
```diff
@@ -35,9 +36,9 @@ type DDB struct {
 }
 
 // New instantiates a new client.
-func New(tableName string) *DDB {
+func New(awsSession *session.Session, tableName string) *DDB {
 	return &DDB{
-		Client:    dynamodb.New(session.Must(session.NewSession())),
+		Client:    dynamodb.New(awsSession, aws.NewConfig().WithMaxRetries(5)),
```
kostaspap, line 41: No need to create a new session - code can use the session created for all AWS clients.

rleighton (approved):

PR #1822, opened 2020-10-21
Fixing issue when retrieving events for an alert

kostaspap:
## Background

While testing 1.11 found this bug. It demonstrates for rules that use threshold. 

## Changes

- Fixing S3 list object operation

## Testing

- pending E2E

### DIFF
--- a/internal/log_analysis/alerts_api/api/get_alert.go
+++ b/internal/log_analysis/alerts_api/api/get_alert.go
@@ -157,7 +157,7 @@ func getEventsForLogType(
 		} else { // not starting from a pagination token
 			// objects have a creation time as prefix we can use to speed listing,
 			// for example: '20200914T021539Z-0e54cab2-80a6-4c27-b622-55ad4d355175.json.gz'
-			listRequest.StartAfter = aws.String(partitionPrefix + alert.CreationTime.Format("20060102T150405Z"))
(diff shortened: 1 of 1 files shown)


internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -157,7 +157,7 @@ func getEventsForLogType(
 		} else { // not starting from a pagination token
 			// objects have a creation time as prefix we can use to speed listing,
 			// for example: '20200914T021539Z-0e54cab2-80a6-4c27-b622-55ad4d355175.json.gz'
-			listRequest.StartAfter = aws.String(partitionPrefix + alert.CreationTime.Format("20060102T150405Z"))
+			listRequest.StartAfter = aws.String(partitionPrefix + nextTime.Format("20060102T150405Z"))
```
kostaspap, line 160: In short: When a rule has a threshold configured, the rules engine will store in S3 all the events that match that rule. 
However, the alert will be created only when the rules engine has stored num(events)> threshold. 

So as `alert.CreationTime` we will have the time that the alert was created. We may have stored events in S3 prior to that. We are already taking care of that in our code (check line 167 of this file) but somehow this optimization missed it.

giorgosp (approved):

rleighton (approved):

PR #1835, opened 2020-10-21
fix rule threshold in summary

### LINKED ISSUE #1826: Frontend displays all Rule Thresholds as 0
### Describe the bug

Frontend displays all Rule Thresholds as 0 in the Rules page. 

### Steps to reproduce

1. Go to Log Analysis -> Rules
2. See that all rules have threshold 0

### Expected behavior

FE should display the appropriate threshold for each rule

### Environment

- Panther version or commit: https://github.com/panther-labs/panther/commit/5ad55d09009d525132109b647abc7c344a46d9e9

s0l0ist:
## Background

Closes #1826 

## Changes

- Add `threshold` to `PolicySummary` swagger. Needed due to the way the `ListRules` API is implemented which uses the `PolicySummary` model during paging which is then converted to `RuleSummary`.
- Add `threshold` to `RuleSummary` swagger
- Added to integration tests
- Frontend changes to restrict the user from inputting a value less than `1` for the threshold (screenshot is from manual override, the field will not let you go below 1 normally)

## Testing

- mage test:ci
- mage test:integration

image
image


austinbyers (approved):

austinbyers:
Perhaps a screenshot showing the correct threshold showing up on the list views?

PR #1839, opened 2020-10-21
Backend bulk update alert statuses

### LINKED ISSUE #1838: Backend bulk update alert statuses
### Description

This issue is to address the backend changes to facilitate updating alert statuses in bulk

The update status API now accepts a list of alert IDs to update. It is capable of handling many alerts (in excess of 1k) by using parallel chunked update requests each processing 10 alerts in series.

### Related Services

Alerts API
- Update Status

### Designs

see epic

### Acceptance Criteria

see epic

s0l0ist:
## Background

Closes #1838 

Also adds some small frontend changes so that this can be released and not break the existing frontend UI.

## Changes

- Update Status API accepts a list of alert IDs to update
- Takes a list of alerts, splits them into chunks of **10**, and dispatches these chunks in parallel to update
- Each chunk or page is a separate goroutine that executes the `UpdateItem` DDB query in series. This will roughly bound the worst-case time to be the sequential operation of updating 10 alert items. This could be optimized to use a worker pool with sub-linear expansion to curve the number of goroutines and thread contention; however, this is overkill.

## Testing

- mage test:ci


internal/log_analysis/alerts_api/table/update_alert.go
```diff
@@ -30,41 +30,50 @@ import (
...
-
-	// Create the condition builder
-	conditionBuilder := createConditionBuilder(input)
-
-	// Build an expression from our builders
-	expression, err := buildExpression(updateBuilder, conditionBuilder)
-	if err != nil {
-		return nil, err
+// UpdateAlertStatus - updates all the specified alerts and returns the updated list
+func (table *AlertsTable) UpdateAlertStatus(input *models.UpdateAlertStatusInput) ([]*AlertItem, error) {
```
kostaspap, line 34: Transactions seem to make it more complicated than it should? Why not just Updating the items? It saves you from doing one extra read and I think will simplify the code as well (just a for loop updating each item would do the job).

s0l0ist: This is a good callout. I suppose we could build out many individual update requests, dispatch each one in parallel and return; however, I wanted this to go a bit faster and try to 'lighten' the load on DDB.

The only way to perform native batch updates is through write transactions. I don't know how this is billed, but I expect it to be cheaper as it is [(N/25) * 2] DDB calls on average instead of [N] where N = # of alerts.

You're spot on with not really needing transactions because it is acceptable for these to fail. The primary reason was purely to make this go a bit faster and be more friendly to DDB while being cheaper.

kostaspap: I see what you mean. Just some more thoughts on this though: 
1. Transactions are more heavy for DDB than individual updates. Transactions require DDB to store the results of each individual operation in a temporary location and revert or commit the result upon success. Since they require DDB to maintain a state which is always going to be more complex and with higher chances of failure. They are definitely more heavy for DDB than normal updates. 
2. Another benefit of individual updates is that it saves you the extra `Get`. You could set the `ReturnValues ` property to `ALL_NEW`; the UpdateOutput will contain the new values. https://docs.aws.amazon.com/sdk-for-go/api/service/dynamodb/#UpdateItemInput
3. It's indeed problem that DDB doesn't offer a nice API to do Updates of individual items in batch. What I am thinking though is that the RTT for a DDB update is around ~10ms, which means it would take 1 second for the max items (100) which I guess is acceptable (I think the users can't even select that many items currently from the UI). 

I think another big benefit is that the code would be much simpler as well?

internal/log_analysis/alerts_api/table/update_alert.go
```diff
@@ -30,41 +30,50 @@ import (
...
+			return nil, err
+		}
+
+		transactWriteItem := &dynamodb.TransactWriteItem{
+			Update: &dynamodb.Update{
+				ConditionExpression:                 expression.Condition(),
+				ExpressionAttributeNames:            expression.Names(),
+				ExpressionAttributeValues:           expression.Values(),
+				Key:                                 alertKey,
+				ReturnValuesOnConditionCheckFailure: aws.String("ALL_OLD"),
```
kostaspap, line 58: `NONE` for this one? Since we are doing a get afterwards.

kostaspap (approved):
I don't see any bug, I have however concerns about the decision to use DDB transactions.

internal/log_analysis/alerts_api/table/update_alert.go
s0l0ist: Yeah this could be `NONE`.

internal/log_analysis/alerts_api/table/update_alert.go
s0l0ist: > Transactions are more heavy for DDB than individual updates. Transactions require DDB to store the results of each individual operation in a temporary location and revert or commit the result upon success. Since they require DDB to maintain a state which is always going to be more complex and with higher chances of failure. They are definitely more heavy for DDB than normal updates.

Right, it's a bit heavier so we add some constant time to the request. The chances of failure I think are acceptable as it falls to DDB internally (I hope).

> Another benefit of individual updates is that it saves you the extra Get. You could set the ReturnValues property to ALL_NEW; the UpdateOutput will contain the new values. https://docs.aws.amazon.com/sdk-for-go/api/service/dynamodb/#UpdateItemInput

Yep, I wish the batch transaction update could handle this...

> It's indeed problem that DDB doesn't offer a nice API to do Updates of individual items in batch. What I am thinking though is that the RTT for a DDB update is around ~10ms, which means it would take 1 second for the max items (100) which I guess is acceptable (I think the users can't even select that many items currently from the UI). We don't even have to launch all the queries in parallel, a for loop would be enough. I don't think we care so much about latency now (until we do :P )

Users can actually select as many alerts as they scroll due to our infinite scroll. For UX, we should never be more than 400 ms. We had filtering at one point that was extremely slow (1-2 seconds) and it was awful before some backend optimizations. By default, we load 50 alerts on modern resolution screens so for this most probable scenario, the bulk update will be near constant-time for any amount of alerts bulk updating.

If I were to do this again and scope out the work, we could have just done the sequential for-loop - but I've _already_ done the hard optimization part so why not use it, right?

internal/log_analysis/alerts_api/api/update_alert_status.go
```diff
@@ -19,27 +19,89 @@ package api
...
+// We are avoiding serialized update requests to ddb and instead will parallelize
+
+func dispatchUpdates(input *models.UpdateAlertStatusInput, maxPageSize uint64) ([]*table.AlertItem, []error) {
+	updateChannel := make(chan updateResult)
+
+	// alert count
+	alertCount := uint64(len(input.AlertIDs))
+
+	// Get the total number of pages. This will be the number of goroutines to create
+	pages := uint64(math.Ceil(float64(alertCount) / float64(maxPageSize)))
```
austinbyers, line 68: Haha why the uint64 everywhere? The default `int` type should be fine

internal/log_analysis/alerts_api/api/update_alert_status.go
```diff
@@ -19,27 +19,89 @@ package api
...
 
-	gatewayapi.ReplaceMapSliceNils(result)
-	return result, nil
+	// Slice up the AlertIDs into chunks to be processed in parallel
+	for page := uint64(0); page < pages; page++ {
+		pageSize := uint64(math.Min(float64((page*maxPageSize)+maxPageSize), float64(alertCount)))
```
austinbyers, line 72: This variable name is a lie :) You're using `pageSize` to mean the end of a slice: it's an index, not the number of items in the page

Simplify to something like:

```go
endIndex := (page+1)*maxPageSize
if endIndex > len(input.AlertIDs) {
    endIndex = alertCount
}

// create shallow copy ...
inputItems := &models.UpdateAlertStatusInput{
    // ...
    AlertIDs: input.AlertIDs[page*maxPageSize : endIndex]
}
```

internal/log_analysis/alerts_api/api/update_alert_status.go
```diff
@@ -19,27 +19,89 @@ package api
...
+
+		// create shallow copy of the input with chunked AlertIDs
```
austinbyers, line 74: This isn't really necessary, you could just send the entire slice and the start/end indexes:

```go
go dispatchUpdate(input, updateChannel, startIndex, endIndex)
```

A slice is just a 3-tuple, and whether you re-slice it here or in the goroutine, the result is the same: a 3-tuple which points to the same underlying array.

internal/log_analysis/alerts_api/api/update_alert_status.go
```diff
@@ -19,27 +19,89 @@ package api
...
+	for _, result := range results {
+		gatewayapi.ReplaceMapSliceNils(result)
+	}
+	return results, nil
+}
+
+// dispatchUpdates - dispatches updates to all alerts in parallel
+// We are avoiding serialized update requests to ddb and instead will parallelize
+
+func dispatchUpdates(input *models.UpdateAlertStatusInput, maxPageSize uint64) ([]*table.AlertItem, []error) {
```
austinbyers, line 61: Rather than returning a slice of errors, I would argue this should follow the Go conventions and just return an `error` - you can flatten the error slice before returning with a simple string join

internal/log_analysis/alerts_api/table/update_alert.go
```diff
@@ -171,3 +180,47 @@ func (table *AlertsTable) update(
...
+	newItems interface{},
+) error {
+	// Perform the batch update transaction
+	_, err := table.Client.TransactWriteItems(transactWriteItems)
+	if err != nil {
+		return &genericapi.AWSError{Method: "dynamodb.TransactWriteItems", Err: err}
+	}
+
+	// Next, we perform a batch get transaction since the update transaction doesn't return any values.
+	// We start by constructing the get input by inspecting the write input
```
austinbyers, line 196: Is a batch get really necessary? If all you're updating is the alert status, and the update was successful, don't you know exactly what the result will be and you can just return that? 

I know there's a `lastUpdatedByTime`, but you can just pre-compute that before writing to Dynamo. Or are there other fields we have to lookup?

austinbyers (commented):
Right, since DynamoDB doesn't have a "batch update", your options are BatchWrite (overwrite the item) or TransactWrite

The BatchWrite is perfect if you know what the full item should be ahead of time, i.e. (alertStatus, lastUpdatedTime, lastUpdatedBy). But even if there are other fields in the same row, you could do a BatchGet followed by a BatchWrite, instead of the other way around:

```
For each alertID batch of X:
    BatchGet(alertIDs)
    BatchWrite(alertStatus, alertID, lastUpdatedBy, otherExistingFieldsStayTheSame...)
```

This shouldn't require many updates to your code: you can still do batching of 25 items and you still have the same operations (BatchGet + BatchWrite), just in the opposite order

This does, however, lead to a possible (but rare) race condition:

1. AlertStatus BatchGet
2. Some other system changes the alert
3. AlertStatus BatchWrite, which removes the changes that just landed

Whether you're ok with this possibility depends on exactly what's stored in the Dynamo item: if it's just alert status and timestamps, I'd say this is acceptable. If you have eventIDs that could be permanently lost, maybe this approach is too risky

----

Using transactions as a way to get parallel updates is definitely clever, I never would have thought of that! But they are twice as expensive and slower and we're not using them for their intended purpose, so I would still say we should avoid them if possible unless we actually need the transaction guarantee.

Another thing to consider with transactions: they will [block all other updates to those items](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis.html#transaction-conflict-handling) until the entire transaction is finished (that's literally the point of a transaction, after all). If the log processor is trying to update the table directly at the same time, we're going to start seeing `TransactionConflictException`

---

The best solution seems to be a standard `UpdateItem`, where you update only the fields you care about (AlertStatus, LastUpdated*). This way, alert status updates will never conflict with log processor updates to the eventIDs or any other field, we don't pay the cost ($$ and latency) for transactions, and you only have to do a single Update for each item (no Get required).

It sucks they don't have a batch operation for it, but you can kinda make one up yourself by having goroutines to update every item in parallel or (more properly) a fixed set of worker routines that handle the updates.

I know that a bulk update of alert statuses is only UI-driven, and so optimizing the architecture isn't incredibly important. We can change the implementation later as long as we're happy with the API design.

But it does become a bigger concern when the public API explicitly encourages automated bulk changes - you could imagine writing a script to close all alerts with a certain tag, for example

internal/log_analysis/alerts_api/api/api.go
```diff
@@ -36,10 +36,11 @@ import (
...
 var (
-	env        envConfig
-	awsSession *session.Session
-	alertsDB   table.API
-	s3Client   s3iface.S3API
+	env            envConfig
+	awsSession     *session.Session
+	alertsDB       table.API
+	s3Client       s3iface.S3API
+	maxDDBPageSize = uint64(25) // This is the max number of items for DDB transactions
```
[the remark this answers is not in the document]
s0l0ist: Yep, will make this a `const`. Needed the type for easier comparisons and calculations for pages.

internal/log_analysis/alerts_api/api/update_alert_status.go
s0l0ist: Over-optimization :)

internal/log_analysis/alerts_api/table/update_alert.go
s0l0ist: We don't have the full alert summary at any point in the request so it is necessary to 'get' it somehow before returning to the frontend.

internal/log_analysis/alerts_api/table/update_alert.go
s0l0ist: okay, because TXNs _could_ cause the log processor to fail to update the alert item in DDB due to locking, I'm going to throw TXNs out.

Instead, I will implement 'paged' goroutines which will use the `UpdateItem` command in series to amortize the worst-case constant time for the frontend. This should be quite fast and we can tune if necessary in the future to have less sequential updates in favor of more parallel goroutines.

Thank you @kostaspap and @austinbyers for your feedback!

internal/log_analysis/alerts_api/api/update_alert_status.go
s0l0ist: I'll rename, but I like my one-liner that doesn't have conditionals for calculating dangerous index offsets.

austinbyers: `math.Min` is effectively the same as a conditional, but IMO your one-liner is pretty hard to read.

At the very least, replace

```
float64((page*maxMageSize) + maxPageSize)
```

with

```
float64((page+1)*maxPageSize)
```

internal/log_analysis/alerts_api/api/update_alert_status_test.go
```diff
@@ -34,42 +39,70 @@ func TestUpdateAlert(t *testing.T) {
...
+		})
 	}
-	expectedSummary := &models.AlertSummary{
-		AlertID:           aws.String("alertId"),
-		Status:            "CLOSED",
-		LastUpdatedBy:     "userId",
-		LastUpdatedByTime: timeNow,
-		DeliveryResponses: []*models.DeliveryResponse{},
+
+	pages := int(math.Ceil(float64(alertCount) / float64(maxDDBPageSize)))
```
kostaspap, line 88: This is the same calculation used in the actual code and could be wrong. How many pages should the code generate for 12345 alerts? You can just put that number in here :)

internal/log_analysis/alerts_api/table/table.go
```diff
@@ -48,7 +48,7 @@ const (
 type API interface {
 	GetAlert(*string) (*AlertItem, error)
 	ListAll(*models.ListAlertsInput) ([]*AlertItem, *string, error)
-	UpdateAlertStatus(*models.UpdateAlertStatusInput) (*AlertItem, error)
+	UpdateAlertStatus(*models.UpdateAlertStatusInput) ([]*AlertItem, error)
```
kostaspap, line 51: nit: It seems that there is no clear separate of layers here. Table interface should operate on its own abstractions (e.g. `AlertItem`). Here have one method operating at two different abstraction layers (`*models.UpdateAlertStatusInput` and `[]*AlertItem`) which is always a warning sign. 

I don't say that we should make this refactoring as part of this PR, but we have to be aware of the tech debt we are adding here and try to avoid it as much as possible.

api/lambda/alerts/models/api.go
```diff
@@ -96,18 +96,18 @@ type ListAlertsInput struct {
...
+//         "alertIds": ["84c3e4b27c702a1c31e6eb412fc377f6"],
 //         "status": "CLOSED"
 //         // userId is added by AppSync resolver (UpdateAlertStatusResolver)
 //         "userId": "5f54cf4a-ec56-44c2-83bc-8b742600f307"
 //     }
 // }
 type UpdateAlertStatusInput struct {
 	// ID of the alert to update
-	AlertID *string `json:"alertId" validate:"hexadecimal,len=32"` // AlertID is an MD5 hash
+	AlertIDs []*string `json:"alertIds" validate:"gt=0,dive,hexadecimal,len=32"` // AlertID is an MD5 hash
```
kostaspap, line 110: `[]*string` -> `[]string`? Not sure we need the `*` here. It would avoid quite few pointer dereferences that we do in the code.

kostaspap (approved):
Looks good! Just some minor comments.

internal/log_analysis/alerts_api/table/table.go
s0l0ist: I see what you mean; however, I didn't want to create/write dynamo-specific code in the API layer. Right now, the API handles the chunking of updates and manages concurrency.

The table interface accepts a list to perform all the logic to build the dynamo request, dispatch, receive, and return data.

At this point, we don't need to have an interface that just accepts a single `AlertItem`. All usages of the top-level API have already been converted to use lists throughout the application.

PR #1840, opened 2020-10-21
VPC Scanning bug fix

nhakmiller:
## Background

In a recent change (#1784) I changed how VPC scanning works to include only a list of security group and network acl IDs, and not a full list of embedded security groups and network acls. Policies that needed to consider the contents of a security group while evaluating a VPC would need to use the `resource_lookup` helper.

As a convenience, I pulled up the default security group ID and the default network ACL ID as these elements are frequently more interesting than other security groups/network acls, and having to do 20 resource lookups just to find the default seemed unwieldly. Unfortunately, I mistyped the check for default security groups so this feature is not working as intended. This PR fixes that bug.

## Changes

- Removed an extraneous trailing space

## Testing

- Deployed into my dev account and checked specifically this field in a scanned VPC.
- Added a check to unit testing for this field
Screen Shot 2020-10-21 at 4 52 43 PM


lindsey-w (approved):
Looks good to me!

nhakmiller:
Adding the `sync-to-release`  label because it's a small fix to a change included in v1.11.

PR #1853, opened 2020-10-23
Custom payload

### LINKED ISSUE #1753: Users should be able to send arbitrary payload to an alert destination
### Description

Users should be able to send arbitrary payload to an alert destination. 

### Related Services

- panther-rules-engine
- panther-log-alert-forwarder
- panther-alert-delivery

### Acceptance Criteria

- Users can send arbitrary payload when sending an alert

kostaspap:
## Background

Closes #1753 

Adding the ability to send custom payload with every alert. 
Notable facts:
1. We introduce a new function `alert_context` in the rules. 
Screen Shot 2020-10-23 at 4 53 07 PM
2. The function is expected to return a `Dict[str, Any]`
3. The value returned by the `alert_context` will be included in the payload sent to SNS/SQS/Webhook. We only want to send to those destinations since the payload can be significant (up to 200kb) and it's meant for automation purposes, we don't want to overload other external systems.

 Name of the property is `alertContext` and is a nested JSON type. See SQS sample:
```
{
  "id": "TestingAlertContext",
  "createdAt": "2020-10-23T13:25:41Z",
  "severity": "HIGH",
  "type": "RULE",
  "link": "https://web-xxxxxxxxx.eu-west-1.elb.amazonaws.com/log-analysis/alerts/58a3c6749a589287d7e9f0d1bd0efc6d",
  "title": "New Alert: TestingAlertContext",
  "name": null,
  "alertId": "58a3c6749a589287d7e9f0d1bd0efc6d",
  "alertContext": {
    "nested": {
      "nestedkey": "nestedValue"
    },
    "string": "teststring",
    "integer": 10
  },
  "description": "",
  "runbook": "",
  "tags": [],
  "version": "Z3qToGrdPi7USQpdmz.WCEt33chA0Zrl"
}
```
4. The value is stored in all the tables in `panther_rule_matches` database. The column name is `p_alert_context` and is of type `string`
Screen Shot 2020-10-23 at 4 47 11 PM
5. The value is also visible from the `Events` tab in the Alert details view: 
Screen Shot 2020-10-23 at 4 53 07 PM

6. In case the function is not specified or it throws an exception, we value a `nil` value in S3. SNS/SQS/Webhook notification will have an empty JSON: 
```
"alertContext": {}
```
7. Max size is 200kb. If this size is exceeded, we don't return anything. (maybe we should return `{"_error": "Threshold exceeded"}`? Same in case of an error, should we return `{"_error": "error description"}`?

## Changes

- Rules Engine now runs the `alert_context` method if a rule matched an event
- Rules Engine stores the context in DDB in the `panther-log-alert-dedup` and also writes the extra field in S3 when it stores the matched events
- `panther-log-alert-forwarder` picks up the data from DDB Stream. It forwards it to the `panther-alert-delivery` (it sends an SQS message.
- `panther-alert-delivery` forwards the message to SQS/SNS/Webhhook

## Testing

- Unit tests
- Tested E2E with SNS & SQS destinatoins


kostaspap:
@jacknagz  no need to see the code - but can you check the description of the PR to see if this is sufficient?

internal/core/alert_delivery/api/send.go
```diff
@@ -56,7 +56,7 @@ func sendAlerts(alertOutputs AlertOutputMap) []DispatchStatus {
 	}
 
 	// Wait until all outputs have finished, gathering all the statuses of each delivery
-	deliveryStatuses := []DispatchStatus{}
+	var deliveryStatuses []DispatchStatus
```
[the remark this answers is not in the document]
alxarch: I think that we can be using both without having to argue each time :)
Literate syntax is better when you want to initialize the slice with some values. 
I usually define all my variables using `:=` just to avoid the different indentations when reading code. I do not think that `foo := []string{}` does any allocations. It's just a zero slice header on the stack. And for the cases where we want to differentiate between empty slice and `nil` slice on the return value we are better served returning an explicit `nil`. Lastly `omitempty` ignores the empty slice on JSON marshal where it's the only place I know these two behave differently.

alxarch (commented):
I see that we chose to use an additional function instead of having the user modify the event in place. While this is the more sanitized choice, it means that the user cannot attach information from external sources used during rule match. Are we planning to address this in a different way? I think it would be useful to have somewhere to store state between calls, both at file level and per event level.

jacknagz:
@kostaspap looks good! I would make it available to every destination though. There's plenty of non-automation use cases for this feature. Thanks!

kostaspap:
> I see that we chose to use an additional function instead of having the user modify the event in place. While this is the more sanitized choice, it means that the user cannot attach information from external sources used during rule match. Are we planning to address this in a different way? I think it would be useful to have somewhere to store state between calls, both at file level and per event level.

Yup, this was discussed in the relevant RFC. 
We opted for introducing this change now and adding the capability for rule methods to pass context around at a later stage. The ability to pass context around needs to be thought more carefully and would affect the way `dedup()`, `title()` functions work as well.

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -206,6 +207,17 @@ func (h *Handler) sendAlertNotification(rule *ruleModel.Rule, alertDedup *AlertD
...
+		var context map[string]interface{}
+		err := jsoniter.UnmarshalFromString(alertDedup.AlertContext, &context)
+		if err != nil {
+			// best effort
+			zap.L().Warn("failed to unmarshal alert context", zap.Error(err))
+		} else {
+			alertNotification.Context = context
+		}
+	}
+
```
rleighton, line 220: Suggestion: if the `msgBody` below is > 200kb ,  re-serialize the `alertNotification` with the `AlertContext` set to:
```
alertContext": {
"messageTooBig": "the alert context was too large to send by 52 bytes",
}
```

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -30,6 +31,7 @@
 
 # Maximum size for a title
 MAX_TITLE_SIZE = 1000
+MAX_ALERT_CONTEXT_SIZE = 200 * 1000  # 200kb
```
rleighton, line 34: Or is it 200*1024?

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -187,6 +197,29 @@ def _get_title(self, event: Dict[str, Any], use_default_on_exception: bool = Tru
...
+            alert_context = self._run_command(self._module.alert_context, event, dict)
+            serialized_alert_context = json.dumps(alert_context)
+        except Exception as err:  # pylint: disable=broad-except
+            if use_default_on_exception:
+                self.logger.warning('alert_context method raised exception. Using default. Exception: %s', err)
+                return None
+            raise
+
+        if len(serialized_alert_context) > MAX_ALERT_CONTEXT_SIZE:
+            # If title exceeds max size, use default
```
rleighton, line 214: Comment seems wrong

internal/log_analysis/rules_engine/src/rule.py
kostaspap: done

internal/log_analysis/rules_engine/src/rule.py
kostaspap: Updated

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
kostaspap: What about :
```
"alertContext": {
"_error": "Alert context was too large to send"
}
```
?

nhakmiller: Since we're already calculating the length, it seems handy to pass along for debugging purposes. e.g.

```
"alertContext": {
"_error": "Alert context was too large to send. Context is XYZ bytes, but max size is 200Kb."
}
```

web/src/pages/CreateRule/CreateRule.tsx
```diff
@@ -27,14 +27,15 @@ import {
   DEFAULT_DEDUP_FUNCTION,
   DEFAULT_RULE_FUNCTION,
   DEFAULT_TITLE_FUNCTION,
+  DEFAULT_ALERT_CONTEXT_FUNCTION,
```
nhakmiller, line 30: At some point we should re-visit how we do this, maybe a button click that injects these comments in or something e.g. "Add title function". As we add more dynamic fields, this will just become more & more cluttered. Not in scope for this PR though.

internal/log_analysis/rules_engine/src/rule.py
nhakmiller: Small thing, but seems like this change didn't get pushed up?

internal/log_analysis/awsglue/schema.go
```diff
@@ -129,6 +129,11 @@ var (
 			Type:    GlueStringType,
 			Comment: "Alert id",
 		},
+		{
+			Name:    "p_alert_context",
+			Type:    GlueStringType,
+			Comment: "Alert id",
```
nhakmiller, line 135: This comment should probably be updated.

nhakmiller (approved):
A few nitpicks, but over all looks good. Straight forward implementation for a straightforward feature :)

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
kostaspap: Yup yup. Pushed this change

internal/log_analysis/awsglue/schema.go
kostaspap: Updated

internal/log_analysis/rules_engine/src/rule.py
kostaspap: I should have written "Will update" - it took longer to push to remote branch. 
Now it is updated :)

kostaspap:
> @kostaspap looks good! I would make it available to every destination though. There's plenty of non-automation use cases for this feature. Thanks!

Done

alxarch:
> Yup, this was discussed in the relevant RFC.
> We opted for introducing this change now and adding the capability for rule methods to pass context around at a later stage. The ability to pass context around needs to be thought more carefully and would affect the way `dedup()`, `title()` functions work as well.

I still think that we could be passing context around by adding stuff to `event` but since I was too busy at the time to support this idea during the RFC discussion, I will not push it now.

internal/core/alert_delivery/outputs/github_test.go
```diff
@@ -65,6 +66,6 @@ func TestGithubAlert(t *testing.T) {
 
 	httpWrapper.On("post", expectedPostInput).Return((*AlertDeliveryResponse)(nil))
 
-	require.Nil(t, client.Github(alert, githubConfig))
+	assert.Nil(t, client.Github(alert, githubConfig))
```
alxarch, line 69: nit: no need to use both require and assert packages. 
On this subject: what I like to do to avoid the annoying `t` argument is
```
assert := require.New(t)
assert.Equal(foo, bar)
assert.NoError(err)
```
Saves some typing.

internal/core/alert_delivery/outputs/jira_test.go
```diff
@@ -24,7 +24,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/assert"
```
alxarch, line 27: Is there some specific reason you prefer the `assert` package? Or just the name?

alxarch (approved):
LGTM
I would prefer to not add a new function name to the rules but it's OK for now.

internal/core/alert_delivery/outputs/github_test.go
alxarch: https://pkg.go.dev/github.com/stretchr/testify/assert#hdr-Example_Usage

alxarch: require and assert are the exact same package. I do not know if importing both increases the build time for tests. Not that important though.

giorgosp (approved):
LGTM. My only thought is what if users already declare `alert_context` in one of their scripts? Maybe it is improbable now, but we need to consider this down the line as we add functions to the rule.

internal/core/alert_delivery/outputs/github_test.go
rleighton: require stops the testing and assert continues, so I often have:
```
require.NoError(t, err) // no need to test asserts if this fails
assert.Equal(t, expected, actual)
assert.Equal(t, expected2, actual2)
assert.Equal(t, expected3, actual3)
```

internal/core/alert_delivery/outputs/opsgenie.go
```diff
@@ -44,9 +45,13 @@ func (client *OutputClient) Opsgenie(
 	runBook := "\n <strong>Runbook:</strong> " + aws.StringValue(alert.Runbook)
 	severity := "\n <strong>Severity:</strong> " + alert.Severity
 
+	// Best effort attempt to marshal Alert Context
+	marshaledContext, _ := jsoniter.MarshalToString(alert.Context)
+	alertContext := "\n <strong>AlertContext:</strong> " + marshaledContext
```
rleighton, line 50: `html.EscapeString(alert.Context)` ? All of these that output HTML probably should be protected from characters like `>`.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -30,6 +31,9 @@
 
 # Maximum size for a title
 MAX_TITLE_SIZE = 1000
+MAX_ALERT_CONTEXT_SIZE = 200 * 1024  # 200kb
```
rleighton, line 34: Does this need to have the header/wrapper JSON in the sent alert subtracted so that the actual send is < 2k? Maybe:
```
ALERT_HEADER_SIZE = 100
MAX_ALERT_CONTEXT_SIZE = (200 * 1024) - ALERT_HEADER_SIZE   # total 200kb
```
It would be cool to test this in an integration test and see where it fails the SNS send.

rleighton (approved):

internal/core/alert_delivery/outputs/github_test.go
kostaspap: > nit: no need to use both require and assert packages.
> On this subject: what I like to do to avoid the annoying `t` argument is
> 
> ```
> assert := require.New(t)
> assert.Equal(foo, bar)
> assert.NoError(err)
> ```
> 
> Saves some typing.

What @rleighton  mentioned above. I noticed that one failing test would stop all the other tests in the package from running. After making the changes, I would run the tests and would see the failing `github_test.go`. Would fix it and then would see the failing `asana_test.go` etc
I figured maybe having an assert would be better here for that reason.

internal/core/alert_delivery/outputs/jira_test.go
kostaspap: See comment above.

internal/log_analysis/rules_engine/src/rule.py
kostaspap: The limit for DDB is 400kb per item (we store this one in DDB) and the limit for SQS/SNS is 256KB. 
I put a limit of 200kb as an approximation - the other fields included in the request will be less than the remaining 56kb

internal/core/alert_delivery/outputs/opsgenie.go
kostaspap: good catch

internal/log_analysis/rules_engine/src/rule.py
rleighton: Perfect. I don't have those limits memorized. How about putting exactly what you said in a comment above?

internal/core/alert_delivery/outputs/opsgenie.go
kostaspap: We should be doing that for all fields actually.

internal/log_analysis/rules_engine/src/rule.py
kostaspap: Added comment.

kostaspap:
> LGTM. My only thought is what if users already declare `alert_context` in one of their scripts? Maybe it is improbable now, but we need to consider this down the line as we add functions to the rule.

That's the same considerations we had when we introduced the `title` and `dedup` function. We probably need to do some overhauling at some point. Passing context between functions, making sure such updates will not overlap with users already defined methods, both would be great improvements.

internal/core/alert_delivery/outputs/opsgenie.go
kostaspap: Had a chat offline: Users might want to render the content as HTML; that's how we do it for `title` function for example. 
Will leave unescaped for now.

PR #1855, opened 2020-10-23
Override `.env` variables for Jest

vorillaz:
## Background

For our specs running, we are using the `dotenv` npm module to load some test environmental variables. Unfortunately, lots of us have some values hardcoded and it causes hiccups and flakiness to some test scenarios. This PR introduces `env` variables overrides **just** for our test suite.

## How to reproduce this one
Run `AWS_REGION="FOO" npm run test web/src/pages/GeneralSettings/GeneralSettings.test.tsx` and you may witness a failure from snapshot testing since `AWS_REGION` gets overridden by the call site.

## Testing
``

### DIFF
--- a/web/__tests__/env.ts
+++ b/web/__tests__/env.ts
@@ -17,9 +17,9 @@
  */
 
 import path from 'path';
-import { loadDotEnvVars, getAppTemplateParams } from '../scripts/utils';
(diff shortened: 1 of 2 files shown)


kostaspap (approved):
I didn't look at the code, but ran `mage test:ci` which now succeeds. Yay!

3nvi (approved):
LGTM 👍

PR #1857, opened 2020-10-23
Improve Log Processor Aggregation

rleighton:
## Background

**tldr;** This PR improves overall cost by 30% and file aggregation by 16 times.

The current implementation of the log processor does not aggregate files well. The consequence of this is that many small files can be emitted, resulting in poor Athena query performance and expensive Snowflake ingest.

We can view log processing with lambdas an optimization problem where we want to MINIMIZE lambda calls (in order to MAXIMIZE the file aggregations) constrained by keeping the SQS queue from growing.

The issue with the current implementation that triggers lambdas from the SQS queue directly is that this is so fast, that each lambda invocation will only get 1 or two files to process such that we get little opportunity to aggregate the data. 

The nature of this change is to switch to a polling model implemented by a CloudWatch event triggering the log processor lambda to read from SQS and the logic to scale up / down is moved inside the log processor. This new approach waits 1 minute and if that single lambda invocation can process all the events, then we only have 1 lambda execution per minute AND we have maximally aggregated the input! If this lambda does not finish all the events in the queue in 1 minute then the next minute's invocation will contribute to draining the SQS queue. In this way under small load spikes are absorbed and we never have more than 1 lambda invocation per minute.

What happens when the sustained load is higher than 1 lambda can handle? In this case the log processor will invoke more lambdas. The general idea (see code for specifics) is the log processor lambda looks at the number of events in the sqs queue every 30 seconds. If the count is larger than 1 lambda can handle (we set this number to 5000 files per invocation) the log processor will invoke another instance of itself. This provides an exponential response to load spikes while also keeping a reasonably smooth response (small oscillations) under steady load.

There is the same amount of "work" done in each approach but the key difference is that in this PR fewer lambdas run longer which gives better aggregation.  While the log processor lambda costs are about the same (because the work is about the same), the better aggregation translates in to lower costs and higher efficiencies down stream from the log processor.

The benchmark I have run lasted 2.5 days and sent a constant stream of 2,000 CloudTrail files per minute (roughly 5TB per day of uncompressed data) from an ec2 instance to:

 - master with log processor queue delay set to 0 (Ireland - eu-west-1)
 - code from this PR with Lambda size 2048 (Canada - ca-central-1)

The nature of the change provides a dramatic reduction in the number of  log processor lambda invocations (but longer executions). The better aggregation translates into cost savings because of:

 - fewer SQS messages
 - fewer SNS messages
 - fewer data catalog updater lambda invocations
 - fewer S3 object API calls
 - fewer  CloudWatch logs (due to fewer lambda calls)
 - fewer CloudTrail logs (due to fewer lambda calls)
 - smaller Snowpipe costs due to better aggregation
 - faster Athena queries over JSON data due to much small number of files

The cost reduction overall is looking to be about 30%.

image

 Below are the counts of the number of S3 object created in each region. Under this load, we see about a 16 times better aggregation using the new code:

- Canada:     225,650

- Ireland:    3,713,254

Here is a comparison of the the old code (top) and the new code (bottom) over 12 hours. Some things to notice:

 - fewer log processor lambda calls (never more than 12 per 5min!) but longer execution time and more memory used due to better aggregation
 - fewer rules engine lambda calls but longer execution time 

**master with log processor queue delay set to 0 (Ireland - us-west-1)**
image

**code from this PR (Canada - ca-central-1)**
image

The differences in SQS behavior is interesting to compare. First the load is constant and looks like:
image

Under this load, notice the top graph below from the old code is very responsive, as events come in they are immediately processed (with little opportunity to aggregate). In contrast, the new code oscillates around a mean level as it adaptively manages the load.

**master with log processor queue delay set to 0 (Ireland - us-west-1)**
image

**code from this PR (Canada - ca-central-1)**
image

Here we have the response of the new code scaling up when hit with 2000 files per minute. It quickly scales up and then falls into the cycle you see above.
image

Billing breakdown by service:

- CloudTrail  (Ireland) $222.26 (Canada) $113.71
- CloudWatch (Ireland) $34.03 (Canada) $24.24
- Lambda (Ireland) $91.11 (Canada) $96.83
  - NOTE: we expect Canada to be about the same or less but I did not exactly control for the sizes of files in each region. Canada had 2 times larger data files than Ireland which accounts for the slightly larger cost for Canada yet this PR still had an overall cost reduction.
- SNS (Ireland) $5.31 (Canada) $1.49
- SQS (Ireland) $26.26 (Canada) $8.95
- S3 (Ireland) $22.35 (Canada) $6.70
  - NOTE: This shows that S3 API calls are where the cost is NOT storage. Since Canada files were 2X larger than Ireland files the final total storage for Canada was 2 times larger than Ireland yet cost was about 1/4.

After the load test above I configure the old (Ireland) and the new (Canada) exactly the same in the `panther_config.yml` using the default self on boarding and let them run for 24 hours. The result is that the cost was about the same with the new code slightly cheaper:
image

## Changes

- Changed log processor to poll sqs every min
- Added a `rescale` go routine that will use the size of the queue to determine if to spawn more lambdas
- Removed the log processor queue delay
- Fixed a bug found while testing in the calculation of the maximum runtime for the lambda

## Testing

- mage test:ci
- mage deploy , compared master branch to this PR under load.


internal/log_analysis/log_processor/common/common.go
```diff
@@ -53,7 +54,7 @@ var (
 
 	Config EnvConfig
 
-	SQSWaitTime int64 // set by env
+	SQSWaitTimeSec int64 // set by env
```
rleighton, line 57: Made variable hold units for easier reading.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -52,54 +53,45 @@ Fewer, bigger files makes Athena queries much faster.
...
-	event events.SQSEvent,
 	processFunc ProcessFunc,
 	generateDataStreamsFunc func([]string) ([]*common.DataStream, error)) (int, error) {
 
-	// these cannot be named return vars because it would cause a data race
+	// this cannot be a named return var because it would cause a data race
 	var sqsMessageCount int
 
 	streamChan := make(chan *common.DataStream, 2*sqsMaxBatchSize) // use small buffer to pipeline events
-	processingDeadlineTime := deadlineTime.Add(-time.Duration(float32(time.Since(deadlineTime)) * processingTimeLimitScalar))
```
rleighton, line 79: This change fixes a bug I found while testing. The negative should not be there. This bug means we never actually stopped processing in 1/2 the lambda allotted time.

I made this into a function so it is easier to read and then added unit tests.

deployments/log_analysis.yml
```diff
@@ -63,11 +63,6 @@ Parameters:
     Description: Log processor Lambda memory allocation
     MinValue: 256 # 128 is too small, risks OOM errors
     MaxValue: 3008
-  LogProcessorSQSDelaySeconds:
```
rleighton, line 66: Clearly this turned  out to be bad idea. Taking out.

deployments/panther_config.yml
```diff
@@ -44,11 +44,7 @@ Infra:
   # Processing capability can be further increased by raising the
   # Lambda concurrent execution limit from default 1000, see:
   # https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html
-  LogProcessorLambdaMemorySize: 1024 # 256 - 3008, in 64MB increments
-
-  # Adding a small queue delay can improve the ability of the log processor
-  # to aggregate processed log files. The value must be between 0 and 20 seconds.
-  LogProcessorSQSDelaySeconds: 5
+  LogProcessorLambdaMemorySize: 2048 # 256 - 3008, in 64MB increments
```
rleighton, line 47: Given the increased aggregation, this needs to be larger than before. We want to avoid having customers change this because of OOMs. When the CloudTrail parser gets more efficient, let's revisit this default.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -38,10 +38,14 @@ import (
 )
 
 const (
-	processingMaxFilesLimit   = 5000 // limit this so there is time to delete from the queue at the end
```
rleighton, line 41: This is related to the bug fix below (see comment). Unrelated to the purpose of this PR but found during testing.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -127,22 +122,21 @@ func streamEvents(
 			}
 
 			// keep reading from SQS to maximize output aggregation
-			messages, messageReceipts, err := sqsbatch.ReceiveMessage(sqsClient,
-				common.Config.SqsQueueURL, common.SQSWaitTime)
```
rleighton, line 131: Avoid long receive waits, we learned the costs can add up over many events.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -173,14 +167,10 @@ func streamEvents(
 	return sqsMessageCount, nil
 }
 
-func lambdaDataStreams(event events.SQSEvent,
```
rleighton, line 176: Removed, we no longer read events in the args of the lambda.

deployments/panther_config.yml
```diff
@@ -111,7 +107,7 @@ Setup:
   # when processing large volumes of data.
   #
   # Access logs are sent either to the S3AccessLogsBucket below or a the panther audit bucket created for you.
-  EnableS3AccessLogs: true
```
rleighton, line 114: This should default false for cost reasons.

internal/log_analysis/log_processor/main/lambda.go
```diff
@@ -38,13 +37,13 @@ func main() {
 	lambda.Start(handle)
 }
 
-func handle(ctx context.Context, event events.SQSEvent) error {
```
rleighton, line 41: We no longer receive events from SQS.

internal/log_analysis/log_processor/processor/rescale.go
```diff
@@ -0,0 +1,106 @@
...
+	processingMaxLambdaInvoke = 1
+)
+
+var (
+	processingScaleDecisionInterval = defaultProcessingScaleDecisionInterval // var so we can set in tests
+)
+
+// scalingDecisions makes periodic adaptive decisions to scale up based on the sqs queue stats, it returns
+// immediately with a boolean stop channel (sending an event to the channel stops execution).
+func scalingDecisions(sqsClient sqsiface.SQSAPI, lambdaClient lambdaiface.LambdaAPI) chan bool {
```
nhakmiller, line 52: This is essentially exactly the orchestration that AWS is doing under the hood when sending SQS -> Lambda, correct? I know none of us are afraid of stepping away from pre-built solutions when they're insufficient, but I worry we lose some of the auto-scaling properties provided by AWS when we try to do this ourselves.

I wonder if a slightly modified architecture could help resolve the core issue. For example, an intermediate "batching" lambda that runs once a minute and simply pulls messages off the current queue then forwards them to a new queue for the log processor. Very similar to what you're doing here, but leaving the auto-scaling logic in AWS's hands. Of course this adds SQS costs, but we could run it as e.g. a 128 MB lambda and it would be pretty cheap overall.

This code is pretty straightforward and seems clean, I just hesitate to implement services that duplicate what AWS has already implemented (and implemented for longer and at a far larger scale than we have).

nhakmiller (commented):
In the hosted accounts we've also observed that the vast majority of s3 costs come from API calls. Glad this is being addressed.

internal/log_analysis/log_processor/processor/rescale.go
rleighton: The scaling code for SQS seems (base on me stressing it) to be very simple, just execute lambdas as fast as possible and you get throttled if you hit your limit. Not very graceful. This is I think what we want, where we have an exponential scale up under load and it is very simple code, no extra queues. I'll add to the above Conversation the graph of the response from where we went from tiny traffic to 2,000 messages per min.

rleighton: I added the graph of the new code scaling up.

rleighton: A batching lambda does not really help at all here if it is queue to queue (sqs limits). We want to have the fewest lambdas as we can and still keep up with the load.

austinbyers: > I wonder if a slightly modified architecture could help resolve the core issue. For example, an intermediate "batching" lambda that runs once a minute and simply pulls messages off the current queue then forwards them to a new queue for the log processor.

I second the suggestion to implement the SQS / dispatching logic with a separate processor instead of trying to manage all of it from the same function. When you have a function that tries to control its own concurrency, you run the risk of all kinds of recursive runaway bugs

rleighton: I'll do another short test where I start at 1000/min, then jump to 3000/min , then jump to 5000/min, then back to 3000/min, and  1000/min to show how this responds.

austinbyers: The way that can work:

1. You have a 128MB cheap function `sqs-dispatcher` invoked every minute that does long-polls of the SQS queue. Once it reaches X messages or waits for Y seconds, it sends a batch to a log processor
2. The log processor deletes the messages from the queue once they've been processed successfully

This way, you automatically invoke as many functions as you need as you pull from the queue, and the polling mechanism is cheap. I think having a 2048MB lambda function polling queues is not the best use of resources

rleighton: > The way that can work:
> 
> 1. You have a 128MB cheap function `sqs-dispatcher` invoked every minute (concurrency limit=1) that does long-polls of the SQS queue. Once it reaches X messages or waits for Y seconds, it sends a batch to a log processor
> 2. The log processor deletes the messages from the queue once they've been processed successfully
> 
> This way, you automatically invoke as many functions as you need as you pull from the queue, and the polling mechanism is cheap. I think having a 2048MB lambda function polling queues is not the best use of resources

Can you explain how this would minimize the lambdas used (and maximize aggregation)? This is the critical point for this change and why it works.

austinbyers (approved):
Talked with @rleighton offline about the two possible implementations:

1. Having a separate queue dispatcher Lambda. This adds a new Lambda function and increases total invocations, but separating the concerns makes the architecture easier to tune and reason about. The dispatcher can be lower memory and therefore cheaper. This is the strategy I've used in the past, but it does introduce a bottleneck on the queue dispatcher.

2. Doing "recursive" scaling as shown here - the same function invokes more copies of itself as needed. Each copy polls the queue directly. This is in theory more efficient, since the messages are polled, processed, and deleted in the same invocation. It also means you can have multiple consumers reading from the queue instead of just one.

While the recursive scaling feels a bit strange, it can actually be a very effective model - I've implemented versions of it before. Since @rleighton has already tested this pretty thoroughly, I'm fine with it! I like avoiding extra Lambda functions if we don't need them. 

Thanks for trying new things and adapting as needed

rleighton:
After the load testing  I configure by the old (Ireland) and the new (Canada) exactly the same in the panther_config.yml using the default self on boarding and let them run for 24 hours. The result is that the cost was about the same with the new code slightly cheaper. See graph added to the end of  "Conversation" above.

internal/log_analysis/log_processor/processor/rescale.go
alxarch: For 'done' channels we should be returning unbuffered `<-chan struct{}` and closing it from the controlling side once we are done. This allows to 'broadcast' the 'done' message to multiple receipients. Even better I would suggest we pass a context.Context, do a `ctx, cancel := context.WithCancel(ctx)` on it and return that back, keeping `cancel()` for the controlling goroutine.

alxarch: > The way that can work:
> 
>     1. You have a 128MB cheap function `sqs-dispatcher` invoked every minute (concurrency limit=1) that does long-polls of the SQS queue. Once it reaches X messages or waits for Y seconds, it sends a batch to a log processor
> 

One proposal for this scheme would be to give access to sources-api to that lambda as well (at least the map of `bucket/prefix` -> `source id`). This way we could switch to aggregation of messages _per-source_.
This would:
1. improve the probability of aggregations since events on a single source are more probable to be of the same log type
2. give us a chance to 'intervene' by splitting log types among sources. 
3. give s a chance to have a 'himem' and a 'lomem' instance of the log processor lambda and dispatch based on log type requirements, *for further cost reduction*.

If the deadline is near all per-source message lists that haven't reached the batch threshold would be aggregated into a single one and sent over to the log processor.


> 
> This way, you automatically invoke as many functions as you need as you pull from the queue, and the polling mechanism is cheap. I think having a 2048MB lambda function polling queues is not the best use of resources

I totally agree and having a separate dispatcher would allow us to test new dispatching strategies without affecting actual  log processor code.

alxarch (commented):
I side with @austinbyers' *original* approach on this and think that having a separate 'cheap' dispatcher lambda for poling queues is the best option. One way to avoid the 'bottleneck' issue is to lift the concurrent limit of 1 and allow the lambda to 'spill over' up to X minutes if there are still more than Y messages on the queue after the first minute.

rleighton:
> I side with @austinbyers' _original_ approach on this and think that having a separate 'cheap' dispatcher lambda for poling queues is the best option. One way to avoid the 'bottleneck' issue is to lift the concurrent limit of 1 and allow the lambda to 'spill over' up to X minutes if there are still more than Y messages on the queue after the first minute.

The issues with a dispatcher in front are 2:
  - a *singleton* won't scale, particularly a small one (small == slow)
  - the whole point of this PR is to minimize the lambda calls to maximize aggregation, this is key to making this work and all the proposals I've seen so far do not have this property (to my understanding).

We definitely should have a follow up on this but I need to push this for the release to reduce Snowpipe costs.

PR #1858, opened 2020-10-23
Make auto remediations more consistent

nhakmiller:
## Background

Auto remediations would not fire in the alert suppress window. This is not working as intended; the auto remediations should trigger on a per resource basis regardless of whether alerts are being sent or not. We already have a separate check in place that ensures we don't repeatedly try (and fail) to remediate a failing resource.

This elusive bug has plagued the last few release tests, because it will _sometimes_ work, depending on how many times you try and how long you wait.

## Changes

- Still attempt auto remediations if auto remediations are enabled and we are in the alert suppress window.

## Testing

- No testing, but I eyeballed the code really hard to be sure :) I can test if someone really wants me to...


nhakmiller:
I was considering adding the `sync-to-release` tag since it's a small bug fix, but I don't think this is so critical that it can't wait for 1.12.

austinbyers (approved):
Nice! I'd say this is a simple and important fix we should `sync-to-release`

PR #1860, opened 2020-10-23
Update default settings

nhakmiller:
## Background

These default values have been observed to drive up costs with questionable security value. Updating them.
 
## Changes

- Changed queue delay to 0 seconds
- Changed default log processor memory to 512 MB
- Disabled s3 access logs by default

## Testing

- We've been playing with these values in some of our hosted accounts to tune them in


deployments/master.yml
```diff
@@ -79,7 +79,7 @@ Parameters:
     Type: String
     Description: Enable S3 access logging for all Panther buckets. This is strongly recommended for security, but comes at an additional cost.
     AllowedValues: [true, false]
-    Default: true
+    Default: false
```
austinbyers, line 82: This feedback loop has already been addressed, I'm not sure if disabling S3 access logs are still necessary @rleighton @kostaspap ?

austinbyers: I'm hesitant to disable S3 access logs completely by default because then Panther will show its own S3 buckets as failing our policies

nhakmiller: That's a good callout. I did this based on the comment by @kostaspap:

```
Maybe a 1.10.2 could have such defaults. Also critical to change default behavior so that we don’t process S3access logs (s3 access logs to endless feedback loop)
```

My interpretation of this was it's critical to disable S3 access logging ingestion by default. But perhaps that wasn't his intent, or maybe we should have a setting to still enable these logs but not ingest them.

austinbyers: #1666 breaks the feedback loop, which already landed in 1.10, so I'm not sure what he was referring to

nhakmiller: Ok, I'll revert this part then.

deployments/master.yml
```diff
@@ -123,11 +123,11 @@ Parameters:
     Description: Log processor Lambda memory allocation. Increase to eliminate out-of-memory errors or reduce processing time (in exchange for higher cost)
     MinValue: 256 # 128 is too small, risks OOM errors
     MaxValue: 3008
-    Default: 1024
+    Default: 512
```
rleighton, line 126: we need to leave it at 1024 or we risk OOMs on CT data

kostaspap: My thinking for reducing the default value for the Log Processor Lambda is that for average scenarios, the 512MB would be enough and would lead to 50% price reduction for Log Processor compute costs.
The OOM problems occur when someone wants to process big CT files: such files require big and pretty “active” infrastructure and the average OSS will likely not encounter it. Users or big customers can always tweak it to make sure it covers their “bigger data” usecases. That's the purpose of a "default" value imo anyway - the one that works best for most people. 

Note that compared to when we tested CT last time, we have reduced the memory allocation required for CT by half -> https://github.com/panther-labs/panther/pull/1373

I think this reduction will benefit the avg OSS user. If some user encounters an occasional OOM we can always instruct them to increase the Lambda memory.

deployments/master.yml
kostaspap: I guess if feedback loop is already in, yeah, we can leave this as is :)

deployments/master.yml
rleighton: Let's leave it. In 1.12.0 I hope we are using https://github.com/panther-labs/panther/pull/1857 which completely changes the dynamic.

s0l0ist (approved):

kostaspap (approved):

PR #1863, opened 2020-10-25
Opsgenie eu region

### LINKED ISSUE #1846: OpsGenie destination doesn't support European region
OpsGenie destination can't be configured with region: currently doesn't support the European endpoints.
(more information at https://github.com/panther-labs/panther/blob/master/internal/core/alert_delivery/outputs/opsgenie.go#L28 )

### Steps to reproduce

1. Go to '.Settings/Destinations'
2. Click on '+ Add Destination'
3. Select OpsGenie
4. Enter API Key from an European OpsGenie account

### Expected behavior

Test send would work.

### Environment

Recommended cloudformation

s0l0ist:
## Background

Closes #1846 

Opsgenie supports two regions, US and EU. The EU region has an extra subdomain layer whereas the US does not.

## Changes

- Adds radio buttons to switch between the two regions (US/EU), by default it uses the US API region.
- Makes sure that existing Opsgenie destinations fallback to 
- Change custom webhook SVG color
- Add destination icon/logo to destination onboarding wizard

## Designs

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/0d293ff7-a6c6-4ddc-9dd0-442c63b91995/commits/ddcbd9c71d07822109f071f05607ba9db0ed2ced/files/8cc3cb34-8b80-405f-97ad-3adafc365a4f/layers/CF79EEA3-5D62-4179-B79B-02845E8CF5F6?mode=design&selected=root-7FFF4C6C-7893-4CD3-81B4-456EE7641EB0

## Screenshots

Screen Shot 2020-11-17 at 5 17 45 PM
Screen Shot 2020-11-17 at 5 17 51 PM

## Testing

- mage test:ci


kostaspap:
@s0l0ist what's up with this PR? Are we still going to fix this (1.13 I assume?)

s0l0ist:
> @s0l0ist what's up with this PR? Are we still going to fix this (1.13 I assume?)

Yes, will merge soon. Nearly forgot about this.

s0l0ist:
cc @dimitriszoitas please take a look and provide a link for the design when you're ready.

api/graphql/schema.graphql
```diff
@@ -608,6 +608,7 @@ type SqsDestinationConfig {
 
 type OpsgenieConfig {
   apiKey: String!
+  europeanServiceRegion: Boolean!
```
3nvi, line 611: In my mind, this should be `region` and accept an enum of `Europe, US`. I think it's a bit more cleaner that way and more understandable both from a developer and from a public-api perspective
```suggestion
  region: OpsgenieRegionEnum!
```

api/graphql/schema.graphql
```diff
@@ -676,6 +677,7 @@ input SqsConfigInput {
 
 input OpsgenieConfigInput {
   apiKey: String!
+  europeanServiceRegion: Boolean!
```
3nvi, line 680: Mirroring the above

```suggestion
  region: OpsgenieRegionEnum!
```

api/graphql/schema.graphql
```diff
@@ -676,6 +677,7 @@ input SqsConfigInput {
...
 }
```
3nvi, line 682: Of course we need an enum to restrict values on the GraphQL level

```graphql
enum OpsgenieRegionEnum {
  Europe
  US
}
```

api/lambda/outputs/models/api.go
```diff
@@ -229,7 +229,8 @@ type JiraConfig struct {
 
 // OpsgenieConfig defines options for each Opsgenie output
 type OpsgenieConfig struct {
-	APIKey string `json:"apiKey"`
+	APIKey                string `json:"apiKey"`
+	EuropeanServiceRegion bool   `json:"europeanServiceRegion"`
```
3nvi, line 233: If we decide to go with what I recommended, this should change to

```suggestion
	Region string   `json:"region" validate:"oneof=Europe US"`
```

web/src/components/forms/OpsgenieDestinationForm/OpsgenieDestinationForm.tsx
```diff
@@ -45,6 +46,7 @@ const OpsgenieDestinationForm: React.FC<OpsgenieDestinationFormProps> = ({
     outputConfig: Yup.object().shape({
       opsgenie: Yup.object().shape({
         apiKey: existing ? Yup.string() : Yup.string().required(),
+        europeanServiceRegion: Yup.boolean(),
```
3nvi, line 49: This should definitely have the `.required()` method regardless of whether we end up keeping it as a boolean or not

```suggestion
        region: Yup.string().required(),
```

web/src/components/forms/OpsgenieDestinationForm/OpsgenieDestinationForm.tsx
```diff
@@ -74,6 +76,16 @@ const OpsgenieDestinationForm: React.FC<OpsgenieDestinationFormProps> = ({
           required={!existing}
           autoComplete="new-password"
         />
+        <Field
```
3nvi, line 79: I feel it’s a bit cleaner if we just had a dropdown.

 I’m sure you add this text to explain what “European Service Region: ON” means, which is fair. People might get confused. That’s exactly why a switch might not be ideal in this use case.

3nvi (commented):
The effort is really good, but I think we can do a bit better on the UI of it (i.e. using a select instead of a switch).

Either way this will most likely be a small change :)

s0l0ist:
Changing to use string enums instead of booleans, @rleighton.

Changing preemptively to use a dropdown. Final UI will be determined by @dimitriszoitas

dimitriszoitas:
@s0l0ist I'm suggesting the layout found in [this mockup](https://share.goabstract.com/ae39d354-01f2-4ee3-80e2-deb444224737?mode=design&sha=55689277f37e9d649e808850945f8f13ae6a4f37)

Since we only have 2 options, it's better to display them both instead of having a switch or a dropdown, so we could go with a radio button solution. Let me know what you think

s0l0ist:
@dimitriszoitas, nice I like it. @3nvi thoughts? This will be in v1.13

internal/core/alert_delivery/outputs/opsgenie_test.go
```diff
@@ -67,7 +67,17 @@ func TestOpsgenieAlert(t *testing.T) {
...
+
+	requestEndpoint := ""
+	switch opsgenieConfig.ServiceRegion {
+	case OpsgenieServiceRegionUS:
+		requestEndpoint = "https://api.opsgenie.com/v2/alerts"
+	case OpsgenieServiceRegionEU:
+		requestEndpoint = "https://api.eu.opsgenie.com/v2/alerts"
+	default:
+		requestEndpoint = "https://api.opsgenie.com/v2/alerts"
+	}
```
kostaspap, line 79: Just `requestEndpoint := "https://api.opsgenie.com/v2/alerts"`. Or even better, let's test that the new code works: 
```
opsGenieRegions := []string{OpsgenieServiceRegionUS, OpsgenieServiceRegionEU}
expectedEndpoints := []string{"https://api.opsgenie.com/v2/alerts", "https://api.eu.opsgenie.com/v2/alerts"}
for i, region := opsGenieRegions {
 expectedPostInput := &PostInput{...}
 httpWrapper.On("post", expectedPostInput)....
 ... etc 
}

```

kostaspap (approved):
Looks good overall, just a small fix for the unit test

3nvi:
> @dimitriszoitas, nice I like it. @3nvi thoughts? This will be in v1.13

Whatever @dimitriszoitas says. Perhaps the text is not needed, but I'm ok with either :)

internal/core/alert_delivery/outputs/opsgenie.go
```diff
@@ -60,8 +63,18 @@ func (client *OutputClient) Opsgenie(
 		AuthorizationHTTPHeader: authorization,
 	}
 
+	requestEndpoint := ""
+	switch config.ServiceRegion {
+	case OpsgenieServiceRegionUS:
```
3nvi, line 68: optional: could we merge 1st and last case (if that's possible in go)

```go
	case OpsgenieServiceRegionUS:
	default:
		requestEndpoint = "https://api.opsgenie.com/v2/alerts"
```

web/src/components/forms/OpsgenieDestinationForm/OpsGenieDestinationForm.test.tsx
```diff
@@ -28,6 +28,7 @@ const emptyInitialValues = {
   outputConfig: {
     opsgenie: {
       apiKey: '',
+      serviceRegion: OpsgenieServiceRegionEnum.Us,
```
3nvi, line 31: lol that's such a weird casing 😄 I know it's autogen and we shouldn't modify it, but it's still fun

web/src/components/forms/OpsgenieDestinationForm/OpsGenieDestinationForm.test.tsx
```diff
@@ -68,12 +72,13 @@ describe('OpsGenieDestinationForm', () => {
       <OpsgenieDestinationForm onSubmit={() => {}} initialValues={emptyInitialValues} />
     );
     const displayNameField = getByLabelText('* Display Name');
+    const serviceRegion = getByLabelText('Service Region', { selector: 'input' });
```
3nvi, line 75: Have you tried if this works?

P.S. That applies to all similar cases found below

```suggestion
    const serviceRegion = getByLabelText('Service Region');
```

web/src/components/forms/OpsgenieDestinationForm/OpsgenieDestinationForm.tsx
```diff
@@ -57,14 +59,27 @@ const OpsgenieDestinationForm: React.FC<OpsgenieDestinationFormProps> = ({
...
           placeholder="How should we name this?"
           required
         />
+        <Box as="fieldset">
+          <Field
+            as={FormikCombobox}
+            name="outputConfig.opsgenie.serviceRegion"
+            items={[OpsgenieServiceRegionEnum.Us, OpsgenieServiceRegionEnum.Eu]}
+            label="Service Region"
+            aria-describedby="serviceRegion-helper"
```
3nvi, line 76:
```suggestion
            aria-describedby="serviceRegion-helper-text"
```

web/src/components/forms/OpsgenieDestinationForm/OpsgenieDestinationForm.tsx
```diff
@@ -57,14 +59,27 @@ const OpsgenieDestinationForm: React.FC<OpsgenieDestinationFormProps> = ({
...
           as={FormikTextInput}
           label="* Display Name"
           placeholder="How should we name this?"
           required
         />
+        <Box as="fieldset">
+          <Field
+            as={FormikCombobox}
+            name="outputConfig.opsgenie.serviceRegion"
+            items={[OpsgenieServiceRegionEnum.Us, OpsgenieServiceRegionEnum.Eu]}
```
3nvi, line 74: if we keep the combobox, we should define a variable 

```js
const opsgenieServiceRegions = [OpsgenieServiceRegionEnum.Us, OpsgenieServiceRegionEnum.Eu];
```

outside the Reactt component and use `<FastField>` instead of `<Field>`

web/src/components/forms/OpsgenieDestinationForm/OpsGenieDestinationForm.test.tsx
```diff
@@ -106,23 +117,28 @@ describe('OpsGenieDestinationForm', () => {
...
     const submitButton = getByText('Update Destination');
     expect(displayNameField).toHaveValue(initialValues.displayName);
     expect(submitButton).toHaveAttribute('disabled');
 
     const newDisplayName = 'New Opsgenie Name';
     fireEvent.change(displayNameField, { target: { value: newDisplayName } });
+
+    fireEvent.mouseDown(serviceRegion);
+    fireEvent.click(getByTestId('service-region-us'));
```
3nvi, line 140: I'm missing something. Where is this `data-testid` attribute added? Can't find it in this PR changes. Could it be a remnant of a previous implementation?

3nvi (commented):
Code looks fine, apart from a couple of minor comments that wouldn't otherwise block me. As soon as we change it to radio, we should do a quick re-review and close it

web/src/components/forms/OpsgenieDestinationForm/OpsGenieDestinationForm.test.tsx
s0l0ist: Doesn't work

web/src/components/forms/OpsgenieDestinationForm/OpsGenieDestinationForm.test.tsx
s0l0ist: I could not figure out how to get the test to simulate clicking on the list of items. After inspecting the dom by doing a dump to the console, this test Id is generated by jest and added as an attribute. So I just used it. 

This is going to change because now we will do radio buttons.

web/src/components/forms/OpsgenieDestinationForm/OpsgenieDestinationForm.tsx
s0l0ist: Well looks like we're tossing it

3nvi:
@s0l0ist do we plan to finish this within the current release to be added as part of 1.13?

s0l0ist:
> @s0l0ist do we plan to finish this within the current release to be added as part of 1.13?

yes

internal/core/alert_delivery/outputs/opsgenie.go
s0l0ist: I'd rather be explicit

3nvi (approved):

Issue #1687, opened 2020-10-07
Remove compliance-api gateway


austinbyers:
### Description

Remove the compliance API gateway (the Lambda handler will remain intact).

Once this first gateway is removed, the other internal gateways will be easier to remove.

### Related Services

All consumers will need to be changed to invoke the `compliance-api` Lambda function directly:

* `analysis-api`
* `resources-api`
* AppSync
* Cloud security backend

PR #1867, opened 2020-10-25
Remove compliance-api gateway

austinbyers:
## Background

The first step of the larger API project is to remove our internal API gateways, which add latency and little value in our current architecture.

This PR removes the `compliance-api` gateway; AppSync and backend services now communicate directly with the `compliance-api` Lambda function, which should reduce overall latency.

Closes: #1687 

## Changes

My goal was to keep this change as small as possible:

* Almost the exact same swagger model was ported to a Go models file
* The output from the Lambda function was left exactly the same

Summary of the changes needed:

* Added a new `gatewayapi.Invoke` helper function which backend services use to invoke the former Gateway handlers directly
    * This handles parsing the `APIGatewayProxyResponse` that all of those functions return - it allows me to keep the API output exactly the same as before
    * It is only temporary; it won't be necessary once we finish internal API consolidation
* Replaced the swagger specification with our standard `api/lambda/compliance/models/api.go` file format
* Updated AppSync to use Lambda instead of API gateway for the compliance-api data source
* Remove all compliance-api references from CloudFormation (and update IAM permissions for calling services)
* Updated integration and unit tests

---

Please do not obsess over details of the service itself - can we add cursor-based pagination? Can we avoid globals and pass context everywhere? Can we add the `errors` package? Etc. The answer is no, not yet - again, I want to keep the service unchanged as much as possible right now; when we consolidate the API services we can address those concerns.

## Testing

- `mage deploy test:integration` (compliance-api thankfully had pretty robust integration tests)
- Upgraded from the `master` branch to this one to ensure no migration issues
- Tested each page in the UI that relates to compliance status:
    - Cloud Security Overview
    - Policy list
    - Policy detail (with embedded resource list)
    - Resource list
    - Resource detail (with embedded policy list)


austinbyers:
Net 7,300 lines of code deleted! Incredibly satisfying 🔴

3nvi:
@austinbyers let's make sure that before we merge it, no conflict exists in the open PRs

pkg/gatewayapi/invoke.go
```diff
@@ -0,0 +1,110 @@
...
+// Invoke a former API gateway proxy Lambda directly.
+//
+// Unmarshals the response body into output and returns (http status code, error).
+// A non-nil error could be caused by:
+//     - failure to marshal request / unmarshal response
+//     - lambda function failed to invoke (does not exist, insufficient permissions)
+//     - lambda function runtime exception (panic, OOM, timeout)
+//     - status code is not 2XX
+//
+// This is similar to genericapi.Invoke and will be obsolete once we consolidate the internal API.
```
nhakmiller, line 61: Planned obsolescence 🍎

nhakmiller (approved):
Mostly reviewed the new code, skimmed a lot of the pieces that looked like translating variable names and moving pointers -> non-pointers. Overall LGTM 🚀

rleighton (approved):
Massive change! Thank you for doing this, that took a lot of work.

This is so big it is hard to really focus on the details but we have lots of tests and time burn this in.

pkg/gatewayapi/invoke.go
austinbyers: I guess it is, never thought of it that way

austinbyers:
> @austinbyers let's make sure that before we merge it, no conflict exists in the open PRs

It's impossible to know *in advance* of all possible PR conflicts, but:

* I checked open PRs; none of them are changing the `compliance-api`
* The graphQL models do *not* change at all in this PR
* I'll check for any conflicts after this merges and help resolve them if necessary

PR #1869, opened 2020-10-26
Fixed wording + issue with initialized filters for booleans

### LINKED ISSUE #1610: Default values in policy page filters unclear
### Describe the bug

I found the UI in the Policy filters a bit confusing. It is unclear what the default filter settings are when someone tries to add a filter. See the screenshots I attach below and the accompanying text.

### Steps to reproduce

1. Go to `Cloud Security`
2. Click  on `Policies`
3. Click on `Filter Options`
4. Select `Status` to be `Pass`
You will see a screen like the one below:

It might seem like a bug in Panther: Someone has selected to display "Passing" policies and it seems like we should be seeing only the "Enabled" ones but we are not... Mystery solved when clicking on "Enabled"

I think we should make more clear what the default filter values are when someone selects to filter on something. 

### Expected behavior

A more clear indication of what is the default value of the filters

### Environment

- Panther version or commit: v1.9.0-RC1

alexmylonas:
## Background

There is small request for text change for a filter in Policies referred to #1610. I also noticed and fixed another small bug that was replicated by:

1. Selecting a boolean filter to `false` (i.e. Enabled Polices to negative).
2. Refresh page
3. Notice that the results are correct but if you expand the filters the above field is not correct filled

Closes #1610 
## Changes

- Text change
- Corrected check for initial values of filters

## Testing

- Locally, multiple cases


web/src/components/utils/GenerateFiltersGroup/GenerateFiltersGroup.tsx
```diff
@@ -80,7 +80,9 @@ function GenerateFiltersGroup<T extends { [key: string]: any }>({
   // time. The value of `initialValues` doesn't get updated as the component updates, since we only
   // need it during form initialization (a.k.a. component mount-time)
   const initialValuesWithDefaults = React.useMemo(() => {
-    return mapValues(filters, (value, name) => initialValues[name] || defaultValues[name]) as T;
+    return mapValues(filters, (value, name) =>
+      initialValues[name] === undefined ? defaultValues[name] : initialValues[name]
```
alexmylonas, line 84: This needed to change because `initialValues[name]` may take `false` value which we result in incorrect initial value

web/src/pages/ListPolicies/ListPoliciesActions/ListPoliciesActions.tsx
```diff
@@ -87,12 +87,12 @@ export const filters = {
...
-      items: ['', 'true', 'false'],
+      items: ['', true, false],
       itemToString: (item: boolean | string) => {
-        if (!item) {
-          return 'All';
+        if (item === '') {
+          return 'All Policies';
         }
-        return item === 'true' ? 'Yes' : 'No';
+        return item === true ? 'Enabled Policies' : 'Disabled Policies';
```
alexmylonas, line 95: Using booleans instead of strings felt more valid

3nvi: I'm indifferent on this

3nvi (approved):
As long as you can guarantee that converting strings to booleans doesn't cause issue with filtering, it LGTM

Issue #1865, opened 2020-10-25
mage test:integration hangs (release 1.11)


rleighton:
### Describe the bug

`mage test:integration` never finishes.

Odd to see this since this should be part of the every release test. Maybe something in my env?

NOTE: I hit this on master doing tests before posting a PR. 

### Steps to reproduce

- checkout release-1.11
- `mage deploy`
- `mage test:integration`

### Expected behavior

Finishes.

### Environment

- release-1.11.0

### Screenshots

![image](https://user-images.githubusercontent.com/18419762/97113369-4b6e8c00-16c0-11eb-8f7b-62288f0fa029.png)

austinbyers:
> Odd to see this since this should be part of the every release test

We do run integration tests in every release, I have already run integration tests for `release-1.11` in both repos and they worked fine.

I tested them again on fresh deploys; they pass for both the current `release-1.11` branch and `master` (#1867) 

```
13:59:29	WARN	[test:integration]	Integration tests will erase all Panther data in account 101802775469 (us-west-2)
Are you sure you want to continue? (yes|no) yes
ok  	github.com/panther-labs/panther/cmd/opstools/requeue	22.351s
ok  	github.com/panther-labs/panther/cmd/opstools/s3queue	1.369s
ok  	github.com/panther-labs/panther/internal/compliance/compliance_api/main	3.416s
ok  	github.com/panther-labs/panther/internal/compliance/resources_api/main	2.757s
ok  	github.com/panther-labs/panther/internal/core/analysis_api/main	8.513s
ok  	github.com/panther-labs/panther/internal/core/layer_manager/main	32.809s
ok  	github.com/panther-labs/panther/internal/core/metrics_api/main	61.813s
ok  	github.com/panther-labs/panther/internal/core/organization_api/main	1.000s
ok  	github.com/panther-labs/panther/internal/core/outputs_api/main	2.086s
ok  	github.com/panther-labs/panther/internal/core/source_api/main	0.749s
ok  	github.com/panther-labs/panther/internal/core/users_api/main	3.545s
ok  	github.com/panther-labs/panther/internal/log_analysis/awsglue	3.488s
ok  	github.com/panther-labs/panther/internal/log_analysis/datacatalog_updater/process	0.670s
ok  	github.com/panther-labs/panther/pkg/awsathena	89.187s
14:04:23	INFO	[test:integration]	test:integration: python policy engine
.
----------------------------------------------------------------------
Ran 1 test in 1.095s

OK
```

[...]

PR #1870, opened 2020-10-26
Update use of custom retriers to only be used on clients (fixes `mage test:integration`)

rleighton:
## Background

`mage test:integration` is currently hanging on the data catalog updater , however the reason is due to the retrier code failing to talk to the ec2 meta data service when creating the session  and retrying over and over. Something must have changed in the SDK as this used to work.

The fix is to add custom retriers only on the clients and not on the session itself where auth is done.

Closes:https://github.com/panther-labs/panther/issues/1865

## Changes

- Move retriers to the client instantiations and not on the aws session

## Testing

- mage test:ci
- mage test:integration
- mage deploy, checked that logs were processing w/out any errors


internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -294,9 +294,9 @@ func getNewS3Client(region *string, creds *credentials.Credentials) (result s3if
 	}
 	// We have seen that in some case AWS will return AccessDenied while accessing data
 	// through STS creds. The issue seems to disappear after some retries
-	session := session.Must(session.NewSession(request.WithRetryer(config,
+	awsSession := session.Must(session.NewSession(config)) // use default retries for fetching creds, avoids hangs!
+	return s3.New(awsSession.Copy(request.WithRetryer(config.WithMaxRetries(s3ClientMaxRetries),
```
kostaspap, line 298: nit: The change here is not really required since the session is used only locally.

kostaspap (approved):

PR #1873, opened 2020-10-26
fix: update options passed on listeners

### LINKED ISSUE #1827: Log Analysis Overview: Selecting a logtype messes up other options
### Describe the bug

On the Log Analysis Overview page, there are several configurable options to change how the graphs look. There is at least the interval option, time frame option, and linear/logarithmic scale option. When you set any of these away from the default value, then select a specific log type on one of the graphs, it silently/invisibly resets the configurable options.

### Steps to reproduce

1. Go to Log Analysis overview page (the 1.11  release environment should work)
1. On the `Alert Severity` graph, select the `Logarithmic` option. The graph should be re-drawn in a logarithmic scale
1. On the left hand side of the graph, select just one severity (that has values). The graph will be re-drawn with just the values of the selected severity, but now in a linear scale. The `Logarithmic` setting will still appear to be selected
1. Select the `Linear` button again. Nothing will happen because the graph is already in a linear scale.
1. Select the `Logarithmic` button again. Now the graph will be re-drawn in a logarithmic scale.

This same behavior applies to the interval and time frame options, and applies whenever you select a single element from the line chart. That is, if you are looking at the events by log type graph with all log types selected, set a non-default interval or timeframe (or both), then select just one log type, then when the graph is re-drawn it will again have the default interval and timeframes set despite the page showing your non-default interval and timeframes.

### Expected behavior

Either:
- Selecting the log type/severity does not reset the other options (preferred)
- Or selecting the log type/severity does reset the other options, and the page is updated to reflect this

### Environment

- Panther version v1.11 candidate release

### Screenshots

These screenshots are a bit awkward because most of the values go away in the logarithmic scale, but notice how in the second screenshot all the alerts are selected and the scale is logarithmic and in the third screenshot only the `Info` severity alerts are selected and the scale is linear, despite the graph clearly showing the `Logarithmic` button selected.

Screen Shot 2020-10-21 at 9 30 12 AM
Screen Shot 2020-10-21 at 9 30 31 AM
Screen Shot 2020-10-21 at 9 31 50 AM

alexmylonas:
## Background

This PR is fixing issues reported on #1827.

Closes #1827 

## Changes

- Update listeners when `chartOptions` are updated
- Kill listeners before enabling them

## Testing

- Locally


3nvi (approved):
Looks good @alexmylonas . I think that does it :)

PR #1874, opened 2020-10-27
Destination Success hook + adding indication to Alert Card

### LINKED ISSUE #1756: Add "failing delivery information" to Alert Card
### Description

Each alert card shows information about the alert, but it doesn't give any info as to whether the alert was successfully delivered or not.

A user can know if the delivery was successful, by visiting the details page of the alert and viewing the banner at the bottom of the page, as shown in the screenshot below

Screen Shot 2020-10-15 at 3 21 55 PM

This task wants to do the same thing for each Alert Card. Specifically, each alert will have a simple text message, notiifying the user that the delivery of this alert has failed. If the delivery was successful, no additional information will be rendered (i.e. keep what we currently have in our UI).

### Related Services

Alerts

### Designs

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/70fbde80-1195-4df9-a583-6962f57e3e19/commits/ba31ffa546856d449e2a15fd451d1b3da8c4db9d/files/c2bbc1a8-4778-4cd6-93a5-4630d8868201/layers/2707278E-4CD9-4F3C-919E-1E3BCC4A0402

### Acceptance Criteria

- The logic that decides whether an alert is successful or not is isolated and placed into a re-usable hook (currently present in the Alert Details page)
- This hook is used within each `AlertCard`
- The output of this hook decides whether the Card should render a message with red text  or not. It could also return the message to render,  along with a status (error/null/success)
- As a side-effect, make sure to hide the "Edit rule" button if the `AlertCard` is within the rule details page (use a prop for that named `hideRuleButton`)
- Snapshot & UI tests are added on the `AlertCard` component

alexmylonas:
## Description

Each Alert can have destinations that can fail to be delivered. We want to showcase when a destination for alert fails to be delivered on `AlertCards`.

- Added a hook for checking if all `deliveryResponses` are successfully delivered
- Added optional text on AlertCards
- Added a prop on AlertCard for hiding "View Rule" LinkButton

Closes #1756

## Design

Mocks [here](https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/70fbde80-1195-4df9-a583-6962f57e3e19/commits/ba31ffa546856d449e2a15fd451d1b3da8c4db9d/files/c2bbc1a8-4778-4cd6-93a5-4630d8868201/layers/2707278E-4CD9-4F3C-919E-1E3BCC4A0402?mode=build&selected=root-89F89A16-A0A3-4626-88D7-4B2C75F0B555)

### Screenshots
Screenshot 2020-10-27 at 10 53 42 AM
Screenshot 2020-10-27 at 11 23 23 AM

### Checklist:

- [x] ~~Make sure the changes are responsive (appearing decently on devices)~~


web/src/components/RelatedDestinations/RelatedDestinations.tsx
```diff
@@ -98,7 +98,7 @@ const RelatedDestinations: React.FC<RelatedDestinationsSectionProps> = ({
 
   return (
     <Flex align="center" spacing={2} mt={1}>
-      {destinations.map(getLogo)}
+      {destinations.length ? destinations.map(getLogo) : 'N/A'}
```
3nvi, line 101:
```suggestion
      {destinations.length ? destinations.map(getLogo) : 'Not configured'}
```

I think it's a bit more explicit likee that, since N/A maps to "something is not available". Perhaps we need a different text color for that as well like the one we use in `GenericItemCard.Value` when something isn't available

web/src/components/cards/AlertCard/AlertCard.test.tsx
```diff
@@ -63,10 +81,32 @@ describe('AlertCard', () => {
     const destination = buildDestination({ outputId, outputType: DestinationTypeEnum.Slack });
     const mocks = [mockListDestinations({ data: { destinations: [destination] } })];
 
-    const { getByAriaLabel, getByAltText } = render(<AlertCard alert={alertData} />, { mocks });
+    const { getByAriaLabel, getByAltText, container } = render(<AlertCard alert={alertData} />, {
```
3nvi, line 84: nice

web/src/components/cards/AlertCard/AlertCard.tsx
```diff
@@ -49,19 +54,21 @@ const AlertCard: React.FC<AlertCardProps> = ({ alert }) => {
...
-                variantColor="navyblue"
-                size="medium"
-              >
-                View Rule
-              </LinkButton>
-            }
-          />
+          {!hideRuleButton && (
+            <GenericItemCard.Value
+              id="link-to-rule"
```
3nvi, line 59:
```suggestion

```

I feel that since the `AlertCard` can be present multiple times in a page, there are going to be multiple elements with the same `id` in the page

web/src/components/cards/AlertCard/AlertCard.tsx
```diff
@@ -28,15 +28,20 @@ import { AlertSummaryFull } from 'Source/graphql/fragments/AlertSummaryFull.gene
...
   alert: AlertSummaryFull;
+  hideRuleButton?: boolean;
 }
 
-const AlertCard: React.FC<AlertCardProps> = ({ alert }) => {
+const AlertCard: React.FC<AlertCardProps> = ({ alert, hideRuleButton = false }) => {
   const { alertDestinations, loading: loadingDestinations } = useAlertDestinations({ alert });
-
+  const { allDestinationDeliveredSuccessfully } = useDestinationsDeliverySuccess({
+    deliveryResponses: alert.deliveryResponses,
```
3nvi, line 42: the `useDestinationsDeliverySuccess` only relates to Alerts, since the "thing" that is delivered is **always** going too be an Alert.

With this in mind, I'd suggest we change the params to be an `alert` instance (either a summary or a details, both will do just fine).

The hook - inside it -  can call `useAlertDestinations` and get the `deliveryResponses` from the `alert` param that you passed.

Seeing the code below, I think that this will lead to a more readable codebase. Thoughts?

web/src/components/cards/AlertCard/AlertCard.tsx
```diff
@@ -89,6 +96,15 @@ const AlertCard: React.FC<AlertCardProps> = ({ alert }) => {
             <UpdateAlertDropdown alert={alert} />
           </Flex>
         </GenericItemCard.ValuesGroup>
+        {!allDestinationDeliveredSuccessfully && (
+          <Flex align="center" spacing={2} mt={2} aria-label="Destination delivery failure">
```
3nvi, line 100:
```suggestion
          <Flex as="section" align="center" spacing={2} mt={2} aria-label="Destination delivery failure">
```

3nvi (approved):
Awesome job @alexmylonas. Really good!

I've left some comments about small enhancements, but nothing that should block this PR. Kudos 🎉

web/src/components/cards/AlertCard/AlertCard.tsx
alexmylonas: My only argument against this that both arguments are already available and used on parent components. 

Do you feel strongly about this?

web/src/components/cards/AlertCard/AlertCard.tsx
alexmylonas: Fixed!

web/src/components/cards/AlertCard/AlertCard.tsx
alexmylonas: Removed

web/src/components/RelatedDestinations/RelatedDestinations.tsx
alexmylonas: Updated to `Not Configured` with smaller `fontSize` to takes less space

PR #1877, opened 2020-10-27
Add gork and fastmatch packages for parsing text into key/value pairs

alxarch:
## Background

Custom logs require to parse text and extract key/value pairs.
This PR provides two packages that implement engines for parsing text.
One is called `gork` and provides `grok`-like regular expression expansion.
The other is called `fastmatch` and provides a much faster alternative that is less flexible
than full regex.

## Changes

- Adds `gork` package
- Adds `fastmatch` package
- The PR does not affect other parts of the code only the packages

## Testing

- Unit tests


giorgosp (commented):
LGTM. I'd just suggest to move the packages under a better directory, since `pkg/x` is not very helpful to someone reading the codebase. Same for `gork` and `fastmatch`. For example, a package like `textparsers` could give a hint at what's inside at least

internal/core/outputs_api/main/integration_test.go
```diff
@@ -80,7 +80,7 @@ func TestIntegrationAPI(t *testing.T) {
 	require.NoError(t, testutils.ClearDynamoTable(awsSession, tableName))
 
 	// Add one of each output type in parallel.
-	t.Run("Add", func(t *testing.T) {
+	t.Run("Set", func(t *testing.T) {
```
kostaspap, line 83: ?

pkg/x/fastmatch/fastmatch.go
```diff
@@ -0,0 +1,172 @@
...
+	prefix     string
+	delimiters []delimiter
+	fields     []string
+	scratch    []rune
+}
+
+type delimiter struct {
+	match string
+	name  string
+	quote byte
```
kostaspap, line 39: Is `match` something like an expression or a pattern? Is it it the actual matched element?  
What is a quote? Can you add a description here, it will be easier for people reviewing the code to understand it.

pkg/x/fastmatch/fastmatch.go
```diff
@@ -0,0 +1,172 @@
...
+		delimiters = append(delimiters, d)
+		if d.name != "" {
+			fields = append(fields, d.name)
+		}
+	}
+	return &Pattern{
+		prefix:     prefix,
+		delimiters: delimiters,
+		fields:     fields,
+	}, nil
```
kostaspap, line 73: Can you add an example of a pattern in a comment? It would help a lot trying to understand the logic

alxarch:
> LGTM. I'd just suggest to move the packages under a better directory, since `pkg/x` is not very helpful to someone reading the codebase. Same for `gork` and `fastmatch`. For example, a package like `textparsers` could give a hint at what's inside at least

The packages are there since they are 'exportable' packages. They are generic in nature and independant from each other. `pkg/` is for such packages. `pkg/x` is for new and somewhat experimental packages. Once we stabilize their API we can move them to `pkg/`. `pkg/` should not have long paths for 'categories' of packages (with the exception of `x/` which marks the package as experimental.  A subsequent PR will implement customlogs text parsing using these packages.

pkg/x/fastmatch/fastmatch.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (d *delimiter) reset(tag, match, prev string) {
+	quote := prevQuote(prev)
+	if quote != nextQuote(match) {
+		quote = 0
+	}
+	d.name = tag
```
kostaspap, line 81: Can we stick to one naming convention for it either `name` or `tag`? It would make reading the code easier.

pkg/x/fastmatch/fastmatch.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	return 0
+}
+
+var (
+	errMatch          = errors.New("match failed")
+	errInvalidPattern = errors.New("invalid pattern")
+)
+
+// MatchString matches src and appends key/value pairs to dst
```
kostaspap, line 111: 1. Why not just `func (p *Pattern) MatchString(dst []string, src string) error`?
2. Can you also extend the description to explain what are the values returned? `error` is clear, but the `[]string` isn't.

pkg/x/fastmatch/fastmatch.go
```diff
@@ -0,0 +1,172 @@
...
+	tail := src
+	if prefix := p.prefix; len(prefix) <= len(tail) && tail[:len(prefix)] == prefix {
+		tail = tail[len(prefix):]
+	} else {
+		return dst, errMatch
+	}
+	matches := dst
+	delimiters := p.delimiters
+	for i := range delimiters {
+		d := &delimiters[i]
```
kostaspap, line 122: Why not just
```
for _, d := range p.delimiters {
}
```
Is it for performance?

pkg/x/gork/builtin.go
```diff
@@ -0,0 +1,102 @@
...
+MONTH              \b(?:Jan(?:uary)?|Feb(?:ruary)?|Mar(?:ch)?|Apr(?:il)?|May|June?|July?|Aug(?:ust)?|Sep(?:tember)?|Oct(?:ober)?|Nov(?:ember)?|Dec(?:ember)?)\b
+MONTHNUM           0?[1-9]|1[0-2]
+MONTHNUM2          0[1-9]|1[0-2]
+MONTHDAY           (?:0[1-9])|(?:[12][0-9])|(?:3[01])|[1-9]
+DAY                \b(?:Mon(?:day)?|Tue(?:sday)?|Wed(?:nesday)?|Thu(?:rsday)?|Fri(?:day)?|Sat(?:urday)?|Sun(?:day)?)\b
+YEAR               (?:\d\d){1,2}
+HOUR               2[0123]|[01]?[0-9]
+MINUTE             [0-5][0-9]
+SECOND             (?:[0-5]?[0-9]|60)(?:[:.,][0-9]+)?
+KITCHEN            %{HOUR}:%{MINUTE}
```
kostaspap, line 78: Is this a thing? ^^

pkg/x/gork/gork.go
```diff
@@ -0,0 +1,284 @@
...
+			return 0, errors.Errorf("invalid pattern name %q in tag %q of pattern %q", name, tag, root)
+		}
+		if field != "" && !validFieldName.MatchString(field) {
+			return 0, errors.Errorf("invalid field name %q in tag %q of pattern %q", field, tag, root)
+		}
+		for _, visited := range visited {
+			if visited == name {
+				return 0, errors.Errorf("recursive pattern %q %v", root, visited)
+			}
+		}
```
kostaspap, line 219: Nice

pkg/x/gork/gork.go
```diff
@@ -0,0 +1,284 @@
...
+		expr:  expr,
+		names: expr.SubexpNames()[1:],
+	}, nil
+}
+
+func (e *Env) lookup(name string) *Pattern {
+	if p, ok := e.patterns[name]; ok {
+		return p
+	}
+	return nil
```
kostaspap, line 262: just 
```
return e.patterns[name]
```
?

pkg/x/gork/gork_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// nolint:lll
+func TestMatchString(t *testing.T) {
+	assert := require.New(t)
+	env := New()
+	src := `%{DATA:remote_ip} %{DATA:identity} %{DATA:user} \[%{HTTPDATE:timestamp}\] "%{DATA:method} %{DATA:request_uri} %{DATA:protocol}" %{DATA:status} %{DATA:bytes_sent}$`
```
kostaspap, line 32: So the pattern finishes with `$` but doesn't start with `^`. Is that intentional for some reason?

kostaspap (commented):
Looks good overall!
While reviewing `grok` was easy, I wouldn't say the same for `fastmatch`, maybe you can add some more comments there (I have put comments appropriate)

kostaspap (approved):

internal/core/outputs_api/main/integration_test.go
alxarch: goland refactor gone wrong. I suggest we adopt an acronym for this: `GRGW`

pkg/x/fastmatch/fastmatch.go
alxarch: nope, tag is the stuff inside the `%{` `}` name is the name of the field. It is possible we want to add 'options' to tags apart from just the name (think go struct tags i.e. `json:"foo,omitempty"`). Thus the distinct names.

pkg/x/fastmatch/fastmatch.go
alxarch: The pairs are appended to dst as stated in the function docs. We *must* return the matches otherwise the caller has no access to them. This is a very common pattern in performance-oriented code. We use append to avoid allocations by reusing a buffer provided by the caller (see `strconv.AppendFoo` and `time.AppendFormat`). Usually it's a `[]byte` buffer but the pattern works with any slice type.

pkg/x/fastmatch/fastmatch.go
alxarch: Yup, we avoid the copy of the full delimiter element to the stack.

pkg/x/gork/builtin.go
alxarch: Apparently so. Saw it in various grok implementations.

pkg/x/gork/gork_test.go
alxarch: Nope, just another pattern.

pkg/x/gork/gork.go
alxarch: like this it's equivalent. The initial code was looking into stuff inside `Pattern` to dirstinguish 'private' and 'public' patterns. That seemed too complex for no good reason.

pkg/x/fastmatch/fastmatch.go
alxarch: And also we always use `[]Foo` instead of `[]*Foo` in performance-sensitive code to have cache-coherence.

pkg/x/fastmatch/fastmatch.go
alxarch: fixed

pkg/x/gork/gork.go
alxarch: fixed

pkg/x/fastmatch/fastmatch.go
alxarch: added some more function docs

pkg/x/fastmatch/fastmatch.go
alxarch: fixed

PR #1878, opened 2020-10-27
Add cloudsecurity database and tables to Glue

rleighton:
## Background

Creates the cloudsecurity database and the resources and compliance tables in Glue.

Closes: https://github.com/panther-labs/panther-enterprise/issues/921
Closes: https://github.com/panther-labs/panther-enterprise/issues/767

## Changes

- rename the current custom resource from "glue" to "log processor tables"
- add a new custom resource for cloud security that creates the tables
- add a pkg that creates the tables
- created an Athena workgroup called AmazonAthenaPreviewFunctionality to get the DDB connector
- added CF to create the DDB connector
- movde the Athena related CF to bootstrap stack and removed code from custom resource

## Testing

- mage test:ci
- mage deploy
- mage teardown
- ran some queries by manually setting up the ddb conngit ector

image

image

image

image


internal/compliance/awsglue/awsglue.go
```diff
@@ -0,0 +1,273 @@
...
+			// per https://github.com/awslabs/aws-athena-query-federation/tree/master/athena-dynamodb
+			"classification": aws.String("dynamodb"),
+			"sourceTable":    aws.String(ResourcesTableDDB),
+			// for attrs with upper case
+			"columnMapping": aws.String(`expiresat=expiresAt,lastmodified=lastModified,integrationtype=integrationType`),
+		},
+		StorageDescriptor: &glue.StorageDescriptor{
+			Location: &locationARN,
+
+			// FIXME: add descriptions to each field
```
rleighton, line 99: @nhakmiller Can pls comment as to the description (i.e., Comment) is for each attribute below?

internal/compliance/awsglue/awsglue.go
```diff
@@ -0,0 +1,273 @@
...
+			// per https://github.com/awslabs/aws-athena-query-federation/tree/master/athena-dynamodb
+			"classification": aws.String("dynamodb"),
+			"sourceTable":    aws.String(ComplianceTableDDB),
+			// for attrs with upper case
+			"columnMapping": aws.String(`errormessage=errorMessage,expiresat=expiresAt,lastupdated=lastUpdated,resourcetype=resourceType`),
+		},
+		StorageDescriptor: &glue.StorageDescriptor{
+			Location: &locationARN,
+
+			// FIXME: add descriptions to each field
```
rleighton, line 196: @nhakmiller Can pls comment as to the description (i.e., Comment) is for each attribute below?

internal/compliance/awsglue/awsglue.go
nhakmiller: Sure:

integrationType: Indicates what type of integration this resource came from. [hardcoded as `aws` right now]
deleted: Whether this is the snapshot of a deleted resource.
integrationId: The unique ID indicating which source integration this resource came from.
attributes: The JSON representation of the resource.
lowerid: The resource ID converted to all lower case letters.
lastmodified: When the most recent scan of this resource occurred.
id: The panther wide unique identifier of the resource.
type: The type of resource (see https://docs.runpanther.io/cloud-security/resources)
expiresat: Unix timestamp representing when this resource will age out of the resources table.

internal/compliance/awsglue/awsglue.go
nhakmiller: Sure


lastupdated - That last date the specified policy was evaluated against the specified resource.
resourceid - The panther wide unique identifier of the resource being evaluated.
policyseverity - The severity of the policy being evaluated.
policyid - The unique identifier of the policy being evaluated.
integrationid - The unique ID indicating which source integration the specified resource originated from.
suppressed - Whether this compliance status is currently being omitted from compliance findings.
expiresat - The unix timestamp when this compliance finding will age out.
resourcetype - The type of the specified resource.
status - Whether the policy evaluation of this resource resulted in a PASS, FAIL, or ERROR state.
errormessage - If an error occurred, the associated error message.

internal/compliance/awsglue/awsglue.go
nhakmiller: Copying 1:1 is fine if that's a technical limitation, but ideally we could omit `lowerid` and possibly `expiresat` from these fields as they're not really relevant to users. `integrationtype` is also kind of meaningless today, but could be useful at some  point in the future.

nhakmiller (approved):
Seems fine, but I wonder how advanced can the connector be? Could we lookup fields in other dynamo tables to enrich stuff? Thinking it might be nice to map `integrationid` -> `integrationlabel` since we do that elsewhere.

This is just a straight dump on dynamo rows -> data lake which I'm guessing is the easiest way to do it, but the dynamo rows were not designed to be viewed by end users necessarily. Nothing harmful here really,  just not optimized for end users to play around with.

This will go a long way to enabling reporting on Cloud Security status though, so kudos for that.

internal/core/custom_resources/resources/athena.go
```diff
@@ -40,8 +51,29 @@ func customAthenaInit(_ context.Context, event cfn.Event) (string, map[string]in
...
+			if awsutils.IsAnyError(err, "InvalidRequestException") {
+				if err := awsathena.WorkgroupAssociateS3(awsSession, previewWorkgroup, props.AthenaResultsBucket); err != nil {
+					return resourceID, nil, fmt.Errorf("failed to associate %s Athena workgroup with %s bucket: %v",
+						previewWorkgroup, props.AthenaResultsBucket, err)
+				}
+			} else {
+				return resourceID, nil, fmt.Errorf("failed to create %s Athena workgroup with %s bucket: %v",
+					previewWorkgroup, props.AthenaResultsBucket, err)
+			}
+		}
```
serbantanasa, line 75: I was unsure in my implementation on how to ensure that dynamo queries are sent to this workgroup, and not the default one. There does not seem to be a programmatic way of switching the default workgroup.

serbantanasa (approved):

internal/core/custom_resources/resources/athena.go
rleighton: Correct. I will make the change in the enterprise sync

deployments/log_analysis.yml
```diff
@@ -121,20 +121,11 @@ Conditions:
...
-      CustomResourceVersion: !Ref CustomResourceVersion
-      ServiceToken: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-cfn-custom-resources
-
   ###### Update Glue Table Schemas for Deployed Tables #####
   UpdateGlueTables:
     DependsOn:
-      - AthenaConfigure
       - UpdaterFunction
-    Type: Custom::UpdateGlueTables
+    Type: Custom::UpdateLogProcessorTables
```
austinbyers, line 128: You cannot change the type of an existing resource - this will break all upgrades:

```
16:14:15	ERROR	[deploy]	    X panther-log-analysis failed (5/9): stack panther-log-analysis change set is FAILED: Update of resource type is not permitted. The new template modifies resource type of the following resources: [UpdateGlueTables]
```

serbantanasa:
I think the rename there is aesthetic and can be rolled back (Russ wanted to clarify which glue tables are being updated, and make it clear that not all glue tables are being defined here to avoid future confusion). 

However, it seems like this kind of issue will make our future deploys quite restricted and/or confusing if we want to stick to non-breaking changes.

rleighton:
I fixed this already

austinbyers:
> However, it seems like this kind of issue will make our future deploys quite restricted and/or confusing if we want to stick to non-breaking changes.

If you want to keep a new resource type, just create a new resource entirely (change the logical ID / yml string). As long as your custom resource is idempotent (can survive a delete + create), then you can change its type by just recreating it with a different ID

> I fixed this already

@rleighton , no, it isn't fixed. I saw your partial revert, but the custom type is still not correct. I just deployed, and I can confirm in master that the type is still `UpdateLogProcessorTables` https://github.com/panther-labs/panther/blob/master/deployments/log_analysis.yml#L131

rleighton:
If I change the logical and physical, then we are good, correct?

These are just tags to us, so no issues with anything persistent

austinbyers:
> If I change the logical and physical, then we are good, correct?

Yes, if you change the logical ID, you will be fine as long as you're fine with the old resource being destroyed. (CloudFormation will create the new ressource and then, after the stack is updated, destroy the old resource.)

rleighton:
Ok, PR to you in a few.

I’ll deploy 1.11 then test the change and post PR

internal/compliance/awsglue/awsglue.go
```diff
@@ -0,0 +1,281 @@
...
+const (
+	CloudSecurityDatabase            = "panther_cloudsecurity"
+	CloudSecurityDatabaseDescription = "Hold tables related to Panther cloud security scanning"
+
+	// https://github.com/awslabs/aws-athena-query-federation/tree/master/athena-dynamodb
+
+	// FIXME: Update the description when the DDB connector is GA
+	ResourcesTableDDB         = "panther-resources"
+	ResourcesTable            = "resources"
+	ResourcesTableDescription = "(ddb.panther_cloudsecurity.panther-resources) The resources discovered by Panther scanning"
```
kostaspap, line 40: 1. What's the `ddb` it starts with? 
2. The tables should be `panther_compliance` - we use underscores not dashes to separate words. Same for resources

internal/core/custom_resources/resources/cloud_security_tables.go
```diff
@@ -0,0 +1,88 @@
...
+		if err := updateCloudSecurityTables(&props); err != nil {
+			zap.L().Error("failed to update glue tables", zap.Error(err))
+			return physicalResourceID, nil, err
+		}
+		return physicalResourceID, nil, nil
+	case cfn.RequestDelete:
+		zap.L().Info("deleting database", zap.String("database", cloudsecglue.CloudSecurityDatabase))
+		if _, err := awsglue.DeleteDatabase(glueClient, cloudsecglue.CloudSecurityDatabase); err != nil {
+			var awsErr awserr.Error
+			if errors.As(err, &awsErr) && awsErr.Code() == glue.ErrCodeEntityNotFoundException {
```
kostaspap, line 59: use `awsutils.IsAnyError`

internal/compliance/awsglue/awsglue.go
rleighton: I'll fix the description, should be "ddb.panther_cloudsecurity.resources)".

internal/core/custom_resources/resources/cloud_security_tables.go
rleighton: Will fix. I will have a PR to tidy a some small things I want to change in OSS based on the merge to enterprise and will put this into it.

Issue #1813, opened 2020-10-21
resources-api has flaky integration test


austinbyers:
### Describe the bug

When running `mage test:integration`, we occasionally see this error:

```
ok  	github.com/panther-labs/panther/cmd/opstools/requeue	24.728s
ok  	github.com/panther-labs/panther/cmd/opstools/s3queue	1.217s
ok  	github.com/panther-labs/panther/internal/compliance/compliance_api/main	11.783s
--- FAIL: TestIntegrationAPI (5.07s)
    --- FAIL: TestIntegrationAPI/GetResource (0.44s)
        --- FAIL: TestIntegrationAPI/GetResource/GetSuccess (0.15s)
            integration_test.go:260:
                	Error Trace:	integration_test.go:260
                	Error:      	Not equal:
                	            	expected: &models.Resource{Attributes:map[string]interface {}{"Panther":"Labs"}, ComplianceStatus:"PASS", Deleted:false, ID:"arn:aws:s3:::my-bucket", IntegrationID:"df6652ff-22d7-4c6a-a9ec-3fe50fadbbbf", IntegrationType:"aws", LastModified:models.LastModified{wall:0xc0a4680, ext:63738837594, loc:(*time.Location)(nil)}, Type:"AWS.S3.Bucket"}
                	            	actual  : &models.Resource{Attributes:map[string]interface {}{"Panther":"Labs"}, ComplianceStatus:"ERROR", Deleted:false, ID:"arn:aws:s3:::my-bucket", IntegrationID:"df6652ff-22d7-4c6a-a9ec-3fe50fadbbbf", IntegrationType:"aws", LastModified:models.LastModified{wall:0xc0a4680, ext:63738837594, loc:(*time.Location)(nil)}, Type:"AWS.S3.Bucket"}

                	            	Diff:
                	            	--- Expected
                	            	+++ Actual
                	            	@@ -4,3 +4,3 @@
                	            	  },
                	            	- ComplianceStatus: (models.ComplianceStatus) (len=4) "PASS",
                	            	+ ComplianceStatus: (models.ComplianceStatus) (len=5) "ERROR",
                	            	  Deleted: (models.Deleted) false,
                	Test:       	TestIntegrationAPI/GetResource/GetSuccess
FAIL
FAIL	github.com/panther-labs/panther/internal/compliance/resources_api/main	5.395s
FAIL
[...]

Issue #1688, opened 2020-10-07
Remove resources-api gateway


austinbyers:
## Description
Remove the resources API gateway (the Lambda handler will remain intact).

## Related Services
All consumers will need to be changed to invoke the resources-api Lambda function directly:

* `remediation-api`
* AppSync
* Cloud Security backend

PR #1880, opened 2020-10-28
Remove resources-api gateway

austinbyers:
## Background

Removes the `resources-api` gateway, just like #1867 removed the `compliance-api` gateway. This removes several thousand lines of code, speeds up deployments, and reduces latency in the web app. 2 down, 2 to go!

Closes: #1688 
Closes: #1813 

## Changes

- Remove `resources-api` gateway: swagger models were ported directly to Go models (no GraphQL changes required)
    - Removed the `orgOverview` endpoint, which isn't used by anyone (it was originally designed to be used for the cloud security overview, but today this is read from the `compliance-api` because they need pass/fail status)
    - Fixed flaky integration test by not caring about the exact compliance status
- Snapshot poller changes (FYI @nhakmiller)
    - Replace `strfmt.DateTime` with `time.Time` - they were the same underlying type; but we want to remove the dependency on the `go-openapi` (swagger) package
    - Restored top-level unit test with proper `resources-api` mocking

## Testing

- Deployed, browsed all cloud security pages to make sure resources are still shown correctly
- All integration tests


internal/compliance/alert_processor/processor/processor_test.go
```diff
@@ -48,31 +48,11 @@ func (m *mockRoundTripper) RoundTrip(request *http.Request) (*http.Response, err
 	return args.Get(0).(*http.Response), args.Error(1)
 }
 
-type mockCompliance struct {
-	gatewayapi.API
-	mock.Mock
-}
-
-func (m *mockCompliance) Invoke(input, output interface{}) (int, error) {
```
austinbyers, line 56: Moved to `pkg/gatewayapi` so it can be shared by multiple unit tests

internal/compliance/aws_event_processor/processor/clients.go
```diff
@@ -22,21 +22,16 @@ import (
...
-
-	transportConfig = client.DefaultTransportConfig().
-			WithBasePath("/" + os.Getenv("RESOURCES_API_PATH")).
-			WithHost(os.Getenv("RESOURCES_API_FQDN"))
-	apiClient  = client.NewHTTPClientWithConfig(nil, transportConfig)
-	httpClient = gatewayapi.GatewayClient(awsSession)
+	awsSession                      = session.Must(session.NewSession())
+	sqsClient       sqsiface.SQSAPI = sqs.New(awsSession)
+	queueURL                        = os.Getenv("SNAPSHOT_QUEUE_URL")
+	resourcesClient gatewayapi.API  = gatewayapi.NewClient(lambda.New(awsSession), "panther-resources-api")
```
kostaspap, line 36: Aren't we removing the API Gateway for resources-api as part of this PR? Why do we need a `gatewayapi.NewClient` to invoke it and not just invoke the Lambda directly?

internal/compliance/remediation_api/remediation/invoke.go
```diff
@@ -55,10 +51,7 @@ var (
 			WithHost(policiesServiceHostname)
 	policiesClient = analysisclient.NewHTTPClientWithConfig(nil, policiesConfig)
 
-	resourcesConfig = resourcesclient.DefaultTransportConfig().
-			WithBasePath(resourcesServicePath).
-			WithHost(resourcesServiceHostname)
-	resourcesClient = resourcesclient.NewHTTPClientWithConfig(nil, resourcesConfig)
+	resourcesClient gatewayapi.API = gatewayapi.NewClient(lambda.New(awsSession), "panther-resources-api")
```
kostaspap, line 54: Note (could be done in a different PR): It would be good to expose the `panther-resources-api` as a constant in the `resourcemodels`  package to avoid typing the same string in many places, e.g. e.g. `resourcemodels.LAMBDA_NAME` or something.

internal/compliance/resources_api/handlers/dynamo.go
```diff
@@ -62,28 +63,28 @@ func (r *resourceItem) Resource(status models.ComplianceStatus) *models.Resource
...
 	condition := existsCondition(resourceID)
 	expr, err := expression.NewBuilder().WithCondition(condition).WithUpdate(update).Build()
 	if err != nil {
 		zap.L().Error("expr.Build failed", zap.Error(err))
 		return &events.APIGatewayProxyResponse{StatusCode: http.StatusInternalServerError}
 	}
 
-	zap.L().Info("submitting dynamo item update",
-		zap.String("resourceId", string(resourceID)))
+	zap.L().Debug("submitting dynamo item update",
```
kostaspap, line 86: We had discussed about high CW costs.... seems that one contributed to it!

internal/compliance/resources_api/handlers/dynamo.go
```diff
@@ -20,36 +20,37 @@ package handlers
...
-	IntegrationType models.IntegrationType `json:"integrationType"`
-	LastModified    models.LastModified    `json:"lastModified"`
-	Type            models.ResourceType    `json:"type"`
+	Attributes      interface{} `json:"attributes"`
+	Deleted         bool        `json:"deleted"`
+	ID              string      `json:"id"`
+	IntegrationID   string      `json:"integrationId"`
+	IntegrationType string      `json:"integrationType"`
+	LastModified    time.Time   `json:"lastModified"`
+	Type            string      `json:"type"`
```
kostaspap, line 44: Is it possible that `Attribute` is ever nil? I think `dynamodbattribute.Marshal` could be unpredictable with this one if nil value is passed to it, I think sometimes it makes a wrong assumption and can mark it as Boolean. Subsequent updates will failing because they were trying to store an attribute of type String on an attribute that it thought was Boolean (it has happened to me in the past)
Maybe worth adding a `dynamodbav:",string"` tag here, unless you have verified that's not the case.

internal/compliance/resources_api/handlers/get_resource.go
```diff
@@ -28,19 +27,20 @@ import (
...
 
-	"github.com/panther-labs/panther/api/gateway/resources/models"
+	"github.com/panther-labs/panther/api/lambda/resources/models"
 	"github.com/panther-labs/panther/pkg/gatewayapi"
 )
 
 // GetResource retrieves a single resource from the Dynamo table.
-func GetResource(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyResponse {
-	resourceID, err := parseGetResource(request)
+func (API) GetResource(input *models.GetResourceInput) *events.APIGatewayProxyResponse {
```
kostaspap, line 35: I think I am missing something. Since we are not using GatewayAPI anymore, why not switch to proper return type?

internal/compliance/resources_api/handlers/list_resources.go
```diff
@@ -57,143 +46,59 @@ var (
...
-	// ***** Sorting *****
-	if sortBy := request.QueryStringParameters["sortBy"]; sortBy != "" {
-		if _, ok := validSortFields[sortBy]; !ok {
-			return nil, errors.New("invalid sortBy: " + sortBy)
-		}
-		result.SortBy = aws.String(sortBy)
+// URL decode parameters and set defaults
+func parseListResources(input *models.ListResourcesInput) error {
+	var err error
+	input.IDContains, err = url.QueryUnescape(input.IDContains)
```
kostaspap, line 81: Do you still have to unescape it?

internal/compliance/resources_api/handlers/list_resources.go
```diff
@@ -284,7 +190,7 @@ func listFilteredResources(scanInput *dynamodb.ScanInput, params *operations.Lis
 		}
 
 		// Compliance status isn't stored in this table, so we filter it out here if needed
-		if params.ComplianceStatus != nil && *params.ComplianceStatus != string(status.Status) {
+		if input.ComplianceStatus != "" && input.ComplianceStatus != status.Status {
```
kostaspap, line 193: If `status.Status` cannot be empty then I guess it would be enough for the check to be `input.ComplianceStatus != status.Status`

If it can be empty, maybe better do something like below? I assume we don't want to append the resource (line 198) if the `status.Status` is empty? 
```
if len(status) == 0{
continue}

if input.ComplianceStatus != status.Status {
return nil
} 

result = append(result, item.Resource(status.Status)
```

internal/compliance/snapshot_poller/models/aws/ecs_cluster.go
```diff
@@ -61,7 +62,7 @@ type EcsService struct {
 	// Fields embedded from ecs.Service
 	CapacityProviderStrategy []*ecs.CapacityProviderStrategyItem
 	// Normalized name for CreatedAt
-	TimeCreated                   *strfmt.DateTime
+	TimeCreated                   *time.Time
```
kostaspap, line 65: 👍

kostaspap (approved):

internal/compliance/aws_event_processor/processor/clients.go
austinbyers: The old client was a wrapper around `http.Client`, this one is a wrapper around the Lambda client, yes. It is indeed invoking the Lambda function directly: https://github.com/panther-labs/panther/blob/master/pkg/gatewayapi/invoke.go

I have a separate helper function for this (instead of `genericapi.Invoke`) because it needs to do an extra step to unpack the `events.APIGatewayProxyResponse` that the `*-api` Lambda functions return.

internal/compliance/remediation_api/remediation/invoke.go
austinbyers: Meh, seems like overkill in my opinion. The constant only exists once per calling function, and once we inline the internal API calls, it will be gone completely

internal/compliance/resources_api/handlers/dynamo.go
austinbyers: Yeah I've been switching a few "info" to "debug" as I've been going through the core apis

internal/compliance/resources_api/handlers/dynamo.go
austinbyers: > Is it possible that Attribute is ever nil?

No, that would mean cloud security submitted a completely empty resource: that should never happen.

We have `validate:"required"` on the Attributes field for AddResourcesInput, but I don't think that validate tag actually works for an `interface{}`. I'll add a test for empty attributes and add extra validation if necessary

internal/compliance/resources_api/handlers/get_resource.go
austinbyers: Because I'm trying to minimize code changes right now so I can remove the internal gateways as fast as possible.

This way, I can leave the `panther-resources-api` response models exactly the same as before. Calling functions (and AppSync) just take a small extra step to unpack the proper model inside the proxy response: that's the `gatewayapi.Client` you were asking about above.

The whole API request/response structure will be overhauled soon, so it doesn't make sense to spend much time fixing it right now.

internal/compliance/resources_api/handlers/list_resources.go
austinbyers: Hmm, good question. The frontend is still escaping it, but you're right that's probably no longer necessary. I'll see if I can remove it from the frontend code

internal/compliance/resources_api/handlers/list_resources.go
austinbyers: > If status.Status cannot be empty then I guess it would be enough for the check to be input.ComplianceStatus != status.Status

Yeah, you're right. Now that we've removed extra pointers and types, the check can be simpler. Good find!

austinbyers: Actually the way it's written now is the simplest. There are two parts: the caller can ask for the compliance status in the _projection_ and in the _filter_. If we get a request like this:

```json
{
    "listResources": {
        "complianceStatus": "",
        "fields": "id,attributes,complianceStatus",
    }
}
```

Then we need to fetch the compliance status (from a separate database), but we do not want to filter on the status. `input.ComplianceStatus` will be an empty string, so we can't just say `input.ComplianceStatus != status.Status`; that condition would never be true.

----

I flipped the logic to hopefully make it easier to understand:

```go
		// Filter on the compliance status (if applicable)
		if input.ComplianceStatus == "" || input.ComplianceStatus == status.Status {
			// Resource passed all of the filters - add it to the result set
			result = append(result, item.Resource(status.Status))
		}

		return nil
```

internal/compliance/resources_api/handlers/dynamo.go
austinbyers: It turns out we weren't validating this correctly: we have to add `validate:"dive"` to descend into a slice of entries! I've used that for a string slice, but I didn't realize it would refuse to descend into a slice without the dive tag.

So I fixed it and added an integration test which confirms that empty attributes are rejected by the API

internal/compliance/resources_api/handlers/list_resources.go
austinbyers: I removed the URL escaping from both the frontend and the backend

austinbyers:
I tested all of the filter/sort options in `listResources` and fixed a few issues I missed the first time around:

* the resource type was not being set correctly (forgot the `json` struct tag)
* filtering by resource type wasn't working because the frontend was sending a CSV string instead of an actual array (I fixed the GraphQL schema and it worked)
* Added `dive` for nested validation

Redeployed and re-ran all integration tests; looks good

internal/compliance/resources_api/handlers/dynamo.go
nhakmiller: We marshal a struct into an interface and then send that along as `Attributes`, so I don't think it could ever be `nil`. At least, it would be a bug in the snapshot poller if it were.

internal/compliance/snapshot_poller/pollers/aws/awstest/utils.go
```diff
@@ -23,13 +23,12 @@ import (
...
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
-	"github.com/go-openapi/strfmt"
 )
 
 // Example output from the AWS API
 var (
 	ExampleTimeParsed, _       = time.Parse(time.RFC3339, "2019-04-02T17:16:30+00:00")
-	ExampleTime                = strfmt.DateTime(ExampleTimeParsed)
+	ExampleTime                = ExampleTimeParsed
```
nhakmiller, line 31: Might have just been able to factor this one out :P

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -64,9 +64,9 @@ const (
 
 var (
 	snapshotPollerSession *session.Session
-	// assumeRoleFunc is the function to return valid AWS credentials.
-	assumeRoleFunc         = assumeRole
-	verifyAssumedCredsFunc = verifyAssumedCreds
+	// AssumeRoleFunc is the function to return valid AWS credentials.
+	AssumeRoleFunc         = assumeRole
```
nhakmiller, line 68: Did we need to export this variable for something? It seems generally specific to the snapshot-poller, and if we wanted to use it elsewhere I would expect it to be moved up to e.g. `pkg` or something.

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -40,12 +41,14 @@ type resourcePoller struct {
...
 
+const integrationType = "aws"
+
 var (
 	// Default region to use when building clients for the individual resource poller
 	// defaultRegion = endpoints.UsWest2RegionID
 	defaultRegion = os.Getenv("AWS_REGION")
 
-	auditRoleName = os.Getenv("AUDIT_ROLE_NAME")
+	AuditRoleName = os.Getenv("AUDIT_ROLE_NAME")
```
nhakmiller, line 51: Again questioning the need to export this, is it used by any other packages?

nhakmiller (commented):
Sorry for the late review! Overall seems good, even though you're screwing around with my baby :( a couple minor comments, if you agree I can do a follow up PR at some point.

internal/compliance/resources_api/handlers/dynamo.go
kostaspap: yeah, good to have the validation at some point. A bug could appear in the snapshot poller at some point, and we wouldn't be able to recover from it - not without deleting an item in DDB. Thanks @austinbyers

PR #1884, opened 2020-10-29
Build alarms with explicit name instead of reference and other small CFN fixes

austinbyers:
## Background

Every resource dependency (`!Ref`, `!Sub`, or `!GetAtt`) slows down both Panther deployment and teardown, because CloudFormation has to wait for the first resource to completely finish before moving on to the next.

In our 3 main stacks (core, cloud-security, log-analysis), I noticed we have many references that are really just sharing names:

- Alarm groups (custom resources) are passed the function/queue/table name
- IAM policies for Lambda functions
- Environment variables for Lambda functions

In all of these cases, the reference can be replaced with the explicit resource name; it doesn't matter what order they are created in. You can create alarms, policies, and environment variables even if the associated resource doesn't exist yet.

----

Here's an example of flattening the dependency tree by removing unnecessary dependencies. Today, we have:

- LogProcessorDLQ
   - LogProcessorDLQAlarms
   - LogProcessorQueue
      - LogProcessorQueueAlarms
      - LogProcessorFunction
        - LogProcessorAlarms

But all of the alarms (and the function's IAM role) can be created immediately because we know what the queue and function names will be:

- LogProcessorDLQ
  - LogProcessorQueue
    - LogProcessorFunction
- LogProcessorDLQAlarms
- LogProcessorQueueAlarms
- LogProcessorAlarms

(If the queue weren't an event source for the function, the function itself could also be created at the beginning.) Besides removing one level of the dependency tree, it pushes most resources to the beginning, which allows CFN to parallelize everything more efficiently.

---

What will bother developers is the duplication of the resource names - the same Dynamo table name, for example, now has to be copied to everywhere else in the template that references that table. Personally, I don't think that's a big deal:

- We already have to list the resource name explicitly for "cross-stack" references
- Typos are caught very easily during deployment (functions will fail fast if their permissions/environment variables are not correct)
- Typing `!Ref MyQueue` vs `panther-my-queue` is about the same amount of work and both options are susceptible to copy-pasta bugs
- If the duplication really bugs everyone, we can use Mappings to copy the same value everywhere, like we already do with Timeout and Memory.

## Changes

- Remove all resource references that aren't necessary for dependency ordering
- Simplified some IAM policies, combining related policy statements and removing redundant SQS permissions when a function is invoked from a queue (the Serverless transform will add those permissions automatically)

## Testing

- Deployed to `us-east-1`, app works fine, no CW alarms
- Deploy time is slightly improved for all 3 stacks:

Deploy Time

- Teardown time is about the same:
Teardown Time

These were tested in us-east-1 as fresh deploy/teardowns. Certainly not scientifically rigorous and subject to AWS "weather," but strongly indicative of an overall improvement in deploy time


deployments/cloud_security.yml
```diff
@@ -148,7 +148,7 @@ Resources:
     Properties:
       AlarmTopicArn: !Ref AlarmTopicArn
       CustomResourceVersion: !Ref CustomResourceVersion
-      QueueName: !GetAtt AlertProcessorQueue.QueueName
+      QueueName: panther-alert-processor-queue
```
kostaspap, line 151: Interesting. So CW will not fail if you try to setup alarms for a queue that doesn't exist? I assume you have check the custom-resources Lambda for errors during deployment?

deployments/cloud_security.yml
```diff
@@ -1367,7 +1331,7 @@ Resources:
           DEBUG: !Ref Debug
           RESOURCES_API_FQDN: !Sub '${ResourcesApiId}.execute-api.${AWS::Region}.${AWS::URLSuffix}'
           RESOURCES_API_PATH: v1
-          SNAPSHOT_QUEUE_URL: !Ref SnapshotQueue
+          SNAPSHOT_QUEUE_URL: !Ref SnapshotQueue # queue is an event source; needs to exist first
```
kostaspap, line 1334: I see that `PollerFunction` has a dependency on `SnapshotQueue` declared in line 1339. Doesn't this mean you could also construct the ARN here as you do in other cases?

kostaspap (approved):
LGTM

deployments/cloud_security.yml
```diff
@@ -1249,7 +1213,7 @@ Resources:
           Statement:
             - Effect: Allow
               Action: dynamodb:GetItem
-              Resource: !GetAtt ResourcesTable.Arn
+              Resource: !Sub arn:${AWS::Partition}:dynamodb:${AWS::Region}:${AWS::AccountId}:table/panther-resources
         - Id: KeyValueStore
```
nhakmiller, line 1217: It's not that big of a deal, but for the rules engine you condensed these four separate inlined policies (resource lookup, KeyValueStore, GetPantherPolicySecrets, GetPantherAnalysisSecrets) into a single inline policy with three statements. For consistency's sake you could do that here as well.

nhakmiller (approved):

deployments/cloud_security.yml
austinbyers: > So CW will not fail if you try to setup alarms for a queue that doesn't exist?

Correct - CloudWatch alarms are just thresholds on _metrics_, they don't care if the associated infrastructure exists yet.

 > I assume you have checked the custom-resources Lambda for errors during deployment?

I checked for CW alarms, and there were none, but I should also search the logs as well, good idea. I'll deploy again and check logs to make sure there weren't any warnings or errors I might have missed

deployments/cloud_security.yml
austinbyers: Yeah, here it doesn't matter whether we do the `!Ref` or the explicit construction because there is already a dependency in the events. I'll go ahead and switch it to be explicit just to be consistent - that way we only use `!Ref` when we really need it

deployments/cloud_security.yml
austinbyers: Sure! Good idea. I might do another pass of all the IAM policies in general, we have a bad habit of making a new inline policy for every individual action

austinbyers:
I'm running out of time today; since this could introduce subtle bugs and I'll be out next week, I'm going to wait to finish this until I get back

austinbyers:
Holding off on merging until I finish removing gateways

austinbyers:
This is once again ready for review. I've updated the branch (wow, it's almost been a month already), addressed the feedback, and am doing a fresh deploy to make sure there are no problems.

I'll make sure each page of the web app still works, integration tests pass, and no CloudWatch alarms were triggered during deployment

deployments/cloud_security.yml
austinbyers: Done

deployments/cloud_security.yml
austinbyers: I updated all the environment variables to construct the resource explicitly, even if the same resource has to be referenced as an event source.

Besides being consistent, the event source may actually become a separate resource, in which case this will help the Lambda functions deploy faster

kostaspap (approved):

deployments/cloud_security.yml
```diff
@@ -191,7 +191,7 @@ Resources:
       Environment:
         Variables:
           DEBUG: !Ref Debug
-          TABLE_NAME: !Ref AlertForwarderTable
+          TABLE_NAME: panther-alert-forwarder
```
rleighton, line 194: I assume this breaks the dependency on the ddb table. If the ddb table schema changes and we do an update to a running system, can't this break things?

deployments/cloud_security.yml
```diff
@@ -288,7 +282,7 @@ Resources:
       AlarmTopicArn: !Ref AlarmTopicArn
       CustomResourceVersion: !Ref CustomResourceVersion
       ServiceToken: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-cfn-custom-resources
-      TableName: !Ref AlertForwarderTable
+      TableName: panther-alert-forwarder
```
rleighton, line 285: Not sure about losing dependencies on ddb tables ... see previous comment.

deployments/cloud_security.yml
```diff
@@ -561,7 +539,7 @@ Resources:
       Description: Compliance API
       Environment:
         Variables:
-          COMPLIANCE_TABLE: !Ref ComplianceTable
+          COMPLIANCE_TABLE: panther-compliance
```
rleighton, line 542: Not sure about losing dependencies on ddb tables ... see previous comment.

deployments/cloud_security.yml
```diff
@@ -671,8 +647,8 @@ Resources:
       Environment:
         Variables:
           DEBUG: !Ref Debug
-          SQS_QUEUE_URL: !Ref RemediationQueue
-          REMEDIATION_LAMBDA_ARN: !GetAtt RemediationFunction.Arn
+          SQS_QUEUE_URL: !Sub https://sqs.${AWS::Region}.${AWS::URLSuffix}/${AWS::AccountId}/panther-remediation-queue
```
rleighton, line 650: Could this introduce a race during deployment?

deployments/cloud_security.yml
```diff
@@ -699,14 +675,14 @@ Resources:
               Action:
                 - sqs:SendMessage
                 - sqs:SendMessageBatch
-              Resource: !GetAtt RemediationQueue.Arn
+              Resource: !Sub arn:${AWS::Partition}:sqs:${AWS::Region}:${AWS::AccountId}:panther-remediation-queue
```
rleighton, line 678: Another race possible? Lambda executes and no q?

deployments/cloud_security.yml
```diff
@@ -927,8 +897,8 @@ Resources:
       Environment:
         Variables:
           DEBUG: !Ref Debug
-          RESOURCES_QUEUE_URL: !Ref ResourcesQueue
-          RESOURCES_TABLE: !Ref ResourcesTable
+          RESOURCES_QUEUE_URL: !Sub https://sqs.${AWS::Region}.${AWS::URLSuffix}/${AWS::AccountId}/panther-resources-queue
+          RESOURCES_TABLE: panther-resources
       FunctionName: panther-resources-api
```
rleighton, line 902: Not sure about losing dependencies on ddb tables ... see previous comment.

deployments/cloud_security.yml
```diff
@@ -927,8 +897,8 @@ Resources:
       Environment:
         Variables:
           DEBUG: !Ref Debug
-          RESOURCES_QUEUE_URL: !Ref ResourcesQueue
-          RESOURCES_TABLE: !Ref ResourcesTable
+          RESOURCES_QUEUE_URL: !Sub https://sqs.${AWS::Region}.${AWS::URLSuffix}/${AWS::AccountId}/panther-resources-queue
```
rleighton, line 900: Possible execution race?

deployments/cloud_security.yml
```diff
@@ -1121,21 +1091,15 @@ Resources:
...
       Runtime: go1.x
       Timeout: !FindInMap [Functions, ResourceProcessor, Timeout]
-      Tracing: !If [TracingEnabled, !Ref TracingMode, !Ref 'AWS::NoValue']
+      Tracing: !If [TracingEnabled, !Ref TracingMode, !Ref AWS::NoValue]
       Policies:
         - Id: SendAndReceiveSqsMessages
           Version: 2012-10-17
           Statement:
-            - Effect: Allow
-              Action:
```
rleighton, line 1134: No longer used?

deployments/cloud_security.yml
```diff
@@ -1308,7 +1262,7 @@ Resources:
         Variables:
           AUDIT_ROLE_NAME: !Sub PantherAuditRole-${AWS::Region}
           DEBUG: !Ref Debug
-          SNAPSHOT_QUEUE_URL: !Ref SnapshotQueue
+          SNAPSHOT_QUEUE_URL: !Sub https://sqs.${AWS::Region}.${AWS::URLSuffix}/${AWS::AccountId}/panther-snapshot-queue
```
rleighton, line 1265: Race on initial deploy possible?

deployments/cloud_security.yml
```diff
@@ -1398,7 +1352,7 @@ Resources:
       Environment:
         Variables:
           DEBUG: !Ref Debug
-          SNAPSHOT_POLLERS_QUEUE_URL: !Ref SnapshotQueue
+          SNAPSHOT_POLLERS_QUEUE_URL: !Sub https://sqs.${AWS::Region}.${AWS::URLSuffix}/${AWS::AccountId}/panther-snapshot-queue
```
rleighton, line 1355: Race on deploy possible?

deployments/core.yml
```diff
@@ -264,17 +264,17 @@ Resources:
       Environment:
         Variables:
           DEBUG: !Ref Debug
-          ORG_TABLE_NAME: !Ref OrganizationTable
+          ORG_TABLE_NAME: panther-organization
```
rleighton, line 267: Not sure about losing dependencies on ddb tables ... see previous comment.

deployments/core.yml
```diff
@@ -319,7 +319,7 @@ Resources:
           POLICY_ENGINE: panther-policy-engine
           RULES_ENGINE: panther-rules-engine
           RESOURCE_QUEUE_URL: !Sub https://sqs.${AWS::Region}.${AWS::URLSuffix}/${AWS::AccountId}/panther-resources-queue
-          TABLE: !Ref AnalysisTable
+          TABLE: panther-analysis
```
rleighton, line 322: Not sure about losing dependencies on ddb tables ... see previous comment.

deployments/core.yml
```diff
@@ -489,7 +485,7 @@ Resources:
         Variables:
           DEBUG: !Ref Debug
           KEY_ID: !Ref OutputsKeyId
-          OUTPUTS_TABLE_NAME: !Ref OutputsTable
+          OUTPUTS_TABLE_NAME: panther-outputs
```
rleighton, line 488: Not sure about losing dependencies on ddb tables ... see previous comment.

deployments/core.yml
```diff
@@ -609,7 +605,7 @@ Resources:
       Environment:
         Variables:
           DEBUG: !Ref Debug
-          ALERT_QUEUE_URL: !Ref AlertQueue
+          ALERT_QUEUE_URL: !Sub https://sqs.${AWS::Region}.${AWS::URLSuffix}/${AWS::AccountId}/panther-alerts-queue
```
rleighton, line 608: Race on deploy?

deployments/core.yml
```diff
@@ -759,7 +746,7 @@ Resources:
           LOG_PROCESSOR_QUEUE_URL: !Sub https://sqs.${AWS::Region}.amazonaws.com/${AWS::AccountId}/panther-input-data-notifications-queue
           LOG_PROCESSOR_QUEUE_ARN: !Sub arn:${AWS::Partition}:sqs:${AWS::Region}:${AWS::AccountId}:panther-input-data-notifications-queue
           SNAPSHOT_POLLERS_QUEUE_URL: !Sub https://sqs.${AWS::Region}.amazonaws.com/${AWS::AccountId}/panther-snapshot-queue
-          TABLE_NAME: !Ref IntegrationsTable
+          TABLE_NAME: panther-source-integrations
```
rleighton, line 749: Not sure about losing dependencies on ddb tables ... see previous comment.

deployments/core.yml
```diff
@@ -1233,17 +1202,17 @@ Resources:
       Environment:
         Variables:
           DEBUG: !Ref Debug
-          LOG_TYPES_TABLE_NAME: !Ref LogTypesTable
+          LOG_TYPES_TABLE_NAME: panther-logtypes
```
rleighton, line 1205: Not sure about losing dependencies on ddb tables ... see previous comment.

deployments/log_analysis.yml
```diff
@@ -184,7 +183,7 @@ Resources:
       Environment:
         Variables:
           DEBUG: !Ref Debug
-          ALERTS_TABLE_NAME: !Ref LogAlertsTable
+          ALERTS_TABLE_NAME: panther-log-alert-info
```
rleighton, line 186: Not sure about losing dependencies on ddb tables ... see previous comment.

deployments/log_analysis.yml
```diff
@@ -312,7 +310,7 @@ Resources:
       Environment:
         Variables:
           DEBUG: !Ref Debug
-          ALERTS_TABLE: !Ref LogAlertsTable
+          ALERTS_TABLE: panther-log-alert-info
```
rleighton, line 313: Not sure about losing dependencies on ddb tables ... see previous comment.

deployments/log_analysis.yml
```diff
@@ -667,23 +658,23 @@ Resources:
...
+      Layers: !If [AttachLayers, !Ref LayerVersionArns, !Ref AWS::NoValue]
       MemorySize: !FindInMap [Functions, Updater, Memory]
       Runtime: go1.x
       Timeout: !FindInMap [Functions, Updater, Timeout]
       Environment:
         Variables:
           ATHENA_WORKGROUP: !Ref AthenaWorkGroup
           DEBUG: !Ref Debug
-          QUEUE_URL: !Ref UpdaterQueue
+          QUEUE_URL: !Sub https://sqs.${AWS::Region}.${AWS::URLSuffix}/${AWS::AccountId}/panther-datacatalog-updater-queue
```
rleighton, line 669: Race on initial deploy?

deployments/log_analysis.yml
```diff
@@ -908,7 +892,7 @@ Resources:
           ALERTS_QUEUE: panther-log-alert-notifications-queue
           S3_BUCKET: !Ref ProcessedDataBucket
           NOTIFICATIONS_TOPIC: !Ref ProcessedDataTopicArn
-          ALERTS_DEDUP_TABLE: !Ref AlertsDedup
+          ALERTS_DEDUP_TABLE: panther-log-alert-dedup
```
rleighton, line 895: Not sure about losing dependencies on ddb tables ... see previous comment.

deployments/log_analysis.yml
```diff
@@ -1106,15 +1068,15 @@ Resources:
...
+      Layers: !If [AttachLayers, !Ref LayerVersionArns, !Ref AWS::NoValue]
       MemorySize: !FindInMap [Functions, MessageForwarder, Memory]
       Runtime: go1.x
       Timeout: !FindInMap [Functions, MessageForwarder, Timeout]
       Environment:
         Variables:
           DEBUG: !Ref Debug
-          STREAM_NAME: !Ref MessageForwarderFirehose
-      Tracing: !If [TracingEnabled, !Ref TracingMode, !Ref 'AWS::NoValue']
+          STREAM_NAME: panther-forwarder-firehose
```
rleighton, line 1078: Race on initial deploy?

rleighton (changes requested):
Nice clean up! 

I have concerns about losing dependencies of the lambda functions on tables and queues. Couldn't there be cases when a function runs soon after an initial  deploy but a table/queue does not exist? This depends on each function but we did not need to worry about it when we had true dependencies.

Updates of ddb table schema/indexes are another concern when deploying over a running system. Now a lambda could read a ddb table but perhaps it has not been changed to the new version of the schema with an expected added column or index. If we kept dependencies then it is not a concern.

Thoughts?

deployments/cloud_security.yml
nhakmiller: Dynamo tables are schema-less, if a lambda refers to new/old columns dynamo will just return nothing if it has nothing there is my understanding. When we start using new columns, we generally try to make it backwards compatible by treating nothing being there as a default situation, but not always possible.

nhakmiller (approved):

deployments/cloud_security.yml
austinbyers: Right, changes to the *schema* would actually require destroying the table, and so we never do schema changes. Adding new columns is managed only in the Lambda functions, not in the Dynamo resource (that's what it means to be noSQL)

However, after chatting offline with @rleighton, he clarified: if we add a new Dynamo index, and update the Lambda functions to use that index, the Lambda update could happen first, and the function would start failing trying to read an index which doesn't exist.

I hadn't thought of that race condition, and there could be others like it. I'm going to go ahead and revert the references in the Lambda functions, but I'll keep the explicit naming in the alarm custom resources. Hopefully, that gives a bit of the extra speed without introducing any new race conditions (the alarm ordering really truly does not matter)

deployments/cloud_security.yml
austinbyers: Ah, no - when you have an event source, the Serverless transform automatically adds the necessary permissions to use it.

Since SQS is feeding this function, we don't need a separate policy to allow reading SQS messages - one will be added automatically

deployments/cloud_security.yml
austinbyers: This one is just used for the Dynamo alarm group, no race condition

deployments/cloud_security.yml
austinbyers: Will revert

deployments/cloud_security.yml
austinbyers: Not likely, but reverting change

deployments/cloud_security.yml
austinbyers: Reverted

deployments/cloud_security.yml
austinbyers: Reverted

deployments/cloud_security.yml
austinbyers: Reverting

deployments/cloud_security.yml
austinbyers: No, because the queue is still referenced in the event-source mapping. Reverting nontheless

deployments/cloud_security.yml
austinbyers: Reverting

deployments/core.yml
austinbyers: Reverting

deployments/core.yml
austinbyers: Reverting

deployments/core.yml
austinbyers: Reverting

deployments/core.yml
austinbyers: Reverting

deployments/core.yml
austinbyers: Reverting

deployments/core.yml
austinbyers: Reverting

austinbyers:
This wasn't a high-priority PR, more of an experiment I played with late one night :) Reverting changes to the Lambda functions, no need to introduce new race conditions just for the sake of slightly faster deploys

But we have hundreds of alarms and those are still building with the explicit resource name. That change alone should be enough to shave a few seconds off of each stack

rleighton (approved):

PR #1901, opened 2020-10-30
Validate S3 bucket (and prefix) are unique among S3 log sources

### LINKED ISSUE #1646: Force distinct bucket/prefix for each S3 source
### Description

Force distinct bucket/prefix for each S3 source

### Related Services

Sources-api

### Designs

Not needed

### Acceptance Criteria

- Putting a new S3 source fails if it has the same bucket and S3 prefix as an existing source

giorgosp:
## Changes

API prevents creating/updating an s3 integration that doesn't have distinct s3 bucket (and prefix, not user-editable though).

Fixes #1646.

## Testing

Manual QA 

Screenshot 2020-10-30 at 2 32 27 AM

### DIFF
--- a/internal/core/source_api/api/put_integration.go
+++ b/internal/core/source_api/api/put_integration.go
@@ -169,6 +169,12 @@ func (api API) integrationAlreadyExists(input *models.PutIntegrationInput) error
 							input.IntegrationLabel),
 					}
 				}
+
+				if existingIntegration.S3Bucket == input.S3Bucket && existingIntegration.S3Prefix == input.S3Prefix {
+					return &genericapi.InvalidInputError{
+						Message: "An S3 integration with the same S3 bucket and prefix already exists.",
+					}
+				}
 			case models.IntegrationTypeSqs:
 				if existingIntegration.IntegrationLabel == input.IntegrationLabel {
 					// Sqs sources need to have different labels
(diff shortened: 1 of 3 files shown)


internal/core/source_api/api/put_integration.go
```diff
@@ -169,6 +169,12 @@ func (api API) integrationAlreadyExists(input *models.PutIntegrationInput) error
 							input.IntegrationLabel),
 					}
 				}
+
+				if existingIntegration.S3Bucket == input.S3Bucket && existingIntegration.S3Prefix == input.S3Prefix {
+					return &genericapi.InvalidInputError{
+						Message: "An S3 integration with the same S3 bucket and prefix already exists.",
+					}
```
kostaspap, line 176: What happens if a customer configures two sources for the same bucket and prefixes:
1. `some-prefix`
2. `some-prefix/another-prefix/`

Should we fail? Or otherwise, we want to make sure that the "longer" prefix takes precedence over the shorter one.

giorgosp: What do you mean precedence?

The combination `bucket-prefix` will be unique even if prefix is empty for an integration, so we won't fail.

giorgosp: Tested it on the dev account, created 2 s3 integrations with same bucket and prefixes:
- prefixa
- prefixa/prefixb

Put S3 objects in the prefixes and the source label ends up correct on Athena:

Screenshot 2020-10-30 at 12 44 43 PM

Screenshot 2020-10-30 at 12 45 13 PM

kostaspap (approved):

PR #1902, opened 2020-10-30
Ignoring S3 files from unregistered sources

kostaspap:
## Background

Sometimes log processor will receive events from unregistered sources. This can happen if e.g. someone has configured bucket-wide SNS notifications but has registered only a subfolder in Panther. 

In those cases it doesn't make sense to fail and retry - just ignore the messages. 

## Changes

- If we receive messages from unregistered S3 path, log a warning and continue

## Testing

- Unit test

### DIFF
--- a/internal/log_analysis/log_processor/sources/s3.go
+++ b/internal/log_analysis/log_processor/sources/s3.go
@@ -116,7 +116,9 @@ func handleNotificationMessage(notification *SnsNotification) (result []*common.
 			}
 			return
 		}
-		result = append(result, dataStream)
+		if dataStream != nil {
+			result = append(result, dataStream)
+		}
 	}
 	return result, err
 }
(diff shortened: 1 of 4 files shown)


rleighton:
Excellent idea.

rleighton (approved):
LGTM.

Note: this will still result in getting paged sometimes (when we get a bunch of warns at once). Still better than failing the batch.

s0l0ist:
@kostaspap is this going to be synced to the release branch?

kostaspap:
> @kostaspap is this going to be synced to the release branch?

This is [already](https://github.com/panther-labs/panther/commit/6566d2e43374abd9c5bdb9797b89d01986fac957) in the release branch. This PR was merged to master before the 1.12 branch was created.

PR #1906, opened 2020-10-30
Title dedup backend fixes

giorgosp:
## Background

Discussed with Aggelos in order to improve the API response when testing rules.

##  Changes
- Fix `testSummary` calculation. True if all tests pass
- Add `passed` field for every test result returned
- Add output and error fields for the new alert_context() function
- Rename `matched` field to `ruleOutput` for consistency

## Testing

- unit + manual QA


api/gateway/analysis/api.yml
```diff
@@ -1201,7 +1201,13 @@ definitions:
         type: string
       ruleId:
         type: string
-      matched:
+      passed:
+        type: boolean
+        description: True if the test returned the expected result. Always false if the result includes an error.
```
kostaspap, line 1206: * Shouldn't this still be `matched`? The expected result for a test would be that `matched=false`. 
* ` Always false if the result includes an error.`: What if the `rule` function passes but the `title` fails with an error. What is going to be the value of the `passed` variable?

giorgosp: There is a separate filed for `matched`. Previously named `matched`, refactored it now to `ruleOutput`.

We only really have unit tests for the `rule()` functions, so  `Passed` returns true if the `rule()` output matched the expected result.

api/graphql/schema.graphql
```diff
@@ -30,7 +30,7 @@ type Mutation {
   suppressPolicies(input: SuppressPoliciesInput!): Boolean
   testPolicy(input: TestPolicyInput!): TestPolicyResponse!
   testRule(input: TestRuleInput!): TestRuleResponse!
-  updateAlertStatus(input: UpdateAlertStatusInput!): AlertSummary
+  updateAlertStatus(input: UpdateAlertStatusInput!): [AlertSummary!]!
```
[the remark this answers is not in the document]
giorgosp: I think I messed up conflict resolution here. Will diff with master

api/gateway/analysis/api.yml
kostaspap: Why do we need both `passed` and `errored`?

api/graphql/schema.graphql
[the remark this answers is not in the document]
3nvi: I tried to remove it, but there were some things breaking. If this is present on master, then leave it as is :D

3nvi (approved):
LGTM

api/gateway/analysis/api.yml
```diff
@@ -1201,7 +1201,13 @@ definitions:
...
+      errored:
+        type: boolean
+        description: True if the result includes an error.
```
kostaspap, line 1209: We already have 
- `genericError`
- `ruleError`
- `dedupError`
- `alertContextError`
- `titleError`

Why do we need the `errored` as well?

kostaspap (approved):

api/gateway/analysis/api.yml
giorgosp: Mostly a convenience for the frontend, so that we do `bool(self.rule_exception or self.title_exception or self.dedup_exception or self.alert_context_exception)` in the backend. Comes from the rule engine

api/gateway/analysis/api.yml
giorgosp: Screenshot 2020-11-02 at 12 51 20 PM

PR #1910, opened 2020-10-30
Restore url-decoding for getResource

austinbyers:
## Background

#1880 introduced a bug: the resource details page did not work for any resource with a `/`. That's because I removed url-decoding from the backend per a PR suggestion, and thought I had removed it from the frontend as well. I did test the new change, but i happened to click a resource without a `/` so the page worked fine. I didn't find this bug until I synced to enterprise and did more testing

With API gateway removed, URL-encoding is no longer necessary. But i couldn't figure out how to remove it from the frontend, so i just restored the decoding in the backend for now.

## Changes

- Restore url-decoding for `getResource` for now

## Testing

- Resource list and details pages in a fresh deploy


nhakmiller (approved):
LGTM

PR #1911, opened 2020-10-31
Using context to manage cancellation of rescaling operation

kostaspap:
## Background

Using context to manage cancellation of rescaling operation

## Changes

- Using context to manage cancellation of rescale operation
- Avoid redundant API call

## Testing

- mage test:ci


internal/log_analysis/log_processor/main/lambda_test.go
```diff
@@ -36,14 +36,6 @@ import (
 	"github.com/panther-labs/panther/pkg/testutils"
 )
 
-var (
```
kostaspap, line 39: It is only used in one test

internal/log_analysis/log_processor/processor/rescale.go
```diff
@@ -49,49 +53,67 @@ var (
...
-	return stopScaling
+func queueDepth(ctx context.Context, sqsClient sqsiface.SQSAPI) (int, error) {
+	getQueueAttributesInput := &sqs.GetQueueAttributesInput{
+		AttributeNames: []*string{
+			aws.String(sqs.QueueAttributeNameApproximateNumberOfMessages), // tells us there is waiting events now
+		},
+		QueueUrl: &common.Config.SqsQueueURL,
+	}
+	getQueueAttributesOutput, err := sqsClient.GetQueueAttributesWithContext(ctx, getQueueAttributesInput)
+	if err != nil && err != context.Canceled {
```
kostaspap, line 89: In case the context is cancelled, also cancel AWS API call

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -111,16 +111,6 @@ func streamEvents(
...
 
-			// under no load we do not read from the sqs queue and just exit
-			totalQueuedMessages, err := queueDepth(sqsClient) // this includes queued and delayed messages
-			if err != nil {
-				readEventErrorChan <- err
-				return
-			}
-			if totalQueuedMessages == 0 {
-				break
-			}
```
kostaspap, line 122: The purpose of this method was to check the messages in the SQS queue - if the messages are 0, don't try to receive them. 
However, this logic is now redundant - we can just try to receive messages from the queue and if 0 are returned, stop.

internal/log_analysis/log_processor/processor/stream_test.go
```diff
@@ -274,41 +279,47 @@ func TestScaleup(t *testing.T) {
...
 	}
-
-	streamTestSqsClient.On("GetQueueAttributes", mock.Anything).Return(streamTestSpikeInEventsQueue, nil).Once()
-	// this will pause scaleupTestDuration to give time for the scale up go routine to run a few cycles
-	streamTestSqsClient.On("ReceiveMessage", mock.Anything).Return(streamTestReceiveMessageOutput, nil).Once()
-
+	// this will pause scaleupTestDuration to give time for the scale up go routine to run twice
+	streamTestSqsClient.On("ReceiveMessage", mock.Anything).Return(streamTestReceiveMessageOutput, nil).Once().
+		Run(func(args mock.Arguments) {
+			wg.Wait()
```
kostaspap, line 294: Using WaitGroup to synchronize the operations instead of relying on time.Sleep

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -65,32 +66,31 @@ func StreamEvents(
...
-	var sqsMessageCount int
-
 	streamChan := make(chan *common.DataStream, 2*sqsMaxBatchSize) // use small buffer to pipeline events
 	processingDeadlineTime := processingDeadlineTime(deadlineTime)
 
 	var accumulatedMessageReceipts []*string // accumulate message receipts for delete at the end
 
 	readEventErrorChan := make(chan error, 1) // below go routine closes over this for errors, 1 deep buffer
 	go func() {
+		scalingDecisionCtx, cancel := context.WithCancel(ctx)
```
rleighton, line 88: better as `cancelScaling`

internal/log_analysis/log_processor/main/lambda.go
```diff
@@ -53,6 +53,6 @@ func process(lc *lambdacontext.LambdaContext, deadline time.Time) (err error) {
 	}()
 
 	logTypesResolver := registry.NativeLogTypesResolver()
-	sqsMessageCount, err = processor.StreamEvents(common.SqsClient, common.LambdaClient, logTypesResolver, deadline)
+	sqsMessageCount, err = processor.StreamEvents(context.TODO(), common.SqsClient, common.LambdaClient, logTypesResolver, deadline)
```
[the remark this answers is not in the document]
rleighton: can you pass in `lc` the `*lambdacontext.LambdaContext` and avoid `context.TODO()` ?

rleighton (dismissed):
Looks good. 

I would run `mage deploy` and let run for a bit and keep sure all the metrics look happy.

rleighton:
Can you remove comment on this line:

https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/processor/rescale.go#L67

alxarch (commented):
I will take a closer look on this Monday. Generally using context looks good. My concerns lie with the closing of channels one after the other when cancel is called. My gut tells me there's either a race or a possible memory leak. I would be expecting a wait group there.

internal/log_analysis/log_processor/processor/stream.go
alxarch: The go idiom is to use ctx and cancel overwriting the original ctx value. This also helps to avoid using the wrong context by mistake. Correct context based code should not require a reference to the original ctx ftom there on.

internal/log_analysis/log_processor/main/lambda.go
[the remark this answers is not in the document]
kostaspap: Done

internal/log_analysis/log_processor/processor/stream.go
kostaspap: Changed to `ctx`

kostaspap:
> Can you remove comment on this line:
> 
> https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/processor/rescale.go#L67

removed

kostaspap:
> My gut tells me there's either a race or a possible memory leak. I would be expecting a wait group there.

I know what you mean. I don't think there's such an issue though. The `streamEvents` function doesn't pass any dependencies to `scalingDecisions`. The `scalingDecisions` invocation probably could be moved to a different place in the code - will do a second round of small changes soon.

kostaspap:
> > My gut tells me there's either a race or a possible memory leak. I would be expecting a wait group there.
> 
> I know what you mean. I don't think there's such an issue though. The `streamEvents` function doesn't pass any dependencies to `scalingDecisions`. The `scalingDecisions` invocation probably could be moved to a different place in the code - will do a second round of small changes soon.

It would be a bigger problem if we were passing any of the channels inside that function.

rleighton:
> > > My gut tells me there's either a race or a possible memory leak. I would be expecting a wait group there.
> > 
> > 
> > I know what you mean. I don't think there's such an issue though. The `streamEvents` function doesn't pass any dependencies to `scalingDecisions`. The `scalingDecisions` invocation probably could be moved to a different place in the code - will do a second round of small changes soon.
> 
> It would be a bigger problem if we were passing any of the channels inside that function.

The rescale operation is really just a function we want run in the background and terminated at the end of the lambda event.

Can you move the spawning of it to the main() function?  I assume when the event is over the lambda `ctx` will send a cancel signal that will shut it down, right?  This seems to be the correct place for it.

rleighton:
CircelCI is failing on: 
```
 github.com/panther-labs/panther/internal/log_analysis/log_processor/main [github.com/panther-labs/panther/internal/log_analysis/log_processor/main.test]
internal/log_analysis/log_processor/main/lambda_test.go:64:9: undefined: process
```

kostaspap:
> CircelCI is failing on:
> 
> ```
>  github.com/panther-labs/panther/internal/log_analysis/log_processor/main [github.com/panther-labs/panther/internal/log_analysis/log_processor/main.test]
> internal/log_analysis/log_processor/main/lambda_test.go:64:9: undefined: process
> ```

Fixed

kostaspap:
> > > > My gut tells me there's either a race or a possible memory leak. I would be expecting a wait group there.
> > > 
> > > 
> > > I know what you mean. I don't think there's such an issue though. The `streamEvents` function doesn't pass any dependencies to `scalingDecisions`. The `scalingDecisions` invocation probably could be moved to a different place in the code - will do a second round of small changes soon.
> > 
> > 
> > It would be a bigger problem if we were passing any of the channels inside that function.
> 
> The rescale operation is really just a function we want run in the background and terminated at the end of the lambda event.
> 
> Can you move the spawning of it to the main() function? I assume when the event is over the lambda `ctx` will send a cancel signal that will shut it down, right? This seems to be the correct place for it.

Yeah, agreed. I tried to move it but it requires changing quite some tests....
I will do in a different PR

internal/log_analysis/log_processor/processor/rescale_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/lambda"
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"github.com/stretchr/testify/mock"
+
+	"github.com/panther-labs/panther/pkg/testutils"
+)
+
+func TestScaleup(t *testing.T) {
```
kostaspap, line 36: Moved (and slightly refactored) from `stream_test.go`

kostaspap:
> > > > My gut tells me there's either a race or a possible memory leak. I would be expecting a wait group there.
> > > 
> > > 
> > > I know what you mean. I don't think there's such an issue though. The `streamEvents` function doesn't pass any dependencies to `scalingDecisions`. The `scalingDecisions` invocation probably could be moved to a different place in the code - will do a second round of small changes soon.
> > 
> > 
> > It would be a bigger problem if we were passing any of the channels inside that function.
> 
> The rescale operation is really just a function we want run in the background and terminated at the end of the lambda event.
> 
> Can you move the spawning of it to the main() function? I assume when the event is over the lambda `ctx` will send a cancel signal that will shut it down, right? This seems to be the correct place for it.

Ok, I went ahead and moved it to the `handle` function. I think it cleans up the tests a bit. Let me know what you think

internal/log_analysis/log_processor/processor/rescale.go
giorgosp: if `err !=nil ` but `err != Cancelled`, do we just ignore the error?

internal/log_analysis/log_processor/processor/rescale.go
```diff
@@ -43,55 +44,68 @@ const (
...
 		zap.L().Debug("scaling up", zap.Int("nLambdas", nLambdas))
 		for i := 0; i < nLambdas; i++ {
-			resp, err := lambdaClient.Invoke(&lambda.InvokeInput{
+			resp, err := lambdaClient.InvokeWithContext(ctx, &lambda.InvokeInput{
 				FunctionName:   box.String("panther-log-processor"),
 				Payload:        []byte(`{"tick": true}`),
 				InvocationType: box.String(lambda.InvocationTypeEvent), // don't wait for response
 			})
-			if err != nil {
+			if err != nil && err != context.Canceled {
```
giorgosp, line 108: You may also want to check for `context.DeadlineExceeded` too

internal/log_analysis/log_processor/processor/rescale.go
kostaspap: Yup.

internal/log_analysis/log_processor/main/lambda.go
```diff
@@ -38,21 +38,34 @@ func main() {
...
 	var sqsMessageCount int
-
 	defer func() {
+		cancelScaling()
 		operation.Stop().Log(err, zap.Int("sqsMessageCount", sqsMessageCount))
 	}()
 
 	logTypesResolver := registry.NativeLogTypesResolver()
-	sqsMessageCount, err = processor.StreamEvents(common.SqsClient, common.LambdaClient, logTypesResolver, deadline)
+	deadline, _ := ctx.Deadline() // TODO Remove deadline, rely on ctx for cancellation
```
rleighton, line 57: Since you are passing ctx to `processor.StreamEvents()` you can remove this line and the `deadline` arg to the method and pull it out of `ctx` where it is used. We use this to derive the the time to read from the sqs queue which is set to 1/2 the deadline (to give time to purge buffers).

internal/log_analysis/log_processor/main/lambda_test.go
```diff
@@ -55,16 +44,12 @@ func TestProcessOpLog(t *testing.T) {
 	common.Config.AwsLambdaFunctionMemorySize = 1024
 	logs := mockLogger()
 	functionName := "myfunction"
-	lc := lambdacontext.LambdaContext{
+	lc := &lambdacontext.LambdaContext{
 		InvokedFunctionArn: functionName,
 	}
 
-	sqsMock := &testutils.SqsMock{}
```
rleighton, line 62: Don't you want to leave this and do?
 ```
ctx := lambdacontext.NewContext(context.WithTimeout(context.Background(),  0*time.Millisecond), lc)
```

I think we want to test that rescale ran on exit.

Comment (kostaspap) [reply internal/log_analysis/log_processor/main/lambda_test.go:62]: Previously, the cancellation of the "ScalingDecision" goroutine was occurring when we would stop pulling from SQS queue.

In the new code, we cancel it later, after we have written to S3 and the SQS messages have been deleted. So if e.g. deleting the messages takes long, the ScalingDecisions will still launch a new AWS Lambda (previously we were not doing that)

My concern on letting it run one last time after it has been cancelled is that we don't really know "when" that will be. Queyring SQS could could block for whatever reason (failures and then some retries...) and we might have a runaway Go routine. We could get around that by putting some more synchronization mechanisms in place, but really starting to think if it is that necessary...

File: internal/log_analysis/log_processor/processor/rescale.go (hunk 8 above)
Comment (kostaspap) [reply internal/log_analysis/log_processor/processor/rescale.go:108]: Updated

File: internal/log_analysis/log_processor/main/lambda.go (hunk 9 above)
Comment (kostaspap) [reply internal/log_analysis/log_processor/main/lambda.go:57]: Updated

File: internal/log_analysis/log_processor/main/lambda_test.go (hunk 10 above)
Comment (rleighton) [reply internal/log_analysis/log_processor/main/lambda_test.go:62]: Really does not matter at all AFAIK. I can't see any way it causes any issue.

**Review by rleighton — COMMENTED** [review]
Looks good. Can you just add back the test that proves the goroutine runs in the main test?

**Comment by kostaspap** [comment]
> Looks good. Can you just add back the test that proves the goroutine runs in the main test?

Added!

File: internal/log_analysis/log_processor/main/lambda_test.go [hunk 11]
```diff
@@ -53,26 +47,47 @@ func mockLogger() *observer.ObservedLogs {
...
+	common.SqsClient = sqsMock
+	emptyQueue := &sqs.GetQueueAttributesOutput{
+		Attributes: map[string]*string{
+			sqs.QueueAttributeNameApproximateNumberOfMessages: aws.String("0"),
+		},
+	}
+	// We use the wg to wait until "scaling decisions" has been called
+	var wg sync.WaitGroup
+	wg.Add(1)
+	sqsMock.On("GetQueueAttributesWithContext", mock.Anything, mock.Anything, mock.Anything).Return(emptyQueue, nil).Once().
```
rleighton, line 61: Very sophisticated mocking, love this.

rleighton (approved):
Very nice refactor! We will scale test this with the more efficient CloudTrail implementation and compare to the previous release.

internal/log_analysis/log_processor/processor/rescale.go
```diff
@@ -43,55 +44,68 @@ const (
...
+		return 0, err
+	}
+
+	return numberOfQueuedMessages, err
 }
 
 // processingScaleUp will execute nLambdas to take on more load
-func processingScaleUp(lambdaClient lambdaiface.LambdaAPI, nLambdas int) {
+func processingScaleUp(ctx context.Context, lambdaClient lambdaiface.LambdaAPI, nLambdas int) {
 	if nLambdas > 0 {
```
alxarch, line 97: better clean this with `if nLambdas <= 0 { return }`

internal/log_analysis/log_processor/processor/rescale.go
```diff
@@ -43,55 +44,68 @@ const (
...
-	processingScaleDecisionInterval = defaultProcessingScaleDecisionInterval // var so we can set in tests
-)
-
-// scalingDecisions makes periodic adaptive decisions to scale up based on the sqs queue stats, it returns
+// ScalingDecisions makes periodic adaptive decisions to scale up based on the sqs queue stats, it returns
 // immediately with a boolean stop channel (sending an event to the channel stops execution).
-func scalingDecisions(sqsClient sqsiface.SQSAPI, lambdaClient lambdaiface.LambdaAPI) chan bool {
-	stopScaling := make(chan bool)
-
+func ScalingDecisions(ctx context.Context, sqsClient sqsiface.SQSAPI, lambdaClient lambdaiface.LambdaAPI, decisionInterval time.Duration) {
```
alxarch, line 49: It would be more clean if this was renamed to `RunScalingDecisions` and it was used as `go RunScalingDecisions(ctx, ...)` without wrapping a `go func(){ ... }` into a new function name that spawns code.

alxarch: This way it is obvious at the call site that a goroutine is being spawned. Probably easier for tests as well.

internal/log_analysis/log_processor/main/lambda_test.go
```diff
@@ -53,26 +47,47 @@ func mockLogger() *observer.ObservedLogs {
...
-	lc := lambdacontext.LambdaContext{
+	lc := &lambdacontext.LambdaContext{
 		InvokedFunctionArn: functionName,
 	}
 
-	sqsMock := &testutils.SqsMock{}
-	common.SqsClient = sqsMock
-	// will be called by scalingDecisions() on exit
-	sqsMock.On("GetQueueAttributes", mock.Anything).Return(emptyQueue, nil).Once()
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Minute)) // Pulling deadline somewhere in the future
```
alxarch, line 79: Please make this more readable by using [`context.WithTimeout`](https://golang.org/pkg/context/#WithTimeout)

internal/log_analysis/log_processor/main/lambda.go
```diff
@@ -32,27 +32,40 @@ import (
...
+	lambdalogger.ConfigureGlobal(ctx, nil)
+	return process(ctx, defaultScalingDecisionInterval)
 }
 
-func process(lc *lambdacontext.LambdaContext, deadline time.Time) (err error) {
+func process(ctx context.Context, scalingDecisionInterval time.Duration) (err error) {
+	lc, _ := lambdacontext.FromContext(ctx)
 	operation := common.OpLogManager.Start(lc.InvokedFunctionArn, common.OpLogLambdaServiceDim).WithMemUsed(lambdacontext.MemoryLimitInMB)
 
+	cancelScaling := scheduleScalingDecisions(ctx, scalingDecisionInterval)
```
alxarch, line 52: best to have the context wiring visible here
```
// Create cancellable deadline for Scaling Decisions go routine
scalingCtx, cancelScaling := context.WithCancel(ctx)
// runs in the background, periodically polling the queue to make scaling decisions
go RunScalingDesisions(scalingCtx, scalingDecisionInterval)
....
```

alxarch (approved):
LGTM, some minor cleanups for more readable context wiring

internal/log_analysis/log_processor/processor/rescale.go
kostaspap: Will do

internal/log_analysis/log_processor/processor/rescale.go
kostaspap: Had the same thought, hesitated to do it. Since we agree on this, i'll go ahead.

kostaspap: Done

internal/log_analysis/log_processor/main/lambda_test.go
kostaspap: Done

internal/log_analysis/log_processor/main/lambda.go
kostaspap: Done

PR #1912, opened 2020-11-01
Remove use of DeployedLogTables()

rleighton:
## Background

This PR removes use of `DeployedLogTables()` and consolidates all the management of Glue tables behind the `datacatalog_updater` lambda.

This refactor is needed in Panther Enterprise to support inclusion of custom logs with Parquet and Snowflake integrations.

There was only 1 use of `DeployedLogTables()` in the custom resource used to update schemas during deployment.
This was replaced by a call to get the deployed logTypes from the source api and then calling the `datacatalog_updater`  to update the related tables. All of the code to update the tables and views had already been moved into the `datacatalog_updater` . One change was added to allow an optional `Sync` flag to kick of a sync of the partitions after the tables and views have been updated.

## Changes

- update custom resource that manages glue tables on deployment to send message to `datacatalog_updater`
- added `ListLogTypes` method to the `source_api`
- updated `DeployedLogTypes()` to include the cloudsec database
- fix stack dependencies in mage deploy that were not correct

## Testing

- mage test:ci
- mage deploy and wait for partitions on the self onboarded log types
- mage deploy + add a column to 3 log parsers + mage deploy and then check that the tables and partitions updated
- mage deploy + remove above column add to 3 log parsers + mage deploy and then check that the tables and partitions updated

### DIFF
--- a/internal/core/custom_resources/resources/log_processor_tables.go
+++ b/internal/core/custom_resources/resources/log_processor_tables.go
@@ -27,18 +27,18 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
-	"github.com/panther-labs/panther/internal/log_analysis/athenaviews"
+	"github.com/panther-labs/panther/internal/core/source_api/apifunctions"
 	"github.com/panther-labs/panther/internal/log_analysis/awsglue"
 	"github.com/panther-labs/panther/internal/log_analysis/datacatalog_updater/process"
-	"github.com/panther-labs/panther/internal/log_analysis/gluetables"
 	"github.com/panther-labs/panther/pkg/awsutils"
 )
 
 type UpdateLogProcessorTablesProperties struct {
 	// TablesSignature should change every time the tables change (for CF master.yml this can be the Panther version)
-	TablesSignature     string `validate:"required"`
-	AthenaWorkGroup     string `validate:"required"`
-	ProcessedDataBucket string `validate:"required"`
+	TablesSignature            string `validate:"required"`
+	AthenaWorkGroup            string `validate:"required"`
+	ProcessedDataBucket        string `validate:"required"`
+	DataCatalogUpdaterQueueURL string `validate:"required"`
 }
 
 func customUpdateLogProcessorTables(ctx context.Context, event cfn.Event) (string, map[string]interface{}, error) {
@@ -51,7 +51,7 @@ func customUpdateLogProcessorTables(ctx context.Context, event cfn.Event) (strin
 			zap.L().Error("failed to parse resource properties", zap.Error(err))
 			return physicalResourceID, nil, err
 		}
-		if err := updateLogProcessorTables(ctx, &props); err != nil {
+		if err := updateLogProcessorTables(&props); err != nil {
 			zap.L().Error("failed to update glue tables", zap.Error(err))
 			return physicalResourceID, nil, err
 		}
@@ -73,7 +73,7 @@ func customUpdateLogProcessorTables(ctx context.Context, event cfn.Event) (strin
 	}
 }
 
-func updateLogProcessorTables(ctx context.Context, props *UpdateLogProcessorTablesProperties) error {
+func updateLogProcessorTables(props *UpdateLogProcessorTablesProperties) error {
 	// ensure databases are all there
 	for pantherDatabase, pantherDatabaseDescription := range awsglue.PantherDatabases {
 		zap.L().Info("creating database", zap.String("database", pantherDatabase))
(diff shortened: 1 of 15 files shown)


tools/mage/deploy/deploy.go
```diff
@@ -400,31 +401,42 @@ func deployMainStacks(settings *PantherConfig, outputs map[string]string) error
 		c <- util.TaskResult{Summary: cfnstacks.Dashboard, Err: deployDashboardStack(outputs["SourceBucket"])}
 	}(results)
 
-	// Log analysis
+	// Wait for above stacks to finish.
```
rleighton, line 404: The stack dependencies were actually not correct, we were just lucky not to hit issues in deployment. Now that there is a custom resource in the log processor that calls the sources_api it is apparent.

internal/log_analysis/gluetables/gluetables.go
```diff
@@ -29,38 +29,25 @@ import (
...
-		deployedLogTables = append(deployedLogTables, gm)
-	}
-
-	return deployedLogTables, nil
-}
-
 // DeployedLogTypes scans glue API to filter log types with deployed tables
 func DeployedLogTypes(ctx context.Context, glueClient glueiface.GlueAPI, logTypes []string) ([]string, error) {
-	dbName := awsglue.LogProcessingDatabaseName
+	dbNames := []string{loganalysisGlue.LogProcessingDatabaseName, cloudsecGlue.CloudSecurityDatabase}
```
rleighton, line 40: Include clousdsec db.

internal/core/source_api/api/list_logtypes.go
```diff
@@ -0,0 +1,57 @@
...
+	// collect them all in a set to ensure uniqueness
+	logTypesSet := make(map[string]struct{})
+	for _, integration := range listOutput {
+		switch integration.IntegrationType {
+		// types related to log processing
+		case models.IntegrationTypeAWS3, models.IntegrationTypeSqs:
+			for _, logType := range integration.LogTypes {
+				logTypesSet[logType] = struct{}{}
+			}
+		}
```
kostaspap, line 48: You should invoke this: https://github.com/panther-labs/panther/blob/master/api/lambda/source/models/integration.go#L66

Then no need for the `switch` statement

internal/core/source_api/apifunctions/apifunctions.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// ListLogTypes gets the current set of logTypes in use
+func ListLogTypes(lambdaClient lambdaiface.LambdaAPI) ([]string, error) {
+	var listLogTypesOutput models.ListLogTypesOutput
+	var listLogTypesInput = &models.LambdaInput{
+		ListLogTypes: &models.ListLogTypesInput{},
+	}
+	if err := genericapi.Invoke(lambdaClient, LambdaName, listLogTypesInput, &listLogTypesOutput); err != nil {
+		return nil, fmt.Errorf("error calling source-api to list logTypes: %v", err)
```
kostaspap, line 44: `errors.Wrapf`?

internal/log_analysis/gluetables/gluetables.go
```diff
@@ -29,38 +29,25 @@ import (
 	"github.com/aws/aws-sdk-go/service/glue/glueiface"
 	"github.com/pkg/errors"
 
-	"github.com/panther-labs/panther/internal/log_analysis/awsglue"
+	cloudsecGlue "github.com/panther-labs/panther/internal/compliance/awsglue"
+	loganalysisGlue "github.com/panther-labs/panther/internal/log_analysis/awsglue"
```
kostaspap, line 33: Package names should be lowercase: https://blog.golang.org/package-names
I believe we want to follow this for import aliases.

internal/core/custom_resources/resources/log_processor_tables.go
```diff
@@ -73,7 +73,7 @@ func customUpdateLogProcessorTables(ctx context.Context, event cfn.Event) (strin
 	}
 }
 
-func updateLogProcessorTables(ctx context.Context, props *UpdateLogProcessorTablesProperties) error {
+func updateLogProcessorTables(props *UpdateLogProcessorTablesProperties) error {
```
kostaspap, line 76: I think you can keep
```
updateLogProcessorTables(_ context.Context, props *UpdateLogProcessorTablesProperties) error
```
Sooner or later we are going to switch to using the context.

internal/core/source_api/apifunctions/apifunctions.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/panther-labs/panther/api/lambda/source/models"
+	"github.com/panther-labs/panther/pkg/genericapi"
+)
+
+// Below are function call entry points to the api that allow callers easily use the lambda interface.
+// There are no referenced external packages (other than `models`) to minimize pulling in unneeded code.
+
+const (
+	LambdaName = "panther-source-api"
+)
```
kostaspap, line 35: Move it to `internal/core/source_api/api/api.go`?

kostaspap (commented):
@giorgosp This PR solves this one I believe: https://github.com/panther-labs/panther-enterprise/issues/1190

internal/core/source_api/api/list_logtypes.go
giorgosp: +1, better use `RequiredLogTypes()` because on enterprise it also contains a special case for CrowdStrike.

rleighton: Perfect! This should help also in merging with enterprise to pick of the other types. Changed.

internal/core/source_api/apifunctions/apifunctions.go
rleighton: Done

internal/log_analysis/gluetables/gluetables.go
rleighton: Done

internal/core/custom_resources/resources/log_processor_tables.go
rleighton: Ok, done

internal/core/source_api/apifunctions/apifunctions.go
rleighton: Good idea. I did not originally because I was worried the package reference would force the linker to pull in all the code in the package, but I tested that just down and Go seems to be smart about this. Moved.

rleighton:
> @giorgosp This PR solves this one I believe: [panther-labs/panther-enterprise#1190](https://github.com/panther-labs/panther-enterprise/issues/1190)

It does look like this should fix that. After I merge to enterprise @giorgosp can test it quickly.

kostaspap (approved):

PR #1914, opened 2020-11-02
Fix pass condition when testing a rule

giorgosp:
## Background

Fix pass condition returned by the API when testing a rule.

## Changes

The test is considered failed iff:
- there is a global error in the rule script (import error, syntax error, etc).
- rule() raises an exception.
- rule() return value is different than the expected value user provided.

Otherwise the test is considered passed. 
`title()/dedup()/alert_context()` exceptions do not fail the tests because:
- they could fail if user provides a test case with missing fields in order to assert the rule doesn't match
- it is consistent with how the rules engine performs log analysis, where users expect them not to run if a rule doesn't match

## Testing
- Manual QA


kostaspap (approved):
Backend code looks ok to me. 
In regards to when a test should be considered failed or not, we need to make sure that the Panther UI and the Panther Analysis Tool demonstrate the same functionality, maybe @nhakmiller can chime in

nhakmiller (commented):
I'm trying to understand both your comment & your code, because they seem to say different things. I would expect this:

1. If a test expects True and the rule returns True with no errors, the test passes
1. If a test expects False and the rule returns False with no errors, the test passes
1. If a test expects True and the rule returns True and the title/dedup/context raises an error, the test fails
1. If a test expects False and the rule returns False and the title/dedup/context raises an error, the test passes

In fact in the last case, I would expect that the title/dedup/context don't even run at all. But that is an implementation detail that doesn't matter to the user; it's ok if they run as long as the output is ignored.

Is this what your code is doing? Based on your code, it seems like you don't do what I would expect in case 4. Based on your comment, however, you seem to be saying you agree with point 4 but disagree with point 3. So I'm just a bit confused on which way its being handled.

internal/core/analysis_api/analysis/rule_engine.go
```diff
@@ -91,8 +91,27 @@ func (e *RuleEngine) TestRule(rule *models.TestPolicy) (*models.TestRuleResult,
...
+		// 1. Users can provide unit tests for the rule() function only,
+		// 2. It most intuitive. Consider the following scenario:
+		//	- User creates a test case where the input is missing some fields and expects rule() to return False.
+		//	- title() fails with KeyError or similar because it uses a field that is missing from the test input.
+		// The test should be successful, because rule() returns False as the user expects.
+		// This is also consistent with log analysis, where if rule won't trigger the alert, the other functions
+		// are not run.
+		var passed bool
+		if len(result.GenericError) > 0 || len(result.RuleError) > 0 {
+			passed = false
```
nhakmiller, line 111: In order to be inline with how the rest of the product works, I think this should be:

```
		if len(result.GenericError) > 0 || len(result.RuleError) > 0 || (result.ExpectedResult && (len(result. TitleError) > 0 ||  len(result.DedupError) > 0 ||  len(result. AlertContextError) > 0)) {
			passed = false
```

or something similar. The test should fail if the expected result is True and also the title, dedup, or alert context functions raise an error.

nhakmiller:
Just to wrap up what we were discussing in Slack:

It is expected behavior that the title/dedup/context functions ONLY run when the rule returns True. Since this is expected behavior, we actually create rules that rely on this behavior, where the title/dedup function relies on a field that will only be present if the rule returns True. In these cases, the title/dedup function may return None or raise an error if run on an  event that would return False.

For these reasons, we should only display the results of title/dedup/context for tests where the expected result is True. If the expected result is False, the behavior of the title/dedup/context functions is undefined so we should not display it or customers may think it means something.

giorgosp:
Updated the code according to the comments.
I'm also truncating the output of `alert_context()` in the test results cause it can be huge json. 
Please take another look of the PR and it should be ready to go!

Screenshot 2020-11-02 at 8 35 31 PM

internal/core/analysis_api/analysis/rule_engine.go
```diff
@@ -91,27 +91,53 @@ func (e *RuleEngine) TestRule(rule *models.TestPolicy) (*models.TestRuleResult,
...
+			testResult.Results[i].TitleError = result.TitleError
+			testResult.Results[i].AlertContextOutput = truncate(result.AlertContextOutput) // truncate, can be huge json
+			testResult.Results[i].AlertContextError = result.AlertContextError
 		}
+		testResult.TestSummary = testResult.TestSummary && passed
 	}
 	return testResult, nil
 }
 
+func truncate(s string) string {
```
nhakmiller, line 120: I might give this a more descriptive name, like `truncateAlertContext` or something similar.

nhakmiller (approved):
LGTM

PR #1921, opened 2020-11-03
fix: add opacity as indication for active tabs

vorillaz:
## Background

This is an optional one that shows the tabs in a "pseudo" disabled state when no rule errors or rule matches are in place.

## Changes.

- Fetching rule errors and rule matches upon mounting
- Activate the  tabs that have data.

## Screenshots.
Screenshot 2020-11-03 at 5 04 27 PM
Screenshot 2020-11-03 at 5 04 13 PM

## Testing.

- Added coverage tests.


kostaspap (approved):
I like it!

alexmylonas (approved):
Code looks solid, I am approving!

I disagree with having a tab that's fade out but still clickable but let's not hang to it.

PR #1927, opened 2020-11-03
fix: sort rules by enabled/disabled status

s0l0ist:
## Background

Fixes sorting by `status` which does not exist for rules. Instead, this should be sorting on enabled/disabled status.

## Changes

- Change sorting `status` filter for rules to be `enabled`

## Testing

- Visually


nhakmiller (approved):
lgtm

PR #1929, opened 2020-11-03
Fix credential generation

nhakmiller:
## Background

We previously made some changes to how the snapshot-poller generates clients in order to support scanning opt-in regions in `ap-east-1`. These changes only worked if the Panther account had global endpoints enabled in all regions. This change (which @kostaspap figured out) should make it work even if global endpoints are not enabled in all regions in the Panther account.

## Changes

- Use regional endpoints whenever creating clients

## Testing

- Deployed to dev account to verify
- Tested with scanner simulation script


rleighton:
@kostaspap Make this change in the log processor too? All new regions will be opt-in ...

kostaspap (approved):

kostaspap:
> @kostaspap Make this change in the log processor too? All new regions will be opt-in ...

Yup, here it is https://github.com/panther-labs/panther/pull/1932

PR #1932, opened 2020-11-04
Using regional STS endpoint

kostaspap:
## Background

Using regional STS endpoint. This gives few benefits: 
- Improved latency
- Support opt-in regions

We are doing this fix also for the snapshot poller: https://github.com/panther-labs/panther/pull/1929

## Changes

- Using local STS endpoint when getting STS credentials for accessing data in S3 buckets

## Testing

- mage test:ci

### DIFF
--- a/internal/log_analysis/log_processor/sources/s3_client_cache.go
+++ b/internal/log_analysis/log_processor/sources/s3_client_cache.go
@@ -166,7 +166,7 @@ var (
 	globalSourceCache = &sourceCache{}
 
 	//used to simplify mocking during testing
-	newCredentialsFunc = stscreds.NewCredentials
+	newCredentialsFunc = getAwsCredentials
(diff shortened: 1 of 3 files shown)


internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -167,7 +166,7 @@ var (
 	globalSourceCache = &sourceCache{}
 
 	//used to simplify mocking during testing
-	newCredentialsFunc = stscreds.NewCredentials
+	newCredentialsFunc = getAwsCredentials
```
kostaspap, line 169: The `session.Copy` performed in line 265 tries to fetch local credentials. I had to mock it out as part of the unit tests

giorgosp (approved):

rleighton (approved):

PR #1933, opened 2020-11-04
Improve AWS.CloudTrail parser performance by porting to the new method

alxarch:
## Background

We are porting 'native' go parsers to the new `pantherlog` method. First on the list is `AWS.CloudTrail` due to the performance overhead  it imposes.

## Changes

- Update all CloudTrail related structs to use `pantherlog.X` field types.
- Implement a new indicator extractor based on `jsoniter` to limit the memory overhead
- Update tests to be YAML-based

## Testing

- Unit tests


internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
```diff
@@ -24,61 +24,57 @@ import (
...
 
 // CloudTrailUserIdentity contains details about the type of IAM identity that made the request.
 type CloudTrailUserIdentity struct {
-	Type             *string                   `json:"type,omitempty"`
-	PrincipalID      *string                   `json:"principalId,omitempty"`
-	ARN              *string                   `json:"arn,omitempty"`
-	AccountID        *string                   `json:"accountId,omitempty"`
-	AccessKeyID      *string                   `json:"accessKeyId,omitempty"`
-	Username         *string                   `json:"userName,omitempty"`
+	Type             pantherlog.String         `json:"type,omitempty"`
```
kostaspap, line 69: do we need all these "omitempty" tags?

alxarch: fixed

internal/log_analysis/log_processor/parsers/awslogs/indicators.go
```diff
@@ -86,3 +87,61 @@ func ScanInstanceID(w pantherlog.ValueWriter, input string) {
 		w.WriteValues(pantherlog.FieldAWSInstanceID, input)
 	}
 }
+
+type AWSRawMessage pantherlog.RawMessage
```
giorgosp, line 91: Is this type alias really needed?

internal/log_analysis/log_processor/parsers/awslogs/testdata/cloudtrail_tests.yml
```diff
@@ -0,0 +1,199 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+name: cloud_trail_generate_data_key
+logType: AWS.CloudTrail
+input: |
+  {
```
giorgosp, line 20: This way of testing is clearly better than the previous.

Idea for future:  It would be nice to have editor/IDE intellisense for json for testdata. It will be difficult to debug json syntax issues now that it is embedded as raw string inside yaml.

internal/log_analysis/log_processor/logtypes/entry.go
```diff
@@ -243,12 +243,12 @@ func checkLogEntrySchema(logType string, schema interface{}) error {
 	if schema == nil {
 		return errors.Errorf("nil schema for log type %q", logType)
 	}
-	data, err := jsoniter.Marshal(schema)
+	data, err := pantherlog.ConfigJSON().Marshal(schema)
```
giorgosp, line 246: Would be easier on the eyes if this line would read `pantherlog.Json.Marshal(schema)`.

giorgosp (approved):

internal/log_analysis/log_processor/parsers/awslogs/indicators.go
```diff
@@ -86,3 +87,61 @@ func ScanInstanceID(w pantherlog.ValueWriter, input string) {
...
+	case jsoniter.ArrayValue:
+		for iter.ReadArray() {
+			extractIndicators(w, iter, key)
+		}
+	case jsoniter.StringValue:
+		value := iter.ReadString()
+		switch key {
+		case "arn", "ARN":
+			ScanARN(w, value)
+		case "instanceId", "instance-id":
```
kostaspap, line 127: Seems we are missing some stuff here: 
https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/parsers/awslogs/extractor.go#L69

internal/log_analysis/log_processor/parsers/awslogs/indicators.go
```diff
@@ -86,3 +87,61 @@ func ScanInstanceID(w pantherlog.ValueWriter, input string) {
 		w.WriteValues(pantherlog.FieldAWSInstanceID, input)
 	}
 }
+
+type AWSRawMessage pantherlog.RawMessage
+
+func ExtractRawMessageIndicators(w pantherlog.ValueWriter, messages ...pantherlog.RawMessage) {
```
kostaspap, line 93: I would suggest writing a test for this similar to what we have for extractor.go : https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/parsers/awslogs/extractor_test.go

It makes much easier to test the individual extractors

internal/log_analysis/log_processor/logtypes/entry.go
alxarch: a) It has to be a function and not a variable otherwise external modules can alter the exported variable
b) The naming convention in go is to use all caps for acronyms regardless of their position in a name (i.e. `JSON` instead of `Json`)

internal/log_analysis/log_processor/parsers/awslogs/indicators.go
alxarch: I think that one extractor for all cases is a bad idea performance-wise. These tests are performed for *every* string field in a JSON tree. I think it would be more appropriate to have a more generic version for this extractor that accepts `map[string]pantherlog.Scanner` for exact key match, prefix match, suffix matching. Will add the missing cases but we need to review this.

internal/log_analysis/log_processor/parsers/awslogs/indicators.go
alxarch: Adding

internal/log_analysis/log_processor/logtypes/entry.go
giorgosp: `pantherlog.JSON().Marshal(schema)` is fine too. Nitpicking,  but `ConfigJSON()` is not a very accurate name

internal/log_analysis/log_processor/parsers/awslogs/indicators.go
alxarch: fixed

internal/log_analysis/log_processor/parsers/awslogs/testdata/cloudtrail_tests.yml
alxarch: I don't know if we can leverage JSON Schema for this

internal/log_analysis/log_processor/parsers/awslogs/indicators.go
alxarch: Nope, forgot that in from initial implementation of the extractor. Fixed

internal/log_analysis/log_processor/parsers/awslogs/indicators.go
alxarch: added test and fixed a bug in ARN scanner

internal/log_analysis/log_processor/logtypes/entry.go
alxarch: The jsoniter package uses `ConfigDefault`, `ConfigFastest` etc. Keeping the name for now until this is somewhow streamlined across to enterprise as well with BuildJSON et al

internal/log_analysis/log_processor/parsers/awslogs/indicators.go
kostaspap: 👍  I'm not against making a more performant solution, we just need to make sure there is no regression

PR #1934, opened 2020-11-04
Rename "Dedup Message" to "Dedup String"

giorgosp:
Rename dedup() output label to "Dedup String" (on the rule test results UI)


3nvi (approved):
LGTM

Issue #1690, opened 2020-10-07
Remove analysis-api gateway


austinbyers:
## Description
Remove the analysis API gateway (the Lambda handler will remain intact).

This is our biggest and complex internal API gateway, so I'm saving it for last so I can first learn how to remove the simpler gateways

## Related Services
All consumers will need to be changed to invoke the analysis-api Lambda function directly:

* AppSync
* Cloud security, log progressing, and historical search backends

PR #1945, opened 2020-11-04
Create DataModel AnalysisType for Support User Defined Schemas

### LINKED ISSUE #1768: Adding Schema Accessible in Rules
### Description

Support field mappings / user defined schema in rules engine. 

### Related Services
WIP
- This issue is strictly for the yml part of schemas - making the logtype -> field mapping available to `udm` method in the rules engine.  
- There is a related issue for events to _use_ the schema via `udm` method.  

### Designs
WIP

### Acceptance Criteria
WIP
 
- Users can add schema definitions in yml
- yml definitions are accessible in the rules engine

lindsey-w:
## Background

In order to support user-defined data models, we have to be able to upload and store the data to be accessible in the rules engine.  This change adds logic that will enable uploading new data models and access those data models.  It defines a new `DATAMODEL` analysis type that is stored in the `panther-analysis` ddb.  Closes #1768 

## Changes

- new AnalysisType: DataModel
- new api endpoints for handling datamodels
- control logic for the api endpoints
- integration tests for: create/modify/get/delete DataModel

## Testing

- `mage test:integration`

## sample data model

```
AnalysisType: DataModel
LogTypes: 
  - OneLogin.Events
DataModelID: onelogin.DataModel
Filename: onelogin_datamodel.py
Enabled: true
Mappings:
  - Name: source_ip
    Field: ipAddress
  - Name: dest_ip
    Field: $.events[*].parameters[?(@.name == 'ipaddr')].value
  - Name: event_type
    Method: get_event_type
```

associated (optional) datamodel py script
```
def get_event_type(event):
    if event.get('event_type_id') == 6:
        return 'failed_login'
    return 'login'
```


lindsey-w (commented):
I made this a draft because I just realized I need to add logic to the bulk upload sections as well.

internal/core/analysis_api/handlers/list.go
```diff
@@ -89,6 +89,11 @@ func ListRules(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyRe
 	return handleList(request, typeRule)
 }
 
+// ListDataModels pages through datamodels from a single organization.
+func ListDataModels(request *events.APIGatewayProxyRequest) *events.APIGatewayProxyResponse {
+	return handleList(request, typeDataModel)
+}
+
```
lindsey-w, line 96: I wasn't sure if I should break this out into its own list logic or reuse some of what was here already.

api/gateway/analysis/api.yml
```diff
@@ -1415,6 +1658,56 @@ definitions:
...
+      - id
+      - lastModified
+
+  Mapping:
+    type: object
+    properties:
+      name:
+        $ref: '#/definitions/sourceName'
+      field:
+        $ref: '#/definitions/field'
```
lindsey-w, line 1696: Should this be an array in order to support making a single data model for multiple LogTypes? Or would this get too complicated (for example: if two logTypes have the field "ipAddress" but in one it means "source_ip" and in the other it means "destination_ip")

nhakmiller (commented):
I'll review this one tomorrow (sorry for the delay!), but it looks like this PR will be one part of a series of PRs for the data model feature correct? I (and a few others) have started using feature branches for that, this could be a good candidate for a feature branch.

internal/core/analysis_api/handlers/create_datamodel.go
```diff
@@ -0,0 +1,181 @@
...
+			Enabled:  item.Enabled,
+			LogTypes: item.ResourceTypes,
+		})
+		return nil
+	})
+
+	if err != nil {
+		zap.L().Error("failed to scan dynamodb for enabled data models", zap.Error(err))
+		return false, err
+	}
```
nhakmiller, line 176: Elsewhere in this function you simply return an error, here you log an error and then return it. Any error that gets returned all the way up will be logged anyways, so unless there is a specific reason I don't know if we need this error as well.

If you're worried about losing context, you can try returning an `errors.WithStack()` or `errors.WithMessage()` to add some context. I don't think it's the standard `errors` package we use, but if you check the codebase for one of those you should be able to find it.

internal/core/analysis_api/handlers/list.go
nhakmiller: Mmm this seems reasonable enough.

internal/core/analysis_api/handlers/modify_datamodel.go
```diff
@@ -0,0 +1,69 @@
...
+		ID:            input.ID,
+		Mappings:      input.Mappings,
+		ResourceTypes: input.LogTypes,
+		Type:          typeDataModel,
+	}
+
+	// check for conflicting enabled DataModels and return an error
+	// so the user doesn't expect this value to be updated
+	if bool(input.Enabled) {
+		enabledCheck, err := isSingleDataModelEnabled(input)
```
nhakmiller, line 50: Again you're checking here in the calling function and inside `isSingleDataModelEnabled` for whether or not input is enabled. Checking in just `isSingleDataModelEnabled` is probably fine.

nhakmiller (commented):
I know @austinbyers is actively working on removing API gateway & swagger right now, when he gets back from PTO next week might be worth syncing with him to make sure you don't have a bunch of crazy merge conflicts with his work.

api/gateway/analysis/api.yml
```diff
@@ -336,7 +410,7 @@ paths:
     # TODO: add paging / merge into ListPolicies endpoint
     get:
       operationId: GetEnabledPolicies
-      summary: List all enabled rules/policies for a customer account for backend processing
+      summary: List all enabled rules/policies/datamodels for a customer account for backend processing
```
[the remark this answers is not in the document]
lindsey-w: I was thinking in the rules engine it might be useful to be able to query for all enabled data models - do you think it would be more appropriate to separate it from these by putting it under /datamodel/enabled?

api/gateway/analysis/api.yml
```diff
@@ -1415,6 +1658,56 @@ definitions:
...
+      lastModified:
+        $ref: '#/definitions/modifyTime'
+      logTypes:
+        $ref: '#/definitions/TypeSet'
+
+    required:
+      - id
+      - lastModified
+
+  Mapping:
```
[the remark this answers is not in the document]
lindsey-w: I was also wondering if `field` and `method` were also too generic - should I also update those to be `MappingField` and `MappingMethod`?

internal/core/analysis_api/handlers/create_datamodel.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// we also need to verify that field and method are mutually exclusive in the input
+	for _, mapping := range result.Mappings {
+		if mapping.Field != "" {
+			if mapping.Method != "" {
+				return nil, errMappingTooManyOptions
+			}
+		} else if mapping.Method == "" {
+			return nil, errFieldOrMethodMissing
+		}
```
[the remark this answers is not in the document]
lindsey-w: I like your version better too - easier to read.

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -224,6 +235,7 @@ func extractZipFile(input *models.BulkUpload) (map[models.ID]*tableItem, error)
 			DisplayName:   models.DisplayName(config.DisplayName),
 			Enabled:       models.Enabled(config.Enabled),
 			ID:            models.ID(config.PolicyID),
+			Mappings:      make([]*models.DataModelMapping, len(config.Mappings)),
```
nhakmiller, line 238: This means that for all non data models, this field will be an empty slice (as opposed to `nil`). I'm not sure if this will matter,  but it might make e.g. JSON serialization or nil checks not behave as expected. We have `omitempty` tags today so it doesn't strictly matter right now I think, but if anything gets changed this would behave slightly differently than the other fields. See this snippet for a demonstration: https://play.golang.org/p/1A3YPXEOeWP.

I would recommend removing this  line, and on lines 253 put an `if len(config.Mappings>0){ analysisItem.Mappings = make([]*models.DataModelMapping, len(confg.Mappings))` or something like that.

internal/core/analysis_api/handlers/create_datamodel.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	// Scan dynamo
+	// Build the dynamodb scan expression
+	projection := expression.NamesList(
+		expression.Name("id"),
+		expression.Name("enabled"),
+		expression.Name("resourceTypes"),
+	)
```
nhakmiller, line 140: You might be able to call `ListEnabledPolicies` here.  Although if that adds any complications don't worry about it, just thinking about saving some duplicated code if applicable.

nhakmiller (commented):
Looking good! A few small comments then good to merge I think.

api/gateway/analysis/api.yml
```diff
@@ -1415,6 +1658,56 @@ definitions:
...
+        $ref: '#/definitions/sourceName'
+      field:
+        $ref: '#/definitions/field'
+      method:
+        $ref: '#/definitions/method'
+    required:
+      - name
+    #oneOf:
+    #  - required: [method, name]
+    #  - required: [field, name]
```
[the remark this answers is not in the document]
lindsey-w: whoops! That was me "I wish this worked" but forgot to remove - fixed.

internal/core/analysis_api/handlers/create_datamodel.go
lindsey-w: I had considered that 🤔  I didn't know which method would be preferred: (1) doing to filtering via the scanInput or (2) doing the filtering after getting back all the enabled data models. I went with (1), but I don't have a particularly good argument for it - do you think it's preferable to return all (using `ListEnabledPolicies`) and doing the filtering in this function?

api/gateway/analysis/api.yml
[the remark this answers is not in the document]
austinbyers: I'm probably going to remove this endpoint completely in #1690 and consolidate to a single `ListDetections` endpoint, but globals and models would probably stay separate. So it doesn't matter too much where you put it, I'll adjust it

internal/core/analysis_api/handlers/create_datamodel.go
austinbyers: I'm going to have a `ListDetections` that will handle policies/rules/scheduled queries, but I think it's fine to keep data models as a separate scan

austinbyers (approved):
Looks pretty good! Thanks so much for adding the integration tests, that will help a lot.

After this merges, I'll work on #1690 to rip out all the swagger. As part of that work, I'll also consolidate some of the Go models and list endpoints

PR #1951, opened 2020-11-04
Mitigate race in creating Athena views

rleighton:
## Background

The Glue Catalog is eventually consistent and if we are too fast the above schema changes will not be visible to Athena

## Changes

- Add 1 sec delay before creating views

## Testing

- mage test:ci

### DIFF
--- a/internal/log_analysis/datacatalog_updater/process/create_tables.go
+++ b/internal/log_analysis/datacatalog_updater/process/create_tables.go
@@ -20,6 +20,7 @@ package process
 
 import (
 	"context"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/sqs"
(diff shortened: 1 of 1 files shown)


s0l0ist (approved):

internal/log_analysis/datacatalog_updater/process/create_tables.go
```diff
@@ -82,6 +83,9 @@ func HandleCreateTablesMessage(ctx context.Context, msg *CreateTablesMessage) er
 		}
 	}
 
+	// the Glue Catalog is eventually consistent and if we are too fast the above schema changes will not be visible to Athena
+	time.Sleep(time.Second)
```
kostaspap, line 87: Instead of just sleeping here for 1 second (we guess it is enough), we should check if the `deployedLogTables` contains the tables we just created. Loop (with a sleep) until we see the tables, then run the Athena query to create a view. Makes more sense?

rleighton: Actually, if we want to validate we need to test that the schema is updated also, not just the table exits. That said, the question that I'm not sure of is if the really works ... even if we can "read our own writes" I'm not sure that is the same as "Athena reads our writes". Do you know?

kostaspap: I have to say I am missing a bit of context of how the issue was identified so I am not sure how we could protect from it. Let's chat offline

PR #1956, opened 2020-11-05
Bulk change Alert statuses [1/2]

alexmylonas:
## Description

This is the first PR for implementing the Bulk change Alert statuses. 

In this PR, i am implementing the new context needed `SelectContext` that keeps track of selections occuring.

Some changes to existing components to handle this and introducing 2 new  checkbox components that handle all operations for this context.

Pending: 

- Perform the actual API
- Appropriatly update alerts displayed with the new status

## Design
You can find mock for [unselected alerts](https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/0d293ff7-a6c6-4ddc-9dd0-442c63b91995/commits/latest/files/c2bbc1a8-4778-4cd6-93a5-4630d8868201/layers/87A29D67-F5C4-408F-87E8-DC77FE7F5454?collectionId=0e79571a-428a-4c7f-bb8d-f6c4e457eb57&collectionLayerId=50e28871-e02c-48b4-8775-0f126b5bf15f&mode=build) and for [selected alerts](https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/0d293ff7-a6c6-4ddc-9dd0-442c63b91995/commits/latest/files/c2bbc1a8-4778-4cd6-93a5-4630d8868201/layers/AA87C9D5-DFA0-420D-8240-521AB11A7C8B?collectionId=0e79571a-428a-4c7f-bb8d-f6c4e457eb57&collectionLayerId=7fefe367-b34a-4370-9257-357cbdba04dd&mode=build&selected=525110283-25D69D6C-9D6F-46A3-991D-5EDE8C1E7FD7)
## Screenshots:
Screenshot 2020-11-05 at 1 33 43 PM

### Checklist:

- [ ] Make sure the related issue is referenced
- [ ] Make sure your changes are accessible (addition of aria-attributes & correct HTML semantics)
- [x] ~~Make sure the changes are responsive (appearing decently on devices)~~


web/src/components/PanelWithSelection/PanelWithSelection.tsx
```diff
@@ -0,0 +1,73 @@
...
+            checked={!!selected.length}
+            onClick={() => {
+              if (selected.length) {
+                deselectAll();
+              } else {
+                selectAll();
+              }
+            }}
+            aria-label="select all checkbox"
+          />
```
alexmylonas, line 61: Checkbox needs to be updated on pounce with a new state, i.e `semi` selected

An alternative would to create a new component extending this.

web/src/components/cards/AlertCard/AlertCard.tsx
```diff
@@ -31,18 +31,51 @@ import useAlertDestinations from 'Hooks/useAlertDestinations';
...
+interface AlertCardWithSelectionProps {
+  alert: AlertSummaryFull;
+  hideRuleButton?: boolean;
+  selectionEnabled: true;
+  selected: string[];
+  onSelect: (string) => void;
+  onDeselect: (string) => void;
+}
+
+type AlertCardProps = AlertCardSimpleProps | AlertCardWithSelectionProps;
```
alexmylonas, line 52: Is this the optimal syntax for conditional props?

web/src/hooks/useMultiselect.tsx
```diff
@@ -0,0 +1,53 @@
...
+      addItem,
+      removeItem,
+      resetSelection,
+      selectAll,
+    }),
+    [selected.length, addItem, resetSelection, selectAll]
+  );
+}
+
+export default useMultiselect;
```
alexmylonas, line 53: Not sure if this the correct approach or we need to add this as context

web/src/pages/ListAlerts/ListAlertSelection/ListAlertSelection.tsx
```diff
@@ -0,0 +1,77 @@
...
+    <Flex justify="flex-end" align="center">
+      <Formik<ListAlertSelectionFormValues>
+        initialValues={initialValues}
+        onSubmit={
+          // TODO: Make this functional
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          (values: ListAlertSelectionFormValues) => {
+            // @ts-ignore
+            // eslint-disable-next-line no-console
+            console.log('values');
```
alexmylonas, line 52: Ignore this

web/src/pages/ListAlerts/ListAlerts.tsx
```diff
@@ -123,13 +124,27 @@ const ListAlerts = () => {
...
               {alertItems.map(alert => (
-                <AlertCard key={alert.alertId} alert={alert} />
+                <AlertCard
+                  key={alert.alertId}
+                  alert={alert}
+                  selectionEnabled
+                  selected={selected}
+                  onSelect={addItem}
+                  onDeselect={removeItem}
+                />
```
alexmylonas, line 147: I am passing some many props here that feel is wrong, any suggestions?

alexmylonas:
@3nvi Updated PR to have a `SelectProvider` & `Context`, please have a look a let me know!

web/src/components/utils/SelectContext/SelectContext.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+export { SelectContext, MemoizedSelectProvider as SelectProvider, withSelectContext, useSelect };
+
+const SelectCheckboxComponent: React.FC<SelectCheckboxProps> = ({ id, ...rest }) => {
+  const { selection, selectItem, deselectItem } = useSelect();
+  const isSelected = selection && selection.find(i => i === id);
+  return (
+    <Checkbox
+      checked={!!isSelected}
+      onClick={() => (isSelected ? deselectItem(id) : selectItem(id))}
```
3nvi, line 99: If we create a `toggleItem`, we would then do `onClick={() =>  toggleItem(id)}`

web/src/hooks/useSelect.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
```
3nvi, line 19: you already implemented a `useSelect` inside the Provider component

web/src/pages/ListAlerts/ListAlertSelection/ListAlertSelection.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+interface ListAlertSelectionFormValues {
+  status: string;
+}
+
+const ListAlertSelection: React.FC = () => {
+  const { selection } = useSelect();
+  return (
+    <Flex justify="flex-end" align="center">
+      <Formik<ListAlertSelectionFormValues>
```
3nvi, line 43: I don't think we need an entire `Formik`form for this. Simply using `<Combobox>` and firing an API request `onChange` would do it 😄

web/src/pages/ListAlerts/ListAlerts.tsx
```diff
@@ -50,7 +52,7 @@ const ListAlerts = () => {
       },
     },
   });
-
+  const { selectAll } = useSelect();
```
3nvi, line 55: using this hook here will cause the entire component to re-render on every selection change. We shouldn't use `useSelect` at all here.

web/src/components/PanelWithSelection/PanelWithSelection.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const PanelWithSelection: React.FC<PanelProps> = ({
+  title,
+  select,
+  filters,
+  children,
+  selectAll,
+}) => {
+  const { selection, resetSelection } = useSelect();
+  return (
```
3nvi, line 38: @alexmylonas I don't like the idea of "binding" the Panel with a selection, since it forces its children to "rerender" on every selection update. Ideally we should keep the `Panel` that we have and just create separate components for the "select all" checkbox and use the `useSelect` **inside it**, so that only the checkbox re-renders

3nvi (commented):
Good work @alexmylonas . Still stuff to go through tho

web/src/components/cards/AlertCard/AlertCard.tsx
```diff
@@ -31,18 +31,51 @@ import useAlertDestinations from 'Hooks/useAlertDestinations';
...
   const { alertDestinations, loading: loadingDestinations } = useAlertDestinations({ alert });
   const { allDestinationDeliveredSuccessfully, loading } = useAlertDestinationsDeliverySuccess({
     alert,
   });
+
+  const isAlertSelected = selected && selected.find(a => a === alert.alertId);
   return (
     <GenericItemCard>
+      <Flex align="start" pr={2}>
+        {selectionEnabled && (
```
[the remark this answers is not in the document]
3nvi: > Indeed there is, the underlying pounce Checkbox component has a padding which i cant seem to override any ideas


Wrap it in a `<Box translate="transform3d(0,-10px,0">...</Box>` or something

> We could have it a default, imo each aria-label checkbox should include the alert id - otherwise how would you know which one it refers too. No?

You are right. Let's just add a **dynamic** `aria-label` to `Checkbox` from **within** the `SelectCheckbox` instead of passing it as a prop to the latter (like the way you did in the `SelectAllCheckbox`)

web/src/components/utils/SelectContext/SelectContext.tsx
```diff
@@ -0,0 +1,115 @@
...
+    () => ({
+      selection,
+      selectAll,
+      deselectItem,
+      selectItem,
+      resetSelection,
+      checkIfSelected,
+      toggleItem,
+    }),
+    [selection, resetSelection, selectAll, selectItem, deselectItem, checkIfSelected]
```
3nvi, line 99:
```suggestion
    [selection, resetSelection, selectAll, selectItem, deselectItem, checkIfSelected, toggleItem]
```

web/src/components/utils/SelectContext/SelectContext.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * @public
+   * Reset selection to an empty array
+   *
+   */
+  const resetSelection = React.useCallback(() => setSelected([]), []);
+
+  // NOTE: Those ids could be passed on hook declaration
+  const selectAll = React.useCallback((ids: string[]) => {
```
3nvi, line 60: minor nitpick: either we code all of the methods as `() => fn()` or `() => { return fn() }`. Let's pick one 😆

web/src/components/utils/SelectContext/SelectContext.tsx
```diff
@@ -0,0 +1,115 @@
...
+  deselectItem: (id: string) => void;
+  resetSelection: () => void;
+  selectAll: (ids: string[]) => void;
+  checkIfSelected: (id) => boolean;
+  toggleItem: (id) => void;
+}
+
+const SelectContext = React.createContext<SelectContextValue>(undefined);
+
+const SelectProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
```
3nvi, line 33: we should add an **optional** prop for `initialSelection` as an array of strings which would power the initial state of the `selection`

web/src/components/utils/SelectContext/SelectAllCheckbox.tsx
```diff
@@ -0,0 +1,45 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import { useSelect } from 'Components/utils/SelectContext/SelectContext';
+import { Checkbox } from 'pouncejs';
+
+interface SelectAllCheckboxProps {
+  ids: string[];
```
3nvi, line 24:
```suggestion
  selectionIds: string[];
```

see my comment below to understand why

web/src/components/utils/SelectContext/SelectCheckbox.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import { Checkbox } from 'pouncejs';
+import { useSelect } from './SelectContext';
+
+interface SelectCheckboxProps {
+  id: string;
```
3nvi, line 24: Wee should change this since it clashes with the native HTML  `id` property. Thus, if a developer wants to add an HTML `id` to this `Checkbox`, he/she now can't. By renaming it to `selectionId` we avoid clashes like that

```suggestion
  selectionId: string;
```

web/src/pages/ListAlerts/ListAlertSelection/ListAlertSelection.tsx
3nvi: Nope, not at all :)

web/src/pages/ListAlerts/ListAlertSelection/ListAlertSelection.tsx
```diff
@@ -0,0 +1,76 @@
...
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          (values: ListAlertSelectionFormValues) => {
+            // @ts-ignore
+            // eslint-disable-next-line no-console
+            console.log('values');
+          }
+        }
+      >
+        <Form>
+          <Flex spacing={4} align="center" pr={4}>
```
3nvi, line 56:
```suggestion
          <Flex spacing={4} align="center">
```

you shouldn't have a `pr` here. As we've said, margins, paddings & positions should be added on thee place where the component is rendered and **not** on the component declaration 😄

web/src/pages/ListAlerts/ListAlertSelection/ListAlertSelection.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <Text>{selection.length} Selected</Text>
+            <Box width={151}>
```
3nvi, line 58: do we need the `width` here? If we do, then let's make it something more explicit like `150` and `160` instead of `151` 😆

web/src/pages/ListAlerts/ListAlerts.tsx
```diff
@@ -123,22 +124,21 @@ const ListAlerts = () => {
...
       <Panel
-        title="Alerts"
-        actions={
-          // Using a Box to add some spacing as
-          // ListAlertFilters tends to cover the whole space available
-          <Box>
-            <ListAlertFilters />
-          </Box>
+        title={
+          <Flex align="center">
```
3nvi, line 130: should there be a `spacing={2}` between them?

web/src/pages/ListAlerts/ListAlerts.tsx
```diff
@@ -123,22 +124,21 @@ const ListAlerts = () => {
...
+            <SelectAllCheckbox ids={alertItems.map(a => a.alertId)} />
```
3nvi, line 131: let's memoize this Alex. Use a `React.useMemo` else this prop will change on every render and the `React.memo` used in `SelectAllCheckbox` won't work 😉

web/src/pages/ListAlerts/ListAlerts.tsx
```diff
@@ -153,4 +153,4 @@ const ListAlerts = () => {
   );
 };
 
-export default withSEO({ title: 'Alerts' })(ListAlerts);
+export default compose(withSEO({ title: 'Alerts' }), withSelectContext)(ListAlerts);
```
3nvi, line 156: This may sound a bit weird, but this has performance issues. Let me explain:

Whenever a checkbox is clicked, the value of this provider changes, which means that the entire tree **below** the  provider component is re-rendered. Thus, if you put the provideer high in the React tree, then **more** components will re-render. To mitigate that, there are 2 alternatives:

1. Move the `<SelecttionProvider>` and make it as **nested** in the component tree as possible. In our use case that would mean moving it right above the `<Panel>` of this file

2. Add a `React.memo` to the component **right below the Provider**. If you remember, the provider changing value means that all the tree re-renders, but the provider **isn't passing props** to any components. That means that if you add a `React.memo` to the immediate child of the provider component, you block this un-necessary re-rendering from cascading to the entire tree. In our use case, if you do 

```ts
export default compose(withSEO({ title: 'Alerts' }), withSelectContext, React.memo)(ListAlerts);
```

you would be fine 😉

3nvi (commented):
Amazing amazing work @alexmylonas . This is great stuff here!

Couple of small bugfixes & comments to address and we are good to proceed

alexmylonas (commented):
Implemented all your suggestions! Have a look and let me know!

web/src/components/cards/AlertCard/AlertCard.tsx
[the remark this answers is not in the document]
3nvi: Just make sure that `-10px` is what wee neeed indeed. I mentioned a number randomly, it might need to be 4 or 8 😄

web/src/components/utils/SelectContext/SelectCheckbox.tsx
```diff
@@ -0,0 +1,40 @@
...
+  selectionId: string;
+}
+
+const SelectCheckboxComponent: React.FC<SelectCheckboxProps> = ({ selectionId, ...rest }) => {
+  const { checkIfSelected, toggleItem } = useSelect();
+  const isSelected = checkIfSelected(selectionId);
+  return (
+    <Checkbox
+      checked={isSelected}
+      aria-label={`select checkbox for ${selectionId}`}
```
3nvi, line 33:
```suggestion
      aria-label={isSelected ? `unselect ${selectionId}` : `select ${selectionId}`}
```

web/src/pages/ListAlerts/ListAlerts.tsx
```diff
@@ -110,6 +111,15 @@ const ListAlerts = () => {
...
 
+  const PanelTitle = React.useMemo(() => {
+    return (
+      <Flex align="center" spacing={2}>
+        <SelectAllCheckbox selectionIds={alertItems.map(a => a.alertId)} />
+        <Text>Alerts</Text>
+      </Flex>
+    );
+  }, [alertItems]);
+
```
3nvi, line 122: When I said `useMemo` I meant to memoize this `alertItems.map(a => a.alertId)` nott the entire rendering 😆 . Can wee just memoize the map result?

3nvi (approved):
Looks awesome @alexmylonas . Left a couple of final comments, but nothing blocking. 

I'm pre-approving this 🎉 . Let's merge it into a feature branch if we can 😃

PR #1961, opened 2020-11-05
Tidy config for Athena workgroups

rleighton:
## Changes

- Add `EnforceWorkGroupConfiguration` to the Panther Athena Workgroup to be consistent with Panther Enterprise where we need the Parquet compaction process to override the s3 path
- Add trailing `/` to s3 paths since the console UI enforces that they end in `/`

## Testing

- mage test:ci
- mage deploy, then checked Athena workgroups in console


serbantanasa (approved):
lgtm

Issue #1938, opened 2020-11-04
Remediation fails with error on most Policies


alexmylonas:
### Describe the bug

There is seems to be an internal server error when trying to remediate any resource, with the following error:

`A custom error was thrown from a mapping template.`

I tested this after i onboarded a **new compliance source** and **enabled real-time notifications**. I am not sure if those are related but could be an indication.

### Steps to reproduce

1. Login in release deployment [here](https://web-892058857.us-east-1.elb.amazonaws.com/)
2. Go to policies
3. Find one that fails and click on it.
4. Find a failing resource in it and try to remediate via the dropdown
5. See the error

### Expected behavior

- Either successfully remediate resource or fail gracefully

### Environment

- Panther version or commit: v1.12-rc (d028a656e09e258331c9e2fb774c119a8cbac324)

### Screenshots
![Screenshot 2020-11-04 at 3 21 56 PM](https://user-images.githubusercontent.com/14179917/98117377-542a4380-1eb2-11eb-8891-897e13bd1260.png)

3nvi:
That's an appsync error @alexmylonas. It might be a FE or a BE issue. We should see what comes back from this resolver and evaluate accordingly

alexmylonas:
Could this be related to @rleighton work with refactoring that took place on the last sprint?

I don't think much else changed either BE or FE regarding this code region, but I will also check the FE part.

3nvi:
Perfect, @alexmylonas can we get an update on this asap?

alexmylonas:
Logs from Cloudwatch on request: 
I am not sure how to interpret this or troubleshot further, any ideas?
```
{
    "level": "warn",
    "ts": 1604573823.9487572,
    "caller": "handlers/post.go:44",
    "msg": "failed to invoke remediation",
    "application": "panther",
    "requestId": "8838c411-28e7-464a-9cd5-cd94b677a2d4",
    "requestMethod": "POST",
    "requestPathParameters": null,
    "requestQueryParameters": null,
    "requestResource": "/remediate",
    "error": "Encountered issue when getting policy: response status code does not match any response statuses defined for this endpoint in the swagger spec (status 403): {}",
[...]

PR #1962, opened 2020-11-05
Fix remediation permissions

nhakmiller:
## Background

Closes https://github.com/panther-labs/panther/issues/1938. The issue is that Austin removed too many permissions when removing the compliance api, this restores some of the permissions that are still needed.

## Changes

- Restores perms

## Testing

- Manually


alexmylonas (approved):
Can't comment on the code, but deployed on my env, and seems to work!

austinbyers (commented):
Yep, I introduced the bug in #1880 . Good catch! My apologies

nhakmiller:
> Good catch! My apologies

@alexmylonas caught it during release testing, and I fixed it. So it was a group effort :) And this  is exactly why we have release testing, so no worries

PR #1964, opened 2020-11-06
Refactoring log pulling

kostaspap:
## Background

Some further cleanup in the log processing. Removing 80 lines of code!

## Changes

- Instead of checking if we have exceeded timeout before every SQS poll, we use ctx expiration instead.

## Testing

- mage test:ci


internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -191,3 +172,19 @@ func highMemoryUsage() (heapUsedMB, memAvailableMB float32, isHigh bool) {
...
+
+func receiveFromSqs(ctx context.Context, sqsClient sqsiface.SQSAPI) ([]*sqs.Message, error) {
+	request := &sqs.ReceiveMessageInput{
+		WaitTimeSeconds:     aws.Int64(0),
+		MaxNumberOfMessages: aws.Int64(sqsMaxBatchSize),
+		QueueUrl:            &common.Config.SqsQueueURL,
+	}
+	receiveMessageOutput, err := sqsClient.ReceiveMessageWithContext(ctx, request)
+
+	if err != nil && err != context.DeadlineExceeded && err != context.Canceled {
```
kostaspap, line 184: Don't report error in case timeout has been reached.

rleighton (approved):
Very nice!

PR #1966, opened 2020-11-06
Fix gzip files being mis-detected as ms-fontobject

giorgosp:
## Background

Sometimes we mis-detect gzipped log files as ms-fontobject.

## Changes

Change the detection code to check directly for the gzip signature instead of relying on the `http.DetectContentType` function. That function checks for many content types and produces a false positive in some cases.

## Testing

- Unit

### DIFF
--- a/internal/compliance/resource_processor/processor/analysis_api.go
+++ b/internal/compliance/resource_processor/processor/analysis_api.go
@@ -19,9 +19,9 @@ package processor
  */
 
 import (
-	"github.com/pkg/errors"
 	"time"
 
+	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
 	"github.com/panther-labs/panther/api/gateway/analysis/client/operations"
--- a/internal/compliance/resource_processor/processor/analysis_api.go
+++ b/internal/compliance/resource_processor/processor/analysis_api.go
@@ -19,6 +19,7 @@ package processor
  */
 
 import (
+	"github.com/pkg/errors"
 	"time"
 
 	"go.uber.org/zap"
@@ -49,8 +50,7 @@ func getPolicies() (policyMap, error) {
 	result, err := analysisClient.Operations.GetEnabledPolicies(
 		&operations.GetEnabledPoliciesParams{HTTPClient: httpClient, Type: string(models.AnalysisTypePOLICY)})
 	if err != nil {
-		zap.L().Error("failed to load policies from analysis-api", zap.Error(err))
-		return nil, err
+		return nil, errors.WithMessage(err, "failed to load policies from analysis-api")
 	}
 	zap.L().Info("successfully loaded enabled policies from analysis-api",
 		zap.Int("policyCount", len(result.Payload.Policies)))
--- a/internal/log_analysis/log_processor/sources/s3.go
+++ b/internal/log_analysis/log_processor/sources/s3.go
@@ -20,6 +20,7 @@ package sources
 
 import (
 	"bufio"
+	"bytes"
 	"compress/gzip"
 	"io"
 	"net/http"
@@ -176,17 +177,16 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
 		// if it's plain text, just return the buffered reader
 		streamReader = bufferedReader
 	} else if strings.HasPrefix(contentType, "application/x-gzip") {
-		var gzipReader *gzip.Reader
-		gzipReader, err = gzip.NewReader(bufferedReader)
+		gzipReader, err := gzip.NewReader(bufferedReader)
 		if err != nil {
 			err = errors.Wrapf(err, "failed to created gzip reader for s3://%s/%s",
 				s3Object.S3Bucket, s3Object.S3ObjectKey)
-			return
+			return nil, err
 		}
 		streamReader = gzipReader
 	} else {
 		err = &ErrUnsupportedFileType{Type: contentType}
-		return
+		return nil, err
 	}
 
 	dataStream = &common.DataStream{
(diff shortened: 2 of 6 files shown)


internal/compliance/resource_processor/processor/analysis_api.go
```diff
@@ -49,8 +50,7 @@ func getPolicies() (policyMap, error) {
 	result, err := analysisClient.Operations.GetEnabledPolicies(
 		&operations.GetEnabledPoliciesParams{HTTPClient: httpClient, Type: string(models.AnalysisTypePOLICY)})
 	if err != nil {
-		zap.L().Error("failed to load policies from analysis-api", zap.Error(err))
-		return nil, err
+		return nil, errors.WithMessage(err, "failed to load policies from analysis-api")
```
kostaspap, line 53: Why this change?

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -163,59 +158,34 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
...
-		// EOF or ErrBufferFull means file is shorter than n
-		case bufio.ErrBufferFull, io.EOF:
-			// not really an error
-		default:
-			return "", err
-		}
-	}
-	return http.DetectContentType(headerBytes), nil
+func checkGzip(r *bufio.Reader) (bool, error) {
+	magicNumber := []byte("\x1F\x8B\x08")
```
kostaspap, line 186: 1. You can make it a `const`
2. Rename from `magicNumber` to something more descriptive
3. This definitely requires some good and very descriptive comment 👍

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -163,59 +158,34 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
...
-	var streamReader io.Reader
-
-	// Checking for prefix because the returned type can have also charset used
-	if strings.HasPrefix(contentType, "text/plain") {
-		// if it's plain text, just return the buffered reader
-		streamReader = bufferedReader
-	} else if strings.HasPrefix(contentType, "application/x-gzip") {
-		var gzipReader *gzip.Reader
-		gzipReader, err = gzip.NewReader(bufferedReader)
+	if isGzip {
```
kostaspap, line 168: What if it is not gzip nor plaintext? How do we handle that? We had cases in the past that users would throw unsupported file types.

kostaspap: We need to check if a file is of gzip or text/plain type and discard it if it is none of that.

giorgosp: The previous code would still fire alarms if the file was of an unsupported content type. I saw this error due to an incident.

There isn't any 100% reliable way to check if a file is plain text. We can only solve (read: assume) by elimination. I can think of two ways:
- Use `http.DetectContetType()` again but check "manually" for gzip before calling that function. This function checks for many content types and lastly checks for utf8 text if non of the previous checks matched.
- Allow the classifier to run, it will fail if the log cannot be parsed (it could be that the file was not plain text but we can't be sure).

I guess we can code a hack by using `http.DetectContentType()` as a helper.

internal/log_analysis/log_processor/sources/s3.go
giorgosp: 1. Yes but didn't want its logic to spill over the whole file, since it is not a core operation of the file.
2. I can rename it to `signature` or sth (magic number is also valid but no so self-explanatory https://en.wikipedia.org/wiki/Magic_number_(programming))
3. Will add a link to the gzip signature or sth, there isn't much to describe

internal/log_analysis/log_processor/sources/s3.go
kostaspap: `sniff.go` has a matcher for `text/plain`. We could put this logic in the code just as you do for GZIP. 
See lines 93-107: https://golang.org/src/net/http/sniff.go

Looking at the `http.DetectContentType` code, it is not assuming that a file is `text/plain` by elimination, but it is checking for specific signatures. I believe we can do the same.

giorgosp: It uses the textSig last, after checking a looot of other signatures. I think we should just use a better package like https://github.com/gabriel-vasile/mimetype

internal/log_analysis/log_processor/sources/s3.go
giorgosp: Added link to gzip signature in `detectContettType()`

internal/log_analysis/log_processor/sources/s3.go
giorgosp: Actually I just kept our `detectContentType()` and resorted to do the gzip check before calling it. Killing two birds with one stone:
- Work around the bug in `http.DetectContentType()`
- Performance optmization, since most log files are gzipped.

The 3rd party packages I checked wouldn't help much because:
- filetype: I couldn't find a text matcher
- mimetype: After find a text type, it would go deeper in the tree to detect the text type as well (csv, json, script, etc), which would impact performance. It also uses a bigger sniff header (3kb compared to 512 bytes we use now).

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -179,43 +180,43 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
 		// if it's plain text, just return the buffered reader
 		streamReader = bufferedReader
 	} else if strings.HasPrefix(contentType, "application/x-gzip") {
-		var gzipReader *gzip.Reader
-		gzipReader, err = gzip.NewReader(bufferedReader)
+		gzipReader, err := gzip.NewReader(bufferedReader)
 		if err != nil {
-			err = errors.Wrapf(err, "failed to created gzip reader for s3://%s/%s",
+			return nil, errors.Wrapf(err, "failed to created gzip reader for s3://%s/%s",
```
kostaspap, line 185: You need to set the `err` to the new error, not just return it. Line 137 is using it.

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -179,43 +180,43 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
...
+		// EOF / ErrBufferFull means file is shorter than sniffLen, but not all detections need so large data.
+		return "", err
 	}
-	return http.DetectContentType(headerBytes), nil
+
+	// Try gzip first, for two reasons:
+	// 1. Performance: It the most usual file type for logs and there is an exact prefix check we can do to detect it.
+	// No need to wait for all the previous checks by http.DetectContentType().
+	// 2. http.DetectContentType() has a bug which mis-detects gzip for ms-fontobject.
+	gzipSignature := []byte("\x1F\x8B\x08") // https://mimesniff.spec.whatwg.org/#matching-an-archive-type-pattern
```
kostaspap, line 214: declare the variable outside of this block? So it doesn't have to be recreated on every function invocation.

internal/log_analysis/log_processor/sources/s3.go
giorgosp: Ah, good point. I hate that op logger. Will fix

internal/log_analysis/log_processor/sources/s3.go
giorgosp: Will do

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -176,45 +177,47 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
...
 			err = errors.Wrapf(err, "failed to created gzip reader for s3://%s/%s",
 				s3Object.S3Bucket, s3Object.S3ObjectKey)
-			return
+			return nil, err
 		}
 		streamReader = gzipReader
 	} else {
-		err = &ErrUnsupportedFileType{Type: contentType}
-		return
+		return nil, &ErrUnsupportedFileType{Type: contentType}
```
kostaspap, line 188: You need to set the `err` to appropriate value - not just return it.

giorgosp: Done

internal/log_analysis/log_processor/sources/s3.go
alxarch: The sniff patterns for 'text/plain' are worthless because they rely on [BOM](https://en.wikipedia.org/wiki/Byte_order_mark)  which is almost never used.

Any test that we can do to detect `text/plain` would only read a part of the file and thus would not be reliable. It would add needless overhead since it would be the same error case as failing to read from the file due to I/O error. 

I would suggest to remove the MIME matching altogether and use `detectGzip(r *bufio.Reader) (bool, err)` or `isGzip(head []byte) bool` and fallback to treating the file as plain text if not. In order to check the file we have already opened the connection to get the reader. Matching on MIME would be useful only if we could get the content type before opening the reader (i.e. by S3 metadata). 

If in the future we want to support other binary formats (i.e. parquet) we will have ways to detect these specifically.

kostaspap (approved):

PR #1967, opened 2020-11-06
Fixing ddb filtering

kostaspap:
## Background

Found this issue while looking at the code today. This could result in alerts-api returning partial results to the FE. 

According to description: 
```
	// The maximum number of items to evaluate (not necessarily the number of matching
	// items). If DynamoDB processes the number of items up to the limit while processing
	// the results, it stops the operation and returns the matching values up to
	// that point, and a key in LastEvaluatedKey to apply in a subsequent operation,
	// so that you can pick up where you left off. Also, if the processed dataset
	// size exceeds 1 MB before DynamoDB reaches this limit, it stops the operation
	// and returns the matching values up to the limit, and a key in LastEvaluatedKey
	// to apply in a subsequent operation to continue the operation. For more information,
	// see Query and Scan (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/QueryAndScan.html)
	// in the Amazon DynamoDB Developer Guide.
	Limit *int64 `min:"1" type:"integer"`

```

We however do some filtering on DDB side, so we are potentially "evaluating" a lot more results. 

## Changes

- Fixing bug where we could return partial alert results 

## Testing

- mage test:ci

### DIFF
--- a/deployments/log_analysis.yml
+++ b/deployments/log_analysis.yml
@@ -96,7 +96,7 @@ Parameters:
 Mappings:
   Functions:
     AlertsApi:
-      Memory: 256
+      Memory: 512
       Timeout: 60
     AlertsForwarder:
       Memory: 128
--- a/internal/log_analysis/alerts_api/table/list.go
+++ b/internal/log_analysis/alerts_api/table/list.go
@@ -82,7 +82,6 @@ func (table *AlertsTable) ListAll(input *models.ListAlertsInput) (
 		KeyConditionExpression:    queryExpression.KeyCondition(),
 		ExclusiveStartKey:         queryExclusiveStartKey,
 		IndexName:                 index,
-		Limit:                     aws.Int64(*queryResultsLimit * 4), //optimization accounting for filtering
 	}
 
 	var lastKey DynamoItem


rleighton (approved):
I think you should do a `mage deploy` + an `AlwaysTrue` rule and test this (more than just `mage test:ci`).

s0l0ist (approved):
This will alleviate the problem, but not solve it completely. 

A real solution would be to store another column with the title in all upper or lower case, transform the query to match that case, and issue the filter query ddb. 

Approving as this seems like a good stopgap.

kostaspap:
> This will alleviate the problem, but not solve it completely.
> 
> A real solution would be to store another column with the title in all upper or lower case, transform the query to match that case, and issue the filter query ddb.
> 
> Approving as this seems like a good stopgap.

This is not solving the performance issue. This is another bug - we would return partial results.

kostaspap:
> > This will alleviate the problem, but not solve it completely.
> > A real solution would be to store another column with the title in all upper or lower case, transform the query to match that case, and issue the filter query ddb.
> > Approving as this seems like a good stopgap.
> 
> This is not solving the performance issue. This is another bug - we would return partial results to the user.

To give more background: The `limit` parameter in the QueryInput is setting how many DDB items the query will *evaluate*, not how many items it will return. 

So let's say that we have 100 items in our table matching the (Partition,Sort) key provided in the QueryInput. The imagine we provide a DDB-side Filter that would match only a subset of the items. 

If we specify a `limit` of , say, 10, DDB does the following: 
1. Fetches the first 10 items based on (Partition, Sort) key
1. Passes them through the DDB-side filter
1. Returns us the items matched in Step #2 and finishes. 

`limit` is not the number of items returned, but the number of items evaluated/scanned.

s0l0ist:
> > > This will alleviate the problem, but not solve it completely.
> > > A real solution would be to store another column with the title in all upper or lower case, transform the query to match that case, and issue the filter query ddb.
> > > Approving as this seems like a good stopgap.
> > 
> > 
> > This is not solving the performance issue. This is another bug - we would return partial results to the user.
> 
> To give more background: The `limit` parameter in the QueryInput is setting how many DDB items the query will _evaluate_, not how many items it will return.
> 
> So let's say that we have 100 items in our table matching the (Partition,Sort) key provided in the QueryInput. The imagine we provide a DDB-side Filter that would match only a subset of the items.
> 
> If we specify a `limit` of , say, 10, DDB does the following:
> 
> 1. Fetches the first 10 items based on (Partition, Sort) key
> 2. Passes them through the DDB-side filter
> 3. Returns us the items matched in Step #2 and finishes.
> 
> `limit` is not the number of items returned, but the number of items evaluated/scanned.

Actually, this code partially addresses performance because the chunk size is larger (up to 1MB worth of data).

The code would normally do those 3 steps you described; however, we loop until we meet our page size or dynamo finishes querying results in [table.Client.QueryPages](https://github.com/panther-labs/panther/blob/master/internal/log_analysis/alerts_api/table/list.go#L91)

So the steps are:
 1. Fetches the first 10 items based on (Partition, Sort) key
 2. Passes them through the DDB-side filter and run the query
 3. If we meet our page size of 25 or if dynamo has no more items from the query, return results
 4. If we don't have enough results to meet our page size, go to step 1

s0l0ist:
After speaking internally, @kostaspap and I agree this will alleviate the performance issue a little, but is not a long term solution.

We need to think about partitioning the data differently to avoid constant sequential querying of data. For example, if we partition by year/month/week we could run 4 queries in parallel for the past 4 weeks to filter results in a much more performant manner. We also need to add new columns for case-insensitive queries so data is _never_ post-filtered after coming from dynamo.

rleighton:
> I think you should do a `mage deploy` + an `AlwaysTrue` rule and test this (more than just `mage test:ci`).

BTW, when you test this with an `AlwaysTrue` rule, let it get a huge number of matches and then see if you can figure out the performance issue we are seeing using this data.

s0l0ist:
> > I think you should do a `mage deploy` + an `AlwaysTrue` rule and test this (more than just `mage test:ci`).
> 
> BTW, when you test this with an `AlwaysTrue` rule, let it get a huge number of matches and then see if you can figure out the performance issue we are seeing using this data.

Right, we have an account with ~4.5k alerts to test with already.

PR #1971, opened 2020-11-06
Slightly change default Log Analysis Overview Date Range

3nvi:
## Background

When viewing the LogAnalysis Overview page you initially see a loading animation and suddenly charts appear. If you change page and press "back" you see the same loading behavior happening although the page was already loaded previously.

The reason is that the API params have changed and the cache got busted. The reason the API params have changed is that the default date range that we sent over to the API is 3 days from **now**. The **now** changes every second which causes Apollo to not serve data from the cache but go straight to the server.

To fix that, this PR forces the default date range to be the end of the current hour (instead of right now), so we can at least serve immediately data from the cache for a max of 1 hour. Because we could have stale data, we obviously do a background update to fetch fresh stuff as soon as we show the "cached" data.

This is not perfect, it has edge cases where it won't give any advantage (i.e. we load the page at `HH:59`), but it serves the purpose.

 
## Changes

- Use a more wide default date range instead of "now"
- Update tests

## Testing

- `npm run test`


web/src/components/GuardedRoute/GuardedRoute.tsx
```diff
@@ -64,7 +64,7 @@ const GuardedRoute: React.FC<GuardedRouteProps> = ({ limitAccessTo, ...rest }) =
     // referrer will exist)
   } else {
     redirectData = {
-      ...location.state.referrer,
+      ...location.state?.referrer,
```
3nvi, line 67: fixes an edge case scenario

alexmylonas (approved):
LGTM!

web/src/helpers/utils.tsx
```diff
@@ -362,12 +362,22 @@ export const getCurrentYear = () => {
...
-  return `${dayjs(date).subtract(days, 'day').toISOString().split('.')[0]}Z`;
+export const getGraphqlSafeDateRange = ({
+  days = 0,
+  hours = 0,
+}: {
+  days?: number;
+  hours?: number;
+}) => {
+  const utcNow = dayjs.utc();
+  const utcDaysAgo = utcNow.subtract(days, 'day').subtract(hours, 'hour');
```
vorillaz, line 373: 🤝

vorillaz (approved):

Issue #1957, opened 2020-11-05
Table updates during deployments are broken


alxarch:
### Describe the bug

The solution provided in #1912 changed the way updates happen for log tables.
The general approach of sending a message to the datacatalog updater SQS was correct.
The problematic choice was to only update the log types used in current sources. This breaks 'all' views if a client for some reason has deleted a source after collecting data.
Creation of views fails if the schema was modified to include new indicator or core fields. In the case of 1.12 release that includes the `p_alert_context` added to rule matches table. 
Since log tables created at any point would be left in the data lake and will be visible for queries, they should be properly updated and included in the 'all' views *regardless of whether they are being used by a source*. 

On a minor note, there was a change to the `DeployedLogTypes` function that includes stuff from the 'panther_cloudsecurity' database which I am not sure should be there.

## Suggested solution

We use the logtypes API to fetch **all available log types** and filter them with `DeployedLogTypes` to get the right set of log types to update.

Update: The code for 'full' database sync during upgrades and 'create tables' during source updates has to be split as the logic is somewhat different in each case. The linked PR addresses that.

rleighton:
We hit the this upgrading the internal account due to the schema change also.

rleighton:
Q: If a log source is deleted (not mentioned in any source), why do we need to up date it?

alxarch:
The current way 'all' views work is to include all tables _deployed to glue_. Updating the views based on log types used in views is more complex since we need to keep modified/deleted sources in sync with the views. My suggestion is to keep this behavior since it will be confusing to have tables in the data explorer that are not present in 'all' views.

The current code uses the same handler for both updating the full database when deploying and when a source is updated. I suggest to split these into two separate events since their logic is quite diffrerent.

alxarch:
Closed by #1975

PR #1975, opened 2020-11-06
Quick fix for deployment updates

alxarch:
## Background

Quickfix for #1957 

## Changes

- Also include deployed tables on sync updates

### DIFF
--- a/internal/log_analysis/datacatalog_updater/process/create_tables.go
+++ b/internal/log_analysis/datacatalog_updater/process/create_tables.go
@@ -32,6 +32,7 @@ import (
 	"github.com/panther-labs/panther/internal/log_analysis/awsglue"
 	"github.com/panther-labs/panther/internal/log_analysis/gluetables"
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/logtypes"
+	"github.com/panther-labs/panther/pkg/stringset"
 )
 
 // CreateTablesMessage is the event that triggers the creation of Glue tables/views for logtypes.
@@ -79,7 +80,7 @@ func HandleCreateTablesMessage(ctx context.Context, msg *CreateTablesMessage) er
 		if err != nil {
 			return err
 		}
-		syncLogTypes = mergeDistinct(msg.LogTypes, deployedLogTypes)
+		syncLogTypes = stringset.Concat(msg.LogTypes, deployedLogTypes)
 	}
 
 	// create/update all tables associated with logTypes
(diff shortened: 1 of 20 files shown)


internal/log_analysis/datacatalog_updater/process/create_tables.go
```diff
@@ -122,3 +137,28 @@ func HandleCreateTablesMessage(ctx context.Context, msg *CreateTablesMessage) er
 
 	return nil
 }
+
+func mergeDistinct(values ...[]string) []string {
```
rleighton, line 141: Should this be ?: 
```
func mergeDistinct(values ...[]string) []string {
	union := make([]string, 0, len(values))
	for _, value := range values {
		union = appendDistinct(union, value)
	}
	return union
}
```

or:
```
func mergeDistinct(values ...[]string) []string {
        uniqueValues := make(map[string]int, len(values))
        i := 0
        for _, value := range values {
		uniqueValues[value] =  i
                 i++
	}
	union := make([]string, len(uniqueValues))
	for  value, index := range uniqueValues {
		union[index] = value
	}
	return union
}
```

alxarch: The idea is to be able to merge arbitrary number of string slices together into a distinct set. I opted to not include a map to avoid allocation. I have used this code before, for this use a map is ok also we don't care about performance really.

alxarch: The first for gets the total size of all 'chunks'

rleighton: Yeah, but summing the characters in all the strings seems wrong (did I read it wrong?) for the size. 

Also shadowing `values` seems also wrong from a read ability pov.

And there are no tests for this code.

alxarch: it's `...[]string` not `...string`

rleighton: Ah, ok.

So how about this:

- add some comments, for example `mergeDistinct takes a set of []string slices and ...`
- fix the loop shadowing
- add some tests
- since you said you used this code elsewhere, perhaps this should be in `pkg/` so this can be re-used

alxarch: Fixed

s0l0ist (approved):

internal/log_analysis/datacatalog_updater/process/create_tables.go
```diff
@@ -67,8 +68,23 @@ func (m CreateTablesMessage) Send(sqsClient sqsiface.SQSAPI, queueURL string) er
...
+		if err != nil {
+			return err
+		}
+		deployedLogTypes, err := gluetables.DeployedLogTypes(ctx, glueClient, availableLogTypes)
+		if err != nil {
+			return err
+		}
+		syncLogTypes = stringset.Concat(msg.LogTypes, deployedLogTypes)
+	}
+
```
kostaspap, line 85: Why do we still need the code in lines 106 to 117?  @alxarch

alxarch: Yes we need it because we have mixed for some reason (?) CreateTables with Sync code, the second call is needed to handle the `CreateTables` properly.

PR #1983, opened 2020-11-07
Add file and line to rule engine exception error messages

rleighton:
## Background

While testing rule errors in Snowflake I noticed that the errors were only of limited use because they included only the exception.
For example, take this rule:
```
def rule(event):
	# Return True to match the log event and trigger an alert.
	if len(event) > 0:
	    1/0
	    
	return False
```
Previously the error message would be:
```
ZeroDivisionError('division by zero')
```

This PR adds the rule file, the line number and the line:
```
division by zero: File "/tmp/rules/AlwaysFail.py", line 4, in rule 1/0
```

## Changes

- add method `error_message` to RuleResult to format the error logged with traceback info
- update tests

## Testing

- mage test:ci
- mage deploy, created a divide by zero rule and query results in Athena


internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -49,17 +49,21 @@ def analyze(self, log_type: str, event: Dict[str, Any]) -> List[EngineResult]:
             self.logger.debug('running rule [%s]', rule.rule_id)
             result = rule.run(event, batch_mode=True)
             if result.errored:
+                short_error_message = repr(result.rule_exception)
+                # TODO(kostaspap): remove error logging once error reporting notification system is in place
+                self.logger.error('failed to run rule %s %s', rule.rule_id, short_error_message)
+                error_type = type(result.rule_exception).__name__
```
rleighton, line 55: The `error_type` moved here from below is the same in Enterprise (this is a backport of testing done there).

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -49,17 +49,21 @@ def analyze(self, log_type: str, event: Dict[str, Any]) -> List[EngineResult]:
             self.logger.debug('running rule [%s]', rule.rule_id)
             result = rule.run(event, batch_mode=True)
             if result.errored:
+                short_error_message = repr(result.rule_exception)
+                # TODO(kostaspap): remove error logging once error reporting notification system is in place
+                self.logger.error('failed to run rule %s %s', rule.rule_id, short_error_message)
```
kostaspap, line 54: You shouldn't log the error here - users will be notified through proper rule error alert

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -59,6 +60,15 @@ class RuleResult:
     alert_context: Optional[str] = None
     alert_context_exception: Optional[Exception] = None
 
+    def error_message(self) -> Optional[str]:
+        """Returns formatted error message with traceback"""
+        if self.rule_exception is not None:
+            trace = traceback.format_tb(self.rule_exception.__traceback__)
+            # we only take last 2 elements of trace which will show the rule file name and line of the error, for example:
+            #    division by zero: File "/tmp/rules/AlwaysFail.py", line 4, in rule 1/0
```
kostaspap, line 68: You should just keep the file name, not the full file path. This is really Panther internal and doesn't provide any value to the users.

internal/log_analysis/rules_engine/tests/test_engine.py
```diff
@@ -127,4 +127,10 @@ def test_analyse_many_rules_one_throws_exception(self) -> None:
             )
         ]
 
+        # since the error message will have a variable file path, test it separately
+        expected_event_matches[1].error_message = result[1].error_message
+        self.assertIsNotNone(result[1].error_message)
+        if result[1].error_message is not None:
+            self.assertTrue(result[1].error_message.endswith('''Exception("Found an issue")'''))
+
```
kostaspap, line 135: See comment above - if you remove the full file path, it should be possible to update the check here as well.

internal/log_analysis/rules_engine/src/engine.py
rleighton: Notice the `TODO`. It says you will do it :)

Is it ok now to remove?

internal/log_analysis/rules_engine/src/rule.py
rleighton: Good idea, will do.

internal/log_analysis/rules_engine/src/engine.py
kostaspap: It was removed on `master`. According to this PR, you are adding it back.

rleighton: Done

internal/log_analysis/rules_engine/src/rule.py
rleighton: Done

internal/log_analysis/rules_engine/tests/test_engine.py
rleighton: Done

kostaspap (approved):

PR #1987, opened 2020-11-09
Backport required changes for CloudTrail parser

alxarch:
## Background

There was a mistake in the previous PR for porting CloudTrail that I only discovered when building the docs in enterprise.
I backported the fixes from that PR and made some minor modification to avoid mistakes when choosing the appropriate 
jsoniter.API for parsers.

Note that part of these changes will be synced again to enterprise to hopefully get the same code across both repos.
Cherry-picking and automation have made some parts of the code diverge significantly causing a lot of confusion. 
This happens because during sync to enterprise new bugs are discovered or extra code needs to be written and it is hard to keep track of all changes that require backporting. I don't think there is a good solution for this TBH.

## Changes

- Backport proper schema handling
- Fix cyclic dependency problems with common module

## Testing

- mage test:go
- enterprise version was e2e tested already


internal/log_analysis/log_processor/common/json.go
```diff
@@ -27,8 +27,12 @@ import (
...
 )
 
-// BuildJSON returns a jsoniter.API instance that is configured to be used for decoding/encoding JSON log events.
-func BuildJSON() jsoniter.API {
+// ConfigForDataLakeWriters returns a jsoniter.API configured to be used for JSON log events written to the data-lake.
+//
+// WARNING: This is meant to be used for encoding ONLY FOR WRITING TO S3 Processed logs.
+//          For all other uses please use pantherlog.ConfigJSON()
+//          You should probably not use it.
+func ConfigForDataLakeWriters() jsoniter.API {
```
giorgosp, line 35: Is it is easy to move it outside the `common` package, I'd suggest to do so. `common`, due its name, is imported everywhere, which may lead to circular imports.

giorgosp (approved):

internal/log_analysis/log_processor/parsers/awslogs/indicators.go
```diff
@@ -28,64 +27,14 @@ import (
...
-		w.WriteValues(pantherlog.FieldAWSAccountID, input)
-	}
-}
-
-func ScanInstanceID(w pantherlog.ValueWriter, input string) {
-	if strings.HasPrefix(input, "i-") {
-		w.WriteValues(pantherlog.FieldAWSInstanceID, input)
-	}
+		pantherlog.FieldAWSInstanceID,
+		pantherlog.FieldAWSTag,
```
kostaspap, line 36: Should we have a `awslogs.Indicators()`? So if we add another one in the future we don't miss it

alxarch: We have moved the definitions inside `pantherlog` to not create a cyclic dependency. For now we should stick to defining all indicator scanners inside `pantherlog` package  'statically'.

PR #1988, opened 2020-11-09
fix: reverse mini-chart for TimeSeries charts

alexmylonas:
## Background

The mini-graph below all charts are displayed in reverse. The cause of this is that `seriesData` are coming from backend in descending timestamp order. While that didn't affect charts cause they are ordered on charts based on timestamp, echarts dont seem to apply the same logic on the mini-chart.

This PR fixes the issue (check before & after screenshot below), the alternative would be to change the API so that it returns data in ascending order.

### Before Screenshot
Screenshot 2020-11-09 at 7 05 28 PM
### After Screenshot
Screenshot 2020-11-09 at 7 04 58 PM

## Changes

- Reversed data before passing it to echart

## Testing

- Locally


3nvi (approved):
LGTM!

PR #1990, opened 2020-11-09
fix: reset pagination upon filtering

vorillaz:
## Background

Fixes the pagination for pagination upon filtering changing.

## Changes

- Reset pagination params upon filters change

## Testing

- `npm run test`


3nvi (approved):
Amazing. Thanks @vorillaz

PR #1992, opened 2020-11-09
Modifications for cloudsecurity feature

alxarch:
## Changes

- Use `jsoniter.RawMessage` for Resource
- Parsers using new pantherlog engine

## Testing

- unit tests


nhakmiller:
In the morning, can you post some examples of what the output of these changes look like in S3? Would be helpful to evaluate it's working as intended.

I'm specifically referring to the files in the data-inputs staging bucket, not the processed data bucket as I understand you  still have some work to do on the log processor.

internal/compliance/datalake_forwarder/forwarder/compliance.go
```diff
@@ -75,30 +79,36 @@ func (sh StreamHandler) processComplianceSnapshot(record events.DynamoDBEventRec
...
+		if err != nil {
+			return nil, err
+		}
+		attr := dynamodb.AttributeValue{}
+		if err := jsoniter.Unmarshal(raw, &attr); err != nil {
+			return nil, err
+		}
+		out[key] = &attr
+	}
+	return out, nil
```
nhakmiller, line 113: For compliance snapshots, previously we were populating the fields of a struct with the struct values of a map by converting those values to the type we need.

With this change, we are going to take a map, marshal each of its struct values into a string, unmarshal each of those strings back into a different struct, insert those structs into a new map, then unmarshal that map into the struct we want? Am I understanding this change correctly?

Does this change the output? What is the benefit to adding this complexity? I'd generally be opposed to this solution that seems more complex to me unless there is a very large benefit to it.

internal/compliance/datalake_forwarder/forwarder/compliance.go
```diff
@@ -75,30 +79,36 @@ func (sh StreamHandler) processComplianceSnapshot(record events.DynamoDBEventRec
...
-		IntegrationID:  image["integrationId"].String(),
-		LastUpdated:    image["lastUpdated"].String(),
-		PolicyID:       image["policyId"].String(),
-		PolicySeverity: image["policySeverity"].String(),
-		ResourceID:     image["resourceId"].String(),
-		ResourceType:   image["resourceType"].String(),
-		Status:         image["status"].String(),
-		Suppressed:     image["suppressed"].Boolean(),
-	}, nil
+func unmarshalMap(attributes map[string]events.DynamoDBAttributeValue, out interface{}) error {
```
nhakmiller, line 92: It seems this `unmarshalMap` function is going to be used by both the resources and compliance forwarder, if so we should move it out of `compliance.go`.

nhakmiller (commented):
A lot of changes here, I don't feel comfortable approving without some example files from both the input data bucket and the processed data bucket (once the log processor side is working).

internal/compliance/datalake_forwarder/forwarder/compliance.go
alxarch: The main benefit is that we don't handle the marshal complexities ourselves. I simplified the conversion logic further to demonstrate this.

alxarch:
> A lot of changes here, I don't feel comfortable approving without some example files from both the input data bucket and the processed data bucket (once the log processor side is working).
If you can provide me with a sample input, I can see if I can run the pantherlog tool on it.

nhakmiller:
> > A lot of changes here, I don't feel comfortable approving without some example files from both the input data bucket and the processed data bucket (once the log processor side is working).
> 
> If you can provide me with a sample input, I can see if I can run the pantherlog tool on it.

Not sure what you mean, what does the `pantherlog` tool do?

I'm asking you to deploy this branch into your dev account and grab some of the files it generates in the `panther-bootsrap-*-input-data-*` bucket and the `panther-bootstrap-*-processed-data-*` bucket so we can see what they look like. Or just a few lines from those files, similar to how I did in this PR https://github.com/panther-labs/panther-enterprise/pull/948.

I just want to validate it looks how we're expecting it to look.

internal/compliance/datalake_forwarder/forwarder/compliance.go
alxarch: ddb attribute values have many complex subtypes that are should all be mapped to/from JSON. `dynamodbattribute` package handles this transparently for us.

nhakmiller (approved):
Seems reasonable enough to me. @kostaspap, @alxarch had asked me to let you handle merging this so I'm going hands off now. More than happy to review/assist with anything else that comes up though.

kostaspap (approved):

PR #1993, opened 2020-11-09
File detection

giorgosp:
## Background

Sometimes we mis-detect gzipped log files as ms-fontobject.

## Changes

Change the detection code to check directly for the gzip signature instead of relying on the `http.DetectContentType` function. That function checks for many content types and produces a false positive in some cases.

## Testing

- Unit


giorgosp:
@s0l0ist forgot to put the `sync-to-release` label on the OSS PR. Mind approving?

s0l0ist (approved):

PR #1995, opened 2020-11-10
Reverting reading change

kostaspap:
## Background

Fixing bug in log processor

## Changes

- Reverting to use bufio Reader instead of scanner

## Testing

- mage test:ci

### DIFF
--- a/internal/log_analysis/log_processor/processor/processor.go
+++ b/internal/log_analysis/log_processor/processor/processor.go
@@ -171,7 +171,6 @@ func (p *Processor) run(outputChan chan<- *parsers.Result) error {
 			}
 			break
 		}
-
 	}
 	if err != nil {
 		err = errors.Wrap(err, "failed to read log line")
--- a/internal/log_analysis/log_processor/processor/processor.go
+++ b/internal/log_analysis/log_processor/processor/processor.go
@@ -20,6 +20,7 @@ package processor
 
 import (
 	"bufio"
+	"io"
 	"sync"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


alxarch (approved):
LGTM

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -156,12 +157,22 @@ func NewFactory(resolver logtypes.Resolver) Factory {
 
 // processStream reads the data from an S3 the dataStream, parses it and writes events to the output channel
 func (p *Processor) run(outputChan chan<- *parsers.Result) error {
-	stream := bufio.NewScanner(p.input.Reader)
-	for stream.Scan() {
-		line := stream.Text()
-		p.processLogLine(line, outputChan)
+	var err error
+	// DO NOT use bufio.NewScanner
+	// bufio.NewScanner has a max buffer size which we might hit as we read bigger log lines
```
giorgosp, line 162: You may also say that bufio.Reader `ReadString()` auto-grows the buffer, with it's pros and cons.

giorgosp (approved):

internal/log_analysis/log_processor/processor/processor.go
alxarch: Since the readers' lifetime is as long as the log file is being processed this should not be a problem. As long as the reader gets GCd at the end of each file.

alxarch: These issues will be resolved when we move to a streaming log reader interface

Issue #1689, opened 2020-10-07
Remove remediation-api gateway


austinbyers:
## Description
Remove the remediation API gateway (the Lambda handler will remain intact).

## Related Services
All consumers will need to be changed to invoke the remediation-api Lambda function directly:

* AppSync
* Cloud security backend

PR #2003, opened 2020-11-10
Remove remediation-api gateway

austinbyers:
## Background

Removing internal API gateways Part 3/4 - the remediation-api

Closes: #1689 

## Changes

- Move swagger models to native Go
- Update AppSync to use Lambda data source instead of gateway
- Update `alert-processor` to invoke `remediation-api` Lambda directly instead of using the gateway
- Remove swagger and gateway
- Change remediation `apihandlers` package name to `handlers` (to match the folder)
- Remove special checks for remediation function not existing - we deploy that function, not the customer

## Testing

- Deployed, was able to `listRemediations` by invoking the `panther-remediation-api` directly
- The policy creation page lists available remediations, as before
- Testing successful resource remediation (in progress)


nhakmiller (approved):
Burn the gateways! I gave it a decent look but the real proof will be in the release testing. Seems in order to me.

austinbyers:
Manual remediation (enabling S3 bucket encryption) worked successfully!

PR #2004, opened 2020-11-10
udm support in rules engine 

### LINKED ISSUE #1812: Access `udm` method in Rules Engine
### Description

Support field mappings / user defined schema in rules engine. 

### Related Services
WIP
- This issue is strictly for the object instantiation / `udm` aspect of enabling field standardization  
- There is a related issue to upload logtype schemas

### Designs
WIP

### Acceptance Criteria
 
- Users can use `udm()` method in rules logic to access fields by standard name

lindsey-w:
## Background

In order to support user-defined data models, we need to implement the `udm` method ("unified data model") to convert unified field name -> log type's field name.  This change does this by: 

1. Importing all enabled `DataModels` and setting them up similar to how rules are imported / setup, using a `log_type_to_data_model` dictionary
2. Converting `event` dictionary to an `EnrichedEvent`. `EnrichedEvent` is an implementation of the abstract `Mapping` class (similar to `dict`)
3. Implementing `udm` method in `EnrichedEvent` to perform either a JSONPath lookup or run a method, based on what the DataModel defines.  

Closes #1812 

## Changes

- new `DataModel` that sets up the unified data model field -> log type field mappings
- new `EnrichedEvent` that implements the `udm` method and has access to DataModels
- unit tests for DataModels and EnrichedEvents

## Testing

- `mage test:ci`


internal/log_analysis/rules_engine/src/__init__.py
```diff
@@ -28,7 +29,7 @@ class EngineResult:
     log_type: str
     dedup: str
     dedup_period_mins: int
-    event: Dict[str, Any]
+    event: Mapping
```
lindsey-w, line 32: Because events can now be either a strict `dict` or an `EnrichedEvent` I modified the various method declarations to accept `Mapping` versus `dict[str, Any]` This prevents the various type errors and reassign errors

internal/log_analysis/rules_engine/src/data_model.py
```diff
@@ -0,0 +1,123 @@
...
+        # data models contains logtype to schema definitions
+        if not ('id' in config) or not isinstance(config['id'], str):
+            raise AssertionError('Field "id" of type str is required field')
+        self.data_model_id = config['id']
+
+        # mappings are required
+        if not ('mappings' in config) and not isinstance(config['mappings'], list):
+            raise AssertionError('Field "mappings" of type list')
+        self.fields: Dict[str, Fields] = dict()  # setup field mappings
+        self.methods: Dict[str, Callable] = dict()  # setup method mappings
```
lindsey-w, line 53: I separated this into two distinct dictionaries as to not mix types within it.  I thought this makes it more clear in the `udm` logic exactly what is going on. Since key lookups are O(1), figured it wouldn't add much overhead to check (max of) 2 dictionaries for the existence of a key

internal/log_analysis/rules_engine/src/enriched_event.py
```diff
@@ -0,0 +1,73 @@
...
+                        'JSONPath [{}] in DataModel [{}], matched multiple fields.'.format(json_path, self.data_model.data_model_id)
+                    )
+        if key in self.data_model.methods.keys():
+            # we are dealing with method
+            method = self.data_model.methods.get(key)
+            if callable(method):
+                # return self.get(method(self._data))
+                return method(self._data)
+        # no matches, return None by default
+        return None
```
lindsey-w, line 73: I wasn't sure if this should raise an error or just return `None` - I went with `None` to prevent errors here.  I figured if a user tries to use the value returned and it is `None` there will be an error in the rule logic anyway.

internal/log_analysis/rules_engine/tests/test_enriched_event.py
```diff
@@ -0,0 +1,123 @@
...
+    def test_udm_multiple_matches(self) -> None:
+        exception = False
+        event = {'dst': {'ip': '1.1.1.1', 'port': '2222'}}
+        data_model = DataModel(
+            {
+                'body': 'def get_source_ip(event):\n\treturn "1.2.3.4"',
+                'versionId': 'version',
+                'mappings': [{
+                    'name': 'destination_ip',
+                    'field': '$.dst.*'
```
kostaspap, line 110: Interesting... Why restrict the users and fail here? We could just return the `{'ip': '1.1.1.1', 'port': '2222'}` right? 

`event.udm('dest').get('ip')`, `event.udm('dest').get('port')` might be easier than forcing them to specify two distinct mappings, one for IP and one for port. 

I am thinking it might actually be easier for users for such cases to use UDM in this way rather than having so specify multiple mappings like

internal/log_analysis/rules_engine/src/enriched_event.py
```diff
@@ -0,0 +1,73 @@
...
+            return self.get(key)
+        # access values via standardized fields
+        if key in self.data_model.fields.keys():
+            # we are dealing with a jsonpath
+            json_path = self.data_model.fields.get(key)
+            if json_path:
+                matches = json_path.find(self._data)
+                if len(matches) == 1:
+                    return matches[0].value
+                if len(matches) > 1:
```
kostaspap, line 60: I am thinking that we might want to return all matches instead of failing? I have put a similar note below, in the test code.

internal/log_analysis/rules_engine/src/data_model.py
```diff
@@ -0,0 +1,123 @@
...
+        self.version = config['versionId']
+
+        self._store_data_models()
+        self._module = self._import_data_model_as_module()
+        self._extract_mappings(config['mappings'])
+
+    def _extract_mappings(self, source_mappings: List[Dict[str, str]]) -> None:
+        name = 'name'
+        field = 'field'
+        method = 'method'
```
kostaspap, line 73: nit: Could be constants

internal/log_analysis/rules_engine/src/data_model.py
```diff
@@ -0,0 +1,123 @@
...
+        for mapping in source_mappings:
+            if name not in mapping.keys():
+                self.logger.error('DataModel [%s] is missing required field: [%s]', self.data_model_id, name)
```
kostaspap, line 76: Seems like we should be throwing an exception here? From what I understand this can happen if the DataModel is invalid/misconfigured. 
I would also advice against logging at error level inside here - the caller can handle this appropriately (it already does log an error actually in case of an exception)

internal/log_analysis/rules_engine/src/data_model.py
```diff
@@ -0,0 +1,123 @@
...
+def _data_model_id_to_path(data_model_id: str) -> str:
+    """Method returns the file path where the rule will be stored"""
+    safe_id = ''.join(x if _allowed_char(x) else '_' for x in data_model_id)
+    path = os.path.join(_DATAMODEL_FOLDER, safe_id + '.py')
+    return path
+
+
+def _allowed_char(char: str) -> bool:
+    """Return true if the character is part of a valid rule ID."""
+    return char.isalnum() or char in {' ', '-', '.'}
```
kostaspap, line 123: Rules have similar methods. Maybe they can be extracted to some common library?

internal/log_analysis/rules_engine/tests/test_enriched_event.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def test_udm_field(self) -> None:
+        event = {'dst_ip': '1.1.1.1', 'dst_port': '2222'}
+        data_model = DataModel(
+            {
+                'body': 'def get_source_ip(event):\n\treturn "1.2.3.4"',
+                'versionId': 'version',
+                'mappings': [{
+                    'name': 'destination_ip',
+                    'field': 'dst_ip'
```
kostaspap, line 71: Should we just ask users to use JSON path everywhere? e.g. ask users to declare this as `$.dst_ip`. 
In my experience, when people are presented with 2 very similar approaches, confusion follows and they end up making mistakes e.g. trying to use JSON path without specifying the `$.` element in front. Every new option means just another thing users have to remember.

internal/log_analysis/rules_engine/tests/test_enriched_event.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def test_udm_json_path(self) -> None:
+        event = {'dst': {'ip': '1.1.1.1', 'port': '2222'}}
+        data_model = DataModel(
+            {
+                'body': 'def get_source_ip(event):\n\treturn "1.2.3.4"',
+                'versionId': 'version',
+                'mappings': [{
+                    'name': 'destination_ip',
+                    'field': '$.dst.ip'
```
kostaspap, line 90: How do we handle cases were an event field has a dot `.` in the name? Some of [Gitlab logs](https://github.com/panther-labs/panther/blob/1ed0e7eef9705b8bd6c6f01171564ba8d41846c9/internal/log_analysis/log_processor/parsers/gitlablogs/api.go#L52) have such fields. Could you add a test for it?

kostaspap (commented):
Love this functionality. Great work!

kostaspap:
Could you also do a benchmark? What is the comparison of the users doing `event.get('field')` vs `event.udm(field)` vs `event.udm(method)`?

internal/log_analysis/rules_engine/src/data_model.py
nhakmiller: I agree, I think any issue in parsing a data model should just raise an exception to immediately toss the whole data model out. I don't like the idea of partially loaded data models.

nhakmiller (commented):
This looks great! Super excited for this feature.

internal/log_analysis/rules_engine/tests/test_enriched_event.py
kostaspap: Fair point.

internal/log_analysis/rules_engine/src/data_model.py
lindsey-w: That makes sense - I'll remove the logging and throw an exception here and in the final `else` block

internal/log_analysis/rules_engine/tests/test_enriched_event.py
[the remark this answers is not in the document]
lindsey-w: Good idea 🤔  I do like `path` - as that could mean `json_path` or a simple `field` reference.

deployments/panther_config.yml
```diff
@@ -60,6 +60,7 @@ Infra:
   # Contributors, take note: the default layer (packaged with Panther releases) bundled in the
   # bootstrap-gateway stack must fit within the ~50MB SAR package limit. boto3, for example, is too big
   PipLayer:
+    - jsonpath-ng==1.5.2
```
austinbyers, line 63: Be sure to update the [relevant documentation](https://docs.runpanther.io/user-guide/analysis/run-time) when this merges to enterprise

internal/log_analysis/rules_engine/src/data_model.py
```diff
@@ -0,0 +1,123 @@
...
+        Args:
+            config: Dictionary that should have the folllowing keys:
+                id: unique data model id
+                mappings: list of log type fields to standard field conversions
+                version: the version of the data model
+                (Optional) body: python body associated with the mappings
+        """
+        self.logger = get_logger()
+        # data models contains logtype to schema definitions
+        if not ('id' in config) or not isinstance(config['id'], str):
```
austinbyers, line 45: Nit: more simply:

```python
if not isinstance(config.get('id'), str):
```

If it doesn't exist, `.get` will return None. Similarly, line 50 and 62

internal/log_analysis/rules_engine/src/data_model.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def _extract_mappings(self, source_mappings: List[Dict[str, str]]) -> None:
+        name = 'name'
+        field = 'field'
+        method = 'method'
+        for mapping in source_mappings:
+            if name not in mapping.keys():
+                self.logger.error('DataModel [%s] is missing required field: [%s]', self.data_model_id, name)
+                continue
+            if field in mapping.keys():
```
austinbyers, line 78: Agreed with Nick: you don't need `.keys()` here. This should be:

```python
if field in mapping:
```

You probably don't need `.keys()` anywhere

requirements.txt
```diff
@@ -30,6 +30,7 @@ Jinja2==2.11.2
 jinja2-time==0.2.0
 jmespath==0.9.5
 jsonpatch==1.26
+jsonpath-ng==1.5.2
```
austinbyers, line 33: You'll also need to add this to `requirements-top-level.txt` (see the comment at the top of that file to understand how we use the two requirements files)

internal/log_analysis/rules_engine/tests/test_enriched_event.py
lindsey-w: I think if they were to use a different json_path it would work this way - e.g. in this example rather than `$.dst.*` using `$.dst` would return the full dictionary and the `.get()` access would work as you expect. The `matches` array would still only be of size == 1 in this case.  So we put the responsibility on the user to define a proper jsonpath. Otherwise, I think it might get complicated dealing with how the jsonpath_ng is returning the data and trying to guess what the user is wanting to return?

internal/log_analysis/rules_engine/tests/test_enriched_event.py
lindsey-w: I didn't think about this! I'm adding a test case. It looks like these will have to be quoted in the mapping definition itself.

lindsey-w:
> Could you also do a benchmark? What is the performance comparison of the users doing `event.get('field')` vs `event.udm(field)` vs `event.udm(method)`?

The following numbers represent the total time it takes to run a rule against 10,000 or 100,000 events. Where the rule was updated to either to (1) dictionary access via `.get()` (2) json path access via `.udm('path')` (3) run a method via `.udm('methodname')`.  The rule was the exact same except for this change.  The events were the same for each. 

Interestingly, for the sample event set of 100,000 - I consistently get something similar to this (where the order in terms of longest to shortest time is the same each test):
```
total time for dictionary access: 3.9716219902038574 
total time for udm path access: 3.391889810562134 
total time for udm method access: 2.36067795753479 
```

But for the sample event set of 10,000 - I consistently get numbers like these (where the order in terms of longest to shortest time is the same each test):
```
total time for dictionary access: 0.2268228530883789 
total time for udm path access: 0.2922179698944092 
total time for udm method access: 0.25775814056396484 
```

Seems like dictionary access swaps places with the path access on smaller set of events 🤔 

Also - this is for a really simple udm method - one that just returns a string.

internal/log_analysis/rules_engine/src/data_model.py
lindsey-w: 👍🏻  Pulled these out into a util

internal/log_analysis/rules_engine/tests/test_enriched_event.py
kostaspap: Got it. Yeah, makes sense. When reading the test I didn't understand exactly the test condition.

kostaspap:
> > Could you also do a benchmark? What is the performance comparison of the users doing `event.get('field')` vs `event.udm(field)` vs `event.udm(method)`?
> 
> The following numbers represent the total time it takes to run a rule against 10,000 or 100,000 events. Where the rule was updated to either to (1) dictionary access via `.get()` (2) json path access via `.udm('path')` (3) run a method via `.udm('methodname')`. The rule was the exact same except for this change. The events were the same for each.
> 
> Interestingly, for the sample event set of 100,000 - I consistently get something similar to this (where the order in terms of longest to shortest time is the same each test):
> 
> ```
> total time for dictionary access: 3.9716219902038574 
> total time for udm path access: 3.391889810562134 
> total time for udm method access: 2.36067795753479 
> ```
> 
> But for the sample event set of 10,000 - I consistently get numbers like these (where the order in terms of longest to shortest time is the same each test):
> 
> ```
> total time for dictionary access: 0.2268228530883789 
> total time for udm path access: 0.2922179698944092 
> total time for udm method access: 0.25775814056396484 
> ```
> 
> Seems like dictionary access swaps places with the path access on smaller set of events 🤔
> 
> Also - this is for a really simple udm method - one that just returns a string.

Huh, this is an interesting one. 

I wanted to make sure that the `path` approach doesn't add significant overhead. It seems that's not the case, numbers are very similar.

kostaspap (approved):
Great work! This is a killer feature!!

deployments/panther_config.yml
lindsey-w: Noted - will do

PR #2006, opened 2020-11-11
Update tables only if integration adds new log types

kostaspap:
## Background

Update tables only if integration adds new log types

## Changes

- Update tables only if integration adds new log types

## Testing

- mage test:ci
- integration tests


giorgosp (approved):

internal/core/source_api/api/update_integration.go
```diff
@@ -209,21 +209,48 @@ func getItem(integrationID string) (*ddb.Integration, error) {
...
+		LogTypes: newLogTypes,
 	}
 	err := m.Send(sqsClient, env.DataCatalogUpdaterQueueURL)
 	if err != nil {
 		return errors.Wrap(err, "failed to create Glue tables")
 	}
 	return nil
 }
+
+func slicesEqual(left, right []string) bool {
```
rleighton, line 238: How about moving this to `pkg/stringset' ?

internal/core/source_api/api/update_integration.go
```diff
@@ -77,16 +77,16 @@ func (api API) UpdateIntegrationSettings(input *models.UpdateIntegrationSettings
 		}
 	}
 
+	if err := updateTables(existingIntegrationItem, input); err != nil {
```
rleighton, line 80: Why move this? Makes the diff bigger.

internal/core/source_api/api/update_integration.go
```diff
@@ -209,21 +209,48 @@ func getItem(integrationID string) (*ddb.Integration, error) {
 	return item, nil
 }
 
-func updateTables(integrationType string, input *models.UpdateIntegrationSettingsInput) error {
-	var logtypes []string
-	switch integrationType {
+func updateTables(item *ddb.Integration, input *models.UpdateIntegrationSettingsInput) error {
```
rleighton, line 212: Unit test for this to confirm behavior?

internal/core/source_api/api/update_integration.go
kostaspap: It needs to. The `normalizeIntegration` modifies the `existingIntegrationitem` - replaces old logtypes with what's inside the input. 

Code could take some further cleanup tbh just didn't want to do bigger changes.

internal/core/source_api/api/update_integration.go
kostaspap: Done

internal/core/source_api/api/update_integration.go
kostaspap: Done. Fixed a couple of tests on the way.

pkg/stringset/stringset.go
```diff
@@ -63,3 +63,25 @@ loopValues:
 	}
 	return dst
 }
+
+// Returns True if the string slices contain the same elements
+// False otherwise
+func SlicesEqual(left, right []string) bool {
```
kostaspap, line 69: Thought of renaming the file to `stringutils` or something, but I thought it could make the PR more difficult to read

rleighton (approved):

internal/core/source_api/api/update_integration.go
alxarch: No, why? Stringset is supposed to be only about distinct sets of strings, not any thing involving string slices.

pkg/stringset/stringset.go
alxarch: Yes, exactly the reason it is not named utils. It is meant to be strictly about distinct sets of string values.

internal/core/source_api/api/update_integration.go
kostaspap: @alxarch should I create a new package `stringslices`? Or maybe just rename the package to `stringutils` or something (see [comment](https://github.com/panther-labs/panther/pull/2006#discussion_r521882904))

alxarch: There is no reason moving this in another package in the first place. Moving any useful little function to another package just fragments the code organisation. My suggestion is to just keep it there and use the rule of 3 to decide if it is worthy of being shared.

pkg/stringset/stringset.go
alxarch: Are you certain that the method works in all possible diff scenarios (add, remove, reorder)? 
I would sort both slices first and go over comparing each value after. 
It is less performant, but does not assume the slices contain distinct items so it is independent from the arguments.
I think this function assumes each of the slices contain unique items.

alxarch (changes requested):
Please add unit tests that verify the comparison function for all diff scenarios (add, remove, reorder) and make sure that user input is guaranteed to hold unique values when we check with the values from the existing record.

pkg/stringset/stringset.go
kostaspap: Right that is the assumption since the LogTypes are unique (we store them as stringset in DDB) so I went for a more performant. However performance is not that important here, I can just sort them and do a deep equals for the slices. 
Better safeguard it against future use.

internal/core/source_api/api/update_integration.go
kostaspap: Fine by me. Moving to `update_integration.go`. Note though that most of the functions in `stringset.go` are also used only in single place (and most of them not even used outside the package).

alxarch: Yes, I agree. Strigset was moved to pass review in urgency. I think there are places where I have had to write the exact same functions again. I need to find where and replace them with `stringset` calls.

alxarch (approved):
LGTM

rleighton:
Probably should run `mage test:integration`

s0l0ist:
Passes integration tests

s0l0ist:
Holding for v1.12.1

internal/core/source_api/api/update_integration.go
```diff
@@ -209,21 +212,40 @@ func getItem(integrationID string) (*ddb.Integration, error) {
...
-		logtypes = input.LogTypes
+		existingLogTypes = item.LogTypes
+		newLogTypes = input.LogTypes
 	case models.IntegrationTypeSqs:
-		logtypes = input.SqsConfig.LogTypes
+		existingLogTypes = item.SqsConfig.LogTypes
+		newLogTypes = input.SqsConfig.LogTypes
+	}
+
+	// If log types haven't been updated, return
```
giorgosp, line 226: If the user removes a log type, the code will still send the CreateTablesMessage. Is this expected behavior? If yes, I suggest we add a comment. This way, someone that suspects about this bug won't lose time trying to reproduce/fix/communicate it.

I'd also add this test case to the tests:
```
{
			first:  []string{"a", "b", "c"},
			second: []string{"a", "b"},
			expect: false,
		},
```

Also, `slicesContainSameElements()` could be simplified to (pseudocode):
```
shouldUpdateTables() {
    for lt in newLogTypes
          if lt not in existingLogTypes
                  return False
    return True
}
```

kostaspap: This is a good point. I wrote the PR initially as a quick fix to an issue we were experiencing in a very specific scenario (we were triggering the "Create table" workflow whenever someone was updating an OAuth2 token). But you are right, we don't want to trigger the "Create table" if the just removed a log type. 

Updated.

giorgosp (approved):

internal/core/source_api/api/update_integration_test.go
```diff
@@ -161,3 +212,64 @@ func TestUpdateIntegrationLastScanEnd(t *testing.T) {
...
+			old:    []string{"Log.A", "Log.B", "Log.A"},
+			new:    []string{"Log.A", "Log.B"},
+			expect: false,
+		},
+		{
+			old:    []string{"Log.A", "Log.B"},
+			new:    []string{"Log.A", "Log.B"},
+			expect: false,
+		},
+		{
```
rleighton, line 253: Cab you add a case where the lengths are the same but true:
```
{
			old:    []string{"Log.A", "Log.B"},
			new:    []string{"Log.C", "Log.D"},
			expect: true,
},
```

internal/core/source_api/api/check_integration.go
```diff
@@ -247,7 +247,7 @@ func checkSqsQueueHealth(input *models.CheckIntegrationInput) *models.SourceInte
 	// is performed before the SQS queue is created.
 	if len(input.SqsConfig.QueueURL) == 0 {
 		health.SqsStatus.Healthy = true
-		health.SqsStatus.Message = "Queue does not exist yet (first time setup)."
+		health.SqsStatus.Message = "Queue does not exist yet (old time setup)."
```
rleighton, line 250: Old?

rleighton (approved):

rleighton:
reminder: @s0l0ist,  I really do not want to give Figma or Boxycharm 1.12 w/out this added.

internal/core/source_api/api/check_integration.go
kostaspap: ah, Goland renamed more than it should

internal/core/source_api/api/update_integration_test.go
kostaspap: Added

internal/core/source_api/api/check_integration.go
kostaspap: Fixed

PR #2010, opened 2020-11-12
add displayName to data models

lindsey-w:
## Background

Add DisplayName to DataModel, similar to Rules/Policies. This allows users to change the default name displayed with various DataModels. 

## Changes

- Added new `DisplayName` field to `UpdateDataModel` object

## Testing

- `mage test:ci`


nhakmiller (approved):
@austinbyers just FYI more changes to  the `analysis-api`.

PR #2013, opened 2020-11-13
Extract domains from CT sourceIPAddress field

kostaspap:
## Background

CloudTrail sometimes includes a domain in the `sourceIPAddress ` field. Extract that to `p_any_domain_names`.

## Changes

- Extracting domain name from the `sourceIPAddress` field.

## Testing

- mage test:ci


internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
```diff
@@ -51,7 +51,7 @@ type CloudTrail struct {
...
 	ServiceEventDetails pantherlog.RawMessage   `json:"serviceEventDetails" description:"Identifies the service event, including what triggered the event and the result."`
 	SharedEventID       pantherlog.String       `json:"sharedEventID" description:"GUID generated by CloudTrail to uniquely identify CloudTrail events from the same AWS action that is sent to different AWS accounts."`
-	SourceIPAddress     pantherlog.String       `json:"sourceIPAddress" panther:"ip" validate:"required" description:"The IP address that the request was made from. For actions that originate from the service console, the address reported is for the underlying customer resource, not the console web server. For services in AWS, only the DNS name is displayed."`
+	SourceIPAddress     pantherlog.String       `json:"sourceIPAddress" panther:"hostname" validate:"required" description:"The IP address that the request was made from. For actions that originate from the service console, the address reported is for the underlying customer resource, not the console web server. For services in AWS, only the DNS name is displayed."`
```
giorgosp, line 54: Will it be added to `p_any_ip_addresses` if it contains an IP?
Also check the `extractIndicators()` below in the file. Seems that we also check there for these indicators in addition to `panther` tags? I'm not familiar with this system but this is confusing.
@alxarch please shed some light :)

kostaspap: > Will it be added to p_any_ip_addresses if it contains an IP?

Yes. This is verified by the other unit test. See [here](https://github.com/panther-labs/panther/blob/db4cfb577dcdee79974127821de235978b691731/internal/log_analysis/log_processor/parsers/awslogs/testdata/cloudtrail_tests.yml#L127) and [here](https://github.com/panther-labs/panther/blob/db4cfb577dcdee79974127821de235978b691731/internal/log_analysis/log_processor/parsers/awslogs/testdata/cloudtrail_tests.yml#L197).

> Also check the extractIndicators() below in the file. Seems that we also check there for these indicators in addition to panther tags? 

Yeah, the logic there is  searchin for more indicators that are nested inside the `additionalEventData`, `requestParameters`, `responseElements`, `serviceEventDetails`. The pantherlog tags work only when the value of a field is `ip`, `host`, etc. It doesn't if we have to search inside the fields. For that we use the scanners that you see

giorgosp (approved):

PR #2020, opened 2020-11-13
Keep only 1 required field for each Cloudflare log type

giorgosp:
## Background

Cloudflare allows users to select which fields should be included
in the events, meaning we can't arbitrarily tag fields as required.

## Changes

Keeps only 1 timestamp field as required for every Cloudflare log type.

## Testing

- unit tests

### DIFF
--- a/internal/log_analysis/log_processor/parsers/cloudflarelogs/cloudflarelogs.go
+++ b/internal/log_analysis/log_processor/parsers/cloudflarelogs/cloudflarelogs.go
@@ -37,24 +37,27 @@ var logTypes = func() logtypes.Group {
 	))
 	return logtypes.Must("Cloudflare",
 		logtypes.ConfigJSON{
-			Name:         "Cloudflare.HttpRequest",
-			Description:  `Cloudflare http request logs`,
+			Name: "Cloudflare.HttpRequest",
+			// nolint:lll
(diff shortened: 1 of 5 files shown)


internal/log_analysis/log_processor/parsers/cloudflarelogs/cloudflarelogs.go
```diff
@@ -38,23 +38,23 @@ var logTypes = func() logtypes.Group {
 	return logtypes.Must("Cloudflare",
 		logtypes.ConfigJSON{
 			Name:         "Cloudflare.HttpRequest",
-			Description:  `Cloudflare http request logs`,
+			Description:  `Cloudflare http request logs. When selecting event fields on the Cloudflare UI, make sure you include the "EdgeStartTimestamp" field as it is required by Panther.`,
```
kostaspap, line 41: 👍

internal/log_analysis/log_processor/parsers/cloudflarelogs/testdata/cloudflare_tests.yml
```diff
@@ -229,6 +242,20 @@ result: |
   	"p_any_trace_ids": ["parent-ray-id", "ray-id"]
   }
 ---
+name: httprequest required fields only
```
kostaspap, line 245: 👍  Very useful logs 😆

kostaspap (approved):

PR #2026, opened 2020-11-13
bug policy scan

lindsey-w:
## Background

Catch 404/NoSuchEntity errors: log at debugging level and return nil rather than raising exception. Issue described [here](https://github.com/panther-labs/panther-enterprise/issues/1187). 

## Changes

- Check for particular error
- new test cases

## Testing

- `mage test:go`


internal/compliance/snapshot_poller/pollers/aws/iam_role.go
```diff
@@ -112,6 +112,11 @@ func iamRoleIterator(page *iam.ListRolesOutput, roles *[]*iam.Role, marker **str
 func getRolePolicy(iamSvc iamiface.IAMAPI, roleName *string, policyName *string) (*string, error) {
 	policy, err := iamSvc.GetRolePolicy(&iam.GetRolePolicyInput{RoleName: roleName, PolicyName: policyName})
 	if err != nil {
+		var awsError awserr.Error
+		if errors.As(err, &awsError) && awsError.Code() == "NoSuchEntity" {
+			zap.L().Debug("role policy could not be found", zap.String("policyName", *policyName))
+			return nil, nil
+		}
```
lindsey-w, line 119: I logged at debug level and return nil rather than raising the exception. I believe (?) this means it won't retry, so I'm not sure if this is what we would like to do?

nhakmiller:
While this is good, the linked issue was related to scanning Lambda functions and the policies attached to them. Could you add this there as well?

internal/compliance/snapshot_poller/pollers/aws/iam_role.go
nhakmiller: Right, it won't retry. Which is the correct behavior I think.

lindsey-w:
> While this is good, the linked issue was related to scanning Lambda functions and the policies attached to them. Could you add this there as well?

I guess I'm not seeing those errors - I'm only seeing instances of `EntityNotFound`.  

Looking [here](https://github.com/panther-labs/panther/blob/master/internal/compliance/snapshot_poller/pollers/aws/lambda_function.go#L149) it does look like the lambda `getPolicy` does log debug message and return nil for missing policies.  Is there another place I should look (for the logs or in the code?)

nhakmiller:
> I guess I'm not seeing those errors - I'm only seeing instances of `NoSuchEntity`.
> 
> Looking [here](https://github.com/panther-labs/panther/blob/master/internal/compliance/snapshot_poller/pollers/aws/lambda_function.go#L149) it does look like the lambda `getPolicy` does log debug message and return nil for missing policies. Is there another place I should look (for the logs or in the code?)

Apologies, I was misreading the error message  🙏 . I haven't checked the logs so you're probably correct

nhakmiller (approved):

nhakmiller:
We could probably apply this same logic to a number of places in the snapshot poller, part of a larger task though. This covers this issue which was at least frequent enough to be noticed by oncall staff.

PR #2030, opened 2020-11-16
Several AWS client-related fixes for log processor

kostaspap:
## Background

Several AWS client-related fixes for log processor. We have encountered these issues while running the log processing at scale.

## Changes

- Change S3Uploader session in the log processor to retry on AccessDenied exceptions. We should never get an AccessDenied error while writing to the Processed Data bucket. The error should appear due to a) Issues on AWS side b) Someone modified the Log Processor Lambda's permissions after deployment. We have seen this error appear from time to time pointing to issues on the AWS side. 
- Seeing sometimes `InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records` error while trying to read data from S3. This issue can appear if we use expired credentials while making a request. Increasing the expiry window from 1' to 2'
- Reducing the period for which we keep the source cache from 5' to 2'. Due to this cache, it could mean that if someone would try sending data to Panther within 5' of a source being onboarded, the data would be dropped. Reducing this period to make that window smaller while making sure we don't overwhelm the source-api with requests. 

## Testing

- mage test:ci


internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -42,11 +42,14 @@ import (
...
-	// How frequently to query the sources_api for new integrations
-	sourceCacheDuration = 5 * time.Minute
+	// sessionDuration is the duration of S3 client STS session
+	sessionDuration = time.Hour
+	// Expirty window for the STS credentials.
+	// Give plenty of time for refresh, we have seen that 1 minute refresh time can sometimes lead to InvalidAccessKeyId errors
+	sessionExpiryWindow   = 2 * time.Minute
+	sourceAPIFunctionName = "panther-source-api"
+	// How frequently to query the panther-sources-api for new integrations
+	sourceCacheDuration = 2 * time.Minute
```
rleighton, line 52: Should `sourceCacheDuration`  be << than `sessionExpiryWindow` to ensure the cached values do not expire?

kostaspap: Nope, not related. The `sourceCache` contains the mappings of (S3 bucket, S3 prefix) -> `AWS Account ID, Role name`. 
It is not related to credentials

giorgosp (approved):
LGTM.

> it could mean that if someone would try sending data to Panther within 5' of a source being onboarded, the data would be dropped.

Still, I believe we should fix the cache invalidation issue and not drop even 2 mins of events. It is also very confusing for both the users and us the developers. At the very least it should be explicit in the UI. I have also been thinking of a solution that calls the `source-api` (only once) to refresh the cache when an s3 read operation fails with `access denied` error.

kostaspap:
> LGTM.
> 
> > it could mean that if someone would try sending data to Panther within 5' of a source being onboarded, the data would be dropped.
> 
> Still, I believe we should fix the cache invalidation issue and not drop even 2 mins of events. It is also very confusing for both the users and us the developers. At the very least it should be explicit in the UI. I have also been thinking of a solution that calls the `source-api` (only once) to refresh the cache when an s3 read operation fails with `access denied` error.

I don't think `access denied` is the right condition - that would be `no client found` error (when we don't have a mapping of S3 bucket -> Prefix to Source ID. 

The SQS forwarder uses such a logic and. has a specialized cache that we could use, see : https://github.com/panther-labs/panther/blob/master/internal/log_analysis/message_forwarder/cache/cache.go

giorgosp:
> I don't think `access denied` is the right condition - that would be `no client found` error that is generated by us (when we don't have a mapping of S3 bucket -> Prefix to Source ID). `access denied` can happen due to AWS issues

We can get `access denied` if the user updates or deletes/recreates the integration and deletes the old Panther role from AWS console.

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -42,11 +42,14 @@ import (
...
-	sessionDurationSeconds = 3600
-	sourceAPIFunctionName  = "panther-source-api"
-	// How frequently to query the sources_api for new integrations
-	sourceCacheDuration = 5 * time.Minute
+	// sessionDuration is the duration of S3 client STS session
+	sessionDuration = time.Hour
+	// Expirty window for the STS credentials.
+	// Give plenty of time for refresh, we have seen that 1 minute refresh time can sometimes lead to InvalidAccessKeyId errors
+	sessionExpiryWindow   = 2 * time.Minute
+	sourceAPIFunctionName = "panther-source-api"
```
alxarch, line 50: suggestion that is irrelevant to the scope of the PR:
We should set a const for the lambda name somewhere in the sources-api code (probably best to do it in `models` to avoid circular deps).
This would allow us to keep track of where the source api lambda is used.
Same applies for all lambdas. We have these strings scattered around the code and the only way to find callers to the lambda is to search for text.

alxarch (approved):

kostaspap:
> > I don't think `access denied` is the right condition - that would be `no client found` error that is generated by us (when we don't have a mapping of S3 bucket -> Prefix to Source ID). `access denied` can happen due to AWS issues
> 
> We can get `access denied` if the user updates or deletes/recreates the integration and deletes the old Panther role from AWS console.

You got me thinking. As I think about it there are a few failure scenarios that we could handle more properly here: 
1. S3 bucket/prefix mapping not found -> Possible due to user sending data before we have refreshed the cache or when users have just not onboarded a source (e.g. configuration SNS notifications bucket wide but just onboarding in Panther a prefix)
2. Access Denied errors. We've seen it happening mostly due to AWS internal issues - get's fixed after retrying the AWS request. I guess it could. happen if the users onboarded a source, but then modified the role so that it doesn't have the right permissions e.g. removed the KMS permissions or the access to the S3 prefix/bucket. It can also happen if they have configured the wrong KMS key
3. Invalid Access Key ID: Can happen if a user deleted and recreate the role that we use to access the data. Refreshing  the sources cache wouldn't help in that case - we would have to delete and recreate the IAM creds. 

Does it sum it up?

giorgosp:
> > > I don't think `access denied` is the right condition - that would be `no client found` error that is generated by us (when we don't have a mapping of S3 bucket -> Prefix to Source ID). `access denied` can happen due to AWS issues
> > 
> > 
> > We can get `access denied` if the user updates or deletes/recreates the integration and deletes the old Panther role from AWS console.
> 
> You got me thinking. These are the failure scenarios that we could handle better here:
> 
>     1. S3 bucket/prefix mapping not found -> Possible due to user sending data before we have refreshed the cache or when users have just not onboarded a source (e.g. configuration SNS notifications bucket wide but just onboarding in Panther a prefix). We could refresh the source cache to get them.
> 
>     2. Access Denied errors. We've seen it happening from time to time due to AWS internal issues - it recovers after retrying the AWS request. I guess it could  happen if the users onboarded a source, but then modified the role so that it doesn't have the right permissions e.g. removed the KMS permissions or the access to the S3 prefix/bucket. It can also happen if they have configured the wrong KMS key
> 
>     3. Invalid Access Key ID: Can happen if a user deleted and recreate the role that we use to access the data. Refreshing  the sources cache wouldn't help in that case - we would have to delete and recreate the IAM creds.
> 
> 
> Does it sum it up?

Yes, for 3 we can't do anything but refreshing the cache on failures would help with the first two points. I think it is worth a ticket.

Issue #1686, opened 2020-10-07
API: Remove internal gateways


austinbyers:
As part of the broader effort to consolidate our API, these internal API gateways will be removed:

* `analysis-api`
* `compliance-api`
* `resources-api`
* `remediation-api`

These are artifacts of a different time in Panther's evolution, and they no longer make sense in our current architecture. They add API latency, are inconsistent with the majority of our backend APIs, and force a dependency on Swagger which has caused several recurring problems in the codebase.

The `public-core-api` gateway will remain for now - its two public endpoints will eventually be moved to the unified public API gateway when that time comes.

### Implementation Overview
The API gateways will be removed, but the Lambda functions serving the requests will remain unchanged for now. Consumers will just need to switch from invoking the HTTPS API gateway to invoking the Lambda function directly

PR #2036, opened 2020-11-17
Remove analysis-api gateway

austinbyers:
## Background

Removes the fourth and final internal gateway. Apologies for the large PR: the `analysis-api` is our oldest and largest API service and there's no easy way to do a partial migration. Most of the changes are just deletions, and the remainder is split across multiple reviewers.

Closes: #1690 
Closes: #1686 (there will be one more PR after this one to remove `swagger` from `mage setup deploy` and to remove `pkg/gatewayapi`)

## Changes

As the last major contributor, @lindsey-w to review the `analysis-api` itself:

* The swagger `api.yml` file has been ported to native go structs (`api/lambda/analysis/models`)
    * `type CoreEntry struct` is the subset of fields that apply to all 4 analysis items (data model, global, rule, and policy)
    * `type PythonDetection struct` is the subset of fields that apply to both rules and policies
    * Embedding these shared structs reduced code duplication considerably, but I'm not sure if it actually made it easier to understand. The JSON structure does not change, only the Go structs. Curious whether the reviewers like this approach or no
* The `GetEnabled*` endpoints have been removed completely - they were always special versions of the `List*` endpoints that selected a specific set of fields
    * Instead, the `List*` endpoints now have a `Fields` option so the caller can ask for just the fields they care about
* All `Update*` endpoints are now almost identical to `Create*` and are defined in the same file
    * This revealed (and fixed) a pre-existing bug: `reports` existed in the Policy definition, but not in the update struct. Now reports can be updated in a policy
* I've separated the Delete endpoints - `DeletePolicies`, `DeleteRules`, `DeleteGlobals`, and `DeleteDataModels`. The input structs are still the same, but their implementations are not - policies need to delete compliance status, and it's a waste of cycles to have rule/model deletion try to delete nonexistent compliance statuses
* Similarly, there are now separate `TestRule` and `TestPolicy` endpoints
* The integration test has been refactored somewhat - each group of tests can now run in parallel

---

@nhakmiller to review associated cloud-security/core changes  (and to help with `analysis-api` models review):
* `alert-processor`, `remediation-api`, `resource-processor`, `alert-delivery`, `layer-manager`, and `metrics-api` are all pointing to the Lambda function instead of the gateway

---

@kostaspap to review associated log-processor changes:

* The `rules-engine`, `alerts-api`,  and `alert-forwarder` now invoke the `analysis-api` Lambda directly instead of routing through API gateway

---

Finally, @3nvi to review the minor changes to the frontend/AppSync

* URL-encoding and CSV arrays are no longer necessary and have been removed as best I could
    * NOTE: In the `listPolicies` call, the frontend is sending a single CSV string in the array instead of an array of strings: `"resourceTypes": ["Type1,Type2"]`. I don't know why this happens; it's not in AppSync. It's not a big deal and I just work around it for now, but if you see a quick fix, I'd love to add it

## Testing

- Fresh deploy
- Upgrade from v1.12
- All integration tests
- Mini-release testing: lots of playing with rules/policies/globals in the web app to make sure it still works

I'd appreciate if some of the reviewers can deploy the branch and interact with the web app to help find any problems I may have missed.

~~There are two issues I'm trying to track down, any leads you know of would be helpful:~~

Turns out the bugs were due to missing permissions in the `layer-manager`, fixed now. All integration tests pass and every interaction in the web UI appears to be in order


web/src/pages/ListRules/ListRules.tsx
```diff
@@ -43,7 +43,7 @@ const ListRules = () => {
   const { loading, error, data } = useListRules({
     fetchPolicy: 'cache-and-network',
     variables: {
-      input: encodeParams(requestParams, ['nameContains']),
```
3nvi, line 46: Do we remove param encoding because it's no longer a `/GET` request?

Same question applies to all places where `encodeParams` has been removed

deployments/appsync.yml
```diff
@@ -1096,29 +1067,23 @@ Resources:
...
       ResponseMappingTemplate: |
         #set ($statusCode = $ctx.result.statusCode)
-        #if($statusCode == 200)
+        #if($statusCode >= 200 && $statusCode < 300)
             true
-        #elseif($statusCode >= 400 && $statusCode < 500)
-            $util.error($util.parseJson($ctx.result.body).message, $statusCode, $ctx.args)
         #else
-            $util.error($ctx.result.body, $statusCode, $ctx.args)
+            $util.error($ctx.result.body, "$statusCode", $input)
```
3nvi, line 1086: nice catch. Thanks for that!

deployments/appsync.yml
```diff
@@ -976,25 +966,24 @@ Resources:
...
       TypeName: Query
       FieldName: policies
-      DataSourceName: !GetAtt AnalysisAPIHttpDataSource.Name
+      DataSourceName: !GetAtt AnalysisAPILambdaDataSource.Name
       RequestMappingTemplate: |
+        #set ($input = $util.defaultIfNull($ctx.args.input, {}))
+        $util.qr($input.put("fields",
+            ["autoRemediationId", "autoRemediationParams", "complianceStatus",
+             "displayName", "enabled", "id", "lastModified", "resourceTypes", "severity", "tags",
+             "threshold", "suppressions", "outputIds", "reports"]))
```
3nvi, line 975: This does seem a bit flimsy for me. Can we not omit that and have the lambda return all  by default whenever a `fields` key is not specified in the input payload?

We can keep it here if this is too much effort, but I feel things can easily go wrong on this level

deployments/appsync.yml
```diff
@@ -976,25 +966,24 @@ Resources:
...
-          "params": {
-            "query": $util.toJson($ctx.args.input),
-            "headers": {
-              "Content-Type": "application/json"
-            }
-          }
+          "version" : "2017-02-28",
+          "operation": "Invoke",
+          "payload": $util.toJson({
+            "listPolicies": $input
```
3nvi, line 980: FYI in the next segments of this file, I'm trusting that the payload keys (i.e. invoked functions) have been typed correctly (can't validate their names one by one :P)

deployments/appsync.yml
```diff
@@ -1384,42 +1336,23 @@ Resources:
...
-          #set($tagsCSV = "")
-          #foreach($tag in $$payload.tags)
-              #set($tagsCSV = $tagsCSV  + "," + $tag)
-          #end
-        #end
-        $util.qr($payload.put("tags", $tagsCSV))
+        #set ($input = $util.defaultIfNull($ctx.args.input, {}))
+        $util.qr($input.put("fields",
+            ["displayName", "enabled", "id", "lastModified", "resourceTypes", "severity", "tags",
+             "threshold", "outputIds", "reports"]))
```
3nvi, line 1344: again same comment as above. Feel free to resolve the current one, just putting it here in case this code-piece got missed

deployments/appsync.yml
```diff
@@ -1521,29 +1436,23 @@ Resources:
...
       ResponseMappingTemplate: |
         #set ($statusCode = $ctx.result.statusCode)
-        #if($statusCode == 200)
+        #if($statusCode >= 200 && $statusCode < 300)
             true
-        #elseif($statusCode >= 400 && $statusCode < 500)
-            $util.error($util.parseJson($ctx.result.body).message, $statusCode, $ctx.args)
         #else
-            $util.error($ctx.result.body, $statusCode, $ctx.args)
+            $util.error($ctx.result.body, "$statusCode", $input)
```
3nvi, line 1455: 👍

deployments/appsync.yml
```diff
@@ -1612,25 +1509,21 @@ Resources:
       ApiId: !Ref ApiId
       TypeName: Query
       FieldName: listGlobalPythonModules
-      DataSourceName: !GetAtt AnalysisAPIHttpDataSource.Name
+      DataSourceName: !GetAtt AnalysisAPILambdaDataSource.Name
       RequestMappingTemplate: |
+        #set ($input = $ctx.args.input)
+        $util.qr($input.put("fields", ["id", "lastModified", "tags"]))
```
3nvi, line 1515: again, I slightly disagree with that. The GraphQL type declares:


```graphql

type GlobalPythonModule {
  body: String!
  description: String!
  id: ID!
  createdAt: AWSDateTime!
  lastModified: AWSDateTime!
}
```

but we are only selecting a handful of fields. What if we change the designs and we need another field from the **existing** type? We'd need to change the resolver to accommodate this.

For this reason, I feeel that returning **all** is always preferable. Eitherway, GraphQL (AppSync) will then automatically filter the response and only send the requested data down the network

deployments/appsync.yml
```diff
@@ -1815,23 +1696,14 @@ Resources:
...
-          "resourcePath": "/v1/policy/test",
-          "params": {
-            "headers": {
-              "Content-Type": "application/json"
-            },
-            "body": $util.toJson({
-              "body": $ctx.args.input.body,
-              "resourceTypes": $ctx.args.input.resourceTypes,
-              "tests": $ctx.args.input.tests,
-              "analysisType": "POLICY"
```
3nvi, line 1832: Seeing the code above, I understand that  `analysisType` param is not needed in the `testPolicy` function anymore.

Just want to double-check

3nvi (commented):
Looks amazing @austinbyers. Thank you so much for all this work. I:

- verified that appsync resolver migration is correct
- verified the web changes

Have a couple of questions, but looks solid so far 🎉

deployments/master.yml
```diff
@@ -240,11 +240,11 @@ Resources:
 
   Appsync:
     Type: AWS::CloudFormation::Stack
+    DependsOn: BootstrapGateway # custom resources lambda function
```
kostaspap, line 243: Isn't the custom Lambda function inside Bootstrap? Why create a dependency to BoostrapGateway?

deployments/master.yml
```diff
@@ -286,11 +285,11 @@ Resources:
 
   Core:
     Type: AWS::CloudFormation::Stack
+    DependsOn: BootstrapGateway # custom resources Lambda function
```
kostaspap, line 288: Same as above

internal/compliance/remediation_api/remediation/invoke.go
```diff
@@ -28,9 +28,7 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
-	analysisclient "github.com/panther-labs/panther/api/gateway/analysis/client"
-	analysisoperations "github.com/panther-labs/panther/api/gateway/analysis/client/operations"
-	analysismodels "github.com/panther-labs/panther/api/gateway/analysis/models"
+	analysismodels "github.com/panther-labs/panther/api/lambda/analysis/models"
```
kostaspap, line 31: Seeing this file reminds me that we could eventually git rid of the extra "remediation" Lambda - the remediation logic could be as well be moved to this API. Not related to this PR, just thinking out loud here :)

internal/compliance/resource_processor/processor/analysis_api.go
```diff
@@ -47,20 +47,37 @@ func getPolicies() (policyMap, error) {
...
-	if err != nil {
-		return nil, errors.WithMessage(err, "failed to load policies from analysis-api")
+	listInput := analysismodels.LambdaInput{
+		ListPolicies: &analysismodels.ListPoliciesInput{
+			Enabled: aws.Bool(true),
+			Fields: []string{
+				// select only the fields we need so it fits in a single response
+				"body", "id", "outputIds", "reports", "resourceTypes", "severity", "suppressions", "tags", "versionId",
+			},
+			PageSize: 1000,
```
kostaspap, line 57: Is this number here dangerous? What's the size of a policy body? Worrying we could easily hit the 6MB limit of a Lambda invocation response. Maybe lower it a bit?

internal/compliance/resource_processor/processor/analysis_api.go
```diff
@@ -47,20 +47,37 @@ func getPolicies() (policyMap, error) {
...
+	listOutput.Paging.TotalPages = 1
+
+	// There should only be one page, but loop over them just in case
+	policies := make(policyMap)
+	for page := 1; page <= listOutput.Paging.TotalPages; page++ {
+		listInput.ListPolicies.Page = page
+
+		if _, err := analysisClient.Invoke(&listInput, &listOutput); err != nil {
+			return nil, errors.WithMessage(err, "failed to load policies from analysis-api")
+		}
```
kostaspap, line 70: nitpick: Had to read the code twice to understand why we were setting the `output.Paging.TotalPages`. 
Maybe change to something like: 
```
currentPage := 1
for {
   listInput.ListPolicies.Page = currentPage
   if _, err := analysisClient(&listInput, &listOutput)
   ...
   if len(output.Policies) == 0 {
       break;
   }
   currentPage ++
}
```

internal/compliance/resource_processor/processor/analysis_api.go
```diff
@@ -47,20 +47,37 @@ func getPolicies() (policyMap, error) {
...
-	for _, policy := range result.Payload.Policies {
-		policies[string(policy.ID)] = policy
+		// Convert list of policies into a map by ID
+		for _, policy := range listOutput.Policies {
+			policies[policy.ID] = policy
+		}
 	}
 
+	zap.L().Info("successfully loaded enabled policies from analysis-api",
+		zap.Int("policyCount", len(policies)))
```
kostaspap, line 79: Debug?

internal/core/analysis_api/analysis/policy_engine.go
```diff
@@ -50,17 +50,15 @@ func NewPolicyEngine(lambdaClient lambdaiface.LambdaAPI, lambdaName string) Poli
...
 }
 
-func (e *PolicyEngine) TestPolicy(policy *models.TestPolicy) (models.TestPolicyResult, error) {
-	testResults := models.TestPolicyResult{}
-
+func (e *PolicyEngine) TestPolicy(policy *models.TestPolicyInput) (*models.TestPolicyOutput, error) {
 	testResources, err := makeTestResources(policy)
 	if err != nil {
-		return testResults, err
+		return nil, err
```
kostaspap, line 56: Any reason why this was not nil before @giorgosp ? Maybe AppSync expects "something" to be returned? Same throughout the code

internal/core/analysis_api/handlers/compliance.go
```diff
@@ -114,11 +114,12 @@ func complianceBatchDelete(policies []*models.DeleteEntry, resourceTypes []strin
 	}
 
 	// Remove the cached status as well
-	for _, policy := range policies {
-		if complianceCache != nil {
+	if complianceCache != nil {
```
kostaspap, line 117: 👍

internal/core/analysis_api/handlers/create_update_rule.go
```diff
@@ -40,23 +37,43 @@ const (
...
+func (API) UpdateRule(input *models.UpdateRuleInput) *events.APIGatewayProxyResponse {
+	return writeRule(input, false)
+}
+
+// Shared by CreateRule and UpdateRule
+func writeRule(input *models.CreateRuleInput, create bool) *events.APIGatewayProxyResponse {
+	// Rule names are embedded in emails, alert outputs, etc. Prevent a possible injection attack
+	if genericapi.ContainsHTML(input.DisplayName) {
+		return &events.APIGatewayProxyResponse{
+			Body:       "invalid display name: " + genericapi.ErrContainsHTML.Error(),
```
kostaspap, line 53: Should we just whitelist Display name characters?

kostaspap (commented):
Great work!! I have put some comments for nitpicks - feel free to ignore them if you don't have time. 

I found some issue with the Rule testing. 

See screenshot: 
Screen Shot 2020-11-17 at 11 27 04 AM

Note that the title is not displayed anywhere in the output. 
Response from AppSync for the above case: 
```
{"data":{"testRule":{"results":[{"id":"0","name":"Test-2882f4","passed":true,"error":{"message":"","__typename":"Error"},"functions":{"ruleFunction":{"output":"true","error":{"message":"","__typename":"Error"},"__typename":"TestDetectionSubRecord"},"titleFunction":{"output":"test","error":{"message":"","__typename":"Error"},"__typename":"TestDetectionSubRecord"},"dedupFunction":{"output":"test","error":{"message":"","__typename":"Error"},"__typename":"TestDetectionSubRecord"},"alertContextFunction":{"output":"","error":{"message":"","__typename":"Error"},"__typename":"TestDetectionSubRecord"},"__typename":"TestRuleRecordFunctions"},"__typename":"TestRuleRecord"}],"__typename":"TestRuleResponse"}}}
```

@giorgosp , @3nvi  could you verify that the Rule testing functionality works as expected?

internal/core/analysis_api/analysis/policy_engine.go
giorgosp: No it was just the function's signature, now we return a pointer so it can be nil.

internal/core/analysis_api/analysis/policy_engine.go
```diff
@@ -50,17 +50,15 @@ func NewPolicyEngine(lambdaClient lambdaiface.LambdaAPI, lambdaName string) Poli
...
 	}
 	engineInput := enginemodels.PolicyEngineInput{
 		Policies: []enginemodels.Policy{
 			{
-				Body:          string(policy.Body),
+				Body:          policy.Body,
```
giorgosp, line 61: Glad we got rid of these funny casts :)

internal/core/analysis_api/handlers/list_data_models.go
```diff
@@ -0,0 +1,75 @@
...
+	})
+	if err != nil {
+		zap.L().Error("failed to scan data models", zap.Error(err))
+		return &events.APIGatewayProxyResponse{StatusCode: http.StatusInternalServerError}
+	}
+
+	// Sort and page
+	sortItems(items, "id", input.SortDir, nil)
+	var paging models.Paging
+	paging, items = pageItems(items, input.Page, input.PageSize)
```
giorgosp, line 63: Out of scope: Seems like DynamoDB supports some kind of pagination https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html
Have we ever considered this? Because, if we have so few items that we can load them all in the lambda's memory and paginate them, may we don't need to bother with pagination at all in the backend.

giorgosp (commented):
Nice job!

lindsey-w (commented):
Wow! 

Deployed in my dev env, all integration tests passed. Ran across the same issue Kostas’ pointed out in the rule testing. I'm still waiting for my `Always True` rule to trigger (been about 15 minutes), but otherwise haven't come across any issues. 

I do like the reduction of code duplication using the `CoreEntry` - but it does make it slightly more difficult to figure out what exactly is required in the various places where it is used. 

Is there a handy way to manually test the API? I was using the api gateway for manual testing when I was developing some of the DataModel stuff, I thought it was useful.  🤷🏻‍♀️

3nvi:
> @3nvi could you verify that the Rule testing functionality works as expected?

It doesn't. I'll try and see what I can do

internal/log_analysis/rules_engine/src/analysis_api.py
```diff
@@ -14,44 +14,75 @@
...
-        request = AWSRequest(method='GET', url=self.url + '/enabled', params={'type': 'RULE'})
-        self.signer.add_auth(request)
-        prepped_request = request.prepare()
+        # There should only be one page, but loop over them just in case
+        list_input: Dict[str, Any] = {
+            'listRules':
+                {
+                    'enabled': True,
+                    # select only the fields we need to minimize the size of the response
+                    'fields': ['body', 'id', 'logTypes', 'outputIds', 'reports', 'severity', 'tags', 'versionId'],
```
lindsey-w, line 37: does specifying 'logTypes' here get translated to 'resourceTypes' in the projection? I can't seem to find that translation.  I think it does for the filter, but possibly not for the fields/projection?

lindsey-w: FWIW, once I changed this to `resourceTypes` my `always true` rule started to alert as expected.

lindsey-w: Deduplication is also acting strangely.  And I think this might also be missing `dedupPeriodMinutes` ?

web/src/pages/ListRules/ListRules.tsx
austinbyers: > Do we remove param encoding because it's no longer a /GET request?

Correct: parameter encoding used to be necessary because the query parameters were passed in the HTTP GET url. With gateway removed, AppSync is talking directly to Lambda and passes the raw JSON payload; no encoding required

deployments/appsync.yml
austinbyers: These fields were always hardcoded - they used to be in the Go backend; I just moved them up to the AppSync layer to make it easier to configure. 

The Python body + the tests are relatively large and are not needed for the list operation; it's a waste of cycles to load and parse them in Lambda. AppSync will drop them before returning to the user, but Lambda has to load them from Dynamo and send them to AppSync. I was also worried that the extra size could lead to running out of space in the Lambda response before the requested page size is reached.

But I think you're right - this is an unnecessary micro-optimization. If I list all policies, the response size is only 721KB, well below the 6MB Lambda limit. So I'll remove the projection for now and we can re-add it later if we need the extra performance

deployments/appsync.yml
austinbyers: I'm trusting that as well, I just made sure to exercise each endpoint in the web app.

When we add more integration/release testing, we'll want to send requests through the public API, not just the internal one. That way we have some automated tests of the GraphQL

deployments/appsync.yml
austinbyers: Actually, I realized this introduced another bug: I copy-pasted this error-handling to all of the modified endpoints, but the `$input` variable is only defined for some endpoints; the rest need to use `$ctx.args.input`.

deployments/appsync.yml
austinbyers: The fundamental tension here is that we are selecting fields at the GraphQL level *and* at the REST level. But yes, I'll remove the projection for now because the extra processing is negligible.

deployments/appsync.yml
austinbyers: Correct: rather than having an `analysisType` to distinguish testing a rule from a policy, I just made them separate endpoints. Easier to understand and develop

api/lambda/analysis/models/api.go
```diff
@@ -0,0 +1,133 @@
...
+type CoreEntryUpdate struct {
+	Body        string   `json:"body" validate:"required"`
+	Description string   `json:"description"`
+	ID          string   `json:"id" validate:"required"`
+	Tags        []string `json:"tags" validate:"omitempty,dive,required"`
+	UserID      string   `json:"userId" validate:"uuid4"`
+}
+
+// Python rules and policies share these fields
+type PythonDetection struct {
```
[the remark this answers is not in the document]
austinbyers: I'm not going to rename all of the code, but yes, @jacknagz has started using "detections", which do make more sense than "analyses"

api/lambda/analysis/models/data_models.go
```diff
@@ -0,0 +1,79 @@
...
+	// Only include data models which apply to one of these log types
+	LogTypes []string `json:"logTypes" validate:"omitempty,dive,required"`
+
+	// ----- Sorting -----
+	SortBy  string `json:"sortBy" validate:"omitempty,oneof=enabled id lastModified logTypes"`
+	SortDir string `json:"sortDir" validate:"omitempty,oneof=ascending descending"`
+
+	// ----- Paging -----
+	PageSize int `json:"pageSize" validate:"min=0"`
+	Page     int `json:"page" validate:"min=0"`
```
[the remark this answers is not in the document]
austinbyers: Not in my opinion. If the items are sufficiently small, you can grab as many as you want. The only limit is the 6MB lambda response, which is actually quite large

api/lambda/analysis/models/policies.go
```diff
@@ -0,0 +1,139 @@
...
+type DeletePoliciesInput struct {
+	Entries []DeleteEntry `json:"entries" validate:"min=1,max=1000,dive"`
+}
+
+type DeleteEntry struct {
+	ID string `json:"id" validate:"required"`
+}
+
+type GetPolicyInput struct {
+	PolicyID  string `json:"policyId" validate:"required"`
```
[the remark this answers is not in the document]
austinbyers: I thought about that; this is just being backwards compatible with the existing swagger definition.

It shouldn't be too hard to change it, I will try that and revert if it proves challenging for some reason. I'm not too worried about the exact details of the API, since we will still have many iterations of the API layer

api/lambda/analysis/models/api.go
```diff
@@ -0,0 +1,133 @@
...
+	LastModifiedBy string    `json:"lastModifiedBy" validate:"uuid4"`
+	Tags           []string  `json:"tags" validate:"omitempty,dive,required"`
+	VersionID      string    `json:"versionId" validate:"required"`
+}
+
+// Creating or updating any item supports these fields
+type CoreEntryUpdate struct {
+	Body        string   `json:"body" validate:"required"`
+	Description string   `json:"description"`
+	ID          string   `json:"id" validate:"required"`
```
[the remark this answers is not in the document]
austinbyers: > We only allow specific characters for ID - Better add validation for it?

Are you talking about the HTML character rejection? That happens in the Go code rather than the struct validation: `genericapi.ContainsHTML`. But a struct tag is easier, I'll switch to that

> Also probably we want to specify a minimum
Why? While not very useful, an ID or Description with just one character is perfectly valid. I suppose the `Body` could have a min requirement, but if it's invalid Python, we'll flag that anyway. `required` implies `min=1` for strings

> and maximum size (e.g. cannot put several MBs since we store them in DDB)
Sure, I suppose. Or we can wait until the put operation fails to give the maximum possible space instead of trying to artificially limit it ahead of time. It's already limited by AppSync and Lambda payload sizes.

I'll go ahead and add a max anyway

api/lambda/compliance/models/api.go
```diff
@@ -386,7 +386,7 @@ type SetStatusEntry struct {
 // The policy-api updates the relevant policy attributes here when they change (severity/suppressions).
 // For these updates, we don't need to re-scan the resources and can instead directly modify the compliance state.
 type UpdateMetadataInput struct {
-	PolicyID     string         `json:"policyId" validate:"required"`
-	Severity     PolicySeverity `json:"severity" validate:"oneof=INFO LOW MEDIUM HIGH CRITICAL"`
-	Suppressions []string       `json:"suppressions"`
+	PolicyID     string   `json:"policyId" validate:"required"`
+	Severity     Severity `json:"severity" validate:"oneof=INFO LOW MEDIUM HIGH CRITICAL"`
```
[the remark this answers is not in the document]
austinbyers: Sadly, it is still needed. `Severity` is just a different type of string, you could have, for example, `var x = Severity("hello world")`.

deployments/master.yml
austinbyers: > Isn't the custom Lambda function inside Bootstrap?

No, it is inside `bootstrap-gateway`, because lambda functions require the source S3 bucket to exist before the template can be packaged. So the source bucket is created in `bootstrap` and the function created in `bootstrap-gateway`

austinbyers: At this point, the name `bootstrap-gateway` is a complete lie, but unfortunately, it's not very easy to change

internal/compliance/remediation_api/remediation/invoke.go
austinbyers: Yes, someday. But it's not a very commonly used feature and is essentially in maintenance mode for now

internal/compliance/resource_processor/processor/analysis_api.go
nhakmiller: > Is this number here dangerous? What's the avg size of a policy (including the body)? Worrying we could easily hit the 6MB limit of a Lambda invocation response. Maybe lower it a bit and rely on pagination?

Today, the zip file that contains all of our policies & rules is about 20Kb. Even assuming very good compression, that's no more than 100Kb uncompressed. But also today we only ship with 100 policies, so this max size will never get hit. Maybe a max page size of 250 reduces the chance of some crazy future situation, but won't actually ever return more than 1 page of results in most cases?

> Also, page size of 1000 seems too much. Does the UI show 1000 policies simultaneously?

This is a backend system, the UI does not interact with it.

austinbyers: The average policy is 4.2KB (including unit tests, which are ignored here); 1000 of them would be 4.2MB, still well within the 6MB limit. I'll add a comment to that effect.

Today, the entire backend just *assumes* it will fit in a single response. I've added pagination here just in case, so this is already an improvement over what we have today. But pagination is actually not handled efficiently at all right now. **Every analysis-api list request is a full table scan**, so the backend needs to use a single request if at all possible.

The details of the table scanning is explained at the top of `list.go` in the `analysis-api`. It's not ideal, but (a) it doesn't actually matter because there are only a few hundred entries, (b) we've been doing it this way since the very beginning, and (c) once we inline the API calls, the Lambda response limit won't even matter

> Does the UI show 1000 policies simultaneously?

No, the UI uses the default page size of 25. But this is the resource-processor, part of the cloud security backend pipeline.

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -94,37 +93,28 @@ func populateAlertData(alertItem *alertTable.AlertItem) (*deliveryModels.Alert,
...
+		},
 	}
-
-	if response == nil {
-		zap.L().Error("Rule response was nil", commonFields...)
+	var rule analysismodels.Rule
+	if _, err := analysisClient.Invoke(&getRuleInput, &rule); err != nil {
+		zap.L().Error("Error retrieving rule", append(commonFields, zap.Error(err))...)
 		return nil, &genericapi.InternalError{Message: genericErrorMessage}
 	}
```
nhakmiller, line 106: Is this case no longer possible? Or was it never possible to begin with?

internal/compliance/resource_processor/processor/analysis_api.go
austinbyers: When would we ever use an info, then? This only triggers once per invocation / once every few minutes. I'm fine with a debug, I just feel we pretty much always remove info logging, I'm curious what you would consider worth logging at the info level

internal/core/analysis_api/handlers/create_update_rule.go
austinbyers: I don't want to make functional changes if I can avoid them - that could break existing deployments and is outside the scope of this change

nhakmiller (commented):
Working my way through!

api/lambda/analysis/models/api.go
```diff
@@ -0,0 +1,133 @@
...
+	// Data models (log analysis)
+	CreateDataModel  *CreateDataModelInput  `json:"createDataModel,omitempty"`
+	DeleteDataModels *DeleteDataModelsInput `json:"deleteDataModels,omitempty"`
+	GetDataModel     *GetDataModelInput     `json:"getDataModel,omitempty"`
+	ListDataModels   *ListDataModelsInput   `json:"listDataModels,omitempty"`
+	UpdateDataModel  *UpdateDataModelInput  `json:"updateDataModel,omitempty"`
+}
+
+// All detection types (global/data-model/policy/rule/query) have these fields in common
+type CoreEntry struct {
```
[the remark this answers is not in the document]
austinbyers: Sure, I was on the fence about whether to use the shared embeddings or not. I'll just revert to a complete flat model like before because then each input struct can have its own validation tags. Even when they share the same fields, they may not share the same validation

internal/core/analysis_api/analysis/policy_engine.go
austinbyers: Me too! That was an artifact of the swagger-generated code

internal/core/analysis_api/handlers/list_data_models.go
austinbyers: Yes, there are two kinds of paging going on:

For *every* List request, the `analysis-api` scans the table for all matching entries. This uses the `scanPages` call to page over all of the Dynamo entries:

```go
// internal/core/analysis_api/handlers/dynamo.go

// Wrapper around dynamoClient.ScanPages that accepts a handler function to process each item.
func scanPages(input *dynamodb.ScanInput, handler func(tableItem) error) error {
```

This is almost certainly one Dynamo page (1000 items), but it will automatically page over all of them if there are multiple.

----

However, Dynamo paging is *not* what we expose to the frontend. We can't. Why? We apply filtering outside of Dynamo (compliance status) and, more importantly, because Dynamo page keys only work when sorting by the partition / sort key. We allow sorting by arbitrary fields; there's no easy way to do that in Dynamo without building lots of indexes.

That being said, we do want to switch to a cursor-based pagination model: where the frontend just gets a `nextPageKey` instead of having to specify a page number. As part of that work (when implementing the public API) we can revisit the paging implementation and hopefully solve it more efficiently, using native Dynamo paging where possible

api/lambda/analysis/models/data_models.go
```diff
@@ -0,0 +1,79 @@
...
+	DisplayName string             `json:"displayName"`
+	Enabled     bool               `json:"enabled"`
+	LogTypes    []string           `json:"logTypes"`
+	Mappings    []DataModelMapping `json:"mappings"`
+}
+
+type DataModelMapping struct {
+	Name   string `json:"name" validate:"required"`
+	Path   string `json:"path" validate:"required_without=Method"`
+	Method string `json:"method" validate:"required_without=Path"`
```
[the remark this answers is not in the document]
austinbyers: I couldn't find a way to enforce "these can't both be defined" without doing it in the Go code, but at least this way we got one of the checks in the validate tags!

internal/log_analysis/rules_engine/src/analysis_api.py
```diff
@@ -14,44 +14,75 @@
...
+                    'pageSize': 1000,
```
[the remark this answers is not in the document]
austinbyers: Yes, that concern would apply here as well, but we're well under the lambda limit (and again, it actually just assumed the whole thing would fit until now)

internal/log_analysis/rules_engine/src/analysis_api.py
austinbyers: > does specifying 'logTypes' here get translated to 'resourceTypes' in the projection? I can't seem to find that translation. I think it does for the filter, but possibly not for the fields/projection?

Good catch! Yes, I suspect there is an issue here with the `logTypes` vs `resourceTypes`, I'll look into it. I bet you're right - there needs to be a translation since the raw dynamo field is called `resourceTypes`. We could also potentially remove the projection entirely, like Aggelos suggested with the frontend

internal/compliance/resource_processor/processor/analysis_api.go
austinbyers: Haha I commented before seeing @nhakmiller 's response :) 

> Maybe a max page size of 250 reduces the chance of some crazy future situation, but won't actually ever return more than 1 page of results in most cases?

Yes, that seems a reasonable compromise. Or even 500

internal/core/alert_delivery/api/deliver_alert.go
austinbyers: I don't think it was ever possible, to be honest. If it doesn't exist, we'll get a 404 error

austinbyers:
> Is there a handy way to manually test the API? I was using the api gateway for manual testing when I was developing some of the DataModel stuff, I thought it was useful. 🤷🏻‍♀️

@lindsey-w , yes: all Lambda functions can be invoked directly using a "test event" in the AWS console. It's actually much much easier than testing API gateway.

For example, just select the "analysis-api" and add the following test event:

```json
{
    "listPolicies": {}
}
```

Then click Test! You'll see both the response and the logs underneath

Screen Shot 2020-11-17 at 3 54 09 PM

internal/compliance/resource_processor/processor/analysis_api.go
kostaspap: That's a good question. I guess what is considered info or debug logging can be subjective sometimes. 

My main concern is that the logging is performed deep inside the call stack - it is hard to say really how frequently this method will be invoked or which method is invoking it. So it's hard to say if it is really useful to be logged every time. For logging performed that deep in the call stack, I generally prefer to use Debug. 
As I see it, info logging should in generally closer to the initial handler. E.g. a log line like "analyzing <N number> of resources using <M number of policies" would be much more useful than 2-3 log lines ("fetched N policies", "fetched M resources", "running analysis",..etc)

E.g. following up the call chain I see that the `getPolicies()` is called by `batchResults.analyze` which is called once per resources [page](https://github.com/panther-labs/panther/blob/4325edf255f2fe8c99f14903224805f5cdc88260/internal/compliance/resource_processor/processor/handler.go#L151-L153). The current resource page is 2k which I guess can be considered significant, still in a deployment of say 100k resources we would end up with 50 log info messages ("successfully loaded enabled policies from analysis-api")

internal/core/analysis_api/handlers/create_update_rule.go
kostaspap: Fair

deployments/appsync.yml
3nvi: oh lol yeah. That's the beauty of having no linter for this VTL thing. Shall I assume that you took care of that now?

deployments/appsync.yml
3nvi: We won't have appsync by then. We'll have golang which will automatically do these sort of validation for us

nhakmiller (approved):
I've looked over alert-processor, remediation-api, resource-processor, alert-delivery, layer-manager, and metrics-api and everything seems to be in order there.

deployments/appsync.yml
austinbyers: I also realized the Appsync error handling needs two steps for these APIs:

* Check if the Lambda function fails completely (panic, OOM, etc). This is the `if $context.error` handling you have on the other functions
* If the Lambda function returns the 4XX or 5XX status code (handled here)

---

Since most resolvers have one of 3 common response templates, I moved those into a `Mapping` (like a CloudFormation constant) so we could use the same ones everywhere they apply.

I also replaced all `$context` references with `$ctx` to be consistent and reduce the size of the template

api/lambda/analysis/models/api.go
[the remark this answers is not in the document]
austinbyers: It was actually a good exercise to revisit all of the input validation (especially after flattening the Go structs), most string fields have a max cap on them now

austinbyers:
I've addressed the biggest pieces of feedback:

* AppSync: fix error handling and remove "fields" projection (@3nvi this is ready for re-review)
* Flatten Go structs (remove `CoreEntry` and `PythonDetection`)
* Use "ID" as the input field name for GetPolicy, GetRule, etc

I still need to fix rules and rule tests and address the other feedback, I'll do that tomorrow.

deployments/appsync.yml
3nvi: 👍

3nvi:
> I've addressed the biggest pieces of feedback:
> 
> * AppSync: fix error handling and remove "fields" projection (@3nvi this is ready for re-review)
> * Flatten Go structs (remove `CoreEntry` and `PythonDetection`)
> * Use "ID" as the input field name for GetPolicy, GetRule, etc
> 
> I still need to fix rules and rule tests and address the other feedback, I'll do that tomorrow.

The changes for rule testing are the following:

- make sure that `error` is present only if an error occurs (within the context of a `TestDetectionSubRecord` type)
- check if `TestDetectionSubRecord` are always returned now (they are not optional anymore) within

```graphql
titleFunction: TestDetectionSubRecord
dedupFunction: TestDetectionSubRecord
alertContextFunction: TestDetectionSubRecord
```

in order to mark them as required. If they are, I'll need to make a small update in the `RuleFormTestResult.tsx` file

s0l0ist (approved):
LGTM 🎉

api/lambda/analysis/models/policies.go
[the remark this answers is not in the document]
austinbyers: Changed to ID!

internal/compliance/resource_processor/processor/analysis_api.go
austinbyers: I've removed the projection and set the pageSize to 250

internal/compliance/resource_processor/processor/analysis_api.go
austinbyers:
```
   if len(output.Policies) == 0 {
       break
   }
```

This condition will never be true - every page should have results. And if we did ask for an empty page, that would be a wasted API call (and full table scan). The stop condition is when the current page matches the total number of pages as indicated in the response.

I'll try to add comments or otherwise clarify

austinbyers: Hopefully this is clearer:

```go
		if listOutput.Paging.ThisPage == listOutput.Paging.TotalPages {
			break
		}
```

internal/compliance/resource_processor/processor/analysis_api.go
austinbyers: Configured to debug

internal/core/analysis_api/handlers/create_update_rule.go
austinbyers: But I did remove this check in Go and move it to an `excludesall` validation on the struct field

internal/log_analysis/rules_engine/src/analysis_api.py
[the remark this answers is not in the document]
austinbyers: Lowered page size to 250 and changed style to match the paging that other calling functions use

internal/log_analysis/rules_engine/src/analysis_api.py
austinbyers: I've removed the projection entirely, we can just get all fields, which is easier to maintain (and the extra size is not that big a deal). Also I've added the `logTypes` -> `resourceTypes` mapping in the listRules handler in case we do use it again

austinbyers:
Ok, I'm done addressing feedback and this is good to merge, but I'll wait for an approval by @kostaspap or @3nvi with regards to the rule testing.

* **Fixed rules**: an AlwaysTrue rule is correctly generating alerts. @lindsey-w was right about the root cause - we weren't mapping the `logTypes` projection field to the `resourceTypes` column in Dynamo. I fixed the mapping but also just removed the projection entirely, it's a premature optimization
* **Fixed rule tests**: the swagger-generated Go models we were using before have `omitempty` in all of the output fields, while the new ones do not. I probably could have just added `omitempty` to the go structs and been done with it... but then I saw the horrifying AppSync resolvers for TestPolicy and TestRule, and I was sad 😢  Poor @3nvi had to completely rebuild the backend response to get it in a format he could use in the frontend. The testing functionality is primarily *for* the frontend, so I just rewrote the test output Go structs to conform to the GraphQL spec, and now AppSync can pass the response along without all the hellish hackery.
* **Fixed a panic in alert-delivery**: It was still expecting the `ANALYSIS_API_HOST` environment variable

--- 

I did another big round of testing:

* Fresh deploy: no CW alarms
* All integration tests
* tested policies and rules: invalid syntax, all tests pass, tests fail (wrong result), tests error (raise exceptions)

Screen Shot 2020-11-19 at 4 46 40 PM
Screen Shot 2020-11-19 at 4 49 29 PM

kostaspap (approved):
I verified that the Rule testing functionality now works as expected. Shiiiiiiiiiip it!!

PR #2037, opened 2020-11-17
Add support for scanning indicators on slice fields

alxarch:
## Background

There are cases where a log contains a field with an array of values where each value can be used as an indicator.

## Changes

- Adds support for using `panther` struct tags on slice fields
- Uses a reflect based method to extract indicator values so that all cases can be covered

## Testing

- mage test:ci


internal/log_analysis/log_processor/pantherlog/jsoniter_test.go
```diff
@@ -87,11 +87,13 @@ func init() {
...
-		Baz  string        `json:"baz" panther:"baz"`
-		Qux  *string       `json:"qux" panther:"qux"`
-		Quux null.String   `json:"quux" panther:"quux"`
+		Foo      *testStringer `json:"foo" panther:"foo"`
+		Bar      testStringer  `json:"bar" panther:"bar"`
+		Baz      string        `json:"baz" panther:"baz"`
+		Qux      *string       `json:"qux" panther:"qux"`
+		Quux     null.String   `json:"quux" panther:"quux"`
+		FooSlice []string      `json:"foos" panther:"foo"`
+		BarSlice []null.String `json:"bars" panther:"bar"`
```
kostaspap, line 96: What about `[]*string`, `[]testStringer`, `[]*testStringer`?

internal/log_analysis/log_processor/pantherlog/jsoniter.go
```diff
@@ -306,6 +306,25 @@ func (*pantherExt) decorateIndicatorField(b *jsoniter.Binding, scanners ...Value
...
+		case el.ConvertibleTo(typStringPtr):
+			b.Encoder = &scanStringSliceEncoder{
+				parent:  b.Encoder,
+				scanner: scanner,
+			}
+		case el.ConvertibleTo(typNullString):
+			b.Encoder = &scanNullStringSliceEncoder{
+				parent:  b.Encoder,
+				scanner: scanner,
+			}
```
kostaspap, line 326: These cases look very similar to the cases in lines 328-356. Could you move them to their own function?

internal/log_analysis/log_processor/pantherlog/jsoniter_test.go
alxarch: Fixed

internal/log_analysis/log_processor/pantherlog/jsoniter.go
alxarch: Fixed

internal/log_analysis/log_processor/pantherlog/jsoniter_test.go
```diff
@@ -87,23 +87,29 @@ func init() {
 func TestPantherExt_DecorateEncoder(t *testing.T) {
 	// Check all possible string types
 	type T struct {
-		Foo  *testStringer `json:"foo" panther:"foo"`
-		Bar  testStringer  `json:"bar" panther:"bar"`
-		Baz  string        `json:"baz" panther:"baz"`
-		Qux  *string       `json:"qux" panther:"qux"`
-		Quux null.String   `json:"quux" panther:"quux"`
+		Foo      testStringer  `json:"foo" panther:"foo"`
+		Bar      testStringer  `json:"bar" panther:"bar"`
```
kostaspap, line 91: Probably just one of them is needed?

internal/log_analysis/log_processor/pantherlog/jsoniter_test.go
```diff
@@ -87,23 +87,29 @@ func init() {
...
+		Baz      string        `json:"baz" panther:"baz"`
+		Qux      *string       `json:"qux" panther:"qux"`
+		Quux     null.String   `json:"quux" panther:"quux"`
+		FooSlice []string      `json:"foos" panther:"foo"`
+		BarSlice []null.String `json:"bars" panther:"bar"`
+		QuxSlice []*string     `json:"quxs" panther:"qux"`
```
kostaspap, line 97: Do we need `[]testStringer`?

internal/log_analysis/log_processor/pantherlog/jsoniter_test.go
alxarch: It used to be `*testStringer` which should theoretiacally work again now. Reflection is a horrible horrible thing.

internal/log_analysis/log_processor/pantherlog/jsoniter_test.go
alxarch: no because `null.String` is already supported based on its implementation of fmt.Stringer

kostaspap (approved):

kostaspap:
@alxarch we should add the "sync-to-release" label?

internal/log_analysis/log_processor/pantherlog/jsoniter_test.go
```diff
@@ -131,11 +138,13 @@ func TestResultEncoder(t *testing.T) {
 		Time     time.Time `json:"tm" event_time:"true"`
 		RemoteIP string    `json:"remote_ip" panther:"ip"`
 		LocalIP  string    `json:"local_ip" panther:"ip"`
+		Tags     []string  `json:"tags" panther:"aws_tag"`
```
kostaspap, line 141: The tag is in the slice - no the individual element (as it appears in custom logs). I guess the indicator gets bubbled up to the slice somehow?

internal/log_analysis/log_processor/pantherlog/jsoniter_test.go
```diff
@@ -131,11 +138,13 @@ func TestResultEncoder(t *testing.T) {
 		Time     time.Time `json:"tm" event_time:"true"`
```
kostaspap, line 138: Can you add a test case for a tested object and put the slice in that object? 

Something like:
```
type A struct {
Tags []string `json:"tags" panther:"aws_tag"`
}
type T struct{
...
NestedStruct A `json:"nestedStruct"
}
```

internal/log_analysis/log_processor/pantherlog/jsoniter_test.go
alxarch: In This PR handles the OSS part of this. How the equivalent struct will be created concerns the EE, customlogs part

alxarch:
> @alxarch we should add the "sync-to-release" label?

IMO this is a new feature. For OSS it is not needed in the release since there are no custom logs. The sync PR on EE should have `sync-to-release` set. 

OTOH adding this to the release does not affect the existing functionality so we can add the label to not have different code on OSS/enterprise release.

internal/log_analysis/log_processor/pantherlog/jsoniter_test.go
alxarch: This case is not relevant to what this PR is fixing. We are concerned only with slices of 'scalar' values. Slices of structs with fields will be handled normally at the struct level

PR #2040, opened 2020-11-17
GraphQL: Use undocumented `Long` scalar

3nvi:
## Background

Currently, we use `Int` and `Float`  as GraphQL scalars whenever we want to depict a numerical value. AppSync uses an 32bit `Int` and `Float` JavaLang representation under the hood, causing some big values ( > 2,000,000,000) to crash our system.

This hasn't been identified so far, because our test data didn't come near to those numbers. AppSync doesn't have any publicly documented scalars (other than `Int` and `Float`) and it doesn't allow you to specify your own custom scalars. The need for a `Long` type has been heavily requested [for more than a year](https://github.com/aws/aws-appsync-community/issues/21), but so far there haven't been any updates from AWS.

Due to the fact that AppSync uses Java under the hood (that's why it can parse VTL templates as well), I took advantage of that and used a `Long` type, knowing that `graphql-java` would map it to a [`Java Long`](https://docs.oracle.com/javase/7/docs/api/java/lang/Long.html) under the hood.

This is an  **undocumented feature**, but it serves our goal until we migrate away from Appsync (cc @austinbyers). A `Long` in java is a 64bit number.

## Changes

- Define a `Long` scalar our GraphQL documents
- Utilize `Long` scalar in our schema
- Update our schema, codegen and builders to let them be aware of this new scalar

## Testing

- `npm run test`

## Notes

I've checked all of the other `Int` declarations in our schema and verified that the OSS doesn't have any additional fields, where such an exception could occur

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -2,6 +2,7 @@ scalar AWSEmail
 scalar AWSDateTime
 scalar AWSTimestamp
 scalar AWSJSON
+scalar Long
 
 schema {
   query: Query
@@ -411,9 +412,9 @@ type OrganizationStatsResponse {
   topFailingResources: [ResourceSummary]
 }
 
-type Series {
-  label: String
-  values: [Int]
+type LongSeries {
+  label: String!
+  values: [Long!]!
 }
 
 type FloatSeries {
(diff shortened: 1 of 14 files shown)


austinbyers (approved):

austinbyers:
Does this need to sync to 1.13 release? If so, don't forget to add the label

3nvi:
> Does this need to sync to 1.13 release? If so, don't forget to add the label

It's up for discussion :D

Issue #1947, opened 2020-11-04
Issue sending to SNS destinations sometimes


nhakmiller:
### Describe the bug

Sometimes there is still an issue when sending to SNS destinations. The send fails with the following message:

```
InvalidParameter: InvalidParameter: Subject status code: 400, request-id <...>
```
### Steps to reproduce

End user reported, still trying to exactly replicate. But it appears to be related to long strings used as the `title` of an alert.

### Expected behavior

SNS destinations work always.

### Environment

How are you deploying or using Panther?

- Panther version v1.11.

### Additional context

Possibly related to the issue encountered and resolved here: https://github.com/panther-labs/panther/pull/1674. It could be that this fix didn't fully resolve the issue, or that there is another similar issue.

PR #2047, opened 2020-11-18
SNS Title Input Validation

horakj:
## Background

Issue sending to SNS destinations sometimes #1947

## Changes

- added SNS Helper function
     - additional Title Input Validation
     - additional SNS API Call after first attempt failure
          - second call replaces Title with "Panther New Alert"

## Testing
- Independent main
- mage test:go
- mage test:ci
- dev panther UI


internal/core/alert_delivery/outputs/sns.go
```diff
@@ -41,6 +44,26 @@ type snsMessage struct {
 // Tests can replace this with a mock implementation
 var getSnsClient = buildSnsClient
 
+// Helper Function handles title input validation and makes SNS API Call
+func snsPublishHelper(snsClient snsiface.SNSAPI, snsMessageInput *sns.PublishInput) (*sns.PublishOutput, error) {
+	fmt.Println("Function Here")
```
nhakmiller, line 49: Looks like some debugging output, lets remove this.

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -41,6 +44,26 @@ type snsMessage struct {
...
+	tmp := *snsMessageInput.Subject
+	// Replace line breaks to preserve original title
+	title := strings.Replace(tmp, "\n", "", 100)
+	// Restrict title length to 100 Chars
+	if len(title) > 100 {
+		*snsMessageInput.Subject = title[0:100]
+	}
+	result, err := snsClient.Publish(snsMessageInput)
+	if err != nil {
```
nhakmiller, line 58: Do you know what type of error gets returned when the title is invalid? Ideally we could handle just this type of error specially and for other errors follow the normal path.

We can do this by first trying to force the error into an `awserr.Error`, and then checking its `Code()` or  `Message()` function to see what the error code/message is. See here for an example: https://github.com/panther-labs/panther/blob/master/internal/compliance/snapshot_poller/pollers/aws/iam_user.go#L84-L93.

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -41,6 +44,26 @@ type snsMessage struct {
...
+		// If error is returned assign new title and make an additional SNS API Call
+		*snsMessageInput.Subject = "New Panther Alert"
+		result, err := snsClient.Publish(snsMessageInput)
+		return result, err
+	}
+	return result, err
```
nhakmiller, line 64: You can re-use these same named variables from above and re-write this to:

```
		result, err = snsClient.Publish(snsMessageInput)
	}
	return result, err
```

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -75,9 +98,10 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
 	}
 
 	title := generateAlertTitle(alert)
-	if len(title) > 100 {
-		title = title[0:100]
-	}
+	// Relocated to snsPublishHelper Function
+	// if len(title) > 100 {
+	// 	title = title[0:100]
+	// }
```
nhakmiller, line 104: Lets cut this dead code here if we don't need it anymore.

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -98,8 +122,9 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
 			Success:    false,
 		}
 	}
-
-	response, err := snsClient.Publish(snsMessageInput)
+	// Relocated to snsPublishHelper Function
+	// response, err := snsClient.Publish(snsMessageInput)
```
nhakmiller, line 126: Cut this as well.

nhakmiller (commented):
Nice first PR! Thanks for grabbing this one.

Can you add a unit test as well? We already have some unit tests around this function, you can just expand those to include at least one case where a newline character is removed.

internal/core/alert_delivery/outputs/sns.go
kostaspap, line 126: Another way to do it (I believe more clean) is to create a new function `func formatTitle(string) string` that you just pass it the existing title and it removes the new line characters and truncates it. 
Then the only change required would be to replace line 110 with 
```
Subject: aws.String(formatTitle(title))
```

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -41,6 +44,26 @@ type snsMessage struct {
...
 
+// Helper Function handles title input validation and makes SNS API Call
+func snsPublishHelper(snsClient snsiface.SNSAPI, snsMessageInput *sns.PublishInput) (*sns.PublishOutput, error) {
+	fmt.Println("Function Here")
+	tmp := *snsMessageInput.Subject
+	// Replace line breaks to preserve original title
+	title := strings.Replace(tmp, "\n", "", 100)
+	// Restrict title length to 100 Chars
+	if len(title) > 100 {
+		*snsMessageInput.Subject = title[0:100]
```
kostaspap, line 55: 1. You can define a local constant `const maxTitleSize = 100` and reuse it
2. Instead of `title[0:100]` you can just write `title[:100]`

internal/core/alert_delivery/outputs/sns.go
horakj: There are a few rare edge cases according to AWS SNS Documentation that can return a format error, the new line error seems to be most common. With the additional api call we can handle the edge cases which are difficult to filter out via regex or replace method, but you are right with making it cleaner.

internal/core/alert_delivery/outputs/sns.go
horakj: Thank you. I will make that adjustment.

internal/core/alert_delivery/outputs/sns.go
horakj: Thank you for the feedback, making adjustment.

internal/core/alert_delivery/outputs/sns.go
horakj: Thank you for the feedback, making adjustment.

internal/core/alert_delivery/outputs/sns.go
horakj: Right now it is a catch all control, changing to the specific syntax error would be more exact.

internal/core/alert_delivery/outputs/sns.go
horakj: Thank you for the feedback, making adjustment.

internal/core/alert_delivery/outputs/sns.go
horakj: How did that get there? Good Catch.

internal/core/alert_delivery/outputs/sns.go
nhakmiller: Either way, I do really like the "send message, try to catch the error, if there is an invalid parameter error resend with different title" approach that you've done.  Which parts are split out into the helper or not don't matter to me  too  much, but definitely keep with this approach I'd say.

internal/core/alert_delivery/outputs/sns.go
horakj: Update. GO does not like removing the Line 62 "return result, err" within the if statement  because they are deemed unused variables. Suggestions?

nhakmiller: Did you also change `:=` to `=` on  line 61? `:=` means "create new variables with the names on the left hand side and assign the values on the right hand side",  whereas  just `=` means "use existing variables with the names on the left hand side and assign to them the values on the right hand side". Kind of a subtle difference there, but one that go will be sure to remind you about.

horakj: Yep, missed that one, thank you.

pkg/testutils/awsmocks.go
```diff
@@ -374,7 +375,7 @@ type SnsMock struct {
 
 func (m *SnsMock) Publish(input *sns.PublishInput) (*sns.PublishOutput, error) {
 	args := m.Called(input)
-	if len(aws.StringValue(input.Subject)) > 100 {
+	if len(aws.StringValue(input.Subject)) > 100 || strings.Contains(aws.StringValue(input.Subject), "\n") {
 		return nil, errors.New("invalid subject")
 	}
```
kostaspap, line 380: The mock shouldn't define such functionality - this code is used by ALL of our codebase. 
In order to mock the SNS client behavior, you should write something like this in your tests: 
```
mockSns := &testutils.SnsMock{}
....
mockSns.On("Publish", mock.Anything).Return(&sns.PublishOutput{}, errors.New"Invalid subject")
```
Not sure how this if condition slipped in or who added it but it shouldn't exist at all. At least, we shouldn't be building on top of it but look to completely remove it if not in this, in a later PR.

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -99,10 +102,29 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
 		}
 	}
 
+	// Remove newlines in title
+	tmp := strings.Replace(*snsMessageInput.Subject, "\n", "", maxTitleSize)
+	// Trim title to the AWS SNS 100 char limit
+	if len(tmp) > maxTitleSize {
+		*snsMessageInput.Subject = tmp[0:maxTitleSize]
+	}
```
kostaspap, line 110: This code should be moved up after line 83 and can be something like:
```
title = strings.ReplaceAll(title, "\n", "")
if len(title) > maxTitleSize {
    title = title[:maxTitleSize-3] + "..." 
}
```
We use the three dots `...` as an indication that input has been truncated in other parts of the code as well.

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -99,10 +102,29 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
...
 	if err != nil {
-		zap.L().Error("Failed to send message to SNS topic", zap.Error(err))
-		return getAlertResponseFromSNSError(err)
+		if reqErr, ok := err.(awserr.RequestFailure); ok {
+			// Catch title edge cases and make SNS API call with generic title
+			if reqErr.StatusCode() == 400 {
+				*snsMessageInput.Subject = "New Panther Alert"
+				response, err = snsClient.Publish(snsMessageInput)
+				if err != nil {
+					zap.L().Error("Failed to send message to SNS topic", zap.Error(err))
```
kostaspap, line 120: I don't think we need to log an error here? The caller will log the error afaik.

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -99,10 +102,29 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
...
+					return getAlertResponseFromSNSError(err)
+				}
+			}
+		} else {
+			zap.L().Error("Failed to send message to SNS topic", zap.Error(err))
```
kostaspap, line 125: See comment above. I believe this should be removed.

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -99,10 +102,29 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
...
+	// Trim title to the AWS SNS 100 char limit
+	if len(tmp) > maxTitleSize {
+		*snsMessageInput.Subject = tmp[0:maxTitleSize]
+	}
 	response, err := snsClient.Publish(snsMessageInput)
+
 	if err != nil {
-		zap.L().Error("Failed to send message to SNS topic", zap.Error(err))
-		return getAlertResponseFromSNSError(err)
+		if reqErr, ok := err.(awserr.RequestFailure); ok {
```
kostaspap, line 114: awserr.Request failure is a pretty broad error here. Is there a specific code returned by AWS when we send invalid request? Then you could be checking that specifically instead of working with HTTP response codes. See e.g. 
https://github.com/panther-labs/panther/blob/0d52197c921541b62afa8f462e5196899b28d1c7/internal/core/custom_resources/resources/cloud_security_tables.go#L58

kostaspap (changes requested):

internal/core/alert_delivery/outputs/sns.go
kostaspap, line 114: This logic should have a unit test

pkg/testutils/awsmocks.go
nhakmiller: I think it adds value. This is a limitation enforced by the AWS API, if other parts of the codebase are making this same mistake isn't it better to catch it in unit testing with something like this? In my ideal world, the mocked client would always return an error whenever the actual API would return an error (based on user input at least). It seems like a more realistic mocked client can only be a good thing.

kostaspap: I see how this would be useful, however this is just a mocking library. Adding logic like this changes its purpose. The only thing that it should do is return the value that it has been instructed to return by the unit test. The moment we start adding functionality like this it stops being a mock and it becomes an in-memory SNS. 

The unit test doesn't even have send a field containing new line to verify that we handle it properly. It just has to instruct the SNS Mock to return an "InvalidInputError" (or whatever the AWS SDK returns) when the code tried to invoke a "Publish".

internal/core/alert_delivery/outputs/sns.go
horakj: We can go one step further and remove an extra line of code
`title := strings.ReplaceAll(generateAlertTitle(alert), "\n", "")`

internal/core/alert_delivery/outputs/sns.go
horakj: `if awsErr.Code() == "InvalidParameter"`

internal/core/alert_delivery/outputs/sns.go
horakj: removed as many as I could without breaking anything originally defined

nhakmiller (approved):

internal/core/alert_delivery/outputs/sns.go
kostaspap: Logging needs to be removed?

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -101,8 +108,17 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
 
 	response, err := snsClient.Publish(snsMessageInput)
 	if err != nil {
-		zap.L().Error("Failed to send message to SNS topic", zap.Error(err))
-		return getAlertResponseFromSNSError(err)
+		// Catch title edge cases and make SNS API call with generic title
+		if awsErr, ok := err.(awserr.Error); ok {
+			if awsErr.Code() == "InvalidParameter" {
```
kostaspap, line 113: We need unit tests for this functionality. 
Basically we need a new unit test that verifies that if the SNS client returns an AWS Error with code `InvalidParameter`, we make a second invocation with the appropriate parameters.

internal/core/alert_delivery/outputs/sns_test.go
```diff
@@ -110,4 +111,7 @@ func TestSendSns(t *testing.T) {
 		Permanent:  false,
 	}, result)
 	client.AssertExpectations(t)
+
+	client.On("Publish", expectedSnsPublishInput).Return(&sns.PublishOutput{}, errors.New("InvalidParameter"))
```
kostaspap, line 115: This doesn't seem to have an effect on the test.

internal/core/alert_delivery/outputs/sns.go
```diff
@@ -134,8 +149,7 @@ func (client *OutputClient) Sns(alert *alertModels.Alert, config *outputModels.S
 func buildSnsClient(awsSession *session.Session, topicArn string) (snsiface.SNSAPI, error) {
 	parsedArn, err := arn.Parse(topicArn)
 	if err != nil {
-		zap.L().Error("failed to parse topic ARN", zap.Error(err))
-		return nil, err
+		return nil, errors.Wrap(err, "failed to parse SNS topic ARN")
```
kostaspap, line 152: The caller is going to log the error - no need for logging here.

internal/core/alert_delivery/outputs/sns_test.go
```diff
@@ -93,13 +89,185 @@ func TestSendSns(t *testing.T) {
...
+	assert.Equal(t, &AlertDeliveryResponse{
+		Message:    "messageId",
+		StatusCode: 200,
+		Success:    true,
+		Permanent:  false,
+	}, result)
+	client.AssertExpectations(t)
+}
+
+func TestTruncateSnsTitle(t *testing.T) {
```
kostaspap, line 111: created new test to verify we do truncate email subjects properly.

kostaspap (approved):

PR #2066, opened 2020-11-19
Minor polling fixes

nhakmiller:
## Background

Fix two bugs reported in Cloud Security scanning systems.

The first bug was in scanning EKS clusters. We require new permissions for the audit role to scan these systems, so we updated the `PantherAuditRole`. In order to be backwards compatible, we capture permission denied errors on the relevant API call and keep scanning. Unfortunately this was not implemented correctly, so we were still logging permission denied errors when scanning EKS clusters.

The second bug is a weird one. The snapshot-poller received several requests to scan a resource of type `ELBV2.LoadBalancer` with the arn `arn:::::`. Clearly this is incorrect, so I checked the `aws-event-processor` logs and it logged some `WARN`s that it had found and extracted an invalid ARN out of a few CloudTrail logs it was processing related to the `ELBV2:ModifyLoadBalancerAttributes` and `ELBV2:DeleteLoadBalancer` API calls.  In attempting to replicate this bug I created, modified, and deleted a few load balancers but I was not able to replicate these warnings or failed scans.

The bug that I was able to fix is that in this situation it still returned an empty ARN (which is what caused the error in the  poller). Now if it detects that it extracted an invalid ARN, it drops the events instead of returning `arn:::::`. But I was not able to determine why the CloudTrail logs had an invalid ARN to begin with. So at least this should not cause an error in the future, but I have no idea what caused this malformed CloudTrail to be processed in the first place.

Closes https://github.com/panther-labs/panther-enterprise/issues/1186 and https://github.com/panther-labs/panther-enterprise/issues/1219.

Open question: should we sync this to release? I'll go with yes for now, but if anyone has any concerns let me know.

## Changes

- Fix permission denied handling when scanning EKS clusters
- Drop bad ARNs in the `aws-event-processor` for ELBV2 events
  - I looked for the same pattern in ELBV2 events in the other classifiers, and found three other instances where we were failing to parse an ARN then still returning a value. Corrected these as well.

## Testing

- Manually tested fix for EKS cluster scanning
- Added new unit test for EKS cluster scanning
- No testing for ELBV2 issue, was not able to replicate the issue. But in the future this case should be handled correctly now.

### DIFF
--- a/internal/compliance/aws_event_processor/processor/elbv2.go
+++ b/internal/compliance/aws_event_processor/processor/elbv2.go
@@ -71,7 +71,7 @@ func classifyELBV2(detail gjson.Result, metadata *CloudTrailMetadata) []*resourc
 			resourceARN, err := arn.Parse(resource.Str)
 			if err != nil {
 				zap.L().Error("elbv2: error parsing ARN", zap.String("eventName", metadata.eventName), zap.Error(err))
-				return changes
+				return nil
 			}
 			if strings.HasPrefix(resourceARN.Resource, "targetgroup/") {
 				continue
@@ -93,7 +93,7 @@ func classifyELBV2(detail gjson.Result, metadata *CloudTrailMetadata) []*resourc
 			lbARN, err := arn.Parse(lb.Get("loadBalancerArn").Str)
 			if err != nil {
 				zap.L().Error("elbv2: error parsing ARN", zap.String("eventName", metadata.eventName), zap.Error(err))
-				return changes
+				return nil
 			}
 			changes = append(changes, &resourceChange{
 				AwsAccountID: metadata.accountID,
--- a/internal/compliance/aws_event_processor/processor/elbv2.go
+++ b/internal/compliance/aws_event_processor/processor/elbv2.go
@@ -130,7 +130,7 @@ func classifyELBV2(detail gjson.Result, metadata *CloudTrailMetadata) []*resourc
 	}
 
 	if parseErr != nil {
-		zap.L().Warn("elbv2: error parsing ARN", zap.String("eventName", metadata.eventName), zap.Any("event", metadata), zap.Error(parseErr))
+		zap.L().Error("elbv2: error parsing ARN", zap.String("eventName", metadata.eventName), zap.Any("event", metadata), zap.Error(parseErr))
 		return nil
 	}
 	return []*resourceChange{{
--- a/internal/compliance/aws_event_processor/processor/elbv2.go
+++ b/internal/compliance/aws_event_processor/processor/elbv2.go
@@ -130,7 +130,7 @@ func classifyELBV2(detail gjson.Result, metadata *CloudTrailMetadata) []*resourc
 	}
 
 	if parseErr != nil {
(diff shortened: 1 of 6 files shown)


internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -121,6 +121,12 @@ func getEKSFargateProfiles(eksSvc eksiface.EKSAPI, clusterName *string) ([]*awsm
 		})
 
 	if err != nil {
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) && awsErr.Code() == "AccessDeniedException" {
```
nhakmiller, line 125: I had to switch this from `eks.ErrorCodeAccessDenied` to `"AccessDeniedException"` because sadly `eks.ErrorCodeAccessDenied` is just `AccessDenied`, and the actual code returned by AWS is `AccessDeniedException` 😞

internal/compliance/snapshot_poller/pollers/aws/eks_cluster_test.go
```diff
@@ -103,6 +105,36 @@ func TestEksClusterDescribeError(t *testing.T) {
...
+func TestEksClusterGetProfiles(t *testing.T) {
+	mockSvc := awstest.BuildMockEksSvc([]string{"ListFargateProfilesPages", "DescribeFargateProfile"})
+
+	out, err := getEKSFargateProfiles(mockSvc, awstest.ExampleEksClusterNameMulti)
+	require.NoError(t, err)
+	assert.NotNil(t, out)
+	assert.Len(t, out, 2)
+}
+
+func TestEksClusterGetProfilesPermissionsError(t *testing.T) {
```
nhakmiller, line 117: This is the unit test that mocks the "AccessDenied" response handling. The other two unit tests are just related to normal behavior.

internal/compliance/aws_event_processor/processor/elbv2.go
```diff
@@ -130,7 +130,8 @@ func classifyELBV2(detail gjson.Result, metadata *CloudTrailMetadata) []*resourc
 	}
 
 	if parseErr != nil {
-		zap.L().Warn("elbv2: error parsing ARN", zap.String("eventName", metadata.eventName), zap.Error(parseErr))
+		zap.L().Warn("elbv2: error parsing ARN", zap.String("eventName", metadata.eventName), zap.Any("event", metadata), zap.Error(parseErr))
```
kostaspap, line 133: All the above reporting ARN parsing problems with error - this does with warn. I'm not sure what's appropriate here, I think we should be consistent?

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
kostaspap: ufff, we did encounter such issue recently with DDB exceptions...

kostaspap (approved):

internal/compliance/aws_event_processor/processor/elbv2.go
nhakmiller: Good point, I'll update them to all log an `error`. Apparently this happens at least occasionally so it would be good to know how often and  maybe find out why.

PR #2068, opened 2020-11-19
Refactor glue schema generation and decouple logtypes module from awsglue 

alxarch:
## Background

The code that maps the fields of go structs to Glue columns has been a source of bugs that appeared when we were adding new log types to Panther. We kept patching it to solve each bug that came up. That was fine until the introduction of custom logs in EE. Now any customer can define their own log schemas and such bugs are now 'out of our control'. 

At the same time the `awsglue` module was handling anythinh Glue-related leading to cyclic dependencies that impeded new feature development. Especially the dependency between `logtypes` and `awsglue` was a major blocker in various cases. 

This PR introduces `glueschema` module that handles conversion of `reflect.Type` values to Glue columns. The code was rewritten from scratch to handle any struct gracefully. It returns errors with path information instead of panicking and links recursive function calls in a way that conversion logic is not scattered (or partially applied) on each case (struct, slice, map etc).

Over the process of rewriting this part, it became apparent that the previous code was not handling case sensitive columns in a predictable way. Column names in AWS Glue [are always lower-case and have certain other restrictions](https://docs.aws.amazon.com/athena/latest/ug/glue-best-practices.html#schema-names). This means that we *cannot* have a row with two columns that have the same lower-case name (i.e. `userAgent`, `UserAgent`). The same applies for sibling fields of a `struct<...>` value. Because of this Glue behavior the default JSON SerDe library fails to read JSON data if two fields in a JSON object have the same name when converted to lower-case. To overcome this we use the `case.insensitive=false` flag for the [JSON SerDe variant we use](https://docs.aws.amazon.com/athena/latest/ug/json-serde.html#openx-json-serde). But this comes with the cost of having to provide name mappings for all field names so that JSON SerDe knows how to map JSON field names to Glue column names. In the previous implementation, such mappings were _not applied in depth-first order_. This means that if a struct had a top-level field of `Application` and then at some nesting level a struct had a field named `application`, the mapping would end up as `mapping.application=application`, 'shadowing' the top level column. The behavior was unpredictable because it depended on the order in which fields were defined. 
**TL;DR**: this PR returns an error if two columns/fields share the same lower-case name at the same level and the JSON SerDe mappings are set by depth-first priority so that 'parent' mappings are never shadowed by 'child' ones.

## Changes

- Introduces `gluechema` package to handle conversion of `reflect.Type` to AWS Glue schema types.
- Fixes case sensitive mappings for JSON SerDe to use a depth-first priority
- Adds transliteration for column names to avoid non-ASCII letters messing up the Queries.
- Adds more rules for symbol expansion when mapping JSON field names to Glue column names.
- Falls back to `_` for any character known to not work for column names.
- Uses better detection for 'required' columns by checking the `validate` tag (only relevant for doc generation).
- Decouples `logtypes` module from `awsglue`
- Cleans up `registry` module to pave the way for `cloudsecurity` work. We keep only a limited set for functions using the 'Native' prefix so that use of `registry` is not ambiguous.

## Testing
- E2E test by deploying and checking tables and views
- Pass all unit tests for the previous schema generation code.
- Add more test cases to ensure errors are returned when column names collide
- Checked all 'native' log types are being registered without errors and that all schemas produced are identical (except for the required flags).
- Verified that the new JSON SerDe mapping method handles field 'shadowing' in a predictable way (depth-first).


internal/log_analysis/awsglue/glueschema/columns.go
```diff
@@ -0,0 +1,58 @@
...
+	if result == name {
+		return name
+	}
+	return strings.Trim(result, "_")
+}
+
+// TODO: [glueschema] Add more mappings of invalid Athena field name characters here
+// NOTE: The mapping should be easy to remember (so no ASCII code etc) and complex enough
+// to avoid possible conflicts with other fields.
+var columnNamesReplacer = strings.NewReplacer(
```
kostaspap, line 52: What about the $ symbol?

alxarch: Good catch that  got dropped in the merge probably.

alxarch: Fixed

deployments/log_analysis.yml
```diff
@@ -84,10 +84,6 @@ Parameters:
     Description: KMS key ID for SQS encryption
     # Example: "484fb80c-4ae5-40d0-b22a-bdd5d0953b3e"
     AllowedPattern: '^[0-9a-f-]{36}$'
-  TablesSignature:
-    Type: String
-    Description: Value from gluetable.DeployedTablesSignature() or the Panther version if using CF
-    MinLength: 1
```
kostaspap, line 90: Why is this removed?

internal/log_analysis/awsglue/glueschema/glueschema.go
```diff
@@ -0,0 +1,356 @@
...
+
+func inferTypeColumns(typ reflect.Type, names map[string][]string) ([]Column, error) {
+	cols, err := inferStructColumns(typ, nil, names)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	return cols, nil
+}
+
+func inferStructColumns(typ reflect.Type, path []string, names collisions) ([]Column, error) {
```
kostaspap, line 111: What is the `path`? Can you add description/comments that explain the logic?

internal/log_analysis/awsglue/glueschema/glueschema.go
```diff
@@ -0,0 +1,356 @@
...
+func InferColumnsWithMappings(schema interface{}) ([]Column, map[string]string, error) {
+	if schema == nil {
+		return nil, nil, errors.New("nil schema value")
+	}
+	names := collisions{}
+	columns, err := inferTypeColumns(reflect.TypeOf(schema), names)
+	if err != nil {
+		return nil, nil, err
+	}
+	mappings := names.caseSensitiveMappings()
```
kostaspap, line 99: The mappings AFAIK are no needed only for collisions but for all fields but I might be wrong. Have you verified this works E2E?

internal/log_analysis/awsglue/glueschema/glueschema.go
```diff
@@ -0,0 +1,356 @@
...
+		field := typ.Field(i)
+		if field.Anonymous {
+			// Possible embedded struct, extend the fields
+			fields = appendStructFieldsJSON(fields, field.Type)
+			continue
+		}
+		// Unexported field
+		if field.PkgPath != "" {
+			continue
+		}
```
kostaspap, line 157: Can you elaborate? Why are we excluding it?

internal/log_analysis/awsglue/glueschema/glueschema.go
```diff
@@ -0,0 +1,356 @@
...
+	for i := range fields {
+		field := &fields[i]
+		fieldPath := append(path, field.Name)
+		col := Column{}
+		if err := inferColumn(&col, field, fieldPath, names); err != nil {
+			return nil, err
+		}
+		if col.Name == "" {
+			continue
+		}
```
kostaspap, line 130: Interesting. Can this really happen? Shouldn't we just throw an error if that happens, why allow it?

internal/log_analysis/awsglue/glueschema/glueschema.go
```diff
@@ -0,0 +1,356 @@
...
+
+// isFieldRequired checks whether a field is required or not.
+func isFieldRequired(field *reflect.StructField) bool {
+	tag, hasValidate := field.Tag.Lookup("validate")
+	if !hasValidate {
+		// If the field does not have a 'validate' tag, it is not required
+		return false
+	}
+	// Otherwise, if the 'validate' tag contains 'omitempty' it is not a required field.
+	return !strings.Contains(tag, "omitempty")
```
kostaspap, line 226: Seems this should be "return True".

A field might as well be:
```
Field string `json:",omitempty" validate:"required"`
```
in which case we should mark it as required.

internal/log_analysis/awsglue/glueschema/glueschema.go
```diff
@@ -0,0 +1,356 @@
...
+		return ArrayOf(elem), nil
+	case reflect.Map:
+		key, err := inferColumnType(typ.Key(), path, nil)
+		if err != nil {
+			return "", err
+		}
+		// Make sure we fail if map key is something exotic
+		if key != TypeString {
+			// We do not want stack for errors
+			return "", newSchemaError(path, "invalid map key %q", key)
```
kostaspap, line 259: I would expand to something like "...only strings allowed as map keys"

internal/log_analysis/awsglue/glueschema/glueschema.go
```diff
@@ -0,0 +1,356 @@
...
+		}
+		// Distinguish map values in path
+		path = append(path, "[]")
```
kostaspap, line 262: Should this be `{}` or something? `[]` is used for arrays.

internal/log_analysis/awsglue/glueschema/glueschema_test.go
```diff
@@ -236,27 +230,71 @@ func TestInferJsonColumns(t *testing.T) {
 		{Name: "CustomSliceField", Type: "baz", Comment: "test field"},
 		{Name: "CustomStructField", Type: "bar", Comment: "test field"},
 	}
-	expectedStructFieldNames := []string{"A", "B", "C", "Field1", "Field2", "InheritedField", "at_sign_remap"}
+	expectedStructFieldNames := map[string]string{
+		"a":                      "A",
+		"at_sign_remap":          "at_sign_remap",
```
kostaspap, line 235: can you add test cases for the other conversions as well?

internal/log_analysis/awsglue/glueschema/glueschema.go
```diff
@@ -0,0 +1,356 @@
...
+				_, exists := out[mapping]
+				if !exists {
+					break
+				}
+			}
+			out[mapping] = caseSensitiveName
+		}
+	}
+	return out
+}
```
kostaspap, line 356: 1. Tests for this? Seems like it should be unit tested
2. Have you verified E2E this functionality?

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
```diff
@@ -79,3 +81,21 @@ func (h *LambdaHandler) fetchAllDeployedLogTypes(ctx context.Context) ([]string,
...
+		entry, err := r.Resolve(ctx, name)
+		if err != nil {
+			return nil, err
+		}
+		if entry == nil {
+			return nil, errors.Errorf("unresolved log type %q", name)
+		}
+		eventSchema := entry.Schema()
+		desc := entry.Describe()
+		meta := awsglue.NewGlueTableMetadata(models.LogData, desc.Name, desc.Description, awsglue.GlueTableHourly, eventSchema)
```
kostaspap, line 97: Hmmmm, why `models.LogData` here? or it doesn't matter in this context? Can you add a comment to that extend?

internal/log_analysis/log_processor/parsers/numerics/numerics.go
```diff
@@ -19,11 +19,19 @@ package numerics
...
 	"strconv"
 
 	"github.com/pkg/errors"
+
+	"github.com/panther-labs/panther/internal/log_analysis/awsglue/glueschema"
 )
 
+func init() {
+	glueschema.MustRegisterMapping(reflect.TypeOf(Integer(0)), glueschema.TypeBigInt)
+	glueschema.MustRegisterMapping(reflect.TypeOf(Int64(0)), glueschema.TypeBigInt)
```
kostaspap, line 32: We've had this discussion before - we didn't want to do the registration here but do it in a single place. We'll eventually have `snowflakeschema` - should we go and update every "type" to make sure it is registered as well?

internal/log_analysis/log_processor/registry/internal/generate.go
```diff
@@ -132,7 +132,5 @@ func init() {
        {{.Name}}.LogTypes(),
 {{ end }}
 	)
-	// Register all log types in the group with the availableLogTypes
-	availableLogTypes.MustRegister(nativeLogTypes)
```
kostaspap, line 136: 👍

kostaspap (changes requested):
This is a good change - but also touches critical parts of the system. How has this been tested E2E? I am mostly worried for the changes in the `mappings`. 

One way I can think of to compare this with the old method is by generating the schemas in the docs - if there's any difference compared to docs generated by the old version, there's an issue :)

deployments/log_analysis.yml
alxarch: Talked to russ about it. It is not used anymore.

internal/log_analysis/awsglue/glueschema/glueschema.go
alxarch: Path, is the field path inside the struct tree. We pass it as an arg so we can get helpful error messages when things go wrong. Since this code runs recursively, a stack trace is useless for discovering which field produced the error.  Normal error wrapping would end up with something like: 
```
failed to parse struct "Foo": failed to parse slice "Bar": failed to parse slice element "Bar[]": failed to parse struct field "Baz": invalid field type "function(string){}"
```
With `path` and `newSchemaError` we get:
```
schema error at "Foo.Bar[].Baz": invalid field type "function(string)"
```

internal/log_analysis/awsglue/glueschema/glueschema.go
alxarch: This produces mappings for all field names. `names` contain all field names discovered in the schema, not only the colliding ones.

internal/log_analysis/awsglue/glueschema/glueschema.go
alxarch: unesported fields are not serializable to/from JSON (unless they are embedded structs, which is covered before this check.

internal/log_analysis/awsglue/glueschema/glueschema.go
kostaspap: You posted few seconds before me it seems :) 
I think it's still useful to have a brief comment about this in the description so readers don't have to go deep into the code to understand its usefulness!

internal/log_analysis/awsglue/glueschema/glueschema.go
kostaspap: A test would be useful as well

internal/log_analysis/awsglue/glueschema/glueschema.go
alxarch: Yes, it is the skip case. I am updating the `fieldColumnName` function to be more readable.

internal/log_analysis/awsglue/glueschema/glueschema.go
alxarch: The example above is not valid for schema purposes. 

Fields marked "omitempty" in JSON, must use "omitempty" for validate also, otherwise they should not have `omitempty` in JSON in the first place.

Nevertheless, the field gets marked as required, appropriately. The `tag` in that function is not `json`, it is `validate`.

What this check tries to achieve, is to catch most `validate` cases where `required` is not used explicitly. 

This was a mistake in the old code. 

If I use
```go
type T struct {
   Foo sting `validate:"eq=Foo"`
}
```
then field `Foo` is required, implicitly.

Keep in mind that `Column.Required` is only relevant for generating docs.

internal/log_analysis/awsglue/glueschema/glueschema.go
alxarch: `[]` is used to index maps as well. For example `foo["bar"]`. The `[]` is meant as a helper to understand that the problematic path is inside the _element_ of  a map/array, not on the field itself. This is only relevant to the error messages displayed so I wouldn't worry about it.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
alxarch: This is one of the complexities that needs to be addressed in a separate PR for cloudsecurity. The place where this function is used only cares about the 'base' table. It 'magically' derives the `rule_errors` and `rule_matches` tables inside `gluetables` and `athenaviews` functions. This is a mistake in code structure IMO. The helpers in those packages should be oblivious to the connections between tables and only use explicit arguments for names, columns etc. We should investigate better logic separation in a subsequent PR. The scope for this is to achieve decoupling.

internal/log_analysis/log_processor/parsers/numerics/numerics.go
alxarch: Yes, this is the only way we won't end up with cyclic dependencies.
Once we get to `snowflakeschema`, the translation needs to happen based on `logschema` and not reflection. All this is a temporary, complex solution based on `reflect` until then.

alxarch: The `numerics` should be removed anyway. Did not want to affect the relevant parsers to do it in this PR.

internal/log_analysis/awsglue/glueschema/glueschema.go
alxarch: 1. Will add a unit test for each case on the `ColumnName` function. For the `InferColumns`, even a single case suffices. 
2. E2E tests for this are my main task for the day, once complete the PR will be undrafted.

internal/log_analysis/awsglue/glueschema/glueschema_test.go
alxarch: Will add on the `ColumnName function`

internal/log_analysis/awsglue/glueschema/glueschema.go
alxarch: fixed

internal/log_analysis/awsglue/glueschema/glueschema.go
kostaspap: Yup, I realized it's for the error message. Was just thinking that maybe having the same symbol both for arrays and maps might be confusing to the user.  Not strong for this one though, up to you.

alxarch:
After some tests on the current Panther code I concluded that we don't really support case sensitive column names. Glue only accepts [lower case column names](https://docs.aws.amazon.com/athena/latest/ug/glue-best-practices.html#schema-names).
JSON SerDe (the variant we use) allows to have case sensitive mappings from JSON to Glue columns. This actually requires to use different column names in Glue than the field names in JSON to be able to support names that collide (i.e. `userAgent` - `UserAgent`). The collisions only occur at the same depth (i.e. `userAgent` and `event.userAgent` do not collide). With all this combined *it is not possible to support 'transparent' mappings for column names that collide*. What we try to do, is avoid JSON SerDe from failing to parse `jsoniter.RawMessage` because of colliding names. But to keep the 'transparent' mapping of column names, we **need to enforce that no colliding column names exist at the same depth for the defined part of the schema**. 
This PR fails to infer Glue columns if such collisions exist. This is the only way to have a predictable behavior for schemas when Querying with Athena. The schemas defined in OSS all seem to comply. We will have to also check the EE-only ones during sync.

internal/log_analysis/awsglue/glueschema/glueschema.go
alxarch: Resolved using a different strategy. See below

alxarch:
Updated the `ColumnName` function to transliterate non-ASCII letters and to handle any non-valid character not in our named character definitions by replacing it with `_`.

internal/log_analysis/awsglue/glueschema/glueschema.go
alxarch: The mapping strategy has changed, see comment below.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
kostaspap: Can you add a comment in the code describing that the `LogData` is really irrelevant here?

internal/log_analysis/log_processor/logtypes/entry.go
```diff
@@ -271,7 +258,7 @@ func checkGlue(schema interface{}) (err error) {
 			}
 		}
 	}()
-	cols, _ := awsglue.InferJSONColumns(schema, awsglue.GlueMappings...)
+	cols, _, _ := glueschema.InferColumnsWithMappings(schema)
```
kostaspap, line 261: Just `glueschema.InferColumns(schema)`.

internal/log_analysis/awsglue/glueschema/columns.go
```diff
@@ -0,0 +1,103 @@
...
+	Comment string
+	// Required marks the column as required.
+	// This information is used in documentation to mark fields as required.
+	// It does not affect Glue schema in some way and this should be removed once
+	// doc generation evolves to not use []glueschema.Column as input.
+	Required bool
+}
+
+// ColumnName normalizes names to be used for Glue table columns
+func ColumnName(name string) string {
```
kostaspap, line 46: We need to use the same renaming strategy when writing the data to S3 as JSON we follow here. I don't think the code does it currently.

kostaspap (commented):
Can you add more information? How did you verify that the new mechanism outputs the same Glue schemas as the previous mechanism?

internal/log_analysis/awsglue/glueschema/columns.go
alxarch: We do use `ColumnName` in the `ConfigForDataLakeWriters`

internal/log_analysis/log_processor/logtypes/entry.go
alxarch: fixed

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
alxarch: fixed

internal/log_analysis/awsglue/glueschema/glueschema.go
alxarch: Added comments for the path argument wherever it is passed to a recursive function.

alxarch:
> Can you add more information? How did you verify that the new mechanism outputs the same Glue schemas as the previous mechanism?

The new mechanism does not output the same schemas as the old code. The old code had problematic cases. 

What I will try to verify though is that any changes that might show up in generated docs won't be backwards incompatible (i.e. datacatalog updater will have problems). This is hard because we moved docs to EE.
Also I am in the middle of E2E testing this on a clean deployment.

kostaspap:
> > Can you add more information? How did you verify that the new mechanism outputs the same Glue schemas as the previous mechanism?
> 
> The new mechanism does not output the same schemas as the old code. The old code had problematic cases.

Where are the differences? Are the differences in any of the built-in data types or in specific Custom logs usecases? See my comment on how we can verify that the new schemas are the same as the old ones. See my [comment](https://github.com/panther-labs/panther/pull/2068#pullrequestreview-535192732) on how we could do that - of course you could do it in a different way if we find one such. 

If the schemas coming out of the new mechanism are different from the ones output by the previous mechanism, this will make the data unqueryable. We definitely don't want that.

alxarch:
> > > Can you add more information? How did you verify that the new mechanism outputs the same Glue schemas as the previous mechanism?
> > 
> > 
> > The new mechanism does not output the same schemas as the old code. The old code had problematic cases.
> 
> Where are the differences? Are the differences in any of the built-in data types or in specific Custom logs usecases? See my comment on how we can verify that the new schemas are the same as the old ones. See my [comment](https://github.com/panther-labs/panther/pull/2068#pullrequestreview-535192732) on how we could do that - of course you could do it in a different way if we find one such.
> 
> If the schemas coming out of the new mechanism are different from the ones output by the previous mechanism, this will make the data unqueryable. We definitely don't want that.

If there are changes in comments, or order of fields, there is no problem to be had. if *types* change for the native logs, we do have a problem but to fix it we should alter the structs to output the correct type, not the schema generation to have wrong logic. I will make a throwaway branch to bring over doc generation from EE so that I can compare.

kostaspap:
> > > > Can you add more information? How did you verify that the new mechanism outputs the same Glue schemas as the previous mechanism?
> > > 
> > > 
> > > The new mechanism does not output the same schemas as the old code. The old code had problematic cases.
> > 
> > 
> > Where are the differences? Are the differences in any of the built-in data types or in specific Custom logs usecases? See my comment on how we can verify that the new schemas are the same as the old ones. See my [comment](https://github.com/panther-labs/panther/pull/2068#pullrequestreview-535192732) on how we could do that - of course you could do it in a different way if we find one such.
> > If the schemas coming out of the new mechanism are different from the ones output by the previous mechanism, this will make the data unqueryable. We definitely don't want that.
> 
> If there are changes in comments, or order of fields, there is no problem to be had. if _types_ change for the native logs, we do have a problem but to fix it we should alter the structs to output the correct type, not the schema generation to have wrong logic. I will make a throwaway branch to bring over doc generation from EE so that I can compare.

Agree with that overall. Only one thing I don't agree: 
>  if _types_ change for the native logs, we do have a problem but to fix it we should alter the structs to output the correct type, not the schema generation to have wrong logic. 

This will change the scope of this PR though since it will require data migration scripts to be in place. I don't want to do this now.

alxarch:
After allowing '-' the docs are the same except for the required columns that don't affect the Glue stuff. We had some columns that were required but not showing as such in the docs.

internal/log_analysis/awsglue/glueschema/columns_test.go
```diff
@@ -0,0 +1,49 @@
...
+	for _, tc := range []testCase{
+		{"@foo", "at_sign_foo"},
+		{"foo,bar", "foo_comma_bar"},
+		{"`foo`", "backtick_foo_backtick"},
+		{"'foo'", "apostrophe_foo_apostrophe"},
+		{"foo.bar", "foo_bar"},
+		{".foo", "_foo"},
+		{"foo-bar", "foo-bar"},
+		{"$foo", "dollar_sign_foo"},
+		{"συνθηματικό", "synthimatiko"},
```
kostaspap, line 42: I was hoping to see `Μύκονοοοος` in here since it seems to be our go-to Greek reference in log processing :P

kostaspap (approved):

internal/log_analysis/athenaviews/athenaviews.go
```diff
@@ -169,7 +170,7 @@ func newPantherViewColumns(tables []*awsglue.GlueTableMetadata, extraColumns []a
 }
 func (pvc *pantherViewColumns) inferViewColumns(table *awsglue.GlueTableMetadata, extraColumns []awsglue.Column) {
 	// NOTE: in the future when we tag columns for views, the mapping  would be resolved here
-	columns, _ := awsglue.InferJSONColumns(table.EventStruct(), awsglue.GlueMappings...)
+	columns, _ := glueschema.InferColumns(table.EventStruct())
```
kostaspap, line 173: We shouldn't ignore the error here. This points to a problem.

alxarch: The main problem is that this code does not return an error in the first place. I can only panic, or refactor this as well.
The error won't be a problem here since we have already checked the log types on registration/creation.

kostaspap: The `glueschema.InferColumns` returns an error . Even if it now panics instead of returning one, this behavior might be altered in the future. 
The `InferColumns` will change overtime and it will rely on the callers to handle errors appropriately.

alxarch: refactored to return errors

kostaspap (approved):

internal/log_analysis/awsglue/glueschema/columns_test.go
alxarch: fixed

Issue #2001, opened 2020-11-10
Improve stream processing and memory usage of log processor Lambda


alxarch:
### Description

The log processor currently has some major memory-related issues 

- ~We never call `Close()` on S3 object's `Body io.ReadCloser`, possibly leaking http connections to S3 service~ Fixed by use of `s3pipe.Downloader`
- ~We use multiple chained buffered readers (`bufio.Reader`) to read each log file which results in unnecessary data copying in memory.~ Not relevant any more with the `s3pipe.Downloader`
- We have a very crude reading method for log entries that in some cases (CloudTrail) requires to load the full file in memory before we can process it. This forces us to set 'fat' Lambda memory requirements to avoid OOM crashes thus increasing processing costs for all logs.

### Related Services

- Log processor

### Designs

To solve these issues we should abstract the reading of log entries from a file using the following interface:

```golang
// LogEntryStream reads log entries from an underlying `io.Reader`.
type LogEntryStream interface {
  // NextLogEntry returns the next log entry in the stream.
  // If `nil` is returned, it means that an error was encountered while reading or the EOF was reached 
  // and there will be no more log entries in the stream.
  //
  // WARNING: To avoid needless memory copying the byte slice returned is only valid until the next call to NextLogEntry or Close
  NextLogEntry() []byte
  // Close releases any memory or resource handlers managed by the stream and returns any error that occured while parsing
  // Once Close is called all future calls must return the same result and any calls to NextLogEntry should return nil
  Close() error
}
```

To produce a `LogEntryStream` we will be using various `LogEntryStreamFactory` implementations depending on the requirements of the particular log file as determined by it's metadata (i.e. source configuration, object key pattern etc).

```golang
type LogEntryStreamFactory interface {
  BuildLogEntryStream (rc io.ReadCloser) (LogEntryStream, error)
}
```

> NOTE: Detection and decompression of `gzip` log streams should happen **before** passing the initial `rc` argument to the appropriate factory. Special care must be taken so that `rc` *also closes* the original `Body  io.ReadCloser` of the S3 object (Handling this transparently is not currently possible without extensive refactoring of the log processor code).

[...]

PR #2071, opened 2020-11-19
Fix connection reset by peer errors in log processor

rleighton:
## Background

Some customers with large numbers of large (> 150MB) files are seeing regular failures due to "connection  reset by peer" errors.

This seems to be common issue with S3 under load and it will simply cut the connection to shed load.

The fix for this is to switch from using `GetObject` to a `Downloader`. The downloader will catch the errors and retry.
In order to use the downloader we need to implement a buffer that can be re-written on error.

I tested this with 162MB compressed (1G uncompressed) s3 access log files. Before this patch there would be failures when there were more than 5 in flight. After this change the errors stopped. 

The load tests revealed a number of opportunities for performance improvements (see changes below).

This is at ~2k files 162MB  per hour (7.7TB / day):

image

Roughly ~4k files 162MB  per hour (~15TB / day):

image

## Changes

- Switch using `GetObject` to a `s3pipe`. 
- Add gradient and dc condition to `rescale`
- Fix error logging for canceled context operations
- Reduce the max time to poll sqs to 1/4 the 15min limit to give more processing time for big files
- Tune uploader concurrency and part size
- Optimize finding largest / oldest buffer in uploader
- Add object size to `s3queue`
- Add max bytes read to processor limits.

## Testing

- mage test:ci
- mage deploy plus load test using 160MB s3access logs


austinbyers (approved):
Great idea, using the `s3.DownloadManager` instead of the `s3.GetObject`!

I don't fully understand why we had to wrap our own IO interface around it, though. Would a [`bytes.Buffer`](https://golang.org/pkg/bytes/#Buffer) be sufficient instead?

internal/log_analysis/log_processor/downloadpipe/downloadpipe.go
```diff
@@ -0,0 +1,109 @@
...
+var downloadPipePool = sync.Pool{
+	New: func() interface{} {
+		return &DownloadPipe{
+			buffer: make([]byte, 0, DownloadPartSize),
+		}
+	},
+}
+
+func NewDownloadPipe(d *s3manager.Downloader) *DownloadPipe {
+	d.Concurrency = 1
```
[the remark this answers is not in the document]
rleighton: Good question. Data comes in chunks and we can't afford to read it all into ram and even if we could we would like to stream in it.  Most data is gzip'd so we need to read a chuck at a time in order.

rleighton:
> Great idea, using the `s3.DownloadManager` instead of the `s3.GetObject`!
> 
> I don't fully understand why we had to wrap our own IO interface around it, though. Would a [`bytes.Buffer`](https://golang.org/pkg/bytes/#Buffer) be sufficient instead?

Good question! So when the downloader is reading it will be writing using WriteAt() which expects to write an offset. If there is a failure it re-writes over the data. So we have to defer reading the chuck until it is done.

internal/log_analysis/log_processor/downloadpipe/downloadpipe.go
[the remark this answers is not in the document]
austinbyers: Why would a greater concurrency result in more RAM usage? It's not going to go beyond the bounds of the buffer no matter how much concurrency you have?

Either way, your explanation seems like a good comment to put in the source

rleighton:
Will do. Look at the WriteAt() interface. That’s the reason

internal/log_analysis/log_processor/downloadpipe/downloadpipe.go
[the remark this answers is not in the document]
rleighton: Done

internal/log_analysis/log_processor/downloadpipe/downloadpipe.go
```diff
@@ -0,0 +1,109 @@
...
+// these buffers are expensive, use a pool
+var downloadPipePool = sync.Pool{
+	New: func() interface{} {
+		return &DownloadPipe{
+			buffer: make([]byte, 0, DownloadPartSize),
+		}
+	},
+}
+
+func NewDownloadPipe(d *s3manager.Downloader) *DownloadPipe {
```
kostaspap, line 49: Better pass the s3client here and let the method create an S3 Downloader.

internal/log_analysis/log_processor/processor/rescale.go
```diff
@@ -117,3 +131,7 @@ func processingScaleUp(ctx context.Context, lambdaClient lambdaiface.LambdaAPI,
 		}
 	}
 }
+
+func IsCanceled(err error) bool {
+	return err == context.Canceled || err == context.DeadlineExceeded || awsutils.IsAnyError(err, request.CanceledErrorCode)
+}
```
kostaspap, line 137: It seems that just the last check is required in this case?

rleighton: Probably. The last I added because I saw it in testing. The ones before you put there. Remove them?

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -85,6 +85,14 @@ func Process(
 				if err := processor.run(resultsChannel); err != nil {
 					return err
 				}
+				// at this point the file has been processed
```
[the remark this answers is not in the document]
rleighton: Good catch.

internal/log_analysis/log_processor/common/common.go
```diff
@@ -39,6 +40,10 @@ import (
 )
 
 const (
+	// NOTE: max buffer size in ram is 50M
+	UploaderConcurrency = 5
+	UploaderPartSize    = 10 * 1024 * 1024
```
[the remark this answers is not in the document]
rleighton: Yeah, will refactor per your suggestions.

internal/log_analysis/log_processor/processor/rescale.go
kostaspap: Yup, remove

internal/log_analysis/log_processor/common/common.go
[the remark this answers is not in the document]
rleighton: Done

internal/log_analysis/log_processor/processor/processor.go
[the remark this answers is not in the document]
rleighton: Fixed

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -150,20 +152,38 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
...
 
+		_, err = downloader.Download(downloadPipe, getObjectInput)
+		if err != nil {
+			err = errors.Wrapf(err, "Download() failed for s3://%s/%s", s3Object.S3Bucket, s3Object.S3ObjectKey)
+			zap.L().Error("s3 download failed", zap.Error(err))
+			closeErr = downloadPipe.CloseWriterWithError(err) // this will cause the reader to fail
+		} else {
+			closeErr = downloadPipe.CloseWriter()
+		}
+	}()
```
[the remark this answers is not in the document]
rleighton: Done

internal/log_analysis/log_processor/downloadpipe/downloadpipe.go
rleighton: Done

internal/log_analysis/log_processor/processor/rescale.go
rleighton: Done

internal/log_analysis/log_processor/downloadpipe/downloadpipe.go
```diff
@@ -0,0 +1,133 @@
...
+		if closeErr != nil {
+			zap.L().Warn("s3 download pipe close failed", zap.Error(closeErr))
+		}
+	}()
+
+	_, err := dp.downloader.Download(dp, dp.getObjectInput)
+	if err != nil {
+		err = errors.Wrapf(err, "Download() failed for s3://%s/%s",
+			*dp.getObjectInput.Bucket, *dp.getObjectInput.Key)
+		zap.L().Error("s3 download failed", zap.Error(err))
```
kostaspap, line 85: Logging here is not needed right? It will be logged by the reader.

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -72,6 +72,18 @@ func Process(
 		// Process streams serially to keep memory requirements low
 		processStreams = func() error {
 			defer close(resultsChannel)
+			// ensure we close the streams before returning to free resources
+			defer func() {
+				for dataStream := range dataStreams {
```
kostaspap, line 77: dataStreams is a channel - I don't think this will work.

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -150,20 +152,38 @@ func readS3Object(s3Object *S3ObjectInfo) (dataStream *common.DataStream, err er
...
-	contentType, err := detectContentType(bufferedReader)
+	downloadPipe := downloadpipe.NewDownloadPipe(downloader)
+
+	// run the downloader in the background writing into the pipe
+	go func() {
+		// while not important to processing, errors on close could indicate other issues
+		var closeErr error
+		defer func() {
+			if closeErr != nil {
+				zap.L().Error("s3 download pipe close failed", zap.Error(closeErr))
```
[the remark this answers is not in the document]
rleighton: Done

internal/log_analysis/log_processor/downloadpipe/downloadpipe.go
rleighton: Correct. I was being overly careful since this new code. I'll add a comment that we can remove in the future

internal/log_analysis/log_processor/processor/processor.go
rleighton: Ugh! You are right.

internal/log_analysis/log_processor/common/common.go
```diff
@@ -39,6 +40,10 @@ import (
 )
 
 const (
+	MaxS3BufferSizeBytes = 50 * 1024 * 1024
+	UploaderPartSize     = 18 * 1024 * 1024 // the size of this affects performance
```
kostaspap, line 44: The [default](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/#pkg-constants) is 5MB so I would expect it to be more performant? Or you think the issue is that having many parts slows it down somehow?

kostaspap (commented):
Is this going to work with CT where we have to pull more data than the Pipe's buffer size? It it possible that the pipe's buffer will "choke" in such case?

internal/log_analysis/log_processor/common/common.go
rleighton: This is just based on playing with it but 5MB was too small, when I upped it then it was a lot faster. My guess is there is a sweet spot.

internal/log_analysis/log_processor/processor/processor.go
rleighton: Fixed it. Made a new function, cleaner also. I was just being impatient. I'm going to let this code burn in over the weekend along with your changes for error handling under high load.

rleighton:
> Is this going to work with CT where we have to pull more data than the Pipe's buffer size? It it possible that the pipe's buffer will "choke" in such case?

Nope because the buffer == partSize. That's the trick here. I'm testing with 160MB files.

kostaspap:
> > Is this going to work with CT where we have to pull more data than the Pipe's buffer size? It it possible that the pipe's buffer will "choke" in such case?
> 
> Nope because the buffer == partSize. That's the trick here. I'm testing with 160MB files.

The 160MB files however contain new line separated events - so the datastream.ReadString('\n') never has to return 160MB. 
In case of CloudTrail it has to return the full payload and keep it in memory.

kostaspap:
> > > Is this going to work with CT where we have to pull more data than the Pipe's buffer size? It it possible that the pipe's buffer will "choke" in such case?
> > 
> > 
> > Nope because the buffer == partSize. That's the trick here. I'm testing with 160MB files.
> 
> The 160MB files however contain new line separated events - so the datastream.ReadString('\n') never has to return 160MB.
> In case of CloudTrail it has to return the full payload and keep it in memory.

Maybe you can test it to make sure it's fine? Should be testable even through unit tests.

rleighton:
> > > > Is this going to work with CT where we have to pull more data than the Pipe's buffer size? It it possible that the pipe's buffer will "choke" in such case?
> > > 
> > > 
> > > Nope because the buffer == partSize. That's the trick here. I'm testing with 160MB files.
> > 
> > 
> > The 160MB files however contain new line separated events - so the datastream.ReadString('\n') never has to return 160MB.
> > In case of CloudTrail it has to return the full payload and keep it in memory.
> 
> Maybe you can test it to make sure it's fine? Should be testable even through unit tests.

I'll test it with https://summitroute.com/blog/2020/10/09/public_dataset_of_cloudtrail_logs_from_flaws_cloud/

rleighton:
> > > Is this going to work with CT where we have to pull more data than the Pipe's buffer size? It it possible that the pipe's buffer will "choke" in such case?
> > 
> > 
> > Nope because the buffer == partSize. That's the trick here. I'm testing with 160MB files.
> 
> The 160MB files however contain new line separated events - so the datastream.ReadString('\n') never has to return 160MB.
> In case of CloudTrail it has to return the full payload and keep it in memory.

Yes, but the buffer paging is _before_ the gzip uncompress so we know it is working.

internal/log_analysis/log_processor/downloadpipe/downloadpipe.go
```diff
@@ -0,0 +1,134 @@
+package downloadpipe
```
alxarch, line 1: I think this can be handled in a [more efficient way](https://gist.github.com/alxarch/3ce77a5a16423142200bb312203c7414).

The idea behind `io.Pipe` is that writes to `io.PipeWriter` block until someone reads data.
By using `Concurrency = 1` *and* the `BufferProvider` field in `s3manager.Downloader`, we can pipe data from each chunk *directly* into the `io.PipeWriter`. This way we minimize the number of times the data gets copied between buffers. Keep in mind that the implementation of `s3manager.Downloader` falls back to `io.Copy` internally to copy data to the `io.WriterAt` and that uses a **32Kb** buffer. That is a lot of buffer duplications for a 5M part.
 
The linked solution improves throughput by pre-fetching the next chunk while the last one is being processed.

alxarch (changes requested):
I am suggesting a more efficient way to handle the downloading part.
I would also like to have the uploader parameters be adjustable per download.

rleighton:
Ok, looking to test @alxarch `s3pipe` variant now

rleighton:
No issues testing with  https://summitroute.com/blog/2020/10/09/public_dataset_of_cloudtrail_logs_from_flaws_cloud/

internal/log_analysis/log_processor/common/common.go
serbantanasa: funny, my old team also converged on 50MB files after empirically testing a broad range of sizes.

rleighton:
I don't think the approach @alxarch sketched out will work because of this: 
https://github.com/aws/aws-sdk-go/blob/v1.35.32/service/s3/s3manager/download.go#L450
Notice on error that the `cleanup` function gets called unconditionally? When there is an error we have idea. We should not advance the buffer in the case of an error. Thoughts on how to handle this?

internal/log_analysis/log_processor/downloadpipe/downloadpipe.go
rleighton: See my comment below. I have the code working with unit tests but I can't figure out how to properly handle errors. We cannot advance the buffer if  there was an error.

internal/log_analysis/log_processor/common/common.go
```diff
@@ -72,7 +77,10 @@ func Setup() {
 
 	s3UploaderSession := Session.Copy(request.WithRetryer(aws.NewConfig().WithMaxRetries(MaxRetries),
 		awsretry.NewAccessDeniedRetryer(MaxRetries)))
-	S3Uploader = s3manager.NewUploader(s3UploaderSession)
+	S3Uploader = s3manager.NewUploaderWithClient(s3.New(s3UploaderSession), func(u *s3manager.Uploader) {
```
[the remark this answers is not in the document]
rleighton: Good idea. I was not going to touch the uploader but it is impacting performance.

internal/log_analysis/log_processor/downloadpipe/downloadpipe.go
alxarch: The error is returned eventually to the download manager and would close the writer and also the reader. What we could do to ensure partial data is not processed is to wrap the buffer so it catches checks error from ReadFrom, and clears itself

alxarch:
The error while reading the body of the chunk can be intercepted. If there was any errors in ReadFrom, we reset the buffer to zero size.

rleighton:
@alxarch and I got unit tests for his `s3pipe` package to work for the "connection reset by peer" case. I'll be replacing my implementation with `s3pipe` and  testing that soon under load. 

Right now there are needed optimizations to uploading that I'm testing. Once done, I'll put it all together. I think these changes will make a big difference in efficiency at high loads.

alxarch:
I cleaned up the `s3pipe` code and tests.
I fixed a possible (unimportant) race condition when closing the `io.ReadCloser`. The race would sometimes report context cancellation errors when using the Close() method. Now only the pipe closed error is returned.
I also modified the `bufio.Reader` we use for reading the S3 file to use a buffer the size of each chunk. The previous code was using the default which is 4Kb in go, that would lead to multiple calls to `io.PipeReader.Read()`, impacting performance.

internal/log_analysis/log_processor/downloadpipe/downloadpipe.go
rleighton: Resolving, will use `s3pipe`

rleighton:
I updated the 'Conversation` with new graphs and an updated summary.

internal/log_analysis/log_processor/main/lambda.go
```diff
@@ -62,14 +62,6 @@ func process(ctx context.Context, scalingDecisionInterval time.Duration) (err er
...
 	logTypesResolver := registry.NativeLogTypesResolver()
 
-	deadline, ok := ctx.Deadline()
-	if !ok {
-		panic("Lambda context doesn't have a deadline!")
-	}
-	// We should poll events for half the Lambda's duration
-	pollingTimeout := time.Until(deadline) / 2
-	ctx, cancel := context.WithTimeout(ctx, pollingTimeout)
-	defer cancel()
```
kostaspap, line 72: Is it possible to put this one back? 
Eventually we'd like to improve on the way we manage the contexts - polling will get its own context, deleting SQS messages its own (i was planning to make that change soon), writing to s3 its own, etc
Moving this here and out of the `processor.PollEvents` was intentional - the caller of `PollEvents` should be defining for how long it should run, not `PollEvents` itself. Agreed that now it doesn't make a big difference but moving it back is reversing one step. It also makes unit testing of the PollEvents more difficult, moving it higher was meant to simplify the testing logic.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -84,9 +100,9 @@ func pollEvents(
 		// continue to read until either there are no sqs messages or we have exceeded the processing time/file limit
 		highMemoryCounter := 0
 
-		for len(accumulatedMessageReceipts) < processingMaxFilesLimit {
+		for len(accumulatedMessageReceipts) < processingMaxFilesLimit && totalBytesRead < processingMaxBytesRead {
```
kostaspap, line 103: This means that each Lambda invocation will process maximum 100MB before terminating - this doesn't seem right.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -116,8 +132,10 @@ func pollEvents(
 			}
 
 			for _, msg := range messages {
-				dataStreams, err := generateDataStreamsFunc(aws.StringValue(msg.Body))
+				// pass lambda context to set FULL deadline to process which is pushed down into downloader
+				dataStreams, err := generateDataStreamsFunc(ctx, aws.StringValue(msg.Body))
 				if err != nil {
+					log.Println("generate streams error!")
```
kostaspap, line 138: leftover?

pkg/testutils/awsmocks.go
```diff
@@ -59,9 +59,14 @@ func (m *S3UploaderMock) Upload(input *s3manager.UploadInput, f ...func(*s3manag
 
 type S3Mock struct {
 	s3iface.S3API
+	Retries int
 	mock.Mock
 }
 
+func (m *S3Mock) MaxRetries() int {
+	return m.Retries
+}
```
kostaspap, line 68: This doesn't seem to be required (also it is not part of `s3iface.S3API`.

alxarch: It is required by the downloader to retry a chunk. The client does `if r, ok := s3Client.(maxRetrier); ok {...` where `maxRetrier` is
```go
type maxRetrier interface {
  MaxRetries() int
}
```
This interface is implemented by a client returned by `s3.New(...)`

internal/log_analysis/log_processor/main/lambda.go
rleighton: The issue is we need 2 deadlines with the new `s3pipe`, the first is the poll deadline (less than the lambda deadline) and the second is the lambda deadline context which is needed for the `s3pipe` reader ... this needs to be the full lambda deadline. The way I handled it is to pass the lambda ctx to the poll function, make a new context for polling and pass the lambda ctx down to the `s3pipe`. Can you suggest an alternative?

kostaspap: Ah, I see the problem now. Hmmm yeah, I don't see an easy way to do this now...Alright, I take my comment back then. We can review it later on, when we do further refactoring of the log processor - there is still some work required to pass the context to the destination etc.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -423,20 +430,48 @@ func (bs *s3EventBufferSet) removeBuffer(buffer *s3EventBuffer) {
 	if len(logTypeToBuffer) == 0 {
 		delete(bs.set, buffer.hour)
 	}
+	// because the priority queue library has no Delete() we set priority to be negative and mark buffer deleted
+	buffer.deleted = true
+	bs.sizePriorityQueue.UpdatePriority(buffer, -1.0)                   // negative number puts it last since valid are positive
+	bs.createTimePriorityQueue.UpdatePriority(buffer, -math.MaxFloat64) // negative max float puts it last since valid are negative
```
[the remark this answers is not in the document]
rleighton: Good catch, I meant to fix that. I as moving fast to find the bottlenecks. The reason we need fast data structures is in the case of large number of unique timestamps from endpoints. I simulated this benchmark by aggregating a few days of s3 access logs. These looks ups became a hot spots. I'll fix the queue to have a Remove().

internal/log_analysis/log_processor/processor/stream.go
rleighton: Yup, will fix

internal/log_analysis/log_processor/processor/stream.go
rleighton: I'll increase this to 250MB but I don't want to go much higher. The main issue is that if we get a really big file that  takes 90% of the deadline to process we don't want to start another and then fail both.

kostaspap: For the case you describe, we could decrease the size of the `streamChan` to e.g. `sqsMaxBatchSize`, or even further.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -127,6 +145,7 @@ func pollEvents(
 					continue
 				}
 				for _, dataStream := range dataStreams {
+					totalBytesRead += dataStream.S3ObjectSize
 					streamChan <- dataStream
```
kostaspap, line 149: Actually here we might want to add a select statement : the `streamChan <- dataStream` operation can block for long term - if the ctx gets cancelled in the meantime, we want to skip pushing the dataStream to the channel.

kostaspap: Could be:
```
				for _, dataStream := range dataStreams {
					select {
					case streamChan <- dataStream:
						totalBytesRead += dataStream.S3ObjectSize
					case <- ctx.Done():
						return
					}
				}
```

internal/log_analysis/log_processor/processor/stream.go
kostaspap: For this to work as expected, we'd also need this: https://github.com/panther-labs/panther/pull/2071/files/5d38408ced8948be88acda1118866c0a4643110c#r528805359

rleighton: Thank link did not work for me :(

BTW, since the biggest size we aggregate to is 50MB, stopping at 250MB should not have a big effect on aggregation, right?

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -423,20 +429,36 @@ func (bs *s3EventBufferSet) removeBuffer(buffer *s3EventBuffer) {
...
-		}
-		return false, nil
-	})
+func (bs *s3EventBufferSet) removeLargestBuffer() (largestBuffer *s3EventBuffer) {
+	largest, err := bs.sizePriorityQueue.Pop() // this takes buffer out of priority queue
+	if err != nil {                            // if q is empty then err is returned
+		return nil
+	}
+	largestBuffer = largest.(*s3EventBuffer)
+	bs.removeBuffer(largestBuffer)
```
kostaspap, line 442: Shouldn't we remove it here from the `createTimePriorityQueue`? Once a buffer is removed from a bufferset, we don't want any reference of it, anywhere.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -423,20 +429,36 @@ func (bs *s3EventBufferSet) removeBuffer(buffer *s3EventBuffer) {
...
 	return largestBuffer
 }
 
+func (bs *s3EventBufferSet) removeTooOldBuffer(now time.Time, maxDuration time.Duration) (oldestBuffer *s3EventBuffer) {
+	oldest, err := bs.createTimePriorityQueue.Pop() // this takes buffer out of priority queue
```
kostaspap, line 447: Does it offer `Peek` functionality? Would be better instead of Popping it and then readding it

internal/log_analysis/log_processor/processor/stream.go
kostaspap: Depends, we pull in logs from many log types that can end up in many different partitions. 
A 3GB Ram lambda running for 15', I don't know, I would honestly expect it to be able able to process more than just only 250MB of gzipped data. 
Also this check doesn't protect against the case where we pull 10 big files towards the end. Better backpressure (smaller `streamChan`) seems like it could handle that case better?

internal/log_analysis/log_processor/destinations/s3.go
rleighton: No but I should add that.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: the removeBuffer() method does that.

internal/log_analysis/log_processor/processor/stream.go
rleighton: Ah, that looks right. Let me test this under high load.

rleighton: I remove the test on size and make `streamChan` blocking to serialize the read and tested for a couple hours under load. 
Works!

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Done

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -71,10 +70,19 @@ func pollEvents(
...
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		panic("lambda context doesn't have a deadline!")
+	}
+	pollingTimeout := time.Until(deadline) / 4
+	pollCtx, cancel := context.WithTimeout(ctx, pollingTimeout)
+	defer cancel()
+
+	streamChan := make(chan *common.DataStream) // must be unbuffered to apply back pressure!
+	var accumulatedMessageReceipts []*string    // accumulate message receipts for delete at the end
```
kostaspap, line 85: In line 138 I would suggest to put a select statement:

```
for _, dataStream := range dataStreams {
    select {
        case streamChan <- dataStream:
        case <- ctx.Done:
                return
    }
}
```

That why we can handle better the case where the ctx has expired but we are still waiting to push work to the streamChan.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -43,8 +42,8 @@ const (
 	// Limit this so there is time to delete from the queue at the end.
 	processingMaxFilesLimit = 5000
 
-	// The max messages per read for SQS (can't find an sqs constant to refer to).
-	sqsMaxBatchSize = 10
+	// How many objects to read per sqs read, the larger value, the the bigger impact on failures
+	sqsReadBatchSize = 1 // keeping this conservative right now, must be smaller than 10
```
kostaspap, line 46: This is going to increase the API calls we make to SQS x10 compared to what we do now. Why make such a big change? 
Already by using an unbuffered `streamChan` we have efficiently applied backpressure to the system.

kostaspap: See also my comment above about using a `select` statement - this will be required to avoid pushing work to the processor after the context has expired.

pkg/priorityq/priorityq.go
```diff
@@ -0,0 +1,140 @@
...
+		value:    v,
+		priority: priority,
+	}
+	heap.Push(p.itemHeap, newItem)
+	p.lookup[v] = newItem
+}
+
+// Pop removes the element with the highest priority from the queue and returns it.
+// In case of an empty queue, an error is returned.
+func (p *PriorityQueue) Pop() (interface{}, error) {
```
alxarch, line 64: I think that we should avoid returning an `error` here.  This is a 'fundamental' collection, the caller should choose how to handle the empty case. If we want the queue to support `nil` interface values we can use `(interface{}, bool)` for return here. IMO there is no point storing `nil` in this queue in the first place. In that case we should just return `interface{}`.

pkg/priorityq/priorityq.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	item := heap.Pop(p.itemHeap).(*item)
+	delete(p.lookup, item.value)
+	return item.value, nil
+}
+
+// Peek returns the item that is on top of the heap but does not remove.
+// In case of an empty queue, an error is returned.
+func (p *PriorityQueue) Peek() (interface{}, error) {
```
alxarch, line 76: Same as above, best return `(interface{}, bool)` or just `interface{}` since supporting `nil` inserts is rather pointless.

pkg/priorityq/priorityq.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+// PriorityQueue represents the queue
+type PriorityQueue struct {
+	itemHeap *itemHeap
+	lookup   map[interface{}]*item
+}
+
+// New initializes an empty priority queue.
+func New() PriorityQueue {
```
alxarch, line 35: Since this is a 'fundamental' container, I would suggest to adjust the code so the zero value is usable. That way we can add is as a field in a struct *without* having to initialize it there also to avoid panics. (see the linked gist for an example)

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -408,6 +412,8 @@ func (bs *s3EventBufferSet) getBuffer(event *parsers.Result) *s3EventBuffer {
 	if !ok {
 		buffer = newS3EventBuffer(logType, hour)
 		logTypeToBuffer[logType] = buffer
+		bs.sizePriorityQueue.Insert(buffer, 0.0)
+		bs.createTimePriorityQueue.Insert(buffer, float64(-buffer.createTime.Unix())) // negative so oldest is on top!
```
alxarch, line 416:
```suggestion
                // Use nanoseconds so we have a better ordering
                since := time.Duration(buffer.crateTime.UnixNano()).Seconds()
		bs.createTimePriorityQueue.Insert(buffer, -since)) // negative so oldest is on top!
```

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -423,20 +429,35 @@ func (bs *s3EventBufferSet) removeBuffer(buffer *s3EventBuffer) {
...
+	largest, err := bs.sizePriorityQueue.Pop() // this takes buffer out of priority queue
+	if err != nil {                            // if q is empty then err is returned
+		return nil
+	}
+	largestBuffer = largest.(*s3EventBuffer)
+	bs.removeBuffer(largestBuffer)
 	return largestBuffer
 }
 
+func (bs *s3EventBufferSet) removeTooOldBuffer(now time.Time, maxDuration time.Duration) (oldestBuffer *s3EventBuffer) {
```
alxarch, line 446: I think going over the list of buffers is faster than using the priority queue here.

internal/log_analysis/log_processor/destinations/s3.go
alxarch: I think using a priority queue is wrong. We will be sorting constantly, on each write. 
We are making `removeLargest` faster by making each `writeEvent` slower.

alxarch: It's best we sort once when the condition of too many buffers / tick is met

alxarch (changes requested):
I think that the PriorityQueue is making things slower rather than faster. We are adding memory overhead and making each `writeEvent` slower to have `removeLargestBuffer` be faster. `removeTooOldBuffer` would be faster without the queue also.
We need to sort the buffers by age on each tick and remove flush the old ones.
If the too-many-buffers condition triggers, we sort by size and flush half of them (the largest ones).

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Where it hurts is when the list gets big (thousands). Happens when timestamps are all over the place.

internal/log_analysis/log_processor/processor/stream.go
rleighton: That was left over , good catch.  I'll change and add the select and stress test. That should work.

pkg/priorityq/priorityq_test.go
```diff
@@ -0,0 +1,150 @@
...
+	pq.Insert("bar", 3)
+	pq.Insert("bar", 1)
+
+	if pq.Len() != 2 {
+		t.Fatal("queue should ignore inserting the same element twice")
+	}
+
+	item, _ := pq.Pop()
+	if item.(string) != "foo" {
+		t.Fatal("queue should ignore duplicate insert, not update existing item")
```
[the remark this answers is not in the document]
rleighton: I'm guessing the author wants a clear distinction between insert and update. If we wanted and `Upsert()` we could add one.

pkg/priorityq/priorityq.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (ih *itemHeap) Push(x interface{}) {
+	it := x.(*item)
+	it.index = len(*ih)
+	*ih = append(*ih, it)
+}
+
+func (ih *itemHeap) Pop() interface{} {
+	old := *ih
+	item := old[len(old)-1]
```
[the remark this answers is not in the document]
rleighton: > I think that the PriorityQueue is making things slower rather than faster. We are adding memory overhead and making each `writeEvent` slower to have `removeLargestBuffer` be faster. `removeTooOldBuffer` would be faster without the queue also.
> We need to sort the buffers by age on each tick and remove flush the old ones.
> If the too-many-buffers condition triggers, we sort by size and flush half of them (the largest ones).

This was bothering me too. I have an easy fix: truncate rank to MB, then the Update() call will look up the value, but only find it has changed once per MB and need to re-order. Since we flush at 50MB that is max 50 updates.

deployments/panther_config.yml
```diff
@@ -44,7 +44,7 @@ Infra:
   # Processing capability can be further increased by raising the
   # Lambda concurrent execution limit from default 1000, see:
   # https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html
-  LogProcessorLambdaMemorySize: 1024 # 256 - 3008, in 64MB increments
+  LogProcessorLambdaMemorySize: 2048 # 256 - 3008, in 64MB increments
```
kostaspap: This could lead to 2x increase in the cost for the log processor. 
I don't see the need to change the default here - people consuming lots of data can configure it as they see appropriate.

internal/log_analysis/log_processor/processor/stream.go
rleighton: Done

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -348,33 +349,36 @@ func (bs *s3EventBufferSet) writeEvent(event *parsers.Result) (sendBuffers []*s3
...
 	}
 	n, err := buf.addEvent(stream.Buffer())
-	bs.totalBufferedMemBytes += uint64(n)
 	if err != nil {
 		return nil, err
 	}
+	bs.totalBufferedMemBytes += uint64(n)
+
+	// update the rank so we can find largest quickly
+	bs.sizePriorityQueue.UpdatePriority(buf, float64(buf.bytes/1024*1024*1024)) // in MB to reduce cost of update
```
kostaspap, line 358: 1. This seems to be GBs
2. Just an idea, priority could be the number of parts - so it could be `bs.sizePriorityQueue.UpdatePriority(buf, float64(buf.bytes/uploaderPartSize))`

kostaspap (approved):

internal/log_analysis/log_processor/destinations/s3.go
rleighton: I like that!

rleighton: Done

internal/log_analysis/log_processor/destinations/s3.go
alxarch: But the heap already contains a slice of thousands of buffers that do get sorted on each insert/pop. That is the part that I don't understand.

pkg/priorityq/priorityq.go
```diff
@@ -0,0 +1,138 @@
...
+	heap queue
+}
+
+// Len returns the number of elements in the queue.
+func (p *PriorityQueue) Len() int {
+	return p.heap.Len()
+}
+
+// Insert inserts a new element into the queue. No action is performed on duplicate elements.
+func (p *PriorityQueue) Insert(v interface{}, priority float64) {
```
alxarch, line 38: If `Insert` does not do anything on duplicate entries, it should at least have a `bool` return value indicating if it succeeded or not, so that callers know they need to update in order to set the priority.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: Oh, you are saying in _this_ case just sort and remove top 1/2. Yes, that is would be better (just more complex). I thought you were saying to remove the pq  entirely.

pkg/priorityq/priorityq.go
rleighton: Sure, will add.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -127,7 +136,12 @@ func pollEvents(
 					continue
 				}
 				for _, dataStream := range dataStreams {
-					streamChan <- dataStream
+					// since streamChan is unbuffered it will block, exit early if deadline is past to delete messages
+					select {
```
rleighton, line 140: Breaking here when the polling deadline is not correct. We need to just commit to processing all messages in the batch read.  If we do not, and we stop,  then we force messages to wait the visibility timeout. I'm taking this out and retesting.

internal/log_analysis/log_processor/destinations/s3.go
rleighton: I'll add a FIXME comment to do this. Right now I want to get this over the finish line.

internal/log_analysis/log_processor/processor/stream.go
kostaspap: This could cause delay for some messages in the scenario you say, but seems like it is needed to avoid starting to process + writing partially data to S3 data (since we cannot finish processing them)

It's better if we can remove it, but as I understood we needed it otherwise we might not be able to process all the pulled  SQS messages in time.

Another approach would be to keep this, but reduce to number of SQS messages we poll (e.g. poll 5 instead of 10. events per batch), so if we ever hit such a scenario, there is a smaller number of affected messages.

kostaspap: Having said that, if removing this line the system works fine, even better. i was under the impression that without such a check and by pulling 10 messages from SQS, we were running into issues.

rleighton: There is an issue. We can timeout if a batch cannot finish  and then the whole batch is lost. I think the answer is simple. The `sqsReadBatchSize` needs to be a stack parameter. Default 10. If there are huge files this can be tuned down . This is really just a SQS cost issue. By setting it to 1 and there are no issues at all but more SQS api calls.

pkg/priorityq/priorityq.go
rleighton: Done

internal/log_analysis/log_processor/destinations/s3.go
rleighton: I misread your question. Let me start over. 

This called once a minute and stops when there are no more "old" buffers, so it will not usually loop over many.

I'm curious why you think  looping over many buffers would be faster? Most of the time the "oldest" is not old enough to flush so there is 1 look up. We only update the priority queue once when the buffer is created.

alxarch: `heap.Push` and `heap.Fix` (used when we insert or update the priority queue) do a full sort of the slice inside the heap.
Since we update the priority queue for size on each event write, it means we sort the slice on each log event.

We have 7 cases that where we might want to flush buffers:
1. On each write if the age of a buffer is over some limit (single)
2. On each write if the size of the particular buffer is over some limit (single)
3. On each write if the number of events of the particular buffer is over some limit (single)
4. On each write if the total size of all buffers is over some limit (batch) (sort by size)
5. On each write if the total number of buffers is over some limit (batch) (sort by size, age) 
6. On each tick (1min) where we want to flush old buffers (batch) (sort by age)
7. When the lambda deadline is near (full)

Out main problem comes from case 7. If we accumulate too many buffers we cannot flush them all at once due to constraints imposed by the S3 API.
Cases 4,5,6 (the batch ones) should not be a problem as long as we can pick the buffers to flush and do the flushing in a separate goroutine _in batch_(i.e. `flushChannel <- []*buffer`) instead of `go flush(*buffer)`

What complicates our problem further are SQS message deletions. Each buffer may contain events from multiple SQS messages and each SQS message's events may end up in multiple buffers. We need a way to be able to tell, when a buffer is flushed, which messages we can delete. What I propose for this problem is:
- add message id on each event
- keep on each buffer a `map[string]int64` to know number of events per message id in that buffer (`bufferEvents`)
- once a message is fully parsed notify an in-memory index of counters that this message is processed and resulted in `n` events.
- once a buffer is flushed, notify the same index using the `bufferEvents` map. If the message is done processing and the total number of flushed messages equals the number of produced messages, _delete the SQS message_.

kostaspap: > Since we update the priority queue for size on each event write, it means we sort the slice on each log event.

This is actually no longer accurate. The code in this PR is: 
```
bs.sizePriorityQueue.UpdatePriority(buf, float64(buf.bytes/uploaderPartSize))
```
The heap.Fix will run only when the `buf.bytes/uploaderPartSize` changes so not on every event.

Deleted the SQS messages once the buffers are written is something that we want to do, but this is out of the scope of this PR.

kostaspap: Deleting the SQS messages when needed after flushing the buffers is out of scope of this PR.

alxarch: > Such a logic would not be necessary. Flushing the buffers on a schedule is preferable as it can handle the case where we buffered some events in memory but we don't receive new events for a period of time. We would still want to flush what's inside that buffer.
I say we keep both. Since we can decide right there that the buffer needs flushing, we distribute the flushing over the tick period and get less buffers to flush on each tick. As long as we do the flushing in a separate goroutine.

alxarch: > Deleting the SQS messages when needed after flushing the buffers is out of scope of this PR

I wholeheartedly agree that scope creep is one of the major issues in this PR. What I want to point out with my suggestion is that we cannot achieve 'true' parallel processing of SQS messages unless we fix SQS message deletion in a distributed way*. 
We need the following goroutines to be working in harmony:
- `poller` polls for messages and pushes process tasks to the `parser` queue _per s3 object_.
   Also notifies `ack` goroutine of the SQS message id and s3 objects in this message.
- `parser`  (can be multiple if we solve #2001) processes those tasks pushing events to `sink` queue.
   Also notifies `ack` goroutine of events produced for an s3 object.
- `sink`  collects events and distributes them to buffers, pushing to `flush` buffers that are ready.
- `flush` (can also be multiple) accepts buffers to be flushed and notifies `ack` of flushed events per message-id/s3-object.
- `ack`  orchestrates SQS message deletions.

Right now most of these happen serially. We do this because... `CloudTrail`. I believe that solving this properly can only happen after #2001. Everything we do until then, is a 'trick' that tunes a serial process for some specific scenario (i.e. high volume) and probably raises the overall run time for another. If we focus on making this truly parallel, all scenarios will be handled gracefully.

*We're also blocking in multiple places where we shouldn't. One such place (I tried to explain this yesterday in the standup) is when we wait for the first chunk to detect if the file is gzipped. This needs to happen in the `parser` goroutine so that `poller` is not blocked waiting on reads from S3.

alxarch: > bs.sizePriorityQueue.UpdatePriority(buf, float64(buf.bytes/uploaderPartSize))

This is an arbitrary limit for gzipped files. Compression ratios almost certaintly won't match across input/output.

kostaspap: > > bs.sizePriorityQueue.UpdatePriority(buf, float64(buf.bytes/uploaderPartSize))
> 
> This is an arbitrary limit for gzipped files. Compression ratios almost certaintly won't match across input/output.

Yup, that's intended.

internal/log_analysis/log_processor/processor/stream.go
kostaspap: That's fair. I think we can go with this for now. We are probably going to do a rehaul on this logic to handle such scenarios better.

rleighton:
All good comments and worth considering but outside of scope.

The missing thing from the this discussion is that such designs need to be
validated with lots of benchmarks, my last week for example was spent
testing and measuring this change.
Nearly all my time. I tried to keep the PR as small as possible while
working at the target customer load. At high volume behavior changes, both
of the software and network / connected services. This makes code like this
very hard. Bugs are really hard to find and changes to the code can have
unintended consequences. “Obviously” good changes turn out to be bad ideas
many times. I definitely think we should revisit this code , particularly
to reduce blast radius of failures. I just want emphasize that changes in
this part of the system are sensitive and require lots of time to test.

Unless there are any final important issues, I’d like to push this. I did
about 5 hours of testing under high load yesterday and saw no issues.

alxarch:
> issues, I’d like to push this. I did about 5 hours of testing under high lo

I understand the urgency and the time invested in solving a blocking issue for a client. I will approve this since it will fix the particular issue. I will add a 'sibling' issue to #2001 to refactor the log processor code with a concurrency-focused approach as the one described in my comments. I think we should avoid further modifications to the log processor pipeline until the pipeline has clear separation of concerns for each step. My proposal is to spent a full sprint to have a feature branch that achieves this (roughly one week coding, one week reviews) and introduce it into master in the start of the sprint after, so we can have stress tests and 'burn-in' for a full release cycle.

alxarch (approved):

rleighton:
I like that idea. I would design the stress tests before you start coding,
ideally check in the code that generates the data and scenarios.
Also, there should be a baseline comparison to the previous version.

PR #2073, opened 2020-11-19
Specify Panther internal SQS queue

glerb:
The SQS queue ARN wasn't templated, requiring a full text ARN to be received as a module input. This makes the specification consistent with the log processing notifications module's style.

## Background

## Changes

## Testing


nhakmiller (approved):
Thanks for this!

PR #2075, opened 2020-11-19
minor udm update

lindsey-w:
## Background

In order to support more complex JSONPath (e.g. `$.events[*].parameters[?(@.name == 'USER_EMAIL')].value`), we must use the `parse` method from `jsonpath_ng.ext`.  Further, we decided that `udm()` should only ever return data using the unified data model, it should not check for a matching key in the original event. 

## Changes

- update import statement
- remove `event.get()` access in the `udm` method

## Testing

- `mage test:python`


internal/log_analysis/rules_engine/src/enriched_event.py
```diff
@@ -46,11 +46,8 @@ def __len__(self) -> int:
 
     def udm(self, key: str) -> Any:
         """Converts standard data model field to logtype field"""
-        # access values via normal logType fields
-        if key in self._data.keys():
-            return self.get(key)
         # access values via standardized fields
-        if key in self.data_model.paths.keys():
```
lindsey-w, line 53: I'm super confused, I thought I had removed this after your last PR comment, @nhakmiller... I looked at the old PR and now I can't even find the comment 😖

kostaspap:
> Further, we decided that udm() should only ever return data using the unified data model, it should not check for a matching key in the original event.

Interesting, any particular reason? It seems handy to be able to use just the `udm` to access fields instead of having to pick between `get`/`udm`. Is that because of potential overlap?

lindsey-w:
> Is that because of potential overlap?

Yes, exactly.  In developing a couple of initial rules / data models I ran into this issue.  Figured it would be best to make the designation that `.udm()` should only ever use the data model provided by the user (in case they want to override how what a field means). Open to other arguments, though.  I had initially thought it would be handy to just develop a rule and not use `.get()` at all.

nhakmiller (approved):

PR #2077, opened 2020-11-20
Oncall Fixes

nhakmiller:
## Background

A few more minor fixes for issues noticed when on call.

## Changes

- The `panther-source-api` was using an outdated template to generate the templates you can download from the UI. This meant EKS scanning would not work. Updated to latest template, I seem to remember us not wanting to use a latest tag for some reason. Also update the "launch stack" link.
- Updated the `aws-event-processor` event blacklist.
- Do some better nil handling to avoid panics in the snapshot poller

## Testing

- Manual testing


kostaspap:
> * I seem to remember us not wanting to use a latest tag for some reason.

Yeah, we don't want that. Someone running panther v1.9 shouldn't deploy 1.13 templates

web/src/components/wizards/ComplianceSourceWizard/StackDeploymentPanel/StackDeployment.tsx
```diff
@@ -47,7 +47,7 @@ const StackDeployment: React.FC = () => {
   const { stackName, body } = data?.getComplianceIntegrationTemplate ?? {};
   const cfnConsoleLink =
     `https://${pantherConfig.AWS_REGION}.console.aws.amazon.com/cloudformation/home?region=${pantherConfig.AWS_REGION}#/stacks/create/review` +
-    `?templateURL=https://s3-us-west-2.amazonaws.com/panther-public-cloudformation-templates/panther-cloudsec-iam/v1.0.0/template.yml` +
+    `?templateURL=https://s3-us-west-2.amazonaws.com/panther-public-cloudformation-templates/panther-cloudsec-iam/v1.12.1/template.yml` +
```
kostaspap, line 50: Could we make that dynamic?

internal/core/source_api/api/get_integration_template.go
```diff
@@ -33,8 +33,8 @@ import (
 
 const (
 	TemplateBucket           = "panther-public-cloudformation-templates"
-	CloudSecurityTemplateKey = "panther-cloudsec-iam/v1.0.1/template.yml"
-	LogAnalysisTemplateKey   = "panther-log-analysis-iam/v1.0.0/template.yml"
+	CloudSecurityTemplateKey = "panther-cloudsec-iam/v1.12.1/template.yml"
```
kostaspap, line 36: Could we make that dynamic by using somme environment value for example? These templates follow the Panther release versions, right?

kostaspap (approved):

web/src/components/wizards/ComplianceSourceWizard/StackDeploymentPanel/StackDeployment.tsx
nhakmiller: With a little digging it appears that `STABLE_PANTHER_VERSION` could be used here. I will put it in here and in the other template link, but I would appreciate if someone from the @panther-labs/web team could validate that it has been done correctly.

internal/core/source_api/api/get_integration_template.go
nhakmiller: I'll see if the backend is tracking this value somehow already,  I'm not sure that it is.

nhakmiller: I spoke too soon, looks like `CustomResourceVersion` should fit the bill.

nhakmiller:
Doing a deployment in `us-east-1` then to double check then I'll merge.

nhakmiller:
> Doing a deployment in `us-east-1` then to double check then I'll merge.

Deployed into dev account, and both the launch now and the download template links worked as expected.

PR #2078, opened 2020-11-20
Don't fail the full batch in case of partial failures

kostaspap:
## Background

Currently the log processor pulls a lot of events through SQS, processes them, and if everything has succeeded, the messages are deleted from the SQS queue. However if any of the following errors occur, the FULL BATCH will fail and all of the messages will be reprocessed. 

The failures can occur: 
1. Polling new messages from SQS queue
2. Accessing the S3 objects & reading first chunk of the S3 objects
3. Reading the full S3 objects
4. Writing to the output

With this PR, I make sure that failures No1, and No2 don't cause the batch to fail. No3 and No4 could be addressed in separate PR and are indeed more complicated. They are also less common - especially No2 can happen very frequently in case of permissions misconfiguration.

## Changes

- See above
- Updated flaky test

## Testing

- Unit tests

### DIFF
--- a/internal/log_analysis/log_processor/processor/stream.go
+++ b/internal/log_analysis/log_processor/processor/stream.go
@@ -116,7 +116,7 @@ func pollEvents(
 			for _, msg := range messages {
 				dataStreams, err := generateDataStreamsFunc(aws.StringValue(msg.Body))
 				if err != nil {
-					zap.L().Warn("Unable to process event", zap.Error(err))
+					zap.L().Warn("Skipping event due to error", zap.Error(err))
 					continue
 				}
 				for _, dataStream := range dataStreams {
--- a/internal/log_analysis/log_processor/processor/stream.go
+++ b/internal/log_analysis/log_processor/processor/stream.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/service/sqs"
 	"github.com/aws/aws-sdk-go/service/sqs/sqsiface"
 	"github.com/pkg/errors"
@@ -35,6 +36,7 @@ import (
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/logtypes"
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/sources"
 	"github.com/panther-labs/panther/pkg/awsbatch/sqsbatch"
+	"github.com/panther-labs/panther/pkg/awsutils"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


kostaspap:
I think the change is pretty self-contained (and unit testable) so it can make it to the release

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -106,19 +107,19 @@ func pollEvents(
 			// keep reading from SQS to maximize output aggregation
 			messages, err := receiveFromSqs(ctx, sqsClient)
 			if err != nil {
-				readEventErrorChan <- err
+				zap.L().Warn("Encountered issue while polling sqs messages. Stopping polling", zap.Error(err))
```
rleighton, line 110: Should this be `Error()` so we get paged? I think a comment is needed to justify `Warn()` . We will not get paged on configuration issues.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -106,19 +107,19 @@ func pollEvents(
...
-				break
+				return
 			}
 
 			for _, msg := range messages {
 				dataStreams, err := generateDataStreamsFunc(aws.StringValue(msg.Body))
 				if err != nil {
-					readEventErrorChan <- err
-					return
+					zap.L().Warn("Skipping event due to error", zap.Error(err))
```
rleighton, line 121: Should this be `Error()` so we get paged? I think a comment is needed to justify `Warn()` . We will not get paged on configuration issues.

rleighton (approved):
Very nice change! Simple and looks correct. I just wonder if we should get paged on the errors.

internal/log_analysis/log_processor/processor/stream.go
kostaspap: This can happen if we can't access/read the data. The only case where this can happen and persist is 1. Some AWS issue 2. Misconfiguration issue (e.g. can't assume the role for whatever reason). Until now it has been usually the 2nd, the first doesn't happen so often because we retry a few times.

Note that here the message will not be deleted but it will return to the queue. If the problem persists, the message will end up in the DLQ - there we will get paged.

kostaspap: I'll add a comment

internal/log_analysis/log_processor/processor/stream.go
kostaspap: Actually this isn't a configuration issue. I will switch to Error, we can switch it back to warn if needed in the future.

alxarch (approved):
This PR seems to be doing what it indents to do :)

kostaspap:
> This PR seems to be doing what it indents to do :)
> I will note here that because SQS batches messages and each message can contain multiple S3 events,
> It might be useful to first produce a slice of all S3Event messages and then handle each batch of S3EventRecord in a separate function. This would allow us to use `defer` to delete each sucessfull S3Event message, either individually or by batch delete at the end of the the events.SQSEvent processing. This falls outside the scope of the PR though as it would shuffle around a lot of code in the log processor.

The issue here is that a message should not be deleted until the Destination has written the contents of the S3 objects referenced in the SQS message object... That's a bit more challenging though. Somehow we should pass the SQS message handler references to the Destination, which should notify the "deletion component" when all the contents of an SQS message have been written to S3. It's totally doable, just work for another PR as you point out :)

PR #2087, opened 2020-11-20
Sort by displayName by default instead of ID

AlexPanther:
**Description in progress**

## Background
closes: https://github.com/panther-labs/panther-enterprise/issues/434

## Changes

- Add backend sort by `displayName` functionality
- Add frontend option to sort by `displayName` in dropdown of rules page
- Add tests

## Testing

- Load rules page to ensure name sort is default
- Select previously implemented sort, then select name sort
- Select name sort when list is already sorted by name


austinbyers (changes requested):
You accidentally committed all of the old `api/gateway` files! After I worked so hard to remove them, now you want to add them back 😭 

Just `git rm -r api/gateway` and that should fix it.

Then please update the PR body with a description and steps for how you tested. You can add "closes: https://github.com/panther-labs/panther-enterprise/issues/434" to automatically close the associated issue when the PR merges

internal/core/analysis_api/handlers/list.go
```diff
@@ -133,10 +138,38 @@ func sortItems(items []tableItem, sortBy, sortDir string, compliance map[string]
 	case "severity":
 		sortBySeverity(items, ascending)
 	default:
-		sortByID(items, ascending)
+		panic("Unexpected sortBy: " + sortBy)
```
austinbyers, line 141: Let's add a comment here, something like:

```go
// Input validation for the caller already happens in the struct validate tags.
// If we reach this code, it means there is a sortBy allowed in the input validation,
// but not supported in the backend, which should never happen
```

internal/core/analysis_api/handlers/list.go
```diff
@@ -133,10 +138,38 @@ func sortItems(items []tableItem, sortBy, sortDir string, compliance map[string]
...
+		left, right := items[i], items[j]
+
+		var leftName, rightName string
+		leftName, rightName = left.DisplayName, right.DisplayName
+		if leftName == "" {
+			leftName = left.ID
+		}
+		if rightName == "" {
+			rightName = right.ID
+		}
```
austinbyers, line 156: Let's add a comment about why we're doing this, since it isn't obvious by looking at it.

E.g.

```go
// The frontend shows display name *or* ID (when there is no display name)
// So we sort the same way it is shown to the user - displayName if available, otherwise ID
```

internal/core/analysis_api/handlers/list_rules.go
```diff
@@ -77,6 +77,9 @@ func stdRuleListInput(input *models.ListRulesInput) {
 	if input.SortDir == "" {
 		input.SortDir = defaultSortDir
 	}
+	if input.SortBy == "" {
```
austinbyers, line 80: Nitpick: let's keep the fields in alphabetical order, so this would go before `input.SortDir`

internal/core/analysis_api/handlers/list_test.go
```diff
@@ -63,3 +63,71 @@ func TestPagePoliciesPageOutOfBounds(t *testing.T) {
...
+		{ID: "h", DisplayName: "b"},
+		{ID: "c", DisplayName: "y"},
+		{ID: "e", DisplayName: "a"},
+		{ID: "g", DisplayName: "b"},
+		{ID: "b", DisplayName: ""},
+	}
+
+	sortByDisplayName(items, true)
+
+	paging, truncation := pageItems(items, 1, 1)
```
austinbyers, line 79: Here you're actually testing two different functions. The `pageItems` function already has unit tests; let's keep the unit tests as isolated as possible and test just the `sortByDisplayName` function.

So you would sort the items, and then just verify their sorted order:

```go
	items := []tableItem{
		{ID: "a", DisplayName: "z"},
		{ID: "h", DisplayName: "b"},
		{ID: "c", DisplayName: "y"},
		{ID: "e", DisplayName: "a"},
		{ID: "g", DisplayName: "b"},
		{ID: "b", DisplayName: ""},
	}

	sortByDisplayName(items, true)

	expected := []tableItem{
		{ID: "...", DisplayName: "..."},
	}
	assert.Equal(t, expected, items)
```

austinbyers (commented):
Implementation looks pretty good though! Just some comments about documentation and testing

internal/core/analysis_api/handlers/list.go
AlexPanther: - [ ] Add comment

internal/core/analysis_api/handlers/list.go
AlexPanther: - [ ] Add comment

internal/core/analysis_api/handlers/list_rules.go
AlexPanther: - [ ] Reorder

internal/core/analysis_api/handlers/list_test.go
AlexPanther: - [ ] Only test `sortByDisplayName`

AlexPanther:
> You accidentally committed all of the old `api/gateway` files! After I worked so hard to remove them, now you want to add them back 😭
> 
> Just `git rm -r api/gateway` and that should fix it.
> 
> Then please update the PR body with a description and steps for how you tested. You can add "closes: [panther-labs/panther-enterprise#434](https://github.com/panther-labs/panther-enterprise/issues/434)" to automatically close the associated issue when the PR merges

Well said 😆

austinbyers (approved):
Congrats on your first change!! 🎉

PR #2088, opened 2020-11-22
Fix tests

### LINKED ISSUE #1600: Flaky unit test in log processor
### Describe the bug

[CircleCI is currently failing on the master branch](https://app.circleci.com/pipelines/github/panther-labs/panther/7481/workflows/1bc211d6-d4f0-44bc-a173-e439d0ed62af/jobs/20423):

```
--- FAIL: TestSendDataIfTimeLimitHasBeenReached (1.58s)
    s3_test.go:363: FAIL:	Upload(string,string)
        		at: [s3_test.go:343]
    s3_test.go:363: FAIL: 0 out of 1 expectation(s) were met.
        	The code you are testing needs to make 1 more call(s).
        	at: [s3_test.go:363]
    s3_test.go:364: FAIL:	Publish(string)
        		at: [s3_test.go:344]
    s3_test.go:364: FAIL: 0 out of 1 expectation(s) were met.
        	The code you are testing needs to make 1 more call(s).
        	at: [s3_test.go:364]
FAIL
coverage: 79.5% of statements
FAIL	github.com/panther-labs/panther/internal/log_analysis/log_processor/destinations	2.051s
```

The same error also happened [5 days ago](https://app.circleci.com/pipelines/github/panther-labs/panther/7336/workflows/bbe38256-d5e8-4fa0-bf89-0e4bd05223e7/jobs/19637) and [15 days ago](https://app.circleci.com/pipelines/github/panther-labs/panther/7005/workflows/faef8d88-c91b-472a-929d-9bfb35477174/jobs/17873), but not before that as best I can tell.

### Expected behavior

Unit tests should always pass in any environment.

kostaspap:
## Background

Fixes #1600 

## Changes

- Fixes flaky test 'TestSendDataIfTimeLimitHasBeenReached'
- Fixes bug when calculating number of buffers in the bufferset. Updated unit tests to verify correct behavior. 

## Testing

- mage test:ci


internal/log_analysis/log_processor/destinations/s3_test.go
```diff
@@ -280,144 +254,109 @@ func TestSendDataIfTotalMemSizeLimitHasBeenReached(t *testing.T) {
...
-			eventChannel <- testResult
-			time.Sleep(destination.maxDuration + (time.Millisecond * 10)) // give time to for timers to expire
+		defer close(eventChannel)
+		for i := 0; i < nevents; i++ {
+			eventChannel <- newSimpleTestEvent().Result()
+			// The destination should flush events every 'maxDuration' time.
+			// While sleeping here, the destination should write to S3
+			// the event we just wrote to the eventChannel
+			time.Sleep(2 * destination.maxDuration)
 		}
```
kostaspap, line 294: Previously we were giving the goroutine that flushes the buffers only 10ms to complete its operation. Under high CPU load (experienced while running the test suite) sometimes the goroutine wouldn't complete in time, thus the issues reported in #1600 . I'm now giving it more time (100ms) while not slowing down the test execution since i lowered the maxDuration to 100ms from 250ms

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -353,18 +354,21 @@ func (bs *s3EventBufferSet) writeEvent(event *parsers.Result) (sendBuffers []*s3
 		return nil, err
 	}
 	// Check if bufferSet has too many entries
-	if len(bs.set) > bs.maxBuffers {
-		// The hope is most of the flushed buffers were done updating (as events often come roughly in time order)
-		bufferReduction := bs.maxBuffers / 2
+	if bs.numBuffers > bs.maxBuffers {
```
kostaspap, line 357: `len(bs.set)` was not the correct way to calculate number of buffers in memory, since it is a `map[time.Time]map[string]*s3EventBuffer`

internal/log_analysis/log_processor/destinations/s3_test.go
```diff
@@ -428,79 +370,108 @@ func TestSendDataToS3FromMultipleHoursBeforeTerminating(t *testing.T) {
 	assert.Equal(t, aws.String("testbucket"), uploadInput.Bucket)
 	assert.True(t, strings.HasPrefix(*uploadInput.Key, expectedS3Prefix) ||
 		strings.HasPrefix(*uploadInput.Key, expectedS3Prefix2)) // order of results is async
+
+	destination.mockS3Uploader.AssertExpectations(t)
+	destination.mockSns.AssertExpectations(t)
 }
 
 func TestSendDataWhenExceedMaxBuffers(t *testing.T) {
```
kostaspap, line 378: This test would fail with the old code.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -353,18 +354,21 @@ func (bs *s3EventBufferSet) writeEvent(event *parsers.Result) (sendBuffers []*s3
...
+		buffersToRemove := (bs.maxBuffers + 1) / 2
```
kostaspap, line 358: In case the maxBuffers was `1`, the previous code wouldn't work.

internal/log_analysis/log_processor/destinations/s3_test.go
```diff
@@ -280,144 +256,110 @@ func TestSendDataIfTotalMemSizeLimitHasBeenReached(t *testing.T) {
...
-		for i := 0; i < nevents-1; i++ {
-			eventChannel <- testResult
-			time.Sleep(destination.maxDuration + (time.Millisecond * 10)) // give time to for timers to expire
+		defer close(eventChannel)
+		for i := 0; i < nevents; i++ {
+			eventChannel <- newSimpleTestEvent().Result()
+			// The destination writes events to S3 every 'maxDuration' time.
+			// While we sleep here, the destination should write to S3
+			// the event we just wrote to the eventChannel
+			time.Sleep(2 * destination.maxDuration)
```
kostaspap, line 296: Previously we were giving the goroutine that flushes the buffers only 10ms to complete its operation. Under high CPU load (experienced while running the test suite) sometimes the goroutine wouldn't complete in time, thus the issues reported in #1600 . I'm now giving it more time (100ms) while not slowing down the test execution since i lowered the maxDuration to 100ms from 250ms

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -423,6 +428,7 @@ func (bs *s3EventBufferSet) removeBuffer(buffer *s3EventBuffer) {
 	if len(logTypeToBuffer) == 0 {
 		delete(bs.set, buffer.hour)
 	}
+	bs.numBuffers--
```
alxarch, line 431: We are not certain at this point that we actually deleted a buffer. We need another check to see if `buffer.logType` exists in `logTypeToBuffer`

alxarch (changes requested):
The changes seem to be fixing the error we were encountering but we need to make sure that the buffer was actually deleted when decrementing the `numBuffers` counter. 
Overall I want to point out that the code handling this is  very hard to follow and feels like more race conditions are lurking in the shadows. We should rewrite this  ASAP to use context.Context and have clear channel and goroutine flow.

internal/log_analysis/log_processor/destinations/s3.go
kostaspap: I added a check. The way the code is right now this cannot happen (we will always remove an existing buffer from a bufferset) but added for safety. 

I am thinking to eventually move the buffers/buffersets to their own file in the package and have more unit tests to verify their behavior.

kostaspap:
> Overall I want to point out that the code handling this is very hard to follow and feels like more race conditions are lurking in the shadows. We should rewrite this ASAP to use context.Context and have clear channel and goroutine flow.

Yes, this code and overall log processor code could take some cleanup. Using context.Context and improving code readability is one thing, managing cancellation of all the major goroutines (poller, parsing, destination) is another, better memory management etc 
I don't think its super urgent, but definitely good to do when we get more confident about how the processing flow should look like.

giorgosp (approved):

alxarch (approved):
LGTM

austinbyers (approved):
Thanks so much for the fix!

PR #2089, opened 2020-11-22
Kill Swagger

austinbyers:
## Background

The PR I've been looking forward to for many months now :) With all of the internal API gateways removed (#1686 ), now we can get rid of `swagger`, `pkg/gatewayapi`, and all the related ugly hacks:

* Injecting the swagger spec directly into the CloudFormation templates before deploying
* Rewriting swagger import headers because of bugs in the `go-swagger` library

## Changes

- Remove `swagger` from all `mage` targets
- Remove as much of `pkg/gatewayapi` as possible
    - `ReplaceMapSliceNils` moved to `genericapi` because we will keep it
    - The rest is deprecated and can be removed once we consolidate our internal API layer
- Fix CloudFormation references to point to the original `bootstrap_gateway.yml` template instead of `out/embedded.bootstrap_gateway.yml`
- Backport an enterprise fix - the bulk upload handler was too long, so the item creation moved into a separate function

## Testing

- `mage clean setup gen fmt test:ci teardown deploy`


kostaspap (approved):

kostaspap:
"Kill it with fire!"

PR #2096, opened 2020-11-23
Update Go/Python libs, prevent dirty repo states

austinbyers:
## Background

Update Go and Python library versions (should resolve some Snyk alerts) and minor mage changes to address developer feedback. In particular, common mage commands (`setup`, `master:deploy`, and `master:publish`) will no longer result in dirty repo states.

## Changes

- Update `pip` versions, Go modules, and CI tool versions
- Remove unused `genericapi.ContainsHTML` function (we use struct tags instead for the same purpose)
- `mage deploy` and `mage master:*` will now automatically re-build the Python layer if the pip libraries in the config file change
    - When @lindsey-w added a new library, many devs were confused because their deployments started breaking; they each had to manually `mage clean` or `rm -r out`. Now, the `buildLayer()` mage function checks not only that the zipfile exists, but that the checksum of the pip versions used to build the layer is the same. If the versions change, a new layer is created automatically during the deploy process
- `mage setup` now automatically runs `go mod tidy`. Otherwise, @s0l0ist noticed that in the enterprise repo, this results in a dirty `go.mod` file because of the `xgo` tool
- `mage master:*` now runs formatting after file generation, again preventing dirty repo states due to autogenerated files that have not been formatted
- Add `eu-west-1` and `eu-central-1` to the available regions for publication (I'll manually add this change to the 1.13 release as well)

## Testing

- Deployed the gateway stack several times to test the layer creation, then no change, then added a new library and the layer was rebuilt


internal/log_analysis/awsglue/table_metadata_test.go
```diff
@@ -144,7 +144,7 @@ func TestGlueTableMetadataSignature(t *testing.T) {
 	gm := NewGlueTableMetadata(models.LogData, "My.Logs.Type", "description", GlueTableHourly, partitionTestEvent{})
 	sig, err := gm.Signature()
 	require.NoError(t, err)
-	assert.Equal(t, "53372e1ee5b73d1e73594335e6df94489d0a759106fa2119fca66844f7ee5618", sig)
+	assert.Equal(t, "2eddbc3d45b070ba2111a6e6ca9f8c8467922ee531c1926c69844216fb19f11b", sig)
```
austinbyers, line 147: @rleighton this value changes every time we update Go module versions, I'm not sure why. I assume we use AWS models somewhere in the signature, and those models get new fields when the libraries update?

rleighton: Yeah, that's likely it. That said, I think we will be removing the need for the signature since the custom resources are triggered on version change.

austinbyers: I changed the test to just verify the *length* of the signature rather than the exact value. This way we don't have to keep updating it, but we still verify the signature is in the format we expect

kostaspap (approved):

lindsey-w (approved):

PR #2097, opened 2020-11-24
Scanning rate limit error fixes

nhakmiller:
## Background

We've received feedback from some users that particularly large environments (especially those with high churn) are still experiencing rate limiting issues.

In one case, an environment with ~100 load balancers making a combined ~20,000 changes / day that require scanning observed over 175,000 failed scans by Panther due to rate limiting. These failed scans also blocked other work from being done, as that other work was blocked by the same rate limits.

This PR tackles a few low hanging fruit in the `aws-event-processor` and `panther-snapshot-poller` to address some service specific rate limit issues, and adds special rate limiting handling for individual resource scans of _any_ resource type.

## Changes

- DynamoDB Table scanning
  - Add a 30 second delay to scan requests for newly created `AWS.DynamoDB.Table` resources, as scans of these resources in the first ~15 seconds or so always fail because the table is in a provisioning state
  - Set dynamodb table batch scanning size to 20, to reduce throttles on the `DescribeTimeToLive` API call
- RDS Instance scanning
  - Don't initiate a scan when snapshots that belong to a no longer existent instance get modified/deleted
  - Add some special logic to handle changes to snapshot attributes. Previously we re-scanned _all_ RDS instances in a given region if the attributes of a single snapshot changed. Now we do some custom handling by passing through the snapshot ID from the aws-event-processor, and looking up the corresponding instance in the snapshot-poller
  - Set max batch size to 20
- And the big one, rate limit error retry handling. For individual resource scans, where we observe rate limiting in high churn environments, I've caught `ThrottlingException: Rate exceeded` AWS errors from any individual resource scanner. Now, instead of returning an error and allowing the scan to be retried normally, we manually re-queue the resource for a scan in a set period of time into the future (with a little randomness to offset multiple throttles at the same time) and then we cache the resource ID. Subsequent scan requests for that resource that come in before our upcoming scan get dropped.
  - With multiple lambda being invoked at once, this isn't a guarantee that we will not process the same resource multiple times. But each lambda container will have it's own cache, so really high churn will still be reduced
  - This only helps with large numbers of changes to relatively small numbers of resources. This is the pattern we observed in user organizations that were experiencing rate limit issues. Small numbers of changes to large numbers of resources has not been reported as an issue yet, but it is something we may want to look into pro-actively

## Testing

- Unit testing and kicking off a few specific & account wide scans to watch for errors
- Some benchmarking of the old & new code on handling high churn (this was serially enqueueing scans for the same resource, an ELBV2 load balancer. This resource type has been observed to be especially susceptible to rate limit issues):
  - old code
    - 100 requests
      - 79 rate limit errors
      - 100 succesfull scans
    - 1001 requests serially
      - 1269 failures
      - 1000 succesfull scans
  - new code
    - 100 requests
      - 9 rate limit errors
      - 27 succesfull scans
      - 75 scans skipped
    - 1000 requests
      - 6 rate limit errors
      - 37 succesfull scans
      - 960 scans skipped
			
This test appears to be highly effective, skipping 96% of scans in some cases! This code won't be nearly as effective in practice, as this test is an extreme situation (1000 scan requests in under a minute for the exact same resource). For reference, one customer who was kind enough to share some data with us showed that at peak around 3000 scans were happening in an hour for about 90 resources. Assuming perfectly optimal distribution, that's one scan per resource per ~33 seconds, which means this method (again assuming perfect distribution of scan requests) would turn ~10 scan requests into one scan request. In practice, there would be bigger bunches of scans where this is more effective due to catching more repeated requests, but also less effective because the lambda concurrency is higher.

Anyways, all these numbers are just estimations. But it seems promising that at least based on some simple tests and some simple math that this should drastically reduce scan requests (and so rate limits) in high churn environments.


internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -49,6 +51,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/aws/aws-sdk-go/service/waf"
 	"github.com/aws/aws-sdk-go/service/wafregional"
+	lru "github.com/hashicorp/golang-lru"
```
nhakmiller, line 54: Using this library because we use it in the log processor already.

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -285,43 +289,66 @@ func singleResourceScan(
...
 		// This error is not retryable
 		return nil, nil
 	}
 
+	if err != nil {
+		var awsErr awserr.Error
+		// Check for rate limit errors. We don't want to blindly retry rate limit errors as this will
+		// cause more rate limit errors, so we re-schedule one new scan several minutes in the future
+		// and suppress all other scans for this resource until that time.
+		if ok := errors.As(err, &awsErr); ok && awsErr.Code() == "ThrottlingException" && awsErr.Message() == "Rate exceeded" {
```
nhakmiller, line 329: I couldn't find these things defined in an AWS library somewhere, although I would prefer to use those if the exist.

austinbyers: Not every code will have a constant, unfortunately

austinbyers (commented):
The GA release is in about 24 hours; this change seems too dangerous to be published without the full release testing because it affects the most important part of the cloud security pipeline.

I would recommend we do *not* include this in 1.13.0, but instead put it into 1.13.1 if we have to for a customer experiencing issues

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -121,6 +124,9 @@ var (
 		awsmodels.PasswordPolicySchema: {}, // Global service
 		awsmodels.WafWebAclSchema:      {}, // Global service
 	}
+
+	clientCache    = make(map[clientKey]cachedClient)
+	rateLimitCache *lru.ARCCache
```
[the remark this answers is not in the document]
nhakmiller: We would need to switch to a FIFO queue in order to use deduplication IDs I believe, I don't know what cost/other considerations there are for FIFO queues vs regular queues. At a minimum, it appears that FIFO queues don't support delayed messages, which is a feature we use a fairly regularly. I'll admit I'm not an expert on FIFO queues though, but in this case I think the architecture change is a bit heavy handed to solve this issue.

nhakmiller:
> The GA release is in about 24 hours; this change seems too dangerous to be published without the full release testing because it affects the most important part of the cloud security pipeline.
> 
> I would recommend we do _not_ include this in 1.13.0, but instead put it into 1.13.1 if we have to for a customer experiencing issues

I see. I guess logically I knew that the sprint would be short on work, but I didn't realize that of course that also means the release is happening sooner. I would have put this PR in earlier but doing all the testing was taking quite a bit of time just because of the nature of it.

This issue is causing at least one customer to be unable to use Panther. I think it would be best to get it in before v1.14. If that means v1.13.1 or v.1.13.0 I'm fine either way.

internal/compliance/aws_event_processor/processor/rds.go
```diff
@@ -78,7 +78,12 @@ func classifyRDS(detail gjson.Result, metadata *CloudTrailMetadata) []*resourceC
 		return nil
 	case "CopyDBSnapshot", "DeleteDBSnapshot", "ModifyDBSnapshot":
 		// Similar to the common case, but looking at the responseElements
-		rdsARN.Resource += detail.Get("responseElements.dBSnapshot.dBInstanceIdentifier").Str
+		instanceID := detail.Get("responseElements.dBSnapshot.dBInstanceIdentifier")
+		// This can happen when a snapshot for a DB that no longer exists is changed
+		if !instanceID.Exists() {
+			return nil
+		}
+		rdsARN.Resource += instanceID.Str
```
kostaspap, line 86: Should this be `rdsARN.Resource = instanceID.Str`?

internal/compliance/aws_event_processor/processor/rds.go
```diff
@@ -108,13 +113,20 @@ func classifyRDS(detail gjson.Result, metadata *CloudTrailMetadata) []*resourceC
...
+		// Since we can't tie this back to the corresponding RDS Instance with just the context of
+		// this event, we send the panther-snapshot-poller the ID of the db snapshot and let the
+		// poller make the appropriate API call to tie this snapshot back to a particular instance.
+		snapshotID := detail.Get("requestParameters.dBSnapshotIdentifier").Str
 		return []*resourceChange{{
 			AwsAccountID: metadata.accountID,
 			EventName:    metadata.eventName,
-			Region:       metadata.region,
+			ResourceID: arn.ARN{
+				Partition: "aws",
```
kostaspap, line 124: How would we deal with this in different partitions? Is the partition available through some other field of the requestParameters field?

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -135,12 +141,19 @@ type cachedClient struct {
...
 func Setup() {
 	awsSession := session.Must(session.NewSession()) // use default retries for fetching creds, avoids hangs!
 	snapshotPollerSession = awsSession.Copy(request.WithRetryer(aws.NewConfig().WithMaxRetries(maxRetries),
 		awsretry.NewConnectionErrRetryer(maxRetries)))
+
+	maxMem, err := strconv.Atoi(os.Getenv("AWS_LAMBDA_FUNCTION_MEMORY_SIZE"))
+	if err != nil {
+		panic(err)
+	}
+	rateLimitCache, err = lru.NewARC(maxMem / 4)
```
kostaspap, line 153: The memory size is in MBs, but the `lru.NewARC` defines the max number of elements in the cache. 
Could you add a comment with the logic behind this calculation?

internal/compliance/snapshot_poller/pollers/aws/dynamodb_table.go
```diff
@@ -38,7 +38,13 @@ import (
 	"github.com/panther-labs/panther/internal/compliance/snapshot_poller/pollers/utils"
 )
 
-const dynamoDBServiceNameSpace = "dynamodb"
+const (
+	dynamoDBServiceNameSpace = "dynamodb"
+	// Several DynamoDB API calls (most noticeably the DescribeTimeToLive API call) have very low
+	// rate limits, so we set the max batch size lower than for most resource types to avoid getting
+	// rate limited.
+	dynamodbBatchSize = 20
```
kostaspap, line 46: By decreasing the number of responses returned by each page, it means we need to make more API calls to retrieve the full results. Seems like this will have the opposite effect of what's intended? Shouldn't we return as many results as possible per API  (`dynamoDBSvc.ListTablesPages`)call but limit the frequency with which the `listTables` function is called?

internal/compliance/snapshot_poller/pollers/aws/poll.go
giorgosp: Yeah, you could create an `isAWSRateExceededError(err)` function if you feel like improving readability. Could be reused in other places too.

internal/compliance/snapshot_poller/pollers/aws/clients.go
kostaspap, line 129: Can you add comments describing the purpose of each of the caches? There seems to be a lot of context around them, having some comments will help the future maintainer of this work.

internal/compliance/snapshot_poller/pollers/aws/poll.go
kostaspap, line 329: nit: Just checking for `ThrottlingException` should be enough?

kostaspap (commented):
Seems that this should this be part of a patch release e.g. 1.13.1? 
1.13 is due to release tomorrow, which means if this is included in the 1.13 it will be released untested. 

@s0l0ist

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -285,43 +289,66 @@ func singleResourceScan(
...
+			// If we parallelize this function, we will need to see if this is already cached before
+			// re-queueing. For now, this is not necessary.
+			err = utils.Requeue(pollermodels.ScanMsg{
+				Entries: []*pollermodels.ScanEntry{scanRequest},
+			}, int64(rateLimitDelay.Seconds())+int64(pageRequeueDelayer.Intn(60)+5))
+			// If the requeue failed, give up and just let lambda error retrying handle it
+			if err != nil {
+				return nil, err
+			}
+			rateLimitCache.Add(*scanRequest.ResourceID, time.Now().Add(rateLimitDelay))
```
rleighton, line 339: Is it a little misleading to call this a cache? Is it more a "scan tracker" ?

rleighton:
> Seems that this should this be part of a patch release e.g. 1.13.1?
> 1.13 is due to release tomorrow, which means if this is included in the 1.13 it will be released untested.
> 
> @s0l0ist

This is an important fix. I also have the patches for performance improvements to the log processor that I am load testing. Perhaps these 2 changes are 13.1 and we burn-in over the weekend and release Monday?

nhakmiller:
> 1.13 is due to release tomorrow, which means if this is included in the 1.13 it will be released untested.

I think I was pretty explicit in the testing I did, this is not untested code.

internal/compliance/aws_event_processor/processor/rds.go
nhakmiller: No, the `instanceID` needs to be prefixed with `db:` which I set above on line 49.

internal/compliance/aws_event_processor/processor/rds.go
nhakmiller: The `aws-event-processor` is not partition aware today, I think it is outside the scope of this change to make it so.

internal/compliance/snapshot_poller/pollers/aws/clients.go
nhakmiller: Hmm I don't know why I put this here, I think I was just putting something random here. I'll just hardcode it to like 5000, that should be fine I think.

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -135,12 +141,19 @@ type cachedClient struct {
...
 }
 
-var clientCache = make(map[clientKey]cachedClient)
-
 func Setup() {
 	awsSession := session.Must(session.NewSession()) // use default retries for fetching creds, avoids hangs!
 	snapshotPollerSession = awsSession.Copy(request.WithRetryer(aws.NewConfig().WithMaxRetries(maxRetries),
 		awsretry.NewConnectionErrRetryer(maxRetries)))
+
+	maxMem, err := strconv.Atoi(os.Getenv("AWS_LAMBDA_FUNCTION_MEMORY_SIZE"))
```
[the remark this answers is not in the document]
nhakmiller: I'll just remove it, not sure why I thought it was needed.

internal/compliance/snapshot_poller/pollers/aws/dynamodb_table.go
nhakmiller: The `ListTablesPages` API call is not the one being rate limited, it's the API calls that we can only make on one resource at a time like `DescribeTimeToLive` or `ListTagsOfResource`. These API calls must be made once per dynamodb table, so by scanning smaller batches of tables at a time we make the API call less frequently.

After we hit the max batch size, we re-queue a scan a few seconds in the future with the next page token to continue this scan in a few seconds. This helps spread the scan out more.

internal/compliance/snapshot_poller/pollers/aws/poll.go
nhakmiller: I guess I use cache pretty generally to just be a collection of things. I can rename it to maybe `rateLimitTracker`.

internal/compliance/snapshot_poller/pollers/aws/dynamodb_table.go
kostaspap: Ok makes sense. I was a bit confused about how this was helping with spreading the load.

kostaspap:
> > 1.13 is due to release tomorrow, which means if this is included in the 1.13 it will be released untested.
> 
> I think I was pretty explicit in the testing I did, this is not untested code.

I probably didn't express myself properly here. It definitely has been tested (your notes were very detailed!) but as with all releases we want to have a few days of burnin in internal accounts

internal/compliance/aws_event_processor/processor/rds.go
austinbyers: Agreed, out of scope right now

austinbyers:
We can sync offline about which release it goes in, I could go either way. Great job on the testing!

You can keep the `sync-to-release` label and, if we want a 1.13.1 release, just wait to merge it until after the official 1.13 release is finished. We don't even have to wait for burn-in, we can release 1.13 and 1.13.1 almost at the same time, and just give 1.13.1 to the critical accounts that need it

You do have a failing Go unit test, by the way, in case you didn't see that

nhakmiller:
> You do have a failing Go unit test, by the way, in case you didn't see that

Ah was sitting on a bundle of PR feedback changes, which happened to include the fix for this test. Updated!

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -285,43 +290,65 @@ func singleResourceScan(
 	var resource interface{}
 	var err error
 
+	// First, check if we've been rate limited recently while attempting to scan this resource. If
+	// so, discard the scan request. There is already one in the ether waiting to be picked up.
+	if timestamp, ok := RateLimitTracker.Get(*scanRequest.ResourceID); ok {
+		if timestamp.(time.Time).After(time.Now()) {
```
kostaspap, line 296: Hmm, shouldn't we remove the item from the RateLimitTracker in an else statement?

kostaspap (approved):

internal/compliance/snapshot_poller/pollers/aws/poll.go
nhakmiller: Good call, will add this.

nhakmiller:
Did a final round of testing after the pr changes were accepted just to make sure.

PR #2098, opened 2020-11-24
Add default sortBy for policies and data models

austinbyers:
## Background

#2087 broke `ListPolicies` and `ListDataModels`. Both the backend and the web app started reporting `analysis-api` panics with the message "unexpected SortBy: "

The problem was that we changed how the list operations handle the default sortBy when none is specified: it used to be handled as the default in a switch statement, but now we expect each List handler to provide their own defaults.

This error would have been caught when running integration tests for #2087 , but I forgot to have Alex run those, so that's my bad.

## Changes

- Add default `sortBy` for policies and data models

## Testing

- Deployed, verified fix to policy list page in the web app
- CloudWatch alarms for `analysis-api` and `resource-processor` were resolved after the fix
- @AlexPanther is working on fixing the `analysis-api` integration tests, where it will be tested more thoroughly

### DIFF
--- a/internal/core/analysis_api/handlers/list.go
+++ b/internal/core/analysis_api/handlers/list.go
@@ -29,7 +29,6 @@ import (
 )
 
 const (
-	defaultSortBy   = "displayName"
 	defaultSortDir  = "ascending"
 	defaultPage     = 1
 	defaultPageSize = 25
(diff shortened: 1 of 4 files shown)


austinbyers:
I'm merging this now (without a review) because `master` is currently broken and I want to unblock the European dev team

But if you're tagged for review, please do still review it and if you have any comments I'll address them in a follow-up PR

s0l0ist:
LGTM

PR #2099, opened 2020-11-24
Add EU regions for publishing

austinbyers:
## Background

We now support publishing Panther to EU regions. I already manually published 1.12.1 to those regions, so we need to get the new regions into 1.13 as well.

This is a PR directly against the release branch; the associated change in `master` is already in #2096

Please merge on approval

## Changes

- Add `eu-west-1` and `eu-central-1` to the region publication set for `mage master:publish`

## Testing

- `mage master:publish` for 1.12.1

### DIFF
--- a/tools/mage/master/deploy.go
+++ b/tools/mage/master/deploy.go
@@ -42,7 +42,7 @@ const (
 	defaultStackName      = "panther"
 )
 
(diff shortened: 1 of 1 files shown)


kostaspap (approved):
LGTM

PR #2105, opened 2020-11-24
Policy alert storage

### LINKED ISSUE #1809: Policy-based Alerts
### Description

The alerting pipeline is separated out into two different paths:
- For rules
- For policies

However, there are stages in the pipeline we'd like to unify. A good first step is to persist alert's generated by policies into the same DDB table where the other alerts are currently stored.

### Related Services

Alert Forwarder (policy)
- Needs to also send to DDB when dispatching to destinations

### Designs

N/A

### Acceptance Criteria

- An alert is triggered via a policy
- User can optionally see the alert delivered to a destination
- User can visit the list alert's page and see a persisted alert that was triggered by a policy
- User can see Resource Type in policy failure alert
- User can see Source in policy failure alert

s0l0ist:
## Background

Backend changes to close #1809 

Gist:
This PR adds support for Policies to store alerts to DDB. We modify the alert schema to accommodate changes for the frontend.

**TODO**: The FE will receive alerts from policies from the BE, but it is not ready to handle them. There's logic in both FE and BE to filter out `policy` alerts. This is set by querying for only the alert-types the FE currently supports (`Rule`, `RuleError`). When the FE designs are ready, the only change for returning policy-alerts is a simple query parameter modification.

## Changes

- Store alerts generated from policies
- Added the following fields to an alert in ddb to provide the FE more context:
  - `PolicyID`
  - `PolicyDisplayName`
  - `PolicySourceID` (source integration)
  - `PolicyVersion`
  - `ResourceTypes`
  - `ResourceID` (this is the failing resource)

## Testing

- mage test:ci
- mage test:integration
- manually trigger alerts from all 3 types, see they're getting delivered/stored/rendered properly.

## Screenshots

Alerts of all types getting delivered
image

Alerts of all types getting stored in DDB
image


web/__tests__/__mocks__/builders.generated.ts
```diff
@@ -292,7 +299,7 @@ export const buildAlertSummary = (overrides: Partial<AlertSummary> = {}): AlertS
       'deliveryResponses' in overrides ? overrides.deliveryResponses : [buildDeliveryResponse()],
     eventsMatched: 'eventsMatched' in overrides ? overrides.eventsMatched : 670,
     ruleId: 'ruleId' in overrides ? overrides.ruleId : '6eb9c948-5a13-4955-bd91-b98801b55bed',
-    type: 'type' in overrides ? overrides.type : AlertTypesEnum.Rule,
+    type: 'type' in overrides ? overrides.type : AlertTypesEnum.RuleError,
```
s0l0ist, line 302: Currently, this causes CI:`npm run test` to fail.

internal/compliance/alert_forwarder/forwarder/forwarder.go
```diff
@@ -19,42 +19,148 @@ package forwarder
...
+		alertType = "Policy"
+	case models.RuleType:
+		alertType = "Rule"
+	case models.RuleErrorType:
+		alertType = "Rule_Error"
+	default:
+		zap.L().Error("Invalid Alert type")
+	}
+	return alertType
+}
```
kostaspap, line 85: Better change signature to
```
func getAlertTypeForLog(alert models.Alert) (string, error)
```
and can let the caller handle it as it wants (log the error/skip/return error etc)

internal/compliance/alert_forwarder/main/lambda.go
```diff
@@ -66,10 +75,31 @@ func reporterHandler(ctx context.Context, event events.DynamoDBEvent) (err error
...
+			"AnalysisType",
+		},
+	})
+	handler = &forwarder.Handler{
+		SqsClient:        sqsClient,
+		DdbClient:        ddbClient,
+		AlertingQueueURL: env.AlertingQueueURL,
+		AlertTable:       env.AlertsTable,
+		MetricsLogger:    metricsLogger,
+	}
```
kostaspap, line 104: nit: You maybe want to run this initialization logic inside `main` instead of the `init()` method. If/when we create a unit test for `internal/compliance/alert_forwarder/main/lambda.go`, the `init()` will run. We don't really want that though - we want it to run when the `main()` is invoked.  I've seen this issue appear in some unit tests, we always ended up moving the initialization code inside `main()`.

internal/core/alert_delivery/api/update_delivery.go
```diff
@@ -31,8 +30,8 @@ func updateAlerts(statuses []DispatchStatus) []*alertModels.AlertSummary {
 	// create a relational mapping for alertID to a list of delivery statuses
 	alertMap := make(map[string][]*alertModels.DeliveryResponse)
 	for _, status := range statuses {
-		// If the alert came from a policy, we need to skip
-		if (status.Alert.Type == deliveryModels.PolicyType) || (status.Alert.AlertID == nil) {
+		// If the alert doesn't have an ID, we need to skip
+		if status.Alert.AlertID == nil {
```
kostaspap, line 34: Hmmm, why do we need this check? 
Alerts from rules have IDs and it seems alerts from policies as well (initialized in `internal/compliance/alert_forwarder/forwarder/forwarder.go#GenerateAlertID `)

internal/log_analysis/alert_forwarder/forwarder/models.go
```diff
@@ -63,6 +62,14 @@ type Alert struct {
...
 	// no Python-generated title is available.
 	AlertDedupEvent
+
+	// Policy-specific fields
+	PolicyID            string   `dynamodbav:"policyId,string"`
+	PolicyDisplayName   string   `dynamodbav:"policyDisplayName,string"`
+	PolicyVersion       string   `dynamodbav:"policyVersion,string"`
+	PolicyIntegrationID string   `dynamodbav:"policyIntegrationId,string"`
+	ResourceTypes       []string `dynamodbav:"resourceTypes,stringset"`
+	ResourceID          string   `dynamodbav:"resourceId,string"` // This is the failing resource
```
kostaspap, line 72: The `log_analysis/alert_forwarder` reads the DDB Stream of `panther-log-alert-dedup`. The only Lambda that writes to that table is the Python rules engine. I don't think this change is needed at all.

internal/log_analysis/alerts_api/api/get_alert_test.go
```diff
@@ -138,6 +138,8 @@ func TestGetAlert(t *testing.T) {
 		Severity:          "INFO",
 		EventCount:        5,
 		LogTypes:          []string{"logtype"},
+		ResourceTypes:     []string{"resourceType"},
+		ResourceID:        "resourceId",
```
kostaspap, line 142: Can we make a separate a test case for the policy scenario? 
This `AlertItem` is about rules (it defines a `ruleId`) and now it has a `resourceId` and `resourceType`.
This doesn't seem correct. It can also be confusing to anyone reading the code in the future. 
This applies to all tests for `get_alert_test.go`.

kostaspap (commented):
Looks good! Some minor comments

api/graphql/schema.graphql
```diff
@@ -158,6 +158,12 @@ interface Alert {
   lastUpdatedBy: ID # gets mapped to a User in the frontend
   lastUpdatedByTime: AWSDateTime # stores the timestamp of the last person who modified the Alert
   updateTime: AWSDateTime! # stores the timestamp from an update from a dedup event
+  policyId: String!
+  policyDisplayName: String!
+  policyIntegrationId: String!
+  policyVersion: String!
+  resourceTypes: [String!]!
+  resourceId: String!
```
3nvi, line 166: we should unflatten this and put it inside a `policy` key. We should do a similar thing for rule-related data as well (i.e. put them inside a `rule` field). This should also make field naming consistent (i.e. `policyIntegrationId` -> `sourceId`)

The same applies to the 2 other changes below

web/src/pages/AlertDetails/AlertDetails.tsx
```diff
@@ -70,14 +73,24 @@ const AlertDetailsPage = () => {
...
+  // FIXME: uncomment when policy-alerts are supported
+  // const { data: policyData, loading: policyLoading } = usePolicyTeaser({
+  //   skip: !alertData || alertData.alert?.type !== AlertTypesEnum.Policy,
+  //   variables: {
+  //     input: {
+  //       id: alertData?.alert?.ruleId,
+  //     },
+  //   },
+  // });
+
```
3nvi, line 93: before we leave this here (even as a comment), we should sync with @mattthetraveler on how this will be presented.

internal/log_analysis/alert_forwarder/forwarder/models.go
s0l0ist: We have several 'alert' structs throughout the codebase and they're all fragmented. I added the fields here because I reference this struct in `compliance/alert_forwarder` to try to re-use the struct.

The other option is to create a new model/struct in `compliance/alert_forwarder` with just the fields that I need. Is that what is preferred? @kostaspap

internal/log_analysis/alerts_api/api/get_alert_test.go
s0l0ist: will fix

internal/core/alert_delivery/api/update_delivery.go
s0l0ist: I thought this was a good sanity check, but it looks like you're correct - all alerts should have `alertID`s now.

internal/compliance/alert_forwarder/forwarder/forwarder.go
s0l0ist: Originally, I didn't want this function to return errors or statuses like that because it affected only our metrics. However, this lambda will in reality, never handle the other cases, so I hardcode the metric values.

api/graphql/schema.graphql
s0l0ist: Better to leave it as `policyIntegrationId` as it isn't supported for rules yet. Flattening will happen in the AppSync layer.

internal/compliance/alert_forwarder/main/lambda.go
s0l0ist: I've copied this format from the `log_analysis/alert_forwarder` so there's some debt there...

api/graphql/schema.graphql
```diff
@@ -189,10 +202,16 @@ type AlertSummary implements Alert {
...
   title: String!
-  logTypes: [String!]!
+  logTypes: [String!]! # From alerts triggered by rules
   lastUpdatedBy: ID # gets mapped to a User in the frontend
   lastUpdatedByTime: AWSDateTime # stores the timestamp of the last person who modified the Alert
   updateTime: AWSDateTime! # stores the timestamp from an update from a dedup event
+  policyId: String!
+  policyDisplayName: String!
+  policyIntegrationId: String!
+  policyVersion: String!
```
austinbyers, line 212: The ! means the field is required, right? But if you're using the same alert struct for rules and policies, then these fields won't be defined for rules.

I would have probably gone with a separate `PolicyAlert` object, just like we have separate `ListPolicies` and `ListRules` even if they're handled by the same API / database

api/lambda/alerts/models/api.go
```diff
@@ -184,22 +184,28 @@ type ListAlertsOutput struct {
...
+	EventsMatched       *int                `json:"eventsMatched" validate:"required"`
+	Severity            *string             `json:"severity" validate:"required"`
+	Status              string              `json:"status,omitempty"`
+	Title               *string             `json:"title" validate:"required"`
+	LastUpdatedBy       string              `json:"lastUpdatedBy,omitempty"`
+	LastUpdatedByTime   time.Time           `json:"lastUpdatedByTime,omitempty"`
+	PolicyID            string              `json:"policyId" validate:"required"`
+	PolicyDisplayName   string              `json:"policyDisplayName" validate:"required"`
+	PolicyIntegrationID string              `json:"policyIntegrationId" validate:"required"`
+	PolicyVersion       string              `json:"policyVersion" validate:"required"`
```
austinbyers, line 206: Again, how are all of these required if rules are using the same fields?

I don't think we're using the `validate` tags at all here, we could just remove them if they aren't needed

internal/compliance/alert_forwarder/main/config.go
```diff
@@ -0,0 +1,58 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * Panther Enterprise is licensed under the terms of a commercial license available from
```
austinbyers, line 24: Remove enterprise license header

internal/log_analysis/alert_forwarder/forwarder/models.go
austinbyers: I like the idea of a new struct, but I don't have much context here

s0l0ist: okay, I'll wait for @kostaspap's feedback

api/graphql/schema.graphql
s0l0ist: Yep, I'll make it optional. It doesn't matter because we don't use `*string` in the go code so all strings will be empty `""` which this satisfies the requirement as it isn't null/undefined.

deployments/cloud_security.yml
```diff
@@ -315,19 +316,39 @@ Resources:
...
+        - Id: ReadTableStreams
+          Version: 2012-10-17
+          Statement:
+            - Effect: Allow
+              Action:
+                - dynamodb:DescribeStream
+                - dynamodb:GetRecords
+                - dynamodb:GetShardIterator
+                - dynamodb:ListStreams
+              Resource: !Sub ${AlertForwarderTable.Arn}/stream/*
```
kostaspap, line 351: Huh, is this needed? I believe since the AlertForwarderTable is listed as an event for this Lambda, the Lambda already will get the appropriate policies.

internal/compliance/alert_forwarder/main/lambda.go
```diff
@@ -66,9 +94,8 @@ func reporterHandler(ctx context.Context, event events.DynamoDBEvent) (err error
 			continue
 		}
 
-		if err = forwarder.Handle(&alert); err != nil {
-			err = errors.Wrap(err, "encountered issue while processing event")
-			return err
+		if err = handler.Do(alert); err != nil {
+			return errors.Wrap(err, "encountered issue while handling policy event")
```
kostaspap, line 98: The oplog manager will log the value of `err`. So now the logged error will not contain the `encountered issue while...` part. This is the reason why we were first setting the `err = ...` then returning it. 
Revert?

internal/compliance/alert_processor/processor/processor.go
```diff
@@ -211,16 +211,19 @@ func getAlertConfigPolicy(event *models.ComplianceNotification) (*alertmodel.Ale
...
-			AnalysisID:          event.PolicyID,
-			AnalysisName:        &policy.DisplayName,
-			CreatedAt:           event.Timestamp,
-			OutputIds:           event.OutputIds,
-			Runbook:             &policy.Runbook,
-			Severity:            string(policy.Severity),
-			Tags:                policy.Tags,
-			Type:                alertmodel.PolicyType,
-			Version:             &event.PolicyVersionID,
+			AnalysisDescription:   &policy.Description,
```
kostaspap, line 214: Seems like you could be setting the alertID here as well? Seems to be better contained. 
So you could move the `internal/compliance/alert_forwarder/forwarder/forwarder.go#GenerateAlertID` here

internal/compliance/resource_processor/processor/handler.go
```diff
@@ -232,12 +232,12 @@ func (r *batchResults) analyze(resources resourceMap, policies policyMap) error
...
-				PolicyID:        policy.ID,
-				PolicyVersionID: policy.VersionID,
-				Timestamp:       time.Now(),
-
+				PolicyID:            policy.ID,
+				PolicyIntegrationID: resource.IntegrationID,
+				PolicyVersionID:     policy.VersionID,
+				ResourceID:          resource.ID,
+				OutputIds:           policy.OutputIDs,
+				Timestamp:           time.Now(),
```
kostaspap, line 240: nit: Should be `time.Now().UTC()` since timezone is important here (doesn't matter in Lambdas, might matter in tests)

internal/log_analysis/alert_forwarder/forwarder/models.go
kostaspap: Yes, it might make sense to do some consolidation. I am not sure if we can have only 1 type of alerts - the alert Delivery payload (now named ... `Alert`) is different from what is stored in the Alerts API. But probably we can do consolidation, in which case if log_analysis/alert_forwarder and compliance/alert_forwarder use the same message, that message should probably live in alerts_api.

internal/compliance/alert_forwarder/main/lambda.go
s0l0ist: Ah thanks!

internal/compliance/alert_forwarder/main/config.go
```diff
@@ -0,0 +1,49 @@
...
+type envConfig struct {
+	AlertsTable      string `required:"true" split_words:"true"`
+	AlertingQueueURL string `required:"true" split_words:"true"`
+}
+
+// Setup parses the environment and builds the AWS and http clients.
+func Setup() {
+	envconfig.MustProcess("", &env)
+
+	awsSession = session.Must(session.NewSession())
```
austinbyers, line 46: Worth adding extra retries here?

```suggestion
	awsSession = session.Must(session.NewSession(aws.NewConfig().WithMaxRetries(10)))

File: internal/compliance/alert_forwarder/main/lambda.go [hunk 17]
```diff
@@ -32,30 +32,58 @@ import (
...
+var (
+	validate = validator.New()
+	handler  *forwarder.Handler
+)
 
 func main() {
-	lambda.Start(reporterHandler)
+	// Required only once per Lambda container
+	Setup()
+	// TODO: revisit this. Not sure why we neeed Dimension sets and why just an array of dimensions is not enough
```
austinbyers, line 49: Is this a copy-pasta comment or is it still applicable?

internal/compliance/alert_processor/processor/processor.go
```diff
@@ -211,17 +213,29 @@ func getAlertConfigPolicy(event *models.ComplianceNotification) (*alertmodel.Ale
...
 		},
 		policy.AutoRemediationID != "", // means we can remediate
 		nil
 }
+
+// generates an ID from the policyID (policy name) and the current timestamp.
+func GenerateAlertID(event *models.ComplianceNotification) *string {
+	key := event.PolicyID + ":" + event.Timestamp.String()
+	keyHash := md5.Sum([]byte(key)) // nolint(gosec)
+	encoded := hex.EncodeToString(keyHash[:])
```
austinbyers, line 239: Clever, I've wondered how to convert a byte array into a slice, the `[:]` would do the trick

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -303,60 +302,116 @@ func filterByLogType(filter *expression.ConditionBuilder, input *models.ListAler
...
 }
 
+// filterByResourceType - filters by list of resource types
+func filterByResourceType(filter *expression.ConditionBuilder, input *models.ListAlertsInput) {
+	if len(input.ResourceTypes) > 0 {
+		// Start with the first known key
+		multiFilter := expression.Name(ResourceTypesKey).Contains(input.ResourceTypes[0])
+
+		// Then add or conditions starting at a new slice from the second index
+		for _, resourceType := range input.LogTypes[1:] {
```
austinbyers, line 312: You're using `ResourceTypes` for the first index and `LogTypes` for the subsequent indices - which is it?

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -303,60 +302,116 @@ func filterByLogType(filter *expression.ConditionBuilder, input *models.ListAler
...
 	if len(input.Type) > 0 {
+		// Start with the first known key
 		var multiFilter expression.ConditionBuilder
-		switch input.Type {
-		case alertdeliverymodels.RuleErrorType:
-			multiFilter = expression.Equal(expression.Name(TypeKey), expression.Value(input.Type))
-		case alertdeliverymodels.RuleType:
-			// Alerts for rule matches don't always have the attribute specified
+
+		// Rule errors don't always have the attribute specified.
```
austinbyers, line 326: This is just a backwards compatibility thing, right? If so, that'd be a helpful note in the comment

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -303,60 +302,116 @@ func filterByLogType(filter *expression.ConditionBuilder, input *models.ListAler
...
+		for _, alertType := range input.Type[1:] {
+			// Rule errors don't always have the attribute specified.
+			if alertType == alertdeliverymodels.RuleErrorType {
+				multiFilter = multiFilter.
+					Or(
+						expression.AttributeNotExists(expression.Name(TypeKey)),
+						expression.Equal(expression.Name(TypeKey), expression.Value(alertType)),
+					)
+			} else {
+				multiFilter = multiFilter.Or(expression.Name(TypeKey).Equal(expression.Value(alertType)))
```
austinbyers, line 347: All of this code is identical to the section above, right? Might as well make it a little helper function:

```go
func alertTypeFilter(alertType string) *expression.ConditionBuilder {
	var multiFilter expression.ConditionBuilder

	// Rule errors don't always have the attribute specified.
	if alertType == alertdeliverymodels.RuleErrorType {
		multiFilter = multiFilter.
			Or(
				expression.AttributeNotExists(expression.Name(TypeKey)),
				expression.Equal(expression.Name(TypeKey), expression.Value(alertType)),
			)
	} else {
		multiFilter = multiFilter.Or(expression.Name(TypeKey).Equal(expression.Value(alertType)))
	}
	return multiFilter
}
```

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -303,60 +302,116 @@ func filterByLogType(filter *expression.ConditionBuilder, input *models.ListAler
...
+	}
 
-// filterByRuleIDContains - filters by a name that contains a string (case insensitive)
-func filterByRuleIDContains(input *models.ListAlertsInput, alert *AlertItem) *AlertItem {
-	if alert != nil && input.RuleIDContains != nil && !strings.Contains(
-		strings.ToLower(alert.RuleID),
-		strings.ToLower(*input.RuleIDContains),
+	if strings.Contains(
+		strings.ToLower(alert.PolicyDisplayName),
+		strings.ToLower(*input.NameContains),
```
austinbyers, line 401: Nit: you call `strings.ToLower(*input.NameContains)` several times in this function. Save a few cycles and compute it once?

```go
nameContains := strings.ToLower(*input.NameContains)
```

That might also make the lines short enough to consolidate, these tall blocks are disorienting me :)

```go
	if strings.Contains(strings.ToLower(alert.ResourceID), nameContains) {
		return alert
	}
```

austinbyers (approved):
Dynamo and CloudFormation code looking pretty good, only one blocking issue I saw in a filter expression, but approving assuming you fix that one

api/lambda/delivery/models/api.go
```diff
@@ -140,7 +140,16 @@ type Alert struct {
...
 
-	// AnalysisDescription is the description of the rule that triggered the alert.
+	// ResourceTypes is the set of resources that could trigger the alert.
+	ResourceTypes []string `json:"resourceTypes,omitempty"`
+
+	// ResourceID is the ID of the failing resource in the policy.
+	ResourceID string `json:"resourceId,omitempty"`
+
+	// AnalysisIntegrationID is the ID of the source integration for the rule/policy that failed.
+	AnalysisIntegrationID string `json:"analysisIntegrationId,omitempty"`
```
kostaspap, line 150: Forward-facing docs use the term `sourceId`. Maybe you want to use that, `integrationId` is older term iirc

deployments/cloud_security.yml
```diff
@@ -315,19 +316,29 @@ Resources:
             BatchSize: 1
       FunctionName: panther-alert-forwarder
       # <cfndoc>
-      # The `panther-alert-forwarder` lambda reads from the ddb stream for the table `panther-alert-forwarder`
-      # and sends them to the `panther-alerts-queue` sqs queue.
+      # This lambda reads from a DDB stream for the `panther-alert-forwarder` table and writes alerts to the `panther-log-alert-info` ddb table.
+      # It also forwards alerts to `panther-alerts-queue` SQS queue where the appropriate Lambda picks them up for delivery.
```
kostaspap, line 320: 👍

internal/compliance/alert_forwarder/forwarder/forwarder.go
```diff
@@ -19,42 +19,118 @@ package forwarder
...
-// Handle forwards an alert to the alert delivery SQS queue
-func Handle(event *models.Alert) error {
-	zap.L().Info("received alert", zap.String("policyId", event.AnalysisID))
+type Handler struct {
+	SqsClient        sqsiface.SQSAPI
+	DdbClient        dynamodbiface.DynamoDBAPI
+	AlertTable       string
+	AlertingQueueURL string
+	MetricsLogger    metrics.Logger
+}
```
kostaspap, line 44: 👍 👍 👍 👍

kostaspap (approved):
Wow, quite a big change! But I've seen you 've done some good cleaup and updated the tests. Kudos!

kostaspap:
This is a big change. Make sure to do some E2E testing. E.g. do rule alerts, rule error alerts work as before?

3nvi:
> This is a big change. Make sure to do some E2E testing. E.g. do rule alerts, rule error alerts work as before?

Would you prefer to target a feature branch or do we just need E2E tests?

kostaspap:
> > This is a big change. Make sure to do some E2E testing. E.g. do rule alerts, rule error alerts work as before?
> 
> Would you prefer to target a feature branch or do we just need E2E tests?

I meant to deploy the code and verify the alerting functionality works as expected. Our current E2E tests don't cover interactions across multiple lambdas - and this PR changes quite a few

api/graphql/schema.graphql
```diff
@@ -99,12 +99,11 @@ input ListAlertsInput {
...
   severity: [SeverityEnum]
   logTypes: [String!]
-  type: AlertTypesEnum
+  resourceTypes: [String!]
+  type: [AlertTypesEnum!]
   nameContains: String
   createdAtBefore: AWSDateTime
   createdAtAfter: AWSDateTime
-  ruleIdContains: String
-  alertIdContains: String
```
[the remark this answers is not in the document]
s0l0ist: Two reasons:
1. They were never actually used
2. The nameContains also searches these fields now

api/lambda/alerts/models/api.go
```diff
@@ -83,17 +81,16 @@ type ListAlertsInput struct {
 	ExclusiveStartKey *string `json:"exclusiveStartKey"`
 
 	// Filtering
-	Type            string     `json:"type" validate:"omitempty,oneof=RULE RULE_ERROR"`
+	Type            []string   `json:"type" validate:"omitempty,dive,oneof=RULE RULE_ERROR POLICY"`
```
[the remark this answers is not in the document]
s0l0ist: I think this is necessary because the query may not specify `Types` at all in which case there's no need to validate anything at all.

internal/compliance/alert_forwarder/main/lambda.go
s0l0ist: copy-pasta, still applicable.

internal/compliance/alert_processor/processor/processor.go
s0l0ist: that would be @kostaspap who figured it out, I stole it 😄

internal/log_analysis/alerts_api/table/list.go
s0l0ist: Whoops... fixed

internal/log_analysis/alerts_api/table/list.go
s0l0ist: Theres a couple of places/filters that are like this and it is because we need to create an initial condition first (limitation of condition expressions) before looping and adding the other conditions present in the argument slice.

api/lambda/alerts/models/api.go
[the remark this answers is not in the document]
austinbyers: Right, but if the list is empty, then `dive` won't do anything, which means it won't validate at all :) Same thing either way, but it's not a big deal, you can leave it as is

internal/log_analysis/alerts_api/table/list.go
austinbyers: Yes, this is a frustrating limitation of the expression library. I'm just saying we can make helper functions to do that instead of copying code, but again, not a big deal either way

api/lambda/delivery/models/api.go
s0l0ist: renamed to `AnalysisSourceID`.

s0l0ist:
> > > This is a big change. Make sure to do some E2E testing. E.g. do rule alerts, rule error alerts work as before?
> > 
> > 
> > Would you prefer to target a feature branch or do we just need E2E tests?
> 
> I meant to deploy the code and verify the alerting functionality works as expected. Our current E2E tests don't cover interactions across multiple lambdas - and this PR changes quite a few

I've deployed, ran integration tests and performed manual tests to ensure it is working as expected.

PR #2106, opened 2020-11-24
Dynamic Fields Support

wey-chiang:
## Background

Adds support for dynamic fields, namely the following fields:

>  Description, Destination Override, Reference, Runbook, Severity

## Changes

Added generated fields to 

- analysisModels - rule_result
- lambda analysis - engine
- lambda analysis - rules
- lambda analysis - api

Updated:
- `internal/compliance/alert_processor/processor/processor.go`
- `internal/core/analysis_api/analysis/rule_engine.go`
- `internal/core/alert_delivery` - Alert Delivery API
- `internal/log_analysis/alert_forwarder` - Alert Forwarder API
- `internal/log_analysis/alerts_api` - Alerts API
- `internal/log_analysis/rules_engine` Python API
To support dynamically generated fields

## Testing

Deployed into `us-east-2` and manually verified through SNS:
--> Static Severity Overriden by Dynamic Severity
--> Dynamic Destination Overriding Dynamic Severity & Static Dest. Override

**Changes from master will need to be merged in and tested**
**Also, some works needs to be done to fix linting issues and unit tests, probably CI tests as well**


internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -54,7 +55,17 @@ func (API) DeliverAlert(input *deliveryModels.DeliverAlertInput) (*deliveryModel
...
+	// First check if there is a destination override set
+	var alertOutputMap AlertOutputMap
+	if alert.DestinationOverride != nil {
+		alertOutputMap, err = getAlertOutputMapping(alert, alert.DestinationOverride)
+		if err == nil {
+			alertOutputMap, err = getAlertOutputMapping(alert, input.OutputIds)
+		}
+	} else {
+		alertOutputMap, err = getAlertOutputMapping(alert, input.OutputIds)
+	}
```
kostaspap, line 67: All the other overrides by the Rules are happening in the forwarding. I really think we should be handling this override there as well - alert delivery should be agnostic to override logic. 
It seems that the alert forwarder could be populating the appropriate `OutputIds` since it already hits the Analysis API to get information about the rule. @s0l0ist thoughts?

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -238,15 +245,54 @@ func (h *Handler) sendAlertNotification(rule *ruleModel.Rule, alertDedup *AlertD
 	return nil
 }
 
-func getAlertTitle(rule *ruleModel.Rule, alertDedup *AlertDedupEvent) string {
-	if alertDedup.GeneratedTitle != nil {
-		return *alertDedup.GeneratedTitle
+func getCustomField(targetField string, rule *ruleModel.Rule, alertDedup *AlertDedupEvent) *string {
```
kostaspap, line 248: I think it's better split this to separate functions. Big switch statements like this are considered problematic - hard to read and a  pain to maintain and they keep growing and growing.  Passing a string as identifier seems fragile as well

internal/log_analysis/rules_engine/src/alert_merger.py
```diff
@@ -157,11 +156,40 @@ def _update_get_conditional(group_info: MatchingGroupInfo) -> AlertInfo:
     }
 
     if group_info.alert_context:
+        update_expression += ', #12=:12'
```
kostaspap, line 159: 👍

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -32,7 +32,11 @@
 MAX_DEDUP_STRING_SIZE = 1000
 
 # Maximum size for a title
-MAX_TITLE_SIZE = 1000
+MAX_CUSTOM_FIELD_SIZE = 1000
```
kostaspap, line 35: 1. Comment needs to be updated
2. Severity override needs to be checked against the valid options (INFO, LOW, MEDIUM...)

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -54,7 +55,17 @@ func (API) DeliverAlert(input *deliveryModels.DeliverAlertInput) (*deliveryModel
...
 
 	// Get our Alert -> Output mappings. We determine which destinations an alert should be sent.
-	alertOutputMap, err := getAlertOutputMapping(alert, input.OutputIds)
+	// First check if there is a destination override set
+	var alertOutputMap AlertOutputMap
+	if alert.DestinationOverride != nil {
+		alertOutputMap, err = getAlertOutputMapping(alert, alert.DestinationOverride)
+		if err == nil {
+			alertOutputMap, err = getAlertOutputMapping(alert, input.OutputIds)
+		}
```
nhakmiller, line 64: Not sure I follow this, we set the `alertOutputMap`, and then if there wasn't an error setting it, we immediately overwrite it with different values?

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -212,35 +289,80 @@ def _get_dedup(self, event: Mapping, title: Optional[str], use_default_on_except
...
-    def _get_title(self, event: Mapping, use_default_on_exception: bool = True) -> Optional[str]:
-        if not self._has_title:
+    def _get_custom_field(self, event: Mapping, target_field: str, use_default_on_exception: bool = True) -> Optional[str]:
+        _custom_field_map = {
+            'title': (self._has_title, self._module.title),
+            'description': (self._has_description, self._module.description),
+            'reference': (self._has_reference, self._module.reference),
+            'severity': (self._has_severity, self._module.severity),
+            'runbook': (self._has_runbook, self._module.runbook),
+        }
```
kostaspap, line 309: Creating this map multiple times per processed event seems like a waste of cycles. This should really be initialized once, when the rule is created.

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -115,19 +126,53 @@ func populateAlertData(alertItem *alertTable.AlertItem) (*deliveryModels.Alert,
...
+	var alertReference models.Reference
+	var alertRunbook models.Runbook
+	// var alertDescription, alertReference, alertRunbook *string
+	var alertDestinationOverride []string
+
+	if alertItem.Description != nil && *alertItem.Description != "" {
+		alertDescription = models.Description(*alertItem.Description)
+	} else {
+		alertDescription = rule.Description
+	}
```
nhakmiller, line 140: Instead of having this:

```
	var alertDescription models.Description
	if alertItem.Description != nil && *alertItem.Description != "" {
		alertDescription = models.Description(*alertItem.Description)
	} else {
		alertDescription = rule.Description
	}
```

I think this is a bit cleaner:

```
	alertDescription = rule.description
	if alertItem.Description != nil && *alertItem.Description != "" {
		alertDescription = models.Description(*alertItem.Description)
	}
```

Additionally, we can use `aws.StringValue` here and further simplify this check to:

```
	alertDescription = rule.description
	if aws.StringValue(alertItem.Description) != "" {
		alertDescription = models.Description(*alertItem.Description)
	}
```

I'd suggest using this pattern for each of these.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -212,35 +289,80 @@ def _get_dedup(self, event: Mapping, title: Optional[str], use_default_on_except
...
+                if use_default_on_exception:
+                    self.logger.warning(
+                        '%s method raised exception. Using default. Exception: %s',
+                        target_field,
+                        err,
+                    )
+                    return None
+                raise
+
+            if len(custom_field) > MAX_CUSTOM_FIELD_SIZE:
```
kostaspap, line 330: The 1000 limit for title was picked on something that we considered would be appropriate for the title. 
Some of the fields could have even bigger length e.g. description could have more characters, while others need different validation e.g. Severity.

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -115,19 +126,53 @@ func populateAlertData(alertItem *alertTable.AlertItem) (*deliveryModels.Alert,
...
+	} else {
+		alertRunbook = rule.Runbook
+	}
+
+	// This will just be nil since a rule should never have a
+	if alertItem.DestinationOverride != nil {
+		alertDestinationOverride = alertItem.DestinationOverride
+	} else {
+		alertDestinationOverride = nil
+	}
```
nhakmiller, line 159: You can simplify this (and remove line 134  above) to just:

```
alertDestinationOverride = alertItem.DestinationOverride
```

If  `alertItem.DestinationOverride` is `nil`,  then the else cause is just redundant.

internal/log_analysis/rules_engine/tests/test_rule.py
```diff
@@ -17,7 +17,7 @@
 from unittest import TestCase
 import json
 
-from ..src.rule import MAX_DEDUP_STRING_SIZE, MAX_TITLE_SIZE, Rule, RuleResult, TRUNCATED_STRING_SUFFIX
+from ..src.rule import MAX_DEDUP_STRING_SIZE, MAX_CUSTOM_FIELD_SIZE, Rule, RuleResult, TRUNCATED_STRING_SUFFIX
```
kostaspap, line 20: We need unit tests to validate that the overrides work

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -156,9 +156,13 @@ func (h *Handler) storeNewAlert(rule *ruleModel.Rule, alertDedup *AlertDedupEven
 	alert := &Alert{
 		ID:                  generateAlertID(alertDedup),
 		TimePartition:       defaultTimePartition,
-		Severity:            string(rule.Severity),
+		Severity:            *getCustomField("severity", rule, alertDedup),
 		RuleDisplayName:     getRuleDisplayName(rule),
-		Title:               getAlertTitle(rule, alertDedup),
+		Title:               *getCustomField("title", rule, alertDedup),
+		Description:         getCustomField("description", rule, alertDedup),
```
nhakmiller, line 162: It seems some of these are pointers and some are just strings? Unless there is a specific reason for this, it  might be preferred to not mix & match them like this and just make them all strings or all string pointers.

kostaspap (commented):
Challenging task but seems to be going in the right direction. Good start @wey-chiang !!

I wanted to note that users can currently test the rule functions (title, dedup, alert_context etc) through UI tests.
We need to make sure that the UI and backend code (analysis API) is updated to handle the new functions as well. Just double checking that you have already planned this work.

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -238,15 +245,54 @@ func (h *Handler) sendAlertNotification(rule *ruleModel.Rule, alertDedup *AlertD
...
+				return ruleDisplayName
+			}
+			return aws.String(string(rule.ID))
+		}
+	case targetField == "description":
+		if alertDedup.GeneratedDescription != nil {
+			return alertDedup.GeneratedDescription
+		} else {
+			return aws.String(string(rule.Description))
+		}
```
nhakmiller, line 265: I think a general pattern we follow in go is to not have `else {  return <..> }` statements. It can be re-written with less indentation like this:

```
		if alertDedup.GeneratedDescription != nil {
			return alertDedup.GeneratedDescription
		}
		return aws.String(string(rule.Description))
```

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -238,15 +245,54 @@ func (h *Handler) sendAlertNotification(rule *ruleModel.Rule, alertDedup *AlertD
...
+	default:
+		zap.L().Error("Unexpected type for getCustomField: ", zap.String("targetField", targetField))
+		return nil
 	}
-	ruleDisplayName := getRuleDisplayName(rule)
-	if ruleDisplayName != nil {
-		return *ruleDisplayName
+}
+
+func getDestinationOverride(alertDedup *AlertDedupEvent) []string {
```
nhakmiller, line 290: It surprised me at first that  this wasn't part of the above function, maybe we could rename the above from  `getCustomField` to `getRuleCustomField` and/or add a comment explaining it, and rename this function to `getAlertDestinationOverride`? Just to disambiguate their purposes.

internal/log_analysis/alert_forwarder/forwarder/models.go
```diff
@@ -36,20 +36,25 @@ const (
 	alertTableUpdateTimeAttribute = "updateTime"
 )
 
+// TODO: Update Generated custom fields dynamodbav tags
```
nhakmiller, line 39: Is this `TODO` still needed? Is this work blocking?

internal/log_analysis/alert_forwarder/forwarder/models.go
```diff
@@ -36,20 +36,25 @@ const (
...
+	RuleID                       string    `dynamodbav:"ruleId"`
+	RuleVersion                  string    `dynamodbav:"ruleVersion"`
+	DeduplicationString          string    `dynamodbav:"dedup"`
+	CreationTime                 time.Time `dynamodbav:"creationTime"`
+	UpdateTime                   time.Time `dynamodbav:"updateTime"`
+	EventCount                   int64     `dynamodbav:"eventCount"`
+	LogTypes                     []string  `dynamodbav:"logTypes,stringset"`
+	AlertContext                 *string   `dynamodbav:"context,string"`
+	Type                         string    `dynamodbav:"type"`
+	GeneratedTitle               *string   `dynamodbav:"title,string"`       // The title that was generated dynamically using Python. Might be null.
```
nhakmiller, line 51: Could you clarify why this field has a json tag, but the ones below are set to `-`? I'm not familiar enough with this code to be sure it's correct/incorrect but it seems strange.

internal/log_analysis/alert_forwarder/forwarder/models.go
```diff
@@ -136,11 +144,39 @@ func FromDynamodDBAttribute(input map[string]events.DynamoDBAttributeValue) (eve
...
+	// Custom Fields
 	generatedTitle := getOptionalAttribute("title", input)
 	if generatedTitle != nil {
 		result.GeneratedTitle = aws.String(generatedTitle.String())
 	}
 
+	generatedDescription := getOptionalAttribute("description", input)
+	if generatedDescription != nil {
+		result.GeneratedDescription = aws.String(generatedDescription.String())
+	}
```
nhakmiller, line 156: I think this can be simplified to just:

```
	result.GeneratedDescription := getOptionalAttribute("description", input)
```

If `getOptionalAttribute` returns nil, then that's the same thing as just not specifying `result.GeneratedDescription` at all which is what happens with your conditional.

internal/log_analysis/alert_forwarder/main/lambda.go
```diff
@@ -98,6 +98,7 @@ func reporterHandler(lc *lambdacontext.LambdaContext, event events.DynamoDBEvent
 			continue
 		}
 
+		// TODO add custom handling for getCustomField
```
nhakmiller, line 101: Is this `TODO` still correct?

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -32,7 +32,11 @@
 MAX_DEDUP_STRING_SIZE = 1000
 
 # Maximum size for a title
-MAX_TITLE_SIZE = 1000
+MAX_CUSTOM_FIELD_SIZE = 1000
+
+# Maximum number of Summary Attributes
+MAX_DESTINATION_OVERRIDE_SIZE = 10
```
nhakmiller, line 38: This comment doesn't align with this variable name.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -137,6 +160,31 @@ def __init__(self, config: Mapping):
         else:
             self._has_title = False
 
+        if hasattr(self._module, 'description'):
+            self._has_description = True
+        else:
+            self._has_description = False
```
nhakmiller, line 166: In case you haven't noticed yet, I am allergic to `else` statements. This could be re-written as:

```
        self._has_description = False
        if hasattr(self._module, 'description'):
            self._has_description = True
```

or with a ternary condition (which can easily be abused, but I think is appropriate here):

```
self._has_description = True if hasattr(self._module, 'description') else False
```

nhakmiller (commented):
Thanks for this! A couple notes:

I know it's a draft PR, but I think the PR description could stand to be cleaned up a bit with some nice formatting. Helps direct attention to where it is needed.

Merging master and the `analysis-api` removal changes will be a challenge as we discussed, @austinbyers could you possibly point us in the right direction on this?

I'm going to tag @giorgosp and @kostaspap for review as well as they work on / own the rules engine primarily so their input will be useful, and given the short time frame before the sprint  is up the more eyes the better I think.

wey-chiang:
> I wanted to note that users can currently test the rule functions (title, dedup, alert_context etc) through UI tests.
> We need to make sure that the UI and backend code (analysis API) is updated to handle the new functions as well. Just double checking that you have already planned this work.
The backend (analysis API) has been updated in `api/gateway/analysis/models/rule_result.go` covers this. 
However, we do not store the outputs in the `panther-analysis` ddb table as they are dynamically generated and stored in the alert itself (in `panther-log-alert-info` and `panther-log-alert-dedup`)

api/lambda/delivery/models/api.go
```diff
@@ -149,6 +149,12 @@ type Alert struct {
 	// Version is the S3 object version for the policy.
 	Version *string `json:"version,omitempty"`
 
+	// Reference is the user-provided triage information
+	Reference *string `json:"reference,omitempty"`
+
+	// DestinationOverride is the custom user-provided destination override
+	DestinationOverride []string `json:"destinationOverride,omitempty" validate:"dive,uuid4"`
```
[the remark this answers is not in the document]
wey-chiang: Removing this to have the destination override in `OutputIds` will require shuffling around some of the logic.

The reason I included the `DestinationOverride` field is so that if the dynamic destination override does not properly map to an output, we can handle that by just using the statically set destination override

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
wey-chiang: I combined them into a single function for brevity -- I'll split them up into separate functions

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: Added a check to the severity override in the `class Rule: run` function

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: Makes sense -- I was hesitant on this as well.
I'll move this into the `__init__` func

internal/core/alert_delivery/api/deliver_alert.go
wey-chiang: Doh, that needs to be changed to `if err != nil`, good catch!

internal/core/alert_delivery/api/deliver_alert.go
wey-chiang: Nice, yeah I felt this part was a bit clunky. Cleaned up!

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
wey-chiang: Makes sense, changed severity and title to consistently be string pointers

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
wey-chiang: Gotcha, this looks cleaner anyways! will fix

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
wey-chiang: I broke out the `getCustomField` func into distinct getter methods -- hopefully that change addresses this concern.

internal/log_analysis/alert_forwarder/forwarder/models.go
wey-chiang: It's unclear to me if the rest of the custom fields require the `dynamodbav` field tags

internal/log_analysis/alert_forwarder/forwarder/models.go
wey-chiang: @kostaspap Perhaps you could enlighten us

internal/log_analysis/alert_forwarder/main/lambda.go
wey-chiang: Nope, will remove.

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: Fixed, good catch

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: Love that, way cleaner!

nhakmiller (commented):
Good cleanup! I'll hold off on approving until master is merged in, as I'm sure there will be some large changes from that. But I'd say it's ready to start that process now.

internal/core/alert_delivery/api/outputs.go
```diff
@@ -36,22 +36,34 @@ func getAlertOutputs(alert *deliveryModels.Alert) ([]*outputModels.AlertOutput,
...
 				}
 			}
 		}
 		return defaultsForSeverity, nil
 	}
 
-	// Otherwise, return the specified output overrides for the alert
+	// If alert has a dynamically set destination override, return the specified output overrides for the alert
 	overrideOutputs := []*outputModels.AlertOutput{}
+	if alert.DestinationOverride != nil && len(alert.DestinationOverride) >= 0 {
```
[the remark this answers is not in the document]
wey-chiang: Yeah, that's the only real situation I can see that happening -- if `alert.DestinationOverride` is an empty but non-nil slice, this case will then be skipped because of the length check and the final case (using the OutputID) will be used.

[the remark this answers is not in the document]
giorgosp: For nil slices, `len(slice) == 0`, so the nil check can be removed. The condition can just become
` if len(alert.DestinationOverride) > 0 {` .

You can also invert the double `for` loop and get rid of the `if` altogether.
```
for _, output := range outputs {
			for _, outputID := range alert.DestinationOverride {
```
would become
```
for _, outputID := range alert.DestinationOverride {
     for _, output := range outputs {
```

By the way, shouldn't the field be called DestinationOverride**s**?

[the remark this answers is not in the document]
nhakmiller: The problem is we still want to enter this conditional if the length is 0 and the slice is not `nil`, so `if len(alert.DestinationOverride) > 0 {` is insufficient. We could condense this to `if alert.DestinationOverride != nil` though.

We also can't get rid of the if statement all together, because we only want to return on line 63 if the `alert.DesinationOverride` is not `nil`, regardless of its length. If it is `nil`, there is different code we want to execute instead.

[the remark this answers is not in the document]
giorgosp: I think we should handle  `return None` and `return []` from `def destination_override()` the same. 
Returning `None` or `[]` would mean that the user doesn't want to override destinations. But if they want to silence the alert, `rule()` should be the place to do that, by returning `False`.

internal/core/alert_delivery/api/outputs.go
```diff
@@ -36,22 +36,34 @@ func getAlertOutputs(alert *deliveryModels.Alert) ([]*outputModels.AlertOutput,
...
 					defaultsForSeverity = append(defaultsForSeverity, output)
 				}
 			}
 		}
 		return defaultsForSeverity, nil
 	}
 
-	// Otherwise, return the specified output overrides for the alert
+	// If alert has a dynamically set destination override, return the specified output overrides for the alert
 	overrideOutputs := []*outputModels.AlertOutput{}
```
giorgosp, line 54: I would rename `overrideOutputs` to `alertOutputs` as it's more accurate.
We want to return the alert's outputs here, and we either get them from overrides or default ones.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -212,35 +273,73 @@ def _get_dedup(self, event: Mapping, title: Optional[str], use_default_on_except
...
+            return custom_field
+        return None
 
-        try:
-            title_string = self._run_command(self._module.title, event, str)
-        except Exception as err:  # pylint: disable=broad-except
-            if use_default_on_exception:
-                self.logger.warning('title method raised exception. Using default. Exception: %s', err)
+    def _get_destination_override(self, event: Mapping) -> Optional[List[str]]:
+        if self._has_destination_override:
```
giorgosp, line 324: Invert the `if` to reduce nesting and improve readability by returning early.
```
if not self._has_destination_override:
   return None

...
```

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -132,23 +156,27 @@ def __init__(self, config: Mapping):
...
+        self._has_severity = hasattr(self._module, 'severity')
+        self._has_runbook = hasattr(self._module, 'runbook')
+        self._has_destination_override = hasattr(self._module, 'destination_override')
+        self._has_dedup = hasattr(self._module, 'dedup')
+        self._has_alert_context = hasattr(self._module, 'alert_context')
 
         self._default_dedup_string = 'defaultDedupString:{}'.format(self.rule_id)
 
+        # Added to support custom fields
+        self._custom_field_map = {
```
giorgosp, line 172: Is this map really needed? Eg, `has_title` can be calculated wherever/whenever we have access to`self._module`

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -132,23 +156,27 @@ def __init__(self, config: Mapping):
...
-            self._has_dedup = True
-        else:
-            self._has_dedup = False
-
-        if hasattr(self._module, 'alert_context'):
-            self._has_alert_context = True
-        else:
-            self._has_alert_context = False
+        # Check if the imported module has these attributes
+        self._has_title = hasattr(self._module, 'title')
```
giorgosp, line 160: If we don't really need `_custom_field_map` below, we could de-clutter `__init()__` by removing all these `self._has_title = hasattr(self._module, 'title')` statements and inline them in the appropriate getters (`_get_title()`) etc.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -212,35 +273,73 @@ def _get_dedup(self, event: Mapping, title: Optional[str], use_default_on_except
...
-        if not self._has_title:
+    def _get_custom_field(self, event: Mapping, target_field: str, use_default_on_exception: bool = True) -> Optional[str]:
+        if target_field not in self._custom_field_map:
+            self.logger.error(
+                'attempted to access field [%s], this should only access custom fields. ',
+                target_field,
+            )
             return None
+        has_field, command = self._custom_field_map[target_field]
+        if has_field:
```
giorgosp, line 294: Same comment here about inverting the `if` to reduce nesting.
This could be actually refactored to 
```
if not hasattr(self.module, field)
    return None

command = getattr(self.module, field)
try:
    self._run_command....
```
Will allow us to get rid of `_custom_field_map`.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -212,35 +273,73 @@ def _get_dedup(self, event: Mapping, title: Optional[str], use_default_on_except
...
+                len(dedup_string),
             )
             num_characters_to_keep = MAX_DEDUP_STRING_SIZE - len(TRUNCATED_STRING_SUFFIX)
             return dedup_string[:num_characters_to_keep] + TRUNCATED_STRING_SUFFIX
 
         return dedup_string
 
-    def _get_title(self, event: Mapping, use_default_on_exception: bool = True) -> Optional[str]:
-        if not self._has_title:
+    def _get_custom_field(self, event: Mapping, target_field: str, use_default_on_exception: bool = True) -> Optional[str]:
```
giorgosp, line 286: Would `_get_field_output` or `_get_function_output` better names for `_get_custom_field()`?
`target_field` could just be `field` or `function`.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -200,7 +257,11 @@ def _get_dedup(self, event: Mapping, title: Optional[str], use_default_on_except
             dedup_string = self._run_command(self._module.dedup, event, str)
         except Exception as err:  # pylint: disable=broad-except
             if use_default_on_exception:
-                self.logger.warning('dedup method raised exception. Defaulting dedup string to "%s". Exception: %s', self.rule_id, err)
+                self.logger.warning(
+                    "dedup method raised exception. Defaulting dedup string to '%s'. Exception: %s",
```
giorgosp, line 261: nit: Could reduce number of lines for this call, to
```
self.logger.warning(
                    'maximum field [%s] length is [%d]. [%s]] for rule with ID [%s] is [%d] characters. Truncating.',
                    target_field, MAX_CUSTOM_FIELD_SIZE, target_field, self.rule_id, len(custom_field))
```
Same for other calls taking too many lines.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -212,35 +273,73 @@ def _get_dedup(self, event: Mapping, title: Optional[str], use_default_on_except
...
+                return custom_field[:num_characters_to_keep] + TRUNCATED_STRING_SUFFIX
+            return custom_field
+        return None
 
-        try:
-            title_string = self._run_command(self._module.title, event, str)
-        except Exception as err:  # pylint: disable=broad-except
-            if use_default_on_exception:
-                self.logger.warning('title method raised exception. Using default. Exception: %s', err)
+    def _get_destination_override(self, event: Mapping) -> Optional[List[str]]:
```
giorgosp, line 323: Seems like we could use the `_get_custom_field()` function for destination overrides too?

wey-chiang: I was running into issues with having the two combined before -- however since I refactored the `self._run_command` function, I'm not sure if this issue is still persistent, I suspect that it is not.

Just merged in from master, will refactor this after I finish up testing

internal/core/alert_delivery/api/outputs.go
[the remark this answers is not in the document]
wey-chiang: Ahh, that sounds like useful functionality we would want to preserve! Will apply the proposed logic

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: Nah, will default to having appropriate getters

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: See above, will favor individual getter methods

wey-chiang:
Addressed the rest of the changes, deploying now to ensure these didn't break anything. Will update the PR with the results.

wey-chiang:
Note the `panther-alert-delivery-api` is showing `json: unsupported type: map[interface {}]interface {}`
Not sure what's causing this atm

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -211,57 +274,139 @@ def _get_dedup(self, event: Mapping, title: Optional[str], use_default_on_except
...
+        if not hasattr(self._module, 'destination_override'):
+            return None
 
-    def _get_alert_context(self, event: Mapping, use_default_on_exception: bool = True) -> Optional[str]:
-        if not self._has_alert_context:
+        try:
+            command = getattr(self._module, 'destination_override')
+            destination_override = self._run_command(command, event, list())
+        except Exception as err:  # pylint: disable=broad-except
+            self.logger.warning('_get_destination_override method raised exception. Exception: %s', err)
```
kostaspap, line 314: The previous code had a variable passed to all methods: `use_default_on_exception`. 
The purpose of this exception: In case of normal operation, if any of the methods that generate the new fields fail, we just want to use the default. 
If any of the methods fail while user is testing through the UI, we want to return the exception.

kostaspap:
> > I wanted to note that users can currently test the rule functions (title, dedup, alert_context etc) through UI tests.
> > We need to make sure that the UI and backend code (analysis API) is updated to handle the new functions as well. Just double checking that you have already planned this work.
> 
> The backend (analysis API) has been updated in `api/gateway/analysis/models/rule_result.go` covers this.
> However, we do not store the outputs in the `panther-analysis` ddb table as they are dynamically generated and stored in the alert itself (in `panther-log-alert-info` and `panther-log-alert-dedup`)

What I meant to say is that this feature requires FE and BE changes. We should make sure that they are planned by FE team. 

If a user tests a rule in the UI and any of the the `reference`, `severity` etc methods fail, the UI should display that failure. 
On the other hand, if any of the reference, severity, etc methods fail during normal operation, we shouldn't fail the processing - we should just use the defaults. We need to make sure that both usecases are tested.

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -104,16 +115,34 @@ func populateAlertData(alertItem *alertTable.AlertItem) (*deliveryModels.Alert,
...
+		alertDescription = *alertItem.Description
+	}
+
+	if aws.StringValue(alertItem.Reference) != "" {
+		alertReference = *alertItem.Reference
+	}
+
+	if aws.StringValue(alertItem.Runbook) != "" {
+		alertRunbook = *alertItem.Runbook
+	}
```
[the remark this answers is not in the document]
wey-chiang: Removed

internal/log_analysis/alert_forwarder/forwarder/forwarder.go
```diff
@@ -195,20 +199,23 @@ func (h *Handler) storeNewAlert(rule *ruleModel.Rule, alertDedup *AlertDedupEven
...
-		Version:      &alertDedup.RuleVersion,
+		CreatedAt:           alertDedup.UpdateTime,
+		OutputIds:           rule.OutputIDs,
+		AnalysisName:        getRuleDisplayName(rule),
+		Severity:            getSeverity(rule, alertDedup),
+		Tags:                rule.Tags,
+		Type:                alertDedup.Type,
+		Title:               getTitle(rule, alertDedup),
+		Version:             &alertDedup.RuleVersion,
+		// Custom Fields
```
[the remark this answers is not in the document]
wey-chiang: Reorganized lines to be alphabetical grouped by sections i.e. custom fields

internal/log_analysis/rules_engine/src/__init__.py
```diff
@@ -34,8 +34,14 @@ class EngineResult:
     error_message: Optional[str] = None
     rule_tags: List[str] = field(default_factory=list)
     rule_reports: Dict[str, List[str]] = field(default_factory=dict)
-    title: Optional[str] = None
     alert_context: Optional[str] = None
+    # custom fields
```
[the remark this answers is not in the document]
wey-chiang: Will change these to consistently reference `generated` fields

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -173,6 +182,38 @@ def run(self, event: Mapping, batch_mode: bool = True) -> RuleResult:
...
+
+        try:
+            rule_result.severity_output = self._get_severity(event)
+            if isinstance(rule_result.severity_output, str):
+                rule_result.severity_output = rule_result.severity_output.upper()
+                if rule_result.severity_output not in SEVERITY_TYPES:
+                    self.logger.error(
+                        "received invalid severity: [%s], expected to be one of: [%s]", str(rule_result.severity_output),
+                        str(SEVERITY_TYPES)
+                    )
```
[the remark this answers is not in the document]
wey-chiang: Removed the logger message and replaced with an AssertionError

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -185,22 +226,44 @@ def run(self, event: Mapping, batch_mode: bool = True) -> RuleResult:
...
+            return json.dumps({ALERT_CONTEXT_ERROR_KEY: alert_context_error})
+
+        return serialized_alert_context
+
     # Returns the dedup string for this rule match
     # If the rule match had a custom title, use the title as a deduplication string
     # If no title and no dedup function is defined, return the default dedup string.
     def _get_dedup(self, event: Mapping, title: Optional[str], use_default_on_exception: bool = True) -> str:
-        if not self._has_dedup:
+        if not hasattr(self._module, 'dedup'):
```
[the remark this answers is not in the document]
wey-chiang: It's the difference between the boolean being looked up in the `__init__` function versus in the `_get_field` functions. 

This _should_ not have a significant impact on performance

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -185,22 +226,44 @@ def run(self, event: Mapping, batch_mode: bool = True) -> RuleResult:
...
             return self._default_dedup_string
 
         try:
-            dedup_string = self._run_command(self._module.dedup, event, str)
+            command = getattr(self._module, 'dedup')
+            dedup_string = self._run_command(command, event, str)
         except Exception as err:  # pylint: disable=broad-except
             if use_default_on_exception:
-                self.logger.warning('dedup method raised exception. Defaulting dedup string to "%s". Exception: %s', self.rule_id, err)
+                self.logger.warning("dedup method raised exception. Defaulting dedup string to '%s'. Exception: %s", self.rule_id, err)
```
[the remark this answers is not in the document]
wey-chiang: Nope, will revert back.

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: Gotcha, will change the `_get_field` functions to accept the `use_default_on_exception` and raise exceptions instead of logging a warning when `use_default_on_exception` is false.

internal/core/alert_delivery/api/outputs.go
```diff
@@ -50,16 +50,25 @@ func getAlertOutputs(alert *deliveryModels.Alert) ([]*outputModels.AlertOutput,
...
+		for _, outputID := range alert.DestinationOverride {
+			if *output.OutputID == outputID {
+				alertOutputs = append(alertOutputs, output)
+			}
+		}
+	}
+
 	// Otherwise, return the specified output overrides for the alert
-	overrideOutputs := []*outputModels.AlertOutput{}
 	for _, output := range outputs {
```
nhakmiller, line 64: We only want to do this if we didn't execute the loop on  lines 55-61 right? I think we might need something like:

```
if  len(alertOutputs) > 0) {
   return alertOutputs, nil
}
```

just before this line. Otherwise we will be appending the static destination overrides to the dynamic destination overrides each time, which is not what we want to do I think.

internal/core/alert_delivery/outputs/jira.go
```diff
@@ -37,10 +37,10 @@ const (
 func (client *OutputClient) Jira(
 	alert *alertModels.Alert, config *outputModels.JiraConfig) *AlertDeliveryResponse {
 
-	description := "*Description:* " + aws.StringValue(alert.AnalysisDescription)
+	description := "*Description:* " + aws.StringValue(&alert.AnalysisDescription)
```
nhakmiller, line 40: `aws.StringValue(xyz)` takes a pointer to a string, and returns that string. If the pointer is `nil`, it returns an empty string.

The `&` operator returns a pointer to the variable it's in front of.

When you do `aws.StringValue(&someVariable)`, what you're doing is taking a string, getting a pointer to that string, then passing that pointer to a function, which checks if that pointer is nil, then returns an empty string if it is nil or returns the string the pointer points to if its not nil.

All that to say, you never need to write `aws.StringValue(&someVariable)`. Instead, just write `someVariable`.  This applies here and in a few places below.

internal/core/alert_delivery/outputs/outputs.go
```diff
@@ -184,24 +184,24 @@ func generateDetailedAlertMessage(alert *alertModels.Alert) string {
...
 func generateAlertTitle(alert *alertModels.Alert) string {
 	if alert.IsResent {
-		return "[Re-sent]: " + *alert.Title
+		return "[Re-sent]: " + alert.Title
 	}
 	switch alert.Type {
 	case alertModels.RuleType:
-		if alert.Title != nil {
-			return "New Alert: " + *alert.Title
+		if &alert.Title != nil {
```
nhakmiller, line 199: This will never be `nil`, this check should instead be:

`if alert.Title != "" { ... }`

`&someString` will never return `nil`, since a pointer is only `nil` if it doesn't point to something and generating a pointer with `&` will always be pointing to something. In go, the non-pointer equivalent of a `nil` check is comparing to the default value. The default value for strings is the empty string `""`.

internal/core/alert_delivery/outputs/slack.go
```diff
@@ -48,7 +48,7 @@ func (client *OutputClient) Slack(alert *alertModels.Alert, config *outputModels
 		},
 		{
 			"title": "Runbook",
-			"value": aws.StringValue(alert.Runbook),
+			"value": aws.String(alert.Runbook),
```
nhakmiller, line 51: This should probably just be:

```
			"value": alert.Runbook,
```

It was expecting a string before, I suspect it will still be expecting a string now (as opposed to a string pointer).

This might be related to the weird issue you were seeing that we talked about on slack.

internal/log_analysis/alerts_api/table/list.go
```diff
@@ -325,8 +325,8 @@ func filterByType(filter *expression.ConditionBuilder, input *models.ListAlertsI
 
 // filterByTitleContains - filters by a name that contains a string (case insensitive)
 func filterByTitleContains(input *models.ListAlertsInput, alert *AlertItem) *AlertItem {
-	if alert != nil && input.NameContains != nil && alert.Title != nil && !strings.Contains(
-		strings.ToLower(*alert.Title),
+	if alert != nil && input.NameContains != nil && &alert.Title != nil && !strings.Contains(
```
nhakmiller, line 328: Same comment as above, but `&alert.Title` will never be `nil`. The comparison should be `alert.Title == ""`.

internal/log_analysis/rules_engine/src/rule.py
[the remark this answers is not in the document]
nhakmiller: Seems this wasn't reverted back yet.

nhakmiller (commented):
Much closer, I think this will be in by today 🤞

internal/core/alert_delivery/api/outputs.go
wey-chiang: Yup, looks like there's no condition in-between the for loops so that check is definitely needed for this to work correctly

internal/core/alert_delivery/outputs/jira.go
wey-chiang: Doh, a lot of places as well. Good catch

internal/core/alert_delivery/api/deliver_alert.go
wey-chiang: So originally the `alert_forwarder` wasn't implementing this logic
However I've since changed the logic to populate `OutputIds` using a getter function that checks for the dynamic override 
This check shouldn't be needed anymore, will clean up.

internal/log_analysis/rules_engine/src/rule.py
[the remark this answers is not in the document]
wey-chiang: Weird, I think this is some weird goland formatting messing this up

nhakmiller (commented):
Deployed into my dev account and the custom alerting is working (sent to SNS -> Email), but  when I view the alert in the UI (or the `panther-alert-info` table) it still just has the default. I wonder if we can change that by modifying the `alerts-api`, or if it requires frontend changes

wey-chiang:
> Deployed into my dev account and the custom alerting is working (sent to SNS -> Email), but when I view the alert in the UI (or the `panther-alert-info` table) it still just has the default. I wonder if we can change that by modifying the `alerts-api`, or if it requires frontend changes

This required FE work since the Alerts page currently pulls from the `panther-analysis` table -- should be pulling from `panther-log-alert-info`

nhakmiller (approved):

internal/core/alert_delivery/api/outputs.go
```diff
@@ -50,16 +50,29 @@ func getAlertOutputs(alert *deliveryModels.Alert) ([]*outputModels.AlertOutput,
 		return defaultsForSeverity, nil
 	}
 
+	// If alert has a dynamically set destination override, return the specified output overrides for the alert
+	alertOutputs := []*outputModels.AlertOutput{}
+	for _, output := range outputs {
+		for _, outputID := range alert.DestinationOverride {
+			if *output.OutputID == outputID {
```
kostaspap, line 57: Safer to do `aws.StringValue(output.OutputID) == outputID`. Ideally, for cases like this the outputID should be a `string` instead of `*string` but you don't have to do it in this PR

api/lambda/delivery/models/api.go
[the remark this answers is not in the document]
kostaspap: I think @giorgosp made a comment about this? How how "Overrides"? We tend to use plural for slices to ease the people reading the code.

internal/core/alert_delivery/api/outputs.go
```diff
@@ -36,8 +36,8 @@ func getAlertOutputs(alert *deliveryModels.Alert) ([]*outputModels.AlertOutput,
 		return nil, err
 	}
 
-	// If alert doesn't have outputs IDs specified, return the defaults for the severity
-	if len(alert.OutputIds) == 0 {
+	// If alert has neither outputs IDs or dynamic dest. override specified, return the defaults for the severity
+	if len(alert.OutputIds) == 0 && alert.DestinationOverride == nil {
```
kostaspap, line 40: Seems better to do:
```
if len(alert.OutputIds) == 0 && len(alert.DestinationOverrides) == 0 {
...
```
Whether DestinationOverride is `nil` or empty slice can depend on the deserialization mechanism we use. Better be safe here.

internal/log_analysis/rules_engine/src/__init__.py
```diff
@@ -20,7 +20,7 @@
 from typing import Any, Dict, Optional, List
 
 
-# pylint: disable=too-many-instance-attributes
+# pylint: disable=too-many-instance-attributes,unsubscriptable-object
```
kostaspap, line 23: Why did we have to add this exclusion for pylint?

internal/log_analysis/rules_engine/tests/test_main.py
```diff
@@ -71,18 +71,28 @@ def test_direct_analysis_event_matching(self) -> None:
...
-                        'genericError': None,
-                        'alertContextError': None,
-                        'alertContextOutput': None,
-                        'dedupError': None,
-                        'dedupOutput': 'defaultDedupString:rule_id',
-                        'errored': False,
                         'id': 'event_id',
-                        'ruleError': None,
                         'ruleId': 'rule_id',
+                        'genericError': None,
```
kostaspap, line 76: Can you add a test that validates the new fields get populated properly?

internal/log_analysis/rules_engine/tests/test_rule.py
```diff
@@ -213,14 +213,20 @@ def test_rule_title_throws_exception(self) -> None:
         rule_body = 'def rule(event):\n\treturn True\ndef title(event):\n\traise Exception("test")'
         rule = Rule({'id': 'test_rule_title_throws_exception', 'body': rule_body, 'versionId': 'versionId'})
 
-        expected_result = RuleResult(matched=True, dedup_output='defaultDedupString:test_rule_title_throws_exception')
+        expected_result = RuleResult(
```
kostaspap, line 216: Please add tests for the new fields.

internal/log_analysis/rules_engine/src/__init__.py
wey-chiang: It was cluttering the `mage test:python` output, figured it made more sense to suppress it

api/lambda/delivery/models/api.go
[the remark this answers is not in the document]
wey-chiang: Sure, I can refactor it to be `Overrides`. Worth noting that the `DestinationOverride` is needed here if the FE wanted to be able to properly display the destination of the alert.

internal/core/alert_delivery/api/outputs.go
wey-chiang: Worth noting that this conditional won't respect the destination suppression (i.e. passing in an empty list).
Fixed!

internal/log_analysis/rules_engine/tests/test_main.py
wey-chiang: Added a unit test `test_direct_analysis_FIELD_exception_fails_test` for all of the generated fields

wey-chiang: Will create a `test_direct_analysis_event_matching_with_generated_fields` that includes output for all of the generated fields.

PR #2108, opened 2020-11-25
Update supported regions (and publish all of them)

austinbyers:
## Background

When deploying from source, Panther supports [many regions](https://docs.runpanther.io/development#supported-regions). But we only publish 3 of them today - this PR updates the supported regions and makes sure we fully support and publish all of them

Closes: https://github.com/panther-labs/aws-management-cloudformation/issues/397

## Changes

- Add 4 new regions which Panther will now work in (thanks to AWS adding AppSync/Cognito/Glue/Athena to these regions)
- Update `mage master:publish` to publish *all* supported regions (not just the US regions)
- Update `mage master:publish` so that each region is published in parallel
    - `docker build` and `docker push` have been split into separate Go helper functions: we only need to build once, but we push to each region

## Testing

- `mage deploy` (make sure the docker build/push flow still works)
- `mage master:deploy` (which shares a lot of code with `master:publish`)
- `mage master:publish` - I didn't actually publish versions in each region, but I tested the parallel packaging, which worked pretty well!


kostaspap (approved):
lgtm

austinbyers:
I deployed the codepipeline stack successfully, it looks like the pipeline is configured correctly.

However, there was an existing problem with the `identity-us-west-2` master stack:

> Output 'GeorgeSkoumasUserName' not found in stack 'arn:aws:cloudformation:us-west-2:948584460855:stack/identity-us-west-2-IamUsers-1L9ZE8BC1UV1T/18933a20-a48c-11ea-8b4d-02bc0b348568'

I've fixed that in this PR as well. Once this merges, I'll monitor the code pipeline and make sure it worked

PR #2112, opened 2020-11-25
Add parsers for Duo logs

giorgosp:
## Background

Add parsers for Duo logs.
https://duo.com/docs/adminapi#logs

## Changes

- Add parsers for Duo authentication (v2), administrator, telephony logs.

## Testing

-  Unit

### DIFF
--- a/internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
+++ b/internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
@@ -174,7 +174,7 @@ func TestResult(t *testing.T, expect string, actual *pantherlog.Result, indicato
 // EqualTimestamp is a helper that checks timestamps for equality with human readable message
 func EqualTimestamp(t *testing.T, expect, actual time.Time, msgAndArgs ...interface{}) {
 	t.Helper()
-	require.False(t, actual.IsZero(), "zero timestamp")
+	require.NotZero(t, actual, msgAndArgs...)
 	require.Equal(t, expect.UTC().Format(time.RFC3339Nano), actual.UTC().Format(time.RFC3339Nano), msgAndArgs...)
 }
 
--- a/internal/log_analysis/log_processor/parsers/duologs/administrator.go
+++ b/internal/log_analysis/log_processor/parsers/duologs/administrator.go
@@ -1,0 +1,36 @@
+package duologs
+
+/**
+ * Panther is a Cloud-Native SIEM for the Modern Security Team.
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import "github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog"
+
+const TypeAdministrator = "Duo.Administrator"
+
+// nolint:lll
+type AdministratorLog struct {
+	//TODO: This event has the same fields as OfflineEnrollmentLog. To distinguish them, we validate that Action below
+	// doesn't contain the values that OfflineEnrollmentLog.Action can take. If, however, Duo adds a value to the
+	// values OfflineEnrollmentLog.Action can take, these events will be misclassified until we update our schemas.
+	Action       pantherlog.String `json:"action" panther:"required" validate:"excludes=o2fa_user_provisioned o2fa_user_deprovisioned o2fa_user_reenrolled" description:"The type of change that was performed."`
+	Description  pantherlog.String `json:"description" description:"String detailing what changed, either as free-form text or serialized JSON."`
+	ISOTimestamp pantherlog.Time   `json:"isotimestamp" panther:"required" event_time:"true" tcodec:"rfc3339" description:"ISO8601 timestamp of the event."`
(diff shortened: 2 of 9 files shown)


internal/log_analysis/log_processor/parsers/duologs/authentication.go
```diff
@@ -0,0 +1,85 @@
...
+
+// nolint:lll
+type AccessDevice struct {
+	Browser             pantherlog.String   `json:"browser" description:"The web browser used for access."`
+	BrowserVersion      pantherlog.String   `json:"browser_version" description:"The browser version."`
+	FlashVersion        pantherlog.String   `json:"flash_version" description:"The Flash plugin version used, if present, otherwise \"uninstalled\"."`
+	Hostname            pantherlog.String   `json:"hostname" panther:"hostname" description:"The hostname, if present, otherwise \"null\"."`
+	IP                  pantherlog.String   `json:"ip" panther:"ip" description:"The access device's IP address, if present, otherwise \"null\"."`
+	IsEncryptionEnabled pantherlog.Bool     `json:"is_encryption_enabled" description:"Reports the disk encryption state as detected by the Duo Device Health app. One of \"true\", \"false\", or \"unknown\"."`
+	IsFirewallEnabled   pantherlog.Bool     `json:"is_firewall_enabled" description:"Reports the firewall state as detected by the Duo Device Health app. One of \"true\", \"false\", or \"unknown\"."`
```
kostaspap, line 51: Bool will not work in the case where it is `unknown`. Should this be string?

internal/log_analysis/log_processor/parsers/duologs/authentication.go
```diff
@@ -0,0 +1,85 @@
...
+	IsPasswordSet       pantherlog.Bool     `json:"is_password_set" description:"Reports the system password state as detected by the Duo Device Health app. One of \"true\", \"false\", or \"unknown\"."`
```
kostaspap, line 52: See above

kostaspap (approved):

giorgosp:
Added parser for Duo offline enrollment logs. It has an interesting caveat since its fields are the same as the administrator log.

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
```diff
@@ -174,7 +174,6 @@ func TestResult(t *testing.T, expect string, actual *pantherlog.Result, indicato
 // EqualTimestamp is a helper that checks timestamps for equality with human readable message
 func EqualTimestamp(t *testing.T, expect, actual time.Time, msgAndArgs ...interface{}) {
 	t.Helper()
-	require.False(t, actual.IsZero(), "zero timestamp")
```
kostaspap, line 177: Why was this removed?

internal/log_analysis/log_processor/parsers/duologs/authentication.go
kostaspap: I have contacted Duo support to clarify the content of these logs. 
You should add a FIXME on these fields and we should update once Duo replies.

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
alxarch: Yup, this is critical to have.

internal/log_analysis/log_processor/parsers/duologs/administrator.go
```diff
@@ -0,0 +1,36 @@
...
+
+const TypeAdministrator = "Duo.Administrator"
+
+// nolint:lll
+type AdministratorLog struct {
+	//TODO: This event has the same fields as OfflineEnrollmentLog. To distinguish them, we validate that Action below
+	// doesn't contain the values that OfflineEnrollmentLog.Action can take. If, however, Duo adds a value to the
+	// values OfflineEnrollmentLog.Action can take, these events will be misclassified until we update our schemas.
+	Action       pantherlog.String     `json:"action" panther:"required" validate:"excludes=o2fa_user_provisioned o2fa_user_deprovisioned o2fa_user_reenrolled" description:"The type of change that was performed."`
+	Description  pantherlog.RawMessage `json:"description" description:"String detailing what changed, either as free-form text or serialized JSON."`
```
alxarch, line 31: The description tag is not a RawMessage. It is a `pantherlog.String`. It states so in the description. When it has JSON form, it is _serialized JSON_. Which means that `RawMessage` not useful here.
The case of 'embedded' JSON is something we do not handle well now. In this particular case where it could also be 'free-form text' it is even harder to solve. I suggest we just use `pantherlog.String` for now.

alxarch (changes requested):
Need to change the `Administrator.Description` to `pantherlog.String`

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
giorgosp: It produces a cryptic error message. Reverting for now to merge this, but we should improve it.

internal/log_analysis/log_processor/parsers/duologs/administrator.go
giorgosp: Updated

internal/log_analysis/log_processor/parsers/duologs/authentication.go
giorgosp: Done

internal/log_analysis/log_processor/parsers/duologs/authentication.go
giorgosp: Done

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
kostaspap: @giorgosp  update the error message?

giorgosp: Updated. Test error is now:
```
Error:      	Should not be zero, but was 0001-01-01 00:00:00 +0000 UTC
        	Test:       	TestDuoParsers/authentication_logs
        	Messages:   	invalid event time
```

was
```
Error:      	Should be false
        	Test:       	TestDuoParsers/authentication_logs
        	Messages:   	zero timestamp
```

alxarch: GoodEnough

alxarch (approved):
LGTM

Issue #1321, opened 2020-08-07
p_any_usernames and p_any_emails


jacknagz:
### Describe the ideal solution

New standard fields for usernames and email addresses to make writing rules quicker, primarily for SaaS logs.

### References

Some sample mappings below (not comprehensive):

GSuite.Reports => actor['email']
AWS.CloudTrail => userIdentity['userName']
OneLogin.Events => user_name
GitLab.Production and API => username

### Additional context

Let's prioritize the SaaS-based logs, then move on to the Cloud/Host logs.

kostaspap:
How do we treat the scenario where a username is also an email? Do we add it in both?

alxarch:
> How do we treat the scenario where a username is also an email? Do we add it in both? @jacknagz @rleighton

Since we allow the schemas to specify multiple indicator scanners this should be on a per-logtype basis by adding *both* scanners to the string.

rleighton:
I would add to both

jacknagz:
I would actually say that we should adhere to strict formats for standard fields, meaning we would not put emails as usernames. Similarly, a hostname should not be classified as a domain.

rleighton:
Except they can legit be both

alxarch:
As a general approach I would avoid a strict check for a value conforming email format. The purpose of the log processor is not to validate the log output. We could end up failing to read data because of a bug in external software or misconfiguration of a server. It is preferable to do minimal checking and such 'mistakes' would end up in query results, potentially helping with detecting abnormal behavior.

If the user marks a field as `panther:"email"` for example, it is not our job to throw an error if the field is not a valid email address. We can still have some sort of fast check (i.e. `contains('@')`) to avoid adding completely invalid input, but I would avoid using a complex regular expression or similar slow processing to check an email value. IMO an invalid email showing up in rules,queries,indicator searches woulg **help** users detect abnormal behavior. Users would be blind to this if we just drop the values.

kostaspap:
> As a general approach I would avoid a strict check for a value conforming email format. The purpose of the log processor is not to validate the log output. We could end up failing to read data because of a bug in external software or misconfiguration of a server. It is preferable to do minimal checking and such 'mistakes' would end up in query results, potentially helping with detecting abnormal behavior.
> 
[...]

PR #2117, opened 2020-11-26
Add `username` and `email` indicators to `pantherlog`

alxarch:
## Background

This PR is related to #1321

## Changes

- Adds `username` indicator scanner and field
- Adds `email` indicator scanner and field
- Adds `username` tag to `AWS.CloudTrail`

## Testing

- mage test:go

### DIFF
--- a/internal/log_analysis/log_processor/pantherlog/meta.go
+++ b/internal/log_analysis/log_processor/pantherlog/meta.go
@@ -76,7 +76,7 @@ const (
 
 // ScanValues implements ValueScanner interface
 func (id FieldID) ScanValues(w ValueWriter, input string) {
-	w.WriteValues(id, input)
+	w.WriteValues(id, strings.TrimSpace(input))
 }
 
 // CoreFields are the 'core' fields Panther adds to each log.
--- a/internal/log_analysis/log_processor/pantherlog/meta.go
+++ b/internal/log_analysis/log_processor/pantherlog/meta.go
@@ -71,6 +71,7 @@ const (
 	FieldAWSARN
 	FieldAWSTag
 	FieldEmail
+	FieldUsername
 )
 
 // ScanValues implements ValueScanner interface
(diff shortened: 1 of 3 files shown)


internal/log_analysis/log_processor/pantherlog/meta.go
```diff
@@ -196,15 +198,25 @@ func init() {
 		NameJSON:    "p_any_aws_tags",
 		Description: "Panther added field with collection of AWS Tags associated with the row",
 	})
-	MustRegisterScanner("ip", ValueScannerFunc(ScanIPAddress), FieldIPAddress)
+	MustRegisterIndicator(FieldEmail, FieldMeta{
```
kostaspap, line 201: Tests?

alxarch: The underlying mechanism for scanning with registered scanners is covered in tests.
The specific indicator scanners are tested in the respective tests for the parsers that use them.
The coverage is pretty thorough if you run the full test suite.

kostaspap: I don't think it is safe to rely on parser tests for that. The parser tests will cover this code to an extend, but their purpose is not that. 

For example, the way this code works: 
1. If a field doesn't have the `@` character, it will not add the indicator
2. If the field has extra spaces e.g. `"test@example.com "`, it will trim the space.

I highly doubt that parser tests will test those scenarios - we could not expect them to do so. This functionality is key part of the scanner though and needs to be tested somehow.

alxarch: > I highly doubt that parser tests will test those scenarios

The YAML tests that check actual JSON results do catch these things. But I will add unit tests anyway.

alxarch: I was mistaken, you meant to use unit tests as booby traps to check for proper empty values on the actual Scanner functions. I thought you meant to check the struct tag mapping for each field.

alxarch: Fixed

kostaspap (approved):

PR #2118, opened 2020-11-26
Fix rule error

### LINKED ISSUE #2113: Syntax Errors Not Showing as Rule Errors
### Describe the bug

When a rule has a syntax error, it does not generate a Rule Error

### Steps to reproduce

1. Write a new rule
2. Add a return statement such as `return foobar = 'test'`
3. CloudWatch alarms fire, but no rule error alert is generated:

```
[ERROR] 2020-11-25T20:26:22.625Z 37ebe67d-1e18-580d-913a-be45de030a8f Failed to import rule Rule.ID Error: [invalid syntax (Rule.ID.py, line 2)]
```

### Expected behavior

A rule error alert should be generated about the Syntax Error

### Environment

Panther Enterprise v1.13.0-RC

### Additional context

None

kostaspap:
## Background

Closes #2113 

## Changes

- `rule_exception` now contains exceptions raised while importing the rule
- Removed the "generic" error - import or "rule function doesn't exist" errors are now handled like generic "rule" errors. Let me know what you think @giorgosp 

## Testing

- unit test
- Tested E2E

Error displayed for invalid module:
Screen Shot 2020-11-30 at 4 00 19 PM

Error displayed for rule non-existent:
Screen Shot 2020-11-30 at 4 06 29 PM

Notification:
Screen Shot 2020-11-30 at 4 08 04 PM

Event details: 

Screen Shot 2020-11-30 at 4 07 54 PM


jacknagz:
Thanks for the quick fix!!

alxarch:
@kostaspap Why did you opt to not differentiate the '_setup_error' from the RuleError?

kostaspap:
> @kostaspap Why did you opt to not differentiate the '_setup_error' from the RuleError?

I was a bit on the fence for this. 

The way I saw it, we have two different types of errors that can happen while running a rule: 
1. Errors that we can recover from. E.g. if any of the title/dedup/alert_context functions throw an error at runtime, we just handle them internally and return the default value for each one of those. 
2. Errors that we cannot recover from: 
  i) Errors importing the rule
  ii) Rule doesn't have a 'rule' method
  iii) Rule method throws an error or doesn't return valid return value

If any of the errors in No2 above happen, we want to store it to the datalake and generate an alert. In my mind all of them somehow are "rule" errors. 

Another way would be to have a `setup`/`generic` error for 2.i & 2.ii and keep the rule error for 2.iii . In case any of the 2.i or 2.ii appear, we send them to the datalake, if 2.iii happens we send that one. 
But I am so-and-so with the approach, wanted some opinions on this

api/lambda/analysis/engine.go
```diff
@@ -97,8 +97,6 @@ type RuleResult struct {
 	DedupError         string `json:"dedupError"`
 	AlertContextOutput string `json:"alertContextOutput"`
 	AlertContextError  string `json:"alertContextError"`
-	// Indicates general error in the Python script (import error, syntax error, etc).
-	GenericError string `json:"genericError"`
```
giorgosp, line 101: `GenericError` was meant to contain such errors. Why remove it?
`RuleError`  is mean't to contain the error only if it happens inside the `rule()` function

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -127,10 +127,15 @@ def __init__(self, config: Mapping):
             self.rule_reports = config['reports']
 
         self._store_rule()
-        self._module = self._import_rule_as_module()
 
-        if not hasattr(self._module, 'rule'):
-            raise AssertionError("rule needs to have a method named 'rule'")
```
giorgosp, line 133: This should probably be a `GenericError`. Maybe it was before? Not sure

api/lambda/analysis/engine.go
kostaspap: Updated

internal/log_analysis/rules_engine/src/rule.py
kostaspap: Updated

giorgosp (approved):
LGTM!

PR #2120, opened 2020-11-26
Port GitLab parsers to pantherlog and add username tags

alxarch:
## Background

## Changes

## Testing


internal/log_analysis/log_processor/parsers/gitlablogs/gitlablogs.go
```diff
@@ -47,47 +49,61 @@ func LogTypes() logtypes.Group {
...
+		NewEvent: func() interface{} {
+			return &Production{}
+		},
+		// This custom validation checks that the event is gitlab.Production not gitlab.API
+		Validate: func(x interface{}) error {
+			event := x.(*Production)
+			if event.Action.Exists || event.EtagRoute.Exists {
+				return pantherlog.ValidateStruct(x)
+			}
+			return errors.New("log entry is probably GitLab.API not GitLab.Production")
```
kostaspap, line 106: This feels a bit like a hack. We use the validator package for validating events. Using the validator's package `require_without` tag seems more appropriate then.

internal/log_analysis/log_processor/parsers/gitlablogs/testdata/api_tests.yml
```diff
@@ -0,0 +1,80 @@
...
+    "time":"2018-10-29T12:49:42.123Z",
+    "severity":"INFO",
+    "duration_s":709.08,
+    "db_duration_s":14.59,
+    "view_duration_s":694.49,
+    "status":200,
+    "method":"GET",
+    "path":"/api/v4/projects",
+    "params":[{"key":"action","value":"git-upload-pack"},{"key":"changes","value":"_any"},{"key":"key_id","value":"secret"},{"key":"secret_token","value":"[FILTERED]"}],
+    "host":"localhost",
```
kostaspap, line 30: This should appear in a "p_any_hostnames" field in the output. It doesn't seem to do so?

internal/log_analysis/log_processor/parsers/gitlablogs/gitlablogs.go
alxarch: It doesn't work though. There seems to be a bug with custom validators.
It worked for `*string` but does not work for `null.String`.

> This feels a bit like a hack. 

Actually it's the other way around :). The `validator` package is a complete hack, held together by duct tape if you take a look under the hood. 
`v10` slightly improves that but it has incompatibilities which I am not willing to resolve in the scope of this PR (might break things elsewhere). Using both is also not an option without duplicating all code for registering custom type validators for `null`.

BTW, the `Validate` field is specifically chosen to be a closure to be able to intervene and manage problematic cases such as this.

kostaspap: What I don't like is that we have validation logic laying around the place. Until now, looking at the schema definitions (Go structs defining the struct) was sufficient to identify what fields are needed and which not. This has been useful when trying to debug issues. Now some logic lays in the tags, some in here. It's easy for someone reviewing the schema to not notice this logic here. 

I agree upgrading the validator is out of scope here, so better add a TODO / FIXME here to explain at least why this logic needed and why the validator package that we use is not sufficient for this case.

internal/log_analysis/log_processor/parsers/gitlablogs/testdata/api_tests.yml
alxarch: We do not have `p_any_hostnames` indicator field. The `hostname` indicator **scanner** checks a value and puts it either in the `p_any_ip_addresses` or `p_any_domain` fields. This is intentional, as hostname values are interchangeable with IP addresses in configurations of services. (note that `localhost` *is* a domain name, just an unqualified one, it just so happens that it is 'hardwired' to resolve to `127.0.0.1` in most systems)

kostaspap: Right right, too many PRs at once, got confused a bit.

internal/log_analysis/log_processor/parsers/gitlablogs/api.go
```diff
@@ -19,81 +19,39 @@ package gitlablogs
...
+	Method                pantherlog.String  `json:"method" validate:"required" description:"The HTTP method of the request"`
+	Path                  pantherlog.String  `json:"path" validate:"required" description:"The URL path for the request"`
+	Params                []QueryParam       `json:"params" description:"The URL query parameters"`
+	Host                  pantherlog.String  `json:"host" panther:"hostname" validate:"required" description:"Hostname serving the request"`
+	UserAgent             pantherlog.String  `json:"ua" description:"User-Agent HTTP header"`
+	Route                 pantherlog.String  `json:"route" validate:"required" description:"Rails route for the API endpoint"`
+	RemoteIP              pantherlog.String  `json:"remote_ip" panther:"ip" description:"The remote IP address of the HTTP request"`
+	UserID                pantherlog.Int64   `json:"user_id" description:"The user id of the request"`
+	UserName              pantherlog.String  `json:"username" panther:"username" description:"The username of the request"`
+	GitalyCalls           pantherlog.Int32   `json:"gitaly_calls" description:"Total number of calls made to Gitaly"`
```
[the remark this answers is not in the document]
alxarch: fixed

kostaspap (approved):

PR #2128, opened 2020-11-27
FE: Fix issues with running web tests locally

3nvi:
## Background

Web tests are timezone-aware and they fail when they don't run in a UTC-based environment since the timezone offset that the local environment is in affects the tests.

Unfortunately, this needed to be setup in a level much higher than our testing suite, since when jest (our test runner) initializes, we already need to have the correct timezone.

This PR changes jest's configuration by making sure to feed it with the correct ENV vars before any test suite runs. This will eventually allow `mage test:web` to run confidently **everywhere**.

## Changes

- add `globalSetup` hook for jest
- rename test setup files to match their purpose
- fix flaky test by removing timeouts and replacing them with `await` calls

## Testing

- `mage test:ci`


vorillaz (approved):
🚢  it

Issue #2031, opened 2020-11-16
Cloud Security events get processed by panther-log-processor


kostaspap:
### Description

Cloud Security resources are stored in the Data lake and queryable through Athena

### Related Services

panther-log-processor

### Designs

None. This is a backend task

### Acceptance Criteria

- Cloud security resources are queryable in the `panther_cloudsecurity` database.

PR #2137, opened 2020-11-27
Refactoring Cloud Security DB creation

kostaspap:
## Background

Work is related to https://github.com/panther-labs/panther/issues/2031
Did some refactoring in the DB management logic. 
Also allowing Log Processor to write data to cloudsecurity database.

## Changes

- Moved all the DB name constants to a new package, `pantherdb` (like `pantherlog`). 
We needed a central place to keep Datalake information that is not restricted only to AWS glue
- Removed some unused methods (see comments) 
- Some smaller refactorings

## Testing

- mage test:ci
- Deployed over existing deployment in dev account. I was able to create a new source and seen logs populated in Athena
- Deployed fresh. Seen all DBs (`panther_logs`, `panther_views`, `panther_temp`, `panther_cloudsecurity`) created properly. Teared it down and seem them being deleted. 
- Pending extra unit tests on the s3 destination


internal/log_analysis/awsglue/partition.go
```diff
@@ -67,23 +66,14 @@ func (gp *GluePartition) GetGlueTableMetadata() *GlueTableMetadata {
...
-	return getTablePrefix(datatype, GetTableName(logType)) + timebin.PartitionPathS3(time)
+func GetPartitionPrefix(database, logType string, timebin GlueTableTimebin, time time.Time) string {
+	return GetTablePrefix(database, pantherdb.GetTable(logType)) + timebin.PartitionPathS3(time)
 }
 
 func (gp *GluePartition) GetPartitionLocation() string {
 	return "s3://" + gp.s3Bucket + "/" + gp.gm.GetPartitionPrefix(gp.time)
 }
 
-// GetPartitionLocation takes an S3 path for an object and returns just the part of the patch associated with the partition
```
kostaspap, line 78: This was not used

internal/log_analysis/awsglue/partition.go
```diff
@@ -169,19 +158,11 @@ func GetPartitionFromS3(s3Bucket, s3ObjectKey string) (*GluePartition, error) {
 	}
 	partition.time = time.Date(year, time.Month(month), day, hour, 0, 0, 0, time.UTC)
 
-	partition.gm = NewGlueTableMetadata(partition.datatype, partition.tableName, "", GlueTableHourly, nil)
+	partition.gm = NewGlueTableMetadata(partition.databaseName, partition.tableName, "", GlueTableHourly, nil)
 
 	return partition, nil
 }
 
-func GetPartitionFromS3Path(s3Path string) (*GluePartition, error) {
```
kostaspap, line 177: This was not used

internal/log_analysis/awsglue/table_metadata.go
```diff
@@ -46,7 +43,6 @@ type PartitionKey struct {
 
 // Metadata about Glue table
 type GlueTableMetadata struct {
-	dataType     models.DataType
```
kostaspap, line 49: Seemed that DB name was sufficient indicator - seemed simpler just to remove the DataType and let Glue deal with databases and table names

internal/log_analysis/awsglue/partition.go
```diff
@@ -67,23 +66,14 @@ func (gp *GluePartition) GetGlueTableMetadata() *GlueTableMetadata {
 	return gp.gm
 }
 
-func GetPartitionPrefix(datatype models.DataType, logType string, timebin GlueTableTimebin, time time.Time) string {
```
kostaspap, line 70: It seemed appropriate that the "glue" methods are operating with databases, tables instead of logtypes and datatypes. I think it clears things up a bit.

internal/log_analysis/awsglue/table_metadata.go
```diff
@@ -46,29 +43,24 @@ type PartitionKey struct {
...
 	tableName    string
 	description  string
-	logType      string
 	prefix       string
 	timebin      GlueTableTimebin // at what time resolution is this table partitioned
 	eventStruct  interface{}
 }
 
 // Creates a new GlueTableMetadata object for Panther log sources
 func NewGlueTableMetadata(
```
kostaspap, line 55: It seemed appropriate that AWS Glue operates with databases and table names. I think it simplifies things a bit

internal/log_analysis/awsglue/table_metadata.go
```diff
@@ -205,18 +191,6 @@ func (gm *GlueTableMetadata) glueTableInput(bucketName string) *glue.TableInput
 	}
 }
 
-func (gm *GlueTableMetadata) Signature() (string, error) {
```
kostaspap, line 208: This was not used

internal/core/custom_resources/resources/log_processor_tables.go
```diff
@@ -71,14 +72,20 @@ func customUpdateLogProcessorTables(ctx context.Context, event cfn.Event) (strin
 		logger.Info("started database sync", zap.Strings("logTypes", requiredLogTypes))
 		return physicalResourceID, nil, nil
 	case cfn.RequestDelete:
-		for pantherDatabase := range awsglue.PantherDatabases {
-			logger.Info("deleting database", zap.String("database", pantherDatabase))
-			if _, err := awsglue.DeleteDatabase(glueClient, pantherDatabase); err != nil {
+		logDatabases := []string{
```
giorgosp, line 75: nit: Why not keep it as a constant?

giorgosp (approved):

internal/core/custom_resources/resources/log_processor_tables.go
kostaspap: It's a slice so it can't be a constant only variable... You mean make a globally accessible variable in the pantherdb?

giorgosp: Yes, didn't mean it literally :P

kostaspap: Updated, I think it's more clear now

internal/log_analysis/awsglue/partition.go
rleighton: FYI used in enterprise here: https://github.com/panther-labs/panther-enterprise/blob/master/internal/log_analysis/datacatalog_compactor/process/compaction.go#L184

internal/log_analysis/awsglue/partition.go
rleighton: FYI used in enterprise here:  https://github.com/panther-labs/panther-enterprise/blob/master/internal/log_analysis/datacatalog_compactor/process/latedata.go#L59

rleighton (approved):

internal/core/pantherdb/pantherdb.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Returns the name of the table for the given log type
+func GetTable(logType string) string {
+	// clean table name to make sql friendly
+	tableName := strings.Replace(logType, ".", "_", -1) // no '.'
+	return strings.ToLower(tableName)
+}
+
+// Returns the database in which exists the
+func GetDatabase(logtype string) string {
```
alxarch, line 85: We're duplicating the logic from `GetDataType` here. Is there a possible scenario in which the two might diverge in the future? I would prefer a `DatabaseName(typ DataType) string` method with a `switch` so that we can start 'binding' `DataType` with db names.

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -143,13 +143,12 @@ func (api *API) getEventsForLogType(
...
-		var dataType logprocessormodels.DataType
+		database := pantherdb.RuleMatchDatabase
 		if alert.Type == alertmodels.RuleErrorType {
-			dataType = logprocessormodels.RuleErrors
-		} else {
-			dataType = logprocessormodels.RuleData
+			database = pantherdb.RuleErrorsDatabase
 		}
-		partitionPrefix := awsglue.GetPartitionPrefix(dataType, logType, awsglue.GlueTableHourly, nextTime)
+		tableName := pantherdb.GetTable(logType)
```
alxarch, line 150: nitpick: It is more usual to avoid `Get` prefixes in go code. `pantherdb.TableName(logType string) string` would be more aligned with ['idiomatic' go](https://golang.org/doc/effective_go.html#Getters). This PR seems to be ideal to get rid of all the 'GetFoo` function names in this package.

internal/log_analysis/awsglue/table_metadata.go
alxarch: In general AFAIK DataType == Database.

internal/log_analysis/awsglue/table_metadata.go
alxarch: Apart from `timebin` stuff, this is practically the same as fields in `glue.TableData`.
If/when we base more stuff on `logschema` to avoid the `eventStruct` this struct will be just a bunch of 'simple' data and code will clean up furtger.

internal/log_analysis/awsglue/table_metadata.go
```diff
@@ -128,19 +112,19 @@ func (gm *GlueTableMetadata) PartitionKeys() (partitions []PartitionKey) {
 }
 
 func (gm *GlueTableMetadata) RuleTable() *GlueTableMetadata {
```
alxarch, line 114: All this is too many methods. Why are fields in this struct private anyway? It just adds methods.
`Prefix()` could be computed when called (it is only used in a single place anyway).
I won't request this to be done here, but this package needs a good scrubbing.

alxarch (approved):
Minor remarks on naming and possible further reduction of the code in `awsglue` package.

internal/log_analysis/alerts_api/api/get_alert.go
kostaspap: Good callout. Will do

internal/log_analysis/awsglue/partition.go
kostaspap: Added back

internal/log_analysis/awsglue/partition.go
kostaspap: This method seems that is is just wrapping `GetPartitionFromS3Path()` and doing `GetPartitionLocation` on the result. I thinking to remove it anyway and just have the caller call the appropriate method. I am worried because this package has a lot of helper functions like this that are really thin wrappers around other public functions...Trying to trim this package down a bit

cmd/opstools/gluerecover/gluerecover/main.go
```diff
@@ -31,7 +31,7 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/panther-labs/panther/cmd/opstools"
-	"github.com/panther-labs/panther/internal/log_analysis/awsglue"
+	"github.com/panther-labs/panther/internal/core/pantherdb"
```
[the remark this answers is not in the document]
kostaspap: I was so and so with this. I though thought we said to put it in the `core`. I am fine putting it elsewhere. Log Analysis might be more appropriate in the sense that it deals more with "all thigns datalake". Moving

internal/core/pantherdb/pantherdb.go
kostaspap: Good idea. Updated

Issue #1384, opened 2020-08-18
Redesign: Source Card listing


3nvi:
### Description

Redesign listing of sources in order to ditch the table UI and replace it by cards, as shown here:

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/849670d3-62c0-462f-b399-bf6174c4352b/commits/d338e01766263f7422c0f695c2f40130c92a8cfc/files/c2bbc1a8-4778-4cd6-93a5-4630d8868201/layers/59865C50-D963-424B-A437-AA149073111B?collectionId=eddb711d-3691-4e4f-ba4d-6eb04ff9a773&collectionLayerId=a52865f6-7359-4d14-a5f8-7308423dff48&mode=build&selected=2656500353-20235FB5-84DC-4DDD-9A58-DCCF86F5753E

### Acceptance Criteria

- Table is not used for the listing of  sources
- Each source gets  represented  by a single card
- Each card has a logo of the source  along with all the details that need to be visible for this source

PR #2140, opened 2020-11-28
FE: Add anonymous analytics to sources

3nvi:
## Background

This PR calibrates sources by increasing the loading times during edit & adding analytics to all create & update events.

Relates to #1384 

## Changes

- Add analytics to S3, SQS & Compliance sources, both during create & update, both for success & failure
- Make sure to preload the ID field from URL params rather than waiting for api callback (helps with `const existing ...` variable that's defined inside wizards)
- Increase left margin of options icon in Cards

## Testing

- `npm run test`


alexmylonas (approved):
LGTM! Add `sync-to-release` if you want it for v1.14

PR #2142, opened 2020-11-28
Validate instance IDs exist before sending

nhakmiller:
## Background

The `aws-event-processor` processes AWS CloudTrail logs and determines if scans need to happen as a result of these logs. Part of the responsibility of the `aws-event-processor` is to extract the `resourceID` (typically the ARN) of the resource being changed and include that in the scan request.

For certain AWS RDS instance related events, the instance ID may not be present in the log if the instance no longer exists.  This is common, for example, when changes happen to RDS instance snapshots of instances that no longer exist.

We were previously passing along these "blank" arns that were causing issues in the `panther-snapshot-poller`.  This change extends some validation of the instance ID extraction to all branches when evaluating RDS Instance related events.

Discovered while investigating on call incidents.

## Changes

- Always validate that the RDS instance ID we extracted exists before returning a scan request

## Testing

- None

### DIFF
--- a/internal/compliance/aws_event_processor/processor/rds.go
+++ b/internal/compliance/aws_event_processor/processor/rds.go
@@ -52,7 +52,12 @@ func classifyRDS(detail gjson.Result, metadata *CloudTrailMetadata) []*resourceC
 	case "AddRoleToDBInstance", "CreateDBInstance", "CreateDBSnapshot", "DeleteDBInstance", "ModifyDBInstance",
 		"PromoteReadReplica", "RebootDBInstance", "RemoveRoleFromDBInstance", "RestoreDBInstanceFromDBSnapshot",
 		"RestoreDBInstanceFromS3", "StartDBInstance", "StopDBInstance":
-		rdsARN.Resource += detail.Get("requestParameters.dBInstanceIdentifier").Str
+		instanceID := detail.Get("requestParameters.dBInstanceIdentifier")
+		if !instanceID.Exists() {
+			zap.L().Warn("unable to extract dBInstanceIdentifier from event", zap.Any("requestParameters", detail.Get("requestParameters").Raw))
+			return nil
+		}
+		rdsARN.Resource += instanceID.Str
 	case "AddTagsToResource", "RemoveTagsFromResource":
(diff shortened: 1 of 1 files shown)


internal/compliance/aws_event_processor/processor/rds.go
```diff
@@ -52,7 +52,12 @@ func classifyRDS(detail gjson.Result, metadata *CloudTrailMetadata) []*resourceC
 	case "AddRoleToDBInstance", "CreateDBInstance", "CreateDBSnapshot", "DeleteDBInstance", "ModifyDBInstance",
 		"PromoteReadReplica", "RebootDBInstance", "RemoveRoleFromDBInstance", "RestoreDBInstanceFromDBSnapshot",
 		"RestoreDBInstanceFromS3", "StartDBInstance", "StopDBInstance":
-		rdsARN.Resource += detail.Get("requestParameters.dBInstanceIdentifier").Str
+		instanceID := detail.Get("requestParameters.dBInstanceIdentifier")
+		if !instanceID.Exists() {
+			zap.L().Warn("unable to extract dBInstanceIdentifier from event", zap.Any("requestParameters", detail.Get("requestParameters").Raw))
```
austinbyers, line 57: If this is expected for instances which do not exist, can this (and the others) just be an "info" message? Should the on-call be paged for lots of warnings here?

austinbyers (approved):

internal/compliance/aws_event_processor/processor/rds.go
nhakmiller: That's a good question, I wasn't sure on this. My thought process was that we've only seen this extremely rarely in practice, so if all of a sudden it starts happening a lot that might be unusual and could mean some assumption we are making that was previously true is no longer true. But an `Info` would be more appropriate since we do expect this to happen sometimes, although rarely.

I think I will swap over to `Info` messages for now, if an issue bubbles up from this then at least we'll have the logs to troubleshoot it but since this could be common if e.g. an instance gets deleted then all 200 of it's snapshots get deleted a week later, there's no reason to page on call staff for that.

lindsey-w (approved):

PR #2143, opened 2020-11-29
FE: Refactor Related Destinations

3nvi:
## Background

**This PR is an optional enhancement that's not planned or required.  Feel free to reject**

Related destinations showcases a `+1` even when it shouldn't do so. In the screenshot below, this has caused 2 issues:

1. It has hidden another destination (of the same type) which **clearly** had the space to be rendered
2. It has shown a `+1` without any information to what this +1 is

Screen Shot 2020-11-28 at 6 10 00 PM

This PR changes that and makes `RelatedDestinations` use the `<LimitItemDisplay>` component

## Changes

- Don't hide destinations (even if they are of the same type) if there's enough space
- Render a tooltip via `LimitItemDisplay` for all additional destinations
- Defer typography  to parent container

## Screenshots

Screen Shot 2020-11-29 at 12 42 58 PM

## Testing

-`npm run test`


kostaspap:
The screenshot shows `JIRA icon, +1, Slack icon`. 

I imagine there are 3 destinations for this alert and we are showing two and we are hiding one? Wouldn't be preferable to show `JIRA icon, Slack icon, +1` ? Maybe sounds like naive question, but that's how I perceive this design.

3nvi:
> The screenshot shows `JIRA icon, +1, Slack icon`.
> 
> I imagine there are 3 destinations for this alert and we are showing two and we are hiding one? Wouldn't be preferable to show `JIRA icon, Slack icon, +1` ? Maybe it is a naive question, but that's how I perceive this design.

The screenshot is just there it to showcase the solution (since all the alerts in my dev env have 2 destinations). We normally allow 3 destinations visible before we show a `+XXX` more

web/src/components/RelatedDestinations/RelatedDestinations.tsx
```diff
@@ -54,58 +52,45 @@ const RelatedDestinations: React.FC<RelatedDestinationsSectionProps> = ({
   verbose = false,
   limit = 3,
 }) => {
+  const sortedDestinations = React.useMemo(() => sortBy(destinations, d => d.outputType), [
```
vorillaz, line 55: It's unclear what the sorting does here, can we use native sort instead (returning 0, 1, -1) with [`String.prototype.localeCompare()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) in place?

web/src/components/RelatedDestinations/RelatedDestinations.tsx
```diff
@@ -54,58 +52,45 @@ const RelatedDestinations: React.FC<RelatedDestinationsSectionProps> = ({
...
-  }
-
   return (
     <Flex align="center" minWidth={85} spacing={2}>
-      {destinations.length ? (
-        destinations.map(getLogo)
-      ) : (
-        <Text fontSize="small">Not configured</Text>
-      )}
+      <LimitItemDisplay limit={limit}>{sortedDestinations.map(getLogo)}</LimitItemDisplay>
```
vorillaz, line 91: `limit` is not actually used by any call sites yet, shall we keep it as a prop?

3nvi: it was not used because it had a default value of `3`. It already is a component prop though, so I'm missing what you are referring to

web/src/components/RelatedDestinations/RelatedDestinations.tsx
3nvi: The whole point of lodash is to make things simpler, so I'd really like to "save" a few lines of code here since `sortBy` is used all over the app.

Would `sortBy(destinations, 'outputType')` be more clear to you or the unclear bit is the underlying sorting function? 


P.S. The comparison is always `<` (i.e. `'aggelos' < 'basilis'`, `1 < 2`, etc.)

web/src/components/RelatedDestinations/RelatedDestinations.tsx
```diff
@@ -54,58 +52,45 @@ const RelatedDestinations: React.FC<RelatedDestinationsSectionProps> = ({
...
     return (
       <RRLink to={urls.settings.destinations.list()}>
-        {destinations.map(destination => (
-          <Flex key={destination.outputId} align="center" mb={2}>
-            {getLogo(destination)}
-            {destination.displayName}
-          </Flex>
-        ))}
+        <Flex inline direction="column" spacing={2}>
+          <LimitItemDisplay limit={limit}>
```
alexmylonas, line 76: I dont think we should add `LimitItemDisplay` for the verbose version

alexmylonas (approved):
Overall looks good I think I have a small comment alongside @vorillaz but i am approving

3nvi:
@vorillaz @alexmylonas You both commented on the limiting of display items. I feel we should either always do it or never do it. We should either:

- Keep it as an optional prop (i.e. `limit`) and force it all the time
- Remove it from this component altogether and just use `<LimitItemDisplay>` as a wrapper of `<RelatedDestinations>` wherever it's used.

Thoughts?

web/src/components/RelatedDestinations/RelatedDestinations.tsx
vorillaz: I am cool proceeding with lodash, no worries.
Just a minor note here, sorting algorithm callback methods should always return 1, -1, or 0. 
```
'aggelos' > 'basilis' // false
'aggelos'.localeCompare('basilis') // -1
```

vorillaz (approved):

PR #2145, opened 2020-11-30
Refactor log processor to use logstream.Stream and context.Context

### LINKED ISSUE #2131: Introduce `LogEntryStream` interface, replacing current line splitter with `LineStream`
### Description

In order to fix #2001, the first step will be to introduce the `LogEntryStream` interface and replace our current 'hardcoded' line splitter with `LineStream`

### Related Services

Log Processor

### Designs

See #2001 

### Acceptance Criteria

- [ ] Current line splitting is changed to use `LineStream`

alxarch:
## Background

Closes #2131 

## Changes

- Adds `logstream` module
- Adds `logstream.LineStream` to parse a log line by line
- Handle gzip auto-detection in `s3pipe` transparently, to avoid unnecessary double buffering
- Pass `context.Context` all the way through to the `processor.run()` method.

## Testing

- mage test:go
- E2E ran for a couple of days


internal/log_analysis/log_processor/s3pipe/s3pipe.go
```diff
@@ -115,17 +132,26 @@ func copyBuffers(w *io.PipeWriter, parts <-chan *bytes.Buffer) {
 			return
 		}
 	}
+	// Make sure a download error that closes parts early does not block the reader forever
+	if peek != nil {
+		peek(nil)
+	}
```
kostaspap, line 138: Interesting edge case! Is it possible to add a unit test for it?

kostaspap (changes requested):

internal/log_analysis/log_processor/processor/logstream/linestream.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		s.err = err
+	}
+	return line
+}
+
+var ErrInvalidUTF8 = goerrors.New("invalid UTF8 encoding")
+
+func (s *LineStream) readLine() ([]byte, error) {
+	line, isPrefix, err := s.r.ReadLine()
```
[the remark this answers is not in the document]
rleighton: If not, add a comment about `isPrefix`. Not clear to a new reader what is happening here.

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -73,12 +73,19 @@ func Process(
...
-				if err := processDataStream(dataStream, resultsChannel, newProcessor); err != nil {
-					return err
+			for {
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				case stream, ok := <-dataStreams:
+					if !ok {
+						return nil
+					}
```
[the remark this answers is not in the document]
rleighton: Doesn't closing the channel have the same effect with less code?

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -123,6 +125,13 @@ func pollEvents(
 			for _, msg := range messages {
 				// pass lambda context to set FULL deadline to process which is pushed down into downloader
 				dataStreams, err := generateDataStreamsFunc(ctx, aws.StringValue(msg.Body))
+				if err == nil {
+					// This is a temporary workaround to ensure all S3 streams are readable.
```
rleighton, line 129: I still do not understand the purpose of this ... can you add more to the comment to elaborate?

internal/log_analysis/log_processor/processor/logstream/linestream.go
```diff
@@ -0,0 +1,137 @@
...
+func (s *LineStream) Err() error {
+	if s.err == io.EOF {
+		return nil
+	}
+	return s.err
+}
+
+// Next reads the next line from the log.
+func (s *LineStream) Next() []byte {
+	if err := s.err; err != nil {
```
giorgosp, line 72: just `if s.err !=nil` ?

internal/log_analysis/log_processor/processor/logstream/linestream.go
```diff
@@ -0,0 +1,137 @@
...
+var ErrInvalidUTF8 = goerrors.New("invalid UTF8 encoding")
+
+func (s *LineStream) readLine() ([]byte, error) {
+	line, isPrefix, err := s.r.ReadLine()
+	// NOTE: ReadLine either returns a non-nil line or it returns an error, never both.
+	if err != nil {
+		return nil, err
+	}
+	if s.numLines == 0 {
+		// Check for valid UTF8 stream on first read.
```
giorgosp, line 94: Maybe update the comment as to why we do this check if numlines == 0??

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -182,3 +196,23 @@ func receiveFromSqs(ctx context.Context, sqsClient sqsiface.SQSAPI) ([]*sqs.Mess
...
+			continue
+		}
+		grp.Go(func() error {
+			return readZero(r)
+		})
+	}
+	return grp.Wait()
+}
+
+func readZero(r io.Reader) error {
```
giorgosp, line 214: what is the point of `readZero()` ?

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -174,31 +184,36 @@ func NewFactory(resolver logtypes.Resolver) Factory {
...
-			if err == io.EOF {
-				err = nil
-				p.processLogLine(line, outputChan)
-			}
+		line := stream.Next()
+		if line == nil {
 			break
 		}
-		p.processLogLine(line, outputChan)
+		p.processLogLine(ctx, string(line), outputChan)
```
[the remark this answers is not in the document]
alxarch: the parsers expect a string. Until that interface changes, the allocation cannot be avoided.

internal/log_analysis/log_processor/processor/logstream/linestream.go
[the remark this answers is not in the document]
alxarch: `ReadBytes` copies the bytes. `ReadLine` returns slices out of the 'live' buffer.

internal/log_analysis/log_processor/s3pipe/s3pipe.go
alxarch: Not sure if it's easy to time, I'll try

internal/log_analysis/log_processor/processor/logstream/linestream.go
alxarch: The comment says 'on first read'. I can move it above if.

internal/log_analysis/log_processor/processor/processor.go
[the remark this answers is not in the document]
alxarch: The `ok` here will be true when `dataStreams` has closed. We want to return then and not wait for `ctx.Done()` so that `resultsChannel` closes ASAP.

internal/log_analysis/log_processor/s3pipe/s3pipe.go
kostaspap: I guess I am not seeing the issue that this is trying to solve. Why is `peek` required? What's the condition that can trigger it?

internal/log_analysis/log_processor/processor/stream.go
kostaspap: This is something I've asked @alxarch to include. 
We have seen the following cases in the past (the reason for this one: https://github.com/panther-labs/panther-enterprise/issues/1454 )
1. User will onboard a new log source
2. User will not give the appropriate KMS permissions to the role assumed by Panther
3. When we try to read the data, we will get "AccessDeniedExceptions". 

The code before this PR, we would read the first chunk of data *inside* the `generateDataStreamsFunc`. So KMS/Access errors would cause line 127 to return an error. We would skip this error and continue with the proper execution. This means that permissions issue for a single S3 object wouldn't fail the batch. 

This PR modifies the way the stream works so that we try to read data at a later stage, when we start to parse the first line. 
This can have the side effect that we detect errors like AccessDenied at a later stage - and right now the way the log processor works, such errors will cause the whole batch to fail. 

This extra code added by @alxarch was to make sure we trigger the read operation earlier (like we were doing until now) to make sure we don't lose the ability to detect issues "early" and react on them.

internal/log_analysis/log_processor/s3pipe/s3pipe.go
```diff
@@ -52,60 +52,77 @@ func (dl Downloader) Download(ctx context.Context, input *s3.GetObjectInput) io.
 	parts := make(chan *bytes.Buffer, 1)
 	// We reset our copy of Downloader to read chunks in sequence and queue them into parts.
 	dl.reset(ctx, w, parts)
-	// We return a reader that will start downloading on first call to Read().
-	// Chunks will be queued and written to the write part of the pipe.
-	return &downloadReader{
+
```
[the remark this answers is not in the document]
alxarch: We don't. This is intentional. We want an independent copy of the struct so we can 'normalize' the fields with our buffer provider without affecting the original.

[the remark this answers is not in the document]
kostaspap: Oh. Can you put a BIG warning sign there? This is a bug waiting to happen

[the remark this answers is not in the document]
alxarch: Added comment explaining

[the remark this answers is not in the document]
kostaspap: Or even make it more explicit by using a new New() method or something that initializes the struct accordingly.

internal/log_analysis/log_processor/s3pipe/s3pipe.go
alxarch: What we are trying to achieve here is to not block forever on the read side of things. 
`peek` is a good name since it 'peeks into' the first chunk. In our particular case, it also closes the `ready` channel once done. All reads will continue only after `ready` is closed. One of the shortcomings of Golang's channels is that you can only close a channel once. So we *have* to call `peek` for the read to proceed, even if it returns an error to the one reading.

internal/log_analysis/log_processor/processor/logstream/linestream.go
alxarch: fixed

internal/log_analysis/log_processor/processor/logstream/linestream.go
alxarch: Added longer explanation

internal/log_analysis/log_processor/processor/stream.go
alxarch: Glad you asked :)
`readZero` is used to kick off the S3 download and get the first chunk *before* reading data from the reader. What this enables us to do, is to detect if the stream is gzipped and 'transparently' swap the underlying `io.Reader` from  `*io.PipeReader`  to `*gzip.Reader`. This way, we can handle gzip streams transparently without having to do unnecessary buffering. (We already have quite a lot of buffering being done to download in chunks)

internal/log_analysis/log_processor/processor/stream.go
alxarch: Also this was not put inside 'generateDataStreamsFunc' so that once we have proper SQS messsage tracking, we can easily remove it here and have the per-message processing happen in a separate goroutine.

internal/log_analysis/log_processor/s3pipe/s3pipe.go
kostaspap: Ok, I have to say looking at the code more closely I was able to identify the issue you are trying to address here.

I have to say it is a bit challenging to follow the code logic and identify this issue. This edge case is really... hidden.

It seems that a lot of the challenges here are because we try to avoid by all means reading any data when `Download` is called. But do we really need to do this? 
1. The way the code is written right now, we anyway read the first chunks serially
2. We can easily parallelize the code in the future to read them in parallel, we just have to make sure that the `Download` function is invoked by different go routines. 

I am worried that we are sacrificing some "ease of maintenance" here to get an edge which we would achieve in a different way. 

Maybe best to chat offline?

internal/log_analysis/log_processor/s3pipe/s3pipe.go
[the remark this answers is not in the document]
alxarch: Added comment explaining the use of a non-pointer receiver. It is common to use the 'pass-by-copy' for structs that will need to be modified for this particular method.

internal/log_analysis/log_processor/s3pipe/s3pipe.go
alxarch: Wiring up the chunked reader is a separate concern from actually reading from it. Even if we decide to read files in parallel, we might want to restrict the number of concurrent readers. The separation of initialization and execution is crucial for controlling parallel code. The edge cases are adressed in this PR and I don't think 'dumping it down' will help when we reach the point where concurrency is handled. Sometimes things are complex by nature. I was recently looking for a good solution to the 'singleflight' problem

kostaspap: > Sometimes things are complex by nature 

We can definitely choose which path to take. I am not sure I understand the reason why we cannot handle this in a performant way. Or what are the differences in performance between the options. 

I definitely know though that if we happen to hit an edge case here, it will be tough to fix it. It is already tough to read and understand it. We are the maintainers of the code, so understanding any piece of it is important, hiding it behind an interface is going to make is easier to use but not easier to debug.

alxarch: The separation is important so that you can control which goroutine will be doing the blocking I/O stuff. 
You want to 'prepare' the work in a 'main' goroutine and then fan out the blocking stuff to 'worker' goroutines.

internal/log_analysis/log_processor/processor/logstream/linestream.go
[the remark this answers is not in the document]
alxarch: Commented the function heavily, since people are not familiar with `bufio.Reader.ReadLine`

internal/log_analysis/log_processor/processor/logstream/linestream.go
```diff
@@ -0,0 +1,150 @@
...
+	line, isPrefix, err := s.r.ReadLine()
+	if err != nil {
+		return nil, err
+	}
+	// Check for valid UTF8 stream on first read.
+	// We want to be doing this here, before handling isPrefix.
+	// This allows us to avoid reading a whole non-UTF8 blob searching for "\n"
+	// NOTE: if we encounter 'weird' log files using UTF16 and/or BOM headers, we need to update this detection
+	if s.numLines == 0 {
+		if !isValidUTF8(line, isPrefix) {
```
kostaspap, line 104: is it possible that the `line` here has size < utf8.UTFMax?

internal/log_analysis/log_processor/processor/logstream/linestream.go
```diff
@@ -0,0 +1,150 @@
...
+	return s.scratch, err
+}
+
+func isValidUTF8(p []byte, partial bool) bool {
+	totalSize := len(p)
+	validSize := 0
+	for len(p) > 0 {
+		r, n := utf8.DecodeRune(p)
+		if r == utf8.RuneError {
+			break
```
kostaspap, line 139: return false?

internal/log_analysis/log_processor/processor/logstream/linestream.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (s *LineStream) readLine() ([]byte, error) {
+	// We use ReadLine instread of ReadBytes("\n") here to avoid unnecessary copying of data.
+	// line is a 'live' slice from the bufio.Reader
+	// isPrefix is set to true if the line did not fit that read buffer, so that we know this is partial data.
+	// err is any read error that occurred.
+	// NOTE: ReadLine either returns a non-nil line or it returns an error, never both.
+	// Another reason for using ReadLine is that with it, we can detect UTF8 on the first line chunk.
+	// If we were using ReadBytes("\n"), and the file was a non-UTF8 blob, we might end up reading the whole file seatching for "\n".
+	line, isPrefix, err := s.r.ReadLine()
```
kostaspap, line 95: Is it possible to add a unit test that tests the scenarios where `isPrefix` is true? I am not sure we have any testing for this part of the code.

kostaspap (approved):
LGTM

PR #2153, opened 2020-11-30
enabled data models bug

lindsey-w:
## Background

Found issue while doing some on call work, where the rules engine was expecting to get enabled DataModels, but it was returning all DataModels (even disabled ones). 

## Changes

- Added integration test to validate / catch this in the future
- Updated the listDataModels to build a ddb filter

## Testing

- `mage test:integration`


internal/core/analysis_api/handlers/list_data_models.go
```diff
@@ -76,3 +79,28 @@ func (API) ListDataModels(input *models.ListDataModelsInput) *events.APIGatewayP
...
+			expression.Name("enabled"), expression.Value(*input.Enabled)))
+	}
+
+	if input.NameContains != "" {
+		filters = append(filters, expression.Contains(expression.Name("lowerId"), input.NameContains).
+			Or(expression.Contains(expression.Name("lowerDisplayName"), strings.ToLower(input.NameContains))))
+	}
+
+	if len(input.LogTypes) > 0 {
+		// a policy with no resource types applies to all of them
```
austinbyers, line 96: Copy-pasta comment needs to be updated for data models

austinbyers (approved):
Sorry about that, I must have broken this when doing the migration. Good find!

Issue #2149, opened 2020-11-30
Query 'getIntegrationTemplate' fails with AccessDenied


alexmylonas:
### Describe the bug

There is an issue when trying to onboard a new S3 log source. 

The issue is apparent when getting to the Setup IAM roles in UI
**The query for getting the CFN template fails with:**
```
[
  {
    "path": [
      "getS3LogIntegrationTemplate"
    ],
    "data": null,
    "errorType": "Lambda:Unhandled",
    "errorInfo": null,
    "locations": [
      {
        "line": 2,
        "column": 3,
        "sourceName": null
      }
    ],
    "message": "AccessDenied: Access Denied\n\tstatus code: 403, request id: C0B7A4B79207E8FD, host id: pX2nfVvi4+g4a4Cc/GOim4mm+N9eYZQz+Hf1CW8xW4o0ogLsH7chP5+lYck2dxUpsuSnr+hioyc="
  }
]
```
### Steps to reproduce

1. Go to `ListLogSource`
2. Click on New Source
3. Add inputs on first screen
4. Click Continue
5. You 'll get a failing pop up for this query

### Expected behavior

Query should be executed properly

### Environment

- Panther version or commit: de762b5b718e8952df8dc6cea4b9ed94fbe5ca3a

### Notes

A possible root cause could be related to this [PR](https://github.com/panther-labs/panther/pull/2077/files)

austinbyers:
Closing as a duplicate of https://github.com/panther-labs/panther-enterprise/issues/1594. I verified this has been fixed in the release environment.

The problem is just that we need to publish the auxiliary templates with `v1.14.0` instead of `v1.14.0-RC`. Since the problem can be fixed by just publishing templates, it's not really a problem with the app itself, and I'm closing the issue.

However, I'm going to take #2158 and work on passing consistent version information everywhere to prevent this from happening again in the future

PR #2158, opened 2020-11-30
fix: references either stable or pre-release version templates

s0l0ist:
## Background

Fixes an issue when downloading CF templates, we only reference the stable version string. This PR changes the FE constants so we can always download CF templates for internal development/pre-release and production/stable releases.

Fixes: #2149 

## Changes

- Use regex to extract the proper version strings
- Deployment from source now reads directly from the version file.

TODO: 
- Master template deployment needs to be fixed and tested.

## Testing

- Manually/Visually


web/src/constants.ts
```diff
@@ -104,9 +104,15 @@ export const RESOURCE_TYPES = [
...
-const VERSION_PARTS = STABLE_PANTHER_VERSION.split('.'); // ["v1", "7", "1]
-const MINOR_PANTHER_VERSION = `${VERSION_PARTS[0]}.${VERSION_PARTS[1]}`.replace('v', ''); // "1.7"
+// We use a regex to capture the version string we're interested in
+// Ex1: Given a string "v1.13.0-RC-release-1.13-abc1234", we find Array ["v1.13.0-RC", "v1.13.0", "1", "13", "0", "-RC"]
+// Ex2: Given a string "v1.13.0-release-1.13-abc1234", we find Array ["v1.13.0", "v1.13.0", "1", "13", "0", undefined]
+const PANTHER_VERSION_MATCH = pantherConfig.PANTHER_VERSION.match(
+  /(v(\d+)\.(\d+)\.(\d+))(?:(-RC)?)/
+);
+export const FULL_PANTHER_VERSION = PANTHER_VERSION_MATCH[0]; // e.g. could be "v1.13.0" or "v1.13.0-RC"
+export const STABLE_PANTHER_VERSION = PANTHER_VERSION_MATCH[1]; // e.g. "v1.13.0"
```
austinbyers, line 114: So there are now 3 exported version strings:

* `pantherConfig.PANTHER_VERSION`
* `FULL_PANTHER_VERSION`
* `STABLE_PANTHER_VERSION`

Looking at those 3 variable names, I would have no idea what the difference is between them.

-----

I would say let's rename like this:

* `STABLE_PANTHER_VERSION` => `PANTHER_PATCH_VERSION` ("v1.13.0")
* `FULL_PANTHER_VERSION` => `QUALIFIED_PANTHER_PATCH_VERSION` ("v1.13.0-RC")

austinbyers: This is following the "major/minor/patch" terminology from semver

web/src/constants.ts
```diff
@@ -104,9 +104,15 @@ export const RESOURCE_TYPES = [
...
-export const STABLE_PANTHER_VERSION = pantherConfig.PANTHER_VERSION.split('-')[0]; // e.g. "v1.7.1"
-const VERSION_PARTS = STABLE_PANTHER_VERSION.split('.'); // ["v1", "7", "1]
-const MINOR_PANTHER_VERSION = `${VERSION_PARTS[0]}.${VERSION_PARTS[1]}`.replace('v', ''); // "1.7"
+// We use a regex to capture the version string we're interested in
+// Ex1: Given a string "v1.13.0-RC-release-1.13-abc1234", we find Array ["v1.13.0-RC", "v1.13.0", "1", "13", "0", "-RC"]
+// Ex2: Given a string "v1.13.0-release-1.13-abc1234", we find Array ["v1.13.0", "v1.13.0", "1", "13", "0", undefined]
+const PANTHER_VERSION_MATCH = pantherConfig.PANTHER_VERSION.match(
+  /(v(\d+)\.(\d+)\.(\d+))(?:(-RC)?)/
+);
+export const QUALIFIED_PANTHER_PATCH_VERSION = PANTHER_VERSION_MATCH[0]; // e.g. could be "v1.13.0" or "v1.13.0-RC"
```
3nvi, line 113: if we have multiple RCs, all users will report that they are using `v1.13.0-RC` although in reality therre may be multiple different RCs.

Is that something that we are ok with?

3nvi (approved):
Approved, but just a minor user-debugging concern. Other than that, it looks good

web/src/constants.ts
austinbyers: We will still show the commit sha to distinguish this case

austinbyers (changes requested):
I thought I had commented somewhere in this PR, but I guess not:

This change doesn't work for the master template, I'm going to take over the PR (or replace it with another one) that fixes all of the versioning problems more holistically

alexmylonas:
@austinbyers Keep in mind the issue #2149 which i am not sure this versioning fix will resolve!

austinbyers:
> @austinbyers Keep in mind the issue #2149 which i am not sure this versioning fix will resolve!

All of the S3 template issues we've seen are the same: we published `v1.14.0-RC` and the app was expecting `v1.14.0`.  The fix, which we've done, is to just re-publish the S3 templates with the latter version. The problem only exists for base versions with a dash in them.

This PR, as written, is a quick hack to fix this more permanently by parsing out the full base version with a regex. So yes, the changes would address the issue you linked. However, this fix is fragile and doesn't work for master-stack deployments.

I am working on a more holistic fix (which will supersede this PR) which will pass the base version everywhere as a separate variable so we don't have to use string parsing to pull it out

PR #2167, opened 2020-12-02
Fix uuid validation and case-insensitive filtering

austinbyers:
## Background

Our release QA revealed a few bugs with the `analysis-api`, which isn't that surprising since we removed the analysis-api gateway in this release.

Fortunately, the fixes are small and easy

## Changes

- Do not require userIDs to be UUID4s: they just need to be non-empty
    - In the community edition, userIDs happen to always be uuids, but that's not always true in enterprise
    - Updated the integration test with a non-uuid userID as a regression test
- ID filtering is now correctly case-insensitive
    - We were lowercasing the display name search, but not the ID search
- Related minor performance improvement: when sorting by display name / ID, we can use the existing dynamo columns "LowerID" and "LowerDisplayName" to avoid having to recompute the lowercase string on each comparison

## Testing

- Deployed to dev account, verified rule/policy case-insensitive filtering by display name and ID
- Fixes verified in release QA environment
- `analysis-api` integration tests

### DIFF
--- a/internal/core/analysis_api/handlers/list.go
+++ b/internal/core/analysis_api/handlers/list.go
@@ -84,7 +84,7 @@ func pythonListFilters(enabled *bool, nameContains, severity string, types, tags
 
 	if nameContains != "" {
 		filters = append(filters, expression.Contains(expression.Name("lowerId"), nameContains).
-			Or(expression.Contains(expression.Name("lowerDisplayName"), strings.ToLower(nameContains))))
+			Or(expression.Contains(expression.Name("lowerDisplayName"), nameContains)))
 	}
 
 	if len(types) > 0 {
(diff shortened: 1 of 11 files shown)


internal/core/analysis_api/handlers/list.go
```diff
@@ -84,7 +84,7 @@ func pythonListFilters(enabled *bool, nameContains, severity string, types, tags
 
 	if nameContains != "" {
 		filters = append(filters, expression.Contains(expression.Name("lowerId"), nameContains).
-			Or(expression.Contains(expression.Name("lowerDisplayName"), strings.ToLower(nameContains))))
+			Or(expression.Contains(expression.Name("lowerDisplayName"), nameContains)))
```
austinbyers, line 87: Note that the `strings.ToLower` was being applied only to the display name, not the ID. Now the lowercasing happens in the List* handlers before calling this function

s0l0ist (approved):

nhakmiller (approved):

Issue #2169, opened 2020-12-02
Rule Errors should have threshold 1


kostaspap:
### Describe the bug

The Rule Error currently uses as threshold the rule threshold. It should have 

### Steps to reproduce

1. Create a rule with syntax error
2. Set a threshold to the rule > 1
3. Send a message
4. See that rule error will not trigger

### Expected behavior

Rule errors should have threshold 1

### Environment

How are you deploying or using Panther?

- Panther version or commit: 1.14-RC

PR #2171, opened 2020-12-02
Fixes Rule Errors threshold and dedup period

kostaspap:
## Background

Closes https://github.com/panther-labs/panther/issues/2169
Updated Rule errors to use the same dedup period as the Rule matches (instead of hardcoded 24h)

## Testing

- mage test:ci
- Deployed and verified E2E


alxarch (approved):

alxarch:
The linked issue does not mention dedup period. It only says the threshold should be 1.

kostaspap:
> The linked issue does not mention dedup period. It only says the threshold should be 1.

Yup, this was a separate issue

Issue #2029, opened 2020-11-14
s3sns opstool


rleighton:
### Description

Add  opstool `s3sns`, similar to `s3queue` but will send s3 file notifications to sns. Use to replay data in subscribed services such as Snowflake and the Rules Engine.

PR #2173, opened 2020-12-02
s3sns opstool

rleighton:
## Background

This PR adds an opstool `s3sns`, similar to s3queue but will send s3 file notifications to sns. 

Used to replay data in subscribed services such as Snowflake and the Rules Engine.

This can be very useful for testing the Rules Engine allowing replay of data.

Closes: https://github.com/panther-labs/panther/issues/2029

## Changes

- Add `s3sns`
- Use new `optools` helpers in `s3queue`
- Factor out the s3 listing code into package `s3list` to be used by `s3queue` and `s3sns`
- Added function `DataTypeFromS3Key)` to `awsglue` pkg

## Testing

- mage test:ci
- ran integration tests
- ran cli to drive data thru the rules engine


cmd/opstools/s3sns/s3sns.go
```diff
@@ -0,0 +1,212 @@
...
+	go func() {
+		s3list.ListPath(s3Client, s3path, limit, notifyChan, errChan, stats)
+		queueWg.Done()
+	}()
+
+	var errorWg sync.WaitGroup
+	errorWg.Add(1)
+	go func() {
+		for err := range errChan { // return last error
+			failed = err
```
kostaspap, line 85: Maybe you want to use`multierr` here: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/processor/processor.go#L115

cmd/opstools/s3sns/s3sns.go
```diff
@@ -0,0 +1,212 @@
...
+
+	queueWg.Wait()
+	close(errChan)
+	errorWg.Wait()
+
+	return failed
+}
+
+// post message per file as-if it was an S3 notification
+func publishNotifications(snsClient snsiface.SNSAPI, lambdaClient lambdaiface.LambdaAPI,
```
kostaspap, line 98: same as above, consider passing a `context.Context` object and just updating the method to return err rather than passing the errChan

cmd/opstools/s3sns/s3sns.go
```diff
@@ -0,0 +1,212 @@
...
+			err = errors.Errorf("%s: failed to invoke %#v", *resp.FunctionError, method)
+			return
+		}
+
+		var availableLogTypes logtypesapi.AvailableLogTypes
+		err = jsoniter.Unmarshal(resp.Payload, &availableLogTypes)
+		if err != nil {
+			err = errors.Wrapf(err, "failed to unmarshal: %s", string(resp.Payload))
+			return
+		}
```
kostaspap, line 196: logtypes API has its own client. 
You can instantiate like this: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/datacatalog_updater/main/lambda.go#L80-L83
You get the types like this: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go#L65

You don't need any custom marshaling/unmarshaling here.

cmd/opstools/s3sns/s3sns.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+}
+
+// logType is not derivable from the s3 path, need to use API
+var (
+	initTablenameToLogType sync.Once
+	tableNameToLogType     map[string]string
+)
+
+func logTypeFromS3Key(lambdaClient lambdaiface.LambdaAPI, s3key string) (logType string, err error) {
```
kostaspap, line 169: Can't you just use https://github.com/panther-labs/panther/blob/master/internal/log_analysis/awsglue/partition.go#L86 ?

cmd/opstools/s3sns/s3sns/main.go
```diff
@@ -0,0 +1,160 @@
...
+	CONCURRENCY = flag.Int("concurrency", 50, "The number of concurrent sqs writer go routines")
+	LIMIT       = flag.Uint64("limit", 0,
+		"If non-zero, then limit the number of files to this number.")
+	TOPIC = flag.String("topic", "panther-processed-data-notifications",
+		"The name of the log processor topic to send notifications.")
+	ATTRIBUTES = flag.Bool("attributes", false,
+		"If true, add SNS attributes that would enable the rule engine and datacatalog updater to receive events.")
+	INTERACTIVE = flag.Bool("interactive", true, "If true, prompt for required flags if not set")
+	VERBOSE     = flag.Bool("verbose", false, "Enable verbose logging")
+	DEBUG       = flag.Bool("debug", false, "Enable debug logging")
```
kostaspap, line 59: Not 100% what is the difference between the two flags. Seems like one would be sufficient?

cmd/opstools/s3sns/s3sns/main.go
```diff
@@ -0,0 +1,160 @@
...
+	stats := &s3list.Stats{}
+	go func() {
+		sig := make(chan os.Signal, 1)
+		signal.Notify(sig, syscall.SIGTERM, syscall.SIGINT)
+		caught := <-sig // wait for it
+		logger.Fatalf("caught %v, sent %d files (%.2fMB) to %s in %v",
+			caught, stats.NumFiles, float32(stats.NumBytes)/(1024.0*1024.0), *TOPIC, time.Since(startTime))
+	}()
+
+	err = s3sns.S3Topic(sess, *ACCOUNT, *S3PATH, s3Region, *TOPIC, *ATTRIBUTES, *CONCURRENCY, *LIMIT, stats)
```
kostaspap, line 119: We pass so many strings as parameters to this topic, I think its easy to mix them app. 
I am thinking maybe better to just groups them in an `Opts` struct and pass that struct in. 
Something like : https://github.com/panther-labs/panther/blob/infer-schemas/cmd/devtools/customlogs/main.go#L56-L59 and https://github.com/panther-labs/panther/blob/infer-schemas/cmd/devtools/customlogs/customlogs/infer.go#L44-L47 . 
Easy to mix up the order of the arguments.

cmd/opstools/s3sns/s3sns/main.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	stats := &s3list.Stats{}
+	go func() {
+		sig := make(chan os.Signal, 1)
+		signal.Notify(sig, syscall.SIGTERM, syscall.SIGINT)
+		caught := <-sig // wait for it
+		logger.Fatalf("caught %v, sent %d files (%.2fMB) to %s in %v",
+			caught, stats.NumFiles, float32(stats.NumBytes)/(1024.0*1024.0), *TOPIC, time.Since(startTime))
+	}()
```
kostaspap, line 117: Why is this needed?

cmd/opstools/s3sns/s3sns.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/panther-labs/panther/internal/log_analysis/awsglue"
+	"github.com/panther-labs/panther/internal/log_analysis/notify"
+	"github.com/panther-labs/panther/internal/log_analysis/pantherdb"
+)
+
+const (
+	topicArnTemplate = "arn:aws:sns:%s:%s:%s"
+)
+
+func S3Topic(sess *session.Session, account, s3path, s3region, topic string, attributes bool,
```
kostaspap, line 50: Another idea would be to create a struct `S3ToSns` and initialize its parameters with this. Then just have a `Run(ctx) error` method

cmd/opstools/s3sns/s3sns.go
```diff
@@ -0,0 +1,212 @@
...
+
+	return s3sns(s3.New(sess.Copy(&aws.Config{Region: &s3region})), sns.New(sess), lambda.New(sess),
+		account, s3path, topic, *sess.Config.Region, attributes, concurrency, limit, stats)
+}
+
+func s3sns(s3Client s3iface.S3API, snsClient snsiface.SNSAPI, lambdaClient lambdaiface.LambdaAPI,
+	account, s3path, topic, topicRegion string, attributes bool,
+	concurrency int, limit uint64, stats *s3list.Stats) (failed error) {
+
+	topicARN := fmt.Sprintf(topicArnTemplate, topicRegion, account, topic)
```
kostaspap, line 61: Instead of using string templates + fmt.Sprintf, you could do something like that: https://github.com/panther-labs/panther/blob/master/internal/core/custom_resources/resources/glue_tables.go#L104-L115

Note that the above code would allow the script to work for other partitions as well e.g. China or GovCloud

cmd/opstools/s3sns/s3sns.go
```diff
@@ -0,0 +1,212 @@
...
+				failed = true
+				continue
+			}
+			logType, err := logTypeFromS3Key(lambdaClient, key)
+			if err != nil {
+				errChan <- errors.Wrapf(err, "failed to get log type from %s", key)
+				failed = true
+				continue
+			}
+			messageAttributes = notify.NewLogAnalysisSNSMessageAttributes(dataType, logType)
```
kostaspap, line 143: There are other parts of the code that so similar operations, so we have already the helper methods available. 
1. You can get Glue partition information from the S3 key: https://github.com/panther-labs/panther/blob/infer-schemas/internal/log_analysis/awsglue/partition.go#L86
2. You get the datatype by providing the logtype: https://github.com/panther-labs/panther/blob/infer-schemas/internal/log_analysis/log_processor/destinations/s3.go#L283

cmd/opstools/s3sns/s3sns.go
```diff
@@ -0,0 +1,212 @@
...
+)
+
+const (
+	topicArnTemplate = "arn:aws:sns:%s:%s:%s"
+)
+
+func S3Topic(sess *session.Session, account, s3path, s3region, topic string, attributes bool,
+	concurrency int, limit uint64, stats *s3list.Stats) (err error) {
+
+	return s3sns(s3.New(sess.Copy(&aws.Config{Region: &s3region})), sns.New(sess), lambda.New(sess),
```
kostaspap, line 53: We have seen SNS returning `tcp connection reset` quite "often" under high loads. Better use the ConnectionRetryer here to avoid retryable errors. 
Also consider increasing the number of retries in general.

kostaspap (approved):
I've put some style comments as well. I am not sure if it makes sense to address all of them given this is a command line tool and not meant to be a long-running process in some lambda or sth.

internal/log_analysis/awsglue/awsglue.go
```diff
@@ -66,3 +68,26 @@ func DataPrefix(databaseName string) string {
 		panic(databaseName + " is not associated with an s3 prefix")
 	}
 }
+
+func DataTypeFromS3Key(s3key string) (dataType pantherdb.DataType, err error) {
```
kostaspap, line 72: See above, this shouldn't be needed. We already have functions that provide this information.

rleighton:
> I've put some style comments as well. I am not sure if it makes sense to address all of them given this is a command line tool and not meant to be a long-running process in some lambda or sth.

Thx for the suggestions. This code was really just copy-pasta-re-factor from the very old `s3queue` done quickly because we needed it. Your suggestions are all good. I'll implement them and upate `s3queue` at the same time.

cmd/opstools/s3sns/s3sns/main.go
rleighton: Tells you what you did after a ^C

cmd/opstools/s3sns/s3sns.go
rleighton: Nope, can't get logType . This hopefully will be fixed with Alex's new work on schema management.

internal/log_analysis/awsglue/awsglue.go
rleighton: I can't find one.

cmd/opstools/s3sns/s3sns.go
rleighton: Done

cmd/opstools/s3sns/s3sns.go
rleighton: Done

PR #2175, opened 2020-12-02
Add backend support for filtering by multiple severities

### LINKED ISSUE #1996: Allow Rule Filtering by multiple Severities
### Description

Users should be able to filter rules by more than one severity at a time

### Acceptance Criteria

- The FE exposes a multi-select for severity selection
- The user can select multiple severities
- Rules that have one of the selected severities can appear as results

### LINKED ISSUE #1997: Support multi-severity filtering for rules
### Description

Currently we can only filter rules through a single severity, but the API should be able to support filtering by more than one. Specifically, the lambda service should accept a list of severities (instead of 1) and return a list of rules matching any of the given severities

### Related Services

analysis-api

### Designs

N/A

### Acceptance Criteria

- The api accepts a list of strings (list of severities) in the `/rules/list` endpoint
- The GraphQL schema gets updated to allow for an optional list of strings to be passed as input (i.e. `[String!]`)
- Backend tests are updated

### Notes

The work should be merged into a feature branch in order not to break `master`

### LINKED ISSUE #1999: Allow filtering Rules by multiple severities in the UI
### Description

Currently we can only filter rules through a single severity, but the UI should be able to support filtering by more than one. Specifically, the user should be able to specify more than one (1) severity and the API should return a list of rules matching any of the given severities

### Related Services

web

### Designs

N/A

### Acceptance Criteria

- The frontend changes the severity `Combobox` to a `MultiCombobox`
- The GraphQL apollo hooks are updated to match the new schema updates
- The tests are updated to match the changes

### Notes

The work should be merged into a feature branch in order not to break `master`

### Caveats

Policies might be affected by this, so we should investigate whether policy filtering continues to work

AlexPanther:
## Background

Customers would benefit from being able to select a more flexible subset of rules based on severity.

Backend: Closes #1997
[Aggelos]: Closes #1999
Total epic: Closes #1996  

## Changes

- Change severity to array from string
- Add/update test
- Add Dynamo filtering/make optimization for readability

## Testing

- Select a filter, apply, select a new filter to override previous filter
- Clear filters


api/graphql/schema.graphql
```diff
@@ -226,7 +226,7 @@ input ListRulesInput {
   nameContains: String
   enabled: Boolean
   logTypes: [String!]
-  severity: SeverityEnum
+  severity: [SeverityEnum]
```
3nvi, line 229: If we specify an array, we can't add `null` values within it 😉 
```suggestion
  severity: [SeverityEnum!]
```

api/lambda/analysis/models/policies.go
```diff
@@ -57,7 +57,7 @@ type ListPoliciesInput struct {
 	ResourceTypes []string `json:"resourceTypes" validate:"max=500,dive,required,max=500"`
 
 	// Only include policies with this severity
-	Severity models.Severity `json:"severity" validate:"omitempty,oneof=INFO LOW MEDIUM HIGH CRITICAL"`
+	Severity []models.Severity `json:"severity" validate:"dive,oneof=INFO LOW MEDIUM HIGH CRITICAL"`
```
3nvi, line 60: just making sure that even after `omitempty` is removed, we can still send `severity: nil` from the FE, in case we don't want to filter by severity at all.

3nvi (commented):
Nice one @AlexPanther. Just a couple of comments from my side 😄

api/lambda/analysis/models/policies.go
austinbyers: Yes, the `nil` value should work fine, but we'll of course deploy and test with and without the filter.

`validate:"omitempty,dive"` and `validate:"dive"` are essentially the same - if there are no elements, the validation won't run either way

internal/core/analysis_api/handlers/list.go
```diff
@@ -74,7 +74,7 @@ var (
 // like lastModified.
 
 // Dynamo filters common to both ListPolicies and ListRules
-func pythonListFilters(enabled *bool, nameContains, severity string, types, tags []string) []expression.ConditionBuilder {
+func pythonListFilters(enabled *bool, nameContains string, severity []compliancemodels.Severity, types, tags []string) []expression.ConditionBuilder {
```
austinbyers, line 77: This line is now too long (which is why `test:go` failed in CI):

```suggestion
func pythonListFilters(
    enabled *bool,
    nameContains string,
    severity []compliancemodels.Severity,
    types, tags []string,
) []expression.ConditionBuilder {
```

austinbyers (commented):
Great start! The Go code looks good!

* Be sure to update the PR description, it's easiest to do this when you first create the PR. How did you test this change? Did you run the integration tests? Did you deploy and try with and without the filter?
* `test:go` is failing - you have a line too long
* `test:web` is failing on `src/pages/ListRules/ListRules.test.tsx` I'm not quite sure how to fix that one, perhaps @3nvi can point us in the right direction tomorrow

3nvi:
> Great start! The Go code looks good!
> 
> * Be sure to update the PR description, it's easiest to do this when you first create the PR. How did you test this change? Did you run the integration tests? Did you deploy and try with and without the filter?
> * `test:go` is failing - you have a line too long
> * `test:web` is failing on `src/pages/ListRules/ListRules.test.tsx` I'm not quite sure how to fix that one, perhaps @3nvi can point us in the right direction tomorrow

The fact that they are failing is what makes me sleep better at night 😄 You changed the `schema.graphql` so the CI rebuilt our fixture builders and TS typings to reflect the new changes. This caused our tests to fail cause with the new builders what we send to the API doesn't match what the API expects (i.e. we send a single severity, but the API expects a list of severities).

Solutions:

1. Omit the GraphQL change and have a "broken" master until FE makes the relevant changes
2. Include the FE effort in the same PR

I'll go ahead with (2) since the related changes should be really small

3nvi:
A side-effect of this is that now policy filtering needs to have the same changes (since policy & filters share the BE code). I'll go ahead and make those changes as well

internal/core/analysis_api/handlers/list.go
```diff
@@ -97,9 +103,12 @@ func pythonListFilters(enabled *bool, nameContains, severity string, types, tags
 		filters = append(filters, typeFilter)
 	}
 
-	if severity != "" {
-		filters = append(filters, expression.Equal(
-			expression.Name("severity"), expression.Value(severity)))
+	if len(severity) > 0 {
+		severityFilter := expression.AttributeNotExists(expression.Name("severity"))
+		for _, severityType := range severity {
+			severityFilter = severityFilter.Or(expression.Contains(expression.Name("severity"), string(severityType)))
```
austinbyers, line 109: Should be `expression.Equals` for severity, not `expression.Contains`

internal/core/analysis_api/handlers/list.go
```diff
@@ -97,9 +103,12 @@ func pythonListFilters(enabled *bool, nameContains, severity string, types, tags
...
+		}
+		filters = append(filters, severityFilter)
```
austinbyers, line 111: If we don't want to do the first expression hack, you can do it like this:

```suggestion
	if len(severity) > 0 {
		severityFilter := expression.Equal(expression.Name("severity"), string(severity[0]))
		for _, severityType := range severity[1:] {
			severityFilter = severityFilter.Or(expression.Equal(expression.Name("severity"), string(severityType)))
		}
		filters = append(filters, severityFilter)
	}
}
```

And then go ahead and make this change for the other loop checks as well (e.g. resource types)

austinbyers:
Thanks so much @3nvi for pushing the frontend changes! So fast!

austinbyers:
> Solutions:
> Omit the GraphQL change and have a "broken" master until FE

@3nvi a broken master should *never* be an option because it will hinder other developers. We would not have merged this PR with failing web tests nor with a broken list rules page.

An explicit feature branch was not necessary here because the existing frontend single-severity filter worked with the API change, so all we had to do was fix the web test.

Of course, thank you again for contributing to the branch! It was a simple change that saved lots of time. Just wanted to make it clear that I agree merging broken changes to master is not acceptable

austinbyers (approved):
I did a fresh deployment and verified:

* Policy severity filtering worked for zero, one, or more options
* Rule severity filtering worked for zero, one, or more options
* Analysis-api integration tests passed

Great job!

3nvi:
> An explicit feature branch was not necessary here because the existing frontend single-severity filter worked with the API change, so all we had to do was fix the web test.

@austinbyers I did try it and it was failing for me. That's why I jumped in with the related changes. It may not be failing on the lambda side, but remember that GraphQL is hardly typed so if you pass in single severity while it expects a list, it will fail.

Some sort of FE changes were 100% needed to at least wrap with brackets `[]` the single severity passed to Appsync

PR #2177, opened 2020-12-02
datamodel single rule test

lindsey-w:
## Background

When adding DataModels to the Rules engine, I missed adding it to the single rule tests that are triggered by the UI in the rule editor.  This change is allow unit test runs from the UI to pass when the rule uses the `udm` method. 

## Changes

- Moved main logic for single rule test into the engine to take advantage of data models imported
- Added unit test for new `analyze_single_rule` method

## Testing

- `mage test:python`


internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -42,6 +42,53 @@ def __init__(self, analysis_api: AnalysisAPIClient) -> None:
...
+        #  via the `udm` method
+        if 'p_log_type' in event and event['p_log_type'] in self.log_type_to_data_models:
+            event = EnrichedEvent(event, self.log_type_to_data_models[event['p_log_type']])
+
+        rule_result = rule.run(event, batch_mode=False)
+        format_exception = lambda exc: '{}: {}'.format(type(exc).__name__, exc) if exc else exc
+        # for tests against rules using the `udm` method, you must specify `p_log_type`
+        # field in each test definition
+        if isinstance(rule_result.rule_exception, AttributeError):
+            if str(rule_result.rule_exception) == '\'dict\' object has no attribute \'udm\'':
```
lindsey-w, line 63: I didn't know of a better way to check for this, open to other suggestions 🙏🏻 

I considered putting this check within the `Rule.run()` (if `udm` is called, ensure `p_log_type` is in the event) - but in normal cases (non-user-defined events), `p_log_type` will _always_ be in the event, so I added logic here instead.

internal/log_analysis/rules_engine/src/data_model.py
```diff
@@ -59,29 +59,29 @@ def __init__(self, config: Dict[str, Any]):
 
         # body is optional in a data model
         self.body = ''
+        self._module = None
         if 'body' in config:
             if not isinstance(config.get('body'), str):
                 raise AssertionError('Field "body" of type str')
             self.body = config['body']
+            self._store_data_models()
+            self._module = self._import_data_model_as_module()
```
lindsey-w, line 68: Figured I'd move this in here for the data models that might not have python logic, save a bit of time.

austinbyers (commented):
Looks good to me! But I'll defer final approval to a member of the data processing team

internal/log_analysis/rules_engine/src/data_model.py
```diff
@@ -59,29 +59,29 @@ def __init__(self, config: Dict[str, Any]):
...
-        self._store_data_models()
-        self._module = self._import_data_model_as_module()
         self._extract_mappings(config['mappings'])
 
     def _extract_mappings(self, source_mappings: List[Dict[str, str]]) -> None:
         for mapping in source_mappings:
             if NAME not in mapping:
                 raise AssertionError('DataModel [{}] is missing required field: [{}]'.format(self.data_model_id, NAME))
-            if PATH in mapping:
+            if mapping.get(PATH):
```
lindsey-w, line 79: Turns out without this check (that the key exists _and_ the value is not-None, rather than just that the key exists), the rules engine will alarm with any data model that has a `method` defined. Not sure why I didn't catch this before

internal/log_analysis/rules_engine/src/engine.py
nhakmiller: I agree this is the correct place for this, the rules engine tends to run hot as far as our lambdas go so keeping unneeded logic out of the critical path of evaluating events is a good thing.

I believe you can do the check like this:
```
MISSING_UDM_EXCEPTION = AttributeError("'dict' object has not attribute 'udm'")

< rest of the code>

if type(rule_result.rule_exception) is type(MISSING_UDM_EXCEPTION) and rule_result.rule_exception.args == MISSING_UDM_EXCEPTION.args:
```

Not sure if this is "better" than what you have, but it's another option at least.

nhakmiller (approved):
LGTM, good catch on this one.

Issue #2176, opened 2020-12-02
Rename the ddb-athena connector


rleighton:
### Description

The ddb Athena connector is implemented as a lambda. Current the name is the same as the catalog name space but we need it prefixed with 'panther-' for teardown to work.

Change name to `panther-ddb-tables`

PR #2186, opened 2020-12-03
Rename Athena ddb connector namespace and lambda

rleighton:
## Background

The ddb Athena connector is implemented as a lambda. The name must be the same as the catalog name space but we need it prefixed with 'panther-' for teardown to work.

This PR changes the name from `ddb` to `panther-ddb-tables`

Closes: https://github.com/panther-labs/panther/issues/2176

## Changes

- Renamed the Athena DDB source group and lambda (which need to have the same name)
- Renamed the connector related resources so they will be deleted and replaced
- Changed supported regions to match Athena2 deployments
- Removed the "Preview" workgroup since this feature is GA 

## Testing

- mage test:ci
- mage deploy + tested Athena queries using ddb tables
- mage teardown + confirmed lambdas deleted


austinbyers (approved):
Did you test teardown *with the deployment role*? The teardown always worked fine before, it only failed when you used the deployment role with it

I don't think you really have to test that right now, I just wanted to explain the problem was with the permissions on the deployment role and not the teardown in general

rleighton:
> Did you test teardown _with the deployment role_? The teardown always worked fine before, it only failed when you used the deployment role with it
> 
> I don't think you really have to test that right now, I just wanted to explain the problem was with the permissions on the deployment role and not the teardown in general

Ah, did not know that. I will test that and update the deployment role as needed.

rleighton:
> > Did you test teardown _with the deployment role_? The teardown always worked fine before, it only failed when you used the deployment role with it
> > I don't think you really have to test that right now, I just wanted to explain the problem was with the permissions on the deployment role and not the teardown in general
> 
> Ah, did not know that. I will test that and update the deployment role as needed.

Deploy and teardown using deployment role worked. No error and all lambdas removed.

PR #2187, opened 2020-12-03
Enabling Cloud Security to data lake 1/2

kostaspap:
## Background

This is 1/2 of remaining PRs to enable Cloud Security data to data lake. Decided to split into two since this PR can have some controversial changes. 

## Changes

- Creating new Cloud Security tables in Glue data catalog
- Updating Cloud Security tables upon deployment

## Testing

- mage test:ci
- Deployed and verify that tables got created: 

Screen Shot 2020-12-03 at 5 40 21 PM

DDB - Glue tables still there: 
Screen Shot 2020-12-03 at 5 41 02 PM

##### NOTE #####

We always considered "Log Type" a unique identifier for data making it to the Data lake. The data was always going to  `panther_logs`, `panther_rule_matches`, `panther_rule_errors` tables. All the logtypes were inside the same namespace. Each log type has not extra context associated with it.

However, by introducing the CloudSecurity log types, this is no longer true. The Cloud Security log types are very different from the rest: 
1. They don't have corresponding Rule tables. The flow is `panther_logs` -> `panther_rule_matches`/`panther_rule_errors`. `panther_cloudsecurity` will NEVER end up in the `panther_rule_matches`. 
2. They should not be displayed in log types selection when someone decides to onboard some data. These log types are "internal".
3. They should not be available to rules. 

Right now, the assumption is that the Cloud Security log types will never have conflict with Log Analysis log types but this is not enforced in the code. It is however tech debt that we are putting here.


deployments/cloud_security.yml
```diff
@@ -109,16 +109,6 @@ Conditions:
...
-  ###### Update Glue Table Schemas for Deployed Tables #####
-  UpdateGlueTables:
-    Type: Custom::UpdateCloudSecurityTables
-    Properties:
-      # Update in case CustomResourceVersion has changed
-      CustomResourceVersion: !Ref CustomResourceVersion
-      ResourcesTableARN: !GetAtt ResourcesTable.Arn
-      ComplianceTableARN: !GetAtt ComplianceTable.Arn
-      ServiceToken: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-cfn-custom-resources
-
```
kostaspap, line 121: The reason this was removed: 
1. We want to be able to run the "Database sync" operation if we update the Cloud Security tables. 
2. Database sync operation is performed by the Data Catalog updater
3. The data catalog updater is deployed in the Log Analysis stack
4. This would need to have a depending on the Data Catalog updater ...

Instead, I moved all the Glue stuff management inside the `log_analysis.yml`

deployments/master.yml
```diff
@@ -336,7 +336,14 @@ Resources:
           Value: panther-cw-dashboards
 
   LogAnalysis:
-    DependsOn: Core
+    DependsOn:
+      - Core
+      - CloudSecurity # FIXME With 1.15 release, the management of `panther_cloudsecurity` database moved
```
kostaspap, line 341: @austinbyers This is required for anyone transitioning from <1.15 to 1.15.... 
In case the race condition described occurs, there will be not `panther_cloudsecurity` database so the `panther-data-catalog-updater` will be throwing errors. 

This is fixed with a redeployment, but didn't want to force everyone deploying from master (non-developers) to redeploy. instead opted to make deployment slower.

internal/compliance/awsglue/awsglue.go
```diff
@@ -54,23 +54,6 @@ var (
 	}
 )
 
-func CreateOrUpdateCloudSecurityDatabase(glueClient glueiface.GlueAPI) error {
-	dbInput := &glue.DatabaseInput{
-		Description: aws.String(pantherdb.CloudSecurityDatabaseDescription),
-		LocationUri: aws.String("dynamo-db-flag"),
```
kostaspap, line 60: The `locationUri` was actually not required.

internal/core/source_api/api/put_integration_test.go
```diff
@@ -120,11 +120,16 @@ func TestAddToSnapshotQueue(t *testing.T) {
 
 func TestPutCloudSecIntegration(t *testing.T) {
 	mockSQS := &testutils.SqsMock{}
-	mockSQS.On("SendMessageBatch", mock.Anything).Return(&sqs.SendMessageBatchOutput{}, nil) // count is hard to get due to batching
 	sqsClient = mockSQS
 	dynamoClient = &ddb.DDB{Client: &modelstest.MockDDBClient{TestErr: false}, TableName: "test"}
 	evaluateIntegrationFunc = func(_ API, _ *models.CheckIntegrationInput) (string, bool, error) { return "", true, nil }
 
+	// Message sent to create Cloud Security tables
+	// TODO Verify payload
```
kostaspap, line 128: I started to fix this unit test, but it was going the wrong way - we don't care about mocking the SQS call, we care about mocking the call to the datacatalog *client* (which uses an SQS client underneath). 

This ended up being much more work and didn't want to increase the scope of this PR.

deployments/master.yml
austinbyers: > instead opted to make deployment slower.

That is very unfortunate, because it will easily add 3-10 minutes per deploy, but I agree a correct deployment is better than a fast one. Is it possible to instead move the `panther_cloudsecurity` database into one of the earlier stacks, like `bootstrap` or `bootstrap-gateway`? That way the main stacks can still run in parallel.

If not, we can remove this depends on after a few releases, then?

You will need to make the same change in `mage deploy` - to fix the dependency ordering so that the log-analysis stack isn't deployed until cloud security is done. (Once we consolidate deployment models, this wouldn't be necessary)

deployments/cloud_security.yml
rleighton: Ok

internal/compliance/awsglue/awsglue.go
rleighton: It used to be for listing in Athena console but I see above that I guess they changed that.

internal/core/custom_resources/resources/glue_tables.go
```diff
@@ -97,3 +104,22 @@ func customUpdateLogProcessorTables(ctx context.Context, event cfn.Event) (strin
...
 }
+
+func createCloudSecurityDDBTables(_ context.Context) error {
+	tableArn := arn.ARN{
+		Partition: "aws",
+		Region:    aws.StringValue(awsSession.Config.Region),
+		AccountID: env.AccountID,
+		Service:   "dynamodb",
+		Resource:  cloudsecglue.ResourcesTableDDB,
+	}
```
rleighton, line 115: nit: you could call `.String()` here and avoid 2 calls below

internal/core/source_api/api/list_logtypes_test.go
```diff
@@ -41,12 +41,6 @@ func TestAPI_ListLogTypes(t *testing.T) {
 				LogTypes:        []string{"one", "two"}, // "one" is duplicate with above
 			},
 		},
-		{ // should not match
```
rleighton, line 44: Curious, why this was removed.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
```diff
@@ -60,20 +60,29 @@ func (h *LambdaHandler) createOrUpdateTablesForLogTypes(ctx context.Context, log
...
-	deployedLogTables, err := resolveLogTables(ctx, h.Resolver, deployedLogTypes...)
+	deployedLogTables, err := resolveTables(ctx, h.Resolver, deployedLogTypes...)
 	if err != nil {
 		return errors.Wrap(err, "failed to resolve tables for logtypes")
 	}
+
+	var tablesInView []*awsglue.GlueTableMetadata
+	for _, table := range deployedLogTables {
+		if table.DatabaseName() == pantherdb.CloudSecurityDatabase {
+			// Don't create views for the Cloud Security tables
```
rleighton, line 78: Should this only be for the ddb tables? I think we want to search over the cloudsec changes tables in indicator search, right?

internal/log_analysis/datacatalog_updater/datacatalog/sync_database_partitions.go
```diff
@@ -50,18 +50,26 @@ func (h *LambdaHandler) HandleSyncDatabasePartitionsEvent(ctx context.Context, e
 	)
 	var tableEvents []*SyncTableEvent
 	for _, dbName := range event.DatabaseNames {
-		// Tables in panther_logs database can have partitions at any point in time.
-		// The rest can only have partitions in the range TableCreateTime <= PartitionTime < now
-		afterTableCreateTime := dbName != pantherdb.LogProcessingDatabase
 		for _, logType := range event.LogTypes {
+			dataType := pantherdb.GetDataType(logType)
+			if dataType == pantherdb.CloudSecurity && dbName != pantherdb.CloudSecurityDatabase {
```
rleighton, line 55: Which tables are these?

internal/core/custom_resources/resources/glue_tables.go
kostaspap: The first `.String()` will generate the ARN of the `cloudsecglue.ResourcesTableDDB` table, the second `.String()` for the `cloudsecglue.ComplianceTableDDB` table... I don't think I can avoid it.

I updated the code though to make the operation more clear.

internal/core/source_api/api/list_logtypes_test.go
kostaspap: Now a compliance source onboarding also triggers the creation of the Glue tables ^^ So the `//should not match` was no longer accurate.

internal/log_analysis/datacatalog_updater/datacatalog/sync_database_partitions.go
kostaspap: So here the code is looping for all Log Types both "log-related" but also "compliance related". imagine "AWS.CloudTrail", "AWS.VpcFlow", etc... but also "Snapshot.ResourceHistory", "Snapshot.ComplianceHistory" . 
It is also looping for all databases, `panther_logs`, `panther_rule_matches`, ... but also `panther_compliance`. 

We should not: 
1. Create Cloud Security tables in `panther_logs` or `panther_rules` DBses
2. Create Log-tables inside `panther_cloudsecurity`. 

I have added a comment with examples.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
kostaspap: Well, that's a good question. In general we want to be able to search for an indicator in the infrastructure, but added this here will not help with that, since the cloudsecurity tables have rows for every change that happened in the infrastructure. They cannot display the state of the infrastructure in a specific moment. 

e.g. let's say someone wants to search for an IP indicator in Panther. They pick a time range (2020-12-07T00:00:00Z -> 2020-12-07T12:00:00Z) and click "Run". 

The results may or may not include the resources (e.g. EC2 instances) that have that indicator depending if a change happened in that resources in the given time frame. I don't want to give the users the false impression that putting an IP in the indicator search will return them all the infrastructure that had that indicator in the given timeframe because that's not what we do.

deployments/master.yml
kostaspap: I actually took another look at it - I was mistaken! I removed this condition since it is not needed.

My main concern here is that there was a following race condition between two operations: 
1. When I removed [this](https://github.com/panther-labs/panther/blob/5adcca3000a76bfba6daed21ba6500f274fa0a9f/deployments/cloud_security.yml#L114), I would trigger [this code](https://github.com/panther-labs/panther/blob/5adcca3000a76bfba6daed21ba6500f274fa0a9f/internal/core/custom_resources/resources/cloud_security_tables.go#L56-L64) to run. This would try to *delete* the `panther_cloudsecurity` database.
2. [This](https://github.com/panther-labs/panther/blob/working-on-cloud-security/internal/core/custom_resources/resources/glue_tables.go#L63-L67) code running on `log_analysis.yml` stack update, would try to *create* the `panther_cloudsecurity` database. 

Since `cloud_security.yml` and `log_analysis.yml` are deployed in parallel, I was under the impression that there was such a race condition.

I was mistaken! The Custom Resources Lambda gets deployed first, before both of these stacks. So #1 from above is not going to happen - the Custom Resource that was managing the `panther_cloudsecurity` tables before this release no longer exists - so it will not try to delete that database. 

False alarm :)

alxarch:
> 1. `panther_cloudsecurity` will NEVER end up in the `panther_rule_matches`.

Are we sure about this? I was under the impression that @nhakmiller was considering the ability to add Rules for these.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
```diff
@@ -87,7 +96,7 @@ func (h *LambdaHandler) fetchAllDeployedLogTypes(ctx context.Context) ([]string,
 	return gluetables.DeployedLogTypes(ctx, h.GlueClient, available)
 }
 
-func resolveLogTables(ctx context.Context, r logtypes.Resolver, names ...string) ([]*awsglue.GlueTableMetadata, error) {
+func resolveTables(ctx context.Context, r logtypes.Resolver, names ...string) ([]*awsglue.GlueTableMetadata, error) {
```
alxarch, line 99: The reason this was called 'LogTables' instead of 'Tables' is that we have 'magic' expansion for the glue table metadata.
I have given up hope on ever managing to properly structure any code that involves `awsglue`. Just add comments explaining that these are the 'base' tables and not the 'expanded' tables that include Rule Errors and Rule Matches.

internal/log_analysis/datacatalog_updater/datacatalog/sync_database_partitions.go
```diff
@@ -49,19 +49,30 @@ func (h *LambdaHandler) HandleSyncDatabasePartitionsEvent(ctx context.Context, e
...
-		// Tables in panther_logs database can have partitions at any point in time.
-		// The rest can only have partitions in the range TableCreateTime <= PartitionTime < now
-		afterTableCreateTime := dbName != pantherdb.LogProcessingDatabase
-		for _, logType := range event.LogTypes {
-			tblName := pantherdb.TableName(logType)
+	for _, logType := range event.LogTypes {
+		dataType := pantherdb.GetDataType(logType)
+		tblName := pantherdb.TableName(logType)
+		for _, dbName := range event.DatabaseNames {
+			if dataType == pantherdb.CloudSecurity && dbName != pantherdb.CloudSecurityDatabase {
```
alxarch, line 56: This part is exactly the reason we need to centralize this somehow. These `if` cases will be scattered all over the code for datacatalog updater and probably elsewhere.

alxarch: We need a way to go from a single log type to all required `awsglue.GlueTableMetadata`

alxarch: Conceptually a logtype requires some tables to work properly. In the future we will be addin more tables such as classification errors etc. Those would be needed for both cloud security and normal logs

alxarch: The way I see this after the introduction of `pantherdb` module is that such information should be coming from there.

alxarch (commented):
I think we need to pause and rethink how we manage tables for logtypes. I have found similar problems while investigating custom log updates. There is no concrete structure or single point of truth for managing the data catalog.

deployments/master.yml
austinbyers: Yay!

deployments/master.yml
```diff
@@ -348,7 +348,8 @@ Resources:
           Value: panther-cw-dashboards
 
   LogAnalysis:
-    DependsOn: Core
+    DependsOn:
+      - Core
```
austinbyers, line 352: Nit: usually we prefer singleton lists to be collapsed; the way it was before was fine (I think this is just leftover from your earlier change that proved to be unnecessary)

internal/core/custom_resources/resources/glue_tables.go
```diff
@@ -97,3 +104,29 @@ func customUpdateLogProcessorTables(ctx context.Context, event cfn.Event) (strin
 		return "", nil, fmt.Errorf("unknown request type %s", event.RequestType)
 	}
 }
+
+func createCloudSecurityDDBTables(_ context.Context) error {
+	resourcesTableArn := arn.ARN{
+		Partition: "aws",
```
austinbyers, line 110: Since we're going to work on supporting multiple partitions, can we make sure new code is partition aware? If there's an easy way to add the correct partition instead of hardcoding, can we do that now? If there's not an easy way to get the partition from the current session, no worries we can add it later

deployments/cloud_security.yml
austinbyers: I don't see any changes to the `log_analysis.yml` stack?

kostaspap: > I don't see any changes to the log_analysis.yml stack in this PR?

These changes now are managed inside the `UpdateLogProcessorGlueTables` custom resource in the `log_analysis.yml`. No change in the template - the change just happened in the custom resource code :)

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
kostaspap: will add comment

deployments/master.yml
kostaspap: Right! Will update (revert)

internal/log_analysis/datacatalog_updater/datacatalog/sync_database_partitions.go
kostaspap: Good call. Updating

internal/core/custom_resources/resources/glue_tables.go
kostaspap: I was able to get it using the endpoint resolver. Updated!

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
giorgosp: You might add a summary of this comment in the code, as to why we don't create views for the Cloud Security tables. If I understood correctly, an one liner like "infra is mutable and historical logs may not accurately show the current picture" would be enough

giorgosp (approved):

rleighton:
Let’s talk about the views. Using indicator search on a resource during
incident response is a major use case for this feature .

internal/core/custom_resources/resources/glue_tables.go
```diff
@@ -56,31 +59,35 @@ func customUpdateLogProcessorTables(ctx context.Context, event cfn.Event) (strin
...
 
-		requiredLogTypes, err := apifunctions.ListLogTypes(ctx, lambdaClient)
+		if err := createCloudSecurityDDBTables(ctx); err != nil {
+			return physicalResourceID, nil, err
+		}
+
+		logTypesInUse, err := apifunctions.ListLogTypes(ctx, lambdaClient)
 		if err != nil {
-			logger.Error("failed to fetch required log types", zap.Error(err))
+			logger.Error("failed to fetch log types in use", zap.Error(err))
```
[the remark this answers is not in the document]
kostaspap: It will be returned to the Lambda caller and eventually be logged there. Removed

internal/log_analysis/pantherdb/pantherdb.go
```diff
@@ -74,6 +75,19 @@ func GetDataType(logtype string) DataType {
 	return LogData
 }
 
+// Returns true if the given logtype has a corresponding table in a Database
+func IsInDatabase(logtype, db string) bool {
```
alxarch, line 79: This tries to fix things but we will be addressing the dependent tables and such stuff in an abstraction. Pending an RFC for it, this should suffice for now.

alxarch (approved):

PR #2189, opened 2020-12-03
fix: date color on generic item card

3nvi:
## Background

The Color of the date on the Cards is not correct, but the correct one is not in our palette. After speaking with @dimitriszoitas we agreed to use `navyblue-200`

### Before

Screen Shot 2020-12-03 at 6 09 44 PM

### After

Screen Shot 2020-12-03 at 6 09 54 PM

## Changes

- Change color of "date" on cards to `navyblue-200`

## Testing

- `npm run test`


vorillaz (approved):
👌

PR #2192, opened 2020-12-04
Fix version bugs

austinbyers:
## Background

#1785 created a shared VERSION file in an attempt to consolidate all version references to the same source of truth. However, the final version string has 4 parts: "(VERSION file)-(branch)-(commit)-(dirty)". This meant services which only needed part of the version had to do string splitting to get just the base version at the beginning. The problem is that when there is a "-" in the VERSION, e.g. "v1.14.0-RC", the string splitting logic no longer works. #2158 attempted to fix this parsing with regexes, but that's clearly just a hack to work around an underlying problem that needs to be addressed.

The main bug affecting the last two releases is that the S3 aux templates were not loading correctly during source onboarding (#2149, https://github.com/panther-labs/panther-enterprise/issues/1594, https://github.com/panther-labs/panther-enterprise/issues/1590):
* The `source-api` was using the CUSTOM_RESOURCE_VERSION instead of the true panther version as the basis of the template URL.
* The frontend was doing a string split to pull out the base version, which discarded the `-RC` suffix we use for release testing
* Worse still, `mage deploy` and `mage master:deploy` were passing the repo version differently and almost led to a completely broken 1.13 release publish

----
Now, instead of passing a single string which has to be split later, we pass the component version pieces and only combine them at the very end when they are needed. I've also simplified the parts: we only care about the VERSION base ("1.14.0") and the commit sha ("12345678").

Those two parts are now passed directly to every service that needs a version. There is no "v" prefix anymore; we add that when the version is actually used (e.g. to generate a template URL)

Closes: #2158 (this is a more holistic solution)

## Changes

- `source-api` is passed the contents of the VERSION file, not the CustomResourceVersion. The VERSION file does not have a "v" prefix, so it has been added when the source-api generates the URL in the Go code.
- "PantherVersion" and "PantherCommit" parameters in CloudFormation now have regex validation to ensure they always have the correct format. PantherVersion looks like "1.X.Y" or "1.X.Y-RC" (no "v" prefix) and PantherCommit is the 8-digit short commit tag
- The default CustomResourceVersion is now the same as the version string shown to the user in the general settings page: VERSION-COMMIT, e.g. "1.14.0-RC (12345678)"
    - This way, every deployment will re-trigger the custom resources if there are new commits. This is helpful during release testing, since we may republish many different `-RC` variants and need to re-apply the changes
    - The same custom resource version is used for both deployment methods
- The commit and the semantic version are passed separately to the web stack. The "v" prefix has been dropped when showing it to the user
- We really only care about the branch and the dirty repo state when we're publishing a release. So those have been removed from the full version string and are instead checked directly during `mage master:publish`
- Fix environment when running web server locally (`npm start`)

### Versions Before
General Settings
Screen Shot 2020-12-03 at 5 41 41 PM

Profile Popup
Screen Shot 2020-12-03 at 5 41 53 PM

(from our 1.14 release environment)

### Versions After

Screen Shot 2020-12-03 at 6 07 46 PM
Screen Shot 2020-12-03 at 6 07 32 PM

Note the simplified versions, and that the profile now shows the full base version (with the -RC suffix)

## Testing

- `mage deploy` and `npm start`
    - verified the version in the profile popup, the general settings page, and the CloudFormation tags
    - verified log analysis and cloud security source onboarding can generate templates correctly (and are loading the correct template versions)
- Ran all the same tests again for `mage master:deploy`


austinbyers:
This is too big a change for me to be comfortable syncing to the release. We've already tested 1.14.0-RC with the hardcoded version fixes, let's just leave it that way so we don't have to re-test

tools/mage/util/version.go
```diff
@@ -20,121 +20,34 @@ package util
...
 	}
-
-	return gitVersion
+	return semver
 }
 
-// Find the most recent published version of Panther in S3, e.g. "1.7.1"
-//
-// This provides an alternative to checking the git tags in the repo.
-func LatestPublishedVersion() (string, error) {
```
austinbyers, line 81: This is a chunk of code that has never been used; it was originally intended for end-to-end release testing

3nvi (approved):
Amazing work @austinbyers . That was much needed since regex parsing was indeed fragile. I'm always amazed how you're able to make perfect FE changes; you've even updated the snapshots. Props 🎉 

Approving the web changes

rleighton (approved):
LGTM. Nice cleanup. Splitting the commit and the version makes this much more maintainable.

alexmylonas:
Confirmed functionality locally!

tools/mage/util/version.go
nhakmiller: The Destroyer of Code living up to his epithet.

nhakmiller (approved):

tools/mage/util/version.go
austinbyers: It's a conspiracy: I just add junk code so I can look good for removing it later

austinbyers:
> This is too big a change for me to be comfortable syncing to the release. We've already tested 1.14.0-RC with the hardcoded version fixes, let's just leave it that way so we don't have to re-test

On second thought, the published template itself is subject to a similar version problem (`mage master:publish` will publish v1.14.0 instead of v1.14.0-RC), so we are going to sync this to the release after all. It has already been tested extensively, and we will continue to test it as we upgrade accounts.

I'm more worried about lingering version bugs than the effects of this change

deployments/core.yml
```diff
@@ -76,6 +76,10 @@ Parameters:
     Description: KMS key for encrypting alert outputs
     # Example: "484fb80c-4ae5-40d0-b22a-bdd5d0953b3e"
     AllowedPattern: '^[0-9a-f-]{36}$'
+  PantherVersion:
+    Type: String
+    Description: The base semantic version of the current deployment (e.g. `1.3.0`)
+    AllowedPattern: '^1\.\d+\.\d+(-.+)?$'
```
s0l0ist, line 82: I don't like this hard match for `1`, lets just make this another `\d+`

deployments/web_server.yml
```diff
@@ -82,10 +82,14 @@ Parameters:
...
+  PantherCommit:
+    Type: String
+    Description: The commit of the current deployment (e.g. `5c2a8a76`)
+    AllowedPattern: '^[0-9a-f]{8}$'
   PantherVersion:
     Type: String
-    Description: The semver of the current deployment (i.e. `v1.3.0`)
-    MinLength: 1
+    Description: The base semantic version of the current deployment (e.g. `1.3.0`)
+    AllowedPattern: '^1\.\d+\.\d+(-.+)?$'
```
s0l0ist, line 92: same as above here

tools/mage/deploy/deploy.go
```diff
@@ -583,7 +584,7 @@ func customResourceVersion() string {
 		return v
 	}
 
-	// By default, just use the major release version so developers do not have to trigger every
-	// custom resource on every deploy.
-	return strings.Split(util.RepoVersion(), "-")[0]
+	// This is the same format as the version shown in the general settings page,
+	// and also the same format used by the master stack.
+	return fmt.Sprintf("%s (%s)", util.Semver(), util.CommitSha())
```
s0l0ist, line 589: nice

tools/mage/master/publish.go
```diff
@@ -44,15 +45,25 @@ func Publish() error {
...
+	}
+
+	// Only allow publishing from the release branch
+	// "git branch --show-current" works only for Git 2.22+, but this is compatible with older versions:
+	branch, err := sh.Output("git", "rev-parse", "--abbrev-ref", "HEAD")
+	if err != nil {
+		return fmt.Errorf("failed to get name of current branch: %s", err)
 	}
-	if !strings.Contains(version, "-release-1.") {
+	if !strings.HasPrefix(branch, "release-1.") {
```
s0l0ist, line 62: Can we make this less about a major version `1`? It will eventually break... Maybe some other regex match?

web/src/components/Navigation/ProfileIcon/ProfileIcon.tsx
```diff
@@ -132,7 +132,7 @@ const ProfileIcon: React.FC = () => {
                   nativeHeight={16}
                   mr={2}
                 />
-                Panther&nbsp;<b>{STABLE_PANTHER_VERSION}</b>
+                Panther&nbsp;<b>{pantherConfig.PANTHER_VERSION}</b>
```
s0l0ist, line 135: We should not pull from `Source/config` directly. Instead, we should make a const inside `Source/constants` so we have a single source of truth.

web/src/components/wizards/ComplianceSourceWizard/StackDeploymentPanel/StackDeployment.tsx
```diff
@@ -48,7 +47,7 @@ const StackDeployment: React.FC = () => {
   const { stackName, body } = data?.getComplianceIntegrationTemplate ?? {};
   const cfnConsoleLink =
     `https://${pantherConfig.AWS_REGION}.console.aws.amazon.com/cloudformation/home?region=${pantherConfig.AWS_REGION}#/stacks/create/review` +
-    `?templateURL=https://s3-us-west-2.amazonaws.com/panther-public-cloudformation-templates/panther-cloudsec-iam/${STABLE_PANTHER_VERSION}/template.yml` +
+    `?templateURL=https://s3-us-west-2.amazonaws.com/panther-public-cloudformation-templates/panther-cloudsec-iam/v${pantherConfig.PANTHER_VERSION}/template.yml` +
```
s0l0ist, line 50: Pull from `Source/constants`

web/src/components/wizards/S3LogSourceWizard/StackDeploymentPanel/StackDeployment.tsx
```diff
@@ -50,7 +49,7 @@ const StackDeployment: React.FC = () => {
   const { stackName, body } = data?.getS3LogIntegrationTemplate ?? {};
   const cfnConsoleLink =
     `https://${pantherConfig.AWS_REGION}.console.aws.amazon.com/cloudformation/home?region=${pantherConfig.AWS_REGION}#/stacks/create/review` +
-    `?templateURL=https://panther-public-cloudformation-templates.s3-us-west-2.amazonaws.com/panther-log-analysis-iam/${STABLE_PANTHER_VERSION}/template.yml` +
+    `?templateURL=https://panther-public-cloudformation-templates.s3-us-west-2.amazonaws.com/panther-log-analysis-iam/v${pantherConfig.PANTHER_VERSION}/template.yml` +
```
s0l0ist, line 52: Pull from `Source/constants`

web/src/helpers/__tests__/analytics.test.tsx
```diff
@@ -21,11 +21,10 @@ import React from 'react';
 import { Button } from 'pouncejs';
 import mixpanel from 'mixpanel-browser';
 import * as Sentry from '@sentry/browser';
-
+import { pantherConfig } from 'Source/config';
```
s0l0ist, line 24: Pull from `Source/constants` and for the other changes in this file

web/src/helpers/analytics.ts
```diff
@@ -18,7 +18,8 @@
 
 import { DestinationTypeEnum } from 'Generated/schema';
 import storage from 'Helpers/storage';
-import { ANALYTICS_CONSENT_STORAGE_KEY, STABLE_PANTHER_VERSION } from 'Source/constants';
+import { pantherConfig } from 'Source/config';
```
s0l0ist, line 21: Pull from `Source/constants`

s0l0ist (changes requested):
Minor changes, overall really good! This will help a lot!

s0l0ist (approved):

deployments/core.yml
austinbyers: Sure

tools/mage/master/publish.go
austinbyers: I'll just limit to a prefix of `release-`

web/src/components/Navigation/ProfileIcon/ProfileIcon.tsx
austinbyers: I considered that, but:

* We already use `pantherConfig.AWS_REGION` and `pantherConfig.AWS_ACCOUNT_ID` throughout the web source
* It seems redundant to `export const PANTHER_VERSION = pantherConfig.PANTHER_VERSION` in the constants file

I'll leave this as is for now since @3nvi approved it as-is and it's consistent with the existent usage, but I'm happy to change this later if we want a different style

austinbyers:
Deployed again, found a rare bug where the commit was being interpreted by CloudFormation as scientific notation 😅 

Bug fixed, verified version strings are still correct

PR #2194, opened 2020-12-04
CLI tool that infers Custom Logs schema out of samples

kostaspap:
## Background

Closes https://app.zenhub.com/workspaces/panther-5e4f99c4fbfd0c5803686679/issues/panther-labs/panther-enterprise/1348
Script that infers the custom logs schema out of a data sample. It only works for JSON data. I am using the same CLI tool (`customlogs`) that we were using until now. But now the users can pass a flag:
- `./customlogs validate` that validates a schema against some logs
- `./customlogs infer` that infers the schema from some logs

Capabilities: 
1. Can infer if a field is required or optional. If a field appears in all sample logs, it is marked as `required`, otherwise `optional`. 
1.  Identifies RFC3339 timestamps

Limitations: 
1. Doesn't mark any timestamp field as `isEventTime`. Users have to do that
1. Doesn't add any indicators
1. All integers are marked as `bigint`. Went for the safe path vs trying to optimize sizes.
1. All floats are marked as `float64`. Went for the safe path vs trying to optimize sizes.

## Changes
- Split the existing `customlogs` main logic to two files, `validate.go` and `infer.go`
- The way the inference works is: 
  1. Loads JSON data line by line
  1. Creates a schema out of this sample
  1. Merges the schema with a "global" schema ("global" -> for all log lines)

## Testing

- Ran it against quite some sample logs that I had collected. Created unit tests for some edge-cases I encountered

@alxarch In the beginning I had created a registry of fields (with full path) -> inferred types. The idea was to keep a slice of all the inferred types and try to reduce them to 1 type (or fail). Implementation turned out to be tricky... I went instead for the pattern you will see in this PR (merging schema line-by-line).


cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+
+	marshalled, err := yaml.Marshal(schema)
+	if err != nil {
+		logger.Fatal("failed to marshal schema", zap.Error(err))
+	}
+	fmt.Println(string(marshalled))
+}
+
+func inferFromFile(logger *zap.SugaredLogger, file string) (logschema.Schema, error) {
```
giorgosp, line 78: Is the logger really needed as a param to every function? Adds a bit of noise

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,466 @@
...
+	schema := logschema.Schema{
+		Version: 0,
+	}
+	fd, err := os.Open(file)
+	if err != nil {
+		return schema, errors.Wrap(err, "failed to read input file")
+	}
+	defer fd.Close()
+
+	scanner := bufio.NewScanner(fd)
```
giorgosp, line 88: Maybe use `bufio.Reader` that has an auto-growing buffer? Scanner will error if a log line is too big

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,466 @@
...
+		// inferring the log schema of that single line
+		lineFields := inferFields(data)
+		// merging the schema of this line with the schema that we have generated from all lines until now
+		schema.Fields, err = mergeFields(schema.Fields, lineFields)
+		if err != nil {
+			return schema, errors.Wrapf(err, "failed while inferring schema from line [%d]", lineNum)
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		logger.Fatal("failed to read input file", zap.Error(err))
```
giorgosp, line 110: Since this function has an error return argument, why not return the error and let it be handled in one place higher in the stack?

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,466 @@
...
+
+		field := logschema.FieldSchema{
+			Name:        key,
+			Description: "The " + key,
+			Required:    true,
+			ValueSchema: *valueSchema,
+		}
+		out = append(out, field)
+	}
+	// sorts the fields by name. Useful so that re-runs of the same tool will generate the same schema
```
giorgosp, line 140: nice

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,466 @@
...
+			return nil, false
+		}
+		return &logschema.ValueSchema{
+			Type:    typ,
+			Element: valueSchema,
+		}, true
+	case logschema.TypeObject:
+		object := value.(map[string]interface{})
+		// If the object has no fields, don't try to infer its type.
+		// Just skip it.
```
giorgosp, line 173: Perhaps it's useful to map it as an uknown object? (Not sure if we support this). So that the existence of this empty object is depicted in the schema, although we don't know its fields yet.

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,466 @@
...
+	case bool:
+		return logschema.TypeBoolean, true
+	case string:
+		return logschema.TypeString, true
+	case json.Number:
+		value := value.(json.Number)
+		if _, err := value.Int64(); err != nil {
+			return logschema.TypeFloat, true
+		}
+		return logschema.TypeBigInt, true
```
giorgosp, line 205: I think we should describe numbers as float64 for a more accurate schema. Even if the field can be parsed as integer in 1 log line, we can't be sure it will be int for all similar events. float64 is also the default decoding for json.

giorgosp (approved):
Nice work! Left some comments

cmd/devtools/customlogs/customlogs/infer.go
kostaspap: Sure. Updating

cmd/devtools/customlogs/customlogs/infer.go
kostaspap: Ah yes. I missed that one. Removed.

cmd/devtools/customlogs/customlogs/infer.go
kostaspap: Actually it was not needed here, removed. 
It is now only present at the top-level function `func Infer(logger *zap.SugaredLogger, opts *InferOpts)`

cmd/devtools/customlogs/customlogs/infer.go
kostaspap: That's actually not possible... a field of type `object` requires some `fields` to be present. 
I had it like that initially but ran into the issue. The unit test tests for this specific case as well.

cmd/devtools/customlogs/customlogs/infer.go
giorgosp: It is not needed there too if we only use a logger initialized in main.

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -452,15 +458,23 @@ func validateSchema(schema logschema.Schema, file string) error {
...
 	}
 
-	scanner := bufio.NewScanner(fd)
-	for scanner.Scan() {
-		_, err := parser.ParseLog(scanner.Text())
+	reader := bufio.NewReader(fd)
+	for {
+		line, err := reader.ReadString('\n')
 		if err != nil {
+			if err == io.EOF {
```
giorgosp, line 465: line might still be populated even if EOF is returned (see ReadString docs)

cmd/devtools/customlogs/customlogs/infer.go
kostaspap: I have to say I am not a huge fan of this one... usually in the logs, if the field is of "float" type, even if the decimal part is 0, they will display it e.g. `1.0`. So it is easy to understand the type. The majority of the number fields tend to be integers as well, floats are usually just a small set e.g. duration, scores etc

cmd/devtools/customlogs/customlogs/infer.go
kostaspap: Good catch! Updated

cmd/devtools/customlogs/customlogs/infer.go
kostaspap: The `Infer` function needs to log messages though, so it needs a logger. The other option would be to make it global var and access it from there, I am thinking passing it to the function is more clear.

giorgosp: I think a global var is fine except if we want different logging config/behavior in different circumstances (like tests etc). This doesn't seem the case here so I'd just make it a global.

cmd/devtools/customlogs/customlogs/infer.go
giorgosp: Not a hard opinion, it would just be safer to use a float type. Users have the incentive to reduce their costs too so I would expect they review and optimise the types.

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,486 @@
...
+		return left, nil
+	}
+	if len(left) == 0 {
+		return right, nil
+	}
+
+	// Map from field name -> Field Schema
+	leftMap := make(map[string]logschema.FieldSchema)
+	for _, schema := range left {
+		if _, ok := leftMap[schema.Name]; ok {
```
alxarch, line 282: This could only happen in case the input JSON is invalid? This seems like something to be extracted into a separate function.

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,486 @@
...
+		merged, err := mergeArrayElementSchema(left.Element, right.Element)
+		if err != nil {
+			return left, err
+		}
+		left.Element = merged
+	}
+	return left, nil
+}
+
+func mergeArrayElementSchema(left, right *logschema.ValueSchema) (*logschema.ValueSchema, error) {
```
alxarch, line 375: This seems like an error in the flow of recursion. There shouldn't be a specific function for array elements.

My suggestion is that we have an `mergeValues(left, right *logschema.ValueSchema) (*logschema.ValueSchema, error)` *entry point*.
It should have a single `switch` on type and handle the cases from there. 
Field values and array elements should be re-entering into `mergeValues`. 
Otherwise, we will end up with duplicate code and bugs will be hard to control in various cases such as in the glue schema case.

For the 'root' object, we can be wrapping a `[]logschema.FieldSchema` into a 'fake'  `logschema.ValueSchema` of type `object`. 

Another thing to note is that we should be passing a `path []string` argument into the recursive functions so that errros can be helpful.

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,486 @@
...
+	if left.Type == right.Type {
+		switch left.Type {
+		case logschema.TypeObject:
```
alxarch, line 378: This switch is duplicated here and in mergeFieldSchema. This is the problem I am talking about above.

alxarch (changes requested):

cmd/devtools/customlogs/customlogs/infer.go
alxarch: A gist following a 'structured recursion' flow. The important part is how functions call each other so that we only have one type switch for infer and one for merge.
https://gist.github.com/alxarch/b3f30c68193f6bda1126ed89fb385f72

alxarch: I would suggest to add this as an exported function in `logschema` so it can be reused.

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,486 @@
...
+	}
+
+	if _, err := time.Parse(time.RFC3339, value); err == nil {
+		return &logschema.ValueSchema{
+			Type:       logschema.TypeTimestamp,
+			TimeFormat: "rfc3339",
+		}
+	}
+
+	return &logschema.ValueSchema{
```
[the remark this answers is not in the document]
kostaspap: That's a good idea, but we can add it later. This is the first iteration of the CLI tool, we can make further improvements later.

cmd/devtools/customlogs/customlogs/infer.go
kostaspap: Updated.

cmd/devtools/customlogs/customlogs/infer.go
kostaspap: Updated.

cmd/devtools/customlogs/customlogs/infer.go
kostaspap: Removed.

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,486 @@
...
+	}
+	switch typ {
+	case logschema.TypeArray:
+		array := value.([]interface{})
+		// If the array has no elements, don't try to infer their type
+		if len(array) == 0 {
+			return nil, false
+		}
+		// We are trying to infer the schema of the array only by looking at the first element
+		valueSchema, ok := inferValueSchema(array[0])
```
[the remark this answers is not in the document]
kostaspap: Fixed

cmd/devtools/customlogs/main.go
```diff
@@ -0,0 +1,80 @@
...
+	if len(os.Args) < 2 {
+		flag.Usage()
+		logger.Fatalf("Need to provide one command")
+	}
+
+	switch cmd := os.Args[1]; cmd {
+	case testCmd:
+		opstools.SetUsage(`-s SCHEMA_FILE [-o OUTPUT_FILE] [INPUT_FILES...]`)
+		opts := &customlogs.TestOpts{
+			Schema: flag.String("s", "", "File file"),
```
alxarch, line 58: Schema file?

cmd/devtools/customlogs/customlogs/infer.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		root = logschema.Merge(root, lineObject)
+	}
+
+	return root, nil
+}
+
+// Validates the schema. It generates a parser of the provided schema
+// and tries to parse the contents of the file.
+func validateSchema(valueSchema *logschema.ValueSchema, file string) error {
```
alxarch, line 127: I don't see any point in re-reading the file through. Parsing won't fail if the schema was infered correctly. It's entirely possible that the fields end up as just `json` types and we are just verifying that jsoniter works. My understanding of the process is that it is a two step thing. First create the schema, and then to test the schema, the user needs to see the results to be able to tell if the schema is fine.

alxarch (approved):

cmd/devtools/customlogs/main.go
kostaspap: ah! yes. Renaming gone wrong.

PR #2195, opened 2020-12-04
Fix: Empty charts & Buggy log scale

### LINKED ISSUE #2178: Log Analysis Overview - empty chart, glitch with logarithmic
### Describe the bug

The `Most Active Rules` doesn't show anything even if the data returned is empty. In fact, it doesn't show data at all when there is supposed to be data -- this is a bug with metrics.

When viewing charts in the **logarithmic** mode, the graph timeline is off by a little. By adjusting the bottom time line bar a tiny bit causes the graph to re-draw it correctly.

### Steps to reproduce

1. Go log analysis overview page
2. Click on `Most Active Rules` tab, see nothing
3. Click on any `Logarithmic` mode, then barely adjust the time window bottom bar, see it re-draw.

### Expected behavior

Should have a placeholder for empty data. It should also show data 

### Environment

- panther v1.14.0-RC

### Screenshots
image

log-analysis-graph-glitch-2
log-analysis-graph-glitch

3nvi:
## Background

When there are absolutely no data in the system, the `series` payload from CloudWatch comes back empty which causes our charting library to render a nothingness. This PR fixes that by identifying empty series and falling back to a default illustration instead.

In addition this PR fixes an issue where data seemed to be missing from the logarithmic scale. In reality this was caused by `0` values since `log(0) = Infinity`. There's no easy fix in that, so a "hack" has been applied to make sure that chart rendering is correct.

Closes #2178 

## Designs

https://app.abstract.com/projects/ab14d6c7-511b-4577-8de9-25ca9cd2918b/branches/master/commits/03d7eb4d679a0c917528fef6829de2720bd5c84b/files/d379a6e4-1ffd-4b9f-adc8-22e5463c4961/layers/F3E9F942-E15B-4488-BA04-CCE92F24CA61?mode=design&selected=root-2B4BD3E8-82A0-4FBE-AF4E-34E4D0B3995F&sha=03d7eb4d679a0c917528fef6829de2720bd5c84b

### Before

Screen Shot 2020-12-04 at 1 08 02 PM

### After

Screen Shot 2020-12-04 at 1 13 27 PM

####

### Before

Screen Shot 2020-12-04 at 1 24 47 PM

### After

Screen Shot 2020-12-04 at 1 25 09 PM

## Changes

- Add `NoDataFound` component & use it wherever we render a chart
- Modify `NoResultsFound` to match the `NoDataFound` API
- Change colors of `ResetButton` to match designs
- Modify all `0` to `0.0001` when log scale is applied

## Testing

- `npm run test`


web/src/components/charts/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -172,10 +174,14 @@ const TimeSeriesChart: React.FC<TimeSeriesChartProps> = ({
...
         },
         data: values
-          .map((v, i) => {
+          .map((value, index) => {
             return {
               name: label,
-              value: [data.timestamps[i], v, data.metadata ? data.metadata[i] : null],
+              value: [
+                data.timestamps[index],
+                value === 0 && scaleType === 'log' ? 0.0001 : value,
```
3nvi, line 182: This is a hack so that log scale doesn't bug out. Funnily, echarts parses 0.0001 as 0, so the tooltip (as seen in the screenshot of the PR's body) shows a value of 0 and not 0.0001 (which is what we wanted)

web/src/pages/LogAnalysisOverview/AlertsBySeverity/AlertsBySeverity.tsx
```diff
@@ -21,24 +21,29 @@ import { Box, Flex } from 'pouncejs';
...
+  const timeseriesData = React.useMemo(
     () => ({
       timestamps,
       series: series.map(serie => ({ ...serie, label: capitalize(serie.label.toLowerCase()) })),
     }),
     [series, timestamps]
   );
 
+  if (!timeseriesData.series.length) {
+    return <NoDataFound title="No alerts are present in the system" />;
```
3nvi, line 40: @nhakmiller to validate the value of `title` as a fallback text for when no metric series exist (see this file and the ones below)

vorillaz (approved):
Nicely done @3nvi , can we also add a simple snapshot test for the pages with the charts?

3nvi:
> Nicely done @3nvi , can we also add a simple snapshot test for the pages with the charts?

Snapshot will just render a canvas. All the info is within the canvas that doesn't benefit from snapshot testing :/

vorillaz:
Argh, you are right. I guess we can’t validate the actual rendering.

--
Theodore Vorillas, Software engineer
Website GitHub Twitter

web/src/components/charts/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -21,7 +21,7 @@ import ReactDOM from 'react-dom';
 import { Box, Flex, theme as Theme, ThemeProvider, useTheme } from 'pouncejs';
 import { formatTime, remToPx, capitalize } from 'Helpers/utils';
 import { FloatSeries, LongSeries } from 'Generated/schema';
-import { EChartOption, ECharts } from 'echarts';
+import type { EChartOption, ECharts } from 'echarts';
```
alexmylonas, line 24: I haven't seen this before. Is it used only when you import types right?

s0l0ist (approved):

web/src/components/charts/TimeSeriesChart/TimeSeriesChart.tsx
alexmylonas: I am not sure i would categorized as a bug. 

What will happen to this chart if you have initial data as `[10, 0, 10000]` (that can easily happen on other charts
Screenshot 2020-12-04 at 4 49 40 PM

web/src/components/charts/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -265,9 +271,8 @@ const TimeSeriesChart: React.FC<TimeSeriesChartProps> = ({
         splitArea: { show: false }, // remove the grid area
       },
       yAxis: {
-        type: scaleType as EChartOption.BasicComponents.CartesianAxis.Type,
+        type: scaleType,
         logBase: 10,
-        min: scaleType === 'log' ? 1 : 0,
```
alexmylonas, line 270: I still feel this is wrong, the value of logarithimc scale takes place when you have big numbers otherwise should (and has) no actual value

alexmylonas (commented):
Good work @3nvi! I have some concern regarding the changes for logarithmic scale.

Also, should we adjust `AlertSummary` component for no data as well?

web/src/components/charts/TimeSeriesChart/TimeSeriesChart.tsx
3nvi: yup. Introduced in TS 3.8 I think

web/src/components/charts/TimeSeriesChart/TimeSeriesChart.tsx
3nvi: ah! Sorry didn't understand that's the reason you had it. Will revert

web/src/components/charts/TimeSeriesChart/TimeSeriesChart.tsx
3nvi: It doesn't render anything at all when the value is `0` on a logarithmic scale (see screenshots). It's clearly a bug by the library since it should handle this internally. Math-wise it's expected as `log(0) =Infinity`.

Do you have a better suggestion here? I went all over the internet and that's the best thing I could find

web/src/pages/LogAnalysisOverview/MostActiveRules/MostActiveRules.tsx
```diff
@@ -58,14 +59,18 @@ const MostActiveRules: React.FC<MostActiveRulesProps> = ({ alertsByRuleID }) =>
...
+        <BarChart
+          gridPosition={gridPosition}
+          barGap={barGap}
+          barWidth={barWidth}
+          data={reversedData}
+          formatSeriesLabel={params => `${params.value} Alerts`}
+          alignment="horizontal"
+        />
+      ) : (
+        <NoDataFound title="Your logs haven't matched any rule yet" />
```
nhakmiller, line 72: Since this is the the `MostActiveRules` graph, might make sense to re-word so that rules are the focus (instead of the logs):

```
No rule activity yet
```

or 

```
No rule matches yet
```

Or something like that

nhakmiller (approved):
Thanks for adding this! Default titles look mostly good, one small optional suggestion. Also glad that logarithmic bug is getting addressed, was kind of awkward during demos.

3nvi:
> Thanks for adding this! Default titles look mostly good, one small optional suggestion. Also glad that logarithmic bug is getting addressed, was kind of awkward during demos.

Not fully  addressed. `0` values will still not show up (since `log10(0) = Infinity`), but everything else will

PR #2212, opened 2020-12-05
Revert to sequential publishing

austinbyers:
## Background

The v1.14 release adds publishing support for 15 different AWS regions! We tried to publish the regions in parallel, but both @s0l0ist and I ran into issues with "Content-MD5 mismatch" or "docker credential failure"

So, we have to revert to sequential region publishing. However, to compensate, I've added the ability to specify which regions you want to publish. For release testing, for example, we don't need to publish all 15 regions, which can take quite awhile.

## Changes

## Testing

- `mage master:publish`

> 17:49:15	INFO	[master:publish]	Publishing panther-community 1.14.0-RC to ca-central-1, ap-northeast-2, eu-north-1, eu-west-1, eu-west-2, us-west-1, ap-northeast-1, ap-southeast-1, ap-southeast-2, us-east-1, us-east-2, us-west-2, ap-south-1, eu-central-1, eu-west-3, sa-east-1

- `REGION=us-east-1,us-east-2,us-west-1,us-west-2 mage master:publish`

> 17:50:01	INFO	[master:publish]	Publishing panther-community 1.14.0-RC to us-east-1, us-east-2, us-west-1, us-west-2

A test publish is in progress (of the master branch with v1.14.0-RC, which I'll overwrite when it syncs to the release branch)

### DIFF
--- a/deployments/web_server.yml
+++ b/deployments/web_server.yml
@@ -85,7 +85,7 @@ Parameters:
   PantherCommit:
     Type: String
     Description: The commit of the current deployment (e.g. `5c2a8a76`)
-    AllowedPattern: '^[0-9a-f]{8}$'
+    AllowedPattern: '^[0-9a-f]+$'
   PantherVersion:
     Type: String
     Description: The base semantic version of the current deployment (e.g. `1.3.0`)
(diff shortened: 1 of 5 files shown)


nhakmiller (approved):
LGTM

austinbyers:
Turns out the "docker credential failure" was actually due to the ECR auth token request not being sent in the correct region. Still, the Content-MD5 errors mean we should stay sequential to be on the safe side for now

PR #2217, opened 2020-12-07
Allow policy engine to assume audit role

nhakmiller:
## Background

When a policy is evaluating a resource, sometimes the policy writer wants additional about that resource. Instead of just always delegating this to the snapshot-poller, which can be a pretty heavy change, if the policy writer could assume the `PantherAuditRole` on an ad hoc basis they could get arbitrary additional information about the resource being evaluated (or a different resource).

Since this is a pretty large amount of permissions to grant to the analysts writing policies, we decided to make this feature optional and disabled by default. For power customers  who want or need this feature, they can enable it.

## Changes

- Optionally allow `policy-engine` to assume `PantherAuditRole` in target accounts

## Testing

- Not tested yet, coordinating with @horakj to make sure it works


nhakmiller:
I haven't tested this yet (even tested deploying it) because I keep running out of time, but pushing this up here so Justin can test it.

horakj (approved):

deployments/cloud_security.yml
```diff
@@ -1165,6 +1170,13 @@ Resources:
               Resource:
                 - !Sub arn:${AWS::Partition}:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:panther-analysis/*
                 - !Sub arn:${AWS::Partition}:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:panther-policies/*
+        - Id: AssumePantherAuditRoles
+          Condition: AllowPolicyEngineScans
```
austinbyers, line 1174: I didn't know you could put a condition inside a policy statement like this, usually the `Condition` only applies to the top-level resource.

I'm not convinced this works, but if you've tested with this setting on and off and can confirm, then great! What you have is better.

But I suspect it needs to be something like:

```yaml
    - !If
      - AllowPolicyEngineScans
      - Id: AssumePantherAuditRoles # value if condition is true - IAM policy definition
        Version: 2012-10-17
        Statement: # etc
      - !Ref AWS::NoValue # value if condition is false - nothing (pretend the item isn't here)
```

deployments/cloud_security.yml
```diff
@@ -107,6 +111,7 @@ Conditions:
   AttachLayers: !Not [!Equals [!Join ['', !Ref LayerVersionArns], '']]
   DebugEnabled: !Equals [!Ref Debug, true]
   TracingEnabled: !Not [!Equals ['', !Ref TracingMode]]
+  AllowPolicyEngineScans: !Equals [true, !Ref AllowPolicyEngineScans]
```
austinbyers, line 114: Nit: flip the order here, to be consistent with `DebugEnabled` and because it's more consistent with typical programming style.

```suggestion
  AllowPolicyEngineScans: !Equals [!Ref AllowPolicyEngineScans, true]
```

austinbyers (commented):
Looks good for now. I suspect long-term we will want to have more generic customization of the policy engine / rules engine IAM role. For example, the user could pass in arbitrary IAM managed policies to be attached, which solves this use case and many others

deployments/cloud_security.yml
austinbyers: Yep, CI failed because the `Condition` is not valid where you have it:

```
W8001 Condition AllowPolicyEngineScans not used
deployments/cloud_security.yml:114:3

E2507 IAM Policy key Condition doesn't exist.
deployments/cloud_security.yml:116:1
```

austinbyers:
This may be related to https://github.com/panther-labs/panther/pull/2561

PR #2225, opened 2020-12-08
Initial refactor of dynamic fields - cleaned up git commit history

wey-chiang:
## Background

Addressed some of the leftover comments by @kostaspap 

## Changes

Refactored DestinationOverride to Overrides everywhere except at the rule engine python module level.
Added unit tests to `test_main.py` and `test_rule.py` for the rules engine

Added reference to alert delivery output to be aligned with the Alert page

## Testing

- `mage test:ci`
- `mage:deploy` with manual testing


internal/core/alert_delivery/api/outputs.go
```diff
@@ -28,16 +28,24 @@ import (
...
 	// fetch available panther outputs
 	outputs, err := getOutputs()
 	if err != nil {
 		return nil, err
 	}
 
+	// Check if the alert outputID
+	alertOutputs := []*outputModels.AlertOutput{}
+	for _, outputID := range alert.OutputIds {
+		if outputID == "SKIP" {
```
kostaspap, line 42: Can we make "SKIP" a constant?

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -352,27 +352,27 @@ def _get_description(self, event: Mapping, use_default_on_exception: bool = True
...
-    def _get_destination_override(self, event: Mapping, use_default_on_exception: bool = True) -> Optional[List[str]]:
-        if not hasattr(self._module, 'destination_override'):
+    def _get_destinations(self, event: Mapping, use_default_on_exception: bool = True) -> Optional[List[str]]:
+        if not hasattr(self._module, 'destinations'):
             return None
 
         try:
-            command = getattr(self._module, 'destination_override')
-            destination_override = self._run_command(command, event, list())
+            command = getattr(self._module, 'destinations')
```
kostaspap, line 360: This change is backwards incompatible. 
Is the old code included in 1.14? How are we going to deal with this backward incompatibility? Or are we "hiding" this function in 1.14? 

I had put a similar comment in the old PR: https://github.com/panther-labs/panther/pull/2165#discussion_r535878711

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -520,3 +520,15 @@ def _run_command(self, function: Callable, event: Mapping, expected_type: Any) -
...
+def _rule_id_to_path(rule_id: str) -> str:
+    """Method returns the file path where the rule will be stored"""
+    safe_id = ''.join(x if _allowed_char(x) else '_' for x in rule_id)
+    path = os.path.join(_RULE_FOLDER, safe_id + '.py')
+    return path
+
+
+def _allowed_char(char: str) -> bool:
+    """Return true if the character is part of a valid rule ID."""
+    return char.isalnum() or char in {' ', '-', '.'}
```
kostaspap, line 534: Hmmm, why do we need that? We have this one: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/rules_engine/src/util.py#L22-L31

kostaspap (approved):
LGTM

My only concern is the backwards incompatibility issue.

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: We could attempt to use both `destination_override` and `destinations` and check against both -- however I am not too concerned about this as this was only silently rolled out in v1.14 and not many people are likely using this feature yet.

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: I think this was left over code from merging, will remove

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: Reverted this back in favor of consistency and maintaining backwards compatability

PR #2226, opened 2020-12-08
Allow s3 sources to update label

nhakmiller:
## Background

It's not possible to update s3 source labels right now. This should be possible.

I'm assuming the decision to omit this functionality was explicit because updating the label requires you to re-deploy the CloudFormation stack to update the `Suffix` parameter. Despite this, in my opinion it is still worthwhile to be able to update the label. Not being able to update the label means I have to go through and delete & re-add each source when I want to change our naming convention, instead of just updating the label. This has continuity issues, as now I can no longer search for all the logs from a single source by using the `p_source_id` field as the new sources will have new unique IDs despite being the same s3 bucket from the same account.

## Changes

- Allow users to update the integration label for s3 integrations

## Testing

- Unit testing


kostaspap (approved):
Ah, this is the only change needed? I am wondering if something else will break along the path. Have you verified E2E your usecase e.g. creating/updating/seeing new label in Data lake?

kostaspap (commented):
The label is used as postfix to the IAM role we assume. I think this will break log processing. 
Have you verified E2E?

giorgosp:
Is the frontend change on a separate PR?

Regarding
>The label is used as postfix to the IAM role we assume. I think this will break log processing.

It will until the log-processors cache is updated. Also the user will have to update the CFN stack for just changing the label, otherwise the ingestion will break forever.

After finishing with the S3 improvements tasks, it will be a step towards decoupling label from CFN stack and not requiring any AWS updates when an s3 integration is updated. Will tag you on the ticket when I create it the next days

nhakmiller:
> The label is used as postfix to the IAM role we assume. I think this will break log processing.
> Have you verified E2E?

Haven't verified E2E yet, was going to do a test deployment today because it was a bit late last night when working on the change. I think we will need to change the log processing role name as well, which appears to be stored directly in dynamo and not inferred from this field each time.

nhakmiller:
> Is the frontend change on a separate PR?

No frontend change is required, the frontend already does not restrict you from doing this. That's why I ran into the issue, I submitted the form and was able to re-name the label and it said updated successfully but it just tossed out this field.

kostaspap (approved):

nhakmiller:
> The label is used as postfix to the IAM role we assume. I think this will break log processing.
> Have you verified E2E?

Added the code to update the log processing role name as well and tested E2E in my dev account. Seems to be working with this change.

PR #2183, opened 2020-12-03
Updated S3 Log Source onboard fields

alexmylonas:
## Background

This a first PR targeting a feature branch since backend is not ready and these changes are not functional. 

Work done is related to https://github.com/panther-labs/panther-enterprise/issues/1545 and https://github.com/panther-labs/panther-enterprise/issues/1520

## Changes

- Remove fields s3Prefix & logTypes
- Added new field s3PrefixLogtypes which is a mapping of prefix and logtypes
i.e.
```
[
  {
    prefix: '',
    logTypes: ['AWS.ALB', ...]
  },
  {
    prefix: 's3',
    logTypes: ['AWS.S3ServerAccess' , ...]
  },
]
```

### Screenshots:
Screenshot 2020-12-03 at 11 56 17 AM

### Notes

For your reviewers, design is still in progress, i would like you to focus and agree on code changes

PR #2232, opened 2020-12-09
Backend: Allow users to specify multiple s3 prefix->logtypes mappings per s3 log source

giorgosp:
## Background

As part of improving the s3 onboarding process, we decided to allow users specify multiple s3 prefixes in the same source.
This is a step towards allowing users to define the s3 prefixes and logtypes for their onboarded s3 sources without having to update the stack, but there is still a very long way to go. See this [commit message](https://github.com/panther-labs/panther/commit/8304c08b32a53e810e550c8b906a9be7fd725dc8) for details.

## Changes

- Added a new field `S3PrefixLogTypes` that is an array of `s3Prefix->logtypes` mappings and deprecates the `S3Prefix` and `LogTypes` fields for an s3 source.
- Removed the s3 prefix filter from the log processing CFN template. 

## Testing

- unit
- E2E: Deploy old version, add some s3 sources, deploy new version, verify that prefixes/logtypes can be read/updated successfully.


api/lambda/source/models/api.go
```diff
@@ -55,9 +55,9 @@ type CheckIntegrationInput struct {
 	EnableRemediation *bool `json:"enableRemediation"`
 
 	// Checks for log analysis integrations
-	S3Bucket string `json:"s3Bucket"`
-	S3Prefix string `json:"s3Prefix"`
-	KmsKey   string `json:"kmsKey"`
+	S3Bucket         string           `json:"s3Bucket"`
+	S3PrefixLogTypes S3PrefixLogtypes `json:"s3PrefixLogTypes,omitempty"`
```
austinbyers, line 59: Changes to the Go input structs also need to be reflected in the GraphQL file (`api/graphql/schema.graphql`), or AppSync and the frontend will break, since the type is now different

api/lambda/source/models/integration.go
```diff
@@ -48,32 +49,90 @@ type SourceIntegrationScanInformation struct {
...
+	StackName string `json:"stackName,omitempty"`
+
+	SqsConfig *SqsConfig `json:"sqsConfig,omitempty"`
 }
 
-func (info *SourceIntegration) RequiredLogTypes() (logTypes []string) {
+// S3PrefixLogtypesMapping contains the logtypes Panther should parse for this s3 prefix.
+type S3PrefixLogtypesMapping struct {
+	S3Prefix string   `json:"s3Prefix,omitempty"`
+	Logtypes []string `json:"logtypes,omitempty" validate:"omitempty,min=1"`
```
austinbyers, line 76: `validate:"omitempty,min=1"` makes no sense - you're saying the array is optional, but if it's defined, it has to have at least one element? That is essentially always true.

If you expect at least one log type, do `validate:"required,min=1"`. If log types are optional, then you could either remove the validation tag completely or validate the strings inside the slice: `validate:"omitempty,dive,required"`

austinbyers, line 76: You name this field `LogTypes` in the other API struct, let's do the same here. (instead of `Logtypes`)

api/lambda/source/models/integration.go
```diff
@@ -48,32 +49,90 @@ type SourceIntegrationScanInformation struct {
...
+func (pl S3PrefixLogtypes) S3Prefixes() []string {
+	prefixes := make([]string, len(pl))
+	for i, m := range pl {
+		prefixes[i] = m.S3Prefix
+	}
+	return prefixes
+}
+
+func (s *SourceIntegration) RequiredLogTypes() (logTypes []string) {
+	// We use a switch to avoid git conflicts with enterprise
```
austinbyers, line 98: Don't need this comment anymore

internal/core/source_api/api/utils.go
```diff
@@ -82,9 +81,13 @@ func itemToIntegration(item *ddb.Integration) *models.SourceIntegration {
 	case models.IntegrationTypeAWS3:
 		integration.AWSAccountID = item.AWSAccountID
 		integration.S3Bucket = item.S3Bucket
-		integration.S3Prefix = item.S3Prefix
+		integration.S3PrefixLogTypes = item.S3PrefixLogTypes
+		if len(integration.S3PrefixLogTypes) == 0 {
+			// Backwards compatibility: Use the old fields, maybe the info is there.
```
austinbyers, line 86: Interesting, so existing integrations (that use a single S3 prefix) will still work?

austinbyers (commented):
Looks like a nice improvement! I know there was some confusion on how best to test the new aux template in a live deployment. I'd recommend the following:

* Publish the new aux template manually in your own dev account (e.g. to your `bootstrap-source` bucket)
* Change the URL for loading this template in the `source-api` [here](https://github.com/panther-labs/panther/blob/master/internal/core/source_api/api/get_integration_template.go#L132-L145) and in the frontend [here](https://github.com/panther-labs/panther/blob/master/web/src/components/wizards/ComplianceSourceWizard/StackDeploymentPanel/StackDeployment.tsx#L48-L50) and [here](https://github.com/panther-labs/panther/blob/master/web/src/components/wizards/S3LogSourceWizard/StackDeploymentPanel/StackDeployment.tsx#L50-L52)

This will also help you understand exactly how the aux templates are referenced during onboarding, and lets you test without having to mess with the real public bucket where we publish these templates

api/lambda/source/models/api.go
giorgosp: Yes they have been handled in the frontend PR https://github.com/panther-labs/panther/pull/2183/

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
```diff
@@ -113,17 +107,9 @@ Resources:
...
-                  Resource: !Sub
-                    - 'arn:aws:s3:::${Bucket}/${Prefix}*'
-                    - Bucket: !FindInMap [PantherParameters, S3Bucket, Value]
-                      Prefix: !FindInMap [PantherParameters, S3Prefix, Value]
-                - Effect: Allow
-                  Action: s3:GetObject
-                  Resource: !Sub 'arn:aws:s3:::${S3Bucket}/${S3Prefix}*'
+              - Effect: Allow
+                Action: s3:GetObject
+                Resource: !Sub 'arn:aws:s3:::${S3Bucket}/*'
```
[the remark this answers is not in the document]
giorgosp: Yes the only reason to not add the prefixes here is to allow users to *eventually* update without updating the CFN stack.
If they definitely to want to prevent us from having access to unrelated files, they could:
- change the role processing role manually (stack will drift but I *think* it's fine)
- configure the bucket notifications with separate SNS topic per prefix, so that we don't get notified for unwanted files.
There is definitely manual work, but there are solutions if it is very important of them.
I don't have a strong opinion on it though, so up to you guys, whatever you think will be better for users.
We should probably get some user feedback before deciding on such things.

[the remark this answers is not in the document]
nhakmiller: That's fair. None of our advanced users use these templates anyways, they all use terraform pretty much. This is more for the users who are newer to AWS, in which case making it easy for them to re-configure the source on a whim is probably best.

api/lambda/source/models/integration.go
```diff
@@ -48,32 +49,89 @@ type SourceIntegrationScanInformation struct {
...
-func (info *SourceIntegration) RequiredLogTypes() (logTypes []string) {
+// S3PrefixLogtypesMapping contains the logtypes Panther should parse for this s3 prefix.
+type S3PrefixLogtypesMapping struct {
+	S3Prefix string   `json:"s3Prefix"`
+	LogTypes []string `json:"logtypes" validate:"required,min=1"`
+}
+
+type S3PrefixLogtypes []S3PrefixLogtypesMapping
+
+func (pl S3PrefixLogtypes) LogTypes() []string {
```
kostaspap, line 81: 1. Switch to  `func (pl *S3PrefixLogTypes)`...same below. See: https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

2. Do we even need this method? I think that our codebase outside this package should just stick to `RequiredLogTypes`.

api/lambda/source/models/integration.go
```diff
@@ -48,32 +49,89 @@ type SourceIntegrationScanInformation struct {
...
+	default:
+		panic("Unexpected integration type " + integType)
+	}
+}
+
+// Return the s3 bucket and prefixes configured for this source.
+// For an s3 source, bucket and prefixes are user inputs.
+func (s *SourceIntegration) S3Info() (bucket string, prefixes []string) {
+	switch s.IntegrationType {
+	case IntegrationTypeAWS3:
```
kostaspap, line 127: Also include IntegrationTypeAWSScan. AWS Scans are also sending data to S3

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
```diff
@@ -27,8 +27,6 @@ Mappings:
       Value: '' # RoleSuffix
     S3Bucket:
       Value: '' # S3Bucket
-    S3Prefix:
```
kostaspap, line 30: I see that this is not versioned?

If we make this change and publish it, is it possible that a user running an older version of Panther will get this instead of the CFN template meant for their version?

internal/core/custom_resources/resources/self_registration.go
```diff
@@ -215,7 +215,7 @@ func updateLogProcessingIntegration(source *models.SourceIntegration, logTypes [
 			IntegrationID:    source.IntegrationID,
 			IntegrationLabel: source.IntegrationLabel,
 			S3Bucket:         source.S3Bucket,
-			LogTypes:         logTypes,
+			S3PrefixLogTypes: models.S3PrefixLogtypes{{S3Prefix: "", LogTypes: logTypes}},
```
kostaspap, line 218: Actually that's a perfect example where having multiple prefixes -> log types for the same source is very useful. 
Can you configure it appropriately? E.g. CloudTrail data go to a separate prefix, GuardDuty as well, VPC FL as well, etc

internal/core/source_api/api/update_integration.go
```diff
@@ -74,26 +74,32 @@ func (api API) UpdateIntegrationSettings(input *models.UpdateIntegrationSettings
...
-		zap.L().Error("failed to normalize integration", zap.Error(err))
-		return nil, err
+	updateIntegrationDBItem(existingItem, input)
+
+	if existingItem.IntegrationType == models.IntegrationTypeSqs {
+		err := UpdateSourceSqsQueue(
+			existingItem.IntegrationID, existingItem.SqsConfig.AllowedPrincipalArns, existingItem.SqsConfig.AllowedSourceArns)
+		if err != nil {
+			zap.L().Error("failed to update integration", zap.Error(err))
+			return nil, err
```
kostaspap, line 93: You want to return `updateIntegrationInternalError`

api/lambda/source/models/integration.go
```diff
@@ -48,32 +49,89 @@ type SourceIntegrationScanInformation struct {
...
+	LogTypes []string `json:"logtypes" validate:"required,min=1"`
+}
+
+type S3PrefixLogtypes []S3PrefixLogtypesMapping
+
+func (pl S3PrefixLogtypes) LogTypes() []string {
+	var logtypes []string
+	for _, m := range pl {
+		logtypes = append(logtypes, m.LogTypes...)
+	}
```
kostaspap, line 85: I imagine you want to bring the unique log types. Check these functions, you might find them useful: https://github.com/panther-labs/panther/blob/master/pkg/stringset/stringset.go

internal/core/source_api/api/put_integration.go
```diff
@@ -170,12 +169,6 @@ func (api API) integrationAlreadyExists(input *models.PutIntegrationInput) error
 							input.IntegrationLabel),
 					}
 				}
-
-				if existingIntegration.S3Bucket == input.S3Bucket && existingIntegration.S3Prefix == input.S3Prefix {
-					return &genericapi.InvalidInputError{
-						Message: "An S3 integration with the same S3 bucket and prefix already exists.",
-					}
-				}
```
kostaspap, line 178: We want some kind of check here. 
1. A user shouldn't add the same prefix twice in the same S3 source
2. A user shouldn't add a prefix/bucket combination that is configuration in a different source.

internal/core/source_api/api/update_integration_test.go
```diff
@@ -280,3 +277,62 @@ func TestSlicesContainSameElements(t *testing.T) {
 		})
 	}
 }
+
+// Tests that an old s3 source that has the deprecated s3 prefix and logtypes fields
+// populated, can be successfully read and updated with the new s3PrefixLogTypes field.
+func TestS3PrefixLogTypes_BackwardsCompatibility(t *testing.T) {
```
kostaspap, line 283: 👍

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
```diff
@@ -208,20 +209,18 @@ func TestSourceCacheStructFind(t *testing.T) {
 	sources := []*models.SourceIntegration{
 		{
 			SourceIntegrationMetadata: models.SourceIntegrationMetadata{
-				IntegrationID:   "1",
-				IntegrationType: models.IntegrationTypeAWS3,
-				S3Bucket:        "foo",
-				S3Prefix:        "",
-				LogTypes:        []string{"Foo.Bar"},
+				IntegrationID:    "1",
```
kostaspap, line 212: Can you add a test with a source that has multiple prefixes?

internal/core/source_api/api/put_integration.go
giorgosp: Added

internal/core/source_api/api/update_integration.go
giorgosp: updated

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
giorgosp: yep added

api/lambda/source/models/integration.go
kostaspap: Ah right, didn't realize it was a slice. Fine to leave as is.

internal/core/custom_resources/resources/self_registration.go
kostaspap: We don't create org trail (we couldn't). We create a normal trail. 
Here you can see the details: 
https://github.com/panther-labs/panther/blob/master/deployments/onboard.yml#L77-L86

The prefixes are controlled by us.

giorgosp: I have looked into it more. If we ever decide to change the prefixes in the CFN templates, nobody will remember to update the self_registration.  Narrowing down the prefixes here is very risky. For what benefit?

The log processor also still classifies based on all the source's  logtypes. We need to update it to narrow down the logtypes by the prefix, when possible. But this is not yet implemented.

giorgosp: I could perhaps add a comment in the CFN templates, so that if we change the prefixes, update the `self_registration` too. But the bug will also occur if a new log type is added there.

kostaspap: The benefits are: 
1. Improved performance
2. We experience a bug from time to time where we try to parse JSON data using CSV parser and the CSV parser triggers an OOM exception. 

it is unlikely we will ever change the prefix, and if we ever do we will capture that issue easily. 
A comment in there should be enough. We don't change prefixes that easily. There are many places in the code were some prefix in the CFN template is also used in the code e.g. see SQS source Firehose, EventBridge Firehose, CloudSecurity firehose and so on.

kostaspap: The feature was developed for usecases exactly like this (one bucket, many log types in different prefixes). We recommend users to take the approach of specifying their Prefix -> Log types correctly. Why shouldn't we do it. 
I believe it should be easy to do.

giorgosp: It's easy but time consuming, factoring in deploying and testing etc. But my main concern is it is brittle and we can't easily unit test that these prefixes match all the configurations in the CFN templates.

The only way to find any problems is mis-classification logs and alerts.

I have almost implemented it, needs some testing.

kostaspap (approved):
LGTM once you address any open comments!

internal/core/custom_resources/resources/self_registration.go
giorgosp: Updated. Deploying now

api/lambda/source/models/integration.go
giorgosp: updated, had to deal with some conflicts now that I got your cloud security logs PR

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -162,13 +163,23 @@ func NewFactory(resolver logtypes.Resolver) Factory {
...
 					Resolver:   resolver,
 					LoadSource: sources.LoadSource,
 				},
 			}, nil
 		case models.IntegrationTypeAWS3, models.IntegrationTypeAWSScan:
-			c, err := sources.BuildClassifier(src, resolver)
+			var availableLogTypes []string
+			if src.IntegrationType == models.IntegrationTypeAWS3 {
+				for _, m := range src.S3PrefixLogTypes {
+					if strings.HasPrefix(input.S3ObjectKey, m.S3Prefix) {
```
kostaspap, line 175: 1. This logic doesn't seem correct. Prefixes might overlap, e.g. you can have prefixes like "test", "test/foo" and S3 key "test/boo/bar.gz". The sources cache in the log processor has logic to deal with this. Maybe you can modify the logic in the sources cache to return you a SourceIntegration only with the appropriate log types. Or, sources cache could just return a different structure and not `SourceIntegration` since that struct is not so useful for the log processor anymore
2. Why do the S3 and AWScan need different logic? In the end, the AWSScan also has S3PrefixLogTypes but has only 1 entry.

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -64,10 +64,11 @@ type s3ClientCacheKey struct {
...
 	// last time the cache was updated
 	cacheUpdateTime time.Time
-	// sources by id
-	index map[string]*models.SourceIntegration
-	// sources by s3 bucket sorted by longest prefix first
-	byBucket map[string][]*models.SourceIntegration
+	// prefixes by s3 bucket. Each prefixes array is ordered by longest first.
+	// An s3 source may have many prefixes, that's why we can't just directly map from bucket to source.
+	prefixesByBucket map[string][]string
+	sourceByPrefix   map[string]*models.SourceIntegration
```
kostaspap, line 70: This doesn't look correct. You can have two sources in different buckets with the same prefixes. They would overlap here.

kostaspap (changes requested):

internal/log_analysis/log_processor/processor/processor.go
giorgosp: 1. Some fields from `SourceIntegration` are used for logging etc.  Your concern is correct although the source cache's logic doesn't help here. Imagine users define these prefixes and log types, which log types should be available for classification if an object key `prefixA/prefixB/log.json` comes in?
- `prefixA/ -> [Log.A]`
- `prefixA/prefixB/ -> [Log.B]`
- `prefixC/ -> [Log.C]`
- `  ""        -> [Log.D]`
We need to use the union of all the matching prefixes, in the example this would be `[Log.A, Log.B, Log.D]`. This is what the code does now.

2. `AWSScan` doesn't have the `S3PrefixLogTypes` field, it's `prefix` and `logtypes` fields are static (`cloudsecurity` prefix and `snapshotlogs.LogTypes()`).

internal/log_analysis/log_processor/sources/s3_client_cache.go
giorgosp: Hmm valid concern, will add a unit test and take it from there

internal/log_analysis/log_processor/processor/processor.go
kostaspap: > We need to use the union of all the matching prefixes, in the example this would be [Log.A, Log.B, Log.D]. This is what the code does now.

That's not how the application works now. The way the application works now is that longer prefixes get priority over shorter ones. We had a test specifically for this see : https://github.com/panther- labs/panther/blob/master/internal/log_analysis/log_processor/sources/s3_client_cache_test.go#L205

This solves cases like the ones we have in our self-onboarding. We have a bucket where: 
1. Cloudtrail data get in <bucket>/AWS/<AccountID>/CloudTrail/
2. S3 Server Access data get in <bucket>/*

For this purpose, we have two prefix mappings:
1. A prefix mapping "/AWS/<AccountID>/CloudTrail/" -> AWS.CloudTrail
2. A prefix mapping for "" -> AWS.S3ServerAccess

Since longer prefix gets priority, any CloudTrail events will use only the CloudTrail parser, S3 server access events will use the S3 Server access parser.

internal/log_analysis/log_processor/sources/sources.go
```diff
@@ -65,56 +57,3 @@ func LoadSourceS3(bucketName, objectKey string) (*models.SourceIntegration, erro
 
 	return result, nil
 }
-
```
alxarch, line 68: The BuildClassifier was left here on purpose.
We don't want the classifier module to depend on the mess that is models.SourceIntegration.
The goal is to replace classifier altogether with a `pantherlog.LogParser` interface at some point.
It is also difficult to track the changes without inline diffs.
Please put it back.

internal/log_analysis/log_processor/processor/processor.go
alxarch: Longer prefixes override shorter ones.
One of the reasons to use prefixes (and to make it easier to add new ones with this PR) is to be able to use the same S3 bucket for conflicting log types. If we don't 'specialize' log types based on prefix length, we will not be able to classify correctly or add 'catch all' log types with `""` prefix.

internal/log_analysis/log_processor/sources/s3_client_cache.go
alxarch: The cache logic relies on sourceByPrefix mapping to multiple sources.

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -112,45 +113,48 @@ func (c *sourceCache) Sync(now time.Time) error {
...
+	return c.sourceByID[id]
 }
 
 // FindS3 looks up a source by bucket name and prefix without updating the cache
-func (c *sourceCache) FindS3(bucketName, objectKey string) *models.SourceIntegration {
-	sources := c.byBucket[bucketName]
-	for _, source := range sources {
-		if strings.HasPrefix(objectKey, source.RequiredS3Prefix()) {
-			return source
+func (c *sourceCache) FindS3(bucket, objectKey string) *models.SourceIntegration {
```
alxarch, line 154: I would suggest to keep the old logic of the cache and return `(src *models.SourceIntegration, logTypes []string)` here.

alxarch (changes requested):
The longest prefix needs to have priority.
The changes to source cache code seem incorrect.

internal/log_analysis/log_processor/sources/sources.go
giorgosp: Moved it back

internal/log_analysis/log_processor/processor/processor.go
giorgosp: Updated. Fixed to use the longest prefix

internal/log_analysis/log_processor/sources/s3_client_cache.go
giorgosp: Update it to return the correct source. Added a test case

internal/log_analysis/log_processor/sources/s3_client_cache.go
giorgosp: Ended up adding a method `S3PrefixLogTypes.LongestPrefixMatch()` to avoid adding many changes

api/lambda/source/models/integration.go
```diff
@@ -48,34 +51,79 @@ type SourceIntegrationScanInformation struct {
...
+func (pl S3PrefixLogtypes) S3Prefixes() []string {
+	prefixes := make([]string, len(pl))
+	for i, m := range pl {
+		prefixes[i] = m.S3Prefix
+	}
+	return prefixes
+}
+
+// Return the S3PrefixLogtypesMapping whose prefix is the longest one that matches the objectKey.
+func (pl S3PrefixLogtypes) LongestPrefixMatch(objectKey string) *S3PrefixLogtypesMapping {
```
kostaspap, line 100: Could switch it to return `(S3PrefixLogTypesMapping, bool)`. See https://github.com/golang/go/wiki/CodeReviewComments#in-band-errors

internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -168,7 +168,16 @@ func NewFactory(resolver logtypes.Resolver) Factory {
 				},
 			}, nil
 		case models.IntegrationTypeAWS3, models.IntegrationTypeAWSScan:
-			c, err := sources.BuildClassifier(src, resolver)
+			var availableLogTypes []string
```
kostaspap, line 171: instead of having a case statement and then an `if` inside, maybe it will be more clean if you create two different case statements, 
```
....
case models.IntegrationTypeAWSScan:
....
```

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -112,26 +117,27 @@ func (c *sourceCache) Sync(now time.Time) error {
...
+	byBucket := make(map[string][]prefixSource)
 	index := make(map[string]*models.SourceIntegration)
 	for _, source := range sources {
-		bucketName := source.RequiredS3Bucket()
-		bucketSources := byBucket[bucketName]
-		byBucket[bucketName] = append(bucketSources, source)
+		bucket, prefixes := source.S3Info()
+		bucketSources := byBucket[bucket]
+		for _, prefix := range prefixes {
+			byBucket[bucket] = append(bucketSources, prefixSource{prefix: prefix, source: source})
```
kostaspap, line 126: This should be `byBucket[bucket] = append(byBucket[bucket], prefixSource{prefix: prefix, source: source})`.

You can verify the bug by changing the line 305 of the unit test to be:
```
src := cache.FindS3("bar", "foo/foo/foo/prefix/qux.json")
```
The test will fail, even though it shouldn't

kostaspap (changes requested):

internal/log_analysis/log_processor/sources/s3_client_cache.go
giorgosp: Good catch! fixed

internal/log_analysis/log_processor/processor/processor.go
giorgosp: updated

api/lambda/source/models/integration.go
giorgosp: Good idea, updated

kostaspap (approved):
LGTM

PR #2235, opened 2020-12-09
Implement custom log type updates without considering race conditions

alxarch:
## Background

Implement updates for custom log types

## Changes

- Adds diff for logschema.Schema
- Blocks backward incompatible updates
- Sends UpdateTable event to datacatalog updater
- Separates `CreateOrUpdate` to `CreateIfNotExists` and `UpdateIfExists` actions used in `CreateTables` and `UpdateTable` events respectively
- Refactor logtypes API to have a sane way to do TransactWriteItems on DDB (`ddbextras`)

## Testing

- mage test:go
- E2E


internal/core/logtypesapi/customlogs.go
```diff
@@ -73,52 +74,94 @@ type CustomLogRecord struct {
...
+	if err := desc.Validate(); err != nil {
+		return nil, NewAPIError("InvalidMetadata", err.Error())
+	}
+	schema := logschema.Schema{}
+	if err := yaml.Unmarshal([]byte(c.LogSpec), &schema); err != nil {
+		return nil, NewAPIError("InvalidSyntax", err.Error())
+	}
+	if _, err := customlogs.Build(desc, &schema); err != nil {
+		return nil, NewAPIError("InvalidLogSchema", err.Error())
+	}
```
kostaspap, line 161: Should these errors be constants here, like we do for the other errors returned to users? internal/core/logtypesapi/api.go

internal/core/logtypesapi/customlogs.go
```diff
@@ -73,52 +74,94 @@ type CustomLogRecord struct {
...
+	if err != nil {
 		return &PutCustomLogOutput{
-			Error: NewAPIError("InvalidMetadata", err.Error()),
+			Error: WrapAPIError(err),
 		}, nil
 	}
-	schema := logschema.Schema{}
-	if err := yaml.Unmarshal([]byte(input.LogSpec), &schema); err != nil {
+	currentRevision := input.Revision
+	if rev := currentRevision; rev == 0 {
```
kostaspap, line 90: nit: just 
```
if currentRevision == 0 {
...
}
```

internal/core/logtypesapi/customlogs.go
```diff
@@ -73,52 +74,94 @@ type CustomLogRecord struct {
...
+		result, err := api.Database.CreateCustomLog(ctx, id, &input.CustomLog)
```
kostaspap, line 91: If I get it right, creating a new Custom Log or not relies on the version passed by the client in the request? 
This seems that it could break easily - a misbehaving client can sent requests for the same version. 
Revision (version?) number should be generated by the backend when someone tries to update an existing log type.

internal/core/logtypesapi/dynamodb.go
```diff
@@ -140,52 +142,87 @@ func (d *DynamoDBLogTypes) BatchGetCustomLogs(ctx context.Context, ids ...string
...
+
+	//for i := int64(1); i < revision; i++ {
+	//	input := dynamodb.DeleteItemInput{
+	//		TableName: aws.String(d.TableName),
+	//		Key:       customRecordKey(id, i),
+	//	}
+	//	if _, err := d.DB.DeleteItemWithContext(ctx, &input); err != nil {
+	//		return err
+	//	}
+	//}
```
kostaspap, line 175: leftovers?

internal/core/logtypesapi/dynamodb.go
```diff
@@ -140,52 +142,87 @@ func (d *DynamoDBLogTypes) BatchGetCustomLogs(ctx context.Context, ids ...string
...
 func (d *DynamoDBLogTypes) DeleteCustomLog(ctx context.Context, id string, revision int64) error {
-	input := dynamodb.TransactWriteItemsInput{
+	tx, err := buildDeleteRecordTx(d.TableName, id, revision)
+	if err != nil {
+		return errors.WithMessage(err, "failed to prepare delete transaction")
+	}
+
+	if _, err := d.DB.TransactWriteItemsWithContext(ctx, tx); err != nil {
+		if txErr, ok := err.(*dynamodb.TransactionCanceledException); ok {
+			switch reason := txErr.CancellationReasons[0]; cancellationReasonCode(reason) {
```
kostaspap, line 152: Is this going to have always 1 element?

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
```diff
@@ -35,7 +35,18 @@ type CreateTablesEvent struct {
 }
 
 func (h *LambdaHandler) HandleCreateTablesEvent(ctx context.Context, event *CreateTablesEvent) error {
-	if err := h.createOrUpdateTablesForLogTypes(ctx, event.LogTypes); err != nil {
+	// This event is only coming from an update to sources API.
+	// We only need to try creating the tables once.
+	logTypes := h.notCreatedYet(event.LogTypes)
+	if len(logTypes) == 0 {
+		return nil
+	}
```
kostaspap, line 43: Why bother put all this logic in here? We can just try to create the table in Glue and if it fails with "AlreadyExists" we are good. This should be a low volume operation anyway. It saves us some lines of code to remove this.

kostaspap (changes requested):

internal/core/logtypesapi/dynamodb.go
alxarch: The TransactionCanceledException has as many elements in CancelationReasons as the items in the TransactWriteItemsInput. So yes, we know we are checking the specific condition for the first element

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
```diff
@@ -35,7 +35,18 @@ type CreateTablesEvent struct {
...
+	// TODO: This event needs to ONLY CREATE tables. Updating a table MUST include syncing the partitions.
+	// This leaves a gap where a table might get updated but no partition sync will happen.
+	// To achieve that we need a mapping from logType to required tables.
+	// Currently this is 'transparently' done in the gluetables package.
+	// This should be handled centrally (i.e. pantherdb)
+	if err := h.createOrUpdateTablesForLogTypes(ctx, logTypes); err != nil {
```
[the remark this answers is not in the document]
alxarch: The createOrUpdateTables handles the creation of ruleErrors and ruleMatches. I tried recreating the function to only create the tables it was given but the result was a lot of code, so I reverted.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
```diff
@@ -44,22 +55,60 @@ func (h *LambdaHandler) HandleCreateTablesEvent(ctx context.Context, event *Crea
...
 	return nil
 }
 
+func (h *LambdaHandler) observeCreatedTable(logType string) {
+	// Store partition in cache as successfully created
+	if h.tablesCreated == nil {
+		h.tablesCreated = make(map[string]struct{})
+	}
+	h.tablesCreated[logType] = struct{}{}
+}
```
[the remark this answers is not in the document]
alxarch: Avoiding extra CreateTable calls was added because we rely on the 'auto-updating' createOrUpdateTables function. It limits the posibillity of actually doing an update of the table here. We want updates to only happen when custom logs are updated or when we deploy. There's still a chance that we might do an out-of-place update but it is significantly lower. The alternative was to have an 'only-create' function but handling the related tables there was  more code than this simple check. I am not happy about the state of things either. This is a hail mary kind of think.

internal/log_analysis/datacatalog_updater/datacatalog/handler.go
```diff
@@ -45,9 +45,14 @@ type LambdaHandler struct {
...
 	Logger                *zap.Logger
+	// Clears a specific log type from the cache
+	ClearLogTypeCache func(name string)
 
-	// Glue partitions known to have been created. (use map[string]string where key == value for map size)
+	// Glue partitions known to have been created.
 	partitionsCreated map[string]struct{}
+
+	// Glue tables known to have been created.
+	tablesCreated map[string]struct{}
```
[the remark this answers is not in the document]
alxarch: Read my reply above as to why this was added

internal/log_analysis/log_processor/customlogs/updates.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/logschema"
+)
+
+func CheckSchemaChange(d *logschema.Change) error {
+	switch d.Type {
+	case logschema.DeleteField:
+		// Note that field renames appear in a change log as a pair of DeleteField, AddField
+		target, path := splitPath(d.Path)
+		return errors.Errorf("cannot delete field %q at %q", target, path)
```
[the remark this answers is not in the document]
alxarch: we can't because this opens the door to renaming fields by two consecutive updates

internal/core/logtypesapi/dynamodb.go
```diff
@@ -196,139 +233,213 @@ func (d *DynamoDBLogTypes) CreateCustomLog(ctx context.Context, id string, param
...
 	}
+	return tx, tx.Validate()
+}
+
+func (d *DynamoDBLogTypes) UpdateCustomLog(ctx context.Context, id string, revision int64, params *CustomLog) (*CustomLogRecord, error) {
+	now := time.Now().UTC()
+	record := CustomLogRecord{
+		CustomLog: *params,
+		LogType:   id,
+		Revision:  revision + 1,
```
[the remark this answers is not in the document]
alxarch: No, it does not work this way. The revision is only for preventing users stepping on each other's toes. The actual guard that does not allow users to override a record is in the id itself. The create transaction (when revision is 0) has a condition that the record does not exist. So a user putting 0 there has no effect.

internal/log_analysis/datacatalog_updater/main/lambda.go
```diff
@@ -82,12 +84,20 @@ func main() {
 		LambdaAPI:  lambdaClient,
 	}
 
+	cachedResolver := logtypes.CachedResolver(logTypeMaxAge, &logtypesapi.Resolver{
+		LogTypesAPI: logtypesAPI,
+	})
```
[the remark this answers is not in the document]
alxarch: We do need a cached resolver, otherwise we would be hitting the LogTypesAPI on every file we process in the log processor.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
alxarch: Answered below

internal/core/logtypesapi/customlogs.go
alxarch: fixed

internal/core/logtypesapi/customlogs.go
alxarch: The constraint comes from the different transaction conditions on create and update. Create ensures the record does not exist. Update ensures that the record is at this revision (it also reads the entry from DDB so user input is there only for avoiding overwrites).

internal/core/logtypesapi/customlogs.go
```diff
@@ -73,52 +74,98 @@ type CustomLogRecord struct {
...
+	schema, err := buildSchema(id, &input.CustomLog)
+	if err != nil {
 		return &PutCustomLogOutput{
-			Error: NewAPIError("InvalidMetadata", err.Error()),
+			Error: WrapAPIError(err),
 		}, nil
 	}
-	schema := logschema.Schema{}
-	if err := yaml.Unmarshal([]byte(input.LogSpec), &schema); err != nil {
+	currentRevision := input.Revision
```
giorgosp, line 89: Versioning is implemented so that 2 different users cannot update the same schema simultaneously?

giorgosp: Does the frontend handle revision error gracefully? I mean telling the user that a different version is on the server and handle it accordingly? Or the frontend just shows a "500 retry later"?

internal/core/logtypesapi/customlogs.go
```diff
@@ -73,52 +74,98 @@ type CustomLogRecord struct {
...
-	if _, err := customlogs.Build(desc, &schema); err != nil {
-		return &PutCustomLogOutput{
-			Error: NewAPIError("InvalidLogSchema", err.Error()),
-		}, nil
+	current, err := api.Database.GetCustomLog(ctx, id, currentRevision)
+	if err != nil {
+		return nil, err
+	}
+	if current == nil {
+		return nil, NewAPIError(ErrRevisionConflict, fmt.Sprintf("record %s@%d does not exist", id, currentRevision))
```
giorgosp, line 102: nit: I think someone reading the logs won't understand the id@revision format. Perhaps change the error texts to something like `record %s at revision %d does not exist` and `record %s is on revision %d, not %d`

internal/core/logtypesapi/customlogs.go
```diff
@@ -73,52 +74,98 @@ type CustomLogRecord struct {
...
-	result, err := api.Database.CreateCustomLog(ctx, id, &input.CustomLog)
+
+	result, err := api.Database.UpdateCustomLog(ctx, id, currentRevision, &input.CustomLog)
 	if err != nil {
 		return &PutCustomLogOutput{
 			Error: WrapAPIError(err),
 		}, nil
 	}
+	if err := api.DataCatalog.SendUpdateTableForLogType(ctx, id); err != nil {
+		return nil, err
```
giorgosp, line 126: If sending the SQS message fails, the UI will probably show a "try again later" message but they won't have the new revision number (because saving in db succeeded). Subsequent requests will probably fail with wrong revision error, so we need to make sure the FE handles revision conflicts well and guides makes it easy for the user to proceed.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
```diff
@@ -44,22 +55,60 @@ func (h *LambdaHandler) HandleCreateTablesEvent(ctx context.Context, event *Crea
...
+		// - CreateDatabaseIfNotExists
+		// - CreateTableIfNotExists
+		// - CreateOrReplaceTable
+		// - CreatePartitionIfNotExists
+		// - CreateOrReplacePartition
+		// - ScanDatabases
+		// - ScanDatabaseTables
+		// - ScanTablePartitions
+		// These actions should be part of an interface that manages the data lake backend.
+		// We need methods that use abstract Database/Table/Partition structs that can contain info for all backends.
```
giorgosp, line 95: 👍

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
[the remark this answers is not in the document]
giorgosp: Since we all agree that a  create-only function is the a cleaner solution I'd suggest we proceed writing it, in the simplest way possible. No need to refactor a lot of code now to make it ideal, we can do small improvements. A copycat of `CreateOrUpdateGlueTables()` would be enough for now imo.
The current logic to avoid updates is more difficult to follow.

internal/log_analysis/datacatalog_updater/main/lambda.go
```diff
@@ -82,12 +84,20 @@ func main() {
...
+
+	// Store the clear cache of the underlying cached resolver.
+	// We use it on custom log updates to ensure we have the latest schema.
+	clearCache := cachedResolver.(interface {
```
giorgosp, line 93: Why not just `clearCache := cachedResolver.Forget` ?

giorgosp (commented):
LGTM generally, leaving final approval to Kostas

internal/core/logtypesapi/customlogs.go
alxarch: Yes, the API errors are embedded in the response and they are shown to the user.

internal/core/logtypesapi/customlogs.go
alxarch: I don't see any way to avoid returning the erorr here.
Error handling is hard. Our hopes here lie with the retries in the underlying AWS session.
Do you have any suggestion on how to handle this gracefully?

internal/log_analysis/datacatalog_updater/main/lambda.go
alxarch: cachedResolver is of type `logtypes.Resolver`. We need the cast to access the other function that exists in the underlying value.

alxarch:
> Since we all agree that a create-only function is the a cleaner solution I'd suggest we proceed writing it, in the simplest way possible. No need to refactor a lot of code now to make it ideal, we can do small improvements. A copycat of CreateOrUpdateGlueTables() would be enough for now imo.
The current logic to avoid updates is more difficult to follow.

I spent 2 hours this morning trying to do this and there is just too much code. The problem once again stems from the non-existent separation of concerns in the `awsglue` package.There's too much code that needs shuffling around to achieve this. 
Factoring in the test changes required, I don't think it's viable for this PR. I will add it as a PR on top of this PR to show the required changes.
With the plan to move to an abstraction layer above Glue/Athena and Snowflake to handle Tables, all that work will be futile.
My suggestion if we insist on this, is to shelve this PR for now and come back to it after we conclude the conversations for the schema management RFC. I agree that having just a good-enough solution with corner cases is not the best option. But this needs bigger changes to be properly fixed.

internal/core/logtypesapi/customlogs.go
giorgosp: I know but if `SendUpdateTableForLogType` fails it will leave the system in an inconsistent state. Perhaps show a message to the user to contact Panther support?

giorgosp: It will be happen very rarily but if it happens it will be a frustrating experience for the user.

internal/log_analysis/datacatalog_updater/main/lambda.go
giorgosp: This is a bit confusing. Why not make `cachedResolver` public and make `CachedResolver()` return a struct instead of interface? This will also follow the idiom of `return structs accept interfaces`.

kostaspap:
> > Since we all agree that a create-only function is the a cleaner solution I'd suggest we proceed writing it, in the simplest way possible. No need to refactor a lot of code now to make it ideal, we can do small improvements. A copycat of CreateOrUpdateGlueTables() would be enough for now imo.
> > The current logic to avoid updates is more difficult to follow.
> 
> I spent 2 hours this morning trying to do this and there is just too much code. The problem once again stems from the non-existent separation of concerns in the `awsglue` package.There's too much code that needs shuffling around to achieve this.
> Factoring in the test changes required, I don't think it's viable for this PR. I will add it as a PR on top of this PR to show the required changes.
> With the plan to move to an abstraction layer above Glue/Athena and Snowflake to handle Tables, all that work will be futile.
> My suggestion if we insist on this, is to shelve this PR for now and come back to it after we conclude the conversations for the schema management RFC. I agree that having just a good-enough solution with corner cases is not the best option. But this needs bigger changes to be properly fixed.

As I see it the change is: 
1. Add a new method "CreateTable" here https://github.com/panther-labs/panther/blob/master/internal/log_analysis/awsglue/table_metadata.go#L45-L52
2. Add a new method CreateTables here: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/gluetables/gluetables.go#L79
3. Invoke that method from the code above. 

It is a cleaner and more robust solution. I think it could be challenging to refactoring the existing code, but we don't have to do that.

alxarch:
> As I see it the change is:
> 
> 1. Add a new method "CreateTable" here https://github.com/panther-labs/panther/blob/master/internal/log_analysis/awsglue/table_metadata.go#L45-L52
> 2. Add a new method CreateTables here: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/gluetables/gluetables.go#L79
> 3. Invoke that method from the code above.

The problem is that all the logic to produce the StorageDescriptor is bundled with the CreateOrUpdate stuff. Starting to move things around to extract functions quickly leads to a major overhaul of the awsglue.go file.

internal/log_analysis/datacatalog_updater/main/lambda.go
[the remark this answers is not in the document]
kostaspap: Are you sure? Can you point me to the code? When I look at the code, I see that we call the resolver only when we update or create a table. The `HandleS3EventRecord` which handles the partition creation doesn't use the resolver.

internal/log_analysis/log_processor/logtypes/resolver.go
```diff
@@ -70,3 +74,72 @@ func (c chainResolver) Resolve(ctx context.Context, name string) (Entry, error)
 	}
 	return nil, nil
 }
+
+// CachedResolver
+func CachedResolver(maxAge time.Duration, r Resolver) Resolver {
```
kostaspap, line 79: Tests for this?

internal/log_analysis/log_processor/logtypes/resolver.go
```diff
@@ -70,3 +74,72 @@ func (c chainResolver) Resolve(ctx context.Context, name string) (Entry, error)
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	delete(c.entries, name)
+}
+
+func (c *cachedResolver) Resolve(ctx context.Context, name string) (Entry, error) {
+	if e := c.find(name); e.IsValid(c.maxAge) {
+		return e.Entry, nil
+	}
+	reply, err, _ := c.group.Do(name, func() (interface{}, error) {
```
kostaspap, line 114: This looks pretty neat. I assume we have to do this because the resolution could potentially include long running tasks (e.g. including I/O)?

internal/log_analysis/log_processor/main/lambda.go
```diff
@@ -68,13 +71,14 @@ func process(ctx context.Context, scalingDecisionInterval time.Duration) (err er
 	logTypesResolver := logtypes.ChainResolvers(
 		registry.NativeLogTypesResolver(),
 		snapshotlogs.Resolver(),
-		&logtypesapi.Resolver{
+		logtypes.CachedResolver(logTypeMaxAge, &logtypesapi.Resolver{
```
kostaspap, line 74: What would be the issue if we use a cached entry that is no longer valid?

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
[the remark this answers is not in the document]
kostaspap: What is the scenario we are trying to protect against using this cache? What will happen if this code gets invoked? 

As as see it, this code will get invoked: every time someone adds a log type (that was already added to one source)  to another source, this code will execute. The cache will not help for many of our scenarios - we have multiple instances of this lambda running, each with their own cache and lambda's get recycled after few hours. What would be the problem if this code gets invoked?

internal/core/logtypesapi/customlogs.go
alxarch: fixed

internal/core/logtypesapi/dynamodb.go
alxarch: I refactored the WriteTransaction logic to use `ddbextras` module. It binds the error handlers with each `TransactWriteItem` so that it is more convenient to define them in a declarative way.

internal/core/logtypesapi/dynamodb.go
alxarch: fixed

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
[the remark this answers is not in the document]
alxarch: I introduced ``  methods and functions to `awsglue.GlueTableMetadata#CreateTableIfNotExists()` and  `gluetables.CreateTablesIfnotExists` that handle the table creation *only* when a view uses a schema. I also removed the created tables cache logic.

internal/log_analysis/datacatalog_updater/datacatalog/handler.go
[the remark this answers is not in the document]
alxarch: Fixed

internal/log_analysis/datacatalog_updater/main/lambda.go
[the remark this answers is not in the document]
alxarch: We need the cached resolver for the log processor. The data catalog could avoid caching the results but since we reolve each available log type multiple times on most requests (resolveTables etc) it would make no sense to rebuild and revalidate the logtype entries over and over again. I can remove it if you want.

internal/core/logtypesapi/customlogs.go
alxarch: fixed

alxarch: Just noting that it is still possible for the update to fail once queued. Which is more of a real issue than failing to send a message to SQS. Cannot address this in this PR. We need state for the pending updates to be store in DDB.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
alxarch: Fixed

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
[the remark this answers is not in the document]
alxarch: I split the logic for each event to createaIfNotExists/updateIfExists

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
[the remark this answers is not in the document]
alxarch: Also UpdateIfExists

internal/core/logtypesapi/customlogs.go
alxarch: fixed

internal/log_analysis/datacatalog_updater/main/lambda.go
alxarch: Fixed

internal/core/logtypesapi/customlogs.go
```diff
@@ -73,52 +74,98 @@ type CustomLogRecord struct {
...
+	current, err := api.Database.GetCustomLog(ctx, id, currentRevision)
+	if err != nil {
+		return nil, err
+	}
+	if current == nil {
+		return nil, NewAPIError(ErrRevisionConflict, fmt.Sprintf("record %s@%d does not exist", id, currentRevision))
+	}
+	if current.Revision != currentRevision {
+		return nil, NewAPIError(ErrRevisionConflict, fmt.Sprintf("record %s@%d is on a different revision", id, currentRevision))
 	}
```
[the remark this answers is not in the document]
alxarch: It is best to fail early here. A user should be informed that the entry was updated before trying to fix any other errors (validation, invalid update, syntax etc)

internal/log_analysis/log_processor/logtypes/resolver.go
alxarch: Yes, singleflight avoids multiple API requests firing  when a record is stale.

internal/log_analysis/log_processor/main/lambda.go
alxarch: The failure scenario probably involves missing fields/indicators etc.

internal/log_analysis/datacatalog_updater/datacatalog/update_table.go
```diff
@@ -0,0 +1,65 @@
...
+		return err
+	}
+	tbl := tableForEntry(entry)
+	updated, err := tbl.UpdateTableIfExists(ctx, h.GlueClient, h.ProcessedDataBucket)
+	if err != nil {
+		return err
+	}
+	// If the table was not updated, it means that it does not exist yet.
+	if !updated {
+		return nil
```
kostaspap, line 48: When can this happen? Probably if we are in inconsistent state? Should it return error here to alert us?

internal/log_analysis/log_processor/main/lambda.go
kostaspap: Ok, we can live with that :)

internal/log_analysis/datacatalog_updater/main/lambda.go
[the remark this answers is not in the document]
kostaspap: Yeah, we can keep the cached resolved for the log processor. 

As I see, removing the cached Resolver from here simplifies the logic - no more need for the "Forget" functionality and removes the change of wrong resolution. 
The I/O required is 1 I/O operation per log type on ever Update or Create table operation, correct? Sounds negligible.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
```diff
@@ -35,12 +35,51 @@ type CreateTablesEvent struct {
...
 
 func (h *LambdaHandler) HandleCreateTablesEvent(ctx context.Context, event *CreateTablesEvent) error {
-	if err := h.createOrUpdateTablesForLogTypes(ctx, event.LogTypes); err != nil {
+	// This event is only coming from an update to sources API.
+	// We only need to try creating the tables once.
+	logTypes := event.LogTypes
+	if len(logTypes) == 0 {
+		return nil
+	}
+	// TODO: This event needs to ONLY CREATE tables. Updating a table MUST include syncing the partitions.
```
giorgosp, line 44: 👍

internal/log_analysis/datacatalog_updater/datacatalog/update_table.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/panther-labs/panther/internal/log_analysis/pantherdb"
+)
+
+type UpdateTablesEvent struct {
+	LogType string
+	TraceID string
+}
+
+func (h *LambdaHandler) HandleUpdateTablesEvent(ctx context.Context, event *UpdateTablesEvent) error {
+	// We need to fetch a fresh entry for this log type
```
giorgosp, line 35: Perhaps update the comment to explain the why we need to fetch a fresh entry?

giorgosp (commented):
Nice work, looks good to me. Only left some comments for better readability for someone new to this code.
Leaving final approval @kostaspap

internal/log_analysis/log_processor/logtypes/resolver.go
alxarch: added

internal/core/logtypesapi/ddbextras/ddbextras.go
```diff
@@ -0,0 +1,318 @@
...
+func IsCancelReason(r *dynamodb.CancellationReason, code string) bool {
+	if r == nil {
+		return false
+	}
+
+	switch c := aws.StringValue(r.Code); c {
+	case code:
+		return true
+	case strings.TrimSuffix(code, "Exception"):
+		return true
```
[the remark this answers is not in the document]
alxarch: There is inconsitency in the names of the constants in the dynamodb package. Some cancelation reasons are returned without the 'Exception' suffix from the API.

internal/log_analysis/datacatalog_updater/datacatalog/update_table.go
alxarch: This can happen in normal case when a custom log type is updated but it is not used by any sources.

kostaspap: gotcha

internal/core/logtypesapi/ddbextras/ddbextras.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+func (p *Put) Cancelled(r *dynamodb.CancellationReason) error {
+	if p.Cancel != nil {
+		return p.Cancel(r)
+	}
+	return nil
+}
+
+type Update struct {
```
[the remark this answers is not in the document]
alxarch: I am addin full comments in the ddbextras package.
The names will remain short as one of the major pain points of the 'official' dynamodb packages are `UnnecessarilyLongNamesAsIfThisWasJAVA`. Always think of the name of the type as `ddbextras.Update` and I would agree that `ddbextras` is a bad name. Maybe something like `transact` would be better.

internal/log_analysis/datacatalog_updater/datacatalog/update_table.go
giorgosp: >This can happen in normal case when a custom log type is updated but it is not used by any sources.
I think it would be helpful if you add this as a comment too

internal/core/logtypesapi/ddbextras/ddbextras.go
```diff
@@ -0,0 +1,318 @@
...
+	TableName    string
+	Key          interface{}
+	Set          map[string]interface{}
+	Add          map[string]interface{}
```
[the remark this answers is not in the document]
alxarch: These are actions of the update expression. An update expression modifies attributes. For more information about how dynamo db transactions work see [the API docs](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html)
I will add the link in the package doc comment

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
alxarch: This is resolved, removing the TODO

internal/core/logtypesapi/customlogs.go
kostaspap: It's less of an issue. If it reaches the SQS it will be retried for several hours before it gets driven to the DLQ.

internal/core/logtypesapi/customlogs.go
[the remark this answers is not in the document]
kostaspap: Sure

internal/log_analysis/datacatalog_updater/datacatalog/update_table.go
alxarch: Fixed

internal/log_analysis/datacatalog_updater/datacatalog/update_table.go
alxarch: Added comment

internal/core/logtypesapi/ddbextras/ddbextras.go
```diff
@@ -0,0 +1,318 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute"
+	"github.com/aws/aws-sdk-go/service/dynamodb/expression"
+)
+
+type WriteTransaction []BatchWriteItem
+
+func (b WriteTransaction) Check(err error) error {
```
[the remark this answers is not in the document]
alxarch: added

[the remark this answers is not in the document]
alxarch: Fixed

kostaspap (approved):

PR #2239, opened 2020-12-09
[FE]: Backport UTC changes

alexmylonas:
## Background

We needed to backport some changes regarding enabling the `DateRangerInput` to use UTC datetimes.

More details on https://github.com/panther-labs/panther-enterprise/pull/1615

## Changes

- Bump pounce to `v0.76.1`
- Update `DateRangeInput` to accept `useUTC` prop
- Updated TimeSeriesChart to accept `useUTC` prop
- Updated formatDatetime to accept `useUTC` arg
- Updated tests & snaps

## Testing

- Locally, we are not using UTC anywhere right now, so no change should be noticed


alexmylonas:
Except for the pounce bump, there is nothing here that could help current release. You think we should do it anyway?

3nvi (approved):
Looks good @alexmylonas. Thanks!

PR #2245, opened 2020-12-10
Generated Alert Fields - FE Support

wey-chiang:
## Background

Changes to the `alerts_api` to support FE work to properly display dynamic fields

## Changes

Main change in logic is in the `alerts_api/utils/utils.go` file:
- Created a new `getRule` function
- `getRule` is only called if any of the generated fields are missing (was having issues getting unit tests to work without this)

## Testing

- `mage test:ci`


wey-chiang:
@3nvi Would you guys be able to extract the `OutputID` from `DeliveryResponse`?  If not, I can extract it out and expose it as `Destinations`
Example JSON:
```
{
  "context": {
    "NULL": true
  },
  "creationTime": {
    "S": "2020-11-29T11:50:44Z"
  },
  "dedup": {
    "S": "TEST CUSTOM TITLE - NEW STUFFS"
  },
  "deliveryResponses": {
    "L": [
      {
        "M": {
          "dispatchedAt": {
            "S": "2020-11-29T11:50:47.405548707Z"
          },
          "message": {
            "S": "569f7a9d-d8e0-5dfe-8567-22a055b7adc7"
          },
          "outputId": {
            "S": "998b396d-10cf-4738-878e-55113701bb86"
          },
          "statusCode": {
            "N": "200"
          },
          "success": {
            "BOOL": true
          }
        }
      },
      {
        "M": {
          "dispatchedAt": {
            "S": "2020-11-29T11:50:47.405557107Z"
          },
          "message": {
            "S": "65b3a179-f4c8-588c-bf1d-f944799f59bd"
          },
          "outputId": {
            "S": "dde678a9-6a14-4f30-8bb2-9da6ea7b603f"
          },
          "statusCode": {
            "N": "200"
          },
          "success": {
            "BOOL": true
          }
        }
      }
    ]
  },
}
```

3nvi:
> @3nvi Would you guys be able to extract the `outputID` from `DeliveryResponse`?
> 
> If not, I can extract it out and expose it as `Destinations`
> 
> Example JSON:
> 
> ```
> {
>   "context": {
>     "NULL": true
>   },
>   "creationTime": {
>     "S": "2020-11-29T11:50:44Z"
>   },
>   "dedup": {
>     "S": "TEST CUSTOM TITLE - NEW STUFFS"
>   },
>   "deliveryResponses": {
>     "L": [
>       {
>         "M": {
>           "dispatchedAt": {
>             "S": "2020-11-29T11:50:47.405548707Z"
>           },
>           "message": {
>             "S": "569f7a9d-d8e0-5dfe-8567-22a055b7adc7"
>           },
>           "outputId": {
>             "S": "998b396d-10cf-4738-878e-55113701bb86"
>           },
>           "statusCode": {
>             "N": "200"
>           },
>           "success": {
>             "BOOL": true
>           }
>         }
>       },
>       {
>         "M": {
>           "dispatchedAt": {
>             "S": "2020-11-29T11:50:47.405557107Z"
>           },
>           "message": {
>             "S": "65b3a179-f4c8-588c-bf1d-f944799f59bd"
>           },
>           "outputId": {
>             "S": "dde678a9-6a14-4f30-8bb2-9da6ea7b603f"
>           },
>           "statusCode": {
>             "N": "200"
>           },
>           "success": {
>             "BOOL": true
>           }
>         }
>       }
>     ]
>   },
> }
> ```

Yeah I think that's what we already do

api/lambda/alerts/models/api.go
```diff
@@ -203,6 +203,10 @@ type AlertSummary struct {
 	PolicyVersion     string              `json:"policyVersion"`
 	ResourceTypes     []string            `json:"resourceTypes"`
 	ResourceID        string              `json:"resourceId"`
+	// Generated Fields Support
```
3nvi, line 206: Looks good. Just making sure we can't dynamically add `tags` to an alert. If we can't, what we have here is more than ok

wey-chiang: Nope, though that does sound like a useful feature -- would be pretty easy to add support for that further down the road

wey-chiang:
Converted to a PR, should be ready to go!

internal/log_analysis/alerts_api/utils/utils.go
```diff
@@ -21,36 +21,70 @@ package utils
...
 		alertType = alertdeliverymodels.RuleType
 	}
 
-	return &models.AlertSummary{
+	// Generated Fields - backwards compatibility support
+	description, reference, runbook := item.Description, item.Reference, item.Runbook
+
+	// Check if we have these fields to avoid an unnecessary API call
+	if aws.String(description) == nil || aws.String(reference) == nil || aws.String(runbook) == nil {
+		alertRule, err := getRule(item.AlertID, analysisClient)
```
nhakmiller, line 72: Don't we want the `RuleID` not the `AlertID` here? These are two different things, but should both be present in the alert.

internal/log_analysis/alerts_api/utils/utils.go
```diff
@@ -21,36 +21,70 @@ package utils
...
+		if err != nil || alertRule == nil {
+			zap.L().Warn("Failed to get Rule with ID ", zap.String("AlertID", item.AlertID))
```
nhakmiller, line 74: We can probably just return here, and then drop the `&& alertRule != nil` check from each of the following if statements.

internal/log_analysis/alerts_api/utils/utils.go
```diff
@@ -21,36 +21,70 @@ package utils
...
 
-	"github.com/panther-labs/panther/api/lambda/alerts/models"
+	alertmodels "github.com/panther-labs/panther/api/lambda/alerts/models"
+	analysismodels "github.com/panther-labs/panther/api/lambda/analysis/models"
 	alertdeliverymodels "github.com/panther-labs/panther/api/lambda/delivery/models"
 	"github.com/panther-labs/panther/internal/log_analysis/alerts_api/table"
+	"github.com/panther-labs/panther/pkg/gatewayapi"
 )
 
+func getRule(resourceID string, analysisClient gatewayapi.API) (*analysismodels.Rule, error) {
```
nhakmiller, line 33: I like this! Lets add a short lived cache here, if dozens/hundreds of alerts are being listed from the same few rules we could save ourselves a lot of API calls and busy waiting time. A cache with maybe a 2 minute lifetime would do the trick, we can just map the rule ID to a `analysismodels.Rule` + timestamp,  and only invoke the analysis-api  if  the rule isn't cached or is cached  & expired.

internal/log_analysis/alerts_api/utils/utils.go
wey-chiang: Good catch! Thanks, will fix

internal/log_analysis/alerts_api/utils/utils.go
wey-chiang: Changed the conditional & made this an if-else so the rest of the statements don't get evaluated

internal/log_analysis/alerts_api/utils/utils.go
```diff
@@ -20,37 +20,110 @@ package utils
...
+// Used to cache a RuleID to an analysismodels.Rule & timestamp
+var (
+	ruleCache        = make(map[ruleKey]cachedRule)
+	ruleCacheTimeout = -2 * time.Minute
 )
 
+// Key used for the rule cache to map to a Rule
+type ruleKey struct {
+	RuleID string
+}
```
nhakmiller, line 44: I think just using a string as the key directly is fine, this `ruleKey` struct is a bit overkill unless I'm missing something.

internal/log_analysis/alerts_api/utils/utils.go
```diff
@@ -20,37 +20,110 @@ package utils
...
 	if len(alertType) == 0 {
 		alertType = alertdeliverymodels.RuleType
 	}
 
-	return &models.AlertSummary{
+	// Generated Fields - backwards compatibility support
+	description, reference, runbook := item.Description, item.Reference, item.Runbook
+
+	// Check if we have these fields to avoid an unnecessary API call
+	if aws.String(description) == nil || aws.String(reference) == nil || aws.String(runbook) == nil {
```
nhakmiller, line 109: I think there might still be some confusion on when you need to use `aws.String()`. Take a look inside the function, it's fairly simple:

```
// String returns a pointer to the string value passed in.
func String(v string) *string {
	return &v
}
```

`aws.String()` is just returning the address of a string. The return value of `aws.String()` cannot ever be `nil`, so this statement will never be entered. The reason that the `aws.String(...)` function exists is because all AWS API calls take string pointers instead of strings, and in go you can't take the address of a string literal.

So since you can't do this:

```
response, err := snsClient.Publish(&sns.PublishInput{SnsTopic: &"myTopicName"}) // Invalid!
```

`aws.String()` lets you do this instead:

```
response, err := snsClient.Publish(&sns.PublishInput{SnsTopic: aws.String("myTopicName")})
```

That's the reason that the `aws.String()` function exists. If you're just trying to figure out if a string has the default string value, it's perfectly ok to have:

```
if description == "" || reference == "" || runbook == "" {
```

and in fact that is the only correct way to do this check. This should be updated here and in the below checks.

internal/log_analysis/alerts_api/utils/utils.go
```diff
@@ -20,37 +20,110 @@ package utils
...
+
+func getRule(resourceID string, analysisClient gatewayapi.API) (*analysismodels.Rule, error) {
+	// Key used to retrieve cached rule
+	cacheKey := ruleKey{
+		RuleID: resourceID,
+	}
+
+	// Return the cached short-lived rule (2 minutes)
+	if cachedRule, exists := ruleCache[cacheKey]; exists {
+		if time.Now().Add(ruleCacheTimeout).Before(cachedRule.Timestamp) {
```
nhakmiller, line 58: This logic looks right to me!

nhakmiller (commented):
Noticed something I missed on the first pass, sorry about that!

internal/log_analysis/alerts_api/utils/utils.go
```diff
@@ -20,37 +20,110 @@ package utils
...
-		result[i] = AlertItemToSummary(item)
+		result[i] = AlertItemToSummary(item, analysisClient)
 	}
 
 	return result
 }
 
 // AlertItemToSummary converts a DDB AlertItem to an AlertSummary
-func AlertItemToSummary(item *table.AlertItem) *models.AlertSummary {
+func AlertItemToSummary(item *table.AlertItem, analysisClient gatewayapi.API) *alertmodels.AlertSummary {
```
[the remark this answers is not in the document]
kostaspap: The caller could just reuse the RuleCache (see my other comment)

3nvi (commented):
API wise it looks ok from my side. Leaving it to the BE team for approvals

internal/log_analysis/alerts_api/utils/utils.go
```diff
@@ -22,35 +22,64 @@ package utils
...
+	"github.com/panther-labs/panther/api/lambda/analysis/models"
 	alertdeliverymodels "github.com/panther-labs/panther/api/lambda/delivery/models"
 	"github.com/panther-labs/panther/internal/log_analysis/alerts_api/table"
 )
 
 // AlertItemsToSummaries converts a list of DDB AlertItem(s) to AlertSummary(ies)
-func AlertItemsToSummaries(items []*table.AlertItem) []*models.AlertSummary {
-	result := make([]*models.AlertSummary, len(items))
+// func AlertItemsToSummaries(items []*table.AlertItem, ruleCache forwarder.RuleCache) []*alertmodels.AlertSummary {
+func AlertItemsToSummaries(alertItems []*table.AlertItem, alertRules []*models.Rule) []*alertmodels.AlertSummary {
```
kostaspap, line 33: We can have many AlertItems pointing to the same Rule. So the `alertRules` can end up having lots of duplicates. 
A way to avoid that is to change `getAlertRules` to return map[string]*models.Rule (ruleId+version -> rule). 
Your loop can for ever alertItem retrieve the corresponding rule (by looking into the map).

internal/log_analysis/alerts_api/api/api.go
```diff
@@ -97,3 +109,40 @@ func decodePaginationToken(token string) (*EventPaginationToken, error) {
...
+		return nil, err
+	}
+	return cachedRule, nil
+}
+
+func (api *API) getAlertRules(alerts []*table.AlertItem) []*models.Rule {
+	alertRules := make([]*models.Rule, 0, len(alerts))
+	for i, item := range alerts {
+		var err error
+		if item.Description == "" || item.Reference == "" || item.Runbook == "" {
```
kostaspap, line 128: The alert forwarder, when it creates a new alert item, it already queries the Rules API to fetch the Description, Reference, Runbook from a rule. 

My understanding is that this code is here only because we want to deal with backwards incompatibility issues: We want to make sure that for alerts created before we had the override capability, we will query the Rule to fetch the appropriate descriptions. 

However, this change seems to go beyond that... If someone has a Rule with empty Description or Reference or Runbook, the code will do the same work twice: 
1. Alert Forwarder will query the Analysis API and fetch the empty rule Description/Reference/Runnbok
2. Alerts API also query the Rules API on every FE request....

Is there some other way to identify if some alert is from "older" code? E.g. maybe it would make sense to make the item.Description, Reference, Runbook be nullable? That way, in older alerts they would be null, while for newer alerts they would be either empty string or populated. 

I am thinking we could have a method like:
```
func IsOldAlert(alert table.Item) bool{
 return item.Description == nil
}
```

with appropriate documentation. Then the code that needs to deal with "older" alerts, could just make use of that function to identify cases where we need to populate the values.

internal/log_analysis/alert_forwarder/forwarder/rule_cache.go
```diff
@@ -61,7 +61,7 @@ func cacheKey(id, version string) string {
 	return id + ":" + version
 }
 
-func (c *RuleCache) getRule(id, version string) (*models.Rule, error) {
+func (c *RuleCache) GetRule(id, version string) (*models.Rule, error) {
```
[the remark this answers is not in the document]
wey-chiang: Ahh this was a change I made in testing but did not intend to include here, will revert this back

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -93,7 +93,9 @@ func (api *API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOu
 		return nil, err
 	}
 
-	alertSummary := utils.AlertItemToSummary(alertItem)
+	alertRule, err := api.getRule(alertItem.RuleID, alertItem.RuleVersion)
```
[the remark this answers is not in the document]
wey-chiang: Changed to use `api.ruleCache.Get`
However, this should not fail if it returns nil -- the rule fields are used as fallbacks (the description, reference and runbook have precedence over the rule fields)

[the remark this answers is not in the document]
wey-chiang: Added a check to log any errors in retrieving the rule

internal/log_analysis/alerts_api/utils/utils.go
```diff
@@ -22,35 +22,64 @@ package utils
...
 	if len(alertType) == 0 {
 		alertType = alertdeliverymodels.RuleType
 	}
 
-	return &models.AlertSummary{
+	// Generated Fields - backwards compatibility support
+	description, reference, runbook := item.Description, item.Reference, item.Runbook
+
+	// Check if we have these fields to avoid an unnecessary API call
+	if description == "" || reference == "" || runbook == "" {
```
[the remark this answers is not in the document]
wey-chiang: I had this duplicated in other places because this function only accepted the description, reference and runbook as parameters -- I'll remove those dupes in other places but I believe this if-statement is necessary here for backwards compatibility

internal/log_analysis/alerts_api/utils/utils_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+type mockDynamoDB struct {
+	dynamodbiface.DynamoDBAPI
+	mock.Mock
+}
+
+func (m *mockDynamoDB) GetItem(input *dynamodb.GetItemInput) (*dynamodb.GetItemOutput, error) {
+	args := m.Called(input)
+	return args.Get(0).(*dynamodb.GetItemOutput), args.Error(1)
+}
```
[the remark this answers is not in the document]
wey-chiang: Nice!

internal/log_analysis/alerts_api/api/api.go
```diff
@@ -97,3 +109,40 @@ func decodePaginationToken(token string) (*EventPaginationToken, error) {
...
+			alertRules[i], err = api.getRule(item.RuleID, item.RuleVersion)
+			if err != nil {
+				alertRule := &models.Rule{
+					Description: "",
+					Reference:   "",
+					Runbook:     "",
+				}
+				alertRules = append(alertRules, alertRule)
+			}
```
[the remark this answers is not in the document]
wey-chiang: I am not sure if that is true (return wrong info to the UI)

Prior to the implementation of dynamic alert fields, alerts did not include description, reference, or runbook. Instead, the FE used the `panther-analysis` API to get these fields. 

Since we are now including them in alerts, these alert fields should have these fields populated already with the dynamically generated fields (including those 3)  and falling back to the rule when there was an error with this logging. 

I don't think we really need this except for backwards compatibility -- so in my mind, we should continue to process regardless of whether or not we are able to retrieve the rule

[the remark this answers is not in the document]
kostaspap: Right, so let's say a user looks at an old alert. The FE makes a request to the backend, and the backend fails while trying to retrieve the Runbook/Description/Reference of that rule.  If I read the code correctly, we will show the user empty strings instead of the Runbook/Description/Reference of the Rule, which is inaccurate, right?

internal/log_analysis/alerts_api/utils/utils.go
wey-chiang: Nice! Yeah that makes a lot of sense

internal/log_analysis/alerts_api/api/api.go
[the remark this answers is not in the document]
wey-chiang: Yes, that would be the case if the Runbook/Description/Reference were not populated in the alert itself -- which should almost never be the case (except for backwards compatibility) because of the changes made to the alert-forwarder.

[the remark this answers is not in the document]
kostaspap: I see, so retrieving Runbook/Description/Reference for "old" alerts is basically best effort. This is new behavior, previously it was not best effort. I'm fine either way, as long as we think that's the appropriate customer experience.

internal/log_analysis/alerts_api/utils/utils.go
```diff
@@ -21,36 +21,62 @@ package utils
...
+	alertmodels "github.com/panther-labs/panther/api/lambda/alerts/models"
+	"github.com/panther-labs/panther/api/lambda/analysis/models"
 	alertdeliverymodels "github.com/panther-labs/panther/api/lambda/delivery/models"
 	"github.com/panther-labs/panther/internal/log_analysis/alerts_api/table"
 )
 
 // AlertItemsToSummaries converts a list of DDB AlertItem(s) to AlertSummary(ies)
-func AlertItemsToSummaries(items []*table.AlertItem) []*models.AlertSummary {
-	result := make([]*models.AlertSummary, len(items))
+// func AlertItemsToSummaries(items []*table.AlertItem, ruleCache forwarder.RuleCache) []*alertmodels.AlertSummary {
```
nhakmiller, line 33: Leftover comment it looks like.

internal/log_analysis/alerts_api/utils/utils.go
```diff
@@ -21,36 +21,62 @@ package utils
...
 	}
 	alertType := item.Type
 	if len(alertType) == 0 {
 		alertType = alertdeliverymodels.RuleType
 	}
 
-	return &models.AlertSummary{
+	// Generated Fields - backwards compatibility support
+	description, reference, runbook := item.Description, item.Reference, item.Runbook
+	if description == nil || reference == nil || runbook == nil {
```
nhakmiller, line 67: I understand that this particular `if` statement has been re-worked a lot, but I think as it stands it makes most sense to remove it.

Each of the individual statuses are just re-checked inside,  so if we leave the three internal `if` statements and remove this parent one I don't the it will be functionally different.

nhakmiller (approved):
Seems good to me, just a few nitpicks.

internal/log_analysis/alerts_api/api/api.go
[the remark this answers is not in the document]
kostaspap: I checked the code again. I see that this is not consistent with that we do in other parts of the code. This has to be reverted to fail. 
E.g. the GetAlert will fail in case of failure to fetch the rule information. This has to be fixed to be consistent. 

@wey-chiang

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -93,7 +93,14 @@ func (api *API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOu
 		return nil, err
 	}
 
-	alertSummary := utils.AlertItemToSummary(alertItem)
+	alertRule, err := api.ruleCache.Get(alertItem.RuleID, alertItem.RuleVersion)
```
kostaspap, line 96: This means we make I/O operation even though we don't need to. We should only be fetching rule description etc for old alerts, not for every alert.

PR #2247, opened 2020-12-10
Send Snapshot information to datalake

### LINKED ISSUE #2031: Cloud Security events get processed by panther-log-processor

kostaspap:
## Background

Closes #2031 

## Changes

- CFN template now sets up the CloudSecurity forwarder Lambda
- Introducing some new `SourceIntegration` functions that allow centralizing the logic of getting S3Prefix, S3Bucket, LogProcessingRole etc . Before the  logic was scattered all over the code. 
- Fixing some bugs I found in the datalake forwarder. I found it easier to troubleshoot the bugs by writing some unit tests. I had to refactor a bit the logic to make it easy to write those tests. 

## Testing

- mage test:ci
- Deployed and verified I can see data in the cloud security database


api/lambda/source/models/integration.go
```diff
@@ -118,8 +146,6 @@ type SqsConfig struct {
 
 	// The Panther-internal S3 bucket where the data from this source will be available
 	S3Bucket string `json:"s3Bucket"`
-	// The S3 prefix where the data from this source will be available
-	S3Prefix string `json:"s3Prefix"`
```
kostaspap, line 122: This is not needed - the prefix for SQS is anyway a constant.

internal/compliance/datalake_forwarder/forwarder/compliance.go
```diff
@@ -73,24 +70,17 @@ func (sh *StreamHandler) processComplianceSnapshot(record *events.DynamoDBEventR
...
 
-func dynamoRecordToCompliance(image map[string]*dynamodb.AttributeValue) (*ComplianceChange, error) {
+func (sh *StreamHandler) recordToCompliance(image map[string]*dynamodb.AttributeValue) (*ComplianceChange, error) {
 	change := ComplianceChange{}
-	if err := dynamodbattribute.UnmarshalMap(image, &change); err != nil {
-		return nil, err
-	}
-	if err := validate.Struct(&change); err != nil {
-		return nil, err
-	}
```
kostaspap, line 94: I am not sure this validation is required here (note that we don't do anything similar for `ResourceChange`). 
1. These should be only in exceptional cases
2. The log processor does this type of validation. It will be easier for debugging purposes to just sent everything to the log processor and if classification fails there, we try to understand why that happened.

internal/compliance/datalake_forwarder/forwarder/forward.go
```diff
@@ -31,18 +30,16 @@ import (
...
-	"github.com/panther-labs/panther/pkg/oplog"
 )
 
 const (
-	ChangeTypeCreate = "created"
-	ChangeTypeDelete = "deleted"
-	ChangeTypeModify = "modified"
+	ChangeTypeCreate = "CREATED"
+	ChangeTypeDelete = "DELETED"
+	ChangeTypeModify = "MODIFIED"
```
kostaspap, line 40: `Status` was `FAIL`|| `PASS`, `Severity` was `INFO`||`HIGH`|| ... 
Seemed appropriate to keep the same capitalization here as well.

internal/compliance/datalake_forwarder/forwarder/forward.go
```diff
@@ -54,29 +51,23 @@ type StreamHandler struct {
 }
 
 // Run is the entry point for the datalake-forwarder lambda
-func (sh *StreamHandler) Run(ctx context.Context, event *events.DynamoDBEvent) (err error) {
-	lc, logger := lambdalogger.ConfigureGlobal(ctx, nil)
-	operation := oplog.NewManager("cloudsec", "datalake_forwarder").Start(lc.InvokedFunctionArn).WithMemUsed(lambdacontext.MemoryLimitInMB)
-	defer func() {
-		operation.Stop().Log(err, zap.Int("numEvents", len(event.Records)))
-	}()
```
kostaspap, line 62: Moved these out of this method and into `main` to simplify unit tests. Unit tests don't need to depend on Lambda runtime

internal/compliance/datalake_forwarder/forwarder/forward.go
```diff
@@ -54,29 +51,23 @@ type StreamHandler struct {
...
 				zap.Error(err),
 				zap.String("eventID", record.EventID),
 				zap.String("eventName", record.EventName),
 				zap.String("eventSourceARN", record.EventSourceArn),
 			)
 			continue
 		}
-		if changes == nil {
-			logger.Warn("Skipping record",
+		if changes == interface{}(nil) {
```
kostaspap, line 69: This took a while to figure out....

internal/compliance/datalake_forwarder/forwarder/forward.go
```diff
@@ -86,16 +77,16 @@ func (sh *StreamHandler) Run(ctx context.Context, event *events.DynamoDBEvent) (
 		}
 		data, err := jsoniter.Marshal(changes)
 		if err != nil {
-			logger.Error("failed to get marshal changes to JSON", zap.Error(err), zap.String("eventId", record.EventID))
+			log.Error("failed to get marshal changes to JSON", zap.Error(err), zap.String("eventId", record.EventID))
 			continue
 		}
-		// TODO: [JSONL] Adding newline here should not be required if the log processor can handle JSON streams
+
```
kostaspap, line 83: Removed this TODO. Even if we support JSON streams in the log processor, we will still want to write the files in JSONL format. Makes it easier to query this data directly (if we ever need to) and makes debugging easier as well.

internal/compliance/datalake_forwarder/forwarder/resources.go
```diff
@@ -26,11 +26,11 @@ import (
...
 	"go.uber.org/zap"
 
-	"github.com/panther-labs/panther/internal/compliance/datalake_forwarder/forwarder/internal/diff"
-	"github.com/panther-labs/panther/internal/compliance/datalake_forwarder/forwarder/internal/events"
+	"github.com/panther-labs/panther/internal/compliance/datalake_forwarder/forwarder/diff"
+	"github.com/panther-labs/panther/internal/compliance/datalake_forwarder/forwarder/events"
 )
 
-type CloudSecuritySnapshotChange struct {
+type ResourceChange struct {
```
kostaspap, line 33: The corresponding struct for compliance was named `ComplianceChange`, felt appropriate to rename.

internal/compliance/datalake_forwarder/forwarder/source_api.go
```diff
@@ -33,30 +33,28 @@ var (
...
 
 // Returns the label for an integration
 // It will return an empty string if the integration doesn't exist
 func (sh *StreamHandler) getIntegrationLabel(integrationID string) (string, error) {
-	_, ok := integrationIDMappings[integrationID]
-	if !ok || time.Since(lastUpdated) > mappingAgeOut {
-		err := sh.updateIntegrationMapping()
-		if err != nil {
+	if time.Since(lastUpdated) > mappingAgeOut {
+		if err := sh.updateIntegrationMapping(); err != nil {
```
kostaspap, line 44: The `if !ok` condition had the potential to cause lots of troubles if we would receive ever data from unregistered resources due to any bug. Just stick to refreshing labels every 1 minute.

internal/compliance/datalake_forwarder/forwarder/forward.go
```diff
@@ -54,29 +51,23 @@ type StreamHandler struct {
...
 				zap.Error(err),
 				zap.String("eventID", record.EventID),
 				zap.String("eventName", record.EventName),
 				zap.String("eventSourceARN", record.EventSourceArn),
 			)
 			continue
 		}
-		if changes == nil {
-			logger.Warn("Skipping record",
+		if !ok {
```
kostaspap, line 69: Previous check did not work as expected - we would return a `interface{}nil`, and the check here would fail, even if we had no changes...

internal/compliance/datalake_forwarder/forwarder/forward.go
```diff
@@ -31,52 +31,46 @@ import (
...
+func (sh *StreamHandler) Run(ctx context.Context, log *zap.Logger, event *events.DynamoDBEvent) (err error) {
 	firehoseRecords := make([]*firehose.Record, 0, len(event.Records))
 	for i := range event.Records {
 		// We should be passing pointers to avoid copy of the record struct
 		record := &event.Records[i]
-		changes, err := sh.getChanges(record)
+		changes, ok, err := sh.getChanges(record)
 		if err != nil {
-			logger.Error("failed to process record",
+			log.Error("failed to process record",
```
giorgosp, line 64: nit: Maybe we should stick to `logger` to differentiate from the std package?

internal/core/source_api/ddb/items.go
```diff
@@ -56,7 +56,6 @@ type IntegrationStatus struct {
 
 type SqsConfig struct {
 	S3Bucket             string   `json:"s3Bucket,omitempty"`
-	S3Prefix             string   `json:"s3Prefix,omitempty"`
```
giorgosp, line 59: Perhaps this is displayed in the frontend?

api/lambda/source/models/integration.go
```diff
@@ -66,14 +66,45 @@ type SourceIntegrationMetadata struct {
...
+	switch s.IntegrationType {
+	case IntegrationTypeSqs:
+		return s.SqsConfig.LogProcessingRole
+	default:
+		return s.LogProcessingRole
+	}
+}
+
+func (s *SourceIntegration) RequiredS3Prefix() string {
+	const sqsS3Prefix = "forwarder"
```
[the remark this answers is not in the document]
kostaspap: Fair. Updated

api/lambda/source/models/integration.go
```diff
@@ -66,14 +66,45 @@ type SourceIntegrationMetadata struct {
...
-		return info.SqsConfig.LogTypes
+	case IntegrationTypeSqs:
+		return s.SqsConfig.LogTypes
 	default:
-		return info.LogTypes
+		return s.LogTypes
+	}
+}
+
+func (s *SourceIntegration) RequiredLogProcessingRole() string {
```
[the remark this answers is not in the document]
kostaspap: There is an overlap with the `LogProcessingRole` field inside the `SourceIntegration`

internal/compliance/datalake_forwarder/forwarder/forward.go
kostaspap: Good idea. Updating

internal/core/source_api/ddb/items.go
kostaspap: Nope, not displayed anywhere.

api/lambda/source/models/integration.go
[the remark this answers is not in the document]
giorgosp: Ah right. We could make the field private but then the json serde won't work. Ideally we would have separate structs for the api response, the main domain object and the db but it's an overkill now. We just have to put up with it I guess.

api/lambda/source/models/integration.go
```diff
@@ -66,14 +66,43 @@ type SourceIntegrationMetadata struct {
...
+		return "cloudsecurity"
+	default:
+		return s.S3Prefix
+	}
+}
+
+func (s *SourceIntegration) RequiredS3Bucket() string {
+	switch s.IntegrationType {
+	case IntegrationTypeSqs:
+		return s.SqsConfig.S3Bucket
```
rleighton, line 103: For safety, should there be as case for IntegrationTypeAWSScan with a panic() if it makes no sense?

internal/compliance/datalake_forwarder/forwarder/forward.go
```diff
@@ -31,41 +31,35 @@ import (
...
 )
 
 const (
-	ChangeTypeCreate = "created"
-	ChangeTypeDelete = "deleted"
-	ChangeTypeModify = "modified"
+	ChangeTypeCreate = "CREATED"
+	ChangeTypeDelete = "DELETED"
+	ChangeTypeModify = "MODIFIED"
 	// TODO add daily syncs
```
rleighton, line 42: I thought this was done ... this is a big gap from a usability POV. We need daily sync's to make the data useable to track history for a resource w/out joins to the resources table.

One of the most important questions is "show me all of my resources AND their changes for a day" ... w/out a daily sync we cannot do this w/out joins. 

Is this hard to add? Once day we want to dump all records and mark them SYNC

api/lambda/source/models/integration.go
kostaspap: Fair. Updating.

internal/compliance/datalake_forwarder/forwarder/forward.go
kostaspap: Not 100% sure about that. Maybe @nhakmiller has more context.

api/lambda/source/models/integration.go
kostaspap: Done

api/lambda/source/models/integration.go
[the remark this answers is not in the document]
kostaspap: Exactly. Maybe we can name it something shorter e.g. `GetBlah` or something. i just followed the pattern we already had with the LogTypes.

internal/compliance/datalake_forwarder/forwarder/forward.go
nhakmiller: Hmm yeah I believe we discussed the easiest fix is to remove this if statement:

https://github.com/panther-labs/panther/pull/2247/files#diff-6ea78b1df1e3d0a9e44c0b876815ca45b75a4b4999b9197c7a47f5a5b369c4b7R109-R112

Then every day during the daily scan we'd get the new snapshots. We had also considered writing a new lambda to pull in snapshots once a day specifically for this purpose, but we weren't sure about if/when/how this feature was going to happen so we postponed that. If we're ok removing the above if statement that will get the desired functionality, or if you want it to be a standalone thing I can take that in a separate follow up PR.

nhakmiller (approved):
Thanks for cleaning this up and taking it over the finish line, sorry I left you so many bugs.

deployments/cloud_security.yml
```diff
@@ -1416,3 +1427,136 @@ Resources:
...
+    Properties:
+      DeliveryStreamName: panther-cloudsecurity-firehose
+      DeliveryStreamType: DirectPut
+      ExtendedS3DestinationConfiguration:
+        BucketARN: !Sub arn:${AWS::Partition}:s3:::${InputDataBucket}
+        Prefix: cloudsecurity/
+        BufferingHints:
+          # Data is flushed once one of the buffer hints are satisfied.
+          IntervalInSeconds: 60
+          SizeInMBs: 128
```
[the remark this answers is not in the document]
kostaspap: This is just the Firehose buffer size :) Not the Lambda trigger size! (I know, I read it wrong the first time I saw it!)

internal/compliance/datalake_forwarder/forwarder/forward.go
kostaspap: No worries, I've updated this PR to include syncs

internal/compliance/datalake_forwarder/forwarder/diff/comparisons.go
```diff
@@ -34,8 +34,8 @@ type Changelog map[string]Diff
 
 // CompJsons compares two JSON strings, and returns a map of paths to changed fields to a tuple of
 // the old and new value of that field
-func CompJsons(left, right string) (map[string]Diff, error) {
-	return compGJsons(gjson.Parse(left), gjson.Parse(right), "")
+func CompJsons(left, right []byte) (map[string]Diff, error) {
```
kostaspap, line 37: By passing in []byte instead of string, we avoid some extra copies of the data in memory. This is particularly important in here, since the objects can be quite large (hundreds of KBs each)

PR #2248, opened 2020-12-10
Implement diff, merge, infer for logschema

alxarch:
## Background

The functionality included in this PR will be used in customlog updates and to infer schemas from samples.
Code is self-contained, only exposing new functionality for the `logschema` package.

## Changes

- Adds `logschema.Diff`, to be used to validate schema updates.
- Adds `logschema.Merge` to be used for auto-generating schemas from samples.
- Adds `logschema.InferJSONValueSchema` to be used for auto-generating schemas from samples.

## Testing

- mage test:go


internal/log_analysis/log_processor/logschema/diff.go
```diff
@@ -0,0 +1,257 @@
...
+		d = append(d, FieldDiff{B: f})
+	}
+	for i := range a {
+		fieldA := &a[i]
+		fieldB := findField(fieldA.Name, b)
+		if fieldB == nil {
+			continue
+		}
+		d = append(d, FieldDiff{A: fieldA, B: fieldB})
+	}
```
kostaspap, line 211: Can you add some comments here explaining the logic?

alxarch:
@kostaspap The infer functionality included here is to demonstrate a maintainable structure for schema inference. The `Diff` stuff is used by updates also.

internal/log_analysis/log_processor/logschema/infer.go
```diff
@@ -0,0 +1,125 @@
...
+			Type:    TypeArray,
+			Element: merged,
+		}
+	case float64:
+		if v != v { // NaN
+			return nil
+		}
+		if float64(int64(v)) == v {
+			return &ValueSchema{Type: TypeBigInt}
+		}
```
kostaspap, line 70: Why is this check needed? We wouldn't want to interpret floats with decimal part 0 e.g `1.0` to be inferred as BigInt

internal/log_analysis/log_processor/logschema/logschema.go
```diff
@@ -57,6 +59,13 @@ type Schema struct {
 	Fields       []FieldSchema           `json:"fields" yaml:"fields"`
 }
 
+func (s *Schema) Clone() *Schema {
+	data, _ := jsoniter.Marshal(s)
+	out := Schema{}
+	_ = jsoniter.Unmarshal(data, &out)
+	return &out
```
kostaspap, line 66: We shouldn't ignore the errors. This shouldn't happen though, so i think we could just panic.

internal/log_analysis/log_processor/logschema/logschema.go
```diff
@@ -57,6 +59,13 @@ type Schema struct {
 	Fields       []FieldSchema           `json:"fields" yaml:"fields"`
 }
 
+func (s *Schema) Clone() *Schema {
```
kostaspap, line 62: if s == nil return nil?

kostaspap (commented):
I imagine the unit tests (there should be a lot of them) will come in a separate PR, right?

internal/log_analysis/log_processor/logschema/diff.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+const (
+	AddField        = "AddField"
+	DeleteField     = "DeleteField"
+	UpdateFieldMeta = "UpdateFieldMeta"
+	UpdateValue     = "UpdateValue"
+	UpdateValueMeta = "UpdateValueMeta"
+	UpdateParser    = "UpdateParser"
+	UpdateMeta      = "UpdateMeta"
```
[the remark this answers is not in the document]
alxarch: Yup, cleaning up and commenting now.

internal/log_analysis/log_processor/logschema/infer.go
alxarch: Yup, needs better check.

alxarch: The thing is that this check is not possible unless we use `json.Number`. AFAIK the `.0` info is only available in the string form of the float. This means that the caller should be using `json.Number` if they want proper int detection. Maybe we remove the check altogether.

internal/log_analysis/log_processor/logschema/logschema.go
alxarch: Yup, we want an exact clone. This is used in contexts where Schema can be nil. It is not the job of this function to panic here. It's job is to produce an exact clone of the input.

internal/log_analysis/log_processor/logschema/logschema.go
alxarch: Fixed

internal/log_analysis/log_processor/logschema/logschema.go
alxarch: Fixed.

internal/log_analysis/log_processor/logschema/infer.go
alxarch: I updated the function documentation and I think the best option here is to just return `TypeFloat` when the value is `float64`.
If the user wants to distinguish between `float` and `int` they should use `json.Number`. Event then I think for inference the only way to properly detect floats is to have many samples. If even one of them has a fractional part, the value will end up as `TypeFloat`.

internal/log_analysis/log_processor/logschema/diff.go
alxarch: Fixed

internal/log_analysis/log_processor/logschema/diff.go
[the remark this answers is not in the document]
alxarch: Fixed

internal/log_analysis/log_processor/logschema/merge.go
```diff
@@ -0,0 +1,154 @@
...
+		return &ValueSchema{
+			Type: TypeString,
+		}
+	default:
+		return a
+	}
+}
+
+// castValue handles merging values of different types while trying to preserve value information.
+// This function assumes a.Type != b.Type
```
kostaspap, line 80: Can you just have it panic or return error if a.Type == b.Type?

internal/log_analysis/log_processor/logschema/merge.go
```diff
@@ -0,0 +1,154 @@
...
+	default:
+		return &ValueSchema{Type: TypeString}
+	}
+}
+
+func castInt(a, b *ValueSchema) *ValueSchema {
+	switch b.Type {
+	case TypeSmallInt:
+		return a.Clone()
+	default:
```
kostaspap, line 151: What about the Float case? We can cast an Int to Float.

internal/log_analysis/log_processor/logschema/merge.go
```diff
@@ -0,0 +1,154 @@
...
+		return &ValueSchema{Type: TypeString}
+	}
+}
+
+func castTimestamp(a, b *ValueSchema) *ValueSchema {
+	switch b.Type {
+	case TypeBigInt:
+		switch a.TimeFormat {
+		case "unix", "unix_ms", "unix_us", "unix_ns":
+			return a.Clone()
```
kostaspap, line 118: Seems that this should be switch to return TypeBigInt instead of Timestamp.

internal/log_analysis/log_processor/logschema/merge.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	case TypeFloat:
+		if a.TimeFormat == "unix" {
+			return a.Clone()
```
kostaspap, line 122: This should be TypeFloat? Casting a timestamp (unix) -> TypeFloat, should return Float

internal/log_analysis/log_processor/logschema/merge.go
```diff
@@ -0,0 +1,154 @@
...
+	return &ValueSchema{Type: TypeString}
+}
+
+func castBigInt(a, b *ValueSchema) *ValueSchema {
+	switch b.Type {
+	case TypeInt, TypeSmallInt:
+		return a.Clone()
+	default:
+		return &ValueSchema{Type: TypeString}
+	}
```
kostaspap, line 135: This seems a bit weird. if b is of type object, we will return TypeString ?

internal/log_analysis/log_processor/logschema/merge_test.go
```diff
@@ -0,0 +1,110 @@
...
+		{"Array,String", &V{Type: TypeArray}, &V{Type: TypeString}, &V{Type: TypeJSON}},
+		{"Array,Float", &V{Type: TypeArray}, &V{Type: TypeFloat}, &V{Type: TypeJSON}},
+		{"Array,BigInt", &V{Type: TypeArray}, &V{Type: TypeBigInt}, &V{Type: TypeJSON}},
+		{"Array,Int", &V{Type: TypeArray}, &V{Type: TypeInt}, &V{Type: TypeJSON}},
+		{"Array,SmallInt", &V{Type: TypeArray}, &V{Type: TypeSmallInt}, &V{Type: TypeJSON}},
+		{"Array,Boolean", &V{Type: TypeArray}, &V{Type: TypeBoolean}, &V{Type: TypeJSON}},
+
+		{"UnixTimestamp,Float", &V{Type: TypeTimestamp, TimeFormat: "unix"}, &V{Type: TypeFloat}, &V{Type: TypeTimestamp, TimeFormat: "unix"}},
+		{"UnixTimestampEvent,Float", &V{Type: TypeTimestamp, TimeFormat: "unix", IsEventTime: true}, &V{Type: TypeFloat}, &V{Type: TypeTimestamp, TimeFormat: "unix", IsEventTime: true}},
+		{"UnixTimestampMS,Float", &V{Type: TypeTimestamp, TimeFormat: "unix_ms"}, &V{Type: TypeFloat}, &V{Type: TypeString}},
```
kostaspap, line 65: unix -> float, seems it should just be float? `unix` timestamp is anyway just number

alxarch: The `unix` tcodec can handle floats for the fractional part of seconds. It is quite common way to express timestsamp in seconds with better resolution.

internal/log_analysis/log_processor/logschema/merge.go
alxarch: No, the `unix` tcodec handles floats setting the fractional part of the number to fractions of a second. All other `unix_` codecs require integer values.

internal/log_analysis/log_processor/logschema/merge.go
alxarch: The idea is that if we have a 'base' value schema set by the user we want to preserve timestamp information if possible. This does not affect the 'just infer schema case without a base' since it can only produce `rfc3339` timestamps.

internal/log_analysis/log_processor/logschema/merge.go
alxarch: The cast functions are in order (as mentioned in the comment in `castValue`) when we reach this point the `Float` case is already handled by `castFloat`. You can think of it as (each value handles conversion for 'lesser' values).

internal/log_analysis/log_processor/logschema/merge.go
alxarch: I can add a panic but it is an unexported function because of all the assumptions it makes.

internal/log_analysis/log_processor/logschema/merge.go
rleighton: Can you add that explanation to the comment above `castValue()`? That would help.

internal/log_analysis/log_processor/logschema/merge_test.go
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// nolint:lll
+func TestMerge(t *testing.T) {
```
rleighton, line 28: Might be good to run the test with `-cover` and see if you have all the core cases exercised.

internal/log_analysis/log_processor/logschema/merge_test.go
kostaspap: > The `unix` tcodec can handle floats for the fractional part of seconds. It is quite common way to express timestsamp in seconds with better resolution.

Yeah I am aware of that. But if I understand the test correct, the result of casting a Timestamp (UnixMs) to Float is a String. Do I read it wrong?

internal/log_analysis/log_processor/logschema/diff.go
```diff
@@ -0,0 +1,285 @@
...
+
+// DiffWalk recursively iterates two value schemas and call `walk` when a change is found.
+// The `basePath` argument is used prepended to the path on each change.
+func DiffWalk(a, b *ValueSchema, walk func(c Change) bool, basePath ...string) {
+	diffWalk(a, b, walk, basePath)
+}
+
+// diffWalk recursively iterates two value schemas and call `walk` when a change is found.
+// The `path` argument is used to attach path information on each change.
+func diffWalk(a, b *ValueSchema, walk func(c Change) bool, path []string) bool {
```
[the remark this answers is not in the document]
alxarch: It aborts the walk if false.

internal/log_analysis/log_processor/logschema/merge_test.go
alxarch: 100% coverage

internal/log_analysis/log_processor/logschema/diff.go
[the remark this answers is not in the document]
kostaspap: Can you add this to the description?

internal/log_analysis/log_processor/logschema/logschema.go
```diff
@@ -57,6 +59,23 @@ type Schema struct {
...
+	if s == nil {
+		return nil
+	}
+	data, err := jsoniter.Marshal(s)
```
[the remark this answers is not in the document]
alxarch: The problem is cyclic structures. 'deep copy' can only happen for stuff that cannot contain pointers to itself

internal/log_analysis/log_processor/logschema/merge.go
kostaspap: Ok, I see that following the logic, we can never arrive here. Resolving question.

internal/log_analysis/log_processor/logschema/diff.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	DiffWalk(valueA, valueB, func(ch Change) bool {
+		c.changes = append(c.changes, ch)
+		return true
+	}, "Fields")
+
+	return c.changes, nil
+}
+
+// DiffWalk recursively iterates two value schemas and call `walk` when a change is found.
```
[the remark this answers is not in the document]
alxarch: fixed

internal/log_analysis/log_processor/logschema/diff.go
```diff
@@ -0,0 +1,285 @@
...
+// The `basePath` argument is used prepended to the path on each change.
+func DiffWalk(a, b *ValueSchema, walk func(c Change) bool, basePath ...string) {
+	diffWalk(a, b, walk, basePath)
+}
+
+// diffWalk recursively iterates two value schemas and call `walk` when a change is found.
```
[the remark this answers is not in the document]
alxarch: fixed

internal/log_analysis/log_processor/logschema/diff.go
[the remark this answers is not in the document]
alxarch: fixed

internal/log_analysis/log_processor/logschema/diff.go
```diff
@@ -0,0 +1,285 @@
...
+			}
+			return walk(ch)
+		}
+		return true
+	default:
+		return true
+	}
+}
+
+func walkObject(a, b []FieldSchema, walk func(c Change) bool, path []string) bool {
```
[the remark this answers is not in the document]
alxarch: fixed

internal/log_analysis/log_processor/logschema/merge.go
alxarch: I moved the cast logic inside merge. (with the comments)

internal/log_analysis/log_processor/logschema/merge.go
alxarch: Moved conversion logic inside Merge

internal/log_analysis/log_processor/logschema/diff.go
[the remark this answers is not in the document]
alxarch: Added a comment on diffWalk. the returning value of `walk` is what controls the recursion. this is quite common signature for such a function. Like the 'visitor' pattern.

internal/log_analysis/log_processor/logschema/merge_test.go
alxarch: Timestamp(unix) + Float -> Timestamp(unix)
Timestamp(unix_ms) + Float -> String. The 'special case' is only `unix`.

alxarch: In general the idea is that Timestamp(unix) is something specific provided probably by human intervention. Float is what the 'dumb' infer code sees. So we preserve such 'external' hints.

kostaspap: Ok, I see you were taking different approach. The way I was seeing it: we tried to "infer" that a field is Timestamp(unix) by just seeing the sample logs. We then try to merge it with a schema where the same field is a float (maybe because we deemed that the value cannot be a timestamp due to being beyond some limits). In that case, the Float "wins". 
 
I see that you consider an approach where we might want to merge an existing schema with some new schemas. In that case, it make sense. The important bit here is that our inference should never try to guess that a timestamp is unix timestamp (e.g. by checking the name of the field)

alxarch: You are right, I fixed the code. The assumption I was making was that we would be able to control types in a granular way so that we could seamlessly update schemas. But Glue/parquet restrictions do not allow any type changes during updates so there is no point in preserving compatibility through `string`.

alxarch:
> I imagine the unit tests (there should be a lot of them) will come in a separate PR, right?

Yes, I have high level tests with complex structs. But they use the testdata schemas and are bound to verifying the Changelog returned by Diff.
The merge is thoroughly tested here.
I will try to have similar tests for Diff as well. As is the case with recursive code we cannot test all possible nesting. So we need to rely on good structure to get full coverage.

kostaspap (approved):

PR #2256, opened 2020-12-10
Change "resource no longer exists" log from Warn() to Debug()

rleighton:
## Before

During delete  "resource no longer exists" errors were logged at warn level but this creates alarms when we get spikes of warnings above threshold.

## After

- Change "resource no longer exists" log from Warn() to Debug()


nhakmiller (approved):
I suppose it is the responsibility of the caller, not the API itself, to determine whether requesting a non-existent resource is unusual enough to warrant a `Warn` message.

PR #2258, opened 2020-12-10
Add Snyk docker container scanning

### LINKED ISSUE #1815: Snyk Setup
### Description

Snyk is a DevSecOps platform that identifies security issues in application dependencies, containers, and infra-as-code. Today, we use a combination of various tools to scan application code (CodeQL) and dependencies (bandit, gosec, and npm-audit), but do not have a tool for container scanning of our web application.

This epic is to track our setup and configuration of Snyk.

### RFC

N/A

### Designs

N/A

### Acceptance Criteria

- Snyk can build and scan our frontend Docker image in CI ([example](https://github.com/snyk/actions/tree/master/docker))
- Snyk can scan our Terraform files for vulnerabilities in CI ([example](https://github.com/snyk/actions/tree/master/iac))
- Snyk can scan our OSS licenses in CI ([might be this?](https://github.com/snyk/actions/tree/master/setup))

s0l0ist:
## Background

Adding support for container vulnerability scanning. This is specifically for our docker container that serves the FE.

Closes #1815 

For local web development, this change requires node `v14.15.x` (LTS). Using `nvm`, simply run `nvm install --lts`

## Changes

- Adds github workflow for container scanning
- Upgrade to `node:14.15-alpine`
- Remove FOSSA

## Testing

- `brew install act`, test github actions locally.


.github/workflows/docker-analysis.yml
```diff
@@ -0,0 +1,63 @@
...
+
+name: 'Docker Analysis'
+on:
+  pull_request:
+    branches:
+      - master
+  schedule:
+    # Run every Sunday at 3pm
+    # https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#schedule
+    - cron: '0 15 * * 0'
```
austinbyers, line 25: Interesting, so we're running both on a schedule and on every PR? We need both?

.github/workflows/docker-analysis.yml
```diff
@@ -0,0 +1,63 @@
...
+          fetch-depth: 2
+
+      # If this run was triggered by a pull request event (and not a scheduled run), then checkout
+      # the head of the pull request instead of the merge commit.
+      - run: git checkout HEAD^2
+        if: ${{ github.event_name == 'pull_request' }}
+
+      - name: Run Snyk to check Docker images for vulnerabilities
+        # Snyk can be used to break the build when it detects vulnerabilities.
+        # In this case we don't want to halt development
```
austinbyers, line 46: I would say we want it to be a failing CI check, but we don't have to enforce it as a required check just yet

austinbyers (approved):
Nice!! I'm approving, but I'd wait for @3nvi to review before merging

.github/workflows/docker-analysis.yml
s0l0ist: I was just copying the style from our code scanning which is set like that.

.github/workflows/docker-analysis.yml
s0l0ist: Ah yes, I forgot about the 'required' check settings. Will fix.

.github/workflows/docker-analysis.yml
3nvi: We need both because:

1. We want them to run on every single PR we make
2. We want to catch any issues with potential package updates during new installations (for packages whose version hasn't been hardcoded all the way to the PATCH version), any admin merges that have bypassed a PR & finally any code that didn't wait for these sec checks to run before it got merge (since those sec checks are optional).

We can change that if you'd like

.github/workflows/docker-analysis.yml
```diff
@@ -0,0 +1,63 @@
...
+        with:
+          image: node:15.2.0-alpine
+          args: '--file=deployments/Dockerfile --severity-threshold=high'
+          # The default parameters do not work when testing locally with 'act' (brew install act). Therefore, we inline theme here which mimics
+          # how they would be interpreted if the above inputs worked.
+          # args: 'snyk test --file=deployments/Dockerfile --severity-threshold=high --docker node:15.2-alpine'
+          #
+          # To test locally, we need to use the image which replicates github actions. It takes a while to download the (6GB) image
+          # and to extract (18GB), be patient!
+          # > act -s SNYK_TOKEN=xxx pull_request -j docker_analysis -P ubuntu-latest=nektos/act-environments-ubuntu:18.04
```
3nvi, line 63: looks ok, but I'm mainly trusting you @s0l0ist on this 😄

deployments/Dockerfile
```diff
@@ -21,7 +21,7 @@
 #
 # ****************     BUILD STAGE     *******************
 #
-FROM node:13.2-alpine AS build-env
```
3nvi, line 24: do we need to update 2 major versions in this PR? I wouldn't want to do it without first testing it out locally & making sure that everything works ok

3nvi (commented):
Thanks for that @s0l0ist . Happy that Fossa is leaving since it was really slow. I'm a bit hesitant to give a green light, unless you can guarantee that upgrading node from 13 to 15 didn't create any issues. As an example, AWS images are still at v12.

Don't get me wrong, I'm **happy** to upgrade, just want to make sure we verified this upgrade 😃

deployments/Dockerfile
s0l0ist: yes - CI will immediately fail on `node:13.x`. I'll test

3nvi: snyk will fail in 1.13? That's harsh 😄

.github/workflows/docker-analysis.yml
austinbyers: Works for me

deployments/Dockerfile
s0l0ist: Dockerfile needed `python` dep to build `iltorb` from source as there's no pre-built bin for node v14.15.x

s0l0ist: Snyk reports a high vuln related to openssl in `node:13.x`. We don't use that dep, but it's also good practice to upgrade to the latest LTS node version.

PR #2259, opened 2020-12-11
Fixes for S3 Onboarding

alexmylonas:
## Background

We have a couple of fixes related to simplifying the S3 onboarding that can be implemented directly to master. 

Closes https://github.com/panther-labs/panther-enterprise/issues/1526, 
Closes https://github.com/panther-labs/panther-enterprise/issues/1596

## Changes

- Updated `LinkButton` to accept disable property
- Hide the `Launch Console` button when editing an S3 source
- Passing `error` prop on the buttons for StackDeployment to be disabled when getting the template fails

## Screenshot
Screenshot 2020-12-11 at 12 59 59 PM

## Testing

- Locally


giorgosp:
Some wording suggestions:
- Replace the large text `Download the autogenerated template ....` with:
```
Download the updated CloudFormation template and update the CloudFormation stack created in the bucket's region during the onboarding process for this source. The default stack name would be ${stackName}.
```
- Replace "Already have your IAM roles setup" with something like "After updating the stack, click below to proceed".

alexmylonas:
> Some wording suggestions:

I have updated the text, i think your suggestions are meant only for a user updates an S3 source.

3nvi (approved):
LGTM @alexmylonas

Issue #1981, opened 2020-11-06
Custom JIRA Tags - Frontend work


nhakmiller:
### Description

Allow users to specify a custom tag in the destination configuration page for JIRA destinations.

### Related Services

The destination onboarding wizard.

### Designs

Not sure if any design work needs to be done, it's just one new field. BUT it's already a fairly big form so maybe some re-design would be needed.

### Acceptance Criteria

You can specify a custom tag setting for JIRA destinations.

3nvi:
@nhakmiller wouldn't we need a backend change for this?

nhakmiller:
> @nhakmiller wouldn't we need a backend change for this?

Indeed we will. As you suggested on the epic https://github.com/panther-labs/panther/issues/1977, I created a frontend and backend task https://github.com/panther-labs/panther/issues/1980.

3nvi:
Oh I missed that! Sorry for that

mattthetraveler:
@3nvi do you think this will require new designs for the new field?

3nvi:
I think we can infer it, unless @dimitriszoitas feels we should do some sort of re-arrangement due to the excess amount of fields

dimitriszoitas:
I'll have to update to mockup either way, but I think that adding a large combo-box below our current fields will suffice. @3nvi let me know if you want to chat a couple of minutes on this.

nhakmiller:
@dimitriszoitas actually I think this design work could be merged with this design work too: https://github.com/panther-labs/panther/issues/2154, they are for the same form. Not sure if it is easier  for you to do both designs  at  once or  one then  the other (of if you're already done with this one), just wanted to throw it out there as an option.

3nvi:
Yeah I agree. We chatted and agreed to add a big single-line multi-combobox as the last field

PR #2268, opened 2020-12-11
Add Jira labels

### LINKED ISSUE #1980: JIRA Custom Tags Backend work
### Description

Allow users to specify a custom tags when forwarding to JIRA destinations.

### Related Services

The sources api and alert-forwarder will need to be updated.

### Designs

It should just be a pass through parameter, not too complicated.

### Acceptance Criteria

A JIRA destination can store a custom tag, and any alerts forwarded to that JIRA destination will be tagged with that custom tag or tags.

s0l0ist:
## Background

Closes #1980 , #1981 

To test the BE, it was much easier to finish the FE to test e2e. @dimitriszoitas please provide your feedback on the design if it needs to be updated. I preemptively added the field, but updating the form should be trivial.

Fixes a small bug where the `Click here to view in the Panther UI` was not rendering as a proper link. This is because we assume a markdown format, but Jira uses a different syntax to display hyperlinks.

Also fixes a bug in Asana where you couldn't update the `Project GIDs` field after the first time you saved the destination. This would affect any destination form field that was not a _string_ due to a BE config where it would try to merge configs using type `map[string]map[string]string` when it should have been `map[string]map[string]interface{}`.

## Changes

- Add `Labels` field to Jira (both BE and FE)
- Fix the `Click here to view in the Panther UI` link format
- Change `mergeConfig` to use a `map[string]map[string]interface{}` to allow saving other data types.

## Testing

- mage test:ci
- manually send test payload, visually inspect Jira

## Screenshots

Screen Shot 2020-12-11 at 10 34 58 PM
Screen Shot 2020-12-11 at 10 35 44 PM
Screen Shot 2020-12-11 at 10 40 54 PM


api/graphql/schema.graphql
```diff
@@ -645,6 +645,7 @@ type JiraConfig {
   apiKey: String!
   assigneeId: String
   issueType: String!
+  labels: [String!]
```
3nvi, line 648: optional: I think that this should be `labels: [String!]!` since the labels key should always be an array; even if it's empty.

api/lambda/outputs/models/api.go
```diff
@@ -219,12 +219,13 @@ type GithubConfig struct {
...
-	APIKey     string `json:"apiKey"`
-	AssigneeID string `json:"assigneeId"`
-	Type       string `json:"issueType"`
+	OrgDomain  string   `json:"orgDomain"`
+	ProjectKey string   `json:"projectKey"`
+	UserName   string   `json:"userName"`
+	APIKey     string   `json:"apiKey"`
+	AssigneeID string   `json:"assigneeId"`
+	Type       string   `json:"issueType"`
+	Labels     []string `json:"labels" validate:"omitempty,dive,min=1"`
```
3nvi, line 228: optional: related to my comment above, perhaps the `Labels` is mandatory (and at worst, it's an empty array)

web/src/components/forms/JiraDestinationForm/JiraDestinationForm.test.tsx
```diff
@@ -68,6 +71,7 @@ describe('JiraDestinationForm', () => {
     const apiKeyField = getByLabelText('* Jira API Key');
     const assigneeIdField = getByLabelText('Assignee ID');
     const issueTypeField = getByLabelText('* Issue Type');
+    const labelsField = getByLabelText('Labels', { selector: 'input' });
```
3nvi, line 74: Optional: In most tests, we write this as:

```suggestion
    const labelsField = getByAllLabelText('Labels')[0];
```

If this change is implemented, it should affect all similar statements in this page

web/src/components/forms/JiraDestinationForm/JiraDestinationForm.tsx
```diff
@@ -99,7 +101,8 @@ const JiraDestinationForm: React.FC<JiraDestinationFormProps> = ({ onSubmit, ini
           required={!existing}
           autoComplete="new-password"
         />
-
+      </SimpleGrid>
+      <SimpleGrid gap={5} columns={3}>
```
3nvi, line 105: I think @dimitriszoitas wanted to have labels on its own row, occupying 100% of the width. Am I right?

web/src/pages/ListDestinations/DestinationCards/JiraDestinationCard.tsx
```diff
@@ -39,6 +39,10 @@ const JiraDestinationCard: React.FC<JiraDestinationCardProps> = ({ destination }
       <GenericItemCard.Value label="Email" value={destination.outputConfig.jira.userName} />
       <GenericItemCard.Value label="Assignee ID" value={destination.outputConfig.jira.assigneeId} />
       <GenericItemCard.Value label="Issue Type" value={destination.outputConfig.jira.issueType} />
+      <GenericItemCard.Value
+        label="Labels"
+        value={destination.outputConfig.jira.labels.join(', ')}
```
3nvi, line 44: According to the current GraphQL schema, the value of `labels` can be `null`, so you should protect against it 😉

3nvi (commented):
great work @s0l0ist . We may need to fix a minor runtime bug, but other than that all's good.

I'd really love @dimitriszoitas opinion here before we merge

web/src/pages/ListDestinations/DestinationCards/JiraDestinationCard.tsx
s0l0ist: Yeah I haven't figured out a good way to sync the schema with what the lambda will return. It actually _never_ returns `null` or `undefined`, but you're right the schema doesn't make that clear.

api/lambda/outputs/models/api.go
austinbyers: Our `genericapi.Router` will automatically guarantee that maps and slices are set to `{}` and `[]` rather than nil, but you still probably don't want the `omitempty` tag

austinbyers (approved):
Nice! Great job 🚀

web/src/components/forms/JiraDestinationForm/JiraDestinationForm.tsx
s0l0ist: I spoke with @dimitriszoitas , the current design implementation I've put together is acceptable.

PR #2275, opened 2020-12-15
Refactor: rename FE component

3nvi:
## Background

We currently have a component named `BulletedLogType` in which we passed `logtypes` as a param. Currently we also pass it `resource types` since we recently added policy-based alerts.

Because of that, we'll be renaming `BulletedLogType` to `BulletedValue` in order to not tie this component to a log type context.

## Changes

- Rename `BulletedLogType` to `BulletedValue`

## Testing

- `npm run test`


web/src/components/BulletedValueList/BulletedLogTypeList.test.tsx
```diff
@@ -18,12 +18,12 @@
 
 import React from 'react';
```
alexmylonas, line 19: Rename test file name!

alexmylonas (approved):
LGTM! One small remnant file name change and its good to go!

PR #2277, opened 2020-12-15
fix: update both docker images

s0l0ist:
## Background

Version bump for the two docker images `panther-buildpack` and `panther-development-pack`

## Changes

- Update the versions to `1.15.0`

## Testing

- N/A


3nvi:
LGTM as long as you publish both the buildpack and development-pack images

austinbyers (approved):

PR #2281, opened 2020-12-15
Alert filtering by title

s0l0ist:
## Background

Fixes a bug where filtering alerts by title was not working at all.
 
## Changes

- fix conditional in the title filter

## Testing

- visually/manually


nhakmiller (approved):
Did you check the change that introduced this bug? I wonder if it was introduced elsewhere as well.

s0l0ist:
> Did you check the change that introduced this bug? I wonder if it was introduced elsewhere as well.

This change was introduced after v1.13, but must have missed testing.

PR #2290, opened 2020-12-16
Generate "DELETED" resource snapshot

### LINKED ISSUE #2289: Deleted Cloud Security resources are not marked as deleted in the data lake
### Describe the bug

Deleted Cloud Security resources are not marked as deleted in the data lake

### Steps to reproduce

1. Deploy Panther
1. Make sure Panther monitors your account
1. Create a new resource e.g. an S3 bucket
1. Delete the resource
1. Wait a bit. Run a query for the resource in the `cloud_security.snapshot_resourcehistory` table. You will not see the resource as deleted.

### Expected behavior

When a resource is deleted in a monitored account, there should be a "deleted" record in the datalake

### Environment

- Panther version or commit: https://github.com/panther-labs/panther/commit/65ed28300bacb469795b047fd9322424a819e4c2

kostaspap:
## Background

Fixes #2289 

## Changes

- Updated code so that it generates DELETED entries in the Cloud Security resource table. 

## Testing

- unit test
- Deployed in my env. Created a bucket, deleted it. Seen the following entry few minutes later: 
Screen Shot 2020-12-16 at 5 10 45 PM


rleighton (commented):
LGTM.

Deferring to NickH for correctness.

internal/compliance/datalake_forwarder/forwarder/resources.go
```diff
@@ -62,14 +62,38 @@ type resourceSnapshot struct {
...
+	}
+
 	// For INSERT and REMOVE events, we don't need to calculate a diff
 	switch lambdaevents.DynamoDBOperationType(record.EventName) {
 	case lambdaevents.DynamoDBOperationTypeInsert:
-		resource, err = sh.processResourceSnapshot(ChangeTypeCreate, record.Change.NewImage)
+		resource, err = sh.processResourceSnapshot(ChangeTypeCreate, newSnapshot)
 	case lambdaevents.DynamoDBOperationTypeRemove:
-		resource, err = sh.processResourceSnapshot(ChangeTypeDelete, record.Change.OldImage)
+		resource, err = sh.processResourceSnapshot(ChangeTypeDelete, oldSnapshot)
```
nhakmiller, line 84: I think we should add a check here for if `oldSnapshot.Deleted` is `True`. If `oldSnapshot.Deleted` is `True` and also the type is `lambdaevents.DynamoDBOperationTypeRemove`, I think we should just drop the record.

Otherwise I think as it is this will send two "resource is deleted" records to the data lake, one for when we see the resource is deleted and so we mark it deleted and one for when the resource TTL expires and we kick it out of the dynamo table.

internal/compliance/datalake_forwarder/forwarder/resources.go
```diff
@@ -205,3 +202,14 @@ func (sh *StreamHandler) processResourceSnapshot(changeType string,
 		ResourceAttributes: attributes,
 	}, nil
 }
+
+func marshalSnapshot(image map[string]*dynamodb.AttributeValue) (*resourceSnapshot, error) {
```
nhakmiller, line 206: Moving this out to its own function is a good idea 👍

nhakmiller (commented):
We now have two ways for a resource is deleted record to be sent to the datalake. One comment about how those ways may overlap. Overall though looks good.

internal/compliance/datalake_forwarder/forwarder/resources.go
kostaspap: Nice catch!

kostaspap: Updated.

nhakmiller (approved):
LGTM

PR #2293, opened 2020-12-16
fix: return alertId when generating the URL

s0l0ist:
## Background

Previously, alerts generated by policies were including a URL that was the sending the user to the `AnalysisID` or policy details page because there was no supporting _alert_ details page for these types of alerts.

This PR ensures all alerts will have the same URL to direct the user to the respective alert details pages.

## Changes

- `return alertURLPrefix + *alert.AlertID` for all alerts regardless of the alert type

## Testing

- Manually
- mage test:go

### DIFF
--- a/internal/core/alert_delivery/outputs/outputs_test.go
+++ b/internal/core/alert_delivery/outputs/outputs_test.go
@@ -29,7 +29,6 @@ import (
 )
 
 func init() {
-	policyURLPrefix = "https://panther.io/policies/"
 	alertURLPrefix = "https://panther.io/alerts/"
 }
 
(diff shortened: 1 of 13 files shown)


internal/core/alert_delivery/outputs/outputs_test.go
```diff
@@ -29,7 +29,6 @@ import (
 )
 
 func init() {
-	policyURLPrefix = "https://panther.io/policies/"
```
austinbyers, line 32: Fun fact: `panther.io` is for sale! Maybe we should buy it @jacknagz ? 🤔

austinbyers (approved):

internal/core/alert_delivery/outputs/outputs_test.go
s0l0ist: waaat yes!

jacknagz: lol of course it's for sale

jacknagz: @willow9886

PR #2297, opened 2020-12-17
Fix flaky test selector

vorillaz:
## Background

Our CI was failing as `findByText` selector with `await` was flaky.

## Changes

- Replace `findByText` with `getByText`
- Add `waitFor` instead of waiting a random amount of time.

## Testing

- `npm run test`


web/__tests__/setupTests.ts
```diff
@@ -31,6 +31,9 @@ import {
   ERROR_REPORTING_CONSENT_STORAGE_KEY,
 } from 'Source/constants';
 
+// 100 seconds
+jest.setTimeout(100000);
```
3nvi, line 35: @vorillaz that's way too much. A failing test will take 100 seconds before it is reported as "failing".

Remember this is the timeout for **each test** not for the aggregated test suites. Jest's default value for this `5000` so I think making it `8000` or `10000` would be ok

3nvi (approved):

PR #2300, opened 2020-12-17
Finally! a streaming CloudTrail parser

alxarch:
## Background

## Changes

## Testing


3nvi:
Yeah! Finally!!!

internal/log_analysis/log_processor/processor/logstream/jsonarray.go
```diff
@@ -0,0 +1,145 @@
...
+	} else if size < MinBufferSize {
+		size = MinBufferSize
+	}
+	return &JSONArrayStream{
+		iter: jsoniter.Parse(jsoniter.ConfigDefault, r, size),
+		seek: path,
+	}
+}
+
+type JSONArrayStream struct {
```
kostaspap, line 42: Can you add some documentation explaining where this is used?

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -125,37 +129,46 @@ func shouldIgnoreS3Object(s3Object *S3ObjectInfo) bool {
...
 
 func buildStream(ctx context.Context, s3Object *S3ObjectInfo) (*common.DataStream, error) {
-	s3Client, sourceInfo, err := getS3Client(s3Object.S3Bucket, s3Object.S3ObjectKey)
+	key, bucket := s3Object.S3ObjectKey, s3Object.S3Bucket
+	s3Client, src, err := getS3Client(bucket, key)
 	if err != nil {
-		err = errors.Wrapf(err, "failed to get S3 client for s3://%s/%s",
-			s3Object.S3Bucket, s3Object.S3ObjectKey)
+		err = errors.Wrapf(err, "failed to get S3 client for s3://%s/%s", bucket, key)
 		return nil, err
```
kostaspap, line 136: nit: just `return nil, errors.Wrap(....)`

internal/log_analysis/log_processor/processor/logstream/jsonarray.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"fmt"
+	"io"
+	"strconv"
+
+	jsoniter "github.com/json-iterator/go"
+	"github.com/pkg/errors"
+)
+
+func NewJSONArrayStream(r io.Reader, size int, path ...string) *JSONArrayStream {
```
kostaspap, line 30: Can you add documentation explaining what the parameters mean?

internal/log_analysis/log_processor/processor/logstream/jsonarray.go
```diff
@@ -0,0 +1,145 @@
...
+				return seekJSONPath(iter, path)
+			}
+			iter.Skip()
+		}
+		if iter.Error == nil {
+			iter.ReportError(opName, fmt.Sprintf("key %q not found", seek))
+		}
+		return false
+	case jsoniter.ArrayValue:
+		n, err := strconv.ParseInt(seek, 10, 64)
```
kostaspap, line 112: Can you add some comments in this method in general? Especially explaining when the code will execute each of the cases.

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -282,3 +295,10 @@ type SnsNotification struct {
 	events.SNSEntity
 	Token *string `json:"Token"`
 }
+
+// nolint:lll
+var rxCloudTrailLog = regexp.MustCompile(`^(?P<account>\d{12})_CloudTrail_(?P<region>[a-z]{2}-[a-z]+-\d+)_(?P<ts>\d{8}T\d{4}Z)_\w+.json.gz$`)
+
+func isCloudTrailLog(key string) bool {
```
kostaspap, line 302: can you add some tests for this method?

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -282,3 +295,10 @@ type SnsNotification struct {
 	events.SNSEntity
 	Token *string `json:"Token"`
 }
+
+// nolint:lll
+var rxCloudTrailLog = regexp.MustCompile(`^(?P<account>\d{12})_CloudTrail_(?P<region>[a-z]{2}-[a-z]+-\d+)_(?P<ts>\d{8}T\d{4}Z)_\w+.json.gz$`)
```
kostaspap, line 300: 1. Local zones are like `us-west-2-lax-1a`. Will this work?
2. Worth adding a link to the docs here: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-log-file-examples.html

internal/log_analysis/log_processor/parsers/awslogs/cloudtrail.go
```diff
@@ -118,42 +116,41 @@ var _ parsers.Interface = (*cloudTrailParser)(nil)
...
+	// This is just in case our special detections in the log processor don't work
+	hybrid := hybridCloudTrail{}
+	if err := jsoniter.UnmarshalFromString(log, &hybrid); err != nil {
+		return nil, errors.Wrap(err, "failed to read CloudTrail record JSON")
+	}
+	if hybrid.Records == nil {
+		// The log entry was a single event
+		return p.parseEvent(results, &hybrid.CloudTrail)
+	}
+	for i := range hybrid.Records {
```
kostaspap, line 129: I really don't like that we have to put this logic here, ideally we could switch the CloudTrail parser to the new parser logic and get rid of custom Go code. If you want to be extra cautious, i think it's fine to keep this for a bit, but maybe log an error if this ever happens - so at least we get notified to fix the regex. Personally I would be fine removing it all together.

alxarch: I think that we need the fallback for cases were log files are delivered in arbitrary paths that do not adhere to the 'official' pattern.

kostaspap: Ok, fine to be extra careful for now

internal/log_analysis/log_processor/processor/logstream/jsonarray.go
alxarch: Added

internal/log_analysis/log_processor/processor/logstream/jsonarray.go
alxarch: Added

internal/log_analysis/log_processor/sources/s3.go
alxarch: Added

internal/log_analysis/log_processor/sources/s3.go
alxarch: Added link and changed pattern for the local regions

internal/log_analysis/log_processor/sources/s3.go
alxarch: Do you mean to not display the s3 URL in the error?

internal/log_analysis/log_processor/processor/logstream/jsonarray.go
alxarch: Added

internal/log_analysis/log_processor/processor/logstream/jsonarray.go
kostaspap: Cool. If I understand the path elements needs to be like `"field name", "<some index>", "field name", "<index>", ...` correct? Might worth pointing it out.

kostaspap (approved):

PR #2303, opened 2020-12-17
fix: web test timeout

3nvi:
## Background

Jest now times out at over 1 minute, but our testing library still has a max-await of 1 second. This PR increases the timeout of our testing library to 10 seconds within CI environments

## Changes

- Increase timeout of testing library

## Testing

- Forced a fail and waited until the test timed out. The timeout occurred 1 min later


kostaspap (approved):
yeassssss

PR #2306, opened 2020-12-17
Rules Engine - Severity Refactor

wey-chiang:
## Background

Came to this realization when discussing with @lindsey-w in this [PR](https://github.com/panther-labs/panther-analysis-dogfood/pull/48#discussion_r544645024)

This discussion alerted me to a flaw in original logic for the `_get_severity` implementation -- namely that there was a illogical code duplication that caused this to not work as originally intended. 

These changes dedup some illogical code and make the generated severity field case-insensitive.

> Making it non-case sensitive should provide a better UX - it's more forgiving in a sense.

Dynamic Alert Severity is currently case sensitive meaning the following will raise an exception.

```
def severity(event):
    return "CrItIcAl"
```

These changes allow users to define severity such as above without issue.

## Changes

- Deduped & simplified valid severity logic -- removed illogical check
- Cleaned up unit test, added default severity to logger for clarity purposes
- Added test case for rule with valid severity: case insensitive 

## Testing

- `mage test:ci`
- Deploying into `us-east-2 to ensure nothing actually breaks


austinbyers (approved):
Nice job! Looks good to me, but I'm not an expert in the rules-engine

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -434,29 +427,21 @@ def _get_severity(self, event: Mapping, use_default_on_exception: bool = True) -
...
+                    'Expected severity to be any of the following: [%s], got [%s] instead.' % (str(SEVERITY_TYPES), severity)
+                )
         except Exception as err:  # pylint: disable=broad-except
             if use_default_on_exception:
                 self.logger.warning(
-                    'severity method for rule with id [%s] raised exception. Using default. Exception: %s', self.rule_id, err
+                    'severity method for rule with id [%s] raised exception. Using default (INFO). Exception: %s', self.rule_id, err
                 )
                 return 'INFO'
             raise
```
nhakmiller, line 444: Even if we're  setting the `severity` to `INFO`, shouldn't we still be reporting the exception? `WARN` messages do no get checked that frequently, so a failing severity function would likely go undetected for some time before a user noticed it manually. Or am I misunderstanding when `use_default_on_exception` is set?

wey-chiang: `use_default_on_exception` is a var passed in as `batch_mode` 
`batch_mode` is always true except when called by `analyze_single_rule` so from my understanding, this will only raise an exception when running a single rule

In terms of how we choose to raise exceptions, these changes do not change anything in that decision process.

nhakmiller: I see, so this is how we expect things to go. Seems reasonable enough.

nhakmiller (approved):

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -434,29 +427,21 @@ def _get_severity(self, event: Mapping, use_default_on_exception: bool = True) -
...
                     'severity method for rule with id [%s] yielded [%s], expected [%s]', self.rule_id, severity, str(SEVERITY_TYPES)
                 )
+                raise AssertionError(
+                    'Expected severity to be any of the following: [%s], got [%s] instead.' % (str(SEVERITY_TYPES), severity)
+                )
         except Exception as err:  # pylint: disable=broad-except
             if use_default_on_exception:
                 self.logger.warning(
-                    'severity method for rule with id [%s] raised exception. Using default. Exception: %s', self.rule_id, err
+                    'severity method for rule with id [%s] raised exception. Using default (INFO). Exception: %s', self.rule_id, err
```
kostaspap, line 441: Shouldn't the default severity be the severity of the rule? Why INFO?

PR #2307, opened 2020-12-18
Parallelize alert s3select

rleighton:
## Background

Modifies the Alerts API to run S3 Select in parallel when retrieving objects from S3

## Changes
- Running up to 50 S3Selects in parallel while searching for events
- Minor fixes in tests

## Testing
- mage test:ci
- E2E

### DIFF
--- a/internal/log_analysis/alerts_api/api/get_alert.go
+++ b/internal/log_analysis/alerts_api/api/get_alert.go
@@ -19,12 +19,9 @@ package api
  */
 
 import (
-	"bytes"
+	"context"
 	"fmt"
-	"sort"
 	"strings"
-	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -37,14 +34,10 @@ import (
 	"github.com/panther-labs/panther/internal/log_analysis/alerts_api/table"
 	"github.com/panther-labs/panther/internal/log_analysis/alerts_api/utils"
 	"github.com/panther-labs/panther/internal/log_analysis/awsglue"
-	"github.com/panther-labs/panther/internal/log_analysis/log_processor/destinations"
 	"github.com/panther-labs/panther/internal/log_analysis/pantherdb"
 )
 
 const (
-	s3SelectQueryBuffer = 1000 // how many listed events we allow in flight
-	s3SelectConcurrency = 50   // not too big or you will get throttled!
-
 	// The format of S3 object suffix that contains the
 	ruleSuffixFormat = "rule_id=%s/"
 
@@ -52,7 +45,7 @@ const (
 )
 
 // GetAlert retrieves details for a given alert
-func (api *API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOutput, err error) {
+func (api *API) GetAlert(input *models.GetAlertInput) (*models.GetAlertOutput, error) {
 	alertItem, err := api.alertsDB.GetAlert(input.AlertID)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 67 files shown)


internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -65,12 +70,20 @@ func (api *API) GetAlert(input *models.GetAlertInput) (result *models.GetAlertOu
...
 
 	var events []string
 	for _, logType := range alertItem.LogTypes {
 		// Each alert can contain events from multiple log types.
 		// Retrieve results from each log type.
+		if logTypeToken, found := token.LogTypeToToken[logType]; found {
+			if logTypeToken.EventIndex == -1 { // if -1 then already searched this logType completely
+				continue
+			}
+		}
```
kostaspap, line 86: Actually, it is possible that more events were added to a logtype we searched before. So we cannot have such a condition....
Having said that, I think we were not handling this scenario properly before neither. We need to update the token to keep track of lastS3event/logtype...Updating it now

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -174,49 +204,146 @@ func (api *API) getEventsForLogType(
...
+		search.s3SelectQueryWaitGroup.Add(1)
+		go func() {
+			for s3SelectQuery := range search.s3SelectQueryChan {
+				s3SelectResult := api.queryS3Object(s3SelectQuery)
+				search.s3SelectResultChan <- &s3SelectResult
+				// protect tally with atomic
+				atomic.AddUint32(&search.s3SelectTotalEventsFound, uint32(len(s3SelectResult.events)))
+				if int(search.s3SelectTotalEventsFound) >= maxResults || s3SelectResult.err != nil {
+					search.s3SelectListChan <- struct{}{} // signal listing to stop
+					break
```
kostaspap, line 278: I'll have to update that - we can receive events out of order e.g. querying object1, object2 object2 may reply first.

rleighton (commented):
Much prettier! I'll pull it and see how it works.

internal/log_analysis/alerts_api/api/s3_search.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+			return nil
+		})
+	}
+
+	// drive requests thru the worker group
+	for _, object := range objects {
+		objectTime, err := timeFromJSONS3ObjectKey(*object.Key)
+		if err != nil {
+			return nil, err
```
kostaspap, line 127: We need to make sure we close the queryChan

rleighton: Line 145 below

kostaspap: There is a return statement in line 127, we will never reach that line.

internal/log_analysis/alerts_api/api/s3_search.go
```diff
@@ -0,0 +1,170 @@
...
+			alertID:    s.alert.AlertID,
+			maxResults: s.maxResults,
+		}
+		queryChan <- s3SelectQuery
+	}
+
+	// this will signal workers to stop
+	close(queryChan)
+	if err := workerGroup.Wait(); err != nil {
+		return nil, err
```
kostaspap, line 147: In case we return here, we will not close the `resultChan`

internal/log_analysis/alerts_api/api/get_alert.go
```diff
@@ -118,44 +122,62 @@ func (api *API) getEventsForLogType(
...
+			return outEvents, &outToken, nil
 		}
 	}
 
-	for ; !nextTime.After(alert.UpdateTime); nextTime = awsglue.GlueTableHourly.Next(nextTime) {
-		if len(result) >= maxResults {
-			// We don't need to return any results since we have already found the max requested
-			break
+	var partitionTime time.Time
+	if token != nil {
```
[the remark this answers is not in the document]
kostaspap: Yeah, it could. I thought maybe the code was more clear this way, but can move it up if you think it's better

kostaspap (approved):

PR #2308, opened 2020-12-18
fix: correct deep merging on alert's events

3nvi:
## Background

When moving to policy-based alerts, the GraphQL structure changed and we introduced more "parent" objects to the alert's `events` key. Because of the fact that our GraphQL library isn't 100% strongly typed, we missed a modification that was affecting endless pagination on alert's events.

This PR fixes that

## Changes

- Fix issue where the paginated events were not being properly merged with the existing ones
- Enforce strong types on pagination params

## Testing

- `npm run test`
- Manually


kostaspap (approved):

kostaspap:
Thanks!

alexmylonas (approved):
LGTM! Optionally add a test case for it!

3nvi:
> LGTM! Optionally add a test case for it!

Testing endless pagination is hard atm. We can't mimic a "scroll behavior" in our tests :/

PR #2309, opened 2020-12-18
Use string slice for PantherAnyString

alxarch:
## Background

We ran into trouble with `[]` in `Umbrella.DNS` logs. The logic we used for storing indicator fields in the parsers using the
old method was complex, had allocation overhead and was error-prone when empty values were involved. 

## Changes

- Switches `parsers.PantherAnyString` to be just `[]string`.

## Testing

- mage test:go
- E2E underway

### DIFF
--- a/internal/log_analysis/log_processor/parsers/pantherlog.go
+++ b/internal/log_analysis/log_processor/parsers/pantherlog.go
@@ -20,18 +20,17 @@ package parsers
 
 import (
 	"net"
-	"reflect"
 	"regexp"
 	"sort"
 	"time"
 
 	jsoniter "github.com/json-iterator/go"
 
-	"github.com/panther-labs/panther/internal/log_analysis/awsglue/glueschema"
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog"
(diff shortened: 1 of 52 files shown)


internal/log_analysis/log_processor/parsers/pantherlog.go
```diff
@@ -60,53 +59,18 @@ type PantherLog struct {
...
-	if err != nil {
-		return err
-	}
-	any.set = make(map[string]struct{}, len(values))
-	for _, entry := range values {
-		any.set[entry] = struct{}{}
-	}
-	return nil
+func (any PantherAnyString) MarshalJSON() ([]byte, error) {
+	sort.Strings(any)
```
kostaspap, line 72: we want avoid duplicates. You use the stringset method here to remove them

alxarch: I thought of doing that. It will be redundant though. We have already added all strings using the stringset method.

alxarch: Changed it to be safe rather than performant.

kostaspap (approved):
Great work!

PR #2319, opened 2020-12-21
Increase updater batching

kostaspap:
## Background

Increase Updater Lambda SQS batch size

## Changes

- Increase batch size for Lambda

## Testing

- Deployed fresh Panther instance. Seen data in Athena

### DIFF
--- a/deployments/log_analysis.yml
+++ b/deployments/log_analysis.yml
@@ -683,7 +683,8 @@ Resources:
           Type: SQS
           Properties:
             Queue: !GetAtt UpdaterQueue.Arn
-            BatchSize: 10
+            BatchSize: 10000 # Max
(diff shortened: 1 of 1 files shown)


deployments/log_analysis.yml
```diff
@@ -683,7 +683,8 @@ Resources:
           Type: SQS
           Properties:
             Queue: !GetAtt UpdaterQueue.Arn
-            BatchSize: 10
+            BatchSize: 10000 # Max
+            MaximumBatchingWindowInSeconds: 15
```
rleighton, line 687: I did some tests, can we make this 30?

kostaspap: Updated

rleighton (approved):

PR #2325, opened 2020-12-23
Add extraction to resource in CloudSecurity

rleighton:
## Background

Resources often have embedded indicators. This PR adds extraction to walk the resource and extract the `p_any` fields.

## Changes

- Add extractor by reviewing the resource models
- Refactor extractor for CloudTrail

## Testing

- mage test:ci
- mage deploy

### DIFF
--- a/internal/compliance/snapshotlogs/resource.go
+++ b/internal/compliance/snapshotlogs/resource.go
@@ -37,6 +37,11 @@ var logTypeResourceHistory = logtypes.MustBuild(logtypes.ConfigJSON{
 	NewEvent: func() interface{} {
 		return &Resource{}
 	},
+	ExtraIndicators: pantherlog.FieldSet{ // these are added by extractors but not used in struct directly
+		pantherlog.FieldDomainName,
+		pantherlog.FieldIPAddress,
+		pantherlog.FieldAWSTag,
+	},
 	Validate: pantherlog.ValidateStruct,
 })
 
@@ -57,11 +62,6 @@ type Resource struct {
 	ARN              pantherlog.String      `json:"arn" panther:"aws_arn" description:"The Amazon Resource Name (ARN) of the resource."`
 	Name             pantherlog.String      `json:"name" description:"The AWS resource name of the resource."`
 	Tags             map[string]string      `json:"tags" description:"A standardized format for key/value resource tags."`
-
-	// dummy fields that are not exported but used to ensure that `p_any` columns are created in the schema for extracted fields
-	DummyDomain pantherlog.String `json:"-" panther:"domain"`
-	DummyIP     pantherlog.String `json:"-" panther:"ip"`
-	DummyTags   pantherlog.String `json:"-" panther:"aws_tag"`
 }
 
 // WriteValuesTo implements pantherlog.ValueWriterTo interface
--- a/internal/compliance/snapshotlogs/resource.go
+++ b/internal/compliance/snapshotlogs/resource.go
@@ -58,8 +58,7 @@ type Resource struct {
 	Name             pantherlog.String      `json:"name" description:"The AWS resource name of the resource."`
 	Tags             map[string]string      `json:"tags" description:"A standardized format for key/value resource tags."`
 
-	// a dummy field that is not exported but used to ensure that `p_any` columns are created in the schema for extracted fields
-	// DummyAny pantherlog.String `json:"-" panther:"aws_account_id,aws_arn,aws_instance_id,domain,ip"`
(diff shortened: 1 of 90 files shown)


internal/log_analysis/log_processor/parsers/awslogs/indicators.go
```diff
@@ -38,22 +38,7 @@ func mustBuildEventSchema(schema interface{}) interface{} {
 }
 
 func ExtractRawMessageIndicators(w pantherlog.ValueWriter, messages ...pantherlog.RawMessage) {
```
kostaspap, line 40: You can just delete this function - any callers of this function should now just call `pantherlog.ExtractRawMessageIndicators`

internal/compliance/snapshotlogs/resource.go
```diff
@@ -52,11 +57,60 @@ type Resource struct {
...
+		case "ID":
+			pantherlog.ScanAWSInstanceID(w, value)
+		case "AccountId", "OwnerId":
+			pantherlog.ScanAWSAccountID(w, value)
+		case "Address", "AssignPublicIp", "PrivateIpAddress", "PrivateIPAddress", "PublicIpAddress", "PublicIPAddress":
+			pantherlog.ScanIPAddress(w, value)
+		case "Domain", "DomainName", "DNSName", "FQDN", "PrivateDnsName", "PublicDnsName":
+			if value != "" {
+				w.WriteValues(pantherlog.FieldDomainName, value)
+			}
```
kostaspap, line 106: You can use the pantherlog method for that: 
```
case "Domain", "DomainName", "DNSName", "FQDN", "PrivateDnsName", "PublicDnsName": 
pantherlog.ScanHostname(w, value)

```

internal/compliance/snapshotlogs/resource.go
```diff
@@ -52,11 +57,60 @@ type Resource struct {
 	ARN              pantherlog.String      `json:"arn" panther:"aws_arn" description:"The Amazon Resource Name (ARN) of the resource."`
 	Name             pantherlog.String      `json:"name" description:"The AWS resource name of the resource."`
 	Tags             map[string]string      `json:"tags" description:"A standardized format for key/value resource tags."`
+
+	// a dummy fields that are not exported but used to ensure that `p_any` columns are created in the schema for extracted fields
+	DummyDomain pantherlog.String `json:"-" panther:"domain"`
```
kostaspap, line 62: So these are needed? This means we have another bug, we are missing the `p_any_aws_tags` column. 
Can you add another field: 
```
DummyAwsTag pantherrlog.String `json:"-" panther:"aws_tag"`
```

kostaspap (commented):
Can you add a new test or extend existing parser tests in `snapshotlogs_test.go` to verify the extraction works as intended?

internal/compliance/snapshotlogs/resource.go
rleighton: Not a bug, by design but I'll defer to @alxarch for the "right" way to declare `p_any` columns when they are NOT referenced in the parser structs. . For `ConfigJSON` type parsers, right now we need add dummy struct members. We _could_ change that and allow a `Schema` to be passed like we do the `Config` type parsers, see: https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/logtypes/entry.go#L116

rleighton: I'll add a dummy for the tags.

rleighton:
> Nice work! Can you add a new test or extend existing parser tests in `snapshotlogs_test.go` to verify the extraction works as intended?

Good catch, I intended to but forgot (fighting deployment bugs fixed in the other PR).

internal/compliance/snapshotlogs/resource.go
rleighton: Done

internal/log_analysis/log_processor/parsers/awslogs/indicators.go
rleighton: Done

rleighton:
> > Nice work! Can you add a new test or extend existing parser tests in `snapshotlogs_test.go` to verify the extraction works as intended?
> 
> Good catch, I intended to but forgot (fighting deployment bugs fixed in the other PR).

Done

kostaspap (approved):

internal/compliance/snapshotlogs/resource.go
rleighton: Done

PR #2326, opened 2020-12-23
Update gluesync and datacatalog updater for cloudsec

rleighton:
## Before

Gluesync and the datacatalog updater did not have config for cloudsec.

Also, the oplog was missing from datacatalog-updater, this broke the dash board and removed error logging.

## Changes

- Add cloudsec tables where necessary
- Add back oplog to datacatalog updater (needed for error logging and dashboard metrics)
- Allow cloudsec tables in views so they can be searched via Indicator Search

## Testing

- mage test:ci
- mage deploy


internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
```diff
@@ -67,22 +64,10 @@ func (h *LambdaHandler) createOrReplaceViewsForAllDeployedLogTables(ctx context.
...
-			// Don't create views for the Cloud Security tables
-			continue
-		}
-		tablesInView = append(tablesInView, table)
-	}
+	deployedLogTables := resolveTables(ctx, h.Resolver, deployedLogTypes...)
 	// update the views for *all* tables based on the log tables.
 	// FIXME: this is confusing, the athenaviews package should not be creating views by expanding table metadata based on hard-wired logic
-	if err := athenaviews.CreateOrReplaceLogViews(h.AthenaClient, h.AthenaWorkgroup, tablesInView); err != nil {
+	if err := athenaviews.CreateOrReplaceLogViews(h.AthenaClient, h.AthenaWorkgroup, deployedLogTables); err != nil {
```
kostaspap, line 70: The views have a 1:1 relationship with databases: 
`panther_logs` -> `panther_views.all_logs`
`panther_rule_matches` -> `panther_views.all_rule_matches`
`panther_rule_errors` -> `panther_views.all_rule_errors`

Since the CloudSec datasets are not "logs", we decided we shouldn't put them in the `panther_logs` database, but in their own database. Following them same concept, we should put them in a different view and not inside the `all_logs` view. 
If we want, we could make eventually a view across *everything* (logs, rule errors, rule matches, cloudsec)

rleighton: Agree. We should be consistent. I'll make a view for the cloudsec database. I'll also add a view over all databases, which is easy and just handy.

rleighton: Done

rleighton:
I changed the configuration to allow rules to be written against this data. This is not just cleaner because cloud security is no longer  an exception, it allows rules to be written against specific changes. There are also other benefits I'll write about on Slack.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
```diff
@@ -97,17 +82,15 @@ func (h *LambdaHandler) fetchAllDeployedLogTypes(ctx context.Context) ([]string,
...
 	var out []*awsglue.GlueTableMetadata
 	for _, name := range names {
 		entry, err := r.Resolve(ctx, name)
-		if err != nil {
-			return nil, err
-		}
-		if entry == nil {
-			return nil, errors.Errorf("unresolved log type %q", name)
+		if err != nil || entry == nil { // don't fail whole operation if inconsistent data...
+			zap.L().Error("unresolved log type", zap.String("name", name), zap.Error(err))
```
alxarch, line 92: Logging an error here is *not* the correct way to handle an unresolved log type here.
The function should return the error to the caller and it should be handled there.
This introduces a global dependency and silently skips the error.
`zap.L()` usage is almost always the wrong way to handle an error.
Logging of errors should only happen 'close' to the top level of handler code.

alxarch (changes requested):
Overall the changes seem fine. Only exception is the change of signature to the `resolveTables` function.

internal/log_analysis/datacatalog_updater/datacatalog/handler.go
```diff
@@ -75,9 +77,17 @@ func (h *LambdaHandler) Invoke(ctx context.Context, payload []byte) ([]byte, err
 	return nil, nil
 }
 
+var opLogManager = oplog.NewManager("log_analysis", "datacatalog_updater")
+
 // HandleSQSEvent handles messages in an SQS event.
-func (h *LambdaHandler) HandleSQSEvent(ctx context.Context, event *events.SQSEvent) error {
-	tasks, err := tasksFromSQSMessages(event.Records...)
+func (h *LambdaHandler) HandleSQSEvent(ctx context.Context, event *events.SQSEvent) (err error) {
+	// NOTE: this logging is needed for alarming and dashboards!
```
alxarch, line 84: The `oplog` manager has a horrible API that heavily relies on global variables and scatters it's logic all around the code base instead of wrapping a handler and using `defer`.

alxarch: A top of my mind approach to this would be to have a `OrchestratedHandler` that can wrap types implementing `lambda.Handler`. 
```go
func OrchestrateHandler(h lambda.Handler) lambda.Handler
```
Any labels we want to put in the logger should be part of the context logger, not relevant to the orchestration.
The resulting handler acts like a *middleware* that uses `defer` to log time/mem usage, errors etc.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
rleighton: So, the issue is I was able to create some inconsistent data (messing in my account) and this caused everything to fail because this function failed ... debugging this was complicated by the fact that the lambda lost all logging of errors , so there were no alarms or logs.

@alxarch If we do not log here AND we do not want to fail, how do you suggest passing this info back to the caller? We do want to know about this situation but we do not want the whole system to fail. I would say an error is not the right way, since the function did not fail.

internal/log_analysis/datacatalog_updater/datacatalog/handler.go
rleighton: Changing `oplog` is a task onto itself, happy to have you do it in anther PR. In the end, we want a standard way of logging at the top of the stack , including stats, etc. This was removed here and ALL logging stopped and a dashboard broke. We may have had production issues as a result but we will not know because of no logs.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
rleighton: @alxarch I made the function return a slice of warning errors and all callers up to the handler, where it is logged. Seems like over engineering but functionally the same so we have these properties:

 - one bad logType does not cause the whole system to collapse
 - we get paged when there is a bad logType

alxarch: I think having an unresolved log type when creating views/tables *is* an irrecoverable error. The system would not collapse. The views just won't get updated.

alxarch: If we organize the management of log types a bit better by not relying on Glue catalog state for our system state we could avoid even having to resolve the tables multiple times. If you consider when the `resolveTables` function is used we should not have any unresolved log types at that point. It would mean something is really wrong. I would still suggest returning an error (even if using `multierror.Append` to collect all such errors) and fail the operation immediately.

internal/core/logtypesapi/main/lambda.go
```diff
@@ -50,18 +51,19 @@ func main() {
 	// Syncing the zap.Logger always results in Lambda errors. Commented code kept as a reminder.
 	// defer logger.Sync()
 
-	session := session.Must(session.NewSession())
+	awsSession := session.Must(session.NewSession())
 	nativeLogTypes := logtypes.CollectNames(registry.NativeLogTypes())
+	nativeLogTypes = append(nativeLogTypes, logtypes.CollectNames(snapshotlogs.LogTypes())...)
```
kostaspap, line 56: We went to great lengths and put lots of effort to make sure we didn't have this behavior.  We didn't want these log types to be available when someone chooses to onboard a source from S3/SQS, same as we didn't want them to be available in Rules. Users have Policies for these resources.

internal/log_analysis/athenaviews/athenaviews_test.go
```diff
@@ -45,12 +45,27 @@ func TestGenerateViewAllLogs(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, expectedSQL, sql)
+}
+
+func TestGenerateViewAllCloudSecurity(t *testing.T) {
+	table1 := awsglue.NewGlueTableMetadata(pantherdb.CloudSecurityDatabase, "table1", "test table1", awsglue.GlueTableHourly, &table1Event{})
+	table2 := awsglue.NewGlueTableMetadata(pantherdb.CloudSecurityDatabase, "table2", "test table2", awsglue.GlueTableHourly, &table2Event{})
+	// nolint (lll)
+	expectedSQL := `create or replace view panther_views.all_cloudsecurity as
+select 'panther_cloudsecurity' AS databaseName,day,hour,month,NULL AS p_any_aws_account_ids,NULL AS p_any_aws_arns,NULL AS p_any_aws_instance_ids,NULL AS p_any_aws_tags,p_any_domain_names,p_any_ip_addresses,p_any_md5_hashes,p_any_sha1_hashes,p_any_sha256_hashes,p_event_time,p_log_type,p_parse_time,p_row_id,p_source_id,p_source_label,year from panther_cloudsecurity.table1
```
kostaspap, line 63: Maybe `p_db_name` given that all panther generated fields use this naming format?

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
rleighton: The `multierr.Append` worked but was messy. I found the clean solution, I move the logging into the resolver that is created in the `main` of the lambda.

internal/log_analysis/athenaviews/athenaviews_test.go
rleighton: Excellent idea. Done.

internal/core/logtypesapi/main/lambda.go
alxarch: I assumed we only needed the snapshot resolver to be chained. Where do we require these logs to be part of the ListAvailableLogTypes endpoint?

alxarch: After talking on a call with Russ we reached the conclusion that it might be best to include it here for now so that the cloud security feature can be implemented. We will probably split the endpoints for sources and rules when we also manage to separate parsers from schemas. Rules need available schemas (including cloudsecurity) and sources need available parsers (a parser can produce results of multiple types). This will be a section in the centralized schema management RFC.

rleighton: Resolving. Can you pls approve? I have an enterprise PR that depends on this one.

alxarch (approved):
If Kostas agrees we should just include the cloudsecurity log types in the list for now so the feature can be unblocked. Otherwise we need to wait for 'harder' PRs like splitting parsers/schemas or find some other weird throwaway solution to only show them in the rules dropdon.

internal/log_analysis/athenaviews/athenaviews.go
```diff
@@ -57,52 +57,99 @@ func GenerateLogViews(tables []*awsglue.GlueTableMetadata) (sqlStatements []stri
...
+		}
+	}
+	sql, err = generateViewAllHelper("all_cloudsecurity", cloudsecTables, []awsglue.Column{})
+	return sql, cloudsecTables, err
 }
 
 // generateViewAllRuleMatches creates a view over all log sources in rule match db the using "panther" fields
-func generateViewAllRuleMatches(tables []*awsglue.GlueTableMetadata) (sql string, err error) {
+func generateViewAllRuleMatches(tables []*awsglue.GlueTableMetadata) (sql string, ruleTables []*awsglue.GlueTableMetadata, err error) {
 	// the rule match tables share the same structure as the logs with some extra columns
```
kostaspap, line 128: Should you add a. 
```
if table.DatabaseName() == pantherdb. LogProcessingDatabase
```

check here? Similar as you do in the methods above.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
```diff
@@ -66,23 +64,15 @@ func (h *LambdaHandler) createOrReplaceViewsForAllDeployedLogTables(ctx context.
 	if err != nil {
 		return errors.Wrap(err, "failed to fetch deployed log types")
 	}
-	// We map the deployed log types to their 'base' log tables.
+	// We map the deployed log types to their 'base' log tables, errors are collected and not fatal
 	deployedLogTables, err := resolveTables(ctx, h.Resolver, deployedLogTypes...)
```
kostaspap, line 68: nit: This is no longer just log tables right? It also includes cloud security? Maybe rename to `deployedTables` for clarity?

internal/log_analysis/datacatalog_updater/main/lambda.go
```diff
@@ -47,6 +47,19 @@ const (
...
 
+type logTypeResolver struct {
+	resolver logtypes.Resolver
+}
+
+func (r *logTypeResolver) Resolve(ctx context.Context, name string) (logtypes.Entry, error) {
+	entry, err := r.resolver.Resolve(ctx, name)
+	if err == nil && entry == nil {
+		// if a logType is not found, this indicates bad data ... log/alarm
+		zap.L().Error("cannot resolve logType", zap.String("logType", name))
```
kostaspap, line 58: Do we need this custom struct here? Why not just log the error in line 100 of internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go ?

kostaspap (approved):

internal/log_analysis/datacatalog_updater/main/lambda.go
rleighton: LOL! That's how I originally . Alex _hates_ logging deep in the stack and insisted I pull it up to the top.

internal/log_analysis/datacatalog_updater/datacatalog/create_tables.go
rleighton: Good idea.

internal/log_analysis/datacatalog_updater/main/lambda.go
kostaspap: In general I agree with the idea that we shouldn't log deep in the call stack but this seems lie over-engineering to me.

rleighton: There a was a long thread on that (now resolved). If a bad log type is ever put into a source then all log types from the source cannot be onboarded because we fail. Very fragile. Breaks onboarding. It happened to me doing stuff in dev but it could just as easily happen in production from a bug. If it did the customer would be stopped. This way we get alerted and the valid log types all work.

PR #2327, opened 2020-12-23
fix: remove newlines from summary in Jira

s0l0ist:
## Background

Fixes an issue where newlines inside an alert would fail to be delivered to Jira destinations.

## Changes

- regex to remove newline characters for the `summary` field.

## Testing


internal/core/alert_delivery/outputs/utils.go
```diff
@@ -109,3 +111,8 @@ func GetOpsGenieRegionalEndpoint(serviceRegion string) string {
 		return "https://api.opsgenie.com/v2/alerts"
 	}
 }
+
+func removeNewLines(input string) string {
+	re := regexp.MustCompile(`\r?\n`)
```
kostaspap, line 116: 1. Huh, interesting, why the `\r`? Couldn't this be just `return strings.ReplaceAll(input, "\n", "")`
2. If we need to keep the `\r` then probably need to change the name - the function doesn't just remove new lines but also carriage return chars. An idea would be to name it sth else e.g. `removeInvalidChars` or `sanitize` or sth and put it inside the `jira.go`

kostaspap (approved):
Seems easy win to add a unit test for the new function.

internal/core/alert_delivery/outputs/utils.go
s0l0ist: I initially thought we should remove `\r` because it could cause problems, but looking back, the error was strictly about new lines. I'll remove the `\r?`

PR #2328, opened 2020-12-24
Using snapshot poller session as the base session for all Snapshot poller clients

kostaspap:
## Background

I have seen a couple of errors reported by the snapshot poller. I saw that we were using the default number of retries in the clients. 

## Testing

- mage test:ci
- Deployed change and did the following: 
   - Updated resource in another region, seen event come through. 
   - Onboard different AWS Account


internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -242,21 +242,18 @@ func getClient(pollerInput *awsmodels.ResourcePollerInput,
...
+	creds := AssumeRoleFunc(pollerInput, SnapshotPollerSession)
 
 	// Second, we need to create a new session in the account to be scanned using the credentials
 	// we just created. This works around a situation where the account being scanned has an opt-in
 	// region enabled that the Panther account does not.
 	//
 	// The region does not matter here, since we are just creating the session. When we create the
 	// client, we will need to specify the region.
-	clientSession, err := session.NewSession(&aws.Config{Credentials: creds})
-	if err != nil {
```
kostaspap, line 254: Using the snapshot poller session as base - and overriding the creds. I assume that was the intention initially anyway?

nhakmiller (commented):
Seems fine, I just want to double check before approving though you did test this with cross account & cross region scanning right?

kostaspap:
> Seems fine, I just want to double check before approving though you did test this with cross account & cross region scanning right?

Good call. I tested cross-region but not cross-account. I will check it now.

kostaspap:
> Seems fine, I just want to double check before approving though you did test this with cross account & cross region scanning right?

Done!

nhakmiller (approved):
🚀

PR #2330, opened 2020-12-29
Add hash scanners to pantherlog

rleighton:
## Background

The hash `p_any` fields had no validation. Added a scanner for each.

## Changes

- Add scanner functions for md5, sha1 and sha256
- Minor fixes to `ScanARN` and `ScanHostname`

## Testing

- mage test:ci


internal/log_analysis/log_processor/pantherlog/scanners.go
```diff
@@ -190,3 +190,37 @@ func ScanEmail(w ValueWriter, input string) {
...
+	if len(input) == 64 && isHex(input) {
+		w.WriteValues(FieldSHA256Hash, input)
+	}
+}
+
+func isHex(s string) bool {
+	if len(s) == 0 {
+		return false
+	}
+	for _, c := range s {
```
alxarch, line 219: A `range` over `s` [will go through UTF8 decoding and iterate over `rune` values](https://golangbot.com/strings/). It's best to use a normal `i` loop to iterate over bytes in this case since hex strings are ASCII. 
```suggestion
	for i := 0; 0 <= i && i < len(s); i++ {
	  if !isHexDigit(s[i]) {
	    return false
	  }
	}
	return true
  }
// this function gets inlined by the compiler
// we use  a function to make the check in the loop above more readable
func isHexDigit(c byte) bool {
  return ('A' <= c && c <= 'F') || ('a' <= c && c <= 'f') || ('0' <= c && c <= '9')
}
```

internal/log_analysis/log_processor/pantherlog/scanners_aws.go
```diff
@@ -40,7 +40,9 @@ func ScanARN(w ValueWriter, input string) {
 		return
 	}
 	w.WriteValues(FieldAWSARN, input)
-	w.WriteValues(FieldAWSAccountID, arn.AccountID)
+	if arn.AccountID != "" {
```
alxarch, line 43: IIRC `WriteValues` does not write empty strings. This is why the check was not there.

alxarch (approved):
LGTM, only a minor (sometimes important) detail about use of `range` over `string` values.

alxarch:
One thing we might consider is if we want to normalize our hex strings to a specific case. It might be important for using them in queries. Does Athena/Snowflake use a specific collation for queries that is case sensitive?

internal/log_analysis/log_processor/pantherlog/scanners_aws.go
rleighton: Yeah, but I'm using this elsewhere and getting `""`.  Given these are "scanners" , I think they should not emit thing that don't match the scan.

internal/log_analysis/log_processor/pantherlog/scanners.go
rleighton: Good point.

rleighton:
> One thing we might consider is if we want to normalize our hex strings to a specific case. It might be important for using them in queries. Does Athena/Snowflake use a specific collation for queries that is case sensitive?

Good point,  I'll add a ToLower().

internal/log_analysis/log_processor/pantherlog/scanners.go
rleighton: Done

PR #2331, opened 2020-12-30
Allow multi-step deployment

TDaglis:
## Background

When deploying the master stack while using the `CustomDomain` parameter, you have a kind of race condition to set up the custom domain to actually point to the load balancer deployed by the stack, before the stack deployment finishes.

## Changes

This PR allows you to break the deployment of the stack into two steps. The first step will deploy just the Bootstrap nested stack and the second step will complete the deployment.
After the first deployment step finishes, you are able to get the ELB URL and set up your DNS records before proceeding to the next and final deployment step.

- ~`FullDeployment`~`BootstrapStackOnly` parameter was added to enable the 2 step deployment
- The default behavior is backward compatible and results in the full deployment of the master stack
- `FullDeployment` condition is used on every resource other than the Bootstrap nested stack
- For the `Onboard` nested stack which already had a condition for its creation, a new compound condition was created to include the `FullDeployment` condition

## Testing

- Successfully deployed the stack using the 2-step process
- Successfully deployed the stack using the default FullDeployment single-step process


s0l0ist (approved):

deployments/master.yml
```diff
@@ -170,6 +170,13 @@ Parameters:
     Description: An existing VPC to deploy the Panther loadbalancer into. If you set this option, you must also specify SecurityGroupID, SubnetOneIPRange, and SubnetTwoIPRange.
     Default: ''
     AllowedPattern: '^(vpc-[0-9a-f]{10,})?$'
+  FullDeployment:
```
austinbyers, line 173: I would rename this `BootstrapStackOnly` and invert the logic. That makes it much more clear exactly what the behavior is

deployments/master.yml
```diff
@@ -170,6 +170,13 @@ Parameters:
...
+    Type: String
+    Description: Flag to indicate stack initialization or full deployment.
```
austinbyers, line 175: If this setting is toggled on an existing deployment, it will wipe out every other stack. Let's make that perfectly clear

```yml
  BootstrapStackOnly:
    Type: String
    Default: false
    Description: If true, only the bootstrap stack will be deployed (WARNING: ALL NON-BOOTSTRAP PANTHER INFRA WILL BE DELETED)
```

deployments/master.yml
```diff
@@ -179,8 +186,10 @@ Mappings:
       Version: ${{PANTHER_VERSION}} # E.g. "1.14.0"
 
 Conditions:
+  FullDeployment: !Equals [!Ref FullDeployment, 'true']
   RegistryProvided: !Not [!Equals [!Ref ImageRegistry, '']]
   OnboardSelf: !Equals [!Ref OnboardSelf, true]
+  CreateOnboard: !And [!Condition FullDeployment, !Condition OnboardSelf]
```
austinbyers, line 192: I don't think you need the `OnboardSelf` condition anymore, this could just be

```suggestion
  CreateOnboard: !And [!Condition FullDeployment, !Equals [!Ref OnboardSelf, true]]
```

(with the logic flipped if we rename the parameter as I suggest above)

austinbyers (commented):
We'll have to be very careful - this setting can never be used on an existing deployment or it will destroy all non-bootstrap resources

TDaglis:
@austinbyers 
I replaced the Parameter with the one you proposed.
I kept the `FullDeployment` condition naming the same and I just updated it to use the new parameter
```
FullDeployment: !Equals [!Ref BootstrapStackOnly, 'false']
```
I kept the condition naming the same since it appears cleaner to me, it is internally used and the full context is provided by the Parameter description. If you prefer to rename the condition as well, just let me know.

I also removed the onboard self condition in favor of the compound condition

austinbyers (approved):
Looks good to me! CI is complaining about a format mismatch, just run `mage fmt` from the root of the repo to fix

TDaglis:
> Looks good to me! CI is complaining about a format mismatch, just run `mage fmt` from the root of the repo to fix

`mage fmt` was updating `package-lock.json.`. After merging the latest updates from master, it appears that `mage fmt` no longer produces chages

PR #2333, opened 2021-01-04
fix: issue with bash echo command

3nvi:
## Background

When a PR is executed against Panther from a **forked** repository, a conditional bash script runs informing the user to run `mage autofmt`.

Currently, there is a bug in the bash script which resulted into CircleCI `setup` crashing and not finishing. This PR fixes it by addressing a small typo

## Changes

- Correctly reference ENV variable
- Remove `!` from the end of a sentence since it opens `dquote`

## Testing

- Locally replicated the fix in my terminal


.circleci/config.yml
```diff
@@ -101,7 +101,7 @@ commands:
             # https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
             if [[ -v CIRCLE_PR_USERNAME ]]; then
                 # We cannot commit and push the differences automatically, inform the user.
-                echo "Hey there $(CIRCLE_PR_USERNAME)!"
+                echo "Hey there $CIRCLE_PR_USERNAME"
```
alxarch, line 104: The mistake was that it should be using `{}` instead of `()`
The `!` does not seem to be affecting quoting in some way as long as you keep the `{}`.
`echo "Hey there ${CIRCLE_PR_USERNAME}!"` works and I suggest we always use `{}` in string substitutions to have a consistent syntax.

alxarch (approved):

.circleci/config.yml
3nvi: Ι tried this in my terminal

```bash
$ CIRCLE_PR_USERNAME=aggelos
$ echo "Hey there ${CIRCLE_PR_USERNAME}!"
```

and it still complains about non-ending string. Can you verify?

austinbyers: Right, let's not end the string with the exclamation point, that doesn't work for me either

`echo "Hey there ${CIRCLE_PR_USERNAME}"` (with the braces as suggested) will be fine

austinbyers (approved):

PR #2339, opened 2021-01-05
Deploy with existing subnets

### LINKED ISSUE #2159: Allow deployment of Panther into existing subnets
### Description

By default, Panther creates a new VPC, Security Group, and two new subnets to deploy  the panther web app load balancer into. This behavior can be overridden, and  users can instead specify an existing VPC & Security Group, and the deployment will create two subnets within that VPC.

Some teams do not have access to create subnets in their network. In this case, it would be preferable to also be able to specify existing subnets and have Panther use those.

### Related Services

The panther load balancer would need to be updated to accept existing subnets, instead of the ones created by Panther.
 
### Designs

None

### Acceptance Criteria

- A user deploys Panther, specifying an existing VPC,  Security Group,  and two existing subnets
- Panther uses these existing networking resources when configuring the web application and the web application load balancer
- Panther does not create any networking infrastructure except for the load balancer.

nhakmiller:
## Background

We already provide users the option to deploy the Panther web application into an existing VPC & Security Group. When a user takes this deployment option, we require them to provide us two IP ranges within that subnet and then we generate subnets with those IP ranges assigned.

We've received several ([several](https://panther-labs-oss.slack.com/archives/CVANBGXS9/p1593681726129800)) requests to allow Panther to be deployed into existing subnets as well. That would allow users to deploy the Panther web application into 100% existing networking infrastructure, which means that existing networking tooling can be taken advantage of to provide a secure place for Panther to run.

This PR adds support for that functionality. Users can now specify the `SubnetOneID` and `SubnetTwoID` parameters. If these parameters are specified, then the `VpcID` and `SecurityGroupID` parameters must also be specified. If `SubnetOneID` and `SubnetTwoID` are specified, then the `SubnetOneIPRange` and `SubnetTwoIPRange` parameters are ignored.

It is still the responsibility of the user to ensure all networking is configured properly and that the web application is reachable by their users if this optional deployment method is used.

Closes #2159.

## Changes

- Allow users to specify existing subnet IDs instead of IP ranges

## Testing

- I did a test deploy with `mage deploy`, deployed an EC2 instance into a VPC/security group with access to the existing security group I deployed Panther into, and was able to access the web app
- I wanted to deploy with `mage master:deploy`, but I could not figure out how to specify the stack parameters when deploying with this method. Would appreciate some input from anyone on core infra if this is possible or not. I did do a `mage master:deploy` with these changes without specifying any parameters, so at least we know it doesn't break current deployments.
- I also tested & confirmed that you cannot upgrade from an existing deployment with Panther created subnets to a deployment with user specified/created subnets. This wasn't a requirement, but it's good to know.


deployments/bootstrap.yml
```diff
@@ -93,7 +93,15 @@ Parameters:
   VpcID:
     Type: String
     Description: An existing VPC to deploy Panther into
-    AllowedPattern: '^(vpc-[0-9a-f]{10,})?$'
+    AllowedPattern: '^(vpc-[0-9a-f]{4,})?$'
```
nhakmiller, line 96: I reduced the required length from 10 to 4 because I observed default VPCs with an ID of only 6 characters, and my research didn't show up a conclusive lower bound on ID length here.

deployments/bootstrap.yml
```diff
@@ -165,6 +173,7 @@ Conditions:
   CreateAlarmSNSTopic: !Equals [!Ref AlarmTopicArn, '']
   IsDeployFromSource: !Equals [!Ref DeployFromSource, true]
   CreateVpc: !Equals [!Ref VpcID, '']
+  CreateSubnets: !Equals [!Ref SubnetOneID, '']
```
nhakmiller, line 176: The assumption being if one is specified the other must be as well.

deployments/bootstrap.yml
```diff
@@ -213,7 +222,7 @@ Resources:
     # trying to set the bucket policy. An easy fix is to just delay this resource until some
     # networking setup has completed, it doesn't particularly matter which resource this depends on,
     # as long as it happens later in the stack creation.
-    DependsOn: PublicSubnetOne
+    DependsOn: TargetGroup
```
nhakmiller, line 225: We are running out of network resources that are unconditionally created! Presumably we will never have users providing their own load balancers...

deployments/bootstrap.yml
3nvi: I'm ok with that

deployments/bootstrap.yml
3nvi: I don't think this will ever be a case.

3nvi (approved):
Great work @nhakmiller . LGTM

deployments/bootstrap.yml
```diff
@@ -906,8 +923,14 @@ Resources:
         - Key: access_logs.s3.bucket
           Value: !Ref AuditLogs
       Subnets:
-        - !Ref PublicSubnetOne
-        - !Ref PublicSubnetTwo
+        - !If
+          - CreateSubnets
+          - !Ref PublicSubnetOne
+          - !Ref SubnetOneID
```
austinbyers, line 929: Nit: I think these can be on one line, just like you have it in the outputs section below:

```yml
    - !If [CreateSubnets, !Ref PublicSubnetOne, !Ref SubnetOneID]
```

austinbyers (approved):
Nice! Much needed

PR #2343, opened 2021-01-05
fix: axios issue from AWS Amplify

3nvi:
## Background

We use Amplify for its `Auth` package. Amplify internally uses  `axios`  which recently had a reported vulnerability. Amplify has since fixed the package, but no release has yet been cut. Even if it did,  we'd require a backport, since we are using `v2.x.x.` while Amplify is on `v3.x.x`.

Until all that happens, we manually forced Amplify to  use `v0.21.1` of axios instead of the current `v0.19.0`. Went through  axios' [changelog](https://github.com/axios/axios/blob/hotfix/0.21.1/CHANGELOG.md) and didn't see any breaking changes.

This will fix our `npm audit` issue for now and we'll upgrade to Amplify's patch version as soon as it becomes available

## Changes

- Force all axios resolutions to latest `v0.21.1`

## Testing

- `npm audit`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,6 +60,19 @@
         "zen-observable": "^0.8.6"
       },
       "dependencies": {
+        "axios": {
+          "version": "0.21.1",
+          "resolved": "https://registry.npmjs.org/axios/-/axios-0.21.1.tgz",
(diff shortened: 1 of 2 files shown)


alexmylonas (approved):
LGTM!

PR #2344, opened 2021-01-05
Update max lambda size to 10240

rleighton:
## Background

The max size (I thought) I fixed earlier but it seems to have been put back to 3008. Lambda now allows 10240. Master stack has this correct.

## Changes

- Change max in log process CF and in the alarm generation code.

## Testing

- mage test:ci
- benchmarks using large lambdas

### DIFF
--- a/internal/core/custom_resources/resources/alarms_lambda.go
+++ b/internal/core/custom_resources/resources/alarms_lambda.go
@@ -40,7 +40,7 @@ const (
 type LambdaAlarmProperties struct {
 	AlarmTopicArn      string `validate:"required"`
 	FunctionName       string `validate:"required"`
-	FunctionMemoryMB   int    `json:",string" validate:"min=128,max=3008"`
+	FunctionMemoryMB   int    `json:",string" validate:"min=128,max=10240"`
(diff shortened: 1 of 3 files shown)


internal/core/custom_resources/resources/alarms_lambda.go
```diff
@@ -40,7 +40,7 @@ const (
 type LambdaAlarmProperties struct {
 	AlarmTopicArn      string `validate:"required"`
 	FunctionName       string `validate:"required"`
-	FunctionMemoryMB   int    `json:",string" validate:"min=128,max=3008"`
+	FunctionMemoryMB   int    `json:",string" validate:"min=128,max=10240"`
```
kostaspap, line 43: Maybe just remove the max from here? AWS will keep updating it, if we put an invalid value, deployment will fail anyway before that.

kostaspap (approved):

internal/core/custom_resources/resources/alarms_lambda.go
rleighton: Done

Issue #2267, opened 2020-12-11
Create Add Data Models Form


mattthetraveler:
### Description

As an analyst, I want to be able create and maintain universal data models in the UI that I can reference when writing new rules and policies in the Python editor (see example [here](https://github.com/panther-labs/panther-analysis/pull/156).

Ideally, this is a separate sub-tab in Settings that presents a view of all the Data Models I have active and where I can also add new Data Models.

### Related Services

[See Backend services here](https://github.com/panther-labs/panther/pull/1945)

### Designs

Here is the[ RFC](https://github.com/panther-labs/panther-enterprise/tree/master/designs/security/simplify-detections)

Here are the [Designs](https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/df9d59c8-eeb1-452d-a5f5-3712b64010be/collections/65b935ef-579f-4620-9530-c359b8d416e3)

### Acceptance Criteria

- To add new Data Model, form should have two sections: Settings and Data Model Mappings
- Within Setting tab, available attributes should be Log Types, Detection ID, and Enabled
- Within Data Model Mappings tab: 
   - there should be a field to drop in a YAML function. This should be mandatory for the user.
   - there should an option to drop in a Python Module. This should be optional for the user.
- Within the Data Models view page:
   - Ability to bulk select Data Models with the option to "Delete"
   - Ability to filter by "Most Recent" and "Last Modified", should default to "Last Modified"

3nvi:
Closeed via #1876

PR #2345, opened 2021-01-05
Data Models [1/4]: Allow Data Model creation

3nvi:
## Background

This is the first out of a series of PRs that will allow a user to add data models into their account. In this PR we provide the data model creation functionality. Specifically:

- The user is allowed to create a data model using our Data Model form

Relates to #2267 

## Screenshots

Screen Shot 2021-01-05 at 7 41 29 PM

## Designs

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/df9d59c8-eeb1-452d-a5f5-3712b64010be/commits/latest/files/46187896-210a-4c7f-92ff-d156f6dc68b2/layers/9B65FBD6-0B23-4E29-B35B-1E3BFBD0D194?collectionId=65b935ef-579f-4620-9530-c359b8d416e3&collectionLayerId=df84b90c-1d18-47f1-bb60-f3fe8ff9d429

## Changes

- Add GraphQL schema for data models
- Add AppSync resolvers
- Create a `DataModelForm`
- Create a `CreateDataModel` page

## Testing

- Because we don't want this feature exposed to master yet, the creation form is only reachable by specifying its URL directly in  the browser tab. To make a manual test visit  `/log-analysis/data-models/new` and play with it
- Extensive Unit & Integration tests have been added

## Notes

- We don't include a nav entry intentionally until the feature is complete
- The UI has changes 2 times during development due to lack of planning, so this PR includes the finalized version of the design and not the original.


web/src/components/forms/DataModelForm/DataModelForm.tsx
```diff
@@ -0,0 +1,254 @@
...
+                  variantColor="darkgray"
+                  to={urls.logAnalysis.dataModels.list()}
+                >
+                  Cancel
+                </LinkButton>
+                <SubmitButton icon="check-outline" variantColor="green">
+                  Save
+                </SubmitButton>
+              </Flex>
+            </Breadcrumbs.Actions>
```
alexmylonas, line 246: nitpick: we usually place breadcrumbs on top of component and/or a separate component

alexmylonas (approved):
Solid work Aggelos! I would personally prefer to base this on a feature branch instead of master but feel free to do what fits you best!

3nvi:
> Solid work Aggelos! I would personally prefer to base this on a feature branch instead of master but feel free to do what fits you best!

Yeah considered it. The  reason I targeted master is that I kept it "hidden", while not breaking anything. Even if master gets shipped like that, there's no issue whatsoever

web/src/components/forms/DataModelForm/DataModelForm.tsx
3nvi: you're right will put them to the top on the subsequent  PR ;)

PR #2352, opened 2021-01-06
Configurable dynamo KV store billing mode

tatehanawalt:
## Background

kv-store currently deploys with the read/write capacity mode as 'provisioned'.  The typical deployment has minimal if any load on the kv-store table. Switching the default to 'on-demand' rather than 'provisioned' will save most deployments a bit of money. 

## Changes

• Adding parameter to cloudformation and panther_config.yml so a user can deploy panther with either provisioned or on-demand read/write capacity modes.
• Set Default read/write capacity mode to 'on-demand' for kv-table

## Testing

• New deployment with yaml config parameter set to 'provisioned'
• New deployment with yaml config parameter set to 'on-demand'
• Update existing deployment from 'on-demand' to 'provisioned'
• `mage test:ci`


deployments/master.yml
```diff
@@ -139,6 +139,11 @@ Parameters:
     Description: Configure Panther to automatically onboard itself as a data source
     AllowedValues: [true, false]
     Default: true
+  ProvisionKvTable:
+    Type: String
+    Description: If true, use PROVISIONED billing and autoscaling for panther-kv-store Dynamo table
+    AllowedValues: [true, false]
+    Default: true
```
austinbyers, line 146: The default should actually be `false` - most customers do not need a pre-provisioned table unless they have rules that leverage them and are processing large volumes of log data

deployments/panther_config.yml
```diff
@@ -66,6 +66,15 @@ Infra:
...
     - requests==2.23.0
 
+  # Select DynamoDB kv-table Read/Write Capacity Mode (provision vs on-demand)
+  # https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadWriteCapacityMode.html
+  # Boolean:
+  #   true  - provision mode. Fixed Billing with static capacity bandwidth.
+  #   false - on-demand mode. Flexible billing with automatic capacity handling.
+  #
+  # You can switch read/write capacity modes once every 24 hours per aws docs.
+  ProvisionKvTable: true
```
austinbyers, line 76: Again, the goal of this change is to actually disable it by default, which will save some money for most deployments where the table isn't used much or at all

The comment can be simplified to something like:

```yml
# Enable provisioned capacity and autoscaling for the key-value Dynamo store available to the Python engines.
# You may want to enable this if you make a high volume of requests to the key-value store, but otherwise
# the default (per-pay-use) billing is typically cheaper for small-to-medium workloads
ProvisionKvTable: false
```

austinbyers (commented):
Nice! The changes look pretty good, but we actually want to *disable* this setting by default.

A pre-provisioned Dynamo table incurs a fixed cost for the table as well as costs for the CloudWatch alarms which monitor its scaling. Since most deployments do not use this table much if at all, disabling it by default saves most deployments a bit of money. It's not much, but it's nice to have simpler/cheaper infra when we can easily do so.

Can you update your PR with:

* A more descriptive title
* An explanation for why we're making the change (you can paraphrase or copy what I wrote above)
* The "Changes" section doesn't have to so low-level. You can say something like "Changes: kv-store provisioning is now configurable; off by default"

austinbyers:
Adding @nhakmiller as the owner of cloud security to see if he agrees with the decision to disable pre-provisioned capacity on the kv table by default

deployments/master.yml
austinbyers: I know this was my original suggestion, but I'm debating whether it makes sense to do this instead:

```yml
KvTableBillingMode:
    Type: String
    Description: Configure how the panther-kv-store Dynamo table should be provisioned and billed
    AllowedValues: [PROVISIONED, PAY_PER_USE]
    Default: PAY_PER_USE
```

This "bubbles up" the real Dynamo configuration to the user, which makes it more explicit for people familiar with Dynamo. But then, it may be easier for users to understand and toggle a boolean value. What do you think?

I'm leaning toward using the more explicit version

nhakmiller: Explicit is good. If we have room to expand the description a bit, I might add something indicated this is used by the policies & rules so people know what this table is and why they might want to increase its capacity. Something like:

```
If true, use PROVISIONED billing and autoscaling for panther-kv-store Dynamo table. This table is used by the custom python policies & rules to store state, and provisioned billing may be cheaper if this feature is used heavily.
```

nhakmiller (commented):
> Adding @nhakmiller as the owner of cloud security to see if he agrees with the decision to disable pre-provisioned capacity on the kv table by default

At face value I don't have a problem with this, very few out of the box detections use this table and those that do are usually at a low volume relatively speaking (tied to user logins for example). So the intuition that most users don't need provisioned capacity seems reasonable.

That being said, I do have to ask was this a decision based on any specific numbers? I've never observed undue dynamo db costs in any of our hosted environments. I'm concerned that at best this is a pre-mature optimization, which isn't the end of the world just curious if we observed/someone reported high dynamodb costs that prompted this.

deployments/master.yml
tatehanawalt: I will update the allowed values to be explicit options rather than true false, expand on the descriptions, include billing implications for each option, And default to on-demand capacity mode.

Since this parameter changes the read/write capacity mode should we name the parameter `KvTableReadWriteCapacityMode` or `KvTableRWCapacity`?

And use allowed values `PROVISIONED` and `ON_DEMAND` with billing implications in the description?

austinbyers: > Since this parameter changes the read/write capacity mode should we name the parameter KvTableReadWriteCapacityMode

No, let's just use the same name Dynamo itself uses: `KvBillingMode`. It's generally understood that "provisioned" billing will require read/write capacity

> KvTableRWCapacity

This would be the parameter name if we wanted to configure the base capacity number (5, 10, 20, etc). Let's keep capacity out of the name

> 'BillingMode' with values PROVISIONED and PAY_PER_USE

CloudFormation is what we care about here - so those values are what we'll put in the parameter options. The web console may choose to display the information in slightly different ways over time; let's stick with the official Dynamo terminology.

austinbyers:
> That being said, I do have to ask was this a decision based on any specific numbers? I've never observed undue dynamo db costs in any of our hosted environments. I'm concerned that at best this is a pre-mature optimization, which isn't the end of the world just curious if we observed/someone reported high dynamodb costs that prompted this.

This is more just tech debt that we've always had (hence the `#TODO` comment above the KV table in the CFN template since the day it was added). It's not a high priority, but it was inspired by our higher-than-expected CloudWatch costs; removing provisioned capacity also removes the 2 CloudWatch alarms that Dynamo uses for autoscaling. Less infra will also lead to slightly faster deploys, etc.

In short - if most people don't need it, it saves a bit of time and money to not deploy it.

But more than anything, this was pedagogical - it's a relatively easy fix that was useful for Tate to learn about CloudFormation and the deployment process.

deployments/core.yml
```diff
@@ -145,6 +149,9 @@ Mappings:
 
 Conditions:
   AttachLayers: !Not [!Equals [!Join ['', !Ref LayerVersionArns], '']]
+  KvProvisioningEnabled: !Equals
+    - !Ref KvTableBillingMode
+    - PROVISIONED
```
austinbyers, line 154: Nit: I think it's fine to keep this in one line:

```suggestion
  KvProvisioningEnabled: !Equals [!Ref KvTableBillingMode, PROVISIONED]
```

deployments/core.yml
```diff
@@ -68,6 +68,10 @@ Parameters:
     Type: String
     Description: SNS topic ARN for data that is sent as input to Log Analysis
     AllowedPattern: '^arn:(aws|aws-cn|aws-us-gov):sns:[a-z]{2}-[a-z]{4,9}-[1-9]:\d{12}:\S+$'
+  KvTableBillingMode:
+    Type: String
+    Description: Configure billing mode for DynamoDB table panther-kv-store. This table is used by the custom python policies & rules to store state, and provisioned billing may be cheaper if this feature is used heavily.
```
austinbyers, line 73: Great description! They're not "custom" python policies and rules - any python can use this table, and in fact we ship some rules (disabled by default) that use this feature.

Just remove the "custom"

austinbyers (approved):
Awesome, great work! Can you update the PR title to something more succinct?

PR #2356, opened 2021-01-07
Change name of destination functions

kostaspap:
## Background

Changing name from `destination_overrides` to `destinations` according to this discussion: https://github.com/panther-labs/panther/pull/2165#discussion_r534324500

It seems that this feature was not released officially in 1.14, so we get some time to fix names like this

## Changes

- Updated name of method

## Testing

- mage test:python

### DIFF
--- a/internal/log_analysis/rules_engine/src/rule.py
+++ b/internal/log_analysis/rules_engine/src/rule.py
@@ -353,7 +353,7 @@ class Rule:
         return description
 
     def _get_destinations(self, event: Mapping, use_default_on_exception: bool = True) -> Optional[List[str]]:
-        if not hasattr(self._module, 'destination'):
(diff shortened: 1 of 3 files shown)


wey-chiang (approved):
LGTM, we need to make sure the docs are updated properly to show the correct change in python syntax from `def destination_override` to `def destinations`. 

I'll update our docs internally on confluence

wey-chiang:
What release are we aiming to include these changes in? @kostaspap

kostaspap:
> What release are we aiming to include these changes in? @kostaspap

1.15!

kostaspap:
> LGTM, we need to make sure the docs are updated properly to show the correct change in python syntax from `def destination_override` to `def destinations`.
> 
> I'll update our docs internally on confluence

Yup, will do!

PR #2359, opened 2021-01-07
Creating immutable events

### LINKED ISSUE #2350: Make Events Immutable in the Rules Engine
### Description

Users should not be able to modify events from the rule engine and have them saved in the data lake.

### Related Services

Rules Engine

### Designs

N/A

### Acceptance Criteria

- Events are immutable in the rule matches

kostaspap:
## Background

Closes #2350 

## Changes

Modifying an event will now throw an exception

## Testing

- mage test:ci
- E2E updated nested item:
Screen Shot 2021-01-08 at 3 59 40 PM
Screen Shot 2021-01-08 at 3 59 27 PM


internal/log_analysis/rules_engine/src/enriched_event.py
```diff
@@ -15,16 +15,16 @@
...
 from collections.abc import Mapping
-from typing import Any, Iterator
+from typing import Any, Iterator, Optional
 
 from .data_model import DataModel
 from .logging import get_logger
 
 
-class EnrichedEvent(Mapping):
+class PantherEvent(Mapping):
```
kostaspap, line 24: I saw that we were already creating our own event in order to offer UDM. This event was making sure that the events are immutable. I am now making sure that all events going through the Rules Engine use this class.

rleighton: This is good but does it handle the case of modifying under a key? Something like:
```
foo = event["key"]
foo["bar"] = "i mutated the event!"
```

kostaspap: Hmmmm, you are right. I don't think it does.

kostaspap: I'm not sure we can avoid a deep copy then. @georgepsarakis any thoughts? We can do make a `Dict[str, Any]` immutable.

rleighton: I liked the idea of "freezing" the event after we read it (one time) then passing that to rules, keep the original for writing because we need to mutate to add the rule specific fields.

rleighton: BTW, the changes in this PR still look like an improvement tho.

rleighton: Hmmm, looks like a big topic: https://stackoverflow.com/questions/4828080/how-to-make-an-immutable-object-in-python

rleighton: Perhaps we deepcopy once, pass that to all rules (yes, if they change it that could affect other rules, too bad). Pass the original event to datalake. This way we are protected.

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -55,7 +55,7 @@ def analyze_single_rule(self, raw_rule: dict, test_spec: Mapping) -> Dict[str, A
         # enrich the event to have access to field by standard field name
         #  via the `udm` method
         if 'p_log_type' in event and event['p_log_type'] in self.log_type_to_data_models:
-            event = EnrichedEvent(event, self.log_type_to_data_models[event['p_log_type']])
+            event = PantherEvent(event, self.log_type_to_data_models[event['p_log_type']])
```
lindsey-w, line 59: Should this `if` be removed and changed to something like:
```
event = PantherEvent(event, self.log_type_to_data_models.get(event.get('p_log_type')))
```
to be consistent with the `analyze` event handling / ensure test events are handled the same way?

lindsey-w: Oh now I see... that would break the `MISSING_UDM_EXCEPTION` logic - which notifies the user that both `p_log_type` must be specified and a DataModel exists for the logType.  These events aren't stored anywhere (I assume?) since they are tests - so maybe this is fine?

internal/log_analysis/rules_engine/src/enriched_event.py
nhakmiller: > Perhaps we deepcopy once, pass that to all rules (yes, if they change it that could affect other rules, too bad). Pass the original event to datalake. This way we are protected.

I like this idea, I'm already having some thoughts on how we could ~~abuse~~ make use of this as a sort of "cache" for expensive operations that multiple rules might want to make.

georgepsarakis: Since we are handling JSON-serializable objects, we only have to account for `list` and `dict` objects. 
As far as I understand, the only point that we would consider mutation is [here](https://github.com/panther-labs/panther/blob/a6bcd4e91022f62ca298f26b77964c916f8c7796/internal/log_analysis/rules_engine/src/output.py#L242-L246), since we are merging the custom rule fields and serializing the event in order to write it back to S3, with the additional fields this time. One note here, I think the [`update` operation](https://github.com/panther-labs/panther/blob/a6bcd4e91022f62ca298f26b77964c916f8c7796/internal/log_analysis/rules_engine/src/output.py#L245) should happen in reverse, because if someone creates a top-level key that has the same name with the `EventCommonFields` attributes, the `event` value will be retained.  

I did some research, keeping these factors in mind:
- JSON serialization performance: converting nested structured to a custom class instance in-place could lead to a custom JSON encoder being required for the final serialization step. This adds a significant overhead, serialization speed drops at 50%.
- `deepcopy` performance: seems that for a sample log event (taken from [here](https://docs.aws.amazon.com/waf/latest/developerguide/logging.html), see below) it requires ~66μs on `Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz`.

Not sure if this sounds negligible, it actually depends on the amount of records processed (I am guessing this is a large number in some cases) and the CPU core speed on Lambda. I am also guessing that it would be best to `deepcopy` before passing to each rule, in order to account for the possible tampering that @rleighton mentions above, with different rules interfering with each other in unexpected ways. Thus, the amount of time spent here is also multiplied with the number of rules for the specific log type (?).

My suggestion would be to initially create shallow copies for each event and then lazily wrap in immutable wrappers their nested objects, when they are retrieved (if ever). I have outlined a possible implementation along with benchmarks and assertions in the following snippet: 

<details>
  <summary><b>Lazily Convert Nested JSON Object to Immutable Objects</b></summary>

```python
from abc import ABC, abstractmethod
from collections.abc import Mapping, Sequence
from copy import deepcopy
import json
from time import time


event = {
    'action': 'BLOCK',
    'formatVersion': 1,
    'httpRequest': {'args': '',
                    'clientIp': '3.3.3.3',
                    'country': 'US',
                    'headers': [{'name': 'Host',
                                 'value': 'localhost:1989'},
                                {'name': 'User-Agent',
                                 'value': 'curl/7.61.1'},
                                {'name': 'Accept', 'value': '*/*'},
                                {'name': 'xssfoo',
                                 'value': '<frameset '
                                          'onload=alert(1)>'},
                                {'name': 'bar', 'value': '10 AND 1=1'}],
                    'httpMethod': 'GET',
                    'httpVersion': 'HTTP/1.1',
                    'requestId': 'rid',
                    'uri': '/foo'},
    'httpSourceId': '-',
    'httpSourceName': '-',
    'nonTerminatingMatchingRules': [],
    'rateBasedRuleList': [],
    'ruleGroupList': [
        {
            'excludedRules': None,
            'nonTerminatingMatchingRules': [
                {
                    'action': 'COUNT',
                    'ruleId': 'RuleB-SQLi',
                    'ruleMatchDetails': [
                        {'conditionType': 'SQL_INJECTION',
                         'location': 'HEADER',
                         'matchedData': ['10',
                                         'and',
                                         '1']}
                    ]
                 }
            ],
         'ruleGroupId': 'arn:aws:wafv2:us-east-1:123456789012:global/rulegroup/hello-world/c05lb698-1f11-4m41-aef4-99a506d53f4b',
         'terminatingRule': {'action': 'BLOCK',
                             'ruleId': 'RuleA-XSS',
                             'ruleMatchDetails': None}}
    ],
    'terminatingRuleId': 'RG-Reference',
    'terminatingRuleMatchDetails': [{'conditionType': 'XSS',
                                     'location': 'HEADER',
                                     'matchedData': ['<', 'frameset']}],
    'terminatingRuleType': 'GROUP',
    'timestamp': 1592361810888,
    'webaclId': 'arn:aws:wafv2:us-east-1:123456789012:global/webacl/hello-world/5933d6d9-9dde-js82-v8aw-9ck28nv9'
}


event_clone = deepcopy(event)
event_json = json.dumps(event)

class ImmutableContainerMixin(ABC):
    _conversions = {}

    @classmethod
    @abstractmethod
    def mutable_type(cls):
        pass

    def __init_subclass__(cls, **kwargs):
        cls.register(cls.mutable_type())

    def __init__(self, container):
        self._container = self._shallow_copy(container)
        self._conversions = tuple(self.__class__._conversions.items())
        self._registered_types = tuple(klass for klass, _ in self._conversions)

    @abstractmethod
    def _shallow_copy(self, obj):
        pass

    @classmethod
    def register(cls, klass):
        cls._conversions[klass] = cls

    def __getitem__(self, item):
        value = self._container[item]
        if isinstance(value, self._registered_types):
            for mutable_type, immutable_type in self._conversions:
                if isinstance(value, mutable_type):
                    value = immutable_type(value)
        return value

    def __repr__(self):
        return f'{self.__class__.__name__}({self._container})'

    def __len__(self) -> int:
        return self._container.__len__()

    def __iter__(self):
        return iter(self._container)


class ImmutableDict(ImmutableContainerMixin, Mapping):
    @classmethod
    def mutable_type(cls):
        return dict

    def _shallow_copy(self, obj):
        return obj.copy()


class ImmutableList(ImmutableContainerMixin, Sequence):
    @classmethod
    def mutable_type(cls):
        return list

    def _shallow_copy(self, obj):
        return tuple(obj)


print('>>> Immutable Event tests')
immutable_event = ImmutableDict(event)

print(immutable_event)
try:
    print(immutable_event['non_existent'])
except KeyError as e:
    print(e.__class__.__name__, e)

print(immutable_event['httpRequest'])
print(immutable_event['httpRequest']['headers'])
print(immutable_event['httpRequest']['headers'][0])

try:
    immutable_event['httpRequest']['headers'] = 1
except TypeError as e:
    print(e.__class__.__name__, e)

print(immutable_event)


event_not_modified = event == event_clone
print('Verify that we did not modify the initial record:', event_not_modified)
assert event_not_modified

final_event_json = json.dumps(event)
print('Verify that the event is JSON-serializable:', json.dumps(event, indent=2))
assert final_event_json == event_json

print('>>> Benchmarks')

start_time = time()
rounds = 1_000_000
for _ in range(rounds):
    deepcopy(event)

print('Deepcopy benchmark:', (time() - start_time) / rounds)

start_time = time()
for _ in range(rounds):
    event.copy()

print('dict.copy benchmark:', (time() - start_time) / rounds)

start_time = time()

l = list(range(1000))
for _ in range(rounds):
    tuple(l)

print('Convert list to tuple benchmark:', (time() - start_time) / rounds)
```
</details>

internal/log_analysis/rules_engine/src/engine.py
kostaspap: I missed this one, thanks for pointing out! There is logic inside the UDM that will trigger an exception if udm function is accessed by no model has been specified!

internal/log_analysis/rules_engine/src/enriched_event.py
```diff
@@ -44,8 +44,22 @@ def __iter__(self) -> Iterator:
     def __len__(self) -> int:
         return len(self._data)
 
+    def __contains__(self, key: object) -> bool:
+        return key in self._data
+
+    def __hash__(self) -> int:
+        return self._data.__hash__()
```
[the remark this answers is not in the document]
kostaspap: Good point, some copy-paste while I was working on it. Will remove.

internal/log_analysis/rules_engine/src/enriched_event.py
kostaspap: I have tried it locally with an S3 Server Access event and I see in your benchmark that using the new ImmutableDict is 2 orders of magnitude faster than the deep copy. It also protects against one rule messing up the event passed to another rule that could have really bad consequences e.g. detections not working.

internal/log_analysis/rules_engine/src/enriched_event.py
```diff
@@ -44,8 +44,22 @@ def __iter__(self) -> Iterator:
     def __len__(self) -> int:
         return len(self._data)
 
+    def __contains__(self, key: object) -> bool:
```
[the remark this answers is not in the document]
kostaspap: Good point

internal/log_analysis/rules_engine/tests/test_immutable.py
```diff
@@ -0,0 +1,65 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from unittest import TestCase
+
+from ..src.immutable import ImmutableDict, ImmutableList
+
```
rleighton, line 20: Probably also want a test for `TestImmutableNestedList` where there is a list inside a dict and you modify it.

internal/log_analysis/rules_engine/tests/test_immutable.py
```diff
@@ -0,0 +1,65 @@
...
+    def setUp(self) -> None:
+        self.initial_dict = {'t': 10, 'a': [{'b': 1, 'c': 2}], 'd': {'e': {'f': True}}}
+        self.immutable_dict = ImmutableDict(self.initial_dict)
+
+    def test_assignment_not_allowed(self) -> None:
+        with self.assertRaises(TypeError):
+            # pylint: disable=E1137
+            self.immutable_dict['a'] = 1  # type: ignore
+
```
rleighton, line 32: I know you test a nested access in `test_modify_event()` but for completeness you probably want that case here as well.

internal/log_analysis/rules_engine/src/immutable.py
```diff
@@ -0,0 +1,94 @@
...
+    def _shallow_copy(self, obj: Any) -> Any:
+        """Creates a shallow copy of the given object"""
+
+    def extra_constructor_arguments(self) -> Dict:  # pylint: disable=R0201
+        """Define additional constructor arguments that must be
+           passed through to nested objects"""
+        return {}
+
+    @no_type_check
+    def __getitem__(self, item):
```
rleighton, line 60: Clever! Lazy eval.

internal/log_analysis/rules_engine/src/enriched_event.py
nhakmiller: I had one question about how this plays out. If `rule` modifies an event to add a new field, can you reference that new field in the `dedup` or `alert_context` functions?

kostaspap: > I had one question about how this plays out. If `rule` modifies an event to add a new field, can you reference that new field in the `dedup` or `alert_context` functions?

Not it would not be possible. We are basically making the event read-only (which we always assumed it to be but never enforced it)

kostaspap: We could support it formally in a different way - e.g. we could expose a new method e.g. `event.context()` that returns a mutable dict that lived during a single rule invocation.

nhakmiller: I don't have a strong opinion either way, just curious because that is what the customer was trying to do so I wanted to know if we should tell them they can/can't still do that. Currently it's also possible with a global variable in the rule, but I've always slightly worried about OOM issues if every single event is storing some dozen bytes in a global dictionary that only gets cleared out when the container resets.

internal/log_analysis/rules_engine/tests/test_immutable.py
kostaspap: Added

internal/log_analysis/rules_engine/tests/test_immutable.py
kostaspap: Added

internal/log_analysis/rules_engine/src/enriched_event.py
kostaspap: > I don't have a strong opinion either way, just curious because that is what the customer was trying to do so I wanted to know if we should tell them they can/can't still do that. Currently it's also possible with a global variable in the rule, but I've always slightly worried about OOM issues if every single event is storing some dozen bytes in a global dictionary that only gets cleared out when the container resets.

I think it is possible to support it, but would require another (probably short) PR. Since the same event is used by multiple rules, any (temporary) state we store in that event will be shared across all rules! So that will help with expensive operations, as you noted above in this thread.

rleighton (approved):

georgepsarakis:
Unfortunately I discovered some bugs that I think we need to address before merging this PR.

1. If the value of a top-level key is a list, an exception is raised, since we will try to pass additional keyworded arguments to `ImmutableList` that do not match its constructor signature.
2. If a `DataModel` is passed to `PantherEvent`, the `jsonpath-ng` search will return the **original nested object which can then be mutated**, in which case the mutation will be present in the original event as well. We should ensure that we always return immutable objects, for all access methods.
3. `Mapping` does implement the equality check magic method, but `Sequence` does not. This may be a source of confusion in some cases, e.g. when someone is comparing a tuple or list with an `ImmutableList` instance, or two `ImmutableDict` instances containing a nested list internally.
4. `ImmutableDict` and `PantherEvent` differ only by the `udm` method, which allows searching by path in the event. Since the paths in the `DataModel` are specified to begin from the top level, returning a `PantherEvent` instance for a nested object and executing the `udm` method on it, would yield incorrect results. It seems more straightforward and simpler to always return `ImmutableDict`.

The above have been addressed, along with unit tests, in the following patch (and will be committed on Monday, unless there are any objections / further suggestions). Note that the implementation is simplified due to the solution of (4):

<details>
  <summary><b>Diff</b></summary>

```diff
diff --git a/internal/log_analysis/rules_engine/src/enriched_event.py b/internal/log_analysis/rules_engine/src/enriched_event.py
index f64aa6e5..b5841e81 100644
--- a/internal/log_analysis/rules_engine/src/enriched_event.py
+++ b/internal/log_analysis/rules_engine/src/enriched_event.py
@@ -34,12 +34,6 @@ class PantherEvent(ImmutableDict):  # pylint: disable=R0901
         super().__init__(event)
         self.data_model = data_model
 
-    def extra_constructor_arguments(self) -> dict:
-        return {'data_model': self.data_model}
-
-    def conversion_overrides(self) -> dict:
-        return {dict: self.__class__}
-
     def udm(self, key: str) -> Any:
         """Converts standard data model field to logtype field"""
         if not self.data_model:
@@ -51,7 +45,7 @@ class PantherEvent(ImmutableDict):  # pylint: disable=R0901
             if json_path:
                 matches = json_path.find(self._container)
                 if len(matches) == 1:
-                    return matches[0].value
+                    return self._ensure_immutable(matches[0].value)
                 if len(matches) > 1:
                     raise Exception(
                         'JSONPath [{}] in DataModel [{}], matched multiple fields.'.format(json_path, self.data_model.data_model_id)
@@ -60,6 +54,6 @@ class PantherEvent(ImmutableDict):  # pylint: disable=R0901
             # we are dealing with method
             method = self.data_model.methods.get(key)
             if callable(method):
-                return method(self._container)
+                return self._ensure_immutable(method(self._container))
         # no matches, return None by default
         return None
diff --git a/internal/log_analysis/rules_engine/src/immutable.py b/internal/log_analysis/rules_engine/src/immutable.py
index a5056b2e..e9d00a0f 100644
--- a/internal/log_analysis/rules_engine/src/immutable.py
+++ b/internal/log_analysis/rules_engine/src/immutable.py
@@ -37,32 +37,24 @@ class ImmutableContainerMixin(ABC):
 
     def __init__(self, container: Any):
         self._container = self._shallow_copy(container)
-        conversions = self.__class__._CONVERSIONS.copy()
-        conversions.update(self.conversion_overrides())
-        self._registered_conversions = tuple(conversions.items())
+        self._registered_conversions = tuple(self.__class__._CONVERSIONS.items())
         self._registered_types = tuple(klass for klass, _ in self._registered_conversions)
 
-    def conversion_overrides(self) -> dict:  # pylint: disable=R0201
-        """Override mutable to immutable type correspondence for this class/instance"""
-        # TODO: redesign the type conversion registration process
-        return {}
-
     @abstractmethod
     def _shallow_copy(self, obj: Any) -> Any:
         """Creates a shallow copy of the given object"""
 
-    def extra_constructor_arguments(self) -> Dict:  # pylint: disable=R0201
-        """Define additional constructor arguments that must be
-           passed through to nested objects"""
-        return {}
-
     @no_type_check
     def __getitem__(self, item):
-        value = self._container[item]
+        value = self._ensure_immutable(self._container[item])
+        return value
+
+    def _ensure_immutable(self, value):
         if isinstance(value, self._registered_types):
             for mutable_type, immutable_type in self._registered_conversions:
                 if isinstance(value, mutable_type):
-                    value = immutable_type(value, **self.extra_constructor_arguments())
+                    value = immutable_type(value)
+                    break
         return value
 
     def __repr__(self) -> str:
@@ -96,6 +88,16 @@ class ImmutableList(ImmutableContainerMixin, Sequence):  # pylint: disable=R0901
     def _shallow_copy(self, obj: list) -> tuple:
         return tuple(obj)
 
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            return self._container == other._container
+        elif isinstance(other, (tuple, list)):
+            # Allow comparison with lists and tuples
+            return len(self._container) == len(other) and \
+                   self._container == tuple(other)
+        else:
+            return False
+
 
 ImmutableList.register()
 ImmutableDict.register()
diff --git a/internal/log_analysis/rules_engine/tests/test_engine.py b/internal/log_analysis/rules_engine/tests/test_engine.py
index 0bda6090..27ddc948 100644
--- a/internal/log_analysis/rules_engine/tests/test_engine.py
+++ b/internal/log_analysis/rules_engine/tests/test_engine.py
@@ -294,13 +294,13 @@ class TestEngine(TestCase):
                 rule_version='version',
                 log_type='log',
                 dedup='TypeError',
-                error_message='\'PantherEvent\' object does not support item assignment: rule_id_1.py, '
+                error_message='\'ImmutableDict\' object does not support item assignment: rule_id_1.py, '
                 'line 2, in rule    event["key"]["nested_key"] = "not_value"',
                 event={'key': {
                     'nested_key': 'value'
                 }},
                 dedup_period_mins=60,
-                title='TypeError("\'PantherEvent\' object does not support item assignment")'
+                title='TypeError("\'ImmutableDict\' object does not support item assignment")'
             )
         ]
 
diff --git a/internal/log_analysis/rules_engine/tests/test_enriched_event.py b/internal/log_analysis/rules_engine/tests/test_enriched_event.py
index 4bf23dbe..938960dd 100644
--- a/internal/log_analysis/rules_engine/tests/test_enriched_event.py
+++ b/internal/log_analysis/rules_engine/tests/test_enriched_event.py
@@ -18,6 +18,7 @@ from unittest import TestCase
 
 from ..src.data_model import DataModel
 from ..src.enriched_event import PantherEvent
+from ..src.immutable import ImmutableList, ImmutableDict
 
 
 class TestEnrichedEvent(TestCase):
@@ -157,3 +158,45 @@ class TestEnrichedEvent(TestCase):
         except Exception:  # pylint: disable=broad-except
             exception = True
         self.assertTrue(exception)
+
+    def test_assignment_not_allowed_without_udm(self):
+        # No DataModel given
+        event = {'dst': {'ip': '1.1.1.1', 'port': '2222'}}
+        enriched_event = PantherEvent(event, None)
+        with self.assertRaises(TypeError):
+            enriched_event['dst'] = 1
+
+    def test_assignment_not_allowed_udm(self):
+        event = {'dst_ip': '1.1.1.1', 'dst_port': '2222',
+                 'extra': {'timestamp': 1, 'array': [1, 2]}}
+        data_model = DataModel(
+            {
+                'versionId': 'version',
+                'mappings': [{
+                    'name': 'destination_ip',
+                    'path': 'dst_ip'
+                }, {
+                    'name': 'extra_fields',
+                    'path': 'extra'
+                }],
+                'id': 'data_model_id'
+            }
+        )
+        enriched_event = PantherEvent(event, data_model)
+        self.assertEqual(ImmutableDict(event['extra']),
+                         enriched_event.udm('extra_fields'))
+        self.assertIsInstance(enriched_event.udm('extra_fields'), ImmutableDict)
+        self.assertIsInstance(enriched_event.udm('extra_fields')['array'], ImmutableList)
+        with self.assertRaises(TypeError):
+            enriched_event.udm('extra_fields')['timestamp'] = 10
+
+    def test_nested_list_immutability(self):
+        event = {
+            'headers': [{
+                'User-Agent': 'Chrome',
+                'Host': 'google.com'
+            }]
+        }
+        enriched_event = PantherEvent(event, None)
+        self.assertIsInstance(enriched_event['headers'], ImmutableList)
+        self.assertIsInstance(enriched_event['headers'][0], ImmutableDict)
diff --git a/internal/log_analysis/rules_engine/tests/test_immutable.py b/internal/log_analysis/rules_engine/tests/test_immutable.py
index c83308bb..3aa82750 100644
--- a/internal/log_analysis/rules_engine/tests/test_immutable.py
+++ b/internal/log_analysis/rules_engine/tests/test_immutable.py
@@ -50,6 +50,16 @@ class TestImmutableDict(TestCase):
         self.assertEqual(self.immutable_dict['a'][0]['b'], 1)
         self.assertEqual(self.immutable_dict['d']['e']['f'], True)
 
+    def test_equality(self):
+        from copy import deepcopy
+
+        # Equality with dict
+        self.assertEqual(self.immutable_dict, self.initial_dict)
+        # Equality with another instance
+        equal_dict = deepcopy(self.initial_dict)
+        other_immutable_dict = ImmutableDict(equal_dict)
+        self.assertEqual(other_immutable_dict, self.immutable_dict)
+
 
 class TestImmutableList(TestCase):
 
@@ -69,6 +79,15 @@ class TestImmutableList(TestCase):
     def test_getitem(self) -> None:
         self.assertEqual(self.immutable_list[0], self.initial_list[0])
 
+    def test_equality(self):
+        # List
+        self.assertEqual(self.initial_list, self.immutable_list)
+        # Tuple
+        self.assertEqual(tuple(self.initial_list), self.immutable_list)
+        # Same class
+        self.assertEqual(ImmutableList(self.initial_list.copy()),
+                         self.immutable_list)
+
 
 class TestImmutableNestedList(TestCase):

```
</details>

kostaspap:
> Unfortunately I discovered some bugs that I think we need to address before merging this PR.
> 
> 1. If the value of a top-level key is a list, an exception is raised, since we will try to pass additional keyworded arguments to `ImmutableList` that do not match its constructor signature.
> 2. If a `DataModel` is passed to `PantherEvent`, the `jsonpath-ng` search will return the **original nested object which can then be mutated**, in which case the mutation will be present in the original event as well. We should ensure that we always return immutable objects, for all access methods.
> 3. `Mapping` does implement the equality check magic method, but `Sequence` does not. This may be a source of confusion in some cases, e.g. when someone is comparing a tuple or list with an `ImmutableList` instance, or two `ImmutableDict` instances containing a nested list internally.
> 4. `ImmutableDict` and `PantherEvent` differ only by the `udm` method, which allows searching by path in the event. Since the paths in the `DataModel` are specified to begin from the top level, returning a `PantherEvent` instance for a nested object and executing the `udm` method on it, would yield incorrect results. It seems more straightforward and simpler to always return `ImmutableDict`.
> 
> The above have been addressed, along with unit tests, in the following patch (and will be committed on Monday, unless there are any objections / further suggestions). Note that the implementation is simplified due to the solution of (4):
> 
> Diff

Sounds good, I ran into #1 while re-testing E2E. You can go ahead and push this patch in the PR.

georgepsarakis:
Thanks for checking @kostaspap, issues addressed in https://github.com/panther-labs/panther/pull/2359/commits/b801e95f60f250428c73791c2f3cc645cc5a0a13.

kostaspap:
Did another E2E test: 

Test updating nested field: 
Screen Shot 2021-01-11 at 12 17 33 PM
Result: 
Screen Shot 2021-01-11 at 12 16 51 PM

Test updating top-level field: 
Screen Shot 2021-01-11 at 12 18 01 PM
Screen Shot 2021-01-11 at 12 17 00 PM

internal/log_analysis/rules_engine/src/enriched_event.py
```diff
@@ -62,6 +54,6 @@ def udm(self, key: str) -> Any:
             # we are dealing with method
             method = self.data_model.methods.get(key)
             if callable(method):
-                return method(self._data)
+                return self._ensure_immutable(method(self._container))
```
kostaspap, line 57: @georgepsarakis This could also be `return method(self._ensure_immutableself._container)` correct? 
I am thinking that `method` could be arbitrary code, that could end up modifying the dictionary.

georgepsarakis: Good catch! Let me check it and get back to you shortly.

georgepsarakis: You are right, mutation in a user-defined method is possible. Unfortunately we cannot use custom classes, because [jsonpath-ng does its own type detection](https://github.com/h2non/jsonpath-ng/blob/de4df602945b72014b4d94dbbfc947ea997fee77/jsonpath_ng/jsonpath.py#L359) and requires `list` and `dict` instances to work correctly. I think we have no choice, but to provide a deep copy of the event, in case a method is used.

georgepsarakis: Test case and fix provided in dc2e851621a5fe97e542e1edea3754d50aa8fe82.

Note that `jsonpath-ng` also provides a separate [method for updating nodes](https://github.com/h2non/jsonpath-ng/blob/de4df602945b72014b4d94dbbfc947ea997fee77/jsonpath_ng/jsonpath.py#L523). I don't think you can achieve mutation using `.find` though.

kostaspap: > You are right, mutation in a user-defined method is possible. Unfortunately we cannot use custom classes, because [jsonpath-ng does its own type detection](https://github.com/h2non/jsonpath-ng/blob/de4df602945b72014b4d94dbbfc947ea997fee77/jsonpath_ng/jsonpath.py#L359) and requires `list` and `dict` instances to work correctly. I think we have no choice, but to provide a deep copy of the event, in case a method is used.

I don't think this should be an issue here though... The `method` is basically a Python method that will be invoked when a user tries to access a specific UDM field. The `jsonpath-ng.find` method doesn't seem to be involved here.

kostaspap: I'm thinking to apply the following patch: 

```
index 026daaa8..ef00bb83 100644
--- a/internal/log_analysis/rules_engine/src/enriched_event.py
+++ b/internal/log_analysis/rules_engine/src/enriched_event.py
@@ -15,7 +15,6 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 from collections.abc import Mapping
-from copy import deepcopy
 from typing import Any, Optional
 
 from .data_model import DataModel
@@ -55,6 +54,6 @@ class PantherEvent(ImmutableDict):  # pylint: disable=R0901
             # we are dealing with method
             method = self.data_model.methods.get(key)
             if callable(method):
-                return self._ensure_immutable(method(deepcopy(self._container)))
+                return self._ensure_immutable(method(self._ensure_immutable(self._container)))
         # no matches, return None by default
         return None
diff --git a/internal/log_analysis/rules_engine/tests/test_enriched_event.py b/internal/log_analysis/rules_engine/tests/test_enriched_event.py
index 982e5e0e..b12e86e2 100644
--- a/internal/log_analysis/rules_engine/tests/test_enriched_event.py
+++ b/internal/log_analysis/rules_engine/tests/test_enriched_event.py
@@ -184,7 +184,8 @@ class TestEnrichedEvent(TestCase):
             }
         )
         enriched_event = PantherEvent(event, data_model)
-        self.assertTupleEqual(enriched_event.udm('source_ip'), (None, 11))
+        with self.assertRaises(TypeError):
+            enriched_event.udm('source_ip')
         self.assertEqual(event_copy, event)
 
     def test_assignment_not_allowed_on_getitem_access(self) -> None:
```

PR #2361, opened 2021-01-07
Cloud Security - Sources API & Snapshot Poller - Filtering and Blacklisting

wey-chiang:
## Background

Users have no mechanism to disable a CloudSecurity source integration forcing them to either a) deal with a potentially overwhelming number of errors that provide no meaningful context or b) to delete the CloudSecurity source altogether.

These changes provide BE support for users to be able to disable a CloudSecurity source integration.

Furthermore, these changes also provide support for the following filtering and blacklisting:

- Regional Blacklisting for all individual pollers apart of the snapshot-poller
- Resource Type Filtering for all individual pollers apart of the snapshot-poller
- Regular Expression Filtering for all individual pollers apart of the snapshot-poller that use resource arns in api calls
  - Support added individually for pollers that do not use a fully-qualified resource arn in api calls

## Changes

## Testing

- `mage test:ci` 
  - Note: this raised issues around web files and some `E0000` for some of the deployment `.yml` files
- Manual deployment to `us-east-1`
- Manual testing using SQS and CloudWatch

---

### Disabled Source Integration
cloudsec-disabled-source

### S3 Regex Filter
cloudsec-filter-s3-regex

### S3 Bucket Filtering
cloudsec-s3-buckets

### S3 Bucket Regex Filtering - Validation
cloudsec-filter-s3-regex-resources-confirmation

### Regional Blacklisting 
cloudsec-filtering-blacklisting-test-results


api/lambda/source/models/api.go
```diff
@@ -83,6 +83,10 @@ type PutIntegrationSettings struct {
 	ScanIntervalMins   int              `json:"scanIntervalMins" validate:"omitempty,oneof=60 180 360 720 1440"`
 	S3Bucket           string           `json:"s3Bucket"`
 	S3PrefixLogTypes   S3PrefixLogtypes `json:"s3PrefixLogTypes,omitempty" validate:"omitempty,min=1"`
+	SourceEnabled      *bool            `json:"sourceEnabled" validate:"omitempty"`
+	RegionBlacklist    []string         `json:"regionBlacklist" validate:"omitempty"`
+	ResourceTypeFilter []string         `json:"resourceTypeFilter" validate:"omitempty"`
+	ARNRegexFilter     string           `json:"arnRegexFilter" validate:"omitempty"`
```
nhakmiller, line 89: I see you've added these to `PutIntegrationSettings`, `UpdateIntegrationSettings`, and `SourceIntegrationMetadata`. Did you update the `PutIntegrationSettings` and `UpdateIntegrationSettings` API endpoints? Not needed right now, but will be needed for eventual frontend support.

nhakmiller (commented):
Thanks for the quick turn around time on this! A few comments (using the suggestions feature where I could!)

deployments/cloud_security.yml
```diff
@@ -1325,6 +1325,12 @@ Resources:
             - Effect: Allow
               Action: lambda:InvokeFunction
               Resource: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-resources-api
+        - Id: InvokeSourceAPI
+          Version: 2012-10-17
+          Statement:
+            - Effect: Allow
+              Action: lambda:InvokeFunction
+              Resource: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-source-api
```
austinbyers, line 1333: Combine this with the resources-api invocation permission above rather than creating a new IAM policy

internal/compliance/snapshot_poller/pollers/aws/configservice.go
```diff
@@ -150,6 +151,18 @@ func buildConfigServiceSnapshot(
 		RecordingGroup: recorder.RecordingGroup,
 		RoleARN:        recorder.RoleARN,
 	}
+	// Check if ResourceID matches the integration's regex filter
```
austinbyers, line 154: For suppressions, we only use glob-style wildcarding rather than full regexes. It seemed to be more intuitive, both for developers and customers.

Consider using glob matching, e.g. `my-dev-bucket-*` rather than a regex `my\-dev\-bucket\-.*` to be consistent with suppressions

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -290,6 +343,16 @@ func singleResourceScan(
 	var resource interface{}
 	var err error
 
+	// Check if resource type is filtered
+	if scanRequest.ResourceType != nil && pollerInput.ResourceTypeFilter != nil {
```
austinbyers, line 347: The name `ResourceTypeFilter` doesn't make it clear whether you're filtering in or out. Is this the set of resources which are _ignored_? In which case maybe `ResourceTypeIgnore` or similar

internal/compliance/snapshot_poller/pollers/utils/source_integration.go
```diff
@@ -0,0 +1,104 @@
...
+	integrationCache                       = make(map[string]cachedIntegration)
+	sess                                   = session.Must(session.NewSession())
+	lambdaClient     lambdaiface.LambdaAPI = lambda.New(sess)
+)
+
+// TODO: If this takes a huge performance hit, update source-api to allow getting a single integration
+func GetIntegration(integrationID string) (integration *models.SourceIntegration, err error) {
+	// Return the cached short-lived integration
+	if cachedIntegration, exists := integrationCache[integrationID]; exists {
+		if time.Now().Add(-1 * time.Minute).Before(cachedIntegration.Timestamp) {
```
austinbyers, line 53: The cache time should be a constant at the top of the file for easy identification

api/lambda/source/models/api.go
```diff
@@ -83,6 +83,10 @@ type PutIntegrationSettings struct {
 	ScanIntervalMins   int              `json:"scanIntervalMins" validate:"omitempty,oneof=60 180 360 720 1440"`
 	S3Bucket           string           `json:"s3Bucket"`
 	S3PrefixLogTypes   S3PrefixLogtypes `json:"s3PrefixLogTypes,omitempty" validate:"omitempty,min=1"`
+	SourceEnabled      *bool            `json:"sourceEnabled" validate:"omitempty"`
+	RegionBlacklist    []string         `json:"regionBlacklist" validate:"omitempty"`
```
austinbyers, line 87: Let's use "Denylist" instead of "Blacklist" in the spirit of inclusivity. Similarly, "Allowlist" instead of "Whitelist"

See, for example: https://gitlab.com/gitlab-org/gitlab/-/issues/7554

> While the etymology of the words isn't steeped in racial undertones or history like other problematic tech word choices, there is no harm in changing them to synonyms that as just as understandable and avoid the question entirely

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -173,6 +174,37 @@ func Poll(scanRequest *pollermodels.ScanEntry) (
 		NextPageToken: scanRequest.NextPageToken,
 	}
 
+	// Retrieve source integration from a short-lived cache for integration filtering options
+	var sourceIntegration *models.SourceIntegration
+	if scanRequest.IntegrationID != nil {
```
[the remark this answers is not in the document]
wey-chiang: Changed it to check if it is nil and to log an error and return early.

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -317,6 +380,16 @@ func singleResourceScan(
...
+		// Check if ResourceID matches the integration's regex filter
+		matched, matchErr := utils.MatchRegexFilter(pollerInput.ARNRegexFilter, *scanRequest.ResourceID)
+		if matched {
+			zap.L().Info("resource filtered based on filter regex",
+				zap.String("regex filter", *pollerInput.ARNRegexFilter), zap.String("resource id", *scanRequest.ResourceID))
+			return nil, nil
+		}
+		if matchErr != nil {
+			return nil, matchErr
+		}
```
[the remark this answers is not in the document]
wey-chiang: Solved this in a number of places

internal/compliance/snapshot_poller/pollers/aws/s3_bucket.go
```diff
@@ -65,6 +65,16 @@ func PollS3Bucket(
...
+	// Check if ResourceID matches the integration's regex filter
+	matched, err := utils.MatchRegexFilter(pollerResourceInput.ARNRegexFilter, resourceARN.Resource)
+	if matched {
+		zap.L().Info("resource filtered based on filter regex",
+			zap.String("regex filter", *pollerResourceInput.ARNRegexFilter), zap.String("resource atn", resourceARN.Resource))
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
wey-chiang: We still need it in the `PollS3Buckets` logic but yeah that makes sense

api/lambda/source/models/api.go
wey-chiang: Looks like custom resources needs to be updated if we want to support these options there

internal/compliance/snapshot_poller/pollers/aws/poll.go
wey-chiang: I'll rename to `ResourceTypeBlacklist` to be consistent with `RegionBlacklist`

internal/compliance/snapshot_poller/pollers/utils/source_integration.go
wey-chiang: Added as a const

api/lambda/source/models/api.go
wey-chiang: 💯

internal/core/custom_resources/resources/self_registration.go
```diff
@@ -192,20 +200,24 @@ func putCloudSecurityIntegration(accountID string) error {
...
+				AWSAccountID:          properties.AccountID,
+				IntegrationLabel:      genLogProcessingLabel(),
+				IntegrationType:       models.IntegrationTypeAWS3,
+				UserID:                systemUserID,
+				S3Bucket:              properties.AuditLogsBucket,
+				S3PrefixLogTypes:      prefixLogTypes,
+				SourceEnabled:         properties.SourceEnabled,
+				RegionDenylist:        properties.RegionDenylist,
+				ResourceTypeDenylist:  properties.ResourceTypeDenylist,
+				ResourceRegexDenylist: properties.ResourceRegexDenylist,
```
nhakmiller, line 220: These fields don't mean anything for log processing sources today, lets not put them in for now.

internal/core/custom_resources/resources/self_registration.go
```diff
@@ -217,18 +229,22 @@ func putLogProcessingIntegration(accountID, auditBucket string, prefixLogTypes m
...
-			S3Bucket:         source.S3Bucket,
-			S3PrefixLogTypes: prefixLogTypes,
+			IntegrationID:         source.IntegrationID,
+			IntegrationLabel:      source.IntegrationLabel,
+			S3Bucket:              source.S3Bucket,
+			S3PrefixLogTypes:      prefixLogTypes,
+			SourceEnabled:         source.SourceEnabled,
+			RegionDenylist:        source.RegionDenylist,
+			ResourceTypeDenylist:  source.ResourceTypeDenylist,
+			ResourceRegexDenylist: source.ResourceRegexDenylist,
```
nhakmiller, line 247: Same comment as above, these new fields only have meaning for scanning/cloud security sources so lets only reference them there. It's kind of weird because we use the same model for all source integrations, but we only populate certain fields based on whats actually used by that integration.

internal/core/source_api/api/put_integration.go
```diff
@@ -257,11 +257,14 @@ func (api API) FullScan(input *models.FullScanInput) error {
...
-		IntegrationLabel: input.IntegrationLabel,
-		IntegrationType:  input.IntegrationType,
+		CreatedAtTime:         time.Now(),
+		CreatedBy:             input.UserID,
+		IntegrationID:         uuid.New().String(),
+		IntegrationLabel:      input.IntegrationLabel,
+		IntegrationType:       input.IntegrationType,
+		RegionDenylist:        input.RegionDenylist,
+		ResourceTypeDenylist:  input.ResourceTypeDenylist,
+		ResourceRegexDenylist: input.ResourceRegexDenylist,
```
nhakmiller, line 267: I would move these down below into the `case models.IntegrationTypeAWSScan` block for the same reasons discussed above.

internal/core/source_api/api/update_integration.go
```diff
@@ -163,6 +163,10 @@ func (api API) validateUniqueConstraints(existingIntegrationItem *ddb.Integratio
 }
 
 func updateIntegrationDBItem(item *ddb.Integration, input *models.UpdateIntegrationSettingsInput) {
+	item.SourceEnabled = input.SourceEnabled
+	item.RegionDenylist = input.RegionDenylist
+	item.ResourceTypeDenylist = input.ResourceTypeDenylist
+	item.ResourceRegexDenylist = input.ResourceRegexDenylist
```
nhakmiller, line 169: Move to `case models.IntegrationTypeAWSScan:` block.

internal/core/source_api/api/utils.go
```diff
@@ -77,6 +77,10 @@ func itemToIntegration(item *ddb.Integration) *models.SourceIntegration {
 	integration.CreatedAtTime = item.CreatedAtTime
 	integration.CreatedBy = item.CreatedBy
 	integration.LastEventReceived = item.LastEventReceived
+	integration.SourceEnabled = item.SourceEnabled
+	integration.RegionDenylist = item.RegionDenylist
+	integration.ResourceTypeDenylist = item.ResourceTypeDenylist
+	integration.ResourceRegexDenylist = item.ResourceRegexDenylist
```
nhakmiller, line 83: Move to `case models.IntegrationTypeAWSScan:` block.

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -226,6 +226,15 @@ func getClient(pollerInput *awsmodels.ResourcePollerInput,
...
 	service string, region string) (interface{}, error) {
 
+	// Check if provided region is blacklisted
+	for _, blacklistedRegion := range pollerInput.RegionDenylist {
+		if region == blacklistedRegion {
+			zap.L().Info("matched blacklisted region - skipping scan",
+				zap.String("region", region))
+			return nil, nil
+		}
+	}
```
nhakmiller, line 236: Don't know why I missed this on the last pass, but I don't think this works. I believe this will cause a panic. `getClient` is called in the individual resource pollers (and the group ones, but those ones already get region filtered so it's a non-issue) and if no error is returned then we assert that the client is of the type we expect. I believe assert `nil` is some interface will panic. See this example usage:

```
	client, err := getClient(pollerResourceInput, ConfigServiceClientFunc, "configservice", region)
	if err != nil {
		return nil, err
	}
	return client.(configserviceiface.ConfigServiceAPI), nil
```

In order for this to work, you would need to update each `getXYZServiceClient` function to know how to handle a `nil, nil` response, and you would have to update each `PollXYZService` function to know how to handle not getting a client back. 

If I'm correct, you can get this behavior to happen by creating a source with a region blacklist and then initiating a single resource scan for a resource in that region.

The two best options in my opinion are:

- Have the `getClient` return a custom error instead of `nil, nil`, let the `getXYZServiceClient` pass that error up (no changes), let each `PollXYZService` pass that error up (no changes), then catch that error in `singleResourceScan` in `poll.go`. We already handle rate limit errors as a special case, this logic would be a cousin to the rate limit logic. Something like `if err == customPantherScannerErrorRegionDenied { zap.L().Info("Skipping denied region in single resource scan"); return nil, nil }`.
- Go with our original plan and add a new helper function that determines the region of the resource based on resource ID & type, and applies region deny listing before entering the `PollXYZService` functions

Of the two, the first seems easier & more robust.

nhakmiller (commented):
A couple nitpicks, and one big issue I missed from last time somehow. Luckily I think there is a fairly straightforward fix that is still aligned with your idea on region filtering

nhakmiller (approved):
LGTM! I think this is actually a really clean solution for the region & resource filtering, nice job.

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -173,6 +174,38 @@ func Poll(scanRequest *pollermodels.ScanEntry) (
 		NextPageToken: scanRequest.NextPageToken,
 	}
 
+	// Retrieve source integration from a short-lived cache for integration filtering options
+	var sourceIntegration *models.SourceIntegration
+	if scanRequest.IntegrationID == nil {
+		zap.L().Error("got a malformed scan request - no integration id", zap.Any("scan request", scanRequest))
```
kostaspap, line 180: Why log error here? Why not return error?

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -173,6 +174,38 @@ func Poll(scanRequest *pollermodels.ScanEntry) (
...
+		return nil, nil
+	}
+	sourceIntegration, err = utils.GetIntegration(*scanRequest.IntegrationID)
+	if err != nil {
+		zap.L().Warn("failed to retrieve integration from source api")
```
kostaspap, line 185: Why log a warning? Why not return an error?

api/lambda/source/models/api.go
```diff
@@ -99,14 +103,18 @@ type ListIntegrationsInput struct {
...
+	IntegrationLabel      string           `json:"integrationLabel" validate:"required,integrationLabel,excludesall='<>&\""`
+	CWEEnabled            *bool            `json:"cweEnabled"`
+	RemediationEnabled    *bool            `json:"remediationEnabled"`
+	ScanIntervalMins      int              `json:"scanIntervalMins" validate:"omitempty,oneof=60 180 360 720 1440"`
+	S3Bucket              string           `json:"s3Bucket" validate:"omitempty,min=1"`
+	S3PrefixLogTypes      S3PrefixLogtypes `json:"s3PrefixLogTypes,omitempty" validate:"omitempty,min=1"`
+	SourceEnabled         *bool            `json:"sourceEnabled" validate:"omitempty"`
+	RegionDenylist        []string         `json:"regionDenylist" validate:"omitempty"`
+	ResourceTypeDenylist  []string         `json:"resourceTypeDenylist" validate:"omitempty"`
+	ResourceRegexDenylist []string         `json:"resourceRegexDenylist" validate:"omitempty"`
```
kostaspap, line 116: What is the regex supposed to match against? ARNs? Names?...?

internal/core/source_api/ddb/items.go
```diff
@@ -53,6 +53,12 @@ type Integration struct {
 	StackName         string   `json:"stackName,omitempty"`
 	LogProcessingRole string   `json:"logProcessingRole,omitempty"`
 
+	// fields for configurable cloud security sources
+	SourceEnabled         *bool    `json:"sourceEnabled" validate:"omitempty"`
```
kostaspap, line 57: Why is this a `*bool` and not `bool`?

internal/compliance/snapshot_poller/pollers/utils/source_integration.go
```diff
@@ -0,0 +1,118 @@
...
+		if glob == "" {
+			continue
+		}
+		// First,  escape any regex special characters
+		escaped := regexp.QuoteMeta(glob)
+
+		// Wildcards in the original pattern are now escaped literals - convert back
+		// NOTE: currently no way for user to specify a glob that would match a literal '*'
+		regex := "^" + strings.ReplaceAll(escaped, `\*`, `.*`) + "$"
+		matcher, err := regexp.Compile(regex)
```
kostaspap, line 103: This method should be removed from here. `utils` is a bad name for a package in Go. In Go, packages are namespaces. 

1. This check should be removed. Instead, you should create a custom validator for the properties inside `ResourcePollerInput.ResourceRegexDenylist. See an example here of custom validators https://github.com/panther-labs/panther/blob/master/api/lambda/source/models/validator.go. 
2. This method should better be exposed as a function of `ResourcePollerInput`. See relevant comment above.

api/lambda/source/models/integration.go
```diff
@@ -67,6 +67,12 @@ type SourceIntegrationMetadata struct {
 	KmsKey            string           `json:"kmsKey,omitempty"`
 	LogProcessingRole string           `json:"logProcessingRole,omitempty"`
 
+	// optional fields for snapshot-poller filtering
+	SourceEnabled         *bool    `json:"sourceEnabled,omitempty"`
```
kostaspap, line 71: 1. `Source` prefix is redundant, just `Enabled` should be enough. 
2. It seems that this should be just `bool`
3. Is the idea here that users can enable/disable specific sources? Do we really need this, can't the users just delete the integration?

internal/compliance/snapshot_poller/pollers/aws/guardduty_detector.go
```diff
@@ -141,6 +146,17 @@ func buildGuardDutyDetectorSnapshot(guardDutySvc guarddutyiface.GuardDutyAPI, de
 	if err != nil {
 		return nil, err
 	}
+
+	// Check if ResourceID matches the integration's regex filter
+	if pollerInput != nil {
+		matched, err := utils.MatchRegexDenylist(pollerInput.ResourceRegexDenylist, *detectorID)
```
kostaspap, line 152: In other places the `ResourceRegexDenylist` is matched against ARNs. Here it is matched against a detectorID. Is the detectorID an arn?

kostaspap (changes requested):
The bulk of the logic is added in the snapshot-pollers package. However, it is the `aws_event_processor` that sends notifications to the snapshot pollers to do the polling. 
Shouldn't the ignore logic be in the `aws_event_processor` package? @nhakmiller

nhakmiller:
> The bulk of the logic is added in the snapshot-pollers package. However, it is the `aws_event_processor` that sends notifications to the snapshot pollers to do the polling.
> Shouldn't the ignore logic be in the `aws_event_processor` package? @nhakmiller

If we did we would just have to duplicate it here to handle daily scans. It's best to have the logic here so it is centralized for all types of scans.

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -226,6 +234,15 @@ func getClient(pollerInput *awsmodels.ResourcePollerInput,
...
 	service string, region string) (interface{}, error) {
 
+	// Check if provided region is in the denylist
+	for _, deniedRegion := range pollerInput.RegionDenylist {
+		if region == deniedRegion {
+			return nil, &RegionDenylistError{
+				Err: errors.New("requested region was in region denylist"),
+			}
+		}
+	}
```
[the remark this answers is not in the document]
nhakmiller: There are two ways to do scans:

- Individual resource scans
- Region wide resource type scans

Region wide scans we can filter before we get here, because we know what region is being requested to be scanned.

For individual resource scans we need this because different resource types sometimes have custom logic to determine what region they are in (IAM resources, s3 resources, guard duty/config resources).

internal/compliance/snapshot_scheduler/scheduler/schedule.go
```diff
@@ -71,8 +71,8 @@ func PollAndIssueNewScans() error {
 	)
 }
 
-// getEnabledIntegrations lists enabled integrations from the snapshot-api.
-func getEnabledIntegrations() (integrations []*models.SourceIntegration, err error) {
+// GetEnabledIntegrations lists enabled integrations from the snapshot-api.
+func GetEnabledIntegrations() (integrations []*models.SourceIntegration, err error) {
```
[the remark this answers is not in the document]
nhakmiller: So we can import it & use it in the `panther-snapshot-poller`. This is the new way we're doing APIs, instead of invoking other lambda functions we just import them and run the code.

api/lambda/source/models/api.go
nhakmiller: Panther resource IDs. If you think  it makes it more clear it could maybe be `ResourceIDRegexDenyList`.

internal/compliance/snapshot_poller/pollers/utils/source_integration.go
nhakmiller: I think a customer validator is overkill for this one thing. Maybe the snapshot-poller needs a validator in general (since right now we're just validating all the inputs by checking them in `if` statements) but I don't think that's in scope for this PR. If you'd like, I can open an issue and assign it to myself for this task.

api/lambda/source/models/integration.go
nhakmiller: Re point 3: we've specifically had requests from users for being able to disable problematic sources without deleting them. Sure they could just delete the source & re-add when they're ready to start scanning again, but this has two downsides:

1. They have to step through the source onboarding wizard again (including looking up the account ID since most people don't memorize  that)
2. When you delete a source, all the resources from that source show up with "unable to find source". By disabling the source it means the resources would still be there in Panther just not being scanned

In general I agree with you it wouldn't be needed, but since we're already in here adding this much more complex logic adding this relatively simple part is a nice quality of life improvement for customers.

internal/compliance/snapshot_poller/pollers/aws/guardduty_detector.go
nhakmiller: Ah good catch, @wey-chiang you should compare to `detectorSnapshot.ID` instead of `detectorID`, as the `detectorSnapshot.ID` is the panther resource ID (which includes the type & region) and is what users would see in the UI under ID.

internal/compliance/snapshot_poller/pollers/aws/poll.go
nhakmiller: Returned errors get retried, if the `IntegrationID` in the request is `nil` we cannot ever succeed at the scan so retrying will just cause more errors.

api/lambda/source/models/integration.go
kostaspap: Thanks @nhakmiller , makes sense.

internal/compliance/snapshot_poller/pollers/aws/poll.go
nhakmiller: Yeah I think returning an error here is probably better, most likely case this would happen:

1. Transient network error meaning communication with the `source-api` is bad
2. The user deleted the source

The second case might be a bit annoying to see a bunch of errors for, but we would also see a bunch of errors if we tried to scan and they had deleted the audit role (and presumably if they delete the source they would prefer not  to have a bunch of inflight messages continue scanning).

internal/compliance/snapshot_poller/pollers/utils/source_integration.go
kostaspap: > I think a customer validator is overkill for this one thing. Maybe the snapshot-poller needs a validator in general (since right now we're just validating all the inputs by checking them in `if` statements) but I don't think that's in scope for this PR. If you'd like, I can open an issue and assign it to myself for this task.

Yeah, I see why maybe that's a bit overkill. What about adding a `ResourcePollerInput.CompileRegex() error` method that tries to compile the provided regexes and populates and internal slice of the compiled regexes? This could be invoked towards the entry point of the code. 
That way, the code that tries to match the regexes against the ARNs/IDs doesn't have to deal with error handling - it will just used the already compiled regexes.

kostaspap:
> > The bulk of the logic is added in the snapshot-pollers package. However, it is the `aws_event_processor` that sends notifications to the snapshot pollers to do the polling.
> > Shouldn't the ignore logic be in the `aws_event_processor` package? @nhakmiller
> 
> If we did we would just have to duplicate it here to handle daily scans. It's best to have the logic here so it is centralized for all types of scans.

Ah right, yeah, makes sense.

internal/compliance/snapshot_poller/pollers/utils/source_integration.go
```diff
@@ -0,0 +1,118 @@
...
+		if integrationID == integration.IntegrationID {
+			result = integration
+		}
+	}
+
+	if result == nil {
+		zap.L().Warn("target integration id not found", zap.String("target integration id", integrationID))
+	}
+	return result, nil
+}
```
kostaspap, line 90: So the snapshot-poller is invoked by `aws-event-processor` and by `snapshot-scheduler`. Both of those Lambdas reach to the sources-api to retrieve SourceIntegrationMetadata. 
Seems they could be passing the blacklist information to the `snapshot-poller` payload, instead of the snapshot poller having to query the sources-api by itself again? I think it could simplify the code a bit (this method + the cache could be removed) and will also reduce the overall calls we make to the sources-api.

internal/compliance/snapshot_poller/pollers/utils/source_integration.go
nhakmiller: Yeah I think that's good, also when scanning several hundred resources just compiling the regex one time then trying to match it 100 times is probably much preferred than compiling & matching 100 times.

internal/compliance/snapshot_poller/pollers/utils/source_integration.go
nhakmiller: Yeah fair enough, the `aws-event-processor` and `snapshot-scheduler` can pass the ignore list information along. Then we also want to change our self re-queue method to re-queue the ignore list information as well.

nhakmiller:
@kostaspap thanks for the thorough review! We were hoping to have this in for 1.15 but seems we just missed it, I'm hoping we can do a `sync-to-release` ideally today so how would you say about this, we:

1. Address the struct formatting nitpicks (about putting similar fields next to each other)
2. Address the regex compile one time in a function thing
3. Update so that the blacklist information is passed through by the snapshot-scheduler & aws-event-processor functions instead of being looked up by the snapshot-poller (since both these functions already make the list-integrations API call)
4. Add some basic unit tests around blacklisting

If these four things were done today, would you be ok with approving this change? @wey-chiang, I think 1, 2 and 4 should be pretty straightforward and number 3 should be accomplishable today unless there are any unforeseen issues. Let me know if you run into any issues.

kostaspap:
> @kostaspap thanks for the thorough review! We were hoping to have this in for 1.15 but seems we just missed it, I'm hoping we can do a `sync-to-release` ideally today so how would you say about this, we:
> 
> 1. Address the struct formatting nitpicks (about putting similar fields next to each other)
> 2. Address the regex compile one time in a function thing
> 3. Update so that the blacklist information is passed through by the snapshot-scheduler & aws-event-processor functions instead of being looked up by the snapshot-poller (since both these functions already make the list-integrations API call)
> 4. Add some basic unit tests around blacklisting
> 
> If these four things were done today, would you be ok with approving this change? @wey-chiang, I think 1, 2 and 4 should be pretty straightforward and number 3 should be accomplishable today unless there are any unforeseen issues. Let me know if you run into any issues.

I see, yeah that's sounds good! I will give a pre-approval now so you can address and merge on your own time. #4 is not super urgent if this has been tested E2E already, just good to have in general so don't block the merge just because of it, I imagine it can be addressed later.

kostaspap (approved):

nhakmiller:
> I see, yeah that's sounds good! I will give a pre-approval now so you can address and merge on your own time. #4 is not super urgent if this has been tested E2E already, just good to have in general so don't block the merge just because of it, I imagine it can be addressed later.

Thanks! We've done some E2E testing, but a think at least one round of basic unit tests will help keep things steady (and give us a base to grow more unit tests onto later if need be).

api/lambda/source/models/api.go
```diff
@@ -74,16 +74,20 @@ type PutIntegrationInput struct {
...
+	AWSAccountID          string           `genericapi:"redact" json:"awsAccountId" validate:"omitempty,len=12,numeric"`
+	CWEEnabled            *bool            `json:"cweEnabled"`
+	RemediationEnabled    *bool            `json:"remediationEnabled"`
+	ScanIntervalMins      int              `json:"scanIntervalMins" validate:"omitempty,oneof=60 180 360 720 1440"`
+	S3Bucket              string           `json:"s3Bucket"`
+	S3PrefixLogTypes      S3PrefixLogtypes `json:"s3PrefixLogTypes,omitempty" validate:"omitempty,min=1"`
+	SourceEnabled         *bool            `json:"sourceEnabled" validate:"omitempty"`
+	RegionDenylist        []string         `json:"regionDenylist" validate:"omitempty"`
+	ResourceTypeDenylist  []string         `json:"resourceTypeDenylist" validate:"omitempty"`
+	ResourceRegexDenylist []string         `json:"resourceRegexDenylist" validate:"omitempty"`
```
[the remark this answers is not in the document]
wey-chiang: Originally, I was using Blacklist which conveyed the intended meaning -- over the course of this PR that has changed; in light of the naming convention we currently use, I think `*Ignore*` makes way more sense, intuitively.

internal/compliance/snapshot_poller/pollers/aws/clients.go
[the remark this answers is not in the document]
wey-chiang: There is also the case of non-arn based scans, a good example of this is the `s3_bucket.go`

There, because we have to retrieve the bucket region, we do the region filtering in the for loop iterating through all the buckets for the `PollS3Buckets`

internal/compliance/snapshot_poller/pollers/aws/configservice.go
```diff
@@ -150,6 +151,16 @@ func buildConfigServiceSnapshot(
...
+	if pollerInput != nil {
+		matched, err := utils.MatchRegexDenylist(pollerInput.ResourceRegexDenylist, *recorder.Name)
+		if err != nil {
+			return nil, err
+		}
+		if matched {
+			return nil, nil
+		}
```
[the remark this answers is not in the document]
wey-chiang: Added a check to both the `PollConfigService` and `PollConfigServices` to handle the `nil` case

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -243,6 +248,17 @@ func buildEksClusterSnapshot(eksSvc eksiface.EKSAPI, clusterName *string) (*awsm
...
+	// Check if ResourceID matches the integration's regex filter
+	if pollerInput != nil {
+		matched, err := utils.MatchRegexDenylist(pollerInput.ResourceRegexDenylist, *details.Arn)
+		if err != nil {
+			return nil, err
+		}
+		if matched {
+			return nil, nil
+		}
+	}
```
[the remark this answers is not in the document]
wey-chiang: @kostaspap 
I'm confused how the `ShouldIgnoreArn` function proposed is any different from the `MatchRegexIgnoreList`

internal/compliance/snapshot_poller/pollers/aws/poll.go
wey-chiang: Will add a comment for clarity

internal/core/source_api/ddb/items.go
wey-chiang: How so? We want to check to ensure that `SourceEnabled` is nil if the user has not specified any option (either enable or disable)
Since `bool` will default to `false`, unless we were to include the enable source option in the source onboarding, this would automatically make all newly created sources as disabled and anytime the integration is updated without this specified, it will also be disabled.

api/lambda/source/models/integration.go
wey-chiang: Renamed `SourceEnabled` to `Enabled`

I am still unsure as to the difference between `bool` and `*bool` here

internal/compliance/snapshot_poller/models/aws/types.go
```diff
@@ -64,12 +67,63 @@ type GenericAWSResource struct {
...
+			zap.L().Error("invalid regex",
+				zap.String("originalPattern", glob),
+				zap.String("transformedRegex", regex),
+				zap.Error(err),
+			)
+			return err
+		}
+		r.CompiledRegexIgnoreList[i] = compiledGlob
+	}
+	return nil
```
nhakmiller, line 107: Instead of doing this, you can make the list with a capacity of 0 and just append entries:

```suggestion
	r.CompiledRegexIgnoreList = make([]*regexp.Regexp, 0, len(r.ResourceRegexIgnoreList))

	for _, glob := range r.ResourceRegexIgnoreList {
		if glob == "" {
			continue
		}
		// First,  escape any regex special characters
		escaped := regexp.QuoteMeta(glob)

		// Wildcards in the original pattern are now escaped literals - convert back
		// NOTE: currently no way for user to specify a glob that would match a literal '*'
		regex := "^" + strings.ReplaceAll(escaped, `\*`, `.*`) + "$"
		compiledGlob, err := regexp.Compile(regex)
		if err != nil {
			// We are building the regex, so it should always be valid
			zap.L().Error("invalid regex",
				zap.String("originalPattern", glob),
				zap.String("transformedRegex", regex),
				zap.Error(err),
			)
			return err
		}
		r.CompiledRegexIgnoreList = append(r.CompiledRegexIgnoreList, compiledGlob)
	}
	return nil
```

Then you won't have `nil`/empty entries in the list, which simplifies the work a bit down below (separate suggestion since apparently there's a max size on suggestions.

internal/compliance/snapshot_poller/models/aws/types.go
```diff
@@ -64,12 +67,63 @@ type GenericAWSResource struct {
...
+		}
+	}
+	for _, compiledRegex := range r.CompiledRegexIgnoreList {
+		if compiledRegex == nil {
+			continue
+		}
+		if compiledRegex.MatchString(resourceID) {
+			return true, nil
+		}
+	}
```
nhakmiller, line 125: With the above suggestion, you don't have to worry about `nil` entries.

```suggestion
	for _, compiledRegex := range r.CompiledRegexIgnoreList {
		if compiledRegex.MatchString(resourceID) {
			return true, nil
		}
	}
```

internal/compliance/snapshot_poller/pollers/aws/configservice.go
```diff
@@ -150,6 +155,14 @@ func buildConfigServiceSnapshot(
 		RecordingGroup: recorder.RecordingGroup,
 		RoleARN:        recorder.RoleARN,
 	}
+	// Check if ResourceID matches the integration's regex filter
+	if pollerInput != nil {
+		if ignore, err := pollerInput.ShouldIgnoreResource(*recorder.Name); ignore || err != nil {
+			if err != nil || ignore {
+				return nil, err
+			}
+		}
```
nhakmiller, line 164: These nested `if` statements are now redundant (I'm assuming this is refactoring leftovers)
```suggestion
		if ignore, err := pollerInput.ShouldIgnoreResource(*recorder.Name); ignore || err != nil {
			return nil, err
		}
```

internal/compliance/snapshot_poller/pollers/aws/eks_cluster.go
```diff
@@ -243,6 +248,15 @@ func buildEksClusterSnapshot(eksSvc eksiface.EKSAPI, clusterName *string) (*awsm
 		return nil, err
 	}
 
+	// Check if ResourceID matches the integration's regex filter
+	if pollerInput != nil {
+		if ignore, err := pollerInput.ShouldIgnoreResource(*details.Arn); ignore || err != nil {
+			if err != nil || ignore {
+				return nil, err
+			}
+		}
```
nhakmiller, line 257: As above:

```suggestion
		if ignore, err := pollerInput.ShouldIgnoreResource(*details.Arn); ignore || err != nil {
			return nil, err
		}
```

internal/compliance/snapshot_poller/pollers/aws/guardduty_detector.go
```diff
@@ -141,6 +149,15 @@ func buildGuardDutyDetectorSnapshot(guardDutySvc guarddutyiface.GuardDutyAPI, de
...
 		return nil, err
 	}
+
+	// Check if ResourceID matches the integration's regex filter
+	if pollerInput != nil {
+		if ignore, err := pollerInput.ShouldIgnoreResource(*detectorSnapshot.ID); ignore || err != nil {
+			if err != nil || ignore {
+				return nil, err
+			}
+		}
```
nhakmiller, line 159:
```suggestion
		if ignore, err := pollerInput.ShouldIgnoreResource(*detectorSnapshot.ID); ignore || err != nil {
			return nil, err
		}
```

internal/compliance/snapshot_poller/pollers/aws/password_policy.go
```diff
@@ -103,6 +103,12 @@ func PollPasswordPolicy(pollerInput *awsmodels.ResourcePollerInput) ([]apimodels
 		awsmodels.PasswordPolicySchema,
 	)
 
+	// Check if ResourceID matches the integration's regex filter
+	if ignore, err := pollerInput.ShouldIgnoreResource(resourceID); ignore || err != nil {
+		if err != nil || ignore {
+			return nil, nil, err
+		}
+	}
```
nhakmiller, line 111:
```suggestion
	if ignore, err := pollerInput.ShouldIgnoreResource(resourceID); ignore || err != nil {
		return nil, nil, err
	}
```

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -314,9 +348,15 @@ func singleResourceScan(
...
-			// Don't return an error here because the scan request is not retryable
+			// This error is not retryable
 			return nil, nil
 		}
+		// Check if ResourceID matches the integration's regex filter
+		if ignore, err := pollerInput.ShouldIgnoreResource(*scanRequest.ResourceID); ignore || err != nil {
+			if err != nil || ignore {
+				return nil, err
+			}
+		}
```
nhakmiller, line 359:
```suggestion
		if ignore, err := pollerInput.ShouldIgnoreResource(*scanRequest.ResourceID); ignore || err != nil {
			return nil, err
		}
```

internal/compliance/snapshot_poller/pollers/aws/s3_bucket.go
```diff
@@ -355,6 +364,15 @@ func PollS3Buckets(pollerInput *awsmodels.ResourcePollerInput) ([]apimodels.AddR
...
 		}
+		// Check if ResourceID matches the integration's regex filter
+		if ignore, err := pollerInput.ShouldIgnoreResource(*bucket.Name); ignore || err != nil {
+			if err != nil {
+				return nil, nil, err
+			}
+			if ignore {
+				continue
+			}
+		}
```
nhakmiller, line 375:
```suggestion
		if ignore, err := pollerInput.ShouldIgnoreResource(*bucket.Name); ignore || err != nil {
			if err != nil {
				return nil, nil, err
			}
			continue
		}
```

I think we can drop the `if ignore`, we know either `ignore` or `err != nil` is true, and since we didn't enter the above block we know it must be `ignore`.

internal/core/custom_resources/resources/self_registration.go
```diff
@@ -40,11 +40,15 @@ const (
...
-	EnableS3AccessLogs bool   `json:",string"`
+	AccountID               string   `validate:"required,len=12"`
+	AuditLogsBucket         string   `validate:"required"`
+	EnableCloudTrail        bool     `json:",string"`
+	EnableGuardDuty         bool     `json:",string"`
+	EnableS3AccessLogs      bool     `json:",string"`
+	Enabled                 *bool    `json:",string"`
+	RegionIgnoreList        []string `json:","`
+	ResourceTypeIgnoreList  []string `json:","`
+	ResourceRegexIgnoreList []string `json:","`
```
nhakmiller, line 51: This is just for self-registration during the initial setup, I think we can just remove these fields.

internal/core/custom_resources/resources/self_registration.go
```diff
@@ -171,17 +175,21 @@ func stringSliceEqual(left, right []string) bool {
...
+				IntegrationLabel:        cloudSecLabel,
+				IntegrationType:         models.IntegrationTypeAWSScan,
+				ScanIntervalMins:        1440,
+				UserID:                  systemUserID,
+				CWEEnabled:              aws.Bool(true),
+				RemediationEnabled:      aws.Bool(true),
+				Enabled:                 properties.Enabled,
+				RegionIgnoreList:        properties.RegionIgnoreList,
+				ResourceTypeIgnoreList:  properties.ResourceTypeIgnoreList,
+				ResourceRegexIgnoreList: properties.ResourceRegexIgnoreList,
```
nhakmiller, line 192: You can just leave these not set/default values, they won't be specified.

nhakmiller (approved):
Mostly some small comments, overall looks good.

internal/core/source_api/ddb/items.go
kostaspap: Well a source can be either enabled or disabled, there is no 3rd option. `nil` doesn't give any value. You can make the FE set this to true or false depending on what you want to be the default.
Putting a *bool here just makes it more painful to handle this property throughout the code since now we need to be careful not to make nil pointer dereference @wey-chiang

api/lambda/source/models/integration.go
kostaspap: > Renamed `SourceEnabled` to `Enabled`
> 
> I am still unsure as to the difference between `bool` and `*bool` here

@wey-chiang see my comment above.

internal/compliance/snapshot_poller/models/aws/types.go
```diff
@@ -64,12 +67,60 @@ type GenericAWSResource struct {
...
+		// NOTE: currently no way for user to specify a glob that would match a literal '*'
+		regex := "^" + strings.ReplaceAll(escaped, `\*`, `.*`) + "$"
+		compiledGlob, err := regexp.Compile(regex)
+		if err != nil {
+			// We are building the regex, so it should always be valid
+			zap.L().Error("invalid regex",
+				zap.String("originalPattern", glob),
+				zap.String("transformedRegex", regex),
+				zap.Error(err),
+			)
```
kostaspap, line 102: You should't log errors so deep in the call stack, please remove. Let the caller decide if they want to log the error or not.

internal/compliance/snapshot_poller/models/aws/types.go
```diff
@@ -64,12 +67,60 @@ type GenericAWSResource struct {
...
+}
+
+func (r *ResourcePollerInput) ShouldIgnoreResource(resourceID string) (ignore bool, err error) {
+	// Check if the regexs have been compiled
+	if r.CompiledRegexIgnoreList == nil {
+		err := r.CompileRegex()
+		if err != nil {
+			return false, err
+		}
+	}
```
kostaspap, line 117: You can just remove this check. This would greatly simplify the work of the callers of `ShouldIgnoreResource` since they now don't have to handle the error case. 
You could make the call panic if `CompileRegex` hasn't been invoked before `ShouldIgnoreResource` is called.

internal/compliance/snapshot_poller/pollers/aws/configservice.go
```diff
@@ -150,6 +155,12 @@ func buildConfigServiceSnapshot(
 		RecordingGroup: recorder.RecordingGroup,
 		RoleARN:        recorder.RoleARN,
 	}
+	// Check if ResourceID matches the integration's regex filter
+	if pollerInput != nil {
+		if ignore, err := pollerInput.ShouldIgnoreResource(*recorder.Name); ignore || err != nil {
+			return nil, err
+		}
+	}
```
kostaspap, line 163: I made this comment in my previous review: 
If you replace `pollerInput *awsmodels.ResourcePollerInput` with `pollerInput awsmodels.ResourcePollerInput` in the call function and make sure that the pollerInput doesn't throw errors, this could just be : 
```
if pollerInput.ShouldIgnoreResource(*recorder.Name){
 return nil
}
```

internal/compliance/snapshot_poller/pollers/aws/password_policy.go
```diff
@@ -103,6 +103,10 @@ func PollPasswordPolicy(pollerInput *awsmodels.ResourcePollerInput) ([]apimodels
 		awsmodels.PasswordPolicySchema,
 	)
 
+	// Check if ResourceID matches the integration's regex filter
+	if ignore, err := pollerInput.ShouldIgnoreResource(resourceID); ignore || err != nil {
```
kostaspap, line 107: pollerInput could be `nil` and here you will get an exception. I would recommend replacing `PollPasswordPolicy(pollerInput *awsmodels.ResourcePollerInput)` with `PollPasswordPolicy(pollerInput awsmodels.ResourcePollerInput)`. I don't see in the codebase any scenario where the ResourcePollerInput would/should be nil. Also see: https://github.com/golang/go/wiki/CodeReviewComments#pass-values

internal/compliance/snapshot_poller/pollers/aws/configservice.go
kostaspap: Also see: https://github.com/golang/go/wiki/CodeReviewComments#pass-values

internal/compliance/snapshot_poller/models/aws/types.go
```diff
@@ -64,12 +67,60 @@ type GenericAWSResource struct {
...
+	Region                  *string
+	Timestamp               *time.Time
+	NextPageToken           *string
+	RegionIgnoreList        []string
+	ResourceTypeIgnoreList  []string
+	ResourceRegexIgnoreList []string
+	CompiledRegexIgnoreList []*regexp.Regexp
+}
+
+func (r *ResourcePollerInput) CompileRegex() error {
```
kostaspap, line 82: Searching this function in this PR I see that it is never called? Can you verify?

kostaspap:
Thanks @wey-chiang, you addressed most of my comments. I've put some new ones, a few of them are style-related but there is also a potential nil-pointer dereference and another issue where we keep compiling the regexes for every new resource scan.

internal/compliance/snapshot_poller/models/aws/types.go
wey-chiang: It is called in the ShouldIgnoreResource func

internal/compliance/snapshot_poller/pollers/aws/configservice.go
wey-chiang: Removed the check since it should never be nil by the logic in the caller.

The `pollerInput *awsmodels.ResourcePollerInput` is used in a number of other files -- particularly in the `get*Client` functions. 

I am in favor of keeping this the way it is (besides removing the check) to maintain consistency with similar code

internal/compliance/snapshot_poller/pollers/aws/password_policy.go
wey-chiang: See above comment

internal/compliance/snapshot_poller/models/aws/types.go
wey-chiang: I've moved it over to the `poll.go:Poll` function, right after the initialize `pollerResourceInput`

internal/core/source_api/api/put_integration.go
```diff
@@ -281,6 +285,10 @@ func generateNewIntegration(input *models.PutIntegrationInput) *models.SourceInt
 		metadata.S3PrefixLogTypes = input.S3PrefixLogTypes
 		metadata.StackName = getStackName(input.IntegrationType, input.IntegrationLabel)
 		metadata.LogProcessingRole = generateLogProcessingRoleArn(input.AWSAccountID, input.IntegrationLabel)
+		metadata.Enabled = input.Enabled
+		metadata.RegionIgnoreList = input.RegionIgnoreList
+		metadata.ResourceTypeIgnoreList = input.ResourceTypeIgnoreList
+		metadata.ResourceRegexIgnoreList = input.ResourceRegexIgnoreList
```
wey-chiang, line 291: 👎

wey-chiang, line 291: 👎

internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -222,6 +256,7 @@ func Poll(scanRequest *pollermodels.ScanEntry) (
 		zap.Any("regions", regions),
 		zap.String("resourceType", *scanRequest.ResourceType),
 	)
+	// For simplicity, region ignoreList is not checked here
 	for _, region := range regions {
 		err = utils.Requeue(pollermodels.ScanMsg{
 			Entries: []*pollermodels.ScanEntry{
```
wey-chiang, line 262: 👎

wey-chiang, line 262: :-1-
```
			Entries: []*pollermodels.ScanEntry{
				{
					AWSAccountID:  scanRequest.AWSAccountID,
					IntegrationID: scanRequest.IntegrationID,
					Region:        region,
					ResourceType:  scanRequest.ResourceType,
				},
```
This should include the options passed in by `ScanEntry`

PR #2362, opened 2021-01-07
bulk upload datamodels fix

lindsey-w:
## Background

In working on getting Bulk Upload data for the FE work, came across a bug that the bulk uploader was allowing mutliple data models to be enabled with the same LogType, which should not be supported. This addresses that issue.

## Changes

- Update BulkUpload to add a check for multiple data models enabled per LogType and return failure if there is one.

## Testing

- `mage test:integration`
- manual testing

Output from `panther_analysis_tool`:
```
[INFO]: Uploading pack to Panther
[WARNING]: Failed to upload to Panther
        status code: 400
        error message: only one DataModel can be enabled per ResourceType
```


internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -363,17 +370,27 @@ func readZipFile(zf *zip.File) ([]byte, error) {
 	return ioutil.ReadAll(f)
 }
 
+func validateUploadedDataModel(item *tableItem) error {
+	if len(item.ResourceTypes) > 1 {
+		return errors.New("only one ResourceType may be specified per DataModel")
```
nhakmiller, line 375: I think this should say `only one LogType may be specified per DataModel` to stay consistent with the phrasing.

nhakmiller (approved):

wey-chiang (approved):
LGTM

PR #2369, opened 2021-01-08
Happy New Year

austinbyers:
## Background

Change `(c) 2020` to `(c) 2021` in license headers and email footers

## Changes

All of them. Interestingly, we almost have exactly 2021 source files

> 2,002 changed files with 2,022 additions and 2,018 deletions.

## Testing

`mage clean setup gen fmt test:ci`

### DIFF
--- a/.github/workflows/code-analysis.yml
+++ b/.github/workflows/code-analysis.yml
@@ -1,5 +1,5 @@
 # Panther is a Cloud-Native SIEM for the Modern Security Team.
-# Copyright (C) 2020 Panther Labs Inc
+# Copyright (C) 2021 Panther Labs Inc
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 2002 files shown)


jacknagz:
@austinbyers you actually don't have to do this because Copyright is effective from the date forward in time

austinbyers:
> @austinbyers you actually don't have to do this because Copyright is effective from the date forward in time

Sure, I know we don't have to. But most places do:

Screen Shot 2021-01-08 at 3 41 25 PM

Up to you

jacknagz:
> > @austinbyers you actually don't have to do this because Copyright is effective from the date forward in time
> 
> Sure, I know we don't have to. But most places do:
> 
> Screen Shot 2021-01-08 at 3 41 25 PM
> 
> Up to you

I don't see a reason to

austinbyers:
> I don't see a reason to

I think the reason is for new files. When we add _new_ code, it's probably not valid to say "(c) 2020" because those source files did not exist last year so we couldn't have had a copyright on them at the time.

An alternative to "(c) 2021" would be "(c) 2020-2021", for example

PR #2373, opened 2021-01-11
Fixing flaky test and few S3 destination bugs

### LINKED ISSUE #2179: [BE] Flaky Unit Tests
### Describe the bug

Every now and then, some backend tests fail. This is an issue to track all of them in an effort to fix the flakyness and make them consistent

### Steps to reproduce

1. Run `mage test:ci` until one of them fails. The failure rate is really low (less than 1 out of 20 times), but it exists

### Expected behavior

Tests should always pass or always fail

### Environment

CircleCI

### Additional context

```
--- FAIL: TestSendDataIfTimeLimitHasBeenReached (1.64s)
    s3_test.go:305: FAIL:	Upload(string,string)
        		at: [s3_test.go:300]
    s3_test.go:305: FAIL: 0 out of 1 expectation(s) were met.
        	The code you are testing needs to make 1 more call(s).
        	at: [s3_test.go:305]
    s3_test.go:306: FAIL:	Publish(string)
        		at: [s3_test.go:301]
    s3_test.go:306: FAIL: 0 out of 1 expectation(s) were met.
        	The code you are testing needs to make 1 more call(s).
        	at: [s3_test.go:306]
```

### Screenshots

Screen Shot 2020-12-03 at 10 32 55 AM

kostaspap:
## Background
Closes #2179 

* Fixing flaky test: TestSendDataIfTimeLimitHasBeenReached
* It was possible for S3 destination to block indefinitely while waiting for new events to become available from the parsers and wouldn't flush buffers on schedule. Fixed.
* Updating behavior of S3 destination so that if it has encountered a failure, we no longer try to flush buffers to S3. 

## Testing

- mage test:ci

### DIFF
--- a/internal/log_analysis/log_processor/destinations/s3.go
+++ b/internal/log_analysis/log_processor/destinations/s3.go
@@ -152,8 +152,6 @@ func (d *S3Destination) SendEvents(parsedEventChannel chan *parsers.Result, errC
 	// accumulate results gzip'd in a buffer
 	failed := false // set to true on error and loop will drain channel
 	bufferSet := d.newS3EventBufferSet()
-	eventsProcessed := 0
-
 	zap.L().Debug("starting to read events from channel")
 	run := true
 	for run {
@@ -212,7 +210,7 @@ func (d *S3Destination) SendEvents(parsedEventChannel chan *parsers.Result, errC
 	close(sendChan)
 	sendWaitGroup.Wait() // wait until all writes to s3 are done
 
-	zap.L().Debug("finished sending s3 files", zap.Int("events", eventsProcessed))
+	zap.L().Debug("finished sending s3 files")
 }
 
 // sendData puts data in S3 and sends notification to SNS
--- a/internal/log_analysis/log_processor/destinations/s3.go
+++ b/internal/log_analysis/log_processor/destinations/s3.go
@@ -153,13 +153,10 @@ func (d *S3Destination) SendEvents(parsedEventChannel chan *parsers.Result, errC
 	failed := false // set to true on error and loop will drain channel
 	bufferSet := d.newS3EventBufferSet()
 	eventsProcessed := 0
-	zap.L().Debug("starting to read events from channel")
-	for event := range parsedEventChannel {
-		if failed { // drain channel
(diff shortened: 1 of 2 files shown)


internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -152,14 +152,9 @@ func (d *S3Destination) SendEvents(parsedEventChannel chan *parsers.Result, errC
...
-	eventsProcessed := 0
 	zap.L().Debug("starting to read events from channel")
-	for event := range parsedEventChannel {
-		if failed { // drain channel
-			continue
-		}
-
-		// Check if any buffer has held data for longer than maxDuration
+	run := true
+	for run {
```
giorgosp, line 157: nit: Do you really need the `run` var? Isn't using `break` enough?

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -170,34 +165,43 @@ func (d *S3Destination) SendEvents(parsedEventChannel chan *parsers.Result, errC
...
+			if !ok {
+				// If channel has been closed,
+				// stop the loop
+				run = false
+				break
+			}
 
-		eventsProcessed++
-	}
+			if failed {
```
giorgosp, line 176: nit: Could this be the loop condition as well or be moved at the top of the loop code? If not, could you write a short comment why?

internal/log_analysis/log_processor/destinations/s3.go
kostaspap: The `break` just breaks the `select` statement... we will re-enter the `for` loop.

giorgosp: oh ok

internal/log_analysis/log_processor/destinations/s3.go
kostaspap: Added!

giorgosp: thanks!

giorgosp (approved):

PR #2374, opened 2021-01-11
Trivial log/error tidying

giorgosp:
## Background
-

## Changes

- Improve some error messages when updating a log source
- Change a log line to be logged at Debug level - there is no need to be at Warn level, after discussing with Nick H.

## Testing

-

### DIFF
--- a/internal/compliance/aws_event_processor/processor/process.go
+++ b/internal/compliance/aws_event_processor/processor/process.go
@@ -381,7 +381,7 @@ func processCloudTrailLog(detail gjson.Result, metadata *CloudTrailMetadata, cha
 	if len(newChanges) > 0 {
 		readOnly := detail.Get("readOnly")
 		if readOnly.Exists() && readOnly.Bool() {
-			zap.L().Warn(
+			zap.L().Debug(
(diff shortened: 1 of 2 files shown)


internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -381,7 +381,7 @@ func processCloudTrailLog(detail gjson.Result, metadata *CloudTrailMetadata, cha
 	if len(newChanges) > 0 {
 		readOnly := detail.Get("readOnly")
```
kostaspap, line 382: Are lines 382-389 even needed or should we just get rid of them? CT `readOnly` value used to return inconsistent results so we are not relying on that anyway.

nhakmiller: I had suggested we reduce it to a debug, since we don't really care that much but it might hypothetically be useful  in debugging a situation where resources were getting scanned a lot or something, not sure. I think removing it entirely or leaving as debug are both ok, up to you.

nhakmiller (approved):

PR #2375, opened 2021-01-11
Add Slack integration logs parser

giorgosp:
## Background

Add parser to support Slack integration logs https://api.slack.com/methods/team.integrationLogs

## Changes

- Add parser to support Slack integration logs

## Testing

- Unit

### DIFF
--- a/internal/log_analysis/log_processor/parsers/slacklogs/integrationlogs.go
+++ b/internal/log_analysis/log_processor/parsers/slacklogs/integrationlogs.go
@@ -1,0 +1,42 @@
+package slacklogs
+
+/**
+ * Panther is a Cloud-Native SIEM for the Modern Security Team.
+ * Copyright (C) 2020 Panther Labs Inc
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import "github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog"
+
+const TypeIntegrationLogs = "Slack.IntegrationLogs"
+
+//nolint:lll
+type IntegrationLog struct {
+	UserID            pantherlog.String `json:"user_id" validate:"required" description:"The id of the user performing the action."`
+	UserName          pantherlog.String `json:"user_name" panther:"username" description:"The username of the user performing the action."`
+	ServiceID         pantherlog.String `json:"service_id" description:"The service id for which this log is about."`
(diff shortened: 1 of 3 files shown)


internal/log_analysis/log_processor/parsers/slacklogs/integrationlogs.go
```diff
@@ -0,0 +1,42 @@
...
+	UserName          pantherlog.String `json:"user_name" panther:"username" description:"The username of the user performing the action."`
+	ServiceID         pantherlog.String `json:"service_id" description:"The service id for which this log is about."`
+	ServiceType       pantherlog.String `json:"service_type" description:"The service type for which this log is about."`
+	AppID             pantherlog.String `json:"app_id" description:"The app id for which this log is about."`
+	AppType           pantherlog.String `json:"app_type" description:"The app type for which this log is about."`
+	Date              pantherlog.Time   `json:"date" validate:"required" tcodec:"unix" event_time:"true" description:"The date when the action happened."`
+	ChangeType        pantherlog.String `json:"change_type" validate:"required" description:"The type of this action (added, removed, enabled, disabled, updated)."`
+	Scope             pantherlog.String `json:"scope" validate:"required" description:"The scope used for this action."`
+	Channel           pantherlog.String `json:"channel" validate:"required" description:"The related channel."`
```
georgepsarakis, line 36: This field seems to refer to the Channel ID, as also seen from the tests. Would it seem better if we named it `ChannelID` instead?

internal/log_analysis/log_processor/parsers/slacklogs/integrationlogs.go
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+import "github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog"
+
+const TypeIntegrationLogs = "Slack.IntegrationLogs"
+
+//nolint:lll
+type IntegrationLog struct {
+	UserID            pantherlog.String `json:"user_id" validate:"required" description:"The id of the user performing the action."`
+	UserName          pantherlog.String `json:"user_name" panther:"username" description:"The username of the user performing the action."`
```
georgepsarakis, line 28: Would the inclusion of this field possibly introduce any privacy considerations? Also I think `UserID` makes it always possible to retrieve the user, while `UserName` may change. Maybe omitting it would also encourage querying by `UserID`, which is more robust? A query would work correctly, even at the event of a username change within the given time range.

internal/log_analysis/log_processor/parsers/slacklogs/integrationlogs.go
kostaspap: > This field seems to refer to the Channel ID, as also seen from the tests. Would it seem better if we named it `ChannelID` instead?

We want to preserve the names of the incoming logs, so we shouldn't be doing any renames.... we can add fields, but not rename existing ones :)

internal/log_analysis/log_processor/parsers/slacklogs/integrationlogs.go
kostaspap: > Would the inclusion of this field possibly introduce any privacy considerations? Also I think `UserID` makes it always possible to retrieve the user, while `UserName` may change. Maybe omitting it would also encourage querying by `UserID`, which is more robust? A query would work correctly, even at the event of a username change within the given time range.

We should process and store all the data sent by the upstream source, we shouldn't intentionally omit any field. No privacy consideration in any case here, keep in mind "we" (Panther) don't see any data send to us by users.

kostaspap (approved):
LGTM

PR #2378, opened 2021-01-12
Allow deploying multiple lambdas and stacks directly

tatehanawalt:
## Background

**Syntax to deploy a single lambda or a single stack has not changed**

This change allows deployment of:
• 1 or more specific lambdas
• 1 or more specific stacks, 
• 1 or more specific lambdas and 1 or more specific stacks (stacks deploy first). 

I.E:
`LAMBDA=lambdaname -- mage deploy`
`STACK=stackname -- mage deploy`

**Deploy 1 or more lambdas or stacks by encasing them in a string and separating them with a space:**
`LAMBDA="lamba1 lambda2 lambda3 etc..." -- mage deploy`
`STACK="stack1 stack2 stack3 etc..." -- mage deploy`

**Deploy specific lambdas and stacks (stacks first, then lambdas):**
`LAMBDA=... STACK=... -- mage deploy`

where:
`LAMBD=...`
can be:
• `LAMBDA=<lambdaname>`
• `LAMBDA="<lambdaname>"`
• `LAMBDA="<lambdaname> <lambdaname>"`

Identical syntax for `STACK=...`

I also added myself to the contributors file.

## Why are you making this change? Please provide a reference to any related issues and PRs

• Deploy a specific subset of lambdas with more than 1 lambda and fewer than the sum of lambdas.
• Deploy a specific subset of stacks with more than 1 stack and fewer than the sum of stacks.
• Deploy subset of lambdas and a subset of stacks

## Changes

You can now deploy a single lambda the same way as before:
The same is true for stacks using the STACK environment variable.

You can define 1 or more lambdas AND 1 or more stacks. Stacks are deployed first.

Right now everything is run sequentially. 

Stacks are run before lambdas. This is because deploying a stack should result in skipping deployment of an individual lambda if the lambda was already deployed in the stack.

### Lower level changes:

Split the env variable value by white space and run the respective deploySingle function for each entry in the set.

Specifically there are 2 added methods:

`PantherNames(setString string) []string`
This function returns setString split by spaces, lowercase each entry, prefix panther- if prefix is not prefixed with panther-

`func RunSet(label string, nameset string, callFn func(string) error) int `
This takes a label (for logging purposes) like 'lambda' or 'stack', a string of item identifiers (set of lambda or stack names split by white spaces), and a `deployIndividual` function 'callFn' which deploys the individual respective item. So if nameset contains lambda function names, our callFn would be `deploySingleLambda`

## Testing

`STACK=<stack name> -- mage deploy`
`STACK="<stack1> <stack2>" -- mage deploy`

`LAMBDA=<lambda name> -- mage deploy`
`LAMBDA="<lambda1> <lambda1>" -- mage deploy`

`STACK="<stack1> <stack2>" LAMBDA="<lambda1> <lambda1>" -- mage deploy`


AlexPanther:
>You can define 1 or more lambdas AND 1 or more stacks
Nice! Love the detailed write-up. Would you mind linking the ticket in the "Background" section? You can say:
- (#1234)
- Closes: (#1234)
Saying "Closes" before the ticket number will update the ticket accordingly when the PR is merged.
Overall, seems like a great change!

tatehanawalt:
> > You can define 1 or more lambdas AND 1 or more stacks
> 
> Nice! Love the detailed write-up. Would you mind linking the ticket in the "Background" section? You can say:
> 
> * (#1234)
> * Closes: (#1234)
> 
> Saying "Closes" before the ticket number will update the ticket accordingly when the PR is merged.
> Overall, seems like a great change!

Hey @AlexPanther, I will link tickets in the future. Theres no ticket on this one. I realized it would be possible without impacting existing syntax and it would be easy/quick to implement.

AlexPanther (approved):

tools/mage/deploy/deploy.go
```diff
@@ -73,20 +73,13 @@ func Deploy() error {
...
 	}
 
-	if function := os.Getenv("LAMBDA"); function != "" {
-		function = strings.ToLower(strings.TrimSpace(function))
-		if !strings.HasPrefix(function, "panther-") {
-			function = "panther-" + function
-		}
-		return deploySingleLambda(function)
-	}
+	// Deploy 1 or more stacks with STACK="name1 name2" (ws delimited)
```
AlexPanther, line 76: Nit: I think writing out "whitespace" would be slightly clearer, especially if you have room on the line before the max char length kicks in

tatehanawalt: Do we have a max line characters spec? I will update the ws->white space and update base branch shortly

AlexPanther: I'm not sure if it's strictly enforced or not, but I believe I have been scolded by GoLand for having too long of line.

tatehanawalt: It looks like our line length for go is 100 characters (not enforced) "Do whatever makes code most legible".

It looks like our style guides are internal. Going open an issue to see if we can get that added to our public contributing docs.

tools/mage/deploy/deploy.go
```diff
@@ -73,20 +73,13 @@ func Deploy() error {
...
+	deployed := RunSet("STACK", os.Getenv("STACK"), deploySingleStack)
 
-	if stack := os.Getenv("STACK"); stack != "" {
-		stack = strings.ToLower(strings.TrimSpace(stack))
-		if !strings.HasPrefix(stack, "panther-") {
-			stack = "panther-" + stack
-		}
-		return deploySingleStack(stack)
+	// Deploy 1 or more lambdas with LAMBDA="name1 name2..." (white space delimited)
+	deployed += RunSet("LAMBDA", os.Getenv("LAMBDA"), deploySingleLambda)
```
austinbyers, line 80: `deploySingleLambda` and `deploySingleStack` are no longer accurate, since the env variable now supports multiple. Can you rename them? Perhaps `deployDirect` or `deploySubset` or something

tools/mage/deploy/deploy.go
```diff
@@ -591,3 +584,31 @@ func customResourceVersion() string {
 	// and also the same format used by the master stack.
 	return fmt.Sprintf("%s (%s)", util.Semver(), util.CommitSha())
 }
+
+// Takes a string  and returns the panther- prefixed, lowercased slice of words(strings) (separated by spaces).
+// e.g "oRg-ApI" -> []string{"panther-org-api"}
+// e.g "one two THREE" -> []string{"panther-one", "panther-two", "panther-three"}
+func PantherNames(setString string) []string {
+	set := strings.Fields(setString)
```
austinbyers, line 592: I didn't know about this function! Learn something new every day

tools/mage/deploy/deploy.go
```diff
@@ -591,3 +584,31 @@ func customResourceVersion() string {
...
+	for index, entry := range set {
```
austinbyers, line 593: Nit: I prefer `i` for indexes:

```suggestion
	for i, entry := range set {
```

Either way is fine, but it's common Go style to use short variable names for receivers, indexes, file handles, etc

tools/mage/deploy/deploy.go
```diff
@@ -591,3 +584,31 @@ func customResourceVersion() string {
...
+		entry = strings.ToLower(strings.TrimSpace(entry))
```
austinbyers, line 594: You shouldn't need the `strings.TrimSpace` since `strings.Fields` is already splitting on the spaces?

tools/mage/deploy/deploy.go
```diff
@@ -591,3 +584,31 @@ func customResourceVersion() string {
...
+func RunSet(label string, nameset string, callFn func(string) error) int {
+	returnCount := 0
+	for _, entry := range PantherNames(nameset) {
+		returnCount++
+		log.Info(label, ": ", entry)
+		if err := callFn(entry); err != nil {
+			log.Error("\n\n ", label, ": ", entry, "\n\n ERROR: ", err.Error(), "\n")
+		}
+	}
+	return returnCount
```
austinbyers, line 613: No need for a separate counter; it will always be exactly the same as `len PantherNames(nameset)`

tools/mage/deploy/deploy.go
```diff
@@ -591,3 +584,31 @@ func customResourceVersion() string {
...
+}
+
+// Takes a type, set of types, and a function to call with each type
+func RunSet(label string, nameset string, callFn func(string) error) int {
+	returnCount := 0
+	for _, entry := range PantherNames(nameset) {
+		returnCount++
+		log.Info(label, ": ", entry)
+		if err := callFn(entry); err != nil {
+			log.Error("\n\n ", label, ": ", entry, "\n\n ERROR: ", err.Error(), "\n")
```
austinbyers, line 610: It's generally not recommended to add extra new lines to log messages - each message should fit on its own line and the logging library will handle the spacing of the different message components (level, timestamp, etc)

austinbyers (commented):
Great idea! I just have some style comments, but the implementation looks good

tools/mage/deploy/deploy.go
austinbyers: We do enforce a 140-character line length: https://github.com/panther-labs/panther/blob/master/.golangci.yml#L59-L60

However, that can be easily disabled with a comment since, as you say, it's not always more legible to break up long lines

tools/mage/deploy/deploy.go
austinbyers: You also don't need to explicitly add the "ERROR" part; the Error level will already be logged (and even colored in)

tools/mage/deploy/deploy.go
austinbyers: EDIT: I see now that you're passing in that function and evaluating it on each item separately. Never mind!

tools/mage/deploy/deploy.go
```diff
@@ -73,20 +73,13 @@ func Deploy() error {
...
-	if stack := os.Getenv("STACK"); stack != "" {
-		stack = strings.ToLower(strings.TrimSpace(stack))
-		if !strings.HasPrefix(stack, "panther-") {
-			stack = "panther-" + stack
-		}
-		return deploySingleStack(stack)
+	// Deploy 1 or more lambdas with LAMBDA="name1 name2..." (white space delimited)
+	deployed += RunSet("LAMBDA", os.Getenv("LAMBDA"), deploySingleLambda)
+	if deployed > 0 {
+		return nil
```
austinbyers, line 82: Maybe a comment here about why we're returning early:

```go
    // Already deployed individual STACKs or LAMBDA functions, skip the full deploy process
    return nil

File: tools/mage/deploy/deploy.go (hunk 5 above)
Comment (tatehanawalt) [reply tools/mage/deploy/deploy.go:594]: Wrote this all kind of late haha nice catch

File: tools/mage/deploy/deploy.go (hunk 6 above)
Comment (tatehanawalt) [reply tools/mage/deploy/deploy.go:613]: Ideally this will eventually only run against valid items so if someone enters an invalid lambda or stack identifier it won't result in an increment.

Comment (austinbyers) [reply tools/mage/deploy/deploy.go:613]: But you're using this counter to stop the deploy process early; you *always* want to return a positive value here or else it will trigger a full deploy, which is probably not your intention.

Comment (tatehanawalt) [reply tools/mage/deploy/deploy.go:613]: I see what you mean. 

if LAMBDA and STACK are not assigned anything the count is intended to be zero which would allow a full deploy but if someone defined anything that is invalid it's an indicator they didn't intend to run a full stack.

File: tools/mage/deploy/deploy.go [hunk 9]
```diff
@@ -591,3 +586,30 @@ func customResourceVersion() string {
...
+		set[i] = entry
+	}
+	return set
+}
+
+// Takes a type, set of types, and a function to call with each type
+func RunSet(label string, nameset string, callFn func(string) error) int {
+	pantherNameSet := PantherNames(nameset)
+	for _, entry := range PantherNames(nameset) {
+		log.Info(label, ": ", entry)
```
austinbyers, line 609: Are you sure this works as intended? I think you want

```suggestion
		log.Infof("%s: %s", label, entry)
```

(sorry I missed that the first time)

tools/mage/deploy/deploy.go
```diff
@@ -591,3 +586,30 @@ func customResourceVersion() string {
...
+		if err := callFn(entry); err != nil {
+			log.Error(zap.String(label, entry), zap.Error(err))
```
austinbyers, line 611: Similarly, this could be:

```suggestion
			log.Errorf("%s: %s failed: %s", label, entry, err)
```

In the mage code, we're not worried about the optimization of the low-level zap types and we just use the "[sugared](https://godoc.org/go.uber.org/zap#SugaredLogger)" formatting for simplicity.

The backend code, for better or worse, uses the more verbose[ zap.Logger](https://godoc.org/go.uber.org/zap#Logger) directly

austinbyers (approved):
Looks good! A couple more logging suggestions, but good to merge after that

s0l0ist (approved):

tools/mage/deploy/deploy.go
tatehanawalt: We need to update this in our confluence go style guides docs

tatehanawalt:
I Renamed the method to CallForEachString

CallForEachString takes a slice of strings and returns a slice of errors from running those strings.
Updated method description

This removes 'PantherNames' from the CallForEach method to support reusability.

tools/mage/deploy/deploy.go
```diff
@@ -73,20 +73,15 @@ func Deploy() error {
...
 
-	if function := os.Getenv("LAMBDA"); function != "" {
-		function = strings.ToLower(strings.TrimSpace(function))
-		if !strings.HasPrefix(function, "panther-") {
-			function = "panther-" + function
-		}
-		return deploySingleLambda(function)
-	}
+	// Deploy 1 or more stacks with STACK="name1 name2" (white space delimited)
+	callStackErrors := CallForEachString("STACK", PantherNames(os.Getenv("STACK")), deploySingleStack)
```
austinbyers, line 77: I thought you wanted this to be private? `callForEachString`

Not a big deal either way, but technically this package exports top-level handlers for the top level `magefile.go`, so we shouldn't export routines unless `mage` will be accessing them directly

tools/mage/deploy/deploy.go
```diff
@@ -591,3 +586,32 @@ func customResourceVersion() string {
...
+
+// Call a method for every string in the callOnSet string slice. Return a slice of errors where the
+// index of the error is the index of the callOnSet string used as the argument in the function call.
+func CallForEachString(label string, callOnSet []string, callFn func(string) error) (callErrors []error) {
+	for _, setEntry := range callOnSet {
+		log.Infof("%s: %v", label, setEntry)
+		err := callFn(setEntry)
+		if err != nil {
+			log.Errorf("%s: %s %v", label, setEntry, err)
+		}
```
austinbyers, line 613: Nit: you could put the error on the same line

```suggestion
		if err := callFn(setEntry); err != nil {
			log.Errorf("%s: %s %v", label, setEntry, err)
		}
```

tools/mage/deploy/deploy.go
```diff
@@ -73,20 +73,15 @@ func Deploy() error {
...
-	if stack := os.Getenv("STACK"); stack != "" {
-		stack = strings.ToLower(strings.TrimSpace(stack))
-		if !strings.HasPrefix(stack, "panther-") {
-			stack = "panther-" + stack
-		}
-		return deploySingleStack(stack)
+	// Deploy 1 or more lambdas with LAMBDA="name1 name2..." (white space delimited)
+	callLambdaErrors := CallForEachString("LAMBDA", PantherNames(os.Getenv("LAMBDA")), deploySingleLambda)
+
+	if len(callStackErrors) > 0 || len(callLambdaErrors) > 0 {
```
austinbyers, line 82: Ah, I almost thought this was a bug.

Your `CallForEachString` function returns a slice of _optional_ errors, one for each string parameter. That means it will return, for example `[]{nil, nil}`. And so, `len(callStackErrors)` would be greater than 0, even if there weren't actually any. When I saw `callStackErrors`, I assumed it would only contain, you know, errors, rather than nil pointers.

That means to actually check for an error condition, every caller would need to loop through the error slice to check for non-nil errors. For that reason, my preference would be that the `CallForEachString` just return a normal `error`, which could be, for example, "X errors encountered" - the details have already been logged.

But it's just a design preference and I'm not going to block the PR over it. I love the way you pass the function as an argument now, though

austinbyers (approved):
🎉

tools/mage/deploy/deploy.go
tatehanawalt: That was mainly when PantherNames was in the method body we can re-use CallForEachString outside of deploy if we ever want to we may need to change logging but we can do that if we need to if we end up reusing this method.

tools/mage/deploy/deploy.go
tatehanawalt: Thanks! I wanted to return the slice of errors so we could deal with any errors specifically. The idea is that if we need to deal with errors we probably save the string slice instead of just passing it and we can iterate through the errors and have the index of the string used to call the function equal to the index of the error returned. 

And I also wanted to retain a count of attempted function calls. So if we pass invalid values deploy doesn't end up initiating a full deploy. 

With a slice of errors the idea is that (assuming the method passed only passes errors on failure) we can get a count of errors, a count of success, and a total count of attempts.

tools/mage/deploy/deploy.go
tatehanawalt: I would but I want to capture nil errors for counting success / error rates and have the returned error set indexable by the callOnSet items. If the caller needs to know which argument resulted in the error they can use something like `callset[indexOf(error)]`

austinbyers: Oh right, I forgot you're using that variable later. Carry on

PR #2386, opened 2021-01-12
Make athena view generation more robust

rleighton:
## Before

If there are  just cloud sec tables deployed then we would not generate a view for `all_logs` which is correct but the code returned for the all_logs view will be `""` which would be evaluated and fail, causing an alarm. This happens during initial deployment when we just have cloudsec tables but no log tables.

## After

- If `""` is returned from view generation then do not add to the returned slice of sql statements.

## Testing

- mage test:go


serbantanasa (approved):
Like the more detailed testing and like the handling of the delayed creation of the panther_logs tables in the onboard stack.

PR #2387, opened 2021-01-12
Fix teardown: run in stages

austinbyers:
## Background

`mage teardown` fails on developer deployments with:

```
14:19:23	ERROR	[teardown]	stack panther-onboard: AWS::CloudFormation::Stack panther-onboard DELETE_FAILED: The following resource(s) failed to delete: [SelfRegistration].
14:19:23	ERROR	[teardown]	stack panther-onboard: Custom::SelfRegistration SelfRegistration DELETE_FAILED: Failed to delete resource. ResourceNotFoundException: Function not found: arn:aws:lambda:us-east-1:571590349641:function:panther-source-api
```

This is due to the fact that `mage teardown` deletes most of the stacks in parallel. But we have some dependencies which need to be honored for a clean teardown. This is yet another reason why we need to move away from managing stacks directly with `mage` and instead handle them nested in the root template (https://github.com/panther-labs/panther-enterprise/issues/1197)

## Changes

- Change `mage teardown` to remove stacks in 3 groups:
    - `web`, `onboard`, `appsync`, and `dashboards` do not have downstream dependencies and can be deleted in parallel
    - `cloud-security` and `log-analysis` can be deleted in parallel, but they both depend on the core stack
    - Finally, delete, in sequential order: `core`, `bootstrap-gateway`, `bootstrap`

This code will be removed entirely once we switch to the root stack deployment in v1.16, so don't worry about nitpicking it. I just want to fix teardown for developers testing the upcoming release

## Testing

- `mage teardown` in an empty region
- `mage teardown` after a fresh deploy (in progress)

### DIFF
--- a/tools/mage/teardown/cfn.go
+++ b/tools/mage/teardown/cfn.go
@@ -55,40 +55,34 @@ func DestroyCfnStacks(masterStack string) error {
 		log.Infof("    √ %s deleted (%d/%d)", result.stackName, finishCount, cfnstacks.NumStacks)
 	}
 
-	// Trigger the deletion of the main stacks in parallel
-	//
-	// The bootstrap stacks have to be last because of the ECS cluster and custom resource Lambda.
-	parallelStacks := []string{
-		cfnstacks.Appsync,
-		cfnstacks.Cloudsec,
-		cfnstacks.Core,
-		cfnstacks.Dashboard,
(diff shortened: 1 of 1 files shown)


AlexPanther (approved):

austinbyers:
Teardown was successful, though of course much slower (which is to be expected)

PR #2392, opened 2021-01-13
Backport LogTypesAPI changes from enterprise

alxarch:
## Background

While syncing the custom log updates PR some changes where required on enterprise to avoid cyclic dependencies.
The hide deleted tables PR also changed the LogTypesAPI code.
 
## Changes

- Refactor external dependencies of `logtypesapi.API` to avoid cyclic dependencies. External actions are defined as function fields and 'wired up' in `main/lambda.go`  
- Adds ListDeletedCustomLogs endpoint to LogTypesAPI (not used on OSS but code is best kept in sync)
- Make sure we don't embed errors for `ListAvailableLogTypes`. The FE code does not handle errors on that endpoint.
- 
## Testing

- mage gen fmt test:go
- E2E test pending (just to verify that the log types dropdown works as expected)


internal/core/logtypesapi/available.go
```diff
@@ -62,3 +62,21 @@ skip:
 	}
 	return dst
 }
+
+// ListDeletedLogTypes lists all deleted log type ids
```
giorgosp, line 66: Perhaps mention here that this method returns the logtypes in alphabetical order and remove the trivial comment in line 72.

Also ListDeletedLogTypes -> ListDeletedCustomLogs in the function comment

internal/core/logtypesapi/available.go
```diff
@@ -62,3 +62,21 @@ skip:
...
+	// Sort deleted log types by name
+	sort.Strings(logTypes)
+	return &DeletedCustomLogs{
+		LogTypes: logTypes,
+	}, nil
+}
+
+type DeletedCustomLogs struct {
+	LogTypes []string  `json:"logTypes,omitempty" description:"A list of ids of deleted log types (omitted if an error occurred)"`
+	Error    *APIError `json:"error,omitempty" description:"An error that occurred while fetching the list"`
```
giorgosp, line 81: `ListDeletedCustomLogs` returns `(*DeletedCustomLogs, error) and `DeletedCustomLogs` also has an `Error` field. Why do we need this redundancy? I see this pattern is used by all the API methods too.

The description of the field is also misleading since `ListDeletedCustomLogs` does `return nil, err` if `api.Database.ListDeletedLogTypes(ctx)` returns an error.

internal/core/logtypesapi/api.go
```diff
@@ -39,17 +38,16 @@ const LambdaName = "panther-logtypes-api"
...
 // LogTypesAPI handles the business logic of log types LogTypesAPI
 type LogTypesAPI struct {
-	NativeLogTypes func() []string
-	Database       LogTypesDatabase
-	LambdaClient   lambdaiface.LambdaAPI
-	DataCatalog    datacatalog.Client
+	NativeLogTypes    func() []string
+	Database          LogTypesDatabase
+	UpdateDataCatalog func(ctx context.Context, logType string, from, to []logschema.FieldSchema) error
+	LogTypeInUse      func(ctx context.Context) ([]string, error)
```
giorgosp, line 44: LogTypeInUse -> LogTypesInUse perhaps?

giorgosp (commented):
LGTM, the only confusing thing is API methods returning either an error or a struct that includes the error field.  

If having `error` in the method signature is required by our current API "framework" boilerplate,  could we be consistent in the methods? eg Either use `return nil, err` and remove the `Error` field from the result or use that field only, when an error occurs.

internal/core/logtypesapi/available.go
alxarch:
```suggestion
// ListDeletedCustomLogs lists all deleted log type ids
```

internal/core/logtypesapi/available.go
alxarch: The one is the 'internal' API and the other is the one exposed to FE. In the 'internal' API we need to know about the error.
Otherwise the code becomes a mess where each `if err != nil {` has to embed the error in-place.
The error is embedded once at the LambdaHandler level (see `main/lambda.go`). This is intentional and the only exception is the `ListAvailableLogTypes` endpoint that does not have embedded error handling in FE.

alxarch: The `Error` field exists in the Output sturcts so that 'models.go' gets generated correctly.

internal/core/logtypesapi/api.go
alxarch: I won't change this here as this is a backport PR and I don't want to have another PR on ee just for this name. I will add a FIXME comment
```suggestion
        // FIXME: Rename to LogTypesInUse
	LogTypeInUse      func(ctx context.Context) ([]string, error)
```

internal/core/logtypesapi/available.go
giorgosp: See for example:
```
func (api *LogTypesAPI) GetCustomLog(ctx context.Context, input *GetCustomLogInput) (*GetCustomLogOutput, error) {
	record, err := api.Database.GetCustomLog(ctx, input.LogType, input.Revision)
	if err != nil {
		return nil, err
	}
	if record == nil {
		return nil, NewAPIError(ErrNotFound, fmt.Sprintf("custom log record %s@%d not found", input.LogType, input.Revision))
	}
	return &GetCustomLogOutput{
		Result: record,
	}, nil
}```

Why the error from `api.Database.GetCustomLog()` is not wrapped as an `APIError`?  
Also, `GetCustomLogOutput` also has an `Error` field but it is never written.

Since this pattern exists already in the codebase let's not block this PR.
But the error handling is inconsistent here and confusing. It would be nice if we could improve it in the next PRs.

**Review by giorgosp — APPROVED** [review]

File: internal/core/logtypesapi/available.go (hunk 2 above)
Comment (alxarch) [reply internal/core/logtypesapi/available.go:81]: > But the error handling is inconsistent here and confusing. It would be nice if we could improve it in the next PRs.

This is not true, the error handling *is* consistent. The detail you are missing is that the `Error` field is there only for the 'external' API. All handlers return `error` as the second result as normal Go code does. Any errors are handled in a single place (`main/lambda.go`) and are embedded in the response. The alternative, as I mentioned above, is to spray the code with 
```golang
if err != nil {
   if !errors.Is(err, *APIError) {
     L(ctx).Error("op failed", zap.Error(err), zap.String("op", "Foo"))
   }
   return FooOutput{
     Error: WrapAPIError(err),
   }, nil
}
```

This is tedious and obstructs program flow.
By handling all errors in a single place we clear the code and ensure that no errors go unlogged by mistake.

PR #2394, opened 2021-01-13
Dynamic Alert fields

### LINKED ISSUE #2166: [FE] Dynamic Alert Fields Support
### Description

As an Analyst, I want to be able to dynamically set the severity of an alarm to:
- Route alarms on critical and production resources to monitored Slack channels or kick off related workflows
- Similarly, they can choose to send lower priority alarms to a lower-priority queue to be triaged
- Tailor the alarm runbook and description
I can do this with the new functions `description`, `severity`, `destination`, etc. Now that I can set these fields, I want to see their test results when running unit tests from the UI.

### Related Services

web:
- Edit Rule
- Alert Details page 

### Designs

Find RFC [here](https://github.com/panther-labs/panther-enterprise/pull/1095/files?short_path=f1e6929#diff-f1e69298df6b23442c7d0ca35b77b0050c235b672d5678bae61b0563acbd6900)

### Acceptance Criteria
- Add all dynamic alert fields to `AlertSummary` and `AlertDetails`
- In Edit Rules page:
  - In the Test Result component, in addition to what is already shown (pass/fail status, title and dedup string) we should also display the dynamic fields (i.e. `reference`, `runbook`, `description`, `severity`, and `destination overrides`) if the related function has been defined for it. These fields **might** be present from the API, depending on whether the user has specified the related functions or not.

- In Alert Details page:
  - Whenever displaying any of the dynamic-alert fields, frontend should make sure to read those values directly from the alert itself and **not** from the rule/policy that's associated to this alert

vorillaz:
## Description

Since the begging of Panther rules were kinda static. With the `dedup()` `rule()` and `title` methods we can use Python to dynamically create alert metadata and test the output. This PR introduces 5 new methods `reference()`, `runbook()`, `description()`, `severity()` and `destinations()`. The `RuleFormTestResult` file has been changed accordingly, there were no needed changes for the Alert details and the Alert summary definitions as the output is parsed by the backend API.
RFC can be found [here](https://github.com/panther-labs/panther-enterprise/pull/1095/files?short_path=f1e6929#diff-f1e69298df6b23442c7d0ca35b77b0050c235b672d5678bae61b0563acbd6900).

Closes #2166 

### How to test this out:
Create a new rule and fill out the form, add a severity, runbook info etc. Override the metadata of the rule using Rule function, you can use some plain Python code as:
```python
def reference(event):
    return 'https://example.com'

def dedup(event):
    return 'Dedup'

def runbook(event):
    return 'https://example.com'
``` 
Run the tests and checkout the generated alerts.

### Designs:
https://share.goabstract.com/239ce451-9027-4a24-a2bb-33af6007360a?mode=build&sha=c0dc616dbe4787a15a7c86b2ea87692f4780bb95

### Screenshots:
Screenshot 2021-01-13 at 2 48 49 PM

Screenshot 2021-01-13 at 2 48 58 PM

Screenshot 2021-01-13 at 2 49 25 PM

Screenshot 2021-01-13 at 2 49 08 PM

Screenshot 2021-01-13 at 2 49 31 PM

### Checklist:


web/src/components/forms/RuleForm/RuleFormTestResult/RuleFormTestResult.tsx
```diff
@@ -88,7 +101,7 @@ const RuleFormTestResult: React.FC<RuleFormTestResultProps> = ({ testResult }) =
             {dedupFunction && (
               <React.Fragment>
                 <Box as="dt" color="navyblue-100">
-                  Dedup String
+                  Dedup Message
```
3nvi, line 104:
```suggestion
                  Dedup String
```

Designs have it wrong, we should keep it as is. Sorry for any confusion

3nvi (approved):
Looks good @vorillaz . Let's just revert the `Dedup String` text and 🚢 it

web/src/components/forms/RuleForm/RuleFormTestResult/RuleFormTestResult.tsx
vorillaz: Fixed, no worries at all :)

PR #2396, opened 2021-01-13
Fixed backwards compatibility bug

kostaspap:
## Background

* Uncommented a unit test after fixing the root cause
* Backporting naming changes to the API's fields (making them uppercase)

## Testing

- mage test:ci


internal/core/source_api/api/update_integration.go
```diff
@@ -241,7 +241,8 @@ func (api *API) updateTables(item *ddb.Integration, input *models.UpdateIntegrat
 	var existingLogTypes, newLogTypes []string
 	switch item.IntegrationType {
 	case models.IntegrationTypeAWS3:
-		existingLogTypes = item.S3PrefixLogTypes.LogTypes()
+		// Need to include `item.LogTypes` for backwards compatibility reasons
+		existingLogTypes = append(item.S3PrefixLogTypes.LogTypes(), item.LogTypes...)
```
kostaspap, line 245: @giorgosp I noticed this after seeing a test failing. The item passed here could be of "old" type, so it could still have some values.

giorgosp: Good catch thanks. We could also make `updateTables` work with a `SourceIntegration` model by using `itemToIntegration`, which handles the backwards-compatibility alone. 

This way we will be consistent in which model we use in our logic and avoid similar issues in the future. Now in some places we use `SourceIntegration` and in some others we use the raw db model.

giorgosp (approved):

internal/core/source_api/api/update_integration.go
kostaspap: Yes agreed, I started doing that but that required a significantly bigger refactoring that I didn't have time to perform now. We can do it in a separate iteration, my main goal was to fix this one.

giorgosp: Np will have it in mind too

Issue #2341, opened 2021-01-05
Support AWS WAF logs


kostaspap:
### Description

Support AWS Web ACL logs: https://docs.aws.amazon.com/waf/latest/developerguide/logging.html

### Acceptance Criteria

- Users can select AWS Web ACL logs when onboarding a new source from S3
- Users can select AWS Web ACL logs when creating a new rule

georgepsarakis:
## Introduction

In order to make the new log type available to Panther, we need to perform the following steps:
- Configure Web ACLs and enable logging in order to obtain log samples and identify their structure.
- Define a new parser under the [awslogs namespace](https://github.com/panther-labs/panther/tree/a54d4ea/internal/log_analysis/log_processor/parsers/awslogs).
- Follow the [documentation guide](https://docs.runpanther.io/v/master/development/writing-parsers) for writing new parsers. An example of the latest log parser framework usage can be found in [Gravitational Teleport Audit log parser definition](https://github.com/panther-labs/panther/blob/a54d4eae3e93e6757dfcb0a6e93eda87445ae785/internal/log_analysis/log_processor/parsers/gravitationallogs/audit.go).
- Perform Panther end-to-end testing, after deploying changes and amend the logging configuration. 
  * Check that new logs are parsed.
  * Write an example alert rule.
  * Check schema in AWS Glue Data Catalog and run sample queries on Athena.

## Cases for Consideration

A Web ACL can contain:
- Rule Groups (Managed or Custom): a reusable set of rules that can be referenced in multiple Web ACL. AWS provides managed rule groups, with rules for stopping vulnerability scanning activity, blocking SQL Injection or XSS attempts.
- Single Rule: contains a possibly complex expression composed by one or more statements that match request parameters. Multiple statements can be evaluated combined with a boolean operator (OR, AND).
- Rate-based Rule: this type of rule sets a request count limit based on the client IP, when the request matches the given criteria.  

Each Rule has a corresponding action:
- `ALLOW`
- `BLOCK`
- `COUNT`

> Note that Rule Groups don't have an assigned action, this is specified through their rules.

Each Rule/Rule Group is also assigned a priority, which modifies the order of processing. This matters because, when a matching rule is found that has a `ALLOW`/`BLOCK` action defined, the processing stops (more on this subject [here](https://docs.aws.amazon.com/waf/latest/developerguide/web-acl-processing.html#rule-group-rule-exclusion)).

[...]

PR #2403, opened 2021-01-14
Introduce AWS WAF Web ACL log parser

georgepsarakis:
## Background

Enable log parsing for [AWS WAF Web ACL](https://docs.aws.amazon.com/waf/latest/developerguide/logging.html).

Addresses https://github.com/panther-labs/panther/issues/2341.

## Changes

- Define new log parser `struct` `WAFWebACL` along with nested object `struct` definitions.
- Register the new parser in `internal/log_analysis/log_processor/parsers/awslogs/awslogs.go`.
- Add a list of test cases specified in #2341, in YAML format.

## Testing

- New log type is available on Panther frontend for selection.
- Rule Engine:
  * Composed a rule that checked for specific log event values.
  * Send HTTP requests that were blocked by WAF rule.
  * Alert was triggered in Panther. 
- Verified log processing by inspecting Lambda Cloudwatch Logs.
- Verified AWS Glue Data Catalog registration by running sample Athena queries.

### DIFF
--- a/internal/log_analysis/log_processor/parsers/awslogs/testdata/waf_web_acl_tests.yml
+++ b/internal/log_analysis/log_processor/parsers/awslogs/testdata/waf_web_acl_tests.yml
@@ -1,0 +1,631 @@
+# Panther is a Cloud-Native SIEM for the Modern Security Team.
+# Copyright (C) 2020 Panther Labs Inc
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+name: waf_web_acl_no_match
+logType: AWS.WAFWebACL
+input: |
+  {
+    "timestamp": 1610452695501,
+    "formatVersion": 1,
+    "webaclId": "arn:aws:wafv2:us-west-2:123456789EXAMPLE:regional/webacl/panther_web_dev/f96fe2f4-a2f4-4c06-a4bc-afd495bd675f",
+    "terminatingRuleId": "Default_Action",
+    "terminatingRuleType": "REGULAR",
+    "action": "ALLOW",
+    "terminatingRuleMatchDetails": [],
+    "httpSourceName": "ALB",
+    "httpSourceId": "123456789EXAMPLE-app/web/768dd9170f08bb0d",
+    "ruleGroupList": [],
+    "rateBasedRuleList": [],
+    "nonTerminatingMatchingRules": [],
+    "httpRequest": {
+      "clientIp": "10.0.0.1",
+      "country": "GR",
+      "headers": [
+        {
+          "name": "Host",
+          "value": "web-123456789.us-west-2.elb.amazonaws.com"
+        },
+        {
+          "name": "user-agent",
+          "value": "curl/7.64.1"
+        },
+        {
+          "name": "accept",
+          "value": "*/*"
+        }
+      ],
+      "uri": "/",
+      "args": "o=11 AND 1=1",
+      "httpVersion": "HTTP/2.0",
+      "httpMethod": "HEAD",
+      "requestId": "1-5ffd8ed7-1c9649c448649a15241ac25a"
+    }
+  }
+result: |
+  {
+      "timestamp": 1610452695501,
+      "formatVersion": 1,
+      "webaclId": "arn:aws:wafv2:us-west-2:123456789EXAMPLE:regional/webacl/panther_web_dev/f96fe2f4-a2f4-4c06-a4bc-afd495bd675f",
+      "terminatingRuleId": "Default_Action",
+      "terminatingRuleType": "REGULAR",
+      "action": "ALLOW",
+      "httpSourceName": "ALB",
+      "httpSourceId": "123456789EXAMPLE-app/web/768dd9170f08bb0d",
+      "httpRequest": {
+        "clientIp": "10.0.0.1",
+        "country": "GR",
+        "headers": [
+          {
+            "name": "Host",
+            "value": "web-123456789.us-west-2.elb.amazonaws.com"
+          },
+          {
+            "name": "user-agent",
+            "value": "curl/7.64.1"
+          },
+          {
+            "name": "accept",
+            "value": "*/*"
+          }
+        ],
+        "uri": "/",
+        "args": "o=11 AND 1=1",
+        "httpVersion": "HTTP/2.0",
+        "httpMethod": "HEAD",
+        "requestId": "1-5ffd8ed7-1c9649c448649a15241ac25a"
+      },
+      "p_event_time": "2021-01-12T11:58:15.501Z",
+      "p_any_ip_addresses": ["10.0.0.1"],
+      "p_any_trace_ids": ["1-5ffd8ed7-1c9649c448649a15241ac25a"],
+      "p_log_type": "AWS.WAFWebACL"
+    }
+---
+name: waf_web_acl_rule_exclusion
+logType: AWS.WAFWebACL
+input: |
+  {
+    "timestamp": 1610450198436,
+    "formatVersion": 1,
+    "webaclId": "arn:aws:wafv2:us-west-2:123456789EXAMPLE:regional/webacl/panther_web_dev/f96fe2f4-a2f4-4c06-a4bc-afd495bd675f",
+    "terminatingRuleId": "Default_Action",
+    "terminatingRuleType": "REGULAR",
+    "action": "ALLOW",
+    "terminatingRuleMatchDetails": [],
+    "httpSourceName": "ALB",
+    "httpSourceId": "123456789EXAMPLE-app/web/768dd9170f08bb0d",
+    "ruleGroupList": [
+      {
+        "ruleGroupId": "AWS#AWSManagedRulesSQLiRuleSet",
+        "terminatingRule": null,
+        "nonTerminatingMatchingRules": [],
+        "excludedRules": [
+          {
+            "exclusionType": "EXCLUDED_AS_COUNT",
+            "ruleId": "SQLi_QUERYARGUMENTS"
+          }
+        ]
+      }
+    ],
+    "rateBasedRuleList": [],
+    "nonTerminatingMatchingRules": [],
+    "httpRequest": {
+      "clientIp": "10.0.0.1",
+      "country": "GR",
+      "headers": [
+        {
+          "name": "Host",
+          "value": "web-123456789.us-west-2.elb.amazonaws.com"
+        },
+        {
+          "name": "user-agent",
+          "value": "curl/7.64.1"
+        },
+        {
+          "name": "accept",
+          "value": "*/*"
+        }
+      ],
+      "uri": "/",
+      "args": "o=10 AND 1=1",
+      "httpVersion": "HTTP/2.0",
+      "httpMethod": "HEAD",
+      "requestId": "1-5ffd8516-7f16578f3b0262c41c10f7a9"
+    }
+  }
+result: |
+  {
+    "timestamp": 1610450198436,
+    "formatVersion": 1,
+    "webaclId": "arn:aws:wafv2:us-west-2:123456789EXAMPLE:regional/webacl/panther_web_dev/f96fe2f4-a2f4-4c06-a4bc-afd495bd675f",
+    "terminatingRuleId": "Default_Action",
+    "terminatingRuleType": "REGULAR",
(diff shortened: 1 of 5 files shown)


internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
```diff
@@ -37,8 +37,11 @@ import (
 	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog/tcodec"
 )
 
+const ConfigAllowSkipEnvVar = "PANTHER_TESTING_LOG_ANALYSIS_LOG_TESTING_ALLOW_SKIP"
```
georgepsarakis, line 40: This is a temporary commit and will be removed from this PR. I just wanted to display the concept and if there is consensus, I can create a separate PR for this change. I found it particularly helpful while debugging individual cases.

internal/log_analysis/log_processor/parsers/awslogs/testdata/waf_web_acl_tests.yml
```diff
@@ -0,0 +1,631 @@
...
+    "timestamp": 1610452695501,
+    "formatVersion": 1,
+    "webaclId": "arn:aws:wafv2:us-west-2:123456789EXAMPLE:regional/webacl/panther_web_dev/f96fe2f4-a2f4-4c06-a4bc-afd495bd675f",
+    "terminatingRuleId": "Default_Action",
+    "terminatingRuleType": "REGULAR",
+    "action": "ALLOW",
+    "terminatingRuleMatchDetails": [],
+    "httpSourceName": "ALB",
+    "httpSourceId": "123456789EXAMPLE-app/web/768dd9170f08bb0d",
+    "ruleGroupList": [],
```
kostaspap, line 30: We don't want that - unnecessary bytes used up. I think if you add an `json:"...,omitempty"` directive, these will be skipped. 
Also Parquet doesn't like empty arrays or maps that much :)

georgepsarakis: Not sure I understand ... this is the input, so it is the log entry as stored on S3. Maybe you meant that we don't want that in the result? If yes, note that empty objects are omitted!

kostaspap: > Not sure I understand ... this is the input, so it is the log entry as stored on S3. Maybe you meant that we don't want that in the result? If yes, note that empty objects are omitted!

🤦  I did a quick look over that PR and probably I got confused while going through the lines.... Yes, exactly what you said. We are good 😄

internal/log_analysis/log_processor/logtypes/logtesting/logtesting.go
```diff
@@ -37,8 +37,11 @@ import (
...
+
 // RunTestsFromYAML reads all test cases in a YAML file and runs them.
 func RunTestsFromYAML(t *testing.T, resolve logtypes.Finder, filename string) {
```
alxarch, line 43:
```suggestion
// If filter are provided, only tests that match the filters are run 
func RunTestsFromYAML(t *testing.T, resolve logtypes.Finder, filename string, filter ...string) {
```

Maybe adding a `filters` vadiadic arg here would be easier.
Using `stringset.Contains` should be fine to filter the test cases.

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog"
+)
+
+// AWS WAF Web ACL Log event structure: https://docs.aws.amazon.com/waf/latest/developerguide/logging.html
+// File naming convention follows the Firehose delivery stream pattern:
+// - https://docs.aws.amazon.com/firehose/latest/dev/basic-deliver.html#s3-object-name
+// - The prefix `aws-waf-logs-` is mandatory for Web ACL logging delivery stream names.
+// nolint:lll,maligned
+type WAFWebACL struct {
+	Action                      pantherlog.String         `json:"action" validate:"required" description:"The action applied by WAF. Possible values for a terminating rule: ALLOW and BLOCK. COUNT is not a valid value for a terminating rule."`
```
alxarch, line 31: You can use `oneof` validation for this.

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
```diff
@@ -0,0 +1,100 @@
...
+	RuleGroupList               []RuleGroupListDetail     `json:"ruleGroupList" description:"The list of rule groups that acted on this request. In the preceding code example, there is only one."`
+	TerminatingRuleID           pantherlog.String         `json:"terminatingRuleId"`
+	TerminatingRuleMatchDetails []RuleMatchDetail         `json:"terminatingRuleMatchDetails" description:"Detailed information about the terminating rule that matched the request. A terminating rule has an action that ends the inspection process against a web request. Possible actions for a terminating rule are ALLOW and BLOCK. This is only populated for SQL injection and cross-site scripting (XSS) match rule statements. As with all rule statements that inspect for more than one thing, AWS WAF applies the action on the first match and stops inspecting the web request. A web request with a terminating action could contain other threats, in addition to the one reported in the log."`
+	TerminatingRuleType         pantherlog.String         `json:"terminatingRuleType" description:"The type of rule that terminated the request. Possible values: RATE_BASED, REGULAR, GROUP, and MANAGED_RULE_GROUP."`
```
alxarch, line 41: Ditto `oneof`

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
```diff
@@ -0,0 +1,100 @@
...
+type ExcludedRule struct {
+	ExclusionType pantherlog.String `json:"exclusionType" description:"A type that indicates that the excluded rule has the action COUNT (most likely value is EXCLUDED_AS_COUNT)."`
+	RuleID        pantherlog.String `json:"ruleId" description:"The ID of the rule within the rule group that is excluded."`
+}
+
+// nolint:lll,maligned
+type RuleDetail struct {
+	RuleID           pantherlog.String  `json:"ruleId" description:"The Rule ID."`
+	Action           pantherlog.String  `json:"action" description:"The configured rule action. For non-terminating rules the value is always COUNT."`
+	RuleMatchDetails *[]RuleMatchDetail `json:"ruleMatchDetails" description:"Detailed information about the rule that matched the request. This field is only populated for SQL injection and cross-site scripting (XSS) match rule statements."`
```
alxarch, line 64: Slices do not need to be pointers to be omitempty
```suggestion
	RuleMatchDetails []RuleMatchDetail `json:"ruleMatchDetails" description:"Detailed information about the rule that matched the request. This field is only populated for SQL injection and cross-site scripting (XSS) match rule statements."`
```

alxarch (approved):
Well done, minor observations.
About the skip thing I would prefer a solution that does not rely on developers not forgetting to remove a skip. We can discuss further.

georgepsarakis:
Thanks for the review @alxarch ! I will reply on all comments shortly.

With regard to test skip functionality, I think we can decouple it entirely from requiring YAML changes that could be committed, by requiring an environment variable that contains a comma-separated list of test names on which we would like to focus. For example, `PANTHER_TESTING_LOG_PARSER_YAML_FOCUS=test_1,test_2`. Either way, we can discuss it at a later time, I'll remove the commit for now 😄 .

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
georgepsarakis: I should have asked what is the best approach with regard to validations.

- Is it better to mark all known top-level fields as required in order to be notified if the structure changes in the future?
- Is it a good strategy to add value constraints? I think leaving it unconstrained allows for more flexibility, as new values may be added transparently.

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
georgepsarakis: You are right, tests pass.

georgepsarakis: Fixed in https://github.com/panther-labs/panther/pull/2403/commits/6cb625ed946f878b556621ddcefb65bc65af091d

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
kostaspap: I would say, in general, go for the 2nd. If we perform strict validation and the upstream gets updated, then these events will be dropped. We should apply strict validation if we think it is really needed[1]

[1] Panther is able to perform classification of logs. If someone adds a source and selects more than one log types, we are going to try to parse the incoming events with all the provided parsers (identified by log types). We try the parsers serially: When a parser manages to parse an event, we consider the classification successful. That's where the `validate` tags are useful - if we are sure that some fields are required in a log, by marking the fields as `required` we make sure that we don't misclassify the logs. Same for requiring the fields to get specific values (like this case)
For this specific case, I would say if you see that these logs have `required` fields that are not present in other AWS logs, that's good enough, we don't have necessarily to specify the exact values.

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
```diff
@@ -0,0 +1,100 @@
...
+	TerminatingRuleID           pantherlog.String         `json:"terminatingRuleId"`
+	TerminatingRuleMatchDetails []RuleMatchDetail         `json:"terminatingRuleMatchDetails" description:"Detailed information about the terminating rule that matched the request. A terminating rule has an action that ends the inspection process against a web request. Possible actions for a terminating rule are ALLOW and BLOCK. This is only populated for SQL injection and cross-site scripting (XSS) match rule statements. As with all rule statements that inspect for more than one thing, AWS WAF applies the action on the first match and stops inspecting the web request. A web request with a terminating action could contain other threats, in addition to the one reported in the log."`
+	TerminatingRuleType         pantherlog.String         `json:"terminatingRuleType" description:"The type of rule that terminated the request. Possible values: RATE_BASED, REGULAR, GROUP, and MANAGED_RULE_GROUP."`
+	Timestamp                   pantherlog.Time           `json:"timestamp" tcodec:"unix_ms" event_time:"true" description:"The timestamp in milliseconds."`
```
kostaspap, line 42: I imagine this should be required field? Given it is the timestamp field.

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
```diff
@@ -0,0 +1,100 @@
...
+	WebACLID                    pantherlog.String         `json:"webaclId" description:"The GUID of the web ACL."`
```
kostaspap, line 43: Seems that this could be required as well?

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
giorgosp: Generally don't think we should add such validations to enum fields, because if the API adds a value our classification will break and will just throw away events until we fix it. 
I suggest we add `oneof` validations only if there isn't any other way to distinctly identify this event.

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
georgepsarakis: Thanks for the thorough explanation, I had a knowledge gap on how classification may work. I went ahead (probably before your comment) and marked certain common fields (present on all inspected logs) as required. Several are unique across log types, so they should provide accurate log type identification.

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
georgepsarakis: @alxarch any objections on https://github.com/panther-labs/panther/pull/2403/files#r558287941 ?

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
georgepsarakis: Sure, done in 868c5808a07b848d70aa3d55662d35c52290fff6.

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
georgepsarakis: Exactly, done in `868c5808a07b848d70aa3d55662d35c52290fff6`.

kostaspap (approved):
🎉

internal/log_analysis/log_processor/parsers/awslogs/waf_web_acl.go
alxarch: The suggestion was only to ensure proper classification. `action` seems like quite a generic field name to only set as `required`. Since you added `required` on other fields as well, as long as you have checked they don't overlap with other schemas, it's fine. 

We have discussed checking that `required` fields in schemas do not overlap between log types used together in a classification. We should revisit the idea.

alxarch: So to answer the question: No objections!

PR #2406, opened 2021-01-14
Indicate missing fields in Rules and Policy creation forms

### LINKED ISSUE #2064: Indicate missing fields when creating new rule or policy
### Description

When creating a new rule or policy it's hard to know what fields are missing before clicking "Create". We should allow the user to click "Create" then indicate which fields are missing with red text. This would make what's missing more obvious.

### Related Services

Rule Creation

### Designs

N/A

### Acceptance Criteria

- "Create" button in Rule or Policy Creation is clickable no matter what's been filled out
- Red text alerts are triggered if user clicks "create" and fields are missing information

kouknick:
## Description

Allow the user to click "Create" then indicate which fields are missing with red text in the rules and policy creation forms.
- "Create" button in Rule or Policy Creation is clickable no matter what's been filled out
- Red text alerts are triggered if user clicks "create" and fields are missing information

Closes #2064 

### Screenshots:
image
image
image
_*Other forms are not affected by this change, the submit button should be disabled as long as the form is not valid_
### Checklist:

- [ ] Make sure you reference the related design document if it exists


web/src/components/buttons/SubmitButton/SubmitButton.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+  <Box position="relative">
+    <Formik
+      validationSchema={validationSchema}
+      initialValues={initialValues}
+      onSubmit={values => {
+        onSubmit(values);
+      }}
+    >
+      <Form>
+        <Field as={FormikTextInput} placeholder="Write something" name="text" label="Text" />
```
alexmylonas, line 41: I think this field needs a `required` property to make more sense!

web/src/components/buttons/SubmitButton/SubmitButton.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+import React from 'react';
+import { fireEvent, render, waitFor } from 'test-utils';
+import * as Yup from 'yup';
+import { Box } from 'pouncejs';
+import { Formik, Form, Field } from 'formik';
+import FormikTextInput from 'Components/fields/TextInput';
+import SubmitButton from './index';
+
+const validationSchema = Yup.object().shape({
+  text: Yup.string().min(10),
```
alexmylonas, line 28:
```suggestion
  text: Yup.string().min(10).required(),
```

web/src/components/buttons/SubmitButton/SubmitButton.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    expect(container).toMatchSnapshot();
+  });
+
+  it('should be disabled by default as long as the form is pristine or invalid', async () => {
+    const { getByText, getByLabelText } = render(<TestForm onSubmit={jest.fn()} />);
+    const saveButton = getByText('Save');
+    const textField = getByLabelText('Text');
+
+    // Type an invalid value to the input (less than 10 characters)
+    fireEvent.change(textField, { target: { value: 'invalid' } });
```
alexmylonas, line 64: If you test if button attribute before change it will not be disabled (due to missing `required`)

alexmylonas (changes requested):
Good job on your first PR! A couple of small comments but its looking good!

web/src/components/buttons/SubmitButton/SubmitButton.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    await waitFor(() => expect(saveButton).toHaveAttribute('disabled'));
+    // Type a valid value to the input
+    fireEvent.change(textField, { target: { value: 'valid text' } });
+    await waitFor(() => expect(saveButton).not.toHaveAttribute('disabled'));
+  });
+
+  it('should be always enabled if pristine and invalid submission are both enabled', async () => {
```
3nvi, line 71: can we add an extra case where the form has been modified via `fireEvent`, it's invalid but `allowInvalidSubmission` is `true`.

In this case, the `SubmitButton` should be enabled

3nvi (approved):
Way to go @kouknick! Props on your first PR 🎉 

That was actually a really nice way to handle it. In my mind, this was more complex, but the solution you reduced it to was super nice! The tests are really nice and clear 😄 

I'm pre-approving pending a minor extra test case & the changes requested by @alexmylonas , but overall super nice

web/src/components/buttons/SubmitButton/SubmitButton.test.tsx
kouknick: Extra test case added! Thanks for the feedback!

web/src/components/buttons/SubmitButton/SubmitButton.test.tsx
kouknick: Nice catch!

alexmylonas (approved):
Solid work! 💯

Issue #2407, opened 2021-01-14
Alert context serialization failure 


georgepsarakis:
### Describe the bug

Alert context callbacks that include an event nested object (or the event itself) cannot be serialized:

```python
def alert_context(event):
    return {'rate_based_rule': event['rateBasedRuleList'][0]}
```

Relates to https://github.com/panther-labs/panther/pull/2359.

### Steps to reproduce

1. Edit or create a rule for which the log type could contain a nested objects.
2. Define an `alert_context` callback that uses a nested object (list, dictionary) as one of its values.
3. Test the rule with a sample JSON that would embed this value in the alert context.
4. See error: `TypeError: Object of type ImmutableDict is not JSON serializable` or `TypeError: Object of type ImmutableList is not JSON serializable`

### Expected behavior

The alert context should be JSON-serializable.

### Environment

How are you deploying or using Panther?

- Panther version or commit: af8e73ce54e4fd60a1fd8cfe10c5c577eddc794f
- OS: MacOS 11.1
- Browser: Chrome 87.0.4280.141

### Screenshots

![image](https://user-images.githubusercontent.com/963304/104586824-adf77880-566e-11eb-846e-4c8c8acd0018.png)

georgepsarakis:
## Suggested Solution

We can consider introducing a [custom JSONEncoder](https://docs.python.org/3.7/library/json.html#json.JSONEncoder) or [provide a `default` lambda](https://docs.python.org/3.7/library/json.html#json.dump) and use it during alert context serialization:

https://github.com/panther-labs/panther/blob/8609cb6b6422846fcfc85eb347d0fbad12f7b00a/internal/log_analysis/rules_engine/src/rule.py#L270

Since the alert context callback is executed only when rule matches are found, the performance impact should not be a consideration:

https://github.com/panther-labs/panther/blob/8609cb6b6422846fcfc85eb347d0fbad12f7b00a/internal/log_analysis/rules_engine/src/rule.py#L219

PR #2409, opened 2021-01-14
JSON serialization for alert context with immutable event objects

georgepsarakis:
## Background

An exception may be raised if alert context contains `ImmutableDict` / `ImmutableList` instances.

See #2407 for details.

## Changes

- Update type hints for `Rule` methods to define `PantherEvent` type instead of `Mapping` for event parameters.
- Add a JSON encoder method that is able to serialize `ImmutableDict` / `ImmutableList` objects and expose it through `PantherEvent`.
- Always pass `PantherEvent` instances when constructing `Rule` instances in tests.
- Provide failing test for alert context that includes event nested objects.

## Testing

- `mage test:python`


kostaspap (commented):
Code change looks good. 
Can you verify that if you follow the steps outlined here https://github.com/panther-labs/panther/issues/2407 , we no longer get error?

georgepsarakis:
Added a encoder test case to ensure `TypeError` is raised for non-serializable objects: https://github.com/panther-labs/panther/pull/2409/commits/448df66bb5740c9248e6e4c5cc3b694cf24c0630#diff-8ca24d631cffea83acfd6ad9687c11eb7a8ccfb0ccb25c52c94dd0a60ad3a7a1R166

georgepsarakis:
> Code change looks good.
> Can you verify that if you follow the steps outlined here #2407 , we no longer get error?

With this branch deployed and alert context callback:

```python
def alert_context(event):
    return {'ips': event['p_any_ip_addresses'] }
```

image

On the master branch:

image

kostaspap (approved):

PR #2410, opened 2021-01-14
fix: issue where alert wouldn't render if rule got deleted

3nvi:
## Background

When a rule is deleted you are expected to be able to still see the alerts that this rule has generated. This used to work, but in the previous sprint we modified some code and now it doesn't. This PR fixes that.

## Changes

- Allow rendering of Alert when associated Rule or Policy has been deleted

## Testing

- Locally


alexmylonas (approved):
LGTM!

PR #2413, opened 2021-01-14
Pass through Log Types filter

nhakmiller:
## Background

The `ListRules` endpoint currently drops the `LogTypes` filter entirely (ignoring it). This fix just passes that filter through.

## Changes

- Pass `LogTypes` filter through when calling `ListRules`

## Testing

- Deployed to dev env and tested manually


nhakmiller:
Going to add `sync-to-release` just because its such a minor but user facing bug.

kostaspap (approved):

PR #2418, opened 2021-01-14
Generated Alert Fields - Destinations Fix

wey-chiang:
## Background

Related to some issues discovered when debugging the FE implementation. 

The analysis-api test record expects a string to be returned from DDB for the destinationsOutput, however it will return a string set (since that's what the rules-engine will store in DDB)

## Changes

- Changed analysis-api model destinationsOutput field to be `[]string`
- Patched the `buildTestSubRecord` for `Destinations` with `buildTestSubRecordList` which handles the case of 

## Testing

- `mage test:ci`
- Deployment and manual testing in `us-east-1`

### DIFF
--- a/internal/core/analysis_api/analysis/rule_engine.go
+++ b/internal/core/analysis_api/analysis/rule_engine.go
@@ -145,7 +145,7 @@ func buildTestSubRecordList(output []string, error string) *models.TestDetection
 	result := &models.TestDetectionSubRecord{}
 
 	if len(output) != 0 {
-		result.Output = aws.String(strings.Join(output, ","))
+		result.Output = aws.String(strings.Join(output, ", "))
 	}
 	if error != "" {
 		result.Error = &models.TestError{Message: error}
--- a/internal/core/analysis_api/analysis/rule_engine.go
+++ b/internal/core/analysis_api/analysis/rule_engine.go
@@ -20,10 +20,10 @@ package analysis
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


nhakmiller (approved):
LGTM, did you check with someone on the frontend that changing this response format is ok? And did you do a test deploy to a dev environment to verify?

nhakmiller:
You also need the `sync-to-release` label.

wey-chiang:
> LGTM, did you check with someone on the frontend that changing this response format is ok? And did you do a test deploy to a dev environment to verify?

From @vorillaz:
> The graphql is expecting that the output is always a string

The changes take the list and concatenates all of the strings with a comma separator -- so the end result is compatible with exactly what is expected.

dest-override-works!

internal/core/analysis_api/analysis/rule_engine.go
```diff
@@ -135,6 +137,22 @@ func buildTestSubRecord(output, error string) *models.TestDetectionSubRecord {
...
 
+func buildTestSubRecordList(output []string, error string) *models.TestDetectionSubRecord {
+	if len(output) == 0 && error == "" {
+		return nil
+	}
+
+	result := &models.TestDetectionSubRecord{}
+
+	if len(output) != 0 {
+		result.Output = aws.String(strings.Join(output, ","))
```
wey-chiang, line 148:
```suggestion
		result.Output = aws.String(strings.Join(output, ", "))
```

austinbyers (approved):

PR #2424, opened 2021-01-15
CircleCI paralllelism for test:web

vorillaz:
## Background

`test:web` was one of the slowest CI checks due to the limitations of Jest sequencer running. I spotted this [PR](https://github.com/facebook/react/pull/19552/files) from the React team and tried out the results on Circle CI with the parallelization in place. Everything works expected in the isolated containers, I guess we can also add some parallel working for `test:go` and `test:python` as well in the near future.

## Some stats.

- Median time for `test:web` without parallelism: **~39s**
- Median time for `test:web` with parallelism: **~20s**

_Stats are coming from the latest 10 open PRs_

## Screenshots.
Screenshot 2021-01-15 at 12 13 27 PM


alexmylonas (commented):
You commited `web/src/graphql/fragments/.PolicyDetailsMain.graphql.icloud` before gitignore 😛

web/__tests__/jestSequencer.js
```diff
@@ -0,0 +1,38 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+const Sequencer = require('@jest/test-sequencer').default;
+
+/* eslint-disable */
+class CustomSequencer extends Sequencer {
+  sort(tests) {
+    if (process.env.CIRCLE_NODE_TOTAL) {
```
3nvi, line 24:
```suggestion
    if (process.env.CI && process.env.CIRCLE_NODE_TOTAL) {
```

just an additional safety net

alexmylonas (approved):
LGTM!

web/__tests__/jestSequencer.js
```diff
@@ -0,0 +1,38 @@
...
+      // In CI, parallelize tests across multiple tasks.
+      const nodeTotal = parseInt(process.env.CIRCLE_NODE_TOTAL, 10);
+      const nodeIndex = parseInt(process.env.CIRCLE_NODE_INDEX, 10);
+      tests = tests
+        .sort((a, b) => (a.path < b.path ? -1 : 1))
+        .filter((_, i) => i % nodeTotal === nodeIndex);
+    }
```
3nvi, line 31: interesting.. Wondering if we can do it for non web tests as well. I know that go has native parallelism, but it's interesting to see if we can do it on a CircleCI executor-level. Cc @austinbyers

web/__tests__/jestSequencer.js
```diff
@@ -0,0 +1,38 @@
+/**
```
3nvi, line 1: can we make it a `.ts` file and add ES6 imports/exports? I think it should work...

3nvi (approved):
Nice call @vorillaz ! Wanted to dig into `parallelism` for a while now.

The good news is that it's even faster! Props! 

The bad news is that we are still bottlenecked by other stuff like `npm run bundlesize` and `mage test:go`.

For the latter, the BE teams should help. For the former, perhaps we can do something to speed it up even more

web/__tests__/jestSequencer.js
vorillaz: CircleCI supports test splitting, I tried using this one with Jest and `jest-junit` with no luck tho.
https://circleci.com/blog/a-guide-to-test-splitting/

austinbyers: > I know that go has native parallelism, but it's interesting to see if we can do it on a CircleCI executor-level.

Worth a try sometime. I have no idea how easy it would be to split up our Go tests or if it'd be any faster since it already has native parallelism, like you say

.gitignore
```diff
@@ -19,6 +19,7 @@ keys/
 
 # MacOS files
 .DS_Store
+.icloud
```
austinbyers, line 22: Previous reviewers have argued this should be in your global git config rather than in the repo. I'm fine with keeping it here just an extra safety measure

austinbyers (approved):
Awesome! Do you know if this will increase the cost significantly? If 3-6 engineers have PRs that are triggering builds at the same time, will we want some kind of parallelism limit?

austinbyers:
@vorillaz @3nvi is this good to merge? It's been open for 20 days

vorillaz:
@austinbyers Since the go tests are serial I can't see a real value for merging this one at the moment. We can definitely get lots of computing power in the future. Our plan supports parallelism and we could speed up closing open PRs or so. Initially, we can even limit this one for nightly or canary runs but one way or another there should be some work from the BE team involved.

3nvi:
@vorillaz We should either close this or merge this. Choose whatever you see fit, since I can personally understand both choices

austinbyers:
> Since the go tests are serial I can't see a real value for merging this one at the moment

Go tests are not serial - they run in parallel (just in a single CI container instead of multiple). Would love to look at CI parallelism for it when we get a chance

Just because one test suite is already faster than another, doesn't mean there isn't value in speeding it up! Test times will grow as the codebase grows, and web tests on average are slower than go tests. So taking the time now to keep them fast seems reasonable to me.

But up to you whether to merge or close

vorillaz:
I closed this one as we are not there yet but we can keep this in mind for future reference. Thanks a lot, Austin for the feeback.

vorillaz:
Closing this one in order to prevent flaky ENOEM issues with panther enterprise

PR #2429, opened 2021-01-18
Modify graph axis units to conveniently display latency values

### LINKED ISSUE #1828: Log Analysis Overview - UX - Graph axis units
### Describe the bug

On the data latency graph, very large values are displayed as seconds. 10,000,000,000 seconds is not a super useful number, it would be preferable to have this in hours or even days as appropriate.

### Steps to reproduce

1. Go to Log Analysis Overview
1. Select a timeframe on the data latency graph that shows results with very large numbers (in the 1.11  release test environment, there is a good example on October 12th)
1. Observe massive scale on the y-axis

### Ideal behavior

The units would automatically adjust to larger units to conveniently display the value.

### Environment

- Panther version 1.11 release candidate

### Screenshots

Screen Shot 2020-10-21 at 9 44 03 AM

kouknick:
## Description
Format graph axis units on the data latency graph. Latency units now automatically adjust to conveniently display larger values.

Closes #1828  

### Screenshots:
image
image

### Checklist:


web/src/components/charts/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -128,7 +144,7 @@ const TimeSeriesChart: React.FC<TimeSeriesChartProps> = ({
   chartType = 'line',
   hideLegend = false,
   hideSeriesLabels = true,
-  units,
+  useTimeSpanUnits,
```
alexmylonas, line 147: Although we only use `units="sec"` in OSS, we use it on Enterprise for other units.  
Therefore, I am inclining to suggest you revert this change and then check if `units="sec"` inside the Chart and use the `formatTimeSpan`

web/src/components/charts/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -119,6 +119,22 @@ function formatDateString(timestamp: Scalars['AWSDateTime'], useUTC: boolean) {
     .toUpperCase()}`;
 }
 
+const formatTimeSpan = (value: number) => {
+  if (value > 60 * 60 * 24 * 30) {
```
alexmylonas, line 123: Could you also add years? From the screenshot it appears we still have a huge a number of months, which I really cant understand how is a valid value.

Extra: ping @nhakmiller to get an understanding of how is that number valid

web/src/components/charts/TimeSeriesChart/ChartTooltip.tsx
```diff
@@ -21,10 +21,13 @@ import { formatDatetime } from 'Helpers/utils';
 
 export interface ChartTooltipProps {
   params: any[];
-  units?: string;
+  /**
+   * The content formatter used to format the series values
+   */
+  formatter?: (value: number) => string;
```
alexmylonas, line 27: Same as below

web/src/components/charts/TimeSeriesChart/TimeSeriesChart.tsx
kouknick: I've totally missed that. Thanks, I'll have that in mind next time!

web/src/components/charts/TimeSeriesChart/TimeSeriesChart.tsx
kouknick: Done!

nhakmiller: This number isn't valid, it's a bug that some log types always show ~292 years latency. It's being worked on.

vorillaz (approved):
Since @alexmylonas says that everything works as expected, he's an echarts aficionado, I would follow.

alexmylonas (approved):
Solid work! 🚢

PR #2443, opened 2021-01-18
[FE] Add outline only during tabbing

3nvi:
## Background

Currently we have an "annoying" effect where the outline gets added to every `<button>` or `<a>` in our app even on simple clicks. That causes an ugly blue wrapper to appear around each clickable element, anytime you click anything that's clickable.

This is something we absolutely need since it plays a huge part in making Panther accessible, but it should only show when user is navigating with the arrows or the tabs and not when simple clicks occur.

In this PR, I borrowed some nice logic from Youtube and made the outline only appear when you are tabbing, remaining hidden during simple clicks. This way the app remains both "pretty" and accessible when needed.

## Changes

- Add a hook that detects user tabbing activity and toggle a class
- Add a global CSS rule to show/hide `outline` on all elements depending on whether the user is tabbing or not
- Add a spacing between nav entries since the outline was overlapping between anchor elements

## Screenshots

Screen Shot 2021-01-18 at 7 09 41 PM

## Testing

- Manually
- `npm run test`


alexmylonas (approved):
LGTM! I haven't personally noticed this behavior everywhere. I can certainly recognize it on Log Analysis Overview.

3nvi:
@vorillaz opinion?

vorillaz (approved):
Overall that good for UI and eye contrast issues but I think we shall revisit the entire approach of "tabbing". I think due to consolidation we shall apply an active state without manipulating the browser's native styling.
But one way or another, the implementation is neat.

PR #2444, opened 2021-01-18
source-api: make UpdateStatus return error if source doesn't exist

giorgosp:
## Background

Make UpdateStatus return an error if integration does not exist. This is needed because the log processor, due to its caching, could try to update the status of an integration that no longer exist. 
Previously the update would succeed, writing an integration with only the status fields

## Changes

- Added a dynamoDB condition to not perform update if integration doesn't exist

## Testing

- integration test


internal/core/source_api/api/update_status.go
```diff
@@ -35,7 +38,13 @@ func (api *API) UpdateStatus(input *models.UpdateStatusInput) error {
 	status := ddb.IntegrationStatus{
 		LastEventReceived: &input.LastEventReceived,
 	}
+
 	err := api.DdbClient.UpdateStatus(input.IntegrationID, status)
+
+	var awsErr awserr.Error
+	if errors.As(err, &awsErr) && awsErr.Code() == dynamodb.ErrCodeConditionalCheckFailedException {
+		return &genericapi.DoesNotExistError{Message: "The source integration does not exist"}
+	}
```
kostaspap, line 47: Should be changed to :
```
if awsutils.IsAnyError(err, dynamodb.ErrCodeConditionalCheckFailedException){
return ...
}
```

We have to write such checks so many times through the codebase, we created a utility function for it ^^

internal/core/source_api/api/update_status_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/panther-labs/panther/api/lambda/source/models"
+	"github.com/panther-labs/panther/internal/core/source_api/ddb"
+	"github.com/panther-labs/panther/pkg/testutils"
+)
+
+func TestAPI_UpdateStatus_FailsIfIntegrationNotExists(t *testing.T) {
+	testutils.IntegrationTest(t) // test runs the API handler locally but hits a real DynamoDB
+
+	awsSession := session.Must(session.NewSession())
```
kostaspap, line 36: This is meant to be an integration test, right? Add it here: https://github.com/panther-labs/panther/blob/master/internal/core/source_api/main/integration_test.go

giorgosp: I believe it's better to have integration tests next to the file they test. This way we know what is tested and what not. A huge file of integration tests isn't very readable.

kostaspap: That's the practice we have been following throughout the code though. Unit tests for `file.go` are placed inside `file_test.go` and integration tests are indexed/named and placed inside an `integration_test.go` that contains all the integration tests for the package or Lambda. 

I am not against changing that pattern however: 
1. We need to find a way to group integration tests together either by putting them in a special folder or giving them a special name etc. This file is named `update_status_test.go` so we would usually expect to find unit tests in here
2. Everyone should start applying it everywhere in the code moving forward

If someone opens a PR  tomorrow and it includes an integration tests, they should follow this pattern.

giorgosp: Also I would expect tests in the `main` package to be e2e tests, i.e. to run `main()` instead of some other struct/function

kostaspap (changes requested):
Btw, we should update the log processor handler as well. We don't want log processing to fail if this method returns such an error

internal/core/source_api/api/update_status_test.go
giorgosp: >We need to find a way to group integration tests together either by putting them in a special folder or giving them a special name etc. 

I believe `testing.IntegrationTest(t)` is enough, not need for extra grouping or folder strucutre. Golang doesn't also have separate files for tests and benchmarks or suggesting to split unit and integration tests. 

>Everyone should start applying it everywhere in the code moving forward
Yes we can help during PR reviews. Tagging also @austinbyers and @nhakmiller here who have worked a lot on the source-api

internal/core/source_api/api/update_status.go
giorgosp: Much cleaner, updated

rleighton:
NOTE: I believe all integration tests need to start with “TestIntegration”
for mage test:integration to work. Check that

internal/core/source_api/api/update_status_test.go
giorgosp: > NOTE: I believe all integration tests need to start with “TestIntegration” for mage test:integration to work. Check that

Thanks Russ, you are right. I moved it back to `integration_test.go` as Kostas suggested and prefixed with `TestIntegration`.

giorgosp:
> Btw, we should update the log processor handler as well. We don't want log processing to fail if this method returns such an error

The `updateIntegrationStatus()` in the log processor just logs the error if it is returned. The problem was in `sourceCache.Update()`. Added a `recover()` to recover from `S3Info()` panics and skip problematic sources.

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -120,11 +121,18 @@ func (c *sourceCache) Update(now time.Time, sources []*models.SourceIntegration)
...
-		for _, prefix := range prefixes {
-			byBucket[bucket] = append(byBucket[bucket], prefixSource{prefix: prefix, source: source})
-		}
-		index[source.IntegrationID] = source
+		func() {
+			defer func() {
+				if r := recover(); r != nil {
+					zap.L().Error(fmt.Sprintf("recovered from panic: %s", r))
+				}
+			}()
```
kostaspap, line 129: The code is calling `S3Info()` on an item returned by the API. Putting the code to do panic recover at such a low level in the code seems a bit ... odd imo. We could validate the result returned from the sources-api Lambda, but it seems that it should be responsibility of the sources-api to always return valid result or error. 

Personally I think it would be fine to leave this out. 

My comment in the previous PR revision was about handling the `genericapi.DoesNotExistError` returned by the sources-api and not failing processing if we encounter it, but as you said we already take care of it and just log a warning.

giorgosp: Ok reverted it

kostaspap (approved):

PR #2446, opened 2021-01-19
Alert routing by type

### LINKED ISSUE #2399: BE Implementation of Alert Routing by Type in Destinations page
### Description

As an analyst, I want to set default alert routing by the alert type (rule match, rule error, etc). Today, this routing is only based on a detection severity (info, low, etc), but it's increasingly common to have destinations configured only for specific purposes, such as handling compliance-based alerts (e.g., my S3 bucket is insecure)

Example user story:
- I want to set up a specific Slack channel as the default alert destination for Rule Errors
- When adding a new Slack destination, I want to see a dropdown of all the alert types and indicate which alert types should be defaulted for Slack

### Related Services

Destinations Page

### Designs

[Designs are here](https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/0d293ff7-a6c6-4ddc-9dd0-442c63b91995/collections/4c465cf2-53df-4aab-830a-f57ad4e3a7f5)

### Acceptance Criteria

- In add/edit alert destination page, the user should be able to define default alert types for the destination
- The user should be able to select from rule matches, rule errors, policy failures, or scheduled rule matches/errors (once available)
- all 3 alert types should be selected by default
- at least one alert type should be selected

s0l0ist:
## Background

This PR adds BE support for delivering alerts to destinations depending on the alert's type (RULE, RULE_ERROR, or POLICY). Closes #2399.

The BE allows the `AlertType` to be empty for backwards compatibility when adding/editing a destination. However, the destinations/ouputs API will always populate the empty `AlertType` list with all the supported alert types.

## Changes

- Adds a new field to several structs called `AlertTypes`.

## Testing

- `mage test:go`
- manually creating a detection (rule/policy) and setting the destination type setting appropriately.


austinbyers (approved):
LGTM, but I'm not an expert on the alerting flow

internal/core/alert_delivery/api/dispatch_alerts.go
```diff
@@ -89,11 +89,19 @@ func getAlertOutputMap(alerts []*deliveryModels.Alert) (AlertOutputMap, error) {
...
 		if err != nil {
 			return alertOutputMap, errors.Wrapf(err, "Failed to fetch outputIds")
 		}
-		uniqueOutputs := getUniqueOutputs(validOutputIds)
+
+		// Next, we filter out any outputs that don't match the Alert Type setting
+		filteredOutputs := filterOutputsByAlertType(alert, outputs)
+
+		// Then, we obtain a list of unique outputs
+		uniqueOutputs := getUniqueOutputs(filteredOutputs)
```
kostaspap, line 102: nitpick: It would be better if the `getAlertOutputs` would just make sure to return unique elements - reading through the code is seems like unintended behavior that it could return duplicate results. 
That way, we don't have to remember to call `getUniqueOutputs` here.

internal/core/alert_delivery/api/outputs.go
```diff
@@ -46,18 +46,18 @@ func getAlertOutputs(alert *deliveryModels.Alert) ([]*outputModels.AlertOutput,
...
-	if alertOutputs, ok := getDestinationOverrides(alert, outputs); ok {
+	// Then, use any destination overrides specified (set in the detection's form)
+	if alertOutputs, ok := getOutputsByDestinationOverrides(alert, outputs); ok {
 		return alertOutputs, nil
 	}
 
-	// Finally, the use the severity rating (default)
-	return getDefaultOutputs(alert, outputs), nil
+	// If no other dynamic/overrides were set, we calculate based on the severity rating (default)
+	return getOutputsBySeverity(alert, outputs), nil
```
kostaspap, line 60: See above - I think it's reasonable that this method will just return unique alert outputs.

internal/core/alert_delivery/api/outputs.go
```diff
@@ -128,6 +132,21 @@ func getDefaultOutputs(alert *deliveryModels.Alert, outputs []*outputModels.Aler
 	return alertOutputs
 }
 
+// filterOutputsByAlertType - returns a list of outputs that match the type specified in the alert.
+// Note: the output.AlertTypes field will always contain a list of at least 1 type as it is backfilled in
+// the outputs API.
+func filterOutputsByAlertType(alert *deliveryModels.Alert, outputs []*outputModels.AlertOutput) []*outputModels.AlertOutput {
+	alertOutputs := []*outputModels.AlertOutput{}
```
kostaspap, line 139: This code is invoked twice, inside the dispatch_alert.go and inside the deliver_alert.go. 

Looking at the code, it seems that `DeliverAlert` and `DispatchAlerts` are nearly identical - the only difference is that `DispatchAlerts` accepts a slice of `deliveryModels.DeliverAlertInput` while `DeliverAlert` just one. 
Other than that, the two methods look identical. Do we need two separate methods, why isn't `DispatchAlert` just invoking `DeliverAlert` in a loop? Seems like a lot of code duplication ^^

internal/core/outputs_api/api/get_outputs_test.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 
+	deliveryModels "github.com/panther-labs/panther/api/lambda/delivery/models"
```
kostaspap, line 29: nit: better `deliverymodels`. Since package names in Go must be lowercase, I think better stick to this convention for aliases.

internal/core/outputs_api/api/utils.go
```diff
@@ -106,10 +109,17 @@ func redactOutput(outputConfig *models.OutputConfig) {
 	}
 }
 
-func configureOutputFallbacks(outputConfig *models.OutputConfig) {
-	if outputConfig.Opsgenie != nil {
-		if outputConfig.Opsgenie.ServiceRegion == "" {
-			outputConfig.Opsgenie.ServiceRegion = outputs.OpsgenieServiceRegionUS
+// configureOutputFallbacks - function used to ensure backwards compatibility by backfilling previously missing fields
+func configureOutputFallbacks(alertOutput *models.AlertOutput) {
```
kostaspap, line 113: Could you add a TODO here? This is all tech debt that should be removed once we get migration code in place.

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -196,32 +196,64 @@ func getAlertOutputMapping(alert *deliveryModels.Alert, outputIds []string) (Ale
 		return alertOutputMap, errors.Wrapf(err, "Failed to fetch outputIds")
 	}
 
-	// Check the provided the input outputIds and generate a list of valid outputs
+	// Check the provided the input outputIds and generate a list of valid outputs. Ends up getting a unique list as well!
```
kostaspap, line 199: Comment goes for line 191, it is out of the scope of this PR but it seems this should be fixed as part of (another) PR.

Do we really need an outputs cache? Looking at the code it seems we shouldn't. We need fetch the outputs once per `DispatchAlerts(ctx context.Context, input []*deliveryModels.DispatchAlertsInput)` invocation, which shouldn't be that frequent. We actually do that already, since we invalidate the cache at line 191. 

Removing the cache will cleanup the code A LOT imo.

internal/core/alert_delivery/api/outputs.go
```diff
@@ -128,6 +132,21 @@ func getDefaultOutputs(alert *deliveryModels.Alert, outputs []*outputModels.Aler
 	return alertOutputs
 }
 
+// filterOutputsByAlertType - returns a list of outputs that match the type specified in the alert.
+// Note: the output.AlertTypes field will always contain a list of at least 1 type as it is backfilled in
+// the outputs API.
```
kostaspap, line 137: Is the note referring to the return slice or the `outputs` argument?

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -196,32 +196,64 @@ func getAlertOutputMapping(alert *deliveryModels.Alert, outputIds []string) (Ale
...
+
+// difference - Finds the difference O(M + N) of outputs: A - B
+func difference(a, b []*outputModels.AlertOutput) []*outputModels.AlertOutput {
+	m := make(map[string]struct{})
+
+	for _, item := range a {
+		m[*item.OutputID] = struct{}{}
+	}
+
+	res := make([]*outputModels.AlertOutput, 0, len(b))
```
kostaspap, line 252: nit: Just `var res []*outputModels.AlertOutput` or `make([]*outputModels.AlertOutput, 0)`? Preallocating `len(b)` actually seems suboptimal here, given the result is going to be `A-B`

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -196,32 +196,64 @@ func getAlertOutputMapping(alert *deliveryModels.Alert, outputIds []string) (Ale
...
 	if len(validOutputIds) == 0 {
 		return alertOutputMap, &genericapi.InvalidInputError{
 			Message: "Invalid destination(s) specified: " + strings.Join(outputIds, ", ")}
 	}
 
+	// Next, we filter out any outputs that don't match the Alert Type setting
+	filteredOutputs := filterOutputsByAlertType(alert, validOutputIds)
+
+	// If there's a difference, return an error message with the IDs that failed
+	diffOutputs := difference(validOutputIds, filteredOutputs)
```
kostaspap, line 210: I am not sure I understand this. Why is this check needed at runtime? The only reason why this can happen, from what I understand, is if our code has a bug, correct? If that's the case, this check should be removed and we should add a unit test to verify the logic.

kostaspap:
Nice work! This PR included changes in 43 files, but reading it was really smooth :)

internal/core/alert_delivery/api/outputs.go
s0l0ist: Yes, the difference is in how the two are invoked:

DeliverAlert - is used when the user manually invokes the resending of an alert (and also sending test alerts)
DispatchAlerts - is triggered via SQS (when the detection engine puts alerts there)

internal/core/alert_delivery/api/deliver_alert.go
s0l0ist: If we were building this from ground zero, I wouldn't have put a cache; however, the optimization has already been done and it is working -- I'm hesitant to take it out now.

The only case when the cache is used is when `DispatchAlerts` is called. `DeliverAlert` is invoked manually by the user and therefore invalidates the cache to ensure it always gets the latest outputs.

kostaspap (approved):
Approving. Some open comments are meant to be addressed in future PRs (if even), rest of comments are minor.

internal/core/alert_delivery/api/dispatch_alerts.go
s0l0ist: Good callout - I've moved this extra logic all into `getAlertOutputs`.

internal/core/outputs_api/api/get_outputs_test.go
s0l0ist: There was a lot of inconsistencies with the module name. I've fixed all of them and added to this PR.

internal/core/alert_delivery/api/deliver_alert.go
s0l0ist: I've taken out the comment change here as a part of the other fix where the `getAlertOutputs` is expected to return unique items.

internal/core/alert_delivery/api/outputs.go
s0l0ist: Moved it to the proper location in the function.

internal/core/alert_delivery/api/deliver_alert.go
s0l0ist: Actually, this can happen for a very important reason. Specifically, if an alert's type (say `RuleErrorType`) and the output is configured to accept `RuleType` and `PolicyType`, then the alert will silently be dropped and not delivered to the output. 

Subsequently, in the delivery history in the Panther UI will not show any entries because it was dropped from not matching the types supported in the destination. However, a user may **not** understand why this is happening and will try to re-send the alert. In that case, we should not return an idempotent successful status - we should calculate if the user's request will differ from what is computed (if the alert will actually go out to the specified destination).

Here, we make this calculation and return an error to the user if the alert will fail to be delivered because the type doesn't match what the destination allows.

internal/core/alert_delivery/api/deliver_alert.go
s0l0ist: yep good catch.

internal/core/alert_delivery/api/deliver_alert.go
```diff
@@ -196,32 +196,64 @@ func getAlertOutputMapping(alert *deliveryModels.Alert, outputIds []string) (Ale
...
+		m[*item.OutputID] = struct{}{}
+	}
+
+	res := make([]*outputModels.AlertOutput, 0, len(b))
+	for _, item := range b {
+		if _, ok := m[*item.OutputID]; !ok {
+			res = append(res, item)
+		}
+	}
+	return res
```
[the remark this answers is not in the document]
s0l0ist: Already had a test for `intersection`, added one for `difference`

internal/core/alert_delivery/api/deliver_alert.go
kostaspap: Interesting, ok now I see. Catching the edge cases there :) 

It makes sense. I know this PR has been already merged, but this exact comment would be very useful if it was inside the code. It will make it easier for someone reading the code to understand the extra context and why this is required and why it should stay there. Maybe you can add this in next PR for this codebase!

PR #2457, opened 2021-01-20
test: mock debounce where needed and changed all waitMs to 1ms

alexmylonas:
## Background

In FE, we widely use `waitMs` for wait for effects to get in place,  that wait time doesn't need to be more than 1ms except for cases where we want to test the formik auto-save functionality.

**Result**: shaved 10-20 sec from test suite

## Changes

- Changed all `waitMs(xx)` to `waitMs(1)` (except formRestoration tests)

## Testing

- `npm run test` locally


web/src/components/utils/FormSessionRestoration/FormSessionRestoration.test.tsx
```diff
@@ -55,7 +55,7 @@ test('correctly stores form values to session', async () => {
   expect(textInput).toHaveValue(testValue);
 
   // wait for debounce to kick in
-  await waitMs(300);
```
3nvi, line 58: we should mock debounce here as well

3nvi (approved):
Looks good @alexmylonas . One more debounce fix and then just make sure you run the tests at least 10 times to make sure that the `waitMs(1)` doesn't create issues.

If 10 runs are all successful, go ahead and merge this

web/src/components/utils/FormSessionRestoration/FormSessionRestoration.test.tsx
alexmylonas: done!

vorillaz (approved):
Nicely done, that's somehow a side effect though.
How about mass replacing every `await waitMs(1);` with a tailormade test helper function?
Essentially `waitMs` is arbitrary and should not be used.

alexmylonas:
> Nicely done, that's somehow a side effect though.
> How about mass replacing every `await waitMs(1);` with a tailormade test helper function?
> Essentially `waitMs` is arbitrary and should not be used.

You mean something like:
```
function wait() {
   return waitMs(1)
}
```

vorillaz:
Yeah, something like this one.
Eventually, we should replace `waitMs` in most cases with `waitForElementToBeRemoved` or similar visual guards.

PR #2461, opened 2021-01-20
Improve rule engine efficiency by using MaximumBatchingWindowInSeconds

rleighton:
## Background

The rule engine uses SQS to dispatch to the lambda. Often this results in multiple lambdas running for events close in time. We would rather have these events processed by 1 lambda because:

 - there are less lambda calls (costs less)
 - it improves file aggregation to process more than single events.

This PR adds the use of MaximumBatchingWindowInSeconds set to 30 seconds to the SQS configuration.
This changes the behavior of SQS dispatch to wait 30sec OR until there are 10 events (files) to process.

During testing of this I ran a  load test of 10 files / sec into master for a bit, that's the first bump under Rules Engine  Invocations in the graph below  ... then I applied the patch below and ran again ... invocations dropped more than 50% with corresponding longer duration with no obvious change in memory usage. This should be both cheaper and provide better aggregation of output files. 

image

## Changes

- Add `MaximumBatchingWindowInSeconds: 30`

## Testing

- mage deploy + load test


kostaspap (approved):
Small change, great benefits. Love it.

PR #2465, opened 2021-01-20
Update pounce version

kouknick:
## Background

Update pounce to the latest version to fix the inconsistent dates in the date range picker.

## Changes

Update pounce to v0.80.1.

## Testing

- npm run test
- Manually tested date range picker


vorillaz (approved):
Solid, 💪

PR #2468, opened 2021-01-20
Verify Cloud-Security and Log-Types for policies, roles, and datamodels on create or modify

tatehanawalt:
## Background

Bulk upload or invoke creation of policies, roles, and datamodels doesn't reject the invokation when the create/update object has invalid log types or resource types. 

This change adds a validation check on log types and resource types before create or modify policies, roles, and datamodels.

## Changes

Validation methods for resource type and log type sets are added to the analysis_api/handlers/clients.go

Resource types are a constant set - internal/compliance/snapshot_poller/models/aws/resourceTypes.go

Methods to update the sets of logs/resource types are also added to the clients file.

One of the update methods queries the logtypes api

A significant portion of the changes are files added for integration testing

## Testing

ci
integration with added tests and testing resources
Manual lambda invoke


internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -391,7 +391,12 @@ func validateUploadedPolicy(item *tableItem) error {
 		item.Severity = compliancemodels.SeverityInfo
 	case models.TypeDataModel:
 		item.Severity = compliancemodels.SeverityInfo
-	case models.TypePolicy, models.TypeRule:
+	case models.TypePolicy:
+		if err := ValidResourceTypeSet(item.ResourceTypes); err != nil {
+			return err
+		}
+		break
```
austinbyers, line 398: You don't need `break`s in Go switch statements. This one can be removed; the one on the next line is fine because there's no other code

internal/core/analysis_api/handlers/create_update_policy.go
```diff
@@ -30,6 +31,26 @@ import (
 	"github.com/panther-labs/panther/pkg/gatewayapi"
 )
 
+var ValidResourceTypes = []string{"AWS.ACM.Certificate", "AWS.CloudFormation.Stack", "AWS.CloudTrail", "AWS.CloudTrail.Meta", "AWS.CloudWatch.LogGroup", "AWS.Config.Recorder", "AWS.Config.Recorder.Meta", "AWS.DynamoDB.Table", "AWS.EC2.AMI", "AWS.EC2.Instance", "AWS.EC2.NetworkACL", "AWS.EC2.SecurityGroup", "AWS.EC2.Volume", "AWS.EC2.VPC", "AWS.ECS.Cluster", "AWS.EKS.Cluster", "AWS.ELBV2.ApplicationLoadBalancer", "AWS.GuardDuty.Detector", "AWS.IAM.Group", "AWS.IAM.Policy", "AWS.IAM.Role", "AWS.IAM.RootUser", "AWS.IAM.User", "AWS.KMS.Key", "AWS.Lambda.Function", "AWS.PasswordPolicy", "AWS.RDS.Instance", "AWS.Redshift.Cluster", "AWS.S3.Bucket", "AWS.WAF.Regional.WebACL", "AWS.WAF.WebACL"}
```
austinbyers, line 34: This should be a map for more efficient inclusion testing:

```go
var ValidResourceTypes = map[string]struct{}{
    "AWS.ACM.Certificate":      struct{}{},
    "AWS.CloudFormation.Stack": struct{}{},
    // ...
}
```

The `map[string]struct{}` is marginally more efficient than a `map[string]bool`, it's a common Go idiom for a "stringset". 

Then, to test inclusion:

```go
if _, exists := ValidResourceTypes[type]; exists {
    // found a type match
}
```

internal/core/analysis_api/handlers/create_update_policy.go
```diff
@@ -30,6 +31,26 @@ import (
...
+
+func ValidResourceType(checkValidResourceType string) bool {
+	for _, existingResourceTypeEntry := range ValidResourceTypes {
+		if checkValidResourceType == existingResourceTypeEntry {
+			return true
+		}
+	}
+	return false
+}
```
austinbyers, line 43: This function will not be needed once you convert to a map above

austinbyers (changes requested):
Be sure to run the associated integration tests and make sure they pass. Then, add an additional integration test (or update an existing one) to test for this case

austinbyers:
Tagging @nhakmiller as the owner of this service - do you see any other easy solution besides just hardcoding these resource types? I figure they don't change very often, we can add some autogeneration code later

internal/core/analysis_api/handlers/create_update_policy.go
austinbyers, line 34: Can you add a comment here and in the web resource constants file that point to each other? E.g.

```go
// These constants are also defined in web/src/....
// Be sure that any changes to this list are reflected in the web source and vice versa
```

internal/core/analysis_api/handlers/bulk_upload.go
tatehanawalt: fixed - thank you

nhakmiller:
> Tagging @nhakmiller as the owner of cloud security - do you see any other easy solution besides just hardcoding these resource types? I figure they don't change very often, we can add some autogeneration code later

Hmm it's not perfect, but `internal/compliance/snapshot_poller/pollers/aws/poll.go` has a map called `ServicePollers` that has a key for every resource type we support. That list has to be manually updated whenever a new resource type is added, but if it's not updated scans of that resource type will crash hard so it usually doesn't get missed. So you could reference that map directly, or add some function to the snapshot-pollers that lets you reference it more cleanly like `ResourceTypeExists` or something that does the lookup. Problem is then you have to import the poller code, which is massive because it depends on a lot of different AWS service SDKs.

Alternatively as you say we can auto generate it. It would be based off the files located in `internal/compliance/snapshot_poller/models/aws/`.

Those two ways were be more guaranteed to be correct, but also way harder than just hardcoding. They do change fairly infrequently, but if we enforce it like this we probably want to add a comment in `poll.go` or something that says "oh yeah also update `internal/core/analysis_api/handlers/create_update_policy.go` whenever a new resource type is added".

austinbyers:
> Problem is then you have to import the poller code, which is massive because it depends on a lot of different AWS service SDKs.

Yeah - the snapshot-poller is currently 9.7 MB compared to analysis-api 5.3 MB. Doubling the size is probably not ideal.

Another idea I hadn't initially considered: the snapshot-poller could respond to an API request for "listResourceTypes." That way, the source of truth stays in the snapshot poller, and we don't have to import its codebase. But then every policy update requires another cross-lambda api call (in addition to the policy-engine validation)

----

Since the constants in the web source are already hardcoded, I'd vote for just taking the easy path for now and hardcoding it here as well. Not worth our time to optimize this just yet. Like you say, let's just add comments everywhere we need these  so we remember to update all of the files together.

We could even add a CI check for it - make sure the web constants, the analysis-api validation, and the snapshot poller map all reference the same resource types

tatehanawalt:
@nhakmiller for simplicity my plan was just to have the simple static set. I'm still using a method to provide the set where it's used for this issue so it will be easy to update in the future.

internal/core/analysis_api/handlers/create_update_policy.go
```diff
@@ -30,6 +31,50 @@ import (
 	"github.com/panther-labs/panther/pkg/gatewayapi"
 )
 
+var ValidResourceTypes = map[string]struct{}{
```
kostaspap, line 34: The snapshot pollers maintain a map of supported ResourceTypes -> Service Name [here](https://github.com/panther-labs/panther/blob/master/internal/compliance/snapshot_poller/pollers/aws/clients.go#L79-L127)

I think it would be better to move that map to a common location, e.g. `pkg/awsresources` or something like that. Then the code here can reference it verify if a provided ResourceType is supported. With this approach we minimize the chance of forgetting to add some resource to the list here.

kostaspap: I just saw the comments write above after putting my own comment. For me, the above solution seems really straightforward and future proof. It might be that it adds some extra AWS dependencies, but I am not sure this would have a measurable impact.

austinbyers: > The snapshot pollers maintain a map of supported ResourceTypes -> Service Name here

Ah, I was only aware of the map from type name to the actual function handlers, which I believe is what Nick was referring to as well.

Yes, it looks like the map(s) you link to could work well. They don't even have to move to the `pkg/` folder - they are still specific to cloud security. They could be `internal/core/compliance/snapshot_poller/awsresources` or something

tatehanawalt:
Merged valid analysis and resource, Ready to start review. I will be completing more details in the morning

internal/core/analysis_api/handlers/create_update_policy.go
austinbyers: It turns out there is no existing map in the cloud security source we can leverage: they don't have the meta types. Hardcoding the types is the only feasible option at the moment

api/graphql/schema.graphql
```diff
@@ -1027,22 +1027,16 @@ input UpdateUserInput {
...
-  modifiedRules: Int!
-
-  totalGlobals: Int!
-  newGlobals: Int!
-  modifiedGlobals: Int!
-
-  totalDataModels: Int!
-  newDataModels: Int!
-  modifiedDataModels: Int!
+type UploadPoliciesResponse {
```
austinbyers, line 1030: I'm guessing you didn't intend to change the graphQL, it was just a bad merge.

Just copy the version of this file from the `master` branch into your branch, then run `mage gen fmt` and that should fix most of the stray web changes

internal/compliance/aws_event_processor/processor/elbv2.go
```diff
@@ -87,16 +87,6 @@ func classifyELBV2(detail gjson.Result, metadata *CloudTrailMetadata) []*resourc
...
-		// If no LB ARN is present, this may be a classic load balancer. We don't support classic
-		// load balancers. We can tell the difference based on the structure of the request parameters
-		if parseErr != nil {
-			lbName := detail.Get("requestParameters.loadBalancerName")
-			if lbName.Exists() {
-				return nil
-			}
-			zap.L().Error("elbv2: error parsing ARN", zap.String("eventName", metadata.eventName), zap.Any("event", metadata), zap.Error(parseErr))
-			return nil
-		}
```
austinbyers, line 99: Another bad merge? It looks like you did a `git pull` but ignored all upstream changes, I know this was a recent change

internal/compliance/snapshot_poller/models/aws/resourceTypes.go
```diff
@@ -0,0 +1,65 @@
...
+// or modifications to this data need to coincide with updates in the other places where this data
+// is hardcoded.
+//
+// Locations may not be in this list! right now this data is hardcoded in
+// • internal/compliance/snapshot_poller/models/aws/ResourceTypes.go
+// • internal/compliance/snapshot_poller/pollers/aws/clients.go
+//
+// • web/src/constants.ts
+//
+var ResourceTypes = map[string]struct{}{
```
austinbyers, line 35: I don't see the `meta` schemas in here, be sure the length of this list matches the `web/src/constants.ts` / UI version

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -91,23 +104,21 @@ var (
 		awsmodels.Ec2VpcSchema:              ec2.ServiceName,
 		awsmodels.EcsClusterSchema:          ecs.ServiceName,
 		awsmodels.EksClusterSchema:          eks.ServiceName,
-		// For every other service, the service name aligns with how SSM refers to the service. For
-		// just the elb and elbv2 service, this is not the case. AWS just had to do it to 'em.
```
austinbyers, line 95: Probably need to keep this comment

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -374,6 +373,13 @@ func validateUploadedDataModel(item *tableItem) error {
 	if len(item.ResourceTypes) > 1 {
 		return errors.New("only one LogType may be specified per DataModel")
 	}
+
+	// We could call the refresh before we traverse the set of new rules. The code you see here
+	// refeshes the log types on each validateLogtypeSet call...
+	if err := validateLogtypeSet(item.ResourceTypes); err != nil {
+		return fmt.Errorf("DataModel contains invalid log type: %s", err.Error())
```
austinbyers, line 380: Use `errors.Errorf` to be consistent with the error handling in this service. The [`errors`](https://godoc.org/github.com/pkg/errors) package provides a nicer way to wrap errors and its stack trace - we try to use it for new code

internal/core/analysis_api/handlers/clients.go
```diff
@@ -74,4 +95,93 @@ func Setup() {
...
+		LambdaName: logtypesapi.LambdaName,
+		LambdaAPI:  lambdaLogTypesClient,
+	}
+
+	refreshLogTypes()
+}
+
+/*
+// This is about to get replaced
+var ValidResourceTypes = map[string]struct{}{
```
austinbyers, line 116: leftover comment I assume

requirements.txt
```diff
@@ -2,17 +2,17 @@ arrow==0.17.0
 astroid==2.4.2
 attrs==20.3.0
 aws-lambda-builders==1.1.0
-aws-sam-cli==1.15.0
-aws-sam-translator==1.33.0
-bandit==1.7.0
+aws-sam-cli==1.11.0
+aws-sam-translator==1.30.1
+bandit==1.6.2
```
austinbyers, line 7: another bad merge

internal/core/analysis_api/main/integration_test.go
```diff
@@ -262,6 +274,13 @@ func TestIntegrationAPI(t *testing.T) {
 		t.Run("BulkUploadInvalid", bulkUploadInvalid)
 		t.Run("BulkUploadSuccess", bulkUploadSuccess)
 	})
+
+	t.Run("BulkUploadInvalid", func(t *testing.T) {
+		t.Run("BulkUploadInvalidPolicyResourceTypesFail", bulkUploadInvalidPolicyResourceTypesFail)
+		t.Run("BulkUploadInvalidRuleLogtypeFail", bulkUploadInvalidRuleLogtypeFail)
+		t.Run("BulkUploadInvalidDatamodelLogtypeFail", bulkUploadInvalidDatamodelLogtypeFail)
+	})
```
austinbyers, line 282: 🎉  Love it

internal/core/analysis_api/main/integration_test.go
```diff
@@ -2611,3 +2781,28 @@ func batchDeleteRules(t *testing.T, ruleID ...string) {
...
 }
+
+// Utility Methods:
+
+// Bulk Data upload helper Zips files in the fsSourcePath to ./<basename(fsSourcePath)>.zip
+// then attempts to upload them to the analysis service
+// BulkZipUploadHelper returns an error because some tests are intended to fail. Acceptable Errors
+// are never expected when managing the files so the test will fail if the error originates from
+// anything besides invoking the service client upload invokation.
+func BulkZipUploadHelper(t *testing.T, fsSourcePath string) (*models.BulkUploadOutput, int, error) {
```
austinbyers, line 2792: 👍

austinbyers (changes requested):
Above and beyond with the integration tests, nice job!

Not sure how it happened, but you've got a bad merge somehow. Either undo every change that you don't intend to include in the PR, or just start over on a new branch and copy over the changes you care about.

You may be able to use `git rebase -i HEAD~10` to throw out bad merge commits, but honestly that's probably more trouble than it's worth :) I usually start over on a new branch when I run into weirdness https://xkcd.com/1597

internal/core/analysis_api/main/integration_test.go
tatehanawalt: I was actually going to ask do we have anything against adding an @author tag to method descriptions? Do we avoid doing that because git-blame exists?

internal/compliance/aws_event_processor/processor/elbv2.go
tatehanawalt: fixed with `git checkout master internal/compliance/aws_event_processor/processor/elbv2.go`

I'm thinking all of my git issues are relating to not setting an upstream. I was avoiding the -u flag  to be explicit which is not working as I intended.

api/graphql/schema.graphql
tatehanawalt: yet another one...

Still getting an error after the fix. I ran:

```
git checkout master api/graphql/schema.graphql
mage clean
mage setup
mage gen fmt
```

But I'm getting a failed gql generation.

Running `npm run graphql-codegen` result error:

```
Error: Unable to find field "uploadPolicies" on type "Mutation"!
```

internal/core/analysis_api/handlers/create_update_policy.go
tatehanawalt: This has been updated to reflect a hardcoded set sourced from the snapshot_pollers src

internal/core/analysis_api/handlers/bulk_upload.go
tatehanawalt: You mean using the github errors package right? Not the [go native errors package](https://golang.org/pkg/errors/#example_New_errorf)

api/graphql/schema.graphql
austinbyers: Revert all of the web changes first, perhaps? Not just the GraphQL schema

tatehanawalt: @austinbyers I did a checked out the entire master/web directory before running this command. still getting the unable to find error

tatehanawalt: Somehow the file `web/src/components/wizards/BulkUploaderWizard/UploadPanel/graphql/uploadPolicies.graphql` made it to my branch but its not in master. file shows its from `7503ee73e (Theodore Vorillas 2020-09-18 19:09:46 +0300`

Removing that file along with the /uploadPolicies.generated.ts fixes the gen

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -161,13 +161,6 @@ var (
 		"RegisterTargets":             {},
 		"DeregisterTargets":           {},
 
-		// These are elb classic events that we don't support but can't differentiate from elbv2
-		// events
-		"RegisterInstancesWithLoadBalancer": {},
-		"ConfigureHealthCheck":              {},
-		"SetLoadBalancerPoliciesOfListener": {},
-		"CreateLoadBalancerPolicy":          {},
-
```
kostaspap, line 170: Why these changes? They seem unrelated to the purpose of this PR?

internal/compliance/snapshot_poller/models/aws/resourceTypes.go
```diff
@@ -0,0 +1,67 @@
+package aws
```
kostaspap, line 1: File names in go are `snake_case` so this should be `resource_types.go`

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -66,7 +65,7 @@ func (API) BulkUpload(input *models.BulkUploadInput) *events.APIGatewayProxyResp
 				if r := recover(); r != nil {
 					zap.L().Error("panicked while processing item",
 						zap.String("id", item.ID), zap.Any("panic", r))
-					results <- writeResult{err: errors.New("panicked goroutine")}
+					results <- writeResult{err: errors.Errorf("panicked goroutine")}
```
kostaspap, line 68: nit: Just `errors.New()`, no need for formatting the string.

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -156,13 +154,13 @@ func extractZipFile(input *models.BulkUploadInput) (map[string]*tableItem, error
 	// Base64-decode
 	content, err := base64.StdEncoding.DecodeString(input.Data)
 	if err != nil {
-		return nil, fmt.Errorf("base64 decoding failed: %s", err)
+		return nil, errors.Errorf("base64 decoding failed: %s", err)
```
kostaspap, line 157: This actually should have been `errors.Wrapf(err, "base64 decoding failed")` to make sure the the returned error has a stack trace. Since you are already updating this code, it think it would be nice to fix this issue ^^

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -372,8 +369,15 @@ func readZipFile(zf *zip.File) ([]byte, error) {
...
 func validateUploadedDataModel(item *tableItem) error {
 	if len(item.ResourceTypes) > 1 {
-		return errors.New("only one LogType may be specified per DataModel")
+		return errors.Errorf("only one LogType may be specified per DataModel")
 	}
+
+	// We could call the refresh before we traverse the set of new rules. The code you see here
+	// refeshes the log types on each validateLogtypeSet call...
+	if err := validateLogtypeSet(item.ResourceTypes); err != nil {
+		return errors.Errorf("DataModel contains invalid log type: %s", err.Error())
```
kostaspap, line 378: See similar comments above. This could be `return errors.Wrap(err, "DataModel contains invalid log types")`

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -388,18 +392,29 @@ func validateUploadedDataModel(item *tableItem) error {
 func validateUploadedPolicy(item *tableItem) error {
 	switch item.Type {
 	case models.TypeGlobal:
+		if len(item.ResourceTypes) > 1 {
+			return errors.Errorf("only one LogType may be specified per DataModel")
```
kostaspap, line 396: Hmmm, i'm not sure this is relevant? This is `TypeGlobal`, not DataModel.

internal/core/analysis_api/handlers/clients.go
```diff
@@ -28,13 +32,22 @@ import (
...
 	"github.com/panther-labs/panther/pkg/gatewayapi"
+
+	// Imports a hardcoded map[string]struct{} Where keys are the set of valid resource types.
+	resourceTypesProvider "github.com/panther-labs/panther/internal/compliance/snapshot_poller/models/aws"
 )
 
 const systemUserID = "00000000-0000-4000-8000-000000000000"
 
+// setting Max Retries to a higher number - we'd like to retry VERY hard before failing.
+const maxRetries = 20
```
kostaspap, line 49: I'm not sure this number make sense. This will cause the code to retry for half an hour.

internal/core/analysis_api/handlers/clients.go
```diff
@@ -46,8 +59,15 @@ var (
 
 	policyEngine analysis.PolicyEngine
 	ruleEngine   analysis.RuleEngine
+
+	//
+	lambdaLogTypesClient *lambda.Lambda
```
kostaspap, line 64: 1. better use the interface `lambdaiface.LambdaAPI`. That way it can be mocked in tests. 
2. I don't think this variable needs to be shared at all?

internal/core/analysis_api/handlers/clients.go
```diff
@@ -74,4 +94,70 @@ func Setup() {
...
+			awsretry.NewConnectionErrRetryer(maxRetries),
+		),
+	)
+	lambdaLogTypesClient = lambda.New(clientsSession)
+	logtypesAPI = &logtypesapi.LogTypesAPILambdaClient{
+		LambdaName: logtypesapi.LambdaName,
+		LambdaAPI:  lambdaLogTypesClient,
+	}
+
+	refreshLogTypes()
```
kostaspap, line 110: Log types get created/delete at runtime. We should fetch the list on every Bulk upload operation, not fetch them only once. The operation is pretty fast, so I wouldn't be too concerned about caching them, etc.

internal/core/analysis_api/handlers/clients.go
```diff
@@ -74,4 +94,70 @@ func Setup() {
...
+}
+
+// Traverse a passed set of resource and return an error if any of them are not found in the current
+// list of valid resource types
+//
+// CAVEAT: This method uses a hardcoded list of existing resource types. If this method is returning
+// unexpected errors the hardcoded list is up to date.
+func ValidResourceTypeSet(checkResourceTypeSet []string) error {
```
kostaspap, line 118: `client.go` doesn't seem to be the right place for this code. `validate.go` maybe?

internal/core/analysis_api/handlers/clients.go
```diff
@@ -46,8 +59,15 @@ var (
 
 	policyEngine analysis.PolicyEngine
 	ruleEngine   analysis.RuleEngine
+
+	//
+	lambdaLogTypesClient *lambda.Lambda
+	logtypesAPI          *logtypesapi.LogTypesAPILambdaClient
+
+	logtypeSetMap map[string]interface{}
```
kostaspap, line 67: nit: We prefer `map[string]struct{}` in such scenarios since it requires less allocations. Allocations don't really matter here, but better change it as a code style best practice.

kostaspap (commented):
Nice job! Mostly style comments, code looks good!

internal/compliance/aws_event_processor/processor/process.go
tatehanawalt: Not sure how this ended up in my commit. Replaced it with the master file.

internal/core/analysis_api/handlers/clients.go
kostaspap: I am assuming these check would run every time someone uploads rules which should be not that frequent.

internal/core/analysis_api/handlers/bulk_upload.go
tatehanawalt: i'm moving my changes to a new branch like 90% of these changes came from a bad merge

internal/core/analysis_api/handlers/clients.go
tatehanawalt: Totally agree. I was expecting this would come up in the pr request so I have all my validation code together to move it where it should go. 

validate.go sounds good to me I will have it there in my updated branch

internal/core/analysis_api/handlers/bulk_upload.go
tatehanawalt: Shouldn't we return the invalid log type? if someone tries to upload bulk data  and the upload fails how would they know how to fix it?

internal/core/analysis_api/handlers/clients.go
tatehanawalt: updated to 5

internal/core/analysis_api/handlers/clients.go
tatehanawalt: I have the logtypes client updated to the lambdaiface.LambdaAPI. 

what do you mean in terms of not being shared? I thought clients was a good place for this because it's a client. Are you saying to move it to the logtypes refresh method and create a new client on each refresh or move it to the validations.go file?

Also Note: I am about to open a new pr on a branch with commits on only the files for this feature.

internal/core/analysis_api/handlers/bulk_upload.go
kostaspap: Ah no, I agree we should. I didn't realize that this is the error shown to the users ^^

`Wrapf` will concatenate the `err` message returned by the `validateLogtypeSet` error with the error message defined here so that information should be included anyway. But maybe that message is not that clear, so feel free to keep this one yeah.

internal/core/analysis_api/handlers/clients.go
kostaspap: Sorry, my comment was not very clear. I've continued the discussion to the new PR.

PR #2470, opened 2021-01-20
Fixes for ELBV2 scanning

nhakmiller:
## Background

This PR fixes one major and three minor bugs related to Cloud Security scanning found during on call.

## Changes

- Add several events to the `aws-event-processor` ignore list that were causing hiccups in the `aws-event-processor`
  - elb `RegisterInstancesWithLoadBalancer`, `ConfigureHealthCheck`, `SetLoadBalancerPoliciesOfListener`, and `CreateLoadBalancerPolicy` are all elb classic events. We don't support elb classic, but are unable to differentiate between elb classic and elbv2 events explicitly. These were causing `warn` messages.
  - lambda `InvokeExecution`. Processing this API call raised errors when trying to extract the account ID. Luckily this event does not indicate to us that we need to scan, as no infrastructure is changing.
- Added special error handling to the elb `ModifyLoadBalancerAttributes` event in the `aws-event-processor`. This is related to the above issue in that we can't differentiate elb classic and elbv2 events. In this case though the event `ModifyLoadBalancerAttributes` may occur for both elb and elbv2 resources, and we can only differentiate between the two by inspecting the format of the request
- This is the big one. We were incorrectly generating the list of available regions when conducting full account ELBV2 scans. This means that the baseline scans (initial scan & daily scan) were skipping all ELBV2 resources. Initially I was concerned this may affect many resource types, but I wrote a quick script to test and it luckily only affects ELBV2 resources.

## Testing

- Unit testing
- Deployed to dev account & manually initiated scans to test the ELBV2 region bug

### DIFF
--- a/internal/compliance/aws_event_processor/processor/elbv2.go
+++ b/internal/compliance/aws_event_processor/processor/elbv2.go
@@ -87,6 +87,16 @@ func classifyELBV2(detail gjson.Result, metadata *CloudTrailMetadata) []*resourc
 		return changes
 	case "CreateListener", "DeleteLoadBalancer", "ModifyLoadBalancerAttributes", "SetIpAddressType", "SetSecurityGroups", "SetSubnets":
 		lbARN, parseErr = arn.Parse(detail.Get("requestParameters.loadBalancerArn").Str)
+		// If no LB ARN is present, this may be a classic load balancer. We don't support classic
+		// load balancers. We can tell the difference based on the structure of the request parameters
+		if parseErr != nil {
+			lbName := detail.Get("requestParameters.loadBalancerName")
+			if lbName.Exists() {
+				return nil
+			}
+			zap.L().Error("elbv2: error parsing ARN", zap.String("eventName", metadata.eventName), zap.Any("event", metadata), zap.Error(parseErr))
+			return nil
+		}
 	case "CreateLoadBalancer":
 		var changes []*resourceChange
 		for _, lb := range detail.Get("responseElements.loadBalancers").Array() {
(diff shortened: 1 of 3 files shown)


internal/compliance/aws_event_processor/processor/elbv2.go
```diff
@@ -129,10 +139,6 @@ func classifyELBV2(detail gjson.Result, metadata *CloudTrailMetadata) []*resourc
 		return nil
 	}
 
-	if parseErr != nil {
-		zap.L().Error("elbv2: error parsing ARN", zap.String("eventName", metadata.eventName), zap.Any("event", metadata), zap.Error(parseErr))
-		return nil
-	}
```
nhakmiller, line 135: The only time `parseErr` was not `nil` was if there was an error on line 89 above. I decided it made the most sense to keep the error handling logic closer to the place the error occurred so I move it up there.

internal/compliance/snapshot_poller/pollers/aws/clients.go
```diff
@@ -92,21 +91,23 @@ var (
...
-		awsmodels.LambdaFunctionSchema:      lambda.ServiceName,
-		awsmodels.PasswordPolicySchema:      iam.ServiceName,
-		awsmodels.RDSInstanceSchema:         rds.ServiceName,
-		awsmodels.RedshiftClusterSchema:     redshift.ServiceName,
-		awsmodels.S3BucketSchema:            s3.ServiceName,
-		awsmodels.WafRegionalWebAclSchema:   waf.ServiceName,
-		awsmodels.WafWebAclSchema:           wafregional.ServiceName,
+		// For every other service, the service name aligns with how SSM refers to the service. For
+		// just the elb and elbv2 service, this is not the case. AWS just had to do it to 'em.
+		awsmodels.Elbv2LoadBalancerSchema: "elb",
```
austinbyers, line 96: 😆

austinbyers (approved):
Love it

PR #2475, opened 2021-01-21
Crowdstrike parsers: Insight special raw & Discover processed events

georgepsarakis:
## Background

Falcon Data Replicator forwards special events (raw & processed) that do not use the convention of defining an event name, thus not being parsed by the [`UnknownEvent` parser](https://github.com/panther-labs/panther/blob/9a217c13af85d8ed328faaf894ac3db708b122be/internal/log_analysis/log_processor/parsers/crowdstrikelogs/unknown.go#L51).

The events are the following:
- Falcon Insight:
  - [`AID_MASTER`](https://developer.crowdstrike.com/crowdstrike/docs/falcon-data-replicator-guide#section-aid-master): Detailed Host and Agent information
  - [`MANAGEDASSETS`](https://developer.crowdstrike.com/crowdstrike/docs/falcon-data-replicator-guide#section-managedassets): Network information of an asset running the sensor.
  - [`NOTMANAGED`](https://developer.crowdstrike.com/crowdstrike/docs/falcon-data-replicator-guide#section-notmanaged): Unmanaged assets discovered by another endpoint in the network.
- Falcon Discover:
  - [`APPINFO`](https://developer.crowdstrike.com/crowdstrike/docs/falcon-data-replicator-guide#section-appinfo): Application and Detection information
  - [`USERINFO`](https://developer.crowdstrike.com/crowdstrike/docs/falcon-data-replicator-guide#section-userinfo): User Account and Logon information

## Changes

- Define 5 additional parsers.
- Add unit tests.

## Testing

- `mage test:go` (synthetic samples for unit testing)


internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
```diff
@@ -0,0 +1,77 @@
...
+	CID                   pantherlog.String `json:"cid" validate:"required" description:"The customer ID."`
+	AccountType           pantherlog.String `json:"AccountType" validate:"required" description:"The type of account set for the user: 'Domain User', 'Domain Administrator', 'Local User'."`
+	DomainUser            pantherlog.String `json:"DomainUser" validate:"required" description:"Indicates if the user's credentials are part of a domain controller: 'Yes', 'No'."`
+	UserName              pantherlog.String `json:"UserName" validate:"required" description:"The username of the system."`
+	UserSidReadable       pantherlog.String `json:"UserSid_readable" validate:"required" panther:"trace_id" description:"The user SID associated with this process."`
+	LastLoggedOnHost      pantherlog.String `json:"LastLoggedOnHost" description:"The host that was last logged into the system."`
+	LocalAdminAccess      pantherlog.String `json:"LocalAdminAccess" description:"Indicates whether a local user is an admin: 'Yes', 'No'."`
+	LoggedOnHostCount     pantherlog.String `json:"LoggedOnHostCount" description:"The number of hosts logged in at _time."`
```
georgepsarakis, line 68: Is it possible to convert this field to integer?

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
```diff
@@ -0,0 +1,77 @@
...
+	LogonInfo             pantherlog.String `json:"LogonInfo" description:"The login information."`
+	LogonTime             pantherlog.Time   `json:"LogonTime" tcodec:"unix" description:"The last login time by this user in epoch format."`
+	LogonType             pantherlog.String `json:"LogonType" description:"Values defined as follows, INTERACTIVE: The security principal is logging on interactively, NETWORK: The security principal is logging on using a network, TERMINAL SERVER: The security principal has logged in via a terminal server."`
+	MonthSinceReset       pantherlog.String `json:"monthsincereset" description:"The number of months since this user's password was last reset."`
```
georgepsarakis, line 72: Same as above, could be converted to integer.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
```diff
@@ -0,0 +1,77 @@
...
+	LastLoggedOnHost      pantherlog.String `json:"LastLoggedOnHost" description:"The host that was last logged into the system."`
+	LocalAdminAccess      pantherlog.String `json:"LocalAdminAccess" description:"Indicates whether a local user is an admin: 'Yes', 'No'."`
+	LoggedOnHostCount     pantherlog.Uint32 `json:"LoggedOnHostCount" description:"The number of hosts logged in at _time."`
+	LogonInfo             pantherlog.String `json:"LogonInfo" description:"The login information."`
+	LogonTime             pantherlog.Time   `json:"LogonTime" tcodec:"unix" description:"The last login time by this user in epoch format."`
+	LogonType             pantherlog.String `json:"LogonType" description:"Values defined as follows, INTERACTIVE: The security principal is logging on interactively, NETWORK: The security principal is logging on using a network, TERMINAL SERVER: The security principal has logged in via a terminal server."`
+	MonthSinceReset       pantherlog.Uint32 `json:"monthsincereset" description:"The number of months since this user's password was last reset."`
```
alxarch, line 72: I would prefer this to be `pantherlog.Int32`. In general the unsigned variants should be reserved for very specific use cases such as network ports and even then their use is questionable. By using `pantherlog.Uint32` the column will end up being stored as `int64` because our storage backends do not have unsigned integers. There are multiple places in the code base where we use unsigned integer variants as a form of 'validation'. IMO this is problematic in several ways:
- By using unsigned variants we are enforcing log validation on our side. This should not be a concern of Panther. A mistake in some log entry should not prevent us from processing it.
- Validation is _implicit_ and is 'fused' with the JSON decoding process. I'd much rather have a `pantherlog.PortNumber` type instead of just `pantherlog.Uint16`. This uses the go type system to convey the programmer's intent clearly.
- Unsigned integers are rarely used outside of a 'low-level' or binary processing context. In those cases, they are used as 'words', 'bitmasks', or as a single opaque sequence of 4 or 8 bytes. When we want an integer number, regardless of whether the number should always be positive, we should use a 'normal' integer. 
- Using unsigned integers hinders interoperability with other systems such as databases where unsigned variants are not available (i.e. Athena). We risk having to resort to storing numbers as strings to cover the full range of uint64 for example or ending up with undefined behavior of how values outside the supported ranges are actually stored. 

To recap: there's no 'bug' here. You should be conscious as to why unsigned variants of integers are used and how they will be processed. 

Note that this is a point where discussion among us has not led to a consensus in the past.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events_test.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/logtypes/logtesting"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog/tcodec"
+)
+
+func TestAppInfoParser(t *testing.T) {
```
alxarch, line 30: We should only add new tests in YAML form. We are in the process of moving schema definitions to an external repo and tests there can only be in the YAML form.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
```diff
@@ -0,0 +1,117 @@
...
+	InterfaceDescription pantherlog.String `json:"InterfaceDescription" description:"The network adapter used for the IP interface."`
+}
+
+// nolint:lll
+type NotManagedAssets struct {
+	// NOTE: this the host local time, which could be unreliable (?)
+	Time                 pantherlog.Time     `json:"_time" validate:"required" tcodec:"unix" event_time:"true" description:"The host's local time in epoch format."`
+	AIP                  pantherlog.String   `json:"aip" validate:"required" panther:"ip" description:"The sensor’s IP, as seen from the CrowdStrike cloud. This is typically the public IP of the sensor. This helps determine the location of a computer, depending on your network."`
+	AIPCount             pantherlog.Uint8    `json:"aipcount" validate:"required" description:"The number of public-facing IP addresses."`
+	LocalIPCount         pantherlog.Uint8    `json:"localipCount" validate:"required" description:"The number of local IP addresses."`
```
alxarch, line 103: Best use `pantherlog.Int32` here. I understand that having more than 255 IP addresses is probably never going to happen but a bug in the 'agent' software might easily push this value outside the range we expect it to be.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events_test.go
```diff
@@ -0,0 +1,214 @@
...
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/logtypes/logtesting"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog/tcodec"
+)
+
+func TestAIDMasterParser(t *testing.T) {
```
alxarch, line 30: Same as above, we should add new tests in YAML form.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
```diff
@@ -0,0 +1,77 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import (
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/logtypes"
+	"github.com/panther-labs/panther/internal/log_analysis/log_processor/pantherlog"
+)
+
+var (
+	TypeAppInfo = logtypes.MustBuild(logtypes.ConfigJSON{
```
alxarch, line 27: For crowdstike logs we should be using the local `mustBuild`.
This is required so that the unknown event parser is aware of all known log types.
Please update all `TypeFooBar` declarations.

alxarch (changes requested):
Some details that need to be addressed:
- ~Use 'local' `mustBuild` for the new types so that the unknown event parser can exclude results of known event types.~
- All new tests should be in YAML form.

Also I added some observations around use of unsigned integer types.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
georgepsarakis: Exactly, this was actually deliberate :)

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
georgepsarakis: Sure, makes sense, will change it.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
```diff
@@ -0,0 +1,77 @@
...
+		Name:         TypePrefix + ".UserInfo",
+		Description:  "User Account & Logon information provided by Falcon Discover",
+		ReferenceURL: "https://developer.crowdstrike.com/crowdstrike/docs/falcon-data-replicator-guide#section-userinfo",
+		NewEvent:     func() interface{} { return &UserInfo{} },
+	})
+)
+
+// nolint:lll
+type AppInfo struct {
+	// NOTE: this the host local time, which could be unreliable (?)
```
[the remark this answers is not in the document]
georgepsarakis: Great, just noticed it and thought to leave a comment so we can discuss it during review. I'll remove the comments, thanks!

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
georgepsarakis: I see, thanks a lot for the detailed explanation!

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// nolint:lll
+type UserInfo struct {
+	// NOTE: this the host local time, which could be unreliable (?)
+	Time                  pantherlog.Time   `json:"_time" validate:"required" tcodec:"unix" event_time:"true" description:"The host's local time in epoch format."`
+	CID                   pantherlog.String `json:"cid" validate:"required" description:"The customer ID."`
+	AccountType           pantherlog.String `json:"AccountType" validate:"required" description:"The type of account set for the user: 'Domain User', 'Domain Administrator', 'Local User'."`
+	DomainUser            pantherlog.String `json:"DomainUser" validate:"required" description:"Indicates if the user's credentials are part of a domain controller: 'Yes', 'No'."`
+	UserName              pantherlog.String `json:"UserName" validate:"required" description:"The username of the system."`
```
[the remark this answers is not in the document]
georgepsarakis: Fixed in https://github.com/panther-labs/panther/pull/2475/files#diff-fc2fa41b652fe180a14b92ba2f9ff34f3c8bbeb8c5bbb47259380489a60fa003R62

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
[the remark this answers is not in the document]
georgepsarakis: Fixed in 10efa2fcb9b19a8b9cb93448bd456b3902d98bbe.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
georgepsarakis: Fixed in 10efa2fcb9b19a8b9cb93448bd456b3902d98bbe.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
georgepsarakis: Fixed in 10efa2fcb9b19a8b9cb93448bd456b3902d98bbe.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events_test.go
georgepsarakis: Fixed in 10efa2fcb9b19a8b9cb93448bd456b3902d98bbe.

alxarch (approved):
LGTM

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
```diff
@@ -0,0 +1,115 @@
...
+	Continent          pantherlog.String  `json:"Continent" description:"The sensor's continent, as seen from the CrowdStrike cloud."`
+	ComputerName       pantherlog.String  `json:"ComputerName" description:"The name of the host."`
+	ConfigIDBuild      pantherlog.String  `json:"ConfigIDBuild" description:"Build number used as part of the ConfigID."`
+	EventPlatform      pantherlog.String  `json:"event_platform" description:"The platform the sensor is running on. Example values: 'Win', 'Lin', 'Mac'."`
+	FirstSeen          pantherlog.Time    `json:"FirstSeen" tcodec:"unix" description:"The first time the sensor was seen by the CrowdStrike cloud in epoch format."`
+	MachineDomain      pantherlog.String  `json:"MachineDomain" description:"The Windows domain name to which the host is currently joined."`
+	OU                 pantherlog.String  `json:"OU" description:"The organizational unit of the host as seen by the sensor (defined by system admin)."`
+	PointerSize        pantherlog.String  `json:"PointerSize" description:"The processor architecture (in decimal, non-hex format): '4' for 32-bit, '8' for 64-bit."`
```
kostaspap, line 72: Should this be an int?

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
```diff
@@ -0,0 +1,75 @@
...
+		NewEvent:     func() interface{} { return &UserInfo{} },
+	})
+)
+
+// nolint:lll
+type AppInfo struct {
+	Time            pantherlog.Time   `json:"_time" validate:"required" tcodec:"unix" event_time:"true" description:"The host's local time in epoch format."`
+	CID             pantherlog.String `json:"cid" validate:"required" description:"The customer ID."`
+	CompanyName     pantherlog.String `json:"CompanyName" validate:"required" description:"The name of the company."`
+	DetectionCount  pantherlog.Int32  `json:"detectioncount" validate:"required" description:"The number of detections."`
```
kostaspap, line 47: Are we sure about the size or should be Int64? Just double-checking here, have been burned few times from many "counters" that went above their limits :P

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
```diff
@@ -0,0 +1,115 @@
...
+	Timezone           pantherlog.String  `json:"Timezone" description:"The sensor's time zone, as seen from the CrowdStrike cloud."`
+	Version            pantherlog.String  `json:"Version" description:"The host's system version."`
+}
+
+// nolint:lll
+type ManagedAssets struct {
+	Time                 pantherlog.Time   `json:"_time" tcodec:"unix" validate:"required" event_time:"true" description:"The host's local time in epoch format."`
+	AID                  pantherlog.String `json:"aid" validate:"required" description:"The sensor ID. This value is unique to each installation of a Falcon sensor. When a sensor is updated or reinstalled, the host gets a new aid. In those situations, a single host could have multiple aid values over time."`
+	CID                  pantherlog.String `json:"cid" validate:"required" description:"The customer ID."`
+	GatewayIP            pantherlog.String `json:"GatewayIP" validate:"required" description:"The gateway of the system where the sensor is installed."`
```
kostaspap, line 87: This can be added as an IP indicator.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
```diff
@@ -0,0 +1,115 @@
...
+	LocalAddressIP4      pantherlog.String `json:"LocalAddressIP4" validate:"required" panther:"ip" description:"The device's local IP address in IPv4 format."`
+	InterfaceAlias       pantherlog.String `json:"InterfaceAlias" description:"The user-friendly name of the IP interface."`
+	InterfaceDescription pantherlog.String `json:"InterfaceDescription" description:"The network adapter used for the IP interface."`
+}
+
+// nolint:lll
+type NotManagedAssets struct {
+	Time                 pantherlog.Time     `json:"_time" validate:"required" tcodec:"unix" event_time:"true" description:"The host's local time in epoch format."`
+	AIP                  pantherlog.String   `json:"aip" validate:"required" panther:"ip" description:"The sensor’s IP, as seen from the CrowdStrike cloud. This is typically the public IP of the sensor. This helps determine the location of a computer, depending on your network."`
+	AIPCount             pantherlog.Int32    `json:"aipcount" validate:"required" description:"The number of public-facing IP addresses."`
```
kostaspap, line 100: Int64?

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
```diff
@@ -0,0 +1,115 @@
...
+	LocalIPCount         pantherlog.Int32    `json:"localipCount" validate:"required" description:"The number of local IP addresses."`
```
kostaspap, line 101: Same here. Should this be Int64 (e.g. for IPv6)

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
georgepsarakis: The exact data type is not specified in the documentation. Not sure what exactly this counter represents, but I assume it is the total number that a specific file has been found within the organization (all sensors). So passing the 2+ Billion threshold would seem a bit unlikely. In any case, sure we can increase it.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
georgepsarakis: Sounds good, I'll replace with `Int32`.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
georgepsarakis: I thought about it and I wasn't sure if we consider this useful or maybe a bit noisy (we could always add this later). My reasoning was that this would be a specific IP for multiple machines, rather than a specific endpoint. I'll add it to the IP list now, thanks!

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
georgepsarakis: Note that this likely refers to a single machine that is not managed (which I guess translates to not having a sensor installed). Will change it though, thanks.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
georgepsarakis: Same reasoning as below, will change it.

georgepsarakis:
Thanks for the feedback @kostaspap. All changes applied in d1203a774af724532f554a10930ddf9495fdc20b.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/discover_processed_events.go
kostaspap: Thanks! Not sure how they count the detections here tbh, if they are e.g. IP matches on network logs, this could easily reach big numbers. Better safe than sorry ^^

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
kostaspap: Yup. If the users go to indicator search and put an IP, if that IP was present in any of the logs, we want to show them to the users. They can decide later if that's important or not ^^

kostaspap (approved):

PR #2479, opened 2021-01-21
Metrics system for system components

kostaspap:
## Background

Introducing new system for reporting system metrics. It uses the AWS's embedded metric format: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html

It has the benefit that it accumulates different metrics in memory and only writes to CW periodically and at the end of each Lambda invocation. 

This PR includes mostly the new metrics code and an example of its usage. Once this PR is merged, I will make another PR to emit more system status metrics from other parts of the system. 

## Changes

- Added new metrics manager

## Testing

- Unit tests
- Deployed and verified metric E2E

Screen Shot 2021-01-22 at 6 51 01 PM
Screen Shot 2021-01-22 at 6 52 04 PM


internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -125,13 +126,6 @@ func pollEvents(
 			for _, msg := range messages {
 				// pass lambda context to set FULL deadline to process which is pushed down into downloader
 				dataStreams, err := generateDataStreamsFunc(ctx, aws.StringValue(msg.Body))
-				if err == nil {
```
kostaspap, line 128: I had to make this change so that I could get metrics per S3 object.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -143,6 +137,20 @@ func pollEvents(
...
+					// The proper solution for this require an SQS message tracker that allows true concurrent processing.
+					// The overall behavior of the system does not change since reading was triggered
+					// when we were detecting MIME types by using `Peek()`.
+					err = kickOffReader(s)
+					common.GetObject.With(
+						metrics.IDDimension, s.Source.IntegrationID,
+						metrics.StatusDimension, statusFromErr(err),
+					).Add(1)
+					if err != nil {
+						zap.L().Warn("Skipping event due to error", zap.Error(err))
```
kostaspap, line 150: Same logging as few lines above

pkg/metrics/counter.go
```diff
@@ -0,0 +1,204 @@
+package metrics
```
kostaspap, line 1: Most of the code is from go-kit's metrics library (https://github.com/go-kit/kit/tree/master/metrics)

internal/log_analysis/log_processor/common/common.go
```diff
@@ -53,7 +55,10 @@ var (
 	SqsClient    sqsiface.SQSAPI
 	SnsClient    snsiface.SNSAPI
 
-	Config EnvConfig
+	Config    EnvConfig
+	CWMetrics metrics.Manager
+
+	GetObject metrics.Counter
```
kostaspap, line 61: Normally if we had a 'handler' the manager would go in there, and I would probably have a struct to keep all metrics

internal/log_analysis/log_processor/common/common.go
```diff
@@ -91,3 +97,13 @@ type DataStream struct {
...
 }
+
+const (
+	MetricLogProcessorGetObject = "GetObject"
+)
+
+func setupMetrics() {
+	CWMetrics = metrics.NewCWEmbeddedMetrics(os.Stdout)
+	GetObject = CWMetrics.NewCounter(MetricLogProcessorGetObject).
+		With(metrics.SubsystemDimension, metrics.SubsystemLogProcessing)
```
alxarch, line 108: We should keep the label _value_ (`metrics.SubsystemLogProcessing`) in this package instead of the `metrics` one. What needs to be consistent across Lambdas is the label itself (`metrics.SubsystemDimension`)

alxarch (changes requested):
This will greatly improve our observability!
I am conerned about 2 things:
1. The change in `kickoff` is probably a regression
2. I think the `Sync` function needs to take care about holding the lock too long

pkg/metrics/cloudwatch.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// If there are no metrics to be reported
+	// Don't log anything
+	if len(metrics) == 0 {
+		return nil
+	}
+
+	const namespace = "Panther"
+	embeddedMetric := EmbeddedMetric{
+		Timestamp: c.timeFunc(),
```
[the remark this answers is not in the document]
kostaspap: It is actually done once for all the metrics - in line 144, you can see that we actually pass all metrics to that object.

pkg/metrics/cloudwatch.go
```diff
@@ -0,0 +1,173 @@
...
+			},
+		},
+	}
+
+	const rootElement = "_aws"
+	c.stream.WriteObjectField(rootElement)
+	c.stream.WriteVal(embeddedMetric)
+	c.stream.WriteObjectEnd()
+	// nolint: errcheck
+	c.writer.Write(append(c.stream.Buffer(), '\n'))
```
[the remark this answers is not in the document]
kostaspap: Nice one! Yeah, I was annoying I had to do the append.

pkg/metrics/counter.go
```diff
@@ -0,0 +1,204 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import "sync"
+
+type Counter interface {
+	With(dimensionValues ...string) Counter
```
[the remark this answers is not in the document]
kostaspap: I could do `dvs`. `lvs` was short for `LabelValues`

pkg/metrics/dimensions.go
```diff
@@ -0,0 +1,41 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+const (
+
+	// SubsystemDimension dimension
```
giorgosp, line 23: Dimensions are known by the users of the `metrics` pkg, wouldn't be better if each dimension was defined close to the code that creates the needed metric?

pkg/metrics/dimensions.go
```diff
@@ -0,0 +1,41 @@
...
+	StatusDimension = "Status"
+	// StatusDimension indicating that a subsystem operation is well
+	StatusOk = "Ok"
+	// StatusDimension indicating that a subsystem is experiencing authZ/N errors
+	StatusAuthErr = "AuthErr"
+	// StatusDimension indicating some general error with the subsystem
+	StatusErr = "Err"
+
+	// IDDimension dimensions
+	IDDimension = "ID"
```
giorgosp, line 40: What is this `IDDimension` is used for? The comment doesn't help much. Should this be `sourceID` or `customlogID` etc?

pkg/metrics/cloudwatch.go
```diff
@@ -0,0 +1,173 @@
...
+		return true
+	})
+
+	// If there are no metrics to be reported
+	// Don't log anything
+	if len(metrics) == 0 {
+		return nil
+	}
+
+	const namespace = "Panther"
```
georgepsarakis, line 137: Perhaps `namespace` should be a `CWEmbeddedMetricsManager` parameter passed during initialization?

internal/log_analysis/log_processor/common/common.go
```diff
@@ -91,3 +97,13 @@ type DataStream struct {
 	S3Bucket     string
 	S3ObjectSize int64
 }
+
+const (
+	MetricLogProcessorGetObject = "GetObject"
+)
+
+func setupMetrics() {
+	CWMetrics = metrics.NewCWEmbeddedMetrics(os.Stdout)
```
georgepsarakis, line 106: The great thing about using the STDOUT stream is that we don't have to introduce an AWS API call, that would possibly force mocking the metric collection on several locations in the code for unit testing:

> On Lambda, you do not need to set this header yourself. Writing JSON to standard out in the embedded metric format is sufficient.

Minor since almost all the code runs on Lambda, but one possible future improvement would be to add support for the [`PutLogEvents` API](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html). I am guessing that now this would only apply to operational CLI tools, but again this unnecessary at this stage and can possibly be abstracted (and tested) rather easily later.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -197,22 +205,28 @@ func receiveFromSqs(ctx context.Context, sqsClient sqsiface.SQSAPI) ([]*sqs.Mess
...
 	_, err := r.Read(buf[:])
 	return err
 }
+
+// Returns the correct Status dimension from the provided error
+func statusFromErr(err error) string {
+	if err == nil {
+		return metrics.StatusOk
+	}
+	if strings.Contains(err.Error(), "AccessDenied") {
```
georgepsarakis, line 228: Isn't this a bit generic? I am guessing that it can become a common method, but I guess this will become apparent with broader usage.

pkg/metrics/counter.go
georgepsarakis: Maybe we should also copy some or all the unit tests?

pkg/metrics/cloudwatch.go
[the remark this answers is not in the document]
georgepsarakis: Are there any size (length) limitations?

pkg/metrics/cloudwatch.go
```diff
@@ -0,0 +1,173 @@
...
+			ticker.Stop()
+			return
+		}
+	}
+}
+
+// Syncs metrics to the underlying system
+// It will writer the metrics in the Embedded Metric Format
+// https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html
+func (c *CWEmbeddedMetricsManager) Sync() error {
```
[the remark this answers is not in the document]
kostaspap: Updated

georgepsarakis (commented):
Great work! I wonder whether we should consider a more extended metric (dimensions) registry, that would allow consistent relationships with the corresponding alarms. Or will the alarm creation happen automatically, as a completely separate process, by somehow scanning the metric list in regular intervals?

pkg/metrics/cloudwatch.go
```diff
@@ -0,0 +1,173 @@
...
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	var metrics []Metric
```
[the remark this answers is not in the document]
kostaspap: This is called every time we do a sync, which is rather infrequent. The allocation we are looking at is the scale of max few objects (max tens of objects in the future). I don't think there is any performance concern here. I think it's easier to read the code and kinda reduces the complexity of maintaining the `CWEmbeddedMetricsManager` struct

pkg/metrics/cloudwatch.go
```diff
@@ -0,0 +1,173 @@
...
+	var v float64
+	for _, f := range a {
+		v += f
+	}
+	return v
+}
+
+func dimensionNames(dimensionValues ...string) []string {
+	dimensions := make([]string, len(dimensionValues)/2)
+	for i, j := 0, 0; i < len(dimensionValues); i, j = i+2, j+1 {
```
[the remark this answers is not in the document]
kostaspap: We shouldn't be concerned too much. These allocations happen when we are about to output the metrics to I/O. The allocations depend on the number of labels and metrics we are going to emit, not on the time volume of processed data or anything.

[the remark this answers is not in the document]
kostaspap: > Code to avoid them would be a bit more hard to read though

Yeah, that's the trick. We need the dimensions in two places: top level fields in the JSON (with their values) and also nested inside the `EmbeddedMetric` field.

pkg/metrics/cloudwatch.go
kostaspap: Not really needed right now. For the foreseeable future, we will be publishing in this namespace.

internal/log_analysis/log_processor/processor/stream.go
```diff
@@ -143,6 +137,20 @@ func pollEvents(
 				}
 
 				for _, s := range dataStreams {
+					// This is a temporary workaround to ensure all S3 streams are readable.
+					// The proper solution for this require an SQS message tracker that allows true concurrent processing.
+					// The overall behavior of the system does not change since reading was triggered
+					// when we were detecting MIME types by using `Peek()`.
+					err = kickOffReader(s)
+					common.GetObject.With(
+						metrics.IDDimension, s.Source.IntegrationID,
```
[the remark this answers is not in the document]
kostaspap: Good idea, will update

pkg/metrics/cloudwatch.go
```diff
@@ -0,0 +1,173 @@
...
+	c.stream.WriteObjectStart()
+
+	c.counters.Reset().Walk(func(name string, dms DimensionValues, values []float64) bool {
+		// Write `"<metric name>" : <value>`
+		c.stream.WriteObjectField(name)
+		c.stream.WriteVal(sum(values))
+		c.stream.WriteMore()
+
+		// Write dimension values
+		for i := 0; i+1 < len(dms); i = i + 2 {
```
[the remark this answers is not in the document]
kostaspap: I just updated the code in there. Seems more work to split it to separate function.

pkg/metrics/cloudwatch.go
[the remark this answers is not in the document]
alxarch: My concern has more to do with holding the lock for too long. While the lock is held *all* goroutines incrementing the metrics are blocked. Reducing allocations and especially doing the actual `Write` **outside the lock**, will minimize this 'stop-the-world' effect. The choice of metrics 'flush' interval should not be affecting the performance of the system overall. If I wanted to log heap memory usage every 200ms for example (heap mem over 1 minute is not very helpful), I should expect that this won't cause the system to block for 10-50ms every 200ms.

pkg/metrics/dimensions.go
kostaspap: Will update to `SourceIDDimension`

internal/log_analysis/log_processor/common/common.go
kostaspap: I don't think we'll need to support that API tbh. AWS recommends for Lambda to just output to stdout or use a custom Lambda logger when it is available (e.g. for Java Lambdas) or use the CloudWatch agent for non-serverless e.g. EC2 instances.

pkg/metrics/dimensions.go
kostaspap: I think it's better for the Dimensions to be available project-wise. Dimensions essentially define the different "partitions" of the system, so they should be pretty standard. The actual values of those ones, yeah, this should be a bit standard.

kostaspap:
> Great work! I wonder whether we should consider a more extended metric (dimensions) registry, that would allow consistent relationships with the corresponding alarms. Or will the alarm creation happen automatically, as a completely separate process, by somehow scanning the metric list in regular intervals?

This is a very good question! I think we might end up with a registry as you suggest. I expect to work more on this in the next days - for the moment I want to get a metrics framework out and will likely revisit the how we register the labels/dimensions etc.

internal/log_analysis/log_processor/processor/stream.go
kostaspap: Will update

pkg/metrics/counter.go
kostaspap: Nice idea, will do

pkg/metrics/cloudwatch.go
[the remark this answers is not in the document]
georgepsarakis: Answering my question: the [maximum log entry size seems to be 256KB](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html), so it shouldn't be a problem here.

pkg/metrics/dimensions.go
alxarch: Centralizing label names for labels that are used in multiple packages is the only way to ensure consistency throughout the system. This improves our ability to build useful parametric dashboards (i.e. same dashboard for all components, one component, some components).

alxarch: Now if a label is specific for a single metric, it is ok to be declared close to the code that defines the metric.
https://prometheus.io/docs/practices/naming/

pkg/metrics/cloudwatch.go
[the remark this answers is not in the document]
kostaspap: I agree with putting the `Write` outside of the block. 
Trying to avoid the extra allocations for Metrics and Dimension values seems really like microoptimizations - we are talking about slices with max a handful of elements. 
Given that `CWEmbeddedMetricsManager` is meant to write the metrics to the logs and logs cost, we will never be emitting them that frequently (in the `ms` scale) because it costs. I nevertheless changed the code to reuse the slices.

internal/log_analysis/log_processor/processor/stream.go
alxarch: Splitting hairs here, but I think we could be abit more elaborate here and use
```golang
var awsErr awserr.Error
if errors.As(err, &awsErr) {
   return err.Code()
}
return metrics.StatusErr
```

alxarch: Using the 'standard' error codes aws services return might come in handy when cross-referencing metrics

pkg/metrics/cloudwatch.go
```diff
@@ -0,0 +1,191 @@
...
+		dims := dms
+		var labelName, labelValue string
+		for len(dims) >= 2 {
+			labelName, labelValue, dims = dims[0], dims[1], dims[2:]
+			c.stream.WriteObjectField(labelName)
+			c.stream.WriteVal(labelValue)
+			c.stream.WriteMore()
+		}
+
+		c.metricsBuf = append(c.metricsBuf, Metric{Name: name, Unit: "Count"})
```
alxarch, line 144: This is probably too late for this PR but we should devise a way to define units on a counter. Perhaps with a different constructor function. The default would be "Count" but it should be overriddable when creating the metric.

pkg/metrics/cloudwatch.go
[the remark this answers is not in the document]
alxarch: I understand the cost argument but some metrics might require more granular intervals. Especially when debugging/testing a change in a Lambda. This is more relevant with Gauge-like metrics which can have spikes that wouldn't be caught with `time.Minute` intervals. For now we're good though.

alxarch (approved):
This will come in handy!
We can explore ways to define units on metrics in a later PR when we have already had the chance to get a feeling of this in production.
LGTM

internal/log_analysis/log_processor/common/common.go
kostaspap: Done

pkg/metrics/cloudwatch.go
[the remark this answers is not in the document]
kostaspap: Code updated to avoid extra allocations.

pkg/metrics/cloudwatch.go
kostaspap: Completely agree with this. I was thinking to change that part soon, it's easy to add.

internal/log_analysis/log_processor/processor/stream.go
kostaspap: I agree, this could be implemented better. Updated to use `awsutils.IsAnyError`. Unfortunately neither `s3` package nor `kms` package define the error code as constant

PR #2480, opened 2021-01-21
poll panther analysis

lindsey-w:
## Background

This add support for polling pack data from github and tracking packs in a ddb. Note: schema mgmt not yet included.

## Changes

- `github.go` : functionality to poll `panther-labs:panther-analysis` repo for new releases
- pack management:
  - GetPack: return info about a particular pack
  - ListPacks: list pack(s) information 
  - PollPacks: poll the github repo for new releases
  - PatchPack: update a pack to a particular version and/or enabled/disable pack

## Testing

- `mage test:go` : added new unit tests surrounding setting up pack updates
- manual testing (see below)
- integration tests: TBD

To test this manually, I forked the panther-analysis repo, pointed my dev account to it, and created test releases.

<details>
<summary>List Packs</summary>
Input:

```
{
  "ListPacks": {}
}
```

Output:

```
{
  "statusCode": 200,
  "headers": {},
  "multiValueHeaders": {},
  "body": "{\"paging\":{\"thisPage\":1,\"totalPages\":1,\"totalItems\":2},\"packs\":[{\"enabled\":false,\"updateAvailable\":true,\"createdBy\":\"00000000-0000-4000-8000-000000000000\",\"description\":\"This pack groups awesome stuff together\",\"displayName\":\"Totally Awesome Pack\",\"enabledVersion\":{\"id\":36673615,\"name\":\"1.15.1\"},\"id\":\"A.Pack.ID.Two\",\"lastModifiedBy\":\"00000000-0000-4000-8000-000000000000\",\"type\":\"\",\"userId\":\"\",\"versionId\":\"\",\"createdAt\":\"2021-01-20T23:24:23.876168212Z\",\"lastModified\":\"2021-01-21T00:26:16.153829327Z\",\"availableVersions\":[{\"id\":36673615,\"name\":\"1.15.1\"},{\"id\":36684752,\"name\":\"1.15.2\"}],\"detectionPatterns\":{\"IDs\":[\"AWS.Console.LoginFailed\",\"AWS.Console.LoginWithoutMFA\"]}},{\"enabled\":true,\"updateAvailable\":true,\"createdBy\":\"00000000-0000-4000-8000-000000000000\",\"description\":\"This pack groups awesome stuff together\",\"displayName\":\"Totally Awesome Pack\",\"enabledVersion\":{\"id\":36673615,\"name\":\"v1.15.1\"},\"id\":\"A.Pack.ID\",\"lastModifiedBy\":\"00000000-1111-4000-8000-000000000000\",\"type\":\"\",\"userId\":\"\",\"versionId\":\"\",\"createdAt\":\"2021-01-21T00:26:16.161973554Z\",\"lastModified\":\"2021-01-21T16:20:03.586397539Z\",\"availableVersions\":[{\"id\":36673615,\"name\":\"1.15.1\"},{\"id\":36684752,\"name\":\"1.15.2\"}],\"detectionPatterns\":{\"IDs\":[\"AWS.CloudTrail.Created\",\"AWS.Console.LoginFailed\",\"AWS.Console.LoginWithoutMFA\"]}}]}"
}
```

</details>

<details>
<summary>Get Pack</summary>

```
{
  "GetPack": {
    "id": "A.Pack.ID"
  }
}
```

Output:

```
{
  "statusCode": 200,
  "headers": {},
  "multiValueHeaders": {},
  "body": "{\"enabled\":true,\"updateAvailable\":true,\"createdBy\":\"00000000-0000-4000-8000-000000000000\",\"description\":\"This pack groups awesome stuff together\",\"displayName\":\"Totally Awesome Pack\",\"enabledVersion\":{\"id\":36673615,\"name\":\"v1.15.1\"},\"id\":\"A.Pack.ID\",\"lastModifiedBy\":\"00000000-1111-4000-8000-000000000000\",\"type\":\"\",\"userId\":\"\",\"versionId\":\"\",\"createdAt\":\"2021-01-21T00:26:16.161973554Z\",\"lastModified\":\"2021-01-21T16:20:03.586397539Z\",\"availableVersions\":[{\"id\":36673615,\"name\":\"1.15.1\"},{\"id\":36684752,\"name\":\"1.15.2\"}],\"detectionPatterns\":{\"IDs\":[\"AWS.CloudTrail.Created\",\"AWS.Console.LoginFailed\",\"AWS.Console.LoginWithoutMFA\"]}}"
}
```

</details>

<details>
<summary>Poll Packs</summary>

```
{
  "PollPacks": {}
}
```

Output:

```
{
  "statusCode": 200,
  "headers": {},
  "multiValueHeaders": {},
  "body": ""
}
```

</details>

<details>
<summary>Patch (update) Packs</summary>

```
{
  "PatchPack": {
    "enabled": true,
    "id": "A.Pack.ID",
    "userId": "00000000-1111-4000-8000-000000000000",
    "enabledVersion": {
      "id": 36684752,
      "name": "v1.15.2"
    }
  }
}
```

Output:

```
{
  "statusCode": 200,
  "headers": {},
  "multiValueHeaders": {},
  "body": "{\"enabled\":true,\"updateAvailable\":false,\"createdBy\":\"00000000-0000-4000-8000-000000000000\",\"description\":\"This pack groups awesome stuff together\",\"displayName\":\"Totally Awesome Pack\",\"enabledVersion\":{\"id\":36684752,\"name\":\"v1.15.2\"},\"id\":\"A.Pack.ID\",\"lastModifiedBy\":\"00000000-1111-4000-8000-000000000000\",\"type\":\"\",\"userId\":\"\",\"versionId\":\"\",\"createdAt\":\"2021-01-21T00:26:16.161973554Z\",\"lastModified\":\"2021-01-21T16:24:13.4496457Z\",\"availableVersions\":[{\"id\":36673615,\"name\":\"1.15.1\"},{\"id\":36684752,\"name\":\"1.15.2\"}],\"detectionPatterns\":{\"IDs\":[\"AWS.CloudTrail.Created\",\"AWS.Console.LoginFailed\",\"AWS.Console.LoginWithoutMFA\"]}}"
}
```

Dynamo entry for detection in pack has description:
A CloudTrail Trail was created, updated, or enabled. THIS IS A NEW DESCRIPTION.\n

</details>

<details>
<summary>Patch (rollback) Packs</summary>

```
{
  "PatchPack": {
    "enabled": true,
    "id": "A.Pack.ID",
    "userId": "00000000-1111-4000-8000-000000000000",
    "enabledVersion": {
      "id": 36673615,
      "name": "v1.15.1"
    }
  }
}
```

Output:

```
{
  "statusCode": 200,
  "headers": {},
  "multiValueHeaders": {},
  "body": "{\"enabled\":true,\"updateAvailable\":true,\"createdBy\":\"00000000-0000-4000-8000-000000000000\",\"description\":\"This pack groups awesome stuff together\",\"displayName\":\"Totally Awesome Pack\",\"enabledVersion\":{\"id\":36673615,\"name\":\"v1.15.1\"},\"id\":\"A.Pack.ID\",\"lastModifiedBy\":\"00000000-1111-4000-8000-000000000000\",\"type\":\"\",\"userId\":\"\",\"versionId\":\"\",\"createdAt\":\"2021-01-21T00:26:16.161973554Z\",\"lastModified\":\"2021-01-21T16:25:20.53470989Z\",\"availableVersions\":[{\"id\":36673615,\"name\":\"1.15.1\"},{\"id\":36684752,\"name\":\"1.15.2\"}],\"detectionPatterns\":{\"IDs\":[\"AWS.CloudTrail.Created\",\"AWS.Console.LoginFailed\",\"AWS.Console.LoginWithoutMFA\"]}}"
}
```

Dynamo entry for detection in pack has description:
A CloudTrail Trail was created, updated, or enabled.\n

</details>


api/lambda/analysis/models/api.go
```diff
@@ -25,6 +25,7 @@ const (
 	TypeRule      DetectionType = "RULE"
 	TypeGlobal    DetectionType = "GLOBAL"
 	TypeDataModel DetectionType = "DATAMODEL"
+	TypePack      DetectionType = "PACK"
 )
```
lindsey-w, line 29: This isn't a detection type; follows the same question from Nick's other PR - if this should be broken out into a different type (especially since it goes into a different table) - but then several of the dynamo func will need to be re-worked.

internal/core/analysis_api/handlers/poll_packs.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/hashicorp/go-version"
+	"go.uber.org/zap"
+
+	"github.com/panther-labs/panther/api/lambda/analysis/models"
+)
+
+func (API) PollPacks(input *models.PollPacksInput) *events.APIGatewayProxyResponse {
```
lindsey-w, line 32: The scheduled cloudwatch event isn't included in this PR; considering adding to this one or making a new PR for it.

wey-chiang (commented):
Left some comments -- will probably have more comments tomorrow. Nice work!

api/lambda/analysis/configmodels.go
```diff
@@ -65,3 +65,17 @@ type Test struct {
...
+	Description      string           `yaml:"Description"`
+	DetectionPattern DetectionPattern `yaml:"DetectionPatterns"`
+	DisplayName      string           `yaml:"DisplayName"`
+	PackID           string           `yaml:"PackID"`
+}
+
+// DetectionPattern defines what makes up a pack
+type DetectionPattern struct {
+	IDs []string `yaml:"IDs"`
+}
```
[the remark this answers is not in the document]
lindsey-w: > Detections would be a better suited name than DetectionPattern since it's just an array of the detection IDs

I was thinking that the patterns could be expanded at some later date (i.e. today its "IDs" but later it could be like what we specify with `panther_analysis_tool`'s `--filter` option?) If its trying too hard to future-proof, I can simplify it.

> Would we need another field to denote a Panther-managed pack?

I had left that and the 'source' information out on this round to simplify things (and since ddb allows us to add columns arbitrarily). I can add the `managed` flag back in though if you think it'll save us some grief in the future when we do support other sources of packs.

api/lambda/analysis/models/packs.go
```diff
@@ -0,0 +1,101 @@
...
+	UserID         string  `json:"userId" validate:"required"`
+}
+
+// PollPacksInput will also update the pack metadata: "availableReleases" and "updateAvailable"
+type PollPacksInput struct {
+	// PollPacksInput allows for specifying a specific relase to poll or not
+	ReleaseVersion Version `json:"releaseVersion"`
+}
+
+type Pack struct {
```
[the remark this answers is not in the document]
lindsey-w: embedded in my other comment

api/lambda/analysis/models/api.go
```diff
@@ -64,6 +65,12 @@ type LambdaInput struct {
 	GetDataModel     *GetDataModelInput     `json:"getDataModel,omitempty"`
 	ListDataModels   *ListDataModelsInput   `json:"listDataModels,omitempty"`
 	UpdateDataModel  *UpdateDataModelInput  `json:"updateDataModel,omitempty"`
+
+	// Detection Packs
+	GetPack   *GetPackInput   `json:"getPack,omitempty"`
+	ListPacks *ListPacksInput `json:"listPacks,omitempty"`
+	PatchPack *PatchPackInput `json:"patchPack,omitempty"`
+	PollPacks *PollPacksInput `json:"pollPacks,omitempty"`
```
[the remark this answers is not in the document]
lindsey-w: Yea, like you mention, I was going with `Patch` simply because its not a complete update.  The other update actions may only be updating a few fields, but they still pass the entire object in, so I was just trying to convey the difference.

api/lambda/analysis/models/packs.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"time"
+)
+
+type DeletePacksInput = DeleteEntriesInput
+
+type GetPackInput struct {
+	ID string `json:"id" validate:"required,max=1000,excludesall='<>&\""`
+	//VersionID string `json:"versionId" validate:"omitempty,len=32"` // TODO: will this be in S3?
```
nhakmiller, line 29: Since we're maintaining different versions in GitHub I don't think we'll also need a "local" versioned copy of packs in s3.

api/lambda/analysis/configmodels.go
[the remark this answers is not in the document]
nhakmiller: In the RFC discussion the other teams were very insistent that lists of IDs would not be sufficient in the long run, and I suppose I've come around to that too. I like the `DetectionPattern` for that reason, leaves an easy place to point to say we are going to support that in a future iteration.

api/lambda/analysis/models/packs.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+type DetectionPattern struct {
+	IDs []string `json:"IDs"`
+}
+
+type Version struct {
+	ID   int64  `json:"id"`
+	Name string `json:"name"`
+}
```
nhakmiller, line 101: What does `Name` refer to in `Version`? Name of a pack? Name of the version?

internal/core/analysis_api/handlers/dynamo.go
```diff
@@ -352,7 +422,23 @@ func dynamoPut(policy *tableItem) error {
 		return err
 	}
 
-	if _, err = dynamoClient.PutItem(&dynamodb.PutItemInput{Item: body, TableName: &env.Table}); err != nil {
+	return dynamoPutItem(env.Table, body)
+}
+
+// Write a single policy to Dynamo.
```
nhakmiller, line 428:
```suggestion
// Write a single pack to Dynamo.
```

internal/core/analysis_api/handlers/get.go
```diff
@@ -45,6 +45,37 @@ func (API) GetDataModel(input *models.GetDataModelInput) *events.APIGatewayProxy
...
+
+	var item *packTableItem
+	item, err = dynamoGetPack(itemID, false)
+	// TODO: do packs need to be in S3?
+	//if versionID == "" {
+	// Get latest version from Dynamo
+	//} else {
+	// Get specific version from S3
+	//item, err = s3Get(itemID, versionID)
+	//}
```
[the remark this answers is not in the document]
nhakmiller: I said the same in another comment, but I think a "local" S3 copy is less relevant because we know the source of truth will always be available in GitHub. Especially since users can't modify packs yet.

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+
+	"github.com/panther-labs/panther/api/lambda/analysis"
+	"github.com/panther-labs/panther/api/lambda/analysis/models"
+)
+
+const (
+	// github org and repo containing detection packs
+	githubOwner      = "panther-labs"
+	githubRepo       = "panther-analysis"
+	signingKeyID     = "2f555f7a-636a-41ed-9a6b-c6192bf55810" // TODO: update: this is a test key
```
nhakmiller, line 51: @wey-chiang per your double key suggestion, would we just have a `firstSigningKeyID` and `secondSingingKeyID` here? And then of course the keys would live in two different places.

Also lets prefix these with `panther` so in the future when customers can onboard their own packs it's clear what these keys are for.

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+	// First, get all of the release data
+	release, _, err := githubClient.Repositories.GetRelease(context.Background(), githubOwner, githubRepo, version.ID)
+	if err != nil {
+		return nil, nil, err
+	}
+	// retrieve the signature file and entire analysis zip
+	for _, asset := range release.Assets {
+		if *asset.Name == "panther-analysis-all.sig" {
+			signatureData, err = downloadGithubAsset(*asset.ID)
+		} else if *asset.Name == "panther-analysis-all.zip" {
```
nhakmiller, line 92: Let's make `panther-analysis-all.sig` and `panther-analysis-all.zip` constants.

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+				zap.String("sourceFile", *asset.Name),
+				zap.String("repository", githubRepo))
+			return nil, nil, err
+		}
+	}
+	return sourceData, signatureData, nil
+}
+
+func downloadURL(url string) ([]byte, error) {
+	response, err := http.Get(url) // nolint:gosec
```
nhakmiller, line 107: Hmm this one raises some eyebrows. is `http.Get()` not using HTTPS? I would assume it is, but then I'm not sure why `nolint:gosec` is required.

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+		return nil, fmt.Errorf("failed to download %s: %v", url, err)
+	}
+	return body, nil
+}
+
+func downloadValidatePackData(version models.Version) error {
+	sourceData, signatureData, err := downloadGithubRelease(version)
+	if err != nil || sourceData == nil || signatureData == nil {
+		zap.L().Error("GitHub release version download failed", zap.Error(err))
+		return err
```
nhakmiller, line 124: Error might be nil here, if somehow err is not nil and sourceData or signatureData is nil. Might be helpful for debugging purposes to add somethings like this
```suggestion
		zap.L().Error("GitHub release version download failed", zap.Error(err), zap.Bool("sourceData nil", sourceData == nil), zap.Bool("signatureData nil", signatureData==nil))
		return err
```

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+	zipReader, err := zip.NewReader(bytes.NewReader(content), int64(len(content)))
+	if err != nil {
+		return nil, nil, fmt.Errorf("zipReader failed: %s", err)
+	}
+	packs := make(map[string]*packTableItem)
+	detections := make(map[string]*tableItem)
+	detectionBodies := make(map[string]string) // map base file name to contents
+
+	// Process the zip file and extract each pack file
+	for _, zipFile := range zipReader.File {
```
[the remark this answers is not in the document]
wey-chiang: I think that's largely dependent on what compensating controls we implement i.e. including in the manifest a hash of the detection body

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+		zap.L().Error("error extracting pack data", zap.Error(err))
+		return err
+	}
+	cacheLastUpdated = time.Now()
+	cacheVersion = version
+	return nil
+}
+
+func extractPackZipFileBytes(content []byte) (map[string]*packTableItem, map[string]*tableItem, error) {
+	// Unzip in memory (the max request size is only 6 MB, so this should easily fit)
```
nhakmiller, line 141: This comment was true because the input was a lambda invocation. I don't believe it is true any longer, since we're pulling zip files from GitHub. Now the average size of all our detections is around 400Kb so I don't think we're at risk of blowing out the memory any time soon, but we should remove this comment or update it.

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+
+			if err != nil {
+				return nil, nil, err
+			}
+
+			// Map the Config struct fields over to the fields we need to store in Dynamo
+			analysisItem := tableItemFromConfig(config)
+			if analysisItem.Type == models.TypeDataModel {
+				// ensure Mappings are nil rather than an empty slice
+				if len(config.Mappings) > 0 {
```
wey-chiang, line 202: You could move this check out to reduce nesting and return early if it is an empty slice

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+
+			if _, exists := detections[analysisItem.ID]; exists {
+				return nil, nil, fmt.Errorf("multiple analysis specs with ID %s", analysisItem.ID)
+			}
+			detections[analysisItem.ID] = analysisItem
+		}
+	}
+
+	// add python bodies
+	// Finish each policy by adding its body and then validate it
```
nhakmiller, line 234:
```suggestion
	// Finish each detection by adding its body and then validate it
```

internal/core/analysis_api/handlers/github.go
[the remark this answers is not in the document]
nhakmiller: Yeah I see a lot of repeated code here, which will be a hassle to maintain. Might be worth to see which bits can be yanked out into a shared function.

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+		opt.Page = response.NextPage
+	}
+	var availableVersions []models.Version
+	// earliest version of panther managed detections that supports packs
+	minimumVersion, _ := version.NewVersion(minimumVersionName)
+	for _, release := range allReleases {
+		version, err := version.NewVersion(*release.Name)
+		if err != nil {
+			// if we can't parse the version, just throw it away?
+			zap.L().Warn("can't parse version", zap.String("version", *release.Name))
```
nhakmiller, line 274: Yeah this seems correct
```suggestion
			// if we can't parse the version, just throw it away
			zap.L().Warn("can't parse version", zap.String("version", *release.Name))
```

internal/core/analysis_api/handlers/update_pack.go
```diff
@@ -0,0 +1,352 @@
...
+}
+
+func updatePackVersions(newVersion models.Version, oldPacks []*packTableItem) error {
+	newPacks, err := setupUpdatePacksVersions(newVersion, oldPacks)
+	if err != nil {
+		return err
+	}
+	for _, newPack := range newPacks {
+		// TODO: Is it ok to keep the previous user id of the person that modified it?
+		// or should this be the "system" userid?
```
wey-chiang, line 199: Maybe have a separate field, i.e. lastModifiedBy (only users) and lastUpdated ?

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+func validateSignature(rawData []byte, signature []byte) error {
+	// use hash of body in validation
+	intermediateHash := sha512.Sum512(rawData)
+	var computedHash []byte = intermediateHash[:]
+	// The signature is base64 encoded in the file, decode it
+	decodedSignature, err := base64.StdEncoding.DecodeString(string(signature))
+	if err != nil {
+		zap.L().Error("error base64 decoding item", zap.Error(err))
+		return err
+	}
```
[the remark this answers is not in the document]
nhakmiller: I like keeping this isolated like this so we can re-work it as needed.

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+				ID:   *release.ID,
+				Name: *release.Name,
+			}
+			availableVersions = append(availableVersions, newVersion)
+		}
+	}
+	return availableVersions, nil
+}
+
+func validateSignature(rawData []byte, signature []byte) error {
```
nhakmiller, line 288: Is it possible to differentiate between failed validations and just generic errors? I suppose in general it's hard to tell a malformed signature from a maliciously tampered with one, but I think a wrong signature might deserve more alarms than say a KMS transient network error.

internal/core/analysis_api/handlers/list_packs.go
```diff
@@ -0,0 +1,130 @@
...
+	if input.Enabled != nil {
+		filters = append(filters, expression.Equal(
+			expression.Name("enabled"), expression.Value(*input.Enabled)))
+	}
+
+	// does this work with a struct like this?
+	if input.EnabledVersion.Name != "" {
+		filters = append(filters, expression.Equal(expression.Name("enabledRelease"),
+			expression.Value(input.EnabledVersion)))
+	}
```
nhakmiller, line 86: This does work. The alternative is make `EnabledVersion` a struct pointer that is nil when not set, but this method is perfectly acceptable.

wey-chiang (changes requested):
Looking good! 

Will need some work to implement the KMS compensating controls

internal/core/analysis_api/handlers/poll_packs.go
```diff
@@ -0,0 +1,140 @@
...
+	// First, check for a new release in the github repo by listing all releases or use
+	// input value to poll for a speicific release
+	var releases []models.Version
+	var err error
+	//if input.ReleaseVersion != (models.Version{}) {
+	//	releases = []models.Version{
+	//		input.ReleaseVersion,
+	//	}
+	//} else {
```
nhakmiller, line 41: Is this dead code? Let's drop it if so.

nhakmiller (commented):
Long PR! I got about halfway through when my attention got pulled, so I figured I'd leave at least what comments I had and come back later.

If we were to start this again from scratch, I might suggest isolating the put/patch/list packs analysis-api changes from the github polling/downloading changes. Might help with the PR size. At this point though I think we're committed though so no worries.

api/lambda/analysis/models/packs.go
lindsey-w: Name is the name of the version (e.g. "v1.14.0") 

I only include both an ID and Name here because github only allows retrieving releases based on an ID, there may be a way for me to _only_ store the name/tag of the version, do a query to get the id, then retrieve the release. (If that is preferable)

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,330 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	// retrieve the signature file and entire analysis zip
+	for _, asset := range release.Assets {
+		if *asset.Name == "panther-analysis-all.sig" {
+			signatureData, err = downloadGithubAsset(*asset.ID)
+		} else if *asset.Name == "panther-analysis-all.zip" {
+			sourceData, err = downloadGithubAsset(*asset.ID)
+		}
```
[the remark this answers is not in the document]
lindsey-w: Were we going to push the manifest in a separate file than the detections? Here I was downloading the "all" zip file which would include the IOCs zip and the manifest, but I can see pulling down just the manifest file vs "everything" each time.  

I'll add one for the public key - so one will live in kms and one will be published with the release?

internal/core/analysis_api/handlers/github.go
lindsey-w: Yea, it does do https... but I'll add some client config rather than use the default to: (1) add a timeout and (2) also see if I can force https versus leaving it up to the caller to call it with 'https' url

internal/core/analysis_api/handlers/update_pack.go
```diff
@@ -0,0 +1,352 @@
...
+			detection.ResourceTypes = newDetection.ResourceTypes // aka LogTypes
+			// detection.OutputIDs = newDetection.OutputIDs
+			detection.Reference = newDetection.Reference
+			detection.Reports = newDetection.Reports
+			detection.Runbook = newDetection.Runbook
+			// detection.Severity = newDetection.Severity
+			detection.Tags = newDetection.Tags
+			detection.Tests = newDetection.Tests
+			// detection.Threshold = newDetection.Threshold
+			newItems = append(newItems, detection)
```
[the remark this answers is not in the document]
lindsey-w: This could be simplified to just call the `writePack()` with the new detections if we want to overwrite everything (no if/else in that case)

internal/core/analysis_api/handlers/poll_packs.go
```diff
@@ -0,0 +1,131 @@
...
+		TableName: &env.PackTable,
+	})
+	if err != nil {
+		// error looking up the existing pack data
+		zap.L().Error("failed to scan panther-analysis-pack table", zap.Error(err))
+		return &events.APIGatewayProxyResponse{StatusCode: http.StatusInternalServerError}
+	}
+	// Finally, check if update is available
+	latestRelease := getLatestRelease(releases)
+	if isNewReleaseAvailable(latestRelease, currentPacks) {
```
lindsey-w, line 56: This is only "forward" looking - it won't retroactively attempt to add any older versions to the pack. 🤔

internal/core/analysis_api/handlers/github.go
lindsey-w: Added specific check for `ErrCodeKMSInvalidSignatureException` - which will log `Error` - other failures will log `Warn`

internal/core/analysis_api/handlers/dynamo.go
```diff
@@ -287,6 +327,38 @@ func dynamoGet(policyID string, consistentRead bool) (*tableItem, error) {
...
+func dynamoGetItemFromTable(table string, id string, consistentRead bool) (*dynamodb.GetItemOutput, error) {
+	response, err := dynamoClient.GetItem(&dynamodb.GetItemInput{
+		ConsistentRead: &consistentRead,
+		Key:            tableKey(id),
+		TableName:      &table,
+	})
+	if err != nil {
+		zap.L().Error("dynamoClient.GetItem failed", zap.Error(err))
+		zap.L().Error("table", zap.Any("table", &table), zap.Any("key", tableKey(id)))
+		return nil, err
```
kostaspap, line 357: See comment above. Better:
```
return nil, errors.Wrapf(err, "failed to get item %s from table %s", table, id)
```

Some comment goes for the rest of the new code

internal/core/analysis_api/handlers/get.go
```diff
@@ -45,6 +45,30 @@ func (API) GetDataModel(input *models.GetDataModelInput) *events.APIGatewayProxy
 	return handleGet(input.ID, input.VersionID, models.TypeDataModel)
 }
 
+func (API) GetPack(input *models.GetPackInput) *events.APIGatewayProxyResponse {
+	itemID, err := url.QueryUnescape(input.ID)
```
kostaspap, line 49: Do you have to do this? This was needed some time ago when we were using GatewayAPI, it shouldn't be required now - AppSync is invoking Lambda directly.

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,224 @@
...
+	pantherGithubRepo  = "panther-analysis"
+	// signing keys information
+	pantherFirstSigningKeyID = "2f555f7a-636a-41ed-9a6b-c6192bf55810" // TODO: update: this is a test key
+	signingAlgorithm         = kms.SigningAlgorithmSpecRsassaPkcs1V15Sha512
+	// source filenames
+	pantherSourceFilename    = "panther-analysis-all.zip"
+	pantherSignatureFilename = "panther-analysis-all.sig"
+	// cache the detection packs to prevent downloading them multiple times
+	// when updating packs (potentially one at a time) via the UI
+	cacheTimeout = 15 * time.Minute
```
kostaspap, line 55: Do we really need a cache here? What is the concern? It would better to avoid using cache if we don't have to. Simplifies code and also ... "caching is hard".

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,224 @@
...
+var (
+	// cache packs and detections
+	cacheLastUpdated time.Time
+	cacheVersion     models.Version
+	detectionCache   = make(map[string]*tableItem)
+	packCache        = make(map[string]*packTableItem)
+)
+
+func downloadGithubAsset(id int64) ([]byte, error) {
+	rawAsset, url, err := githubClient.Repositories.DownloadReleaseAsset(context.Background(), pantherGithubOwner, pantherGithubRepo, id)
```
kostaspap, line 69: 1. This should be `context.TODO()`
2. If there is an error, we should return early
3. After the error handling, better do `defer rawAsset.Close()`

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,224 @@
...
+func downloadGithubRelease(version models.Version) (sourceData []byte, signatureData []byte, err error) {
+	// Setup options and client
+	// First, get all of the release data
+	release, _, err := githubClient.Repositories.GetRelease(context.Background(), pantherGithubOwner, pantherGithubRepo, version.ID)
+	if err != nil {
+		return nil, nil, err
+	}
+	// retrieve the signature file and entire analysis zip
+	for _, asset := range release.Assets {
+		if *asset.Name == pantherSignatureFilename {
```
kostaspap, line 90: better `aws.String(asset.Name)`. This is a response from an external party, we need to be extra careful to avoid nil pointer dereferences.

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,224 @@
...
+	transport := &http.Transport{
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: false,
+			MinVersion:         tls.VersionTLS12,
+		},
+	}
+	client := &http.Client{
+		Timeout:   10 * time.Second,
+		Transport: transport,
+	}
```
kostaspap, line 119: Instead of defining the client in here on every request, better define one, in the `internal/core/analysis_api/handlers/clients.go` file. Otherwise, testing this code could be very challenging.

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,224 @@
...
+	result, err := kmsClient.Verify(signatureVerifyInput)
+	if err != nil {
+		if awsutils.IsAnyError(err, kms.ErrCodeKMSInvalidSignatureException) {
+			zap.L().Error("signature verification failed", zap.Error(err))
+			return err
+		}
+		zap.L().Warn("error validating signature", zap.Error(err))
+		return err
+	}
+	if *result.SignatureValid {
```
kostaspap, line 219:
```
if aws.BoolValue(result.SignatureValid)
```
for safety since we don't really control that response.

internal/core/analysis_api/handlers/github.go
```diff
@@ -0,0 +1,224 @@
...
+
+var (
+	// cache packs and detections
+	cacheLastUpdated time.Time
+	cacheVersion     models.Version
+	detectionCache   = make(map[string]*tableItem)
+	packCache        = make(map[string]*packTableItem)
+)
+
+func downloadGithubAsset(id int64) ([]byte, error) {
```
kostaspap, line 68: Maybe this code could be inside `pkg/github` package so other parts of the system can re-use. @alxarch

internal/core/analysis_api/handlers/update_pack.go
```diff
@@ -0,0 +1,355 @@
...
+	"go.uber.org/zap"
+
+	"github.com/panther-labs/panther/api/lambda/analysis/models"
+	"github.com/panther-labs/panther/pkg/gatewayapi"
+)
+
+func (API) PatchPack(input *models.PatchPackInput) *events.APIGatewayProxyResponse {
+	// This is a partial update, so lookup existing item values
+	var item *packTableItem
+	item, err := dynamoGetPack(input.ID, false)
```
kostaspap, line 39: Seems this should be consistent read? Otherwise we might patch with wrong data.

internal/core/analysis_api/handlers/update_pack.go
```diff
@@ -0,0 +1,355 @@
...
+
+func updatePackEnabledVersion(input *models.PatchPackInput, item *packTableItem) *events.APIGatewayProxyResponse {
+	// First, update the pack metadata in case the detection pattern has been updated
+	err := updatePackToVersion(input, item)
+	if err != nil {
+		zap.L().Error("Error updating pack metadata", zap.Error(err))
+		return &events.APIGatewayProxyResponse{StatusCode: http.StatusInternalServerError}
+	}
+	// get new version of the pack
+	newPack, err := dynamoGetPack(input.ID, false)
```
kostaspap, line 74: Same here. It should be a consistent read. How often is `dynamoGetPack` going to get invoked? Are there cost considerations for allowing it to be inconsistent?

internal/core/analysis_api/handlers/update_pack.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+	// Build the scan input
+	// include all detection types
+	scanInput, err := buildScanInput(
+		[]models.DetectionType{
+			models.TypeRule,
+			models.TypePolicy,
+			models.TypeDataModel,
+			models.TypeGlobal,
```
kostaspap, line 318: Do we have to add everything explicitly here? Is it possible to avoid it? I am thinking that this would be something really easy to miss.

internal/core/analysis_api/handlers/util.go
```diff
@@ -238,6 +238,57 @@ func writeItem(item *tableItem, userID string, mustExist *bool) (int, error) {
...
+//
+// The following fields are set automatically (need not be set by the caller):
+//     CreatedAt, CreatedBy, LastModified, LastModifiedBy, VersionID
+//
+// To update an existing item,              mustExist = aws.Bool(true)
+// To create a new item (with a unique ID), mustExist = aws.Bool(false)
+// To allow either an update or a create,   mustExist = nil (neither)
+//
+// The first return value indicates what kind of change took place (none, new item, updated item).
+func writePack(item *packTableItem, userID string, mustExist *bool) error {
```
kostaspap, line 251: nit: Move this code to `pack.go`? This file ends up having logic from all parts of the analysis api, will be really hard to maintain.

kostaspap (commented):
This is a lot of code! Thanks for doing this. 
I have put some style comments and few comments about other issues. 

I see that this PR is still a draft, not sure if this is going to become a "real" PR or is it going to be broken down to multiple ones? 
I have to say it is difficult to do meaningful PR over 2k lines of code so in the future it would be better to break it down to smaller chunks.

lindsey-w:
> I see that this PR is still a draft, not sure if this is going to become a "real" PR or is it going to be broken down to multiple ones?
> I have to say it is difficult to do meaningful PR over 2k lines of code so in the future it would be better to break it down to smaller chunks.

I was considering taking in current comments then doing two new PRs - one around the pack definitions/changes and one around the polling/github. Apologies to everyone - this is a pretty ridiculous PR to ask folks to look at in one shot.

api/lambda/analysis/models/packs.go
```diff
@@ -0,0 +1,97 @@
...
+	ID string `json:"id" validate:"required,max=1000,excludesall='<>&\""`
+}
+
+type ListPacksInput struct {
+	// ----- Filtering -----
+	// Only include packs which are enabled or disabled
+	Enabled *bool `json:"enabled"`
+
+	// Only include packs which have this enabledRelease
+	EnabledVersion Version `json:"enabledVersion"`
```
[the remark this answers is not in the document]
lindsey-w: I changed the name of this a couple of times because I also didn't like `enabledVersion` or `enabledRelease` (which is why the comment doesn't match the name - I'll fix that).  I have a hard time with naming conventions XD. This is supposed to represent the current version of the pack that is installed.  I didn't think `currentVersion` really fit either since that could also be interpreted as the "latest" release.

deployments/core.yml
```diff
@@ -444,6 +451,37 @@ Resources:
       ServiceToken: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-cfn-custom-resources
       TableName: panther-analysis
 
+  AnalysisPackTable:
+    Type: AWS::DynamoDB::Table
```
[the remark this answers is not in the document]
lindsey-w: yea, that's the idea

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -158,104 +158,127 @@ func extractZipFile(input *models.BulkUploadInput) (map[string]*tableItem, error
...
+			if err != nil {
+				return nil, nil, err
+			}
 
-		if analysisItem.Type == models.TypeDataModel {
-			// ensure Mappings are nil rather than an empty slice
-			if len(config.Mappings) > 0 {
+			// Map the Config struct fields over to the fields we need to store in Dynamo
+			analysisItem := tableItemFromConfig(config)
+			if analysisItem.Type == models.TypeDataModel && len(config.Mappings) > 0 {
```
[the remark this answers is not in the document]
lindsey-w: would't make much sense 😆  Updating..

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -158,104 +158,127 @@ func extractZipFile(input *models.BulkUploadInput) (map[string]*tableItem, error
...
+	// add python bodies
+	// Finish each detection by adding its body and then validate it
+	for _, detection := range detections {
+		if body, ok := detectionBodies[detection.Body]; ok {
+			detection.Body = body
+			if err := validateUploadedPolicy(detection); err != nil {
+				return nil, nil, err
 			}
-		} else if policy.Type != models.TypeDataModel {
+		} else if detection.Type != models.TypeDataModel {
```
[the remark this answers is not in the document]
lindsey-w: Right, this is checking for the `TypeDataModel` - which isn't necessarily a 'detection' is handled like globals and grouped with the other detections.

internal/core/analysis_api/handlers/github.go
lindsey-w: I guess just trying to prevent us from having to download a release multiple times when someone updates packs one at a time - where each request to update will download the same release.  Perhaps not that big of a concern since this shouldn't happen that often and `panther-analysis` releases aren't that big.

api/lambda/analysis/models/packs.go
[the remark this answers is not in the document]
kostaspap: What about just `version` or `packVersion`?

internal/core/analysis_api/handlers/github.go
kostaspap: Refactoring the code to be a Client struct might be lot of work and is not critical - I think putting the shared code in a place might be useful though. Better coordinate with @alxarch on the latter, since I am not 100% sure which parts of this he might find useful.

api/lambda/analysis/models/packs.go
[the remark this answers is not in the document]
nhakmiller: +1 for `packVersion`.

internal/core/analysis_api/handlers/update_pack.go
```diff
@@ -0,0 +1,355 @@
...
+	} else if input.Enabled != item.Enabled {
+		// Otherwise, we are simply updating the enablement status of the pack and the
+		// detections in this pack.
+		return updatePackEnablement(input, item)
+	}
+	// Nothing to update, report success
+	return gatewayapi.MarshalResponse(item.Pack(), http.StatusOK)
+}
+
+func updatePackEnabledVersion(input *models.PatchPackInput, item *packTableItem) *events.APIGatewayProxyResponse {
```
nhakmiller, line 66: Maybe I'm just tired/low on coffee, but each of the function names in this file seem practically identical and I'm having an extremely difficult time differentiating them/understanding them. Could you add a comment above each function explaining what it does? Like

```
// updatePackEnabledVersion changes what version of a pack is enabled in Dynamo. This function also handles all the additional work needed when updating a pack's version
func updatePackEnabledVersion(input *models.PatchPackInput, item *packTableItem) *events.APIGatewayProxyResponse {
<...>

// updatePackEnablement handles enabling and disabling packs, including enabling/disabling all the detections in that pack
func updatePackEnablement(input *models.PatchPackInput, item *packTableItem) *events.APIGatewayProxyResponse {
```

internal/core/analysis_api/handlers/util.go
nhakmiller: I think `writeItem` makes sense here because many different functions call it for many different reasons. I suspect `writePack` will only be called by a few functions related to enabling/disabling packs. I agree with Kostas this could probably live in a separate file or with the rest of the pack code.

nhakmiller (commented):
Looking good, a few more comments. One more iteration and I think we can merge this into the feature branch, since we're going to be iterating & testing on this a lot before merging to master I think it doesn't need to be perfect.

internal/core/analysis_api/handlers/update_pack.go
lindsey-w: Nope, it is really confusing.  😆  I'll work on adding comments and always open to suggestions

api/lambda/analysis/models/packs.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// ----- Projection -----
+	// Fields to return in the response (default: all)
+	Fields []string `json:"fields" validate:"max=20,dive,required,max=100"`
+
+	// ----- Sorting ----- TODO: not supported in first version
+
+	// ----- Paging -----
+	PageSize int `json:"pageSize" validate:"min=0,max=1000"`
+	Page     int `json:"page" validate:"min=0"`
```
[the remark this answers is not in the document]
kostaspap: >  I feel like it would be strange for just this one to use the new method of pagination.

Makes sense

internal/core/analysis_api/handlers/github.go
kostaspap: > What is context.TODO()?

The context object should normally be passed from the APIs entry point down the call stack to be used by functions like this. So in this case, the "proper" way to do it would require the handler method (the entry point to the lambda) passing the `context.Context` object provided by the Lambda runtime down the call stack... until it reached `func downloadGithubAsset(ctx context.Context, id int64) ([]byte, error)`. We would then pass that ctx to the githubClient function.

Since we don't do that, we could pass the `context.TODO` to satisfy the Github Client's requirements. From `context.TODO` docs
<pre>
// TODO returns a non-nil, empty Context. Code should use context.TODO when
// it's unclear which Context to use <b>or it is not yet available (because the
// surrounding function has not yet been extended to accept a Context
// parameter).</b>
</pre>

Ideally: 
1. All new code should be passing `context.Context` object as argument down the call stack. 
2. The top level functions can invoke private functions for the time being with `context.TODO()`
3. Eventually we wire the handler's context with the API public functions context

but that's a bigger change.

internal/core/analysis_api/handlers/update_pack.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+	if item == nil {
+		return &events.APIGatewayProxyResponse{
+			Body:       fmt.Sprintf("Cannot find %s (%s)", input.ID, models.TypePack),
+			StatusCode: http.StatusNotFound,
+		}
+	}
+	// Update the enabled status and enabledRelease if it has changed
+	// Note: currently only support `enabled` and `enabledRelease` updates from the `patch` operation
+	if input.EnabledVersion.Name != "" && input.EnabledVersion.Name != item.EnabledVersion.Name {
```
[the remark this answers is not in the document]
lindsey-w: If someone is enabling a pack, but not updating it.  I can change the input to make `Enabled` and `PackVersion` `required` on patches - so this would simply to `old version != new version`

internal/core/analysis_api/handlers/update_pack.go
lindsey-w: I decided to (slightly) simplify this by just always calling the same function to update the pack - whether the user is updating enabled/disabled, the pack version, or both.  This removed an if/else and a whole function altogether.  This also has the benefit in that ... if for whatever reason the update/rollback fails in the middle (pack was updated but the detections weren't) - then when calling this again the even though it looks like the pack is on the right version, it will still try to do the update.  This has the downside of doing a complete update every time (even if everything is as it should be) - but I think the benefits outweigh the costs?

internal/core/analysis_api/handlers/update_pack.go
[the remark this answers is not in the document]
lindsey-w: Hmm, so enabling a pack in the UI wouldn't enable the detections in it?

pkg/github/github.go
```diff
@@ -0,0 +1,134 @@
...
+	if rawAsset != nil {
+		defer rawAsset.Close()
+		body, err = ioutil.ReadAll(rawAsset)
+	} else if url != "" {
+		body, err = downloadURL(client, url)
+	}
+	return body, err
+}
+
+func downloadURL(client *GithubClient, url string) ([]byte, error) {
```
alxarch, line 119: We should be passing `context.Context` and use it in the request.
We should use `http.Client.Do()` instead of `Get` to do that. If you take a look into the code of `Get()` it just calls `Do`.

pkg/github/github.go
```diff
@@ -0,0 +1,134 @@
...
+		if response.NextPage == 0 {
+			break
+		}
+		opt.Page = response.NextPage
+	}
+	return allReleases, nil
+}
+
+func downloadGithubAsset(client *GithubClient, owner string, repository string, id int64) ([]byte, error) {
+	rawAsset, url, err := client.Github.Repositories.DownloadReleaseAsset(context.TODO(), owner, repository, id)
```
alxarch, line 104: We should be passing `context.Context` from an argument here and let `context.TODO()` to be decided by the caller.

pkg/github/github.go
```diff
@@ -0,0 +1,134 @@
...
+				// If we failed to download an asset, return the error
+				return nil, err
+			}
+			assetData[aws.StringValue(releaseAsset.Name)] = rawData
+		}
+	}
+	return assetData, nil
+}
+
+func (c *GithubClient) ListAvailableGithubReleases(owner string, repository string) ([]*github.RepositoryRelease, error) {
```
alxarch, line 84: We should be passing `context.Context` here and wire it through to all functions.

alxarch (commented):
I think the approach to listing releases will be problematic due to API rate limiting.

pkg/github/github.go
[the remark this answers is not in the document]
alxarch: After some discussion with @georgepsarakis it seems that the API returns the releases in (almost) reverse order basd on `published at`.
This is undocumented in the API but we can rely on it for now.

In this case fetching *only one page* should work. 100 releases back should be of no interest to anyone for our use case.

The order is not very strict as can be seen [here](https://github.community/t/releases-api-cant-figure-out-order/741)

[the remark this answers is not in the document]
alxarch: We should make sure that calls to the GitHub API are issued only after _explicit user action_ and automatically (i.e. when accessing a page) to avoid hitting the rate limit.

internal/core/analysis_api/handlers/update_pack.go
nhakmiller: Yeah I think so. Plus this is called fairly rarely so we don't have to worry if it's a bit inefficient.

pkg/github/github.go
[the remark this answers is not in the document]
nhakmiller: My understanding is this API call is only made once a day, I don't think there will even be a button users can press to initiate it. Sort of a background update process. Might be good to add a comment about this rate limit though so we know not to expand it too much in the future.

internal/core/analysis_api/handlers/update_pack.go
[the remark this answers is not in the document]
nhakmiller: basically I want to  be able to support this:
- User enables pack
- User sees one rule in the pack they don't like
- User disables rule they don't like
- User gets updates to pack, rule stays disabled.

Maybe there should be a "userDisabled" field, which overrides the "enabled" field? Probably a future improvement at this point, what you have here seems like a good start.

nhakmiller (approved):

PR #2481, opened 2021-01-21
Update dependencies

nhakmiller:
## Background

Snyk detected some out of date python dependencies with vulnerabilities and available updates. Applying those updates here.

## Changes

- Re-installed python dependencies

## Testing

- Unit tests


requirements.txt
```diff
@@ -2,17 +2,17 @@ arrow==0.17.0
 astroid==2.4.2
 attrs==20.3.0
 aws-lambda-builders==1.1.0
-aws-sam-cli==1.11.0
-aws-sam-translator==1.30.1
-bandit==1.6.2
+aws-sam-cli==1.15.0
```
austinbyers, line 5: I'll be removing this one soon!

austinbyers (approved):
Do the libraries in the Python layer need to be updated as well?

nhakmiller:
> Do the libraries in the Python layer need to be updated as well?

Might be a good idea, is this just updating their versions in `panther_config.yml`? If so I can add that.

austinbyers:
> Might be a good idea, is this just updating their versions in panther_config.yml? If so I can add that.

Yes, as well as the [associated documentation](https://docs.runpanther.io/user-guide/analysis/run-time) (once it syncs to enterprise)

nhakmiller:
> > Might be a good idea, is this just updating their versions in panther_config.yml? If so I can add that.
> 
> Yes, as well as the [associated documentation](https://docs.runpanther.io/user-guide/analysis/run-time) (once it syncs to enterprise)

Strangely enough they appear to be at the same version as the packages in `requirements.txt` even after the updates, I guess these packages don't change too often.

PR #2483, opened 2021-01-22
[FE] Showcase dynamic alert field in details page

### LINKED ISSUE #2166: [FE] Dynamic Alert Fields Support

### LINKED ISSUE #2482: [FE] Dynamic Alert Fields - Alerts Page Fix
Alerts Page still uses the analysis-api to display the Alert reference and runbook.

It looks like it is getting the dynamic alert severity and destinations correct here.

Also may want to display the alert description since that is also dynamic now.

3nvi:
## Background

Users are able to dynamically override data for an alert generated by a rule (such as description, reference, etc.), but  - up until now - those weren't showcased in the UI, since we always showed the values that the rule had.

This PR fixes it, by making sure that all dynamic fields originate from the alert item itself.

Closes #2482
Closes #2166 

## Changes

- Add missing fields to schema
- Use new fields in Alert Details page

## Screenshots
Screen Shot 2021-01-22 at 11 07 24 AM

## Testing

- `npm run test`

## Notes

All dynamic fields **automatically** fallback to the values that the rule has, so no need to add a fallback in the UI ourselves

### DIFF
--- a/web/src/pages/AlertDetails/RuleAlertDetails/AlertDetailsInfo/AlertDetailsInfo.tsx
+++ b/web/src/pages/AlertDetails/RuleAlertDetails/AlertDetailsInfo/AlertDetailsInfo.tsx
@@ -17,7 +17,7 @@
  */
 
 import React from 'react';
-import { Box, Card, Flex, Link, SimpleGrid } from 'pouncejs';
+import { Box, Card, Flex, Link, SimpleGrid, Text } from 'pouncejs';
 import Linkify from 'Components/Linkify';
 import { Link as RRLink } from 'react-router-dom';
 import urls from 'Source/urls';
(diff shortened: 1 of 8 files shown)


vorillaz (approved):
LGTM, you can add a simple snapshot testing just for reference.
Pre approving this one.

web/src/pages/AlertDetails/RuleAlertDetails/AlertDetailsInfo/AlertDetailsInfo.tsx
```diff
@@ -40,44 +40,43 @@ const AlertDetailsInfo: React.FC<AlertDetailsInfoProps> = ({ alert, rule }) => {
...
-          <SimpleGrid columns={2} spacing={5}>
-            <Flex direction="column" spacing={2}>
-              <Box
-                color="navyblue-100"
-                fontSize="small-medium"
-                aria-describedby="runbook-description"
-              >
-                Runbook
+      <Card variant="dark" as="section" p={4}>
+        {alert.description && <Text mb={6}>{alert.description}</Text>}
```
alexmylonas, line 44: i am guessing `alert` will always be defined

alexmylonas (approved):

web/src/pages/AlertDetails/RuleAlertDetails/AlertDetailsInfo/AlertDetailsInfo.tsx
3nvi: yeah if it's not, we catch it on a component higher up

PR #2487, opened 2021-01-22
Fix some issues in schema.json for AJV

alxarch:
## Background

@3nvi is trying to use AJV for FE validation of schemas. These changes are required to pass 'strict' ajv mode

## Changes

- Fix AJV 'strict' errors

## Testing

- mage test:go


3nvi (approved):
LGTM

PR #2494, opened 2021-01-22
Add projection to list policies call

nhakmiller:
## Background

When the resource-processor needs to evaluate an updated resource, it requests a list of all the policies from the analysis-api. When it makes this request, it just requests all enabled policies with no projections. This means the compliance status field is included in the response. The analysis-api doesn't store the compliance status of policies directly, instead it queries the compliance-api and requests this information. The compliance-api in turn scans the compliance table and summarizes the results. For large number of resources/policies, this can take a very long time causing timeouts in the resource-processor & analysis-api.

The resource-processor doesn't actually use the compliance status for anything, so all this time is wasted. This change adds a projection to the `ListPolicies` call that does not include the compliance status so the analysis-api does not need to look this information up.

## Changes

- Add a projection to the `ListPolicies` API call in the resource-processor

## Testing

- I timed before and after this change in my dev account and these are the average durations of invocations of the resource-processor:
  - Cache hit (baseline): ~300 ms (average across 10 invocations), with occasional outliers around 1000 ms
  - Pre-change cache miss: ~1,491 ms (average across 6 invocations)
  - Post-change cache miss: ~820 ms (average across 5 invocations)

There is some variation in the times due to the fact that the policy-engine will take longer/shorter to run based on the number of results which I'm not accounting for. But this confirms that there is time savings here even in my small dev account. I also confirmed that the expensive `DescribeOrgComplianceStatus` call is no longer being made by the analysis-api.


internal/compliance/resource_processor/processor/analysis_api.go
```diff
@@ -49,6 +49,18 @@ func getPolicies() (policyMap, error) {
 	// Load from analysis-api
 	listInput := analysismodels.LambdaInput{
 		ListPolicies: &analysismodels.ListPoliciesInput{
+			// Project out compliance status, because it takes a long time to calculate and we don't
```
austinbyers, line 52: Maybe "Exclude compliance status because it takes a long time..."

"Project out" sounds like you're including it in the projection

austinbyers (approved):
Awesome, thanks! Not sure why I didn't do that sooner

lindsey-w (approved):
Nice find

PR #2498, opened 2021-01-25
Handle unescaped space in request URL for ALB log entries

georgepsarakis:
## Background

The current parser flow perceives the record as invalid if it finds more/less than 3 request parameter segments in the [`"request"` field](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-log-entry-syntax) and rejects it. However, the Load Balancer can accept requests with an unescaped space character, in which case it also marks the log entry with the [classification reason as `SpaceInUri`](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#classification-reasons). 

## Changes

- Accept log entries with URLs containing unescaped space character(s) and recompose URL correctly, after splitting on the space delimiter.
- Add failing test case.

## Testing

- `mage test:go`


internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -178,3 +178,34 @@ func (event *ALB) updatePantherFields(p *ALBParser) {
 	event.AppendAnyDomainNamePtrs(event.DomainName)
 	event.AppendAnyAWSARNPtrs(event.ChosenCertARN, event.TargetGroupARN)
 }
+
+func (p *ALBParser) extractRequestParams(record []string) (map[string]string, error) {
```
alxarch, line 182: We should pass the csv fields directly to keep csv index handling in the parent function.

```suggestion
func (p *ALBParser) extractRequestParams(requestLine, classificationReason string) (map[string]string, error) {
```
We can then call this with `p.extractRequestParameters(record[12], record[28])`.
This will make unit testing the method easier also.
Reading `record[12]` or `record[28]` in the middle of this function seems like something that should be someone else's concern.

internal/log_analysis/log_processor/parsers/awslogs/alb_test.go
```diff
@@ -235,6 +234,53 @@ func TestHTTPSNoTarget(t *testing.T) {
...
+		RequestProcessingTime:  aws.Float64(-1),
+		TargetProcessingTime:   aws.Float64(-1),
+		ResponseProcessingTime: aws.Float64(-1),
+		ELBStatusCode:          aws.Int(301),
+		TargetStatusCode:       nil,
+		ReceivedBytes:          aws.Int(234),
+		SentBytes:              aws.Int(391),
+		RequestHTTPMethod:      aws.String("GET"),
+		RequestHTTPVersion:     aws.String("HTTP/1.1"),
+		RequestURL:             aws.String("http://example.com:80/hello world"),
```
alxarch, line 261: Did we observe this in actual ALB events?
This seems very sloppy on their part.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -178,3 +178,34 @@ func (event *ALB) updatePantherFields(p *ALBParser) {
...
+
+	// See https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#classification-reasons
+	classificationReason := record[28]
+	spaceInURIError := classificationReason == "SpaceInUri"
+
+	if !spaceInURIError {
+		return nil, errors.New("invalid record - request parameters could not be parsed")
+	}
+
+	// Reconstruct the initial URL since it contains unescaped space characters
```
alxarch, line 208: I am not sure we are out of the woods here.
A URL containing unencoded spaces is malformed.
Should we be 'fixing' the encoding so that the value can be properly be parsed by url parsers in rules etc?

alxarch: If we return an error we will be 'dropping' the event (probably triggering mis-classification).
We want to store the event even if it has a `SpaceInUri` cancellation reason.

alxarch (changes requested):
I think the event should not be rejected if it has a `SpaceInUri` error.
It is an ALB event and should be stored as one.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
georgepsarakis: We could but this is somehow debatable. It seems to leak extraction details to the main method, whereas it could be entirely the function's concern to retrieve the request parameters from the record. The helper field extraction method is also not public, but rather wraps some of the `Parse` method logic, so it is expected to be tied to internal field position knowledge.

As far as unit testing is concerned, since the method is private, I don't think we can/should test it, but rather perform more extensive testing through `Parse`, if necessary.

Finally, if you notice, the classification reason is not utilized anywhere else (not sure if we should add it at some point). 

Thoughts?

internal/log_analysis/log_processor/parsers/awslogs/alb.go
alxarch: https://play.golang.org/p/5pMmnqvbt92

internal/log_analysis/log_processor/parsers/awslogs/alb_test.go
georgepsarakis: Yes, this is an acceptable error for ALB.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
georgepsarakis: I don't think we should alter it, because we may be shadowing some issue with invalid/malicious input detection. For the record, the Python built-in URL parser seems to accept the malformed URL:

```python
>>> urlparse('http://example.com/hello world')
ParseResult(scheme='http', netloc='example.com', path='/hello world', params='', query='', fragment='')
```

alxarch: The python example uses `kwargs` so it knows `path` should be url encoded.

georgepsarakis: Not sure I get your point, parsing seems to be performed correctly. I'll ask a question on the issue, to get some more feedback on whether it makes sense to replace the raw log field value in this case.

alxarch: I think that the current behavior of the PR should work properly. No need to 'fix' the URL.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
alxarch: My view on this is the inverse. Having the arbitrary `28` index inside the 'inner' function seems like passing irrelevant information to a function that should be a simple mapping of RequestLing -> url parameters.
The function needs not be a method of `p` either.
It is a 'pure' function.
It should not even depend on the `SpaceInUri` thing TBH. 

A request line is valid as long as we can split it into `{method} {uri} {version}`. In this regard, it could just be
```go
func splitRequestLine(line string) (method, uri, version string)
```
The caller can handle the special cases having more context than the 'split' function.


All this is a discussion of code structure. The actual PR will work fine as is. I am just trying to convey my preferences regarding code aesthetics and composability here.

alxarch: On a micro level the current `parseRequestParameters`
has to allocate a `[]string` for the parts *and* a `map[string]sting` _on each line_.
ALB events are quite chatty. Having a zero-allocation split here might be noticeable in terms of performance.

```golang
func splitRequestLine(line string) (method, uri, version string) {
   if pos := strings.IndexByte(line, ' '); 0 <= pos && pos < len(line) {
       method, line := line[:pos], line[pos+1:]
       if pos := strings.LastIndexByte(line, ' '); 0 <= pos && pos < len(line) {
         uri, version := line[:pos], line[pos+1:]
         return method, uri, version
       }
   }
   return line, "", ""
}
```

Again, this is all just educational. I am not requesting this to be changed.

georgepsarakis: > The function needs not be a method of p either.

True, I'll change that.

Your observations make sense as more generic considerations, but I believe you are thinking it in terms of a usable and intuitive public API, which is not the case here though. This is actually functionality that is tied to `Parse`, but just extracted to a separate function for clarity (and so that we don't end up with a 300-line function at some point 😄 ). Thus it makes sense to have knowledge of these esoteric parameters (field position, classification reason hint). If you visualize the whole function body inlined in `Parse` then I believe you'll likely agree that it makes more sense to be written like this.

georgepsarakis: > Having a zero-allocation split here might be noticeable in terms of performance.

Good note, I'll research it a bit more, thanks.

georgepsarakis:
I performed a quick benchmark between the master branch version and the current branch, using two similar log entries differing only on qualifying for having a URL with a space character.

```go
// go test -bench=BenchmarkALBParser -test.run=XXX
func BenchmarkALBParser(b *testing.B) {
	// Parse a log entry N times
	log := "http 2018-08-26T14:17:23.186641Z app/my-loadbalancer/50dc6c495c0c9188 192.168.131.39:2817 " +
		"10.0.0.2:80 0.000 0.001 0.000 200 200 34 366 \"GET http://www.example.com:80/hello world HTTP/1.1\" " +
		"\"curl/7.46.0\" - - arn:aws:elasticloadbalancing:us-east-2:123456789012:targetgroup/my-targets/73e2d6bc24d8a067 " +
		"\"Root=1-58337262-36d228ad5d99923122bbe354\" \"-\" \"-\" 0 2018-08-26T14:17:23.186641Z \"forward\" \"-\" \"-\" \"-\" \"-\" \"Acceptable\" \"SpaceInUri\""

	// log := "http 2018-08-26T14:17:23.186641Z app/my-loadbalancer/50dc6c495c0c9188 192.168.131.39:2817 " +
	// 	"10.0.0.3:80 0.000 0.001 0.000 200 200 34 366 \"GET http://www.example.com:80/ HTTP/1.1\" " +
	// 	"\"curl/7.46.0\" - - arn:aws:elasticloadbalancing:us-east-2:123456789012:targetgroup/my-targets/73e2d6bc24d8a067 " +
	// 	"\"Root=1-58337262-36d228ad5d99923122bbe354\" \"-\" \"-\" 0 2018-08-26T14:17:23.186641Z \"forward\" \"-\" \"-\""
	//
	parser := (&ALBParser{}).New()
	for n := 0; n < b.N; n++ {
		parser.Parse(log)
	}
}
```

Observations:
- Parse performance was close to 6K ns (5950 - best of 5).
- With the `map` version it may reach 6.5K ns (+8% - worst of 5).
- With the static array version the difference seems smaller, the overhead for the space-in-URI log entry is ~70 ns.

I would say that it's difficult to measure reliably, but I guess that using simpler structures makes sense in this case.

PS: I didn't go ahead with measuring [this suggestion](https://github.com/panther-labs/panther/pull/2498#discussion_r563635847), since I am not sure if the errors are meaningful for classification ... Seems it could work efficiently though, so I'll leave a comment in the code so that we can revisit this in the future.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
```diff
@@ -178,3 +178,30 @@ func (event *ALB) updatePantherFields(p *ALBParser) {
 	event.AppendAnyDomainNamePtrs(event.DomainName)
 	event.AppendAnyAWSARNPtrs(event.ChosenCertARN, event.TargetGroupARN)
 }
+
+// TODO(optimization): https://github.com/panther-labs/panther/pull/2498#discussion_r563635847
+func extractRequestParams(requestInfo string) ([3]string, error) {
```
georgepsarakis, line 183: Classification reason check removed after a discussion with Kostas.

internal/log_analysis/log_processor/parsers/awslogs/alb.go
georgepsarakis: Function simplified significantly in 6393ee5a0fbccabeacb90f8b3c07d46ffce0f5b3.

alxarch (approved):
LGTM

PR #2502, opened 2021-01-25
[FE] Fix custom log modal deletion

3nvi:
## Background

When deleting a Custom Log item, we remove the item from the list even when the BE has returned an error. This removes the custom log from the UI but doesn't restore it back unless the user refreshes.

This PR addresses that

## Changes

- Check for BE errors before removing a Custom log from the UI


web/src/components/modals/DeleteCustomLogModal/DeleteCustomLogModal.tsx
```diff
@@ -54,9 +65,7 @@ const DeleteCustomLogModal: React.FC<DeleteCustomLogModalProps> = ({ customLog,
         listAvailableLogTypes: queryData => {
           return {
             ...queryData,
-            logTypes: deletionResponse.error
-              ? queryData.logTypes
-              : queryData.logTypes.filter(logType => logType !== customLog.logType),
+            logTypes: queryData.logTypes.filter(logType => logType !== customLog.logType),
```
3nvi, line 68: this was a remnant from our optimistic response

alexmylonas (approved):
Tested and worked as intended by not removing the custom log source

PR #2510, opened 2021-01-26
Backport generic modal

vorillaz:
## Description

Backported from Panther enterprise: https://github.com/panther-labs/panther-enterprise/pull/2019

### Checklist:


3nvi:
@vorillaz Code looks good, but I'm not sure about the approach.

This modal can replace all of our other modals, but so far our approach is to use purpose-based modals rather than create abstract ones. Is there any reason we don't just add the specific modals we need?

In my mind we should either replace all of our modalss with a generic approach (and change the `showModal` call) or just create  the purpose-specific modals that are needed. 

I'd suggest we do the latter and then open a separate PR where we change our `showModal` call to something like:

```
showModal({ component: MyModal, props: myProps })
```

which may help us a lot

alexmylonas:
I do see value in what @3nvi says and we should do that in a next iteration.

But sometimes we have some very small arbitrary modals that a Generic approach we serve us well instead of having to create a new Modal folder and update `ModalManager` and `ModalContext`

3nvi:
@alexmylonas That's what I'm saying. With my proposed approach, we wouldn't even need a ModalManager anymore

vorillaz:
As @alexmylonas said, I see no real value, adding highly specific modals across the app. The idea is just popping an information modal with a tailor-made body and title.

3nvi:
I agree, we can go ahead with this now, but I'd also want another PR (not necessarily from you) where we remove `ModalManager` and replace our `showModal({ type: MODALS.WHATEVER, props: ModalProps }` with `showModal({ component: MyModal, props: myProps })`.

This will make the generic modal not needed anymore and will allow us to spawn modals in an instant

alexmylonas (approved):
I am approving! I have also added #2513 so we can track the refactor

PR #2516, opened 2021-01-27
Allow users to return event from alert_context

### LINKED ISSUE #2515: Alert_Context fails if user passes event
### Describe the bug

If the users decides to return the provided event from `alert_context`, the `alert_context` method will fail

### Steps to reproduce

1. Go to Log Analysis and select to create a new rule
2. Create a new rule with body:
```
def rule(event): 
    return event

def alert_context(event): 
    return event
```
3. See error

### Expected behavior

Users should be able to return event through alert_context method

### Environment

kostaspap:
## Background

Closes #2515

## Changes

- Allow users to return object from `alert_context` method

## Testing

- mage test:ci
- E2E after deploying to my account and defining a rule:
```
def rule(event): 
    return True

def alert_context(event): 
    return event
```

### DIFF
--- a/internal/log_analysis/rules_engine/tests/test_rule.py
+++ b/internal/log_analysis/rules_engine/tests/test_rule.py
@@ -302,7 +302,7 @@ class TestRule(TestCase):  # pylint: disable=too-many-public-methods
 
         expected_alert_context = json.dumps(event)
         expected_result = RuleResult(
-            matched=True, dedup_output='defaultDedupString:test_alert_context_immutable_event', alert_context=expected_alert_context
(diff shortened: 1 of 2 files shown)


internal/log_analysis/rules_engine/tests/test_rule.py
```diff
@@ -294,6 +294,18 @@ def test_alert_context_immutable_event(self) -> None:
         )
         self.assertEqual(expected_result, rule.run(PantherEvent(event, None)))
 
+    def test_alert_context_returns_full_event(self) -> None:
```
kostaspap, line 297: This test was failing without the changes in `rule.py`

georgepsarakis (approved):
👍

PR #2517, opened 2021-01-27
Handle Managed Schemas in the LogTypesAPI

### LINKED ISSUE #2437: Add support for handling managed schemas in the LogTypesAPI
LogTypes API must be updated to support both user-defined and managed log types.
- Add 'Managed' attribute to distinguish between user-defined and managed schemas
- Add `GetSchema` endpoint to return any schema
- Add `ListManagedSchemaUpdates` endpoint to list managed schema updates from `panther-analysis` repo
- Update transactions for `CustomLog` endpoints to have conditions on  `Managed != true`
- Add the `UpdateManagedSchemas` endpoint to load a managed schemas manifest from a URL and update all entries found within.

### LINKED ISSUE #2440: Update LogTypesAPI resolver to handle managed schemas
LogTypesAPI resolver should handle both custom and managed schemas
All components that require log type resolution should use only this resolver
Special care must be taken for CLI tools that require logtypes resolution since all logtypes should be resolved based on the ManagedSchemas repository.

alxarch:
## Background

LogTypes API  updated to support both user-defined and managed log types.

Closes #2440 
Closes #2437 

## Changes

- Add `logtypesapi.SchemaRecord` to reperesent both user-defined and managed schemas.
- Replace `logtypesapi.LogTypesDatabase` interface with `logtypesapi.SchemaDatabase` to handle operations for all schemas.
- Update `CustomLog` endpoints to use `SchemaDatabase` in a backwards compatible way  (e.g. conditions on  `Managed != true`)
- Add `GetSchema` endpoint to return any schema
- Add `ListManagedSchemaUpdates` endpoint to list managed schema updates from `panther-analysis` repo
- Add the `UpdateManagedSchemas` endpoint to load a managed schemas manifest from a URL and update all entries found within.
- Embeds the minimum required release version for managed schemas using `go-bindata`
- Add a `build.sh` to use when we want to update the minimum required release version for managed schemas embedded in panther
- Adds a new custom resource  `Custom::ManagedSchemas` to handle managed schema updates during a deployment

## Testing

- mage test:go
- e2e deployed and checked that entries in DDB are created and that log processor/ data catalog updater can resolve log types properly using the logtypes API


alxarch:
This is a long PR. The changes to LogTypesAPI are extensive because we need to keep the current endpoints that were designed with only custom logs in mind. In order to manage both user-defined and managed logs in the same db, we introduce the `SchemaRecord` type that represents both user-defined and panther-managed schemas.

The `LogTypesAPI` is pivoting towards `SchemasAPI`. The idea behind this is that defining, updating and managing schemas should not be constrained to log types only. Since a full switch to SchemasAPI cannot happen until FE catches up, we have to retain the CustomLog endpoints along side the Schema endpoints. 

This PR implements the minimal endpoints required so that managed schemas can work properly throughout the system. Namely:
- GetSchema
- ListManagedSchemaUpdates
- UpdateManagedSchemas

internal/log_analysis/managedschemas/github.go
```diff
@@ -0,0 +1,87 @@
...
+
+type GitHubRepository struct {
+	Repo   string
+	Owner  string
+	Client *github.Client
+}
+
+const githubAssetName = "managed-schemas.zip"
+
+func (p *GitHubRepository) ReleaseFeed(ctx context.Context, sinceTag string) ([]Release, error) {
```
kostaspap, line 39: In the end we followed a different approach compared to what @lindsey-w did, right? I am guessing we can look into merging the two codebases down the line.

kostaspap (commented):
Good stuff, nothing pops out in a first pass I did. 

Deferring approval until PR has a bit more information regarding testing.

internal/log_analysis/managedschemas/github.go
kostaspap, line 39: Can you add a descriptions to the public methods?

internal/log_analysis/managedschemas/managedschemas.go
```diff
@@ -0,0 +1,154 @@
...
+func findArchiveFile(z *zip.Reader, name string) *zip.File {
+	for _, f := range z.File {
+		if f.Name == name {
+			return f
+		}
+	}
+	return nil
+}
+
+func ReadYAMLManifest(release string, r io.Reader) ([]ManifestEntry, error) {
```
kostaspap, line 109: Can you add a comment describing the structure on high level? Just enough to give the reader some context.

internal/core/logtypesapi/customlogs.go
```diff
@@ -211,28 +201,21 @@ type DelCustomLogOutput struct {
...
 func (api *LogTypesAPI) ListCustomLogs(ctx context.Context) (*ListCustomLogsOutput, error) {
-	available, err := api.Database.IndexLogTypes(ctx)
-	if err != nil {
-		return nil, err
-	}
-	custom := available[:0]
-	for _, logType := range available {
-		if strings.HasPrefix(logType, customlogs.LogTypePrefix) {
-			custom = append(custom, logType)
+	records := make([]*SchemaRecord, 0, 16)
```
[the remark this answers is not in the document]
alxarch: Just a non-zero pre-allocation. 
When using `append` if the capacity of a slice is not sufficient, `go` doubles* the capacity of the slice and copies data.
So the capacity of an empty slice goes:
`0, 1, 2, 4, 8, 16, 32,...`
By pre-allocating to `16` you can avoid the extra steps on the initial allocations.
This is obviously not needed here. Will remove, it is just a habit of initializing slices to some size from performance sensitive code.

*it actually sets it to `2*capacity+1`

internal/log_analysis/datacatalog_updater/main/lambda.go
```diff
@@ -83,17 +83,14 @@ func main() {
...
-		registry.NativeLogTypesResolver(),
-		snapshotlogs.Resolver(),
-		&logtypesapi.Resolver{
-			LogTypesAPI: logtypesAPI,
-		},
-	)
+	apiResolver := &logtypesapi.Resolver{
+		LogTypesAPI:    logtypesAPI,
+		NativeLogTypes: logtypes.MustMerge("native", registry.NativeLogTypes(), snapshotlogs.LogTypes()),
+	}
```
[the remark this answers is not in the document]
alxarch: Yes, one of the core points this PR addresses

internal/log_analysis/managedschemas/github.go
alxarch: The approach we took had to be functioning before the full PR on their side so it is bare-bones.
Once we have their code and have fully explored our requirements for both packs and managed schemas we can find some common way to handle this if needed.

kostaspap (approved):
Code looks good.
Note that this should be tagged with `sync-to-release` tag.

alxarch:
Ran into some trouble while deploying for e2e tests.
I will fix any issues and re-request a review.

internal/log_analysis/managedschemas/github.go
alxarch: Added

internal/log_analysis/managedschemas/managedschemas.go
alxarch: Added

alxarch:
Fixed various issues with the exported YAML files.
Also fixed an issue brought up by @georgepsarakis that involved semver comparisons when updating a managed schema.

internal/log_analysis/managedschemas/build.sh
```diff
@@ -0,0 +1,70 @@
...
+cat <<EOF > "release.go"
+// Code generated for package $PKG_NAME by build.sh DO NOT EDIT. (@generated)
+package $PKG_NAME
+
+const ReleaseVersion = "${RELEASE_TAG}"
+
+EOF
+
+# Clean up
+rm -rf "${TMP_DIR}"
```
georgepsarakis, line 70: I know that we create a temporary directory in a safely manner above, so it is unlikely for accidents to happen, but this seems a bit disconcerting. As an alternative you could do something like this:

```
BUILD_DIR="$TMPDIR/panther-managed-schemas-build/"$(date "+%Y%m%d%H%M%S")
mkdir -p "$BUILD_DIR"
...
rm -rf "$TMPDIR/panther-managed-schemas-build/*"
```

alxarch: AFAIK `mktemp -d` is random each time. Also it allows the environment to determine where it wants its temporary directories to reside.
What kind of accidents are you referring to?

internal/core/logtypesapi/dynamodb.go
```diff
@@ -33,89 +33,217 @@ import (
...
-	attrRevision   = "revision"
-	attrLogType    = "logType"
+	attrRecordKind   = "RecordKind"
+	attrDisabled     = "IsDeleted"
+	attrRevision     = "revision"
+	attrDescription  = "description"
+	attrReferenceURL = "referenceURL"
+	attrSpec         = "logSpec"
+	attrManaged      = "managed"
+	attrLogType      = "logType"
```
kostaspap, line 54: Can you share some more context? What is happening here? Are we changing the way we store schemas in DDB?

internal/core/logtypesapi/dynamodb.go
```diff
@@ -33,89 +33,217 @@ import (
...
 
-func (d *DynamoDBLogTypes) BatchGetCustomLogs(ctx context.Context, ids ...string) ([]*CustomLogRecord, error) {
-	var records []*CustomLogRecord
+func buildUpdateManagedSchemaTx(tableName string, record SchemaRecord) transact.Transaction {
+	currentRevision := record.Revision - 1
+	return transact.Transaction{
+		&transact.Update{
+			TableName: tableName,
+			Key:       schemaRecordKey(record.Name, 0),
+			Set: map[string]interface{}{
```
kostaspap, line 154: Please add comments there. What is the transaction logic here? Why do we have to use transactions here?

internal/core/logtypesapi/dynamodb.go
```diff
@@ -33,89 +33,217 @@ import (
...
+			ReturnValuesOnConditionCheckFailure: dynamodb.ReturnValueAllOld,
+			// We convert these failures to APIErrors here
+			Cancel: func(r *dynamodb.CancellationReason) error {
+				if transact.IsConditionalCheckFailed(r) {
+					rec := ddbSchemaRecord{}
+					if e := dynamodbattribute.UnmarshalMap(r.Item, &rec); e != nil {
+						return e
+					}
+					if !rec.Managed {
+						return NewAPIError(ErrRevisionConflict, fmt.Sprintf("schema record %q is not managed", rec.RecordID))
```
kostaspap, line 204: Is `ErrRevisionConflict` the correct error code here?

internal/core/logtypesapi/dynamodb.go
```diff
@@ -33,89 +33,217 @@ import (
...
+func (d *DynamoDBSchemas) IndexLogTypes(ctx context.Context) ([]string, error) {
+	input := dynamodb.GetItemInput{
+		TableName:            aws.String(d.TableName),
+		ProjectionExpression: aws.String(attrAvailableLogTypes),
+		Key:                  mustMarshalMap(statusRecordKey()),
+	}
+
+	output, err := d.DB.GetItemWithContext(ctx, &input)
+	if err != nil {
+		L(ctx).Error(`failed to get DynamoDB item`, zap.Error(err))
```
kostaspap, line 225: Shouldn't we log the error at the caller?

internal/core/logtypesapi/dynamodb.go
```diff
@@ -33,89 +33,217 @@ import (
...
+		return nil, err
+	}
+
+	item := struct {
+		AvailableLogTypes []string
+	}{}
+	if err := dynamodbattribute.UnmarshalMap(output.Item, &item); err != nil {
+		L(ctx).Error(`failed to unmarshal DynamoDB item`, zap.Error(err))
```
kostaspap, line 233: Same as above. I think logging should be higher in the call stack?

internal/core/logtypesapi/dynamodb.go
```diff
@@ -156,77 +284,67 @@ func (d *DynamoDBLogTypes) DeleteCustomLog(ctx context.Context, id string, revis
...
+	if enabled {
+		updIndex.Add = attr
+	} else {
+		updIndex.Delete = attr
+	}
+	return append(tx, updIndex)
 }
 
-func (d *DynamoDBLogTypes) CreateCustomLog(ctx context.Context, id string, params *CustomLog) (*CustomLogRecord, error) {
+func (d *DynamoDBSchemas) CreateUserSchema(ctx context.Context, id string, upd SchemaUpdate) (*SchemaRecord, error) {
```
kostaspap, line 318: Is `UserSchema` what we see in other places as `CustomSchema`? It would be better for readability purposes to stick to one terminology.

internal/core/logtypesapi/dynamodb.go
```diff
@@ -156,77 +284,67 @@ func (d *DynamoDBLogTypes) DeleteCustomLog(ctx context.Context, id string, revis
...
-		LogType:   id,
-		Revision:  1,
-		CustomLog: *params,
-		UpdatedAt: now,
-	}
-	tx := buildCreateRecordTx(d.TableName, result)
+	record := SchemaRecord{
+		Name:         id,
+		Revision:     1,
+		Managed:      false,
```
kostaspap, line 323: We use `Custom` and `Managed`. Instead of having `managed` true/false, should we just have `custom` true/false? or maybe we should have a `type` with value string that can take either `Managed` or `Custom` value? This might require bigger changes in this PR, would like to hear your thoughts.

internal/core/logtypesapi/dynamodb.go
```diff
@@ -254,31 +372,31 @@ func buildCreateRecordTx(tbl string, record CustomLogRecord) transact.Transactio
...
-				attrAvailableLogTypes: newStringSet(record.LogType),
+				attrAvailableLogTypes: newStringSet(record.Name),
 			},
 			Key: statusRecordKey(),
 		},
 	}
 }
 
-func (d *DynamoDBLogTypes) UpdateCustomLog(ctx context.Context, id string, revision int64, params *CustomLog) (*CustomLogRecord, error) {
+func (d *DynamoDBSchemas) UpdateUserSchema(ctx context.Context, id string, rev int64, upd SchemaUpdate) (*SchemaRecord, error) {
```
kostaspap, line 391: Why do we need separate Update functions but on `ListDeletedLogTypes ` function?

internal/core/logtypesapi/dynamodb.go
```diff
@@ -400,10 +515,10 @@ func newStringSet(strings ...string) *dynamodb.AttributeValue {
 	}
 }
 
-func (d *DynamoDBLogTypes) ListDeletedLogTypes(ctx context.Context) ([]string, error) {
+func (d *DynamoDBSchemas) ListDeletedLogTypes(ctx context.Context) ([]string, error) {
```
kostaspap, line 518: nit: renamed to `ListDisabledSchemas`?

internal/core/logtypesapi/inmem.go
```diff
@@ -21,46 +21,39 @@ package logtypesapi
 import (
 	"context"
 	"fmt"
+	"strings"
```
kostaspap, line 24: Fine to keep this change here but a more general thought: Having to maintain a in-memory version of the schema's repo will be challenge to maintain long term as it requires its own tests to verify it is actually working as expected etc. A mocking library could be used for unit tests that don't want to interact with DDB.

internal/core/logtypesapi/managed_schemas.go
```diff
@@ -0,0 +1,148 @@
...
+	return &UpdateManagedSchemasOutput{
+		Records: records,
+	}, nil
+}
+
+func loadManifest(ctx context.Context, release, manifestURL string) ([]managedschemas.ManifestEntry, error) {
+	if release == managedschemas.ReleaseVersion {
+		return managedschemas.LoadDefaultManifest()
+	}
+	return managedschemas.LoadReleaseManifestFromURL(ctx, manifestURL)
```
kostaspap, line 113: Why is this check needed? can you add a comment?

internal/core/logtypesapi/resolver.go
```diff
@@ -31,23 +30,21 @@ import (
...
 // Resolve implements logtypes.Resolver
 func (r *Resolver) Resolve(ctx context.Context, name string) (logtypes.Entry, error) {
-	if !strings.HasPrefix(name, customlogs.LogTypePrefix) {
-		return nil, nil
-	}
-	reply, err := r.LogTypesAPI.GetCustomLog(ctx, &GetCustomLogInput{
-		LogType:  name,
+	reply, err := r.LogTypesAPI.GetSchema(ctx, &GetSchemaInput{
+		Name:     name,
 		Revision: 0,
```
kostaspap, line 45: Is revision 0 meant to be the latest revision?

internal/log_analysis/log_processor/logschema/infer.go
```diff
@@ -247,6 +247,12 @@ func inferFieldSchema(s *FieldSchema, field reflect.StructField) error {
 		value.IsEventTime = isEventTime
 	}
 	value.TimeFormat = field.Tag.Get("tcodec")
+	if value.Type == TypeTimestamp {
+		value.TimeFormat = "rfc3339"
+		if format, ok := field.Tag.Lookup("tcodec"); ok {
+			value.TimeFormat = format
+		}
+	}
```
kostaspap, line 255: Why is this part of this PR? What is this change for?

kostaspap (commented):
Woah, big PR! Put some comments, nothing seems like a blocker, mostly changes related to readability

deployments/log_analysis.yml
```diff
@@ -115,9 +115,17 @@ Conditions:
...
+    Type: Custom::ManagedSchemas
+    DependsOn: UpdaterFunction
+    Properties:
+      CustomResourceVersion: !Ref CustomResourceVersion
+      ServiceToken: !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:panther-cfn-custom-resources
+
   ###### Update Glue Table Schemas for Deployed Tables #####
   UpdateLogProcessorGlueTables:
-    DependsOn: UpdaterFunction
+    DependsOn: UpdateManagedSchemas
```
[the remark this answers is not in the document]
alxarch:
```suggestion
    DependsOn: UpdateManagedSchemas # The data catalog updater needs to have up-to-date schemas 
```

internal/core/logtypesapi/dynamodb.go
alxarch: We are now storing both managed and user-defined schemas. The difference between them is that managed schemas have `managed` field set to `true` and they have `release` set to their version tag. Also managed schemas keep no record of past revisions.

internal/core/logtypesapi/inmem.go
alxarch: Yeah, the logic was to keep inmem for tests but that is also cumbersome. I should just add a mock for the `SchemasDatabase` interface and use that in tests.

internal/core/logtypesapi/managed_schemas.go
alxarch: We want to avoid having to call github on deployment

internal/core/logtypesapi/resolver.go
alxarch: Yes

internal/log_analysis/log_processor/logschema/infer.go
alxarch: The exported YAML files were not working properly.

internal/core/logtypesapi/dynamodb.go
alxarch: I'll remove this, it's unused

internal/log_analysis/managedschemas/build.sh
georgepsarakis: It's better to depend on (some) constant parts rather than entirely on variables for catastrophic commands. I am not sure how this could happen with the current script, I just thought to point it out if it takes one more concern out of the equation, we can always revisit this later.

internal/core/logtypesapi/api.go
```diff
@@ -38,31 +39,36 @@ const LambdaName = "panther-logtypes-api"
...
+	// GetSchema gets a single schema record
+	GetSchema(ctx context.Context, id string, revision int64) (*SchemaRecord, error)
+
+	// UpdateSchema updates a managed schema to the release version provided
+	UpdateUserSchema(ctx context.Context, id string, rev int64, upd SchemaUpdate) (*SchemaRecord, error)
+
+	// UpdateManagedSchema updates a managed schema to the release version provided
+	UpdateManagedSchema(ctx context.Context, id string, release string, upd SchemaUpdate) (*SchemaRecord, error)
+
+	ToggleSchema(ctx context.Context, id string, enabled bool) error
```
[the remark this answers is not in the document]
alxarch: added

internal/log_analysis/managedschemas/build.sh
alxarch: Seeing as this is to be run once when updating the minimum required version, I'll just remove the Clean up altogether

internal/core/logtypesapi/customlogs.go
```diff
@@ -136,31 +130,27 @@ func (api *LogTypesAPI) checkUpdate(a, b *logschema.Schema) error {
...
-	if err := yaml.Unmarshal([]byte(c.LogSpec), &schema); err != nil {
+	if err := yaml.Unmarshal([]byte(spec), &schema); err != nil {
 		return nil, NewAPIError(ErrInvalidSyntax, err.Error())
 	}
+	if desc != "" {
+		schema.Description = desc
+	}
+	if refURL != "" {
+		schema.ReferenceURL = refURL
+	}
```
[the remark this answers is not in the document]
alxarch: What this does is to use the values in the form fields as _overrides_ for the values in the YAML.
This is the behavior we want here. Having form fields that overlap with the values in the YAML is not ideal but it allows the user to focus on defining the fields in YAML.
We also do not want to be modifying the user's YAML _input_ so the only point where we can apply these overrides is in `buildSchema`, _before_ validating the user input.  OTOH we are reusing this function to handle managed schema updates where the overrides would be empty.
I updated the code to distinguish between `buildAndValidateUserSchema` and `buildAndValidateManagedSchema`.

internal/core/logtypesapi/dynamodb.go
alxarch: Changed to AlreadyExists

internal/core/logtypesapi/dynamodb.go
alxarch: Removed the unused function

internal/core/logtypesapi/dynamodb.go
alxarch: Removed unused function

internal/core/logtypesapi/dynamodb.go
alxarch: Yes we stick to one. `SchemaDatabase` has `Schema` actions.
The LogTypesAPI has to retain the `CustomLog` endpoints for compatibility.

internal/core/logtypesapi/dynamodb.go
alxarch: See above, the terms are `UserSchema` and `ManagedSchema`. Since entries already in DDB are `UserSchemas` we have to use `Managed: true` as the boolean flag.

internal/core/logtypesapi/dynamodb.go
alxarch: We need separate separate functions because `UserSchema` records are keeping historical records of previous revisions.
The `SchemaDatabase` interface is to be further simplified in an upcoming PR.
I don't understand the second part of the question after 'but'.

internal/core/logtypesapi/inmem.go
alxarch: The `SchemaDatabase` interface will be simplified in a subsequent PR. This will trim down on the code we require to maintain for `InMem`

internal/core/logtypesapi/dynamodb.go
alxarch: We use transactions to have coherent logic and error handling between endpoints. We could be using a simple `UpdateItem` for managed schemas but it would deviate from what we use for UserSchemas. Also building the update expression is much simpler using the `transact` package than doing it manually.

internal/core/logtypesapi/dynamodb.go
kostaspap: I meant to write "Why we don't have different `ListDeleted` functions for User/Managed schemas". It is clear to me now though.

Issue #2467, opened 2021-01-20
Upgrade AWS Go SDK to v2


austinbyers:
### Description

AWS Go SDK v2 is now [generally available](https://aws.amazon.com/about-aws/whats-new/2021/01/aws-sdk-for-go-version-2-now-generally-available/)!

There are substantial performance improvements in the new library (which, in turn, lead to cost savings), so it's definitely worth our time to switch sooner than later. We'll wait a few weeks for them to patch any issues before we start upgrading Panther services. The upgrade need not happen all at once - each service owner can migrate their own service when they have a few spare cycles to do so.

This epic is not necessarily anyone's dedicated responsibility yet; it's just a convenient place to track which services have been upgraded and any roadblocks we encounter along the way.

See [migration docs](https://aws.github.io/aws-sdk-go-v2/docs/migrating/)

### Acceptance Criteria

- Eventually, all aws go sdk imports are switched to `github.com/aws/aws-sdk-go-v2`

Check off each service once the migration is complete:

#### `cmd`

- [ ] `devtools`
- [ ] `opstools`

#### `internal/compliance`

- [ ] `alert-forwarder`
- [ ] `alert-processor`
- [ ] `aws-event-processor`
- [ ] `compliance-api`
- [ ] `datalake-forwarder`
- [ ] `remediation-api`
- [ ] `remediation-processor`
- [ ] `resources-api`
- [ ] `snapshot-poller`
- [ ] `snapshot-scheduler`

#### `internal/core`

- [ ] `alert-delivery`
- [ ] `analysis-api`
- [ ] `custom-resources`
- [ ] `database-api` (enterprise)
- [ ] `layer-manager`
- [ ] `logtypesapi`
- [ ] `metrics-api`
- [ ] `organization-api`
- [ ] `outputs-api`
- [ ] `source-api`
- [ ] `users-api`

#### `internal/log-analysis`

- [ ] `alert-forwarder`
- [ ] `alerts-api`
- [ ] `datacatalog-compactor` (enterprise)
- [ ] `datacatalog-compactor-reaper` (enterprise)
- [ ] `datacatalog-updater`
- [ ] `log-processor`
- [ ] `log-puller` (enterprise)
- [ ] `message-forwarder`

#### `pkg`

- [ ] `awsathena`
- [ ] `awsbatch`
- [ ] `awscfn`
- [ ] `awscostexplorer`
- [ ] `awsretry`
- [ ] `awssqs`
- [ ] `awsutils`
[...]

PR #2524, opened 2021-01-27
Replace sam-cli with our own packaging

austinbyers:
## Background

Deploying the root stack (`mage master:deploy`) takes so long (30+ mins) that developers can't feasibly use it for day-to-day work. But they will soon have to - we're going to remove the old deploy process entirely.

The first step in migrating deployment methods is to make the root-stack deployment faster. It takes such a long time primarily because `sam package` uploads assets to S3 sequentially, whereas `mage deploy` packages several stacks in parallel.

The solution then, is to rip out the `aws-sam-cli` entirely and build our own packaging code. This way, we have full control over the parallelism, and as an added bonus we don't have to shell out to another process to re-parse the templates. I'm always happy to remove dependencies we don't need

As a result of these changes, the root stack deployment is 30% faster!

Part of https://github.com/panther-labs/panther-enterprise/issues/1815

### New Packaging Code
Added packaging code to `tools/mage/pkg`. The code is primarily designed for the root-stack deployment, but until we finish the migration I've also adapted the existing `mage deploy` to use the new packaging code.

- Instead of doing all the build steps at the beginning of the deploy process (`go build`, `docker build`, etc), these operations have been pushed into the packaging step. This way, builds and uploads can happen in parallel with each other, and every stack builds only the exact set of resources it needs. In other words, instead of walking the entire source tree to build every Lambda function, we build the exact paths needed for each stack "just-in-time" right before we upload them to S3.
- Removed `aws-sam-cli` and updated requirements file accordingly: `boto3` updated to the version currently supported in the lambda environment.

Here's a sample of the new deploy output:

```
16:24:35	INFO	[deploy]	uploading 14.7 MB file out/lambda/panther-alert-processor.zip (sha1:595abcf002) to S3
16:24:36	INFO	[deploy]	uploading 15.9 MB file out/lambda/panther-alert-delivery-api.zip (sha1:1b79a35a9e) to S3
16:24:38	INFO	[deploy]	uploading 14.0 MB file out/lambda/panther-compliance-api.zip (sha1:29bc4e492c) to S3
16:24:38	INFO	[deploy]	uploading 14.7 MB file out/lambda/panther-cloudsecurity-datalake-forwarder.zip (sha1:7cd9d6331b) to S3
16:24:39	INFO	[deploy]	uploading 55.3 kB file out/lambda/panther-policy-engine.zip (sha1:0473eeb37d) to S3
16:24:40	INFO	[deploy]	uploading 94.3 kB file out/lambda/panther-aws-remediation.zip (sha1:b88c2a1f91) to S3
16:24:40	INFO	[deploy]	uploading 15.2 MB file out/lambda/panther-resources-api.zip (sha1:78f6a52b1e) to S3
16:24:46	INFO	[deploy]	packaged out/deployments/pkg.core.yml
16:24:47	INFO	[deploy]	create CloudFormation stack panther-core
16:24:48	INFO	[deploy]	packaged out/deployments/pkg.cloud_security.yml
16:24:48	INFO	[deploy]	uploading 56.0 kB file out/deployments/pkg.cloud_security.yml (sha1:a23aff343e) to S3
16:24:48	INFO	[deploy]	create CloudFormation stack panther-cloud-security
16:26:24	INFO	[deploy]	    √ panther-appsync finished (4/9)
16:26:53	INFO	[deploy]	    √ panther-cloud-security finished (5/9)
16:27:06	INFO	[deploy]	    √ panther-core finished (6/9)
16:27:07	INFO	[deploy]	docker build web server (deployments/Dockerfile)
[+] Building 0.0s (0/1)
[+] Building 0.6s (2/3)
 => [internal] load build definition from Dockerfile                                                                                                                                0.0s
 => => transferring dockerfile: 37B                                                                                                                                                 0.0s
[+] Building 1.5s (19/19) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                0.0s
 => => transferring dockerfile: 37B                                                                                                                                                 0.0s
 => [internal] load .dockerignore                                                                                                                                                   0.0s
 => => transferring context: 35B                                                                                                                                                    0.0s
 => [internal] load metadata for docker.io/library/node:14.15-alpine                                                                                                                1.0s
 => [build-env 1/8] FROM docker.io/library/node:14.15-alpine@sha256:2ae9624a39ce437e7f58931a5747fdc60224c6e40f8980db90728de58e22af7c                                                0.0s
 => [internal] load build context                                                                                                                                                   0.1s
 => => transferring context: 113.15kB                                                                                                                                               0.1s
 => CACHED [stage-1 2/7] RUN mkdir /code                                                                                                                                            0.0s
 => CACHED [stage-1 3/7] WORKDIR /code                                                                                                                                              0.0s
 => CACHED [build-env 2/8] RUN apk update && apk add --no-cache --update nasm libtool alpine-sdk autoconf automake python                                                           0.0s
 => CACHED [build-env 3/8] RUN mkdir /code                                                                                                                                          0.0s
 => CACHED [build-env 4/8] WORKDIR /code                                                                                                                                            0.0s
 => CACHED [build-env 5/8] ADD package.json package-lock.json ./                                                                                                                    0.0s
 => CACHED [build-env 6/8] RUN npm i                                                                                                                                                0.0s
 => CACHED [build-env 7/8] ADD web web                                                                                                                                              0.0s
 => CACHED [build-env 8/8] RUN npm run build                                                                                                                                        0.0s
 => CACHED [stage-1 4/7] COPY --from=build-env /code/web/dist web/dist                                                                                                              0.0s
 => CACHED [stage-1 5/7] ADD package.json .                                                                                                                                         0.0s
 => CACHED [stage-1 6/7] ADD web/scripts web/scripts                                                                                                                                0.0s
 => CACHED [stage-1 7/7] RUN npm install express express-static-gzip ejs chalk dotenv helmet                                                                                        0.0s
 => exporting to image                                                                                                                                                              0.0s
 => => exporting layers                                                                                                                                                             0.0s
 => => writing image sha256:861868080c0009c4d233c463386f201df1cd8f545657298fe6edfb7da11140c8                                                                                        0.0s
16:27:10	INFO	[deploy]	uploading 19.4 MB file out/lambda/panther-log-processor.zip (sha1:02c4ddabbc) to S3
16:27:10	INFO	[deploy]	pushing docker image 101802775469.dkr.ecr.ca-central-1.amazonaws.com/panther-web:861868080c00 to remote repo
```

### New Config File
The `deployments/panther_config.yml` file will be removed once we remove the old deploy logic. We need something in its place - today the only way to configure root stack parameters is an environment variable, which won't handle spaces well.

I've added a new `deployments/root_confg.yml` which is _not_ checked into the repo. It is autogenerated if it doesn't exist yet, and looks like this:

```yml
# Root stack configuration - edit to configure your own parameter overrides

RootStackName: panther
PipLayer:
- jsonpath-ng==1.5.2
- policyuniverse==1.3.2.2
- requests==2.23.0
ParameterOverrides:
  CloudWatchLogRetentionDays: "14"
  CompanyDisplayName: PantherDev
  FirstUserEmail: user@example
  FirstUserGivenName: Austin
```

This makes it easy for developers to add their own config without worrying about checking it in accidentally. Even better, this persists across deployments - you won't have to constantly type in your name and email for every deploy. `mage clean` will remove the file.

As its name suggests, this file is only for the root stack; the current `mage deploy` will still use the old config file for now

### CFN Changes
- The CFN templates have been tweaked slightly to accommodate the new packaging. For example, we don't need to pass in the ECR registry and image ID; it can be embedded in the template just like packaging embeds S3 paths. This cleans up a fair bit of CFN, which is a nice bonus
- The `master.yml` stack has been renamed `root.yml` to avoid confusion with the `master` git branch. The `mage master:*` commands will be renamed later
- Developer assets (S3 bucket / ECR repo) have been added to their own `panther-dev` stack - this stack has to be deployed *before* Panther itself. Once we finish consolidating the deployment models, this means we'll be able to remove the `SourceBucket` and `PantherWebImageRepository` resources from the bootstrap stack. I like this - developer resources don't need to be in the official releases

### Bug Fixes
- A `mage clean setup deploy` would sometimes fail with "too many open files" when zipping the Python layer. This was due to an unclosed file handler in `pkg/shutil`
- We now show the `docker build` output on each deploy - newer versions of docker are much better about managing the stdout so it doesn't consume the entire terminal
    - `docker build` (and therefore the entire deploy) was actually failing in verbose mode (`mage -v deploy`) because we were getting the imageID from the command output. But that output only exists in quiet mode, which we don't use anymore. So now we use the `--iidfile` flag to get the final image ID from a tempfile, which is safer than parsing command output anyway

## Testing

- `mage deploy`
- `mage master:deploy`
- `mage master:publish` - I published "v1.16.0-dev" to one region to make sure I didn't break the publication code - publishing releases should be much faster with the new packaging code
- Deploy benchmark:

Packaging _ Deploy Benchmark

For each deploy, I did a `go clean -cache`, `docker system prune -a`, and `mage clean setup`. I'm based in SF and have fiber internet.

As you can see, this branch is about 30% faster for root-stack deployment! In fact, it's about as fast as `mage deploy` is today. This speedup should be enough to make root-stack development feasible.


austinbyers:
I'll need the reviewers' help to decide where to put the `docker build` and how to show its progress.

Today, in the `master` branch, `docker build` is run essentially by itself before deploying the web stack. With the change in this PR, the docker build has been moved into the packaging logic, meaning it can run in parallel with other `go build` packaging. This has worked pretty well in my tests so far.

However, the default `docker build` output (which is really nice) completely clobbers the other log messages:

```
18:08:08	INFO	[master:deploy]	packaged out/deployments/pkg.bootstrap.yml
[+] Building 0.3s (2/3)
[+] Building 0.5s (2/3)
 => [internal] load build definition from Dockerfile                                                                                                                                0.0s
[+] Building 1.6s (4/18)
[+] Building 1.7s (4/18)
[+] Building 2.1s (4/18)
[+] Building 4.4s (4/18)
[+] Building 33.6s (7/18)
```

Each of those new lines is when `mage` tried to log a new S3 upload, but Docker was running and overwrote the terminal.

I really like showing the build output: especially on a fresh deploy, seeing all the docker build steps is helpful.

So here are some options - which would you prefer we do?

1. Keep this change as-is - the clobbered output is weird and maybe confusing, but mostly just aesthetic
2. Change the docker progress output to "plain" (`--progress plain`). This removes all the fancy progress trackers and shows the raw container output. This can intermingle with the `mage` messages well, but the container output is far too verbose (specifically, `webpack.Progress` spits out several thousand lines). If we tune the container output, this could work well
3. Hide the docker output by default, which is what most of you are used to. No progress information on the docker build
4. Run the `docker build` by itself (with default output) _before_ the rest of the `go build` parallel packaging. This will result in slower deploys overall, but will be less resource-intensive

tools/mage/master/deploy.go
```diff
@@ -19,111 +19,79 @@ package master
...
 	"github.com/panther-labs/panther/tools/cfnstacks"
 	"github.com/panther-labs/panther/tools/mage/clients"
 	"github.com/panther-labs/panther/tools/mage/deploy"
 	"github.com/panther-labs/panther/tools/mage/logger"
+	"github.com/panther-labs/panther/tools/mage/pkg"
 	"github.com/panther-labs/panther/tools/mage/util"
 )
 
-const defaultStackName = "panther"
+const devStackName = "panther-dev"
```
kostaspap, line 38: We also have a `defaultRootStackName` in `tools/mage/master/config.go`. Where is each needed?

tools/mage/master/publish.go
```diff
@@ -142,25 +134,42 @@ func getPublicationApproval(log *zap.SugaredLogger, regions []string) error {
...
 	bucket, s3Key, s3URL := s3MasterTemplate(region)
 
-	// Publish S3 assets and ECR docker image
-	ecrRegistry := fmt.Sprintf("349240696275.dkr.ecr.%s.amazonaws.com/panther-community", region)
-	pkg, err := pkgAssets(log, ecr.New(awsSession), region, bucket, ecrRegistry, dockerImageID)
+	packager := pkg.Packager{
+		Log:            log,
+		AwsSession:     awsSession,
+		Bucket:         bucket,
+		EcrRegistry:    fmt.Sprintf("349240696275.dkr.ecr.%s.amazonaws.com/panther-community", region),
```
kostaspap, line 152: Can you add a comment about the account id? It is panther's account that hold the public data, right?

tools/mage/pkg/pkg.go
```diff
@@ -0,0 +1,418 @@
...
+//
+// Supports the following resource types:
+//     AWS::AppSync::GraphQLSchema (DefinitionS3Location)
+//     AWS::CloudFormation::Stack (TemplateURL)
+//     AWS::ECS::TaskDefinition (Image)
+//     AWS::Lambda::LayerVersion (Content)
+//     AWS::Serverless::Function (CodeUri)
+//
+// Returns the path to the packaged template (in the out/ folder)
+func (p Packager) Template(path string) (string, error) {
```
kostaspap, line 114: Better `(p *Packager)` -> https://github.com/golang/go/wiki/CodeReviewComments#receiver-type  . Same for all functions with a receiver.

tools/mage/pkg/pkg.go
```diff
@@ -0,0 +1,418 @@
...
+	// safest approach because we can access just the keys we care about and leave the rest alone.
+	var body map[string]interface{}
+	if err := util.ParseTemplate(path, &body); err != nil {
+		return "", err
+	}
+
+	// Start the worker routines
+	resources := body["Resources"].(map[string]interface{})
+	jobs := make(chan cfnResource, len(resources))
+	results := make(chan cfnResource, len(resources))
```
kostaspap, line 127: nit: Usually safer to close them in a defer statement. 
```
jobs := make(chan cfnResource, len(resources))
defer close(jobs)
results := make(chan cfnResource, len(resources))
defer close(results)
```

Also by locating closing of channel next to its creation, it's more easy to identify cases where a channel could be leaked (I think the code never closed `results` channel)

tools/mage/pkg/pkg.go
```diff
@@ -0,0 +1,418 @@
...
+
+	workers := numWorkers
+	if len(resources) < workers {
+		// Some stacks have very few resources (e.g. aux templates);
+		// no need to spin up workers that won't have anything to do.
+		workers = len(resources)
+	}
+	for w := 1; w <= workers; w++ {
+		go p.resourceWorker(fmt.Sprintf("[%d] %s", w, path), jobs, results)
+	}
```
kostaspap, line 137: In other parts of the system we use the `errgroup.WithContext` pattern - it has the benefit that it cancels running tasks if any task reported an error -> https://github.com/panther-labs/panther/blob/master/cmd/opstools/gluesync/gluesync/main.go#L82

Putting this here would require some refactoring and I don't think it's that important for the code here, just thought of sharing :)

kostaspap (approved):
Nice stuff!

rleighton:
LGTM. Nice work

tools/mage/pkg/pkg.go
austinbyers: Hmm, none of those conditions that they list apply in this case, though. The config struct is small and I do *not* want any of the receivers to be able to edit the config struct. The config should be static / only changed by the caller. That's why I chose not to make it a pointer

tools/mage/master/deploy.go
austinbyers: There are two stacks deployed - the first is the new `panther-dev` stack which contains the S3 bucket / ECR repo. Then the main root stack is deployed, but that stack name is user-configurable (which is why the default is defined in the config.go instead of here)

tools/mage/master/publish.go
austinbyers: Yes, it is Panther's public account (where we publish releases). I'll make the accountID a constant and add a comment

tools/mage/pkg/pkg.go
austinbyers: > it has the benefit that it cancels running tasks if an individual task reported an error

In this code (`mage`), a returned error will immediately be bubbled up and halt the entire program. I assume in that case Go or the OS will automatically kill the other goroutines?

I'll definitely take a look though, thanks for sharing!

deployments/cloud_security.yml
```diff
@@ -130,7 +130,7 @@ Resources:
       # * Failure of this sqs queue will impact delivery of alerts to output destinations.
       # * Failed events will go into the `panther-alert-processor-queue-dlq`. When the system has recovered they should be re-queued to the `panther-alert-processor-queue` using the Panther tool `requeue`.
       # </cfndoc>
-      MessageRetentionPeriod: 1209600 # Max duration - 14 days
+      MessageRetentionPeriod: '1209600' # Max duration - 14 days
```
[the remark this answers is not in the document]
tatehanawalt: good to know

tools/mage/pkg/pkg.go
austinbyers: `errgroup` turned out to be a bit difficult to add here because I'm using return values, not just errors. Each worker returns the modified resource. I suppose alternatively they could modify the shared template with mutex locking, but in the interest of time I'll leave this as is.

But there is a lot of `mage` code that could probably benefit from the `errgroup`, so I'll definitely keep it in mind!

tools/mage/pkg/pkg.go
austinbyers: I'll go ahead and switch to a pointer receiver anyway - future devs may expect to be able to modify the struct via receiver

austinbyers:
I've addressed the feedback but also made a number of larger changes after more testing, so I've asked for another round of reviews.

* I isolated the `docker build`: running `docker build` at the same time as several `go builds` was too much; my laptop started lagging. That's fine, keeping docker by itself also lets us keep the cleaner docker output. The `docker push` still happens in parallel with the other packaging code
* I kept running into issues with S3. Specifically, the `HeadObject` check would keep getting stuck in some regions (but only sometimes), and I have no idea why. So.... I flipped all of the new deploy code to AWS SDK v2 (part of #2467) and it immediately worked much better.

I didn't convert all of `mage` to AWS SDK v2 yet, but the deploy code is the majority of the AWS API calls, so it'll be easy now to finish that migration.

austinbyers:
Another nice feature of this new packaging logic: `mage setup` is no longer required for deployment/publishing, since the `aws-sam-cli` pip library was removed!

Now you can deploy with just Go and `mage`; `mage setup` is only necessary for running tests

tools/mage/pkg/docker.go
```diff
@@ -0,0 +1,131 @@
...
+func dockerLogin(ecrServer, username, password string) error {
+	// We are going to replace Stdin with a pipe reader, so temporarily
+	// cache previous Stdin
+	existingStdin := os.Stdin
+	// Make sure to reset the Stdin.
+	defer func() {
+		os.Stdin = existingStdin
+	}()
+	// Create a pipe to pass docker password to the docker login command
+	pipeReader, pipeWriter, err := os.Pipe()
```
kostaspap, line 108: I believe we should be closing the `pipeReader` maybe in a defer statement. 
```
pipeReader, pipeWriter, err := os.Pipe()
if err != nil {
    return fmt.Errorf("failed to open pipe: %v", err)
}
defer func() {
    pipeReader.Close()
    pipeWriter.Close()
}
```

kostaspap (approved):
This looks good! 
Could you also check if some of the existing commands work e.g. 
```
STACK=log-analysis mage deploy
LAMBDA=panther-log-processor mage deploy
```

This PR is for improving the root stack deployment, I'm suggesting the tests to make sure the changes didn't break existing functionality.

deployments/dev.yml
```diff
@@ -0,0 +1,86 @@
...
+    Properties:
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
+          - ServerSideEncryptionByDefault:
+              SSEAlgorithm: AES256
+      LifecycleConfiguration:
+        Rules:
+          # Once a stack is deployed, its resources in S3 can be safely removed.
+          - ExpirationInDays: 14
+            Status: Enabled
```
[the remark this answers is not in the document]
austinbyers: Hmm good question, I'm not sure - as soon as you deploy, the source is actually sent to AWS Lambda and stored there. S3 is just a temporary staging place.

Immediately after you deploy, you could erase the S3 source bucket and Lambda wouldn't care in the slightest.

But I'm not sure if during a rollback CFN tries to fetch the old S3 source or the old Lambda source, that probably depends on how exactly the serverless transform works. If we explicitly created lambda versions, then the S3 source definitely wouldn't matter; CFN would revert to the old Lambda version snapshot. But since we don't use versioning (because of concerns updating memory/timeout on live deployments), maybe it would fail to rollback

I'll bump it to 30 days to be safe, especially since some customers will be deploying from source indefinitely. I'll make a note to research that rollback situation

requirements-top-level.txt
```diff
@@ -11,15 +11,12 @@
...
 # pip3 freeze --all > requirements.txt
 # deactivate
-aws-sam-cli
 bandit
-boto3~=1.14.40  # Matches version in Lambda: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html
-cfn-lint
+boto3~=1.16.31  # Matches version in Lambda: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html
 cfn-flip
+cfn-lint
 jsonpath-ng
```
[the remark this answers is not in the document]
austinbyers: Panther itself actually doesn't require any pip libraries. The `requirements.txt` is only dev dependencies for running tests (we could rename it `dev-requirements.txt`)

The pip libraries you list are for the default Python layer, which is only applicable when running policies/rules in Panther itself. Developers do not need to install those libraries to test anything in the Panther source code, so no, they do not need to be added to the requirements here.

@tatehanawalt will be working on a better way to manage these layer libraries which will hopefully make that distinction more clear (they'll be passed as parameters to CloudFormation).

tools/mage/pkg/docker.go
austinbyers: Sure, will do (I just moved the code, but happy to fix it)

austinbyers:
> Could you also check if some of the existing commands work e.g.

Yep, the code does change the current `mage deploy` flow as well, so I've been testing both deploy methods and will give it another round of testing before I merge now that the deploy is fixed (#2592 )

austinbyers:
> Now you can deploy with just Go and mage; mage setup is only necessary for running tests

It turned out this wasn't quite true - we still needed `cfn-flip` for parsing YAML templates with short-form CloudFormation functions.

But AWS finally has a Go library which can parse raw CloudFormation! https://pkg.go.dev/github.com/aws-cloudformation/rain/cfn/parse

By switching to that, I was able to remove `cfn-flip` entirely, which will also make packaging slightly faster (we don't have to shell out to Python and we don't have to make so many round trips to disk).

The `rain` library uses v3 of the Go YAML library, so I've switched all of `mage` to yaml v3 to be consistent. I've done many deploys with both methods as well as LAMBDA and STACK specific deploys

deployments/dashboards.yml
```diff
@@ -22,595 +22,595 @@ Resources:
...
-                          "stat": "Sum",
-                          "period": 300,
-                          "start": "-PT12H",
-                          "end": "P0D",
-                          "title": "Input SQS Queue Performance"
-                      }
-                  }
-              ]
-          }
+                    {
```
austinbyers, line 27: The yaml v3 library has a different default indent. It's configurable, but I didn't want to mess with it right now

PR #2548, opened 2021-01-28
ResourceType and LogType validations for policies, roles, and datamodels 

### LINKED ISSUE #619: Validate Log and Resource types
### Description

The Panther backend should validate the Log & Resource Types

### Acceptance Criteria

1. Panther backend validates Log Types
2. Panther backend validates Resource types
3. In case some create/update operation for a rule/policy includes an invalid log or resource type, backend should throw exception.

tatehanawalt:
## Background

Bulk upload or invoke creation of policies, roles, and datamodels doesn't reject the invokation when the create/update object has invalid log or resource types.

This change adds a validation check on log types and resource types before create or modify policies, roles, and datamodels.

## Changes:

Created validations for logtypes and resourcetype in the analysis-api validation.go file. 

Validations are called on sets of log or resource types.

Added validation calls to the bulk upload handler in `tableItemFromConfig`

Added validation calls to create_update_datamodel, create_update_policy, and create_update_role handlers.

Minor modification to the existing integration testing bulk upload resources where a datamodel contained an invalid type.

Moved bulk upload integration testing resources to their own directory in the analysis api main directory.

###### individual item handlers:
The create_update_datamodel handler contained a method 'validateUpdateDataModel'
I have added this method to role and policy with respective names (`validateUpdateRole`, `validateUpdatePolicy`)
Right now the added methods to role and policy only contain their respective type validations. I added these methods for consistency across the 3 type handlers.

## New changes from the previous branch:
• Bulk upload validations occur before allocating the tableitem.
• Updated lambda client to a lambdaiface.
• Moved validations to validation.go.
• Added integration tests for individual invalid type handlers (bulk testing was implemented in the previous branch).
• Removed fmt from bulk upload, create datamodel, create policy, and create role handlers.
• Various nits and syntax improvements. 
• Validation methods in role and policy handlers in a format similar to the existing datamodel validation method

##### Caveats:
Calling the validation before allocating the table item saves some memory. 
The tableItemFromConfig method calls validations before creating a new tableItem. There are also checks on the tableItem. This means we have very similar switch statements before and after allocating a new tableitem.

## Testing:
ci
integrations added for bulk uploads containing resources with at least 1 invalid type each. 
Manual lambda invokation

closes #619


tatehanawalt:
@kostaspap there were a couple comments on the previous pr that haven't been closed yet.

https://github.com/panther-labs/panther/pull/2468#discussion_r566090384
https://github.com/panther-labs/panther/pull/2468#discussion_r566092703

++
@austinbyers 

Other than the two links I think I have the other comments addressed in this new branch. Apologies on the new pr. I had most of the unintended changes removed but with removing some of the files I didn't want to unintentionally remove anyone else's work when merging the old branch back with master.

internal/core/analysis_api/handlers/validation.go
```diff
@@ -0,0 +1,84 @@
...
+	if err := refreshLogTypes(); err != nil {
+		return err
+	}
+	for _, logtype := range logtypes {
+		if !logtypeIsValid(logtype) {
+			return errors.Errorf("%s", logtype)
+		}
+	}
+	return nil
+}
```
kostaspap, line 84: This could be just:

```suggestion
func validateLogtypeSet(logtypes []string) error {
	availableLogTypes, err := logtypesAPI.ListAvailableLogTypes(context.TODO())
	if err != nil {
		return err
	}
	for _, logtype := range logtypes {
		if !stringset.Contains(availableLogTypes.LogTypes, logtype) {
			return errors.Errorf("%s", logtype)
		}
	}
	return nil
}
```

It achieves same functionality, it's also readable and saves few lines of code - we no longer need `refreshLogTypes` ,`logtypeIsValid` functions :)  

The important here is to get rid of the `var logtypeSetMap map[string]struct{}` package variable. We should try to keep package variables like that to a minimum: 
1. Any part of the package code can access them and mutate them. They are very useful sometimes (e.g. for caching values between Lambda invocations) but we should avoid them if possible since they add some code maintenance overhead since we need to make sure that e.g. no two goroutines try to modify them at the same time. 
2. They make unit testing challenging since unit tests running in parallel can modify them in parallel and lead to panics/failed tests.

internal/core/analysis_api/main/integration_test.go
```diff
@@ -748,6 +767,34 @@ func createPolicySuccess(t *testing.T) {
...
+			Suppressions: policy.Suppressions,
+			Tags:         policy.Tags,
+			Tests:        policy.Tests,
+			UserID:       userID,
+		},
+	}
+	var result models.Policy
+	statusCode, err := apiClient.Invoke(&input, &result)
+	assert.Equal(t, http.StatusBadRequest, statusCode)
+	assert.Error(t, err)
```
kostaspap, line 795: nit: Since the error is visible to the customers, does it makes sense to validate it as well?

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -388,18 +402,22 @@ func validateUploadedDataModel(item *tableItem) error {
 func validateUploadedPolicy(item *tableItem) error {
 	switch item.Type {
 	case models.TypeGlobal:
+		if len(item.ResourceTypes) > 1 {
+			return errors.New("only one LogType may be specified per DataModel")
```
kostaspap, line 406: A made a similar [comment](https://github.com/panther-labs/panther/pull/2468#discussion_r566090847) in the other PR: 
This case is for `models.TypeGlobal` but your error message is talking about `DataModel`. This doesn't seem accurate? 

If we should have such a condition for Data models, this should be in the `models.TypeDataModel` case cc @lindsey-w

internal/core/analysis_api/handlers/clients.go
```diff
@@ -74,4 +86,17 @@ func Setup() {
...
+			aws.NewConfig().WithMaxRetries(maxRetries),
+			awsretry.NewConnectionErrRetryer(maxRetries),
+		),
+	)
+
+	lambdaLogTypesClient = lambda.New(logtypesClientsSession)
+	logtypesAPI = &logtypesapi.LogTypesAPILambdaClient{
+		LambdaName: logtypesapi.LambdaName,
+		LambdaAPI:  lambdaLogTypesClient,
+	}
```
kostaspap, line 101: Continuing discussion from [here](https://github.com/panther-labs/panther/pull/2468#discussion_r566472734): 

This could be just:
```
logtypesAPI = &logtypesapi.LogTypesAPILambdaClient{
		LambdaName: logtypesapi.LambdaName,
		LambdaAPI:  lambda.New(logtypesClientsSession),
	}
```

The `lambdaLogTypesClient lambdaiface.LambdaAPI` declared in line 59 is used only here, it doesn't have to be package visible. We should promote visibility of variables to package or public only if needed for the reason I explained in a similar comment in this PR.

internal/core/analysis_api/handlers/validation.go
kostaspap: Also note the use of `context.TODO()` vs `context.Background()` used in the `refreshLogTypes()`. 

From `context.TODO()` docs: 
<pre>
// TODO returns a non-nil, empty Context. Code <b>should use context.TODO</b> when
// it's unclear which Context to use or <b>it is not yet available (because the
// surrounding function has not yet been extended to accept a Context
// parameter)</b>
</pre>

kostaspap (commented):
Good job! LGTM, I have only few comments, mostly style, only one regarding functionality.

internal/core/analysis_api/handlers/bulk_upload.go
lindsey-w: Yea, I think this might need to be moved under the data model case

internal/core/analysis_api/main/integration_test.go
austinbyers: Meaning validate the error message itself? Yeah, good idea.

E.g.

```go
require.Error(t, err)
assert.Equal(err, "expected error msg", err.Error())
```

internal/core/analysis_api/handlers/bulk_upload.go
austinbyers: Globals do not have associated resource/log types, no extra validation is needed.

This does indeed need to move to under the data model

internal/core/analysis_api/handlers/validation.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"context"
+
+	"github.com/pkg/errors"
+
+	resourceTypesProvider "github.com/panther-labs/panther/internal/compliance/snapshot_poller/models/aws"
+)
+
+// Populated by a call to refreshLogTypes
+var logtypeSetMap map[string]struct{}
```
austinbyers, line 30: Since you're not caching this, you don't need it to be a package variable

internal/core/analysis_api/handlers/validation.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	logtypeSetMap = make(map[string]struct{})
+	for _, logtype := range logtypes.LogTypes {
+		logtypeSetMap[logtype] = struct{}{}
+	}
+	return nil
+}
+
+// Return the existence of the passed logtype in the current logtypes.
+// NOTE: Accuret results require an updated logtypeSetMap - currently accomplished using the call to
```
austinbyers, line 62: Typo: `accurate` instead of `accuret`

internal/core/analysis_api/main/integration_test.go
austinbyers: Or `assert.Contains(t, err.Error(), "expected error substring")` to not be so brittle

austinbyers (commented):
Getting there! I agree with Kostas' comments about style, but the implementation looks good

internal/core/analysis_api/handlers/validation.go
tatehanawalt: Thanks for the context tip and I like the syntax code recommendation

kostaspap (approved):
Pre-approving the PR, really happy we do these validations

internal/core/analysis_api/handlers/validation.go
tatehanawalt: @kostaspap are you referring to the "github.com/jjeffery/stringset" package?

austinbyers: We actually have our own `pkg/stringset`, but our implementation is a `[]string`, which is very odd, I don't know why we have that, to be honest. Seems very inefficient

My preference - stick to `map[string]struct{}` rather than importing an extra unnecessary package

tatehanawalt: I could go both ways I like the syntax of @kostaspap and I very much like limiting external dependencies.

@austinbyers I put this together while I was waiting to hear what package to use, is this what you are thinking? I also have @kostaspap  suggestion working. 
```
	logtypeSetMap := make(map[string]struct{})
	for _, logtype := range availableLogTypes.LogTypes {
		logtypeSetMap[logtype] = struct{}{}
	}
	for _, lt := range logtypes {
		if _, found := logtypeSetMap[lt]; !found {
			return errors.Errorf("%s", lt)
		}
	}
```

austinbyers: Yes, you've got the right idea. A local map in which you test inclusion 👍

internal/core/analysis_api/handlers/clients.go
tatehanawalt: I see what you mean now, updating to the above in the next push

internal/core/analysis_api/handlers/validation.go
```diff
@@ -0,0 +1,67 @@
...
+
+	resourceTypesProvider "github.com/panther-labs/panther/internal/compliance/snapshot_poller/models/aws"
+)
+
+// Traverse a passed set of resource and return an error if any of them are not found in the current
+// list of valid resource types
+//
+// CAVEAT: This method uses a hardcoded list of existing resource types. If this method is returning
+// unexpected errors the hardcoded list is up to date.
+func ValidResourceTypeSet(checkResourceTypeSet []string) error {
```
austinbyers, line 34: This function is exported but `validateLogtypeSet` is not

Let's be consistent; if they don't need to be exported, then let's not

internal/core/analysis_api/handlers/validation.go
```diff
@@ -0,0 +1,67 @@
...
+			return errors.Errorf("%s", writeResourceTypeEntry)
+		}
+	}
+	return nil
+}
+
+// Retrieve a set of log types from the logtypes api and validate every entry in the passed set
+// is a value found in the logtypes-api returned set
+//
+// CAVEAT: This method will trigger a request to the log-types api EVERY time it is called.
```
austinbyers, line 47: Does that mean this is called for every single rule during a bulk upload? That's going to be too much.

I know I said we don't need caching, but we definitely don't want a bulk upload to be making a hundred API calls. Find some way to avoid that:

* You could have a separate function fetch the log type map. Call that function once during the bulk upload, and use that same map for validation every time
* Cache the `logtypeSetMap` globally for 60 seconds or so

internal/core/analysis_api/main/integration_test.go
```diff
@@ -748,6 +767,37 @@ func createPolicySuccess(t *testing.T) {
...
+
+	//nolint
```
austinbyers, line 797: `// nolint: lll` to disable just the specific linter rather than all of them

Or, even better, use string concatenation instead of having a really long line

austinbyers (commented):
I have a concern about the bulk upload

internal/core/analysis_api/handlers/validation.go
tatehanawalt: I had a refresh method in the initial commit. theres some minor updates but its pretty close to a copy and paste to re add it
```
// Request the logtypes-api for the current set of logtypes and assign the result list to 'logtypeSetMap'
func refreshLogTypes() {
	// Temporary get log types for testing
	logtypes, err := logtypesAPI.ListAvailableLogTypes(context.Background())
	if err != nil {
		return
	}

	logtypeSetMap = make(map[string]interface{})
	for _, logtype := range logtypes.LogTypes {
		logtypeSetMap[logtype] = nil
	}
}
```

internal/core/analysis_api/main/integration_test.go
tatehanawalt: also just now seeing the assert.Equal params of expected, actual are swapped there. Already removed the nolints but i'm gonna fix that assertion

internal/core/analysis_api/handlers/validation.go
austinbyers: We try to avoid global state going forward, so I think the first option (just separate the log type lookup so the bulk upload can pre-fetch them) is probably preferable. If you keep global state, you're going to have to check the time on it and expire the cache

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -251,14 +260,88 @@ func extractZipFile(input *models.BulkUploadInput) (map[string]*tableItem, error
...
+	case models.TypeDataModel, models.TypeRule:
+		if len(resourceTypes) == 0 {
+			resourceTypes = config.LogTypes
+		}
+
+		// This is essentially a duplicate of the validateLogtypeSet method in validations.go
+		// In bulk upload we only refresh the set of logtypes once. Otherwise we would refresh the set
+		// for each log type in the bulk set.
+		for _, rscType := range resourceTypes {
+			if _, found := logtypes[rscType]; !found {
```
austinbyers, line 331: Nit: we try to avoid global variables going forward unless there's a good reason for them. This looks like the only place the global variable is read, can we pass it as a function parameter instead?

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -251,14 +260,88 @@ func extractZipFile(input *models.BulkUploadInput) (map[string]*tableItem, error
...
+	var resourceTypes []string = config.ResourceTypes
+
+	// Validations for valid log and resource types
+	switch itemType {
+	case models.TypeDataModel, models.TypeRule:
+		if len(resourceTypes) == 0 {
+			resourceTypes = config.LogTypes
+		}
+
+		// This is essentially a duplicate of the validateLogtypeSet method in validations.go
```
austinbyers, line 327: Instead of duplicating the code, you could just adapt `validateLogtypeSet` to accept an existing `logtypes` map as an optional parameter.

BulkUpload would send that argument, the others would leave it `nil`. The `validatLogtypeSet` would look up the map if it's not already defined

austinbyers (approved):
Looks good! Minor style comment about the global and the code duplication, but fine either way

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -251,13 +260,85 @@ func extractZipFile(input *models.BulkUploadInput) (map[string]*tableItem, error
...
+	return models.UnitTest{
+		ExpectedResult: test.ExpectedResult,
+		Name:           test.Name,
+		Resource:       log,
+	}, err
+}
+
+// Validate the analysis item's Resource type or Log type depending on the config AnalysisType.
+// Passing the logtypes allows us to retrieve the set of valid log types once for a set of validations
+func bulkValidateLogAndResourceTypes(config analysis.Config, logtypes map[string]struct{}) error {
```
tatehanawalt, line 305: @austinbyers tagging for review - sending you a slack now but this achieves eliminating the global state

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -251,13 +260,85 @@ func extractZipFile(input *models.BulkUploadInput) (map[string]*tableItem, error
...
+func readZipFile(zf *zip.File) ([]byte, error) {
+	f, err := zf.Open()
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err := f.Close(); err != nil {
+			zap.L().Error("error closing zip file", zap.Error(err))
+		}
+	}()
```
austinbyers, line 338: If the zip file can't be closed for some reason, we don't really care that much. Log a warning at most, or just don't even worry about it:

```go
defer f.Close()
```

But we definitely don't need to be paged for a file that can't close since we already read it.

One advantage of running in Lambda is that containers are transient

----

I know you just moved this code, but might as well fix it while you're in here :)

internal/core/analysis_api/handlers/validation.go
```diff
@@ -0,0 +1,80 @@
...
+		return errors.Errorf("%s", *firstMissing)
+	}
+	return nil
+}
+
+// Returns the first set entry not found as a key in the searchMap
+func FirstSetItemNotInMapKeys(itemSet []string, searchMap map[string]struct{}) *string {
+	for _, item := range itemSet {
+		if _, found := searchMap[item]; !found {
+			return &item
```
austinbyers, line 76: Using the address of a loop variable can cause problems, let's do `return aws.String(item)` here to be safe (to guarantee a fresh heap allocation).

Or just return a string instead of a `*string`

internal/core/analysis_api/handlers/bulk_upload.go
```diff
@@ -156,18 +155,23 @@ func extractZipFile(input *models.BulkUploadInput) (map[string]*tableItem, error
...
 	zipReader, err := zip.NewReader(bytes.NewReader(content), int64(len(content)))
 	if err != nil {
-		return nil, fmt.Errorf("zipReader failed: %s", err)
+		return nil, errors.Errorf("zipReader failed: %s", err)
 	}
 
 	policyBodies := make(map[string]string) // map base file name to contents
 	result := make(map[string]*tableItem)
 
+	logtypes, err := getLogTypesSet()
```
austinbyers, line 170: 👍

austinbyers (approved):
Nice job! This will be really helpful for both us and customers

PR #2549, opened 2021-01-28
RegionIgnoreListError TypeAssertionError 

wey-chiang:
## Background

While investigating problems in AWS, the following error message was found
`interface conversion: error is *errors.withStack, not *aws.RegionIgnoreListError: TypeAssertionError`
This PR is meant to address that type assertion.

## Changes

- Changed logic from `if err == err.(*RegionIgnoreListError)` to `if errors.As(err, &e)`
- Added test case to s3_bucket_test.go to make sure the `RegionIgnoreListError` is returned when region is ignored.

## Testing

- `mage test:ci`


internal/compliance/snapshot_poller/pollers/aws/poll.go
```diff
@@ -374,7 +374,8 @@ func singleResourceScan(
 
 	if err != nil {
 		// Check for region ignoreList error
-		if err == err.(*RegionIgnoreListError) {
+		var e *RegionIgnoreListError
+		if errors.As(err, &e) {
```
austinbyers, line 378: The `&e` is now a double pointer (`**RegionIgnoreListError`). That might work, but probably not what you want.

```suggestion
		var e RegionIgnoreListError
		if errors.As(err, &e) {
```

internal/compliance/snapshot_poller/pollers/aws/s3_bucket_test.go
```diff
@@ -277,3 +279,25 @@ func TestS3BucketPollerError(t *testing.T) {
 	assert.Nil(t, marker)
 	assert.Error(t, err)
 }
+
+func TestS3BucketPollerRegionIgnoreListError(t *testing.T) {
```
kostaspap, line 283: This test passes even if the change in the `poll.go` is commented out.

Please first create a test that will trigger the same error we've seen in production, then write the code to fix it.

kostaspap (changes requested):

internal/compliance/snapshot_poller/pollers/aws/s3_bucket_test.go
austinbyers: Good catch!

internal/compliance/snapshot_poller/pollers/aws/poll.go
wey-chiang: So upon second review, it appears that I need to use a double pointer here. Following the [Go Blog](https://blog.golang.org/go1.13-errors): 

> `errors.As` will panic if the target is not a non-nil pointer to either a type that implements error, or to any interface type. 

Instantiating `e` as `var e RegionIgnoreListError` causes `errors` to complain: `*target must be interface or implement error`

austinbyers (approved):

internal/compliance/snapshot_poller/pollers/aws/s3_bucket_test.go
wey-chiang: > This test passes even if the change in the poll.go is commented out.

Doh! 😓  Good catch, this should no longer hold true with the latest set of commits.

internal/compliance/snapshot_poller/pollers/aws/poll_test.go
```diff
@@ -21,10 +21,93 @@ package aws
...
+
+	RateLimitTracker, _ = lru.NewARC(10)
+	KmsClientFunc = awstest.SetupMockKms
+
+	// Provide stub poller session
+	SnapshotPollerSession = &session.Session{}
+
+	_, err := singleResourceScan(sampleScanRequest, sampleResourcePollerInput)
+
+	assert.Equal(t, nil, err)
```
kostaspap, line 70: nit: 
```suggestion
	assert.NoError(t, err)
```

kostaspap (approved):

PR #2556, opened 2021-01-29
[Feature] [FE]: Unified detections page [1/2]

### LINKED ISSUE #1078: Allow filtering of Rules based on Creator

Currently, rules  can be filtered  by many fields, but it's still hard to know who added what. Specifically, Panther has a lot of pre-created rules, but users can also add their own. Rules created by Panther itself co-exist  with the ones created by the users themselves, so we need a way to filter based on whether the role was autogenerated or not.

As a first step, we should allow to quickly filter based on  whether the rule was created by Panther or not. If that  turns out to be easy, we can also enhance this functionality by allowing filtering for each individual user.

The exact same thing needs to happen for the Policies page (in Cloud Security)

### Describe your use cases

Organizations want to see what they themselves have added VS what was already there

Weekly

Alternatives is to sort  by creation date. The rules at the "top" are the custom ones, created by members of the organization

### Additional context

I personally  think that the quickest & biggest value will be created by providing a simple filter to separate organization-created Rules vs Panther pre-created rules. Filtering by individual users is nice, but can **potentially** be added later down the line.

3nvi:
## Background

This PR implements the core logic for the detections listing page, allowing a user to view all policies & rules in a single unified page. This PR includes listing & filtering, but excludes the mass-actions/mass-selections of cards, which will come in a later PR.

Closes #1821 
Closes #1078 

## Designs

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/df9d59c8-eeb1-452d-a5f5-3712b64010be/collections/c02562f3-dedb-4408-85f6-512a4c981477

## Screenshots

Screen Shot 2021-01-29 at 5 00 02 PM
Screen Shot 2021-01-29 at 5 00 08 PM

## Changes

- Create `ListDetections` page
- Remove `ListRules` page
- Remove `ListPolicies` page
- Update graphql cache modifications

## Testing

- `npm run test`


web/codegen/buildGraphqlResourcesPlugin.js
```diff
@@ -163,6 +162,17 @@ exports.plugin = (schema, documents, config) => {
...
+      const typeName = node.name.value;
+      const { fields } = node;
+      return {
+        typeName,
+        mockFn: () => {
+          const mockFields = fields ? fields.map(({ mockFn }) => mockFn(typeName)).join('\n') : '';
+          return createGraphQLResourceFunction(typeName, mockFields, false);
+        },
+      };
+    },
```
3nvi, line 175: our script that generates fixtures didn't have support for GraphQL interfaces. With this change, we make sure we create fixtures (fake data) for our `Interface` definitions as well 😃

web/src/components/badges/StatusBadge/StatusBadge.tsx
```diff
@@ -21,8 +21,8 @@ import { Badge, BadgeProps, Box, Tooltip } from 'pouncejs';
 import { ComplianceStatusEnum } from 'Generated/schema';
 
 export const STATUS_COLOR_MAP: { [key in StatusBadgeProps['status']]: BadgeProps['color'] } = {
-  [ComplianceStatusEnum.Pass]: 'green-400' as const,
-  [ComplianceStatusEnum.Fail]: 'red-300' as const,
+  [ComplianceStatusEnum.Pass]: 'green-500' as const,
+  [ComplianceStatusEnum.Fail]: 'red-200' as const,
   [ComplianceStatusEnum.Error]: 'orange-400' as const,
```
3nvi, line 26: colors have changed to something more "pastel"-like

web/src/components/buttons/LinkButton/LinkButton.tsx
```diff
@@ -16,42 +16,18 @@
...
+import { ButtonProps, Button } from 'pouncejs';
+import LinkButtonWrapper, { LinkButtonWrapperProps } from '../utils/LinkButtonWrapper';
 
-type ButtonWithoutAs = Omit<ButtonProps, 'as'>;
-type ToProp = Pick<RRLinkProps, 'to'>;
-export type LinkButtonProps = ButtonWithoutAs & ToProp & { external?: boolean };
+type LinkButtonProps = LinkButtonWrapperProps & Omit<ButtonProps, 'as'>;
 
 const LinkButton: React.FC<LinkButtonProps> = ({ disabled, external, to, children, ...rest }) => {
-  let linkProps: LinkProps;
```
3nvi, line 27: we need an `Link-IconButton` that looks like `LinkButton` but uses an `IconButton` underneath the hood, rather than a `Button`.

We abstract the necessary "Link" part of it (in a component named `LinkWrapper`, so we can use it to create a `LinkButton` and `LinkIconButton`

web/src/components/cards/RuleCard/RuleCard.tsx
```diff
@@ -56,7 +56,9 @@ const RuleCard: React.FC<RuleCardProps> = ({ rule }) => {
           <GenericItemCard.Date date={formatDatetime(rule.lastModified)} />
           <RuleCardOptions rule={rule} />
         </GenericItemCard.Header>
-
+        <Text fontSize="small" as="span" color="cyan-500">
+          Rule
+        </Text>
```
3nvi, line 61: Designs called for that

web/src/pages/CreateDetection/CreatePolicy/CreatePolicy.tsx
```diff
@@ -49,7 +48,6 @@ export const initialValues: Required<AddPolicyInput> = {
 const CreatePolicy: React.FC = () => {
   const { history } = useRouter();
   const [createPolicy, { error }] = useCreatePolicy({
-    refetchQueries: [{ query: ListPoliciesDocument, variables: { input: {} } }],
```
3nvi, line 52: no need to refetch this anymore since we don't have a dedicated policies page

web/src/pages/CreateDetection/CreateRule/CreateRule.tsx
```diff
@@ -53,7 +52,6 @@ export const initialValues: Required<AddRuleInput> = {
 const CreateRule: React.FC = () => {
   const { history } = useRouter();
   const [createRule, { error }] = useCreateRule({
-    refetchQueries: [{ query: ListRulesDocument, variables: { input: {} } }],
```
3nvi, line 56: same as above

api/graphql/schema.graphql
```diff
@@ -53,6 +53,7 @@ type Mutation {
 type Query {
   alert(input: GetAlertInput!): AlertDetails
   alerts(input: ListAlertsInput): ListAlertsResponse
+  detections(input: ListDetectionsInput): ListDetectionsResponse!
```
alexmylonas, line 56:
```suggestion
  detections(input: ListDetectionsInput!): ListDetectionsResponse!
```

api/graphql/schema.graphql
```diff
@@ -87,6 +88,30 @@ type Query {
...
+  createdBy: String
+  lastModifiedBy: String
+  initialSet: Boolean
+  tags: [String!]
+  sortBy: ListDetectionsSortFieldsEnum # defaults to `id`
+  sortDir: SortDirEnum # defaults to `ascending`
+  pageSize: Int # defaults to `25`
+  page: Int # defaults to `1`
+}
+
```
alexmylonas, line 114: Removed unused queries and inputs for list policies and rules?

web/src/components/cards/PolicyCard/PolicyCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+                <RelatedDestinations
+                  destinations={detectionDestinations}
+                  loading={loadingDetectionDestinations}
+                />
+              }
+            />
+          </GenericItemCard.ValuesGroup>
+          <GenericItemCard.ValuesGroup>
+            <Flex ml="auto" mr={0} align="flex-end" spacing={4}>
+              <StatusBadge status={policy.complianceStatus} />
```
alexmylonas, line 80:
```suggestion
  
```
I don't see this `PASS` badge on designs.

web/src/helpers/analytics.ts
```diff
@@ -51,6 +51,7 @@ export enum PageViewEnum {
   ComplianceOverview = 'Compliance Overview',
   ListRules = 'List Rules',
```
alexmylonas, line 52:
```suggestion
```

web/src/urls.ts
```diff
@@ -36,13 +36,14 @@ const urlEncode = (str: string) => encodeURIComponent(str).replace(/%3A/g, unesc
...
     home: () => '/detections/',
+    list: () => urls.detections.home(),
     create: () => `${urls.detections.home()}new/`,
   },
   compliance: {
     home: () => '/cloud-security/',
     overview: () => `${urls.compliance.home()}overview/`,
     policies: {
-      list: () => `${urls.compliance.home()}policies/`,
+      list: () => `${urls.detections.list()}policies/`,
```
alexmylonas, line 46: I am not sure why we should keep those links under policies and rules

We should remove them imo.

alexmylonas (commented):
Amazing work Aggelos! Deployed and validated it works well!

I have a couple of minor comments but nothing too important

api/graphql/schema.graphql
3nvi: We want to be able to get detections without providing **any filters at all**. I believe optional is the best way to go here :/

api/graphql/schema.graphql
3nvi: ah fair!

web/src/components/cards/PolicyCard/PolicyCard.tsx
3nvi: Yeah I asked Jim to add it, since, without it, there's no way for someone to know if a policy detection is failing or not (you currently **can** on the table).

Nice catch though!

web/src/helpers/analytics.ts
3nvi: 👍

web/src/urls.ts
3nvi: Short Answer: We need them.

When you go to `/detections/123` we don't know if `123` is a rule or a policy (to load the appropriate React component). Thus we do it like do `/detections/policies/123` so we can know that `123` is a policy. 

Now in the breadcrumbs, you'll see the options `Detections > Policies > 123` and `Policies` is clickable ,so it has to redirect somewhere. To make this redirect possible, we add a `<Redirect>` component in `routes.tsx`. To configure it correctly,  we need the `from` value that's `urls.policies.list()`

web/src/components/cards/PolicyCard/PolicyCard.tsx
alexmylonas: Gotcha

web/src/urls.ts
alexmylonas: And I am guessing we cant simply query by Id and let the BE understand if it's a policy or a rule. 
I am covered thank you!

alexmylonas (approved):
Thanks for addressing my comments, approving joyfully!

PR #2557, opened 2021-01-29
[FE][BUGFIX]: Prevent crashing when a destination override has been deleted

3nvi:
## Background

Currently if you assign a destination override to a rule and then you deleted this destination, the entire Rules page crashes. This is because of a bug in our FE code that never expects a Rule to contain an `outputId` for a Destination that is not present in our DB.

This PR addressed that by filtering out "dead destinations" and returning a "Not Configured" message to the user

## Screenshots

Screen Shot 2021-01-27 at 12 06 01 PM

## Changes

- Filter out nullish values from the detection <-> destination overrides association

## Testing

- `npm run test`


austinbyers (approved):
Thanks for the quick fix! 🎉

austinbyers:
I'm not sure about the `sync-to-release` label though - we're not going to do a 1.15.3 patch for this; 1.16 is already starting next week. The 1.16 branch doesn't exist yet, so no need to sync to release

vorillaz (approved):
Thanks for this one.

Issue #2560, opened 2021-01-29
Panther Rules Engine CFN Parameter for Attached Managed Policies ARN


horakj:
### Description

Adding a parameter in the root stack which will allow the ability to attach managed IAM Policies (i.e adding S3 and SecretsManager read permissions to the rules engine IAM role). This adds the ability to use the full power of Python such as making boto3 calls for threat intel lists in that are in S3 or storing an API Token in SecretsManager which can be used to make third party API calls.

### Related Services

Panther Rules Engine

### Designs

Example Managed Policy which has an ARN

RulesEngineAssumeReadOnlyPolicy:
    Type: AWS::IAM::ManagedPolicy
    Properties:
      ManagedPolicyName: panther-rule-engine-cross-account-readonly-permissions
      Description: Panther Rules Engine Cross Account Read-only Managed policy
      PolicyDocument:
        Version: '2012-10-17'
        Statement:
          - Effect: Allow
            Action: sts:AssumeRole
            Resource: !Sub arn:aws:iam::${TargetAccountID}:role/PantherRulesEngineCrossAccountReadOnlyRole
          - Effect: Allow
            Action:
              - secretsmanager:GetResourcePolicy
              - secretsmanager:GetSecretValue
              - secretsmanager:DescribeSecret
              - secretsmanager:ListSecretVersionIds
            Resource: arn:aws:secretsmanager:*
      Roles:
        - !Ref PantherRulesEngineRoleName

### Acceptance Criteria

- A CFN Parameter that allows for managed policy ARNs to be added the Rules Engine IAM role
- Cross Account CFN's which can be deployed on top of the Panther Stacks

@mattthetraveler @nhakmiller

PR #2561, opened 2021-01-29
Panther Rules Engine Cross Account IAM Permissions CFNS

horakj:
## Background

Customers have requested a way to create dynamic lists or make third part API calls which can be used in the Rules engine in Panther. This method will create the cross account access need to make the request happen. 

Full resolve below. This will work right now.

Issue: Panther Rules Engine CFN Parameter for Attached Managed Policies ARN #2560 

## Changes

- 1 CFN for a Target or Customer account which creates an IAM role that can be assumed by the Panther Rules engine for read-only s3 and Secrets Manager access
- 1 CFN for the Panther Account Rules Engine IAM role which can assume the role created above

## Testing

- Deployed in Dev Account
- Created Rule that uses boto3's sts and s3 api calls


deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
```diff
@@ -0,0 +1,66 @@
...
+      Policies:
+        - PolicyName: panther-rule-engine-cross-account-read-only-role
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action:
+                  - s3:Get*
+                  - s3:List*
+                Resource: !Sub arn:aws:s3:::${TargetBucketName}/*
```
austinbyers, line 52: The `s3:ListBucket` permission actually [requires](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html#amazons3-actions-as-permissions) just the bucket resource. So you would need:

```yml
- Effect: Allow
  Action:
    - s3:Get*
    - s3:List*
  Resource:
    - !Sub arn:aws:s3:::${TargetBucketName}
    - !Sub arn:aws:s3:::${TargetBucketName}/*
```

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
```diff
@@ -0,0 +1,66 @@
...
+Parameters:
+  PantherMasterAccountID:
+    Type: String
+    Description: Enter the 12 Digit AWS Panther Master AccountID located in Panther under General > Settings.
+  PantherRulesEngineRoleName:
+    Type: String
+    Description: Enter the name of the Panther Rules Engine Role Name which can be found under IAM by searching "rules" in the Panther Master Account.
+  TargetBucketName:
+    Type: String
+    Description: Enter the S3 Bucket that the Panther Rules Engine can Read from.
```
austinbyers, line 29: If the bucket is encrypted with KMS, the user will also need to provide the KMS key ID. You can add that as a `TODO` if we don't need it right now

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
```diff
@@ -0,0 +1,66 @@
...
+                Resource: !Sub arn:aws:s3:::${TargetBucketName}/*
+              - Effect: Allow
+                Action:
+                  - secretsmanager:GetResourcePolicy
+                  - secretsmanager:GetSecretValue
+                  - secretsmanager:DescribeSecret
+                  - secretsmanager:ListSecretVersionIds
+                Resource: arn:aws:secretsmanager:*
+
+      RoleName: PantherRulesEngineCrossAccountReadOnlyRole
```
austinbyers, line 61: IAM is a global namespace, so hardcoding the role name would mean it would only work in one region. Perhaps we can do

```yml
RoleName: !Sub PantherRulesEngineCrossAccountReadOnlyRole-${AWS::Region}
```

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
```diff
@@ -0,0 +1,66 @@
...
+                  - s3:Get*
+                  - s3:List*
+                Resource: !Sub arn:aws:s3:::${TargetBucketName}/*
+              - Effect: Allow
+                Action:
+                  - secretsmanager:GetResourcePolicy
+                  - secretsmanager:GetSecretValue
+                  - secretsmanager:DescribeSecret
+                  - secretsmanager:ListSecretVersionIds
+                Resource: arn:aws:secretsmanager:*
```
austinbyers, line 59: This would allow fetching every secret in the account! That's probably far too broad - there should be a secretID that can be passed in instead

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-panther.yml
```diff
@@ -0,0 +1,53 @@
...
+          - Effect: Allow
+            Action: sts:AssumeRole
+            Resource: !Sub arn:aws:iam::${TargetAccountID}:role/PantherRulesEngineCrossAccountReadOnlyRole
+          - Effect: Allow
+            Action:
+              - secretsmanager:GetResourcePolicy
+              - secretsmanager:GetSecretValue
+              - secretsmanager:DescribeSecret
+              - secretsmanager:ListSecretVersionIds
+            Resource: arn:aws:secretsmanager:*
```
austinbyers, line 46: Why do we need these permissions directly? They're already part of the role we're assuming

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-panther.yml
```diff
@@ -0,0 +1,53 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+AWSTemplateFormatVersion: 2010-09-09
+Description: Panther Rules Engine Read-only Cross Account Assume Role and Secrets Manager Managed Policy for Panther Account
```
austinbyers, line 18: This stack is meant to be deployed in Panther's account?

I know we're trying to get a solution as fast as possible, but this should really be a setting passed through to the main panther stacks. Otherwise, we'll have to deploy this as a separate template for every customer that needs it, and we'll have two different templates that are affecting the same IAM role, which means they could clobber each other.

Tagging @nhakmiller and @s0l0ist to get their input as well - is this a sufficient solution in the short term?

----

I think it's actually pretty simple to do this the "correct" way:

* Add a top-level `AssumablePythonRoles` parameter in the root stack (`master.yml`). This would be an optional list of IAM roles that both of the python engines are allowed to assume
    * Or we could have different parameters for policy vs rules engine
* Pass that parameter to the `log-analysis` stack (rules-engine) and the `cloud-security` stack (policy-engine)
* Add an inline policy to the rules-engine / policy-engine which allows assuming the roles if one has been defined:

```yml
Statement:
   - !If
     - AssumablePythonRolesDefined
     - Effect: Allow
       Action: sts:AssumeRole
       Resource: !Ref AssumablePythonRoles
     - !Ref AWS::NoValue
```

This way, we can manage this setting the same as all of our other settings on a user deployment. I know you may need help with this, perhaps @s0l0ist can help with that implementation

austinbyers (changes requested):
Thanks for putting this together! You're becoming a CloudFormation expert in no time

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
horakj: Thank you, will add that!

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
horakj: Nice, Thank you Ill update that.

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
horakj: Easy fix, wil do.

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-panther.yml
horakj: The logic here is in case the customer wants to give us a secret to store in the panther account (SaaS) this would allow for quicker processing. Thoughts?

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-panther.yml
horakj: Yes I would be happy for any help that would be on the main stack. This PR was the now solution but the above is the correct. Should I drop the this PR for Aux templates or keep it?

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-panther.yml
austinbyers: If we're going with the assume role strategy, usually all of the permissions are attached to the role you assume, and only that role. Then this role only allows `sts:AssumeRole`

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-panther.yml
austinbyers: >  Should I drop the this PR for Aux templates or keep it?

We still need the customer template for sure. The one for Panther I think we should combine with #2217 - pass in a list of roles the engines are allowed to assume. I'll let @nhakmiller make the final call as to how to proceed with the Panther side - I can help next week if needed

horakj: I updated both templates to reflect the changes, if good we can proceed with figuring out how to combine with the issue above

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
horakj: I am going to put that as a TODO for now, TY sir!

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
```diff
@@ -0,0 +1,75 @@
...
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action:
+                  - s3:Get*
+                  - s3:List*
+                Resource:
+                  - !Sub arn:aws:s3:::${TargetBucketName}
+                  - !Sub arn:aws:s3:::${TargetBucketName}/*
```
austinbyers, line 58: Use `:${AWS::Partition}:` instead of `:aws:`, again

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
```diff
@@ -0,0 +1,75 @@
...
+    Type: String
+    Default: NoSecret
+    Description: Enter the name of the Secret in Secrets Manger that will be accessible from the Panther Rules Engine.
+  TargetBucketName:
+    Type: String
+    Description: Enter the S3 Bucket that the Panther Rules Engine can Read from.
+
+Resources:
+  PantherRulesEngineIamRole:
+    Type: 'AWS::IAM::Role'
```
austinbyers, line 37: Nit: we don't need quotes here (nor on the version)

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
austinbyers: I don't see the TODO comment... can we add it?

austinbyers (approved):
So here's how we'll move forward (I synced with @nhakmiller )

I will finish https://github.com/panther-labs/panther/pull/2217 - we're going to add a configuration option to Panther which allows users to pass in an arbitrary list of managed IAM policies.

-----

So I think you've got the right setup:

1. The customer deploys an IAM role in their target account which Panther can assume
2. We deploy a managed policy in the Panther account which is allowed to assume that role (or do anything else)
3. The managed policy is passed as a configuration option to Panther. We just redeploy Panther to re-attach the policy

So you've got steps 1 and 2 in this PR and I'll work on step 3 in the [other PR](https://github.com/panther-labs/panther/pull/2217)

But the important point is that we will *not* be manually changing Panther infrastructure. Manually attaching a managed policy to a Panther deployment will be problematic

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
```diff
@@ -0,0 +1,75 @@
...
+Parameters:
+  PantherMasterAccountID:
+    Type: String
+    Description: Enter the 12 Digit AWS Panther Master AccountID located in Panther under General > Settings.
+  PantherRulesEngineRoleName:
+    Type: String
+    Description: Enter the name of the Panther Rules Engine Role name which can be found in the AWS Console under IAM by searching "rules" in the Panther Master Account.
+  SecretName:
+    Type: String
+    Default: NoSecret
```
austinbyers, line 29: The better way to do this would be to have an empty default string and then only attach an IAM statement if a secret is defined

austinbyers: (I'll get that change for you)

deployments/auxiliary/cloudformation/panther-iam-roles-rule-engine-account-customer.yml
horakj: I added to the top of the parameters now so it is easier to find

austinbyers:
@horakj take a look, I made some changes:

* Turned the secret name into an arn; it's possible it could be in yet another account
* Added the option for KMS decryption
* Added CloudFormation conditions for KMS and Secret access - the policy is added only if their respective parameters were defined
* Removed the Panther managed policy. With #2647, we can pass an IAM role as a top-level Panther parameter (`PythonAssumableRoleArns`) and the Python engines will be allowed to assume it

Review the changes, test them in your account, and then you're good to merge once you're happy with them!

austinbyers (approved):

horakj:
Tested in dev environment with no issues

PR #2563, opened 2021-01-29
Aux CFN - Athena Third Party BI Connection IAM Permissions 

horakj:
## Background

 Generic template the covers multiple BI tools using Athena that require an access key and secret key.

## Changes

- 1 CFN Aux Template

## Testing

- Works with Mode BI and Apache Superset


deployments/auxiliary/cloudformation/panther-bi-iam-permissions.yml
```diff
@@ -0,0 +1,112 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+AWSTemplateFormatVersion: 2010-09-09
+Description: IAM permissions needed to connect to Athena through third party tools.
+
+Resources:
+  Athena:
+    Type: AWS::IAM::User
+    Properties:
+      UserName: Athena_analytics_readonly
```
austinbyers, line 24: Let's make this a configurable parameter (but you can keep this as the default value).

deployments/auxiliary/cloudformation/panther-bi-iam-permissions.yml
```diff
@@ -0,0 +1,112 @@
...
+          - Effect: Allow
+            Action:
+              - s3:AbortMultipartUpload
+              - s3:GetBucketLocation
+              - s3:GetObject
+              - s3:ListBucket
+              - s3:ListBucketMultipartUploads
+              - s3:ListMultipartUploadParts
+              - s3:PutObject
+            Resource: arn:aws:s3:::panther-bootstrap-athenaresults-*
```
austinbyers, line 71: The bucket names are not so predictable. With the root stack deployment, for example: `myRootStackName-bootstrap-n86t3jcxl5be-athenaresults-1axkels3jum5m`

So our options are:
* Require the entire bucket name as a parameter, or
* Add a parameter for the root stack name (default: "panther") and then allow `${RootStack}-*-athenaresults*`

deployments/auxiliary/cloudformation/panther-bi-iam-permissions.yml
```diff
@@ -0,0 +1,112 @@
...
+          - Effect: Allow
+            Action:
+              - s3:AbortMultipartUpload
+              - s3:GetBucketLocation
+              - s3:GetObject
+              - s3:ListBucket
+              - s3:ListBucketMultipartUploads
+              - s3:ListMultipartUploadParts
+              - s3:PutObject
+            Resource: arn:aws:s3:::panther-bootstrap-historicaldata-*
```
austinbyers, line 91: Since you have the same set of permissions for each bucket, you can combine these 3 statements into one statement with multiple resources:

```suggestion
          - Effect: Allow
            Action:
              - s3:AbortMultipartUpload
              - s3:GetBucketLocation
              - s3:GetObject
              - s3:ListBucket
              - s3:ListBucketMultipartUploads
              - s3:ListMultipartUploadParts
              - s3:PutObject
            Resource:
              - arn:aws:s3:::panther-bootstrap-processeddata-*
              - arn:aws:s3:::panther-bootstrap-historicaldata-*
              - arn:aws:s3:::panther-bootstrap-athenaresults-*
```

(with the caveat about bucket names I mentioned above)

deployments/auxiliary/cloudformation/panther-bi-iam-permissions.yml
```diff
@@ -0,0 +1,112 @@
...
+    Type: AWS::IAM::AccessKey
+    Properties:
+      Status: Active
+      UserName: !Ref Athena
+
+Outputs:
+  AccessKey:
+    Value: !Ref AthenaAccessKey
+  SecretKey:
+    Value: !GetAtt AthenaAccessKey.SecretAccessKey
```
austinbyers, line 112: CloudFormation outputs are not encrypted and are visible to anyone with read permissions on the stack - we definitely should not output it here.

> CloudFormation does not redact or obfuscate any information you include in the Outputs section. We strongly recommend you do not use this section to output sensitive information, such as passwords or secrets.

Users may just have to generate the access key themselves after we create the IAM user

austinbyers (changes requested):
Good start!

Doesn't have to be now, but we do maintain terraform copies of all the aux templates as well, if it's easy to add a tf version that'd be awesome

deployments/auxiliary/cloudformation/panther-bi-iam-permissions.yml
nhakmiller: I agree. We do this internally because we the the only people/systems that have access to the account at all already have more access than the keys provide. In general, we shouldn't assume that other environments are setup the same as ours.

deployments/auxiliary/cloudformation/panther-bi-iam-permissions.yml
horakj: Sounds good.

deployments/auxiliary/cloudformation/panther-bi-iam-permissions.yml
```diff
@@ -0,0 +1,98 @@
...
+              - s3:GetBucketLocation
+              - s3:GetObject
+              - s3:ListBucket
+              - s3:ListBucketMultipartUploads
+              - s3:ListMultipartUploadParts
+              - s3:PutObject
+            Resource:
+              - !Sub arn:aws:s3:::${RootStackName}-*-athenaresults-*
+              - !Sub arn:aws:s3:::${RootStackName}-*-processeddata-*
+              - !Sub arn:aws:s3:::${RootStackName}-*-historicaldata-*
```
austinbyers, line 84: Replace `:aws:` with `:${AWS::Partition}:` to ensure this can work in AWS China and gov-cloud in the future

deployments/auxiliary/cloudformation/panther-bi-iam-permissions.yml
```diff
@@ -0,0 +1,98 @@
...
+          - Effect: Allow
+            Action:
+              - s3:GetBucketLocation
+              - s3:ListAllMyBuckets
+              - s3:ListBucket
```
austinbyers, line 89: GetBucketLocation and ListBucket are already included above, why do we have to grant access to every S3 bucket here?

I think the only one you need to keep in this block is `s3:ListAllMyBuckets`

deployments/auxiliary/cloudformation/panther-bi-iam-permissions.yml
```diff
@@ -0,0 +1,95 @@
...
+
+AWSTemplateFormatVersion: 2010-09-09
+Description: Athena Third Party BI Tool Permissions (Create access key post deployment for IAM user created).
+
+Parameters:
+  UserName:
+    Type: String
+    Default: PantherAthenaBI
+    Description: Enter a name for the IAM user that will be created.
+  RootStackName:
```
austinbyers, line 25: This may not work for customers who deployed directly from source, but that will only apply to the community edition and those deployments will be migrated in a few weeks anyway

austinbyers (approved):

deployments/auxiliary/cloudformation/panther-bi-iam-permissions.yml
horakj: My dev is deployed from source and I have no issue but I am using the enterprise repo.

austinbyers: `mage deploy` direct-stack buckets will look like this:

* `panther-bootstrap-historicaldata-1dagafs5yass5`

There is no "root stack" name in this case, but I see why your panther default is working. We do actually have some production accounts with this setup, and their bucket names will not change even with the migration.

`mage master:deploy` buckets will look like this:

* `myRootStackName-bootstrap-n86t3jcxl5be-historicaldata-1axkels3jum5m`

----

So yes, this pattern

* `!Sub arn:${AWS::Partition}:s3:::${RootStackName}-*-historicaldata-*`

will work for both cases (if you want to be extra certain, you can verify with a `mage master:deploy` setup)

austinbyers (approved):

PR #2568, opened 2021-01-31
Minor tool fixes

rleighton:
## Background

Minor tools fixes found by using them for testing.

## Changes

- Ignore rate when not set to adjust concurrency in `s3queue`
- fix bugs introduced in final refactor for s3access log generation in `filegen`

## Testing

- mage build:tools
- generate files then use s3queue to drive load into Panther


cmd/devtools/filegen/fields.go
```diff
@@ -40,6 +40,12 @@ func String(n int) string {
 	return uniuri.NewLen(n)
 }
 
+var alphaNumChars = []byte(`abcdedfhijklmnopqrstuvwxyz0123456789`)
+
+func StringAlphaNum(n int) string {
```
kostaspap, line 45: The `String()` function above already returns alphaNum - just it includes also uppercase characters. If that's the case, maybe it would enough just to invoke `strings.ToLower(filegen.String(...))` in the part of the code that need it? 
If you think another method is needed, probably it should be called differently e.g. `AlphaNumLowerCase` or sth

kostaspap (approved):
Changes look good

cmd/devtools/filegen/fields.go
rleighton: I wanted to make it clear that `String()` could be any utf8 string in the signature ... we might change libraries.  That fact that the current implementation is somewhat restrictive could change.

PR #2570, opened 2021-02-01
Refactor date range picker

kouknick:
## Description

From now on the Datepicker accepts props to handle the dates in utc or local timezones.
This PR removes the custom logic to perform the utc to local time conversion.

### Checklist:


vorillaz (approved):
🙏

PR #2573, opened 2021-02-01
Fix log latency metric 1/2

kostaspap:
## Background

* While working on  https://github.com/panther-labs/panther-enterprise/issues/1753 I identified a bug: If in the same Lambda invocation we had multiple metrics with different dimension values, only one dimension value would be output to output. The issue would cause one of the tests added in this PR(`multiple dimension values`)  to fail. 

* I have also update the `metrics.Counter` to support units

## Changes

- Changed metrics.Counter to support different units
- Fixed bug mentioned above

## Testing

- unit tests
- Deployed to account and used to emit some metrics from Log processing. 

Screen Shot 2021-02-02 at 4 24 34 PM
Screen Shot 2021-02-02 at 4 32 13 PM

### DIFF
--- a/pkg/metrics/cloudwatch.go
+++ b/pkg/metrics/cloudwatch.go
@@ -33,7 +33,7 @@ type Manager interface {
 	// It will sync the metrics every `interval` duration
 	Run(ctx context.Context, interval time.Duration)
 	// Returns a new Counter
-	NewCounter(name string) Counter
+	NewCounter(name, unit string) Counter
 	// Sync the metrics to the underlying system
 	Sync() error
 }
@@ -48,9 +48,7 @@ type CWEmbeddedMetricsManager struct {
 	stream *jsoniter.Stream
 	// Function that return the current time in milliseconds
 	// Can be overwritten in unit tests
-	timeFunc      func() int64
-	metricsBuf    []Metric
-	dimensionsBuf [][]string
+	timeFunc func() int64
 }
 
 // New returns a CWEmbeddedMetricsManager object that may be used to create metrics.
@@ -70,13 +68,13 @@ func NewCWEmbeddedMetrics(writer io.Writer) *CWEmbeddedMetricsManager {
 }
 
 // NewCounter returns a counter. Observations are aggregated and emitted once
-// per write invocation.
-// Panics if the client has been closed
-func (c *CWEmbeddedMetricsManager) NewCounter(name string) Counter {
+// per Sync invocation.
+func (c *CWEmbeddedMetricsManager) NewCounter(name, unit string) Counter {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 	return &DimensionsCounter{
 		name: name,
+		unit: unit,
 		obs:  c.counters.Observe,
 	}
 }
@@ -119,13 +117,11 @@ func (c *CWEmbeddedMetricsManager) sync() ([]byte, error) {
 	c.stream.Reset(nil)
 	c.stream.Error = nil
 
-	// Reusing slices to avoid extra allocations
-	c.metricsBuf = c.metricsBuf[:0]
-	c.dimensionsBuf = c.dimensionsBuf[:0]
+	timeNow := c.timeFunc()
 
-	c.stream.WriteObjectStart()
+	c.counters.Reset().Walk(func(name, unit string, dms DimensionValues, values []float64) bool {
+		c.stream.WriteObjectStart()
 
-	c.counters.Reset().Walk(func(name string, dms DimensionValues, values []float64) bool {
 		// Write `"<metric name>" : <value>`
 		c.stream.WriteObjectField(name)
 		c.stream.WriteVal(sum(values))
@@ -141,35 +137,30 @@ func (c *CWEmbeddedMetricsManager) sync() ([]byte, error) {
 			c.stream.WriteMore()
 		}
 
-		c.metricsBuf = append(c.metricsBuf, Metric{Name: name, Unit: "Count"})
-		c.dimensionsBuf = append(c.dimensionsBuf, dimensionNames(dms...))
+		embeddedMetric := EmbeddedMetric{
+			Timestamp: timeNow,
+			CloudWatchMetrics: []MetricDirectiveObject{
+				{
+					Namespace:  Namespace,
+					Dimensions: []DimensionSet{dimensionNames(dms...)},
+					Metrics:    []Metric{{Name: name, Unit: unit}},
+				},
+			},
+		}
 
+		const rootElement = "_aws"
+		c.stream.WriteObjectField(rootElement)
+		c.stream.WriteVal(embeddedMetric)
+		c.stream.WriteObjectEnd()
+		c.stream.WriteRaw("\n")
 		return true
 	})
 
 	// If there are no metrics to be reported
 	// Don't log anything
-	if len(c.metricsBuf) == 0 {
+	if len(c.stream.Buffer()) == 0 {
 		return nil, nil
 	}
-
-	const namespace = "Panther"
-	embeddedMetric := EmbeddedMetric{
-		Timestamp: c.timeFunc(),
-		CloudWatchMetrics: []MetricDirectiveObject{
-			{
-				Namespace:  namespace,
-				Dimensions: c.dimensionsBuf,
-				Metrics:    c.metricsBuf,
-			},
-		},
-	}
-
-	const rootElement = "_aws"
-	c.stream.WriteObjectField(rootElement)
-	c.stream.WriteVal(embeddedMetric)
-	c.stream.WriteObjectEnd()
-	c.stream.WriteRaw("\n")
 
 	return c.stream.Buffer(), c.stream.Error
 }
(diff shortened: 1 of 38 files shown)


internal/log_analysis/log_processor/processor/processor.go
```diff
@@ -258,17 +259,9 @@ func (p *Processor) processLogLine(ctx context.Context, line string, outputChan
...
-	for _, parserStats := range p.classifier.ParserStats() {
-		p.operation.Log(err, zap.Any(statsKey, *parserStats))
-		logType.Value = parserStats.LogType
-		pMetrics[0].Value, pMetrics[1].Value, pMetrics[2].Value =
-			parserStats.BytesProcessedCount, parserStats.EventCount, parserStats.CombinedLatency
-		common.BytesProcessedLogger.Log(pMetrics, logType)
+	for _, stats := range p.classifier.ParserStats() {
+		p.operation.Log(err, zap.Any(statsKey, *stats))
+		logmetrics.BytesProcessed.With(metrics.LogTypeDimension, stats.LogType).Add(float64(stats.BytesProcessedCount))
+		logmetrics.EventsProcessed.With(metrics.LogTypeDimension, stats.LogType).Add(float64(stats.EventCount))
```
kostaspap, line 265: Will move this ones to `processLogLine` function in a separate PR.

pkg/metrics/counter.go
```diff
@@ -151,40 +148,12 @@ func (n *node) observe(dvs DimensionValues, value float64) {
...
 	child, ok := n.children[head]
 	if !ok {
-		child = &node{}
+		child = &node{unit: n.unit}
 		n.children[head] = child
 	}
 	child.observe(tail, value)
 }
 
-func (n *node) add(dvs DimensionValues, delta float64) {
```
kostaspap, line 160: This function was not used by our code - removed.

pkg/metrics/dimensions.go
```diff
@@ -19,19 +19,11 @@ package metrics
...
 	SubsystemDimension = "Subsystem"
 
-	// StatusDimension dimension
 	StatusDimension = "Status"
-	// StatusDimension indicating that a subsystem operation is well
-	StatusOk = "Ok"
-	// StatusDimension indicating that a subsystem is experiencing authZ/N errors
-	StatusAuthErr = "AuthErr"
-	// StatusDimension indicating some general error with the subsystem
-	StatusErr = "Err"
```
kostaspap, line 33: Moved dimension values to individual Lambda package.

pkg/metrics/cloudwatch.go
```diff
@@ -141,36 +137,31 @@ func (c *CWEmbeddedMetricsManager) sync() ([]byte, error) {
 			c.stream.WriteMore()
 		}
 
-		c.metricsBuf = append(c.metricsBuf, Metric{Name: name, Unit: "Count"})
-		c.dimensionsBuf = append(c.dimensionsBuf, dimensionNames(dms...))
+		embeddedMetric := EmbeddedMetric{
```
kostaspap, line 140: Initially I thought of writing directly to `c.stream`. However the code was becoming quickly too complicated for people to read so opted to go for this approach instead.

giorgosp (approved):

georgepsarakis (commented):
Nice work (as far as I understand it 😅 ) ! So the majority of changes concerns passing the unit through to the underlying metrics aggregator framework, right?

Also, quick question, was the test case `"clears buffer after syncing"` reproducing some bug that you mentioned (at least that's what I remember)?

kostaspap:
> Nice work (as far as I understand it 😅 ) ! So the majority of changes concerns passing the unit through to the underlying metrics aggregator framework, right?
> 
> Also, quick question, was the test case `"clears buffer after syncing"` reproducing some bug that you mentioned (at least that's what I remember)?

This PR addresses two issues: Adding units to the counter and also fixing a bug. I describe the bug in the PR description, the test that reproduced the bug is `multiple dimension values`.

georgepsarakis:
> I describe the bug in the PR description, the test that reproduced the bug is multiple dimension values.

You are right, sorry I missed that 😞 .

pkg/metrics/cloudwatch.go
```diff
@@ -70,13 +68,13 @@ func NewCWEmbeddedMetrics(writer io.Writer) *CWEmbeddedMetricsManager {
 }
 
 // NewCounter returns a counter. Observations are aggregated and emitted once
-// per write invocation.
-// Panics if the client has been closed
-func (c *CWEmbeddedMetricsManager) NewCounter(name string) Counter {
+// per Sync invocation.
+func (c *CWEmbeddedMetricsManager) NewCounter(name, unit string) Counter {
 	c.mtx.Lock()
```
alxarch, line 73: I don't see why we need to lock the manager here

alxarch: Normally I would suggest to hold the lock and have `counters` create the node for the requested counter so that future calls to `NewCounter` cannot change the units. But that's not how this library works.

alxarch (approved):
Seems fine.
A bit concerned about how units are propagated in `Space` and whether there's room to end up with a single metric having two different units.

pkg/metrics/counter.go
alxarch: How come we're not using this? Have we modified the `add` logic for go-kit's `Space`?

pkg/metrics/cloudwatch.go
```diff
@@ -141,36 +137,31 @@ func (c *CWEmbeddedMetricsManager) sync() ([]byte, error) {
...
+		c.stream.WriteVal(embeddedMetric)
+		c.stream.WriteObjectEnd()
+		c.stream.WriteRaw("\n")
 		return true
 	})
 
 	// If there are no metrics to be reported
 	// Don't log anything
-	if len(c.metricsBuf) == 0 {
+	if len(c.stream.Buffer()) == 0 {
```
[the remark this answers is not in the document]
kostaspap: Nope we shouldn't. Omitting 0 is different from not omitting anything at all. In CW world it is possible to differentiate handling of these cases to catch issues like metrics agent not sending data to the metrics endpoint.

pkg/metrics/counter.go
```diff
@@ -29,22 +29,24 @@ type Counter interface {
...
+		unit: d.unit,
 		dvs:  d.dvs.With(dvs...),
 		obs:  d.obs,
 	}
 }
 
 // Add implements metrics.Counter.
 func (d *DimensionsCounter) Add(delta float64) {
-	d.obs(d.name, d.dvs, delta)
+	d.obs(d.name, d.unit, d.dvs, delta)
```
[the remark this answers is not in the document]
kostaspap: I am not sure I would like to tie the metric name to the unit. In monitoring systems I've worked in the past it was possible to do conversion between units e.g. a metric `duration` could be switched to be displayed in hours/minutes/seconds/microseconds/nanoseconds. It was actually an anti-pattern to name the metric with a unit.

pkg/metrics/cloudwatch.go
kostaspap: Good catch! Removed the lock. the underlying `Space` already uses appropriate locking.

pkg/metrics/counter.go
kostaspap: It seems it is used only by `Gauge` type of metrics: https://github.com/go-kit/kit/blob/0ca22796c7622119f521a79bfd73a8b4d33a9ca4/metrics/cloudwatch/cloudwatch.go#L334-L336

kostaspap:
> Seems fine.
> A bit concerned about how units are propagated in `Space` and whether there's room to end up with a single metric having two different units.

Yeah, that's actually a possibility. An idea would be to fail(panic) if this happens. What do you think?

pkg/metrics/counter.go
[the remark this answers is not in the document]
giorgosp: I too think the metric should be in one unit.  Will there be 2 call sites where the one writes a metric in unitA and the other one in unitB?

`displayed` is the keyword here. It would be simpler if the all events of a metric would be in the same unit. The presentation layer can transform it, no?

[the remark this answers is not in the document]
kostaspap: > `displayed` is the keyword here. It would be simpler if the all events of a metric would be in the same unit. The presentation layer can transform it, no?

It is possible, but the idea here is that it ends up being painful to have to rename stuff in dashboards, alarms etc. It also becomes more challenging to compare metrics like duration/bytes processed/etc between two different systems because one decided to emit processing time in seconds and the other in milliseconds. So it is easier to get rid of metric names all together and let the metric system do its magic and display everything in seconds or milliseconds (because it has the ability to do so automatically)

I think prometheus suggests what it suggests because of the way it works in the backend and I am clearly biased because of other metrics systems I've used in the past that had the capability to automatically convert metrics to a standard unit (which was the system that CloudWatch is trying to become)

In any case, I made the suggestion that the code could fail if it tries to register a metric with the same name but with different unit. This unfortunately has to be done at runtime, but any other check of metrics (e.g. inferring the unit from the name of the metric) would also have to be done at runtime.

pkg/metrics/cloudwatch.go
kostaspap: I've also added a test that runs things concurrently, so to either `go test race` or the test itself could identify issues there.

pkg/metrics/cloudwatch.go
kostaspap: It was particularly annoying because there are so many arrays (even multi dimensional arrays) in there.

kostaspap:
This is PR 1/2 for fixing the metrics and since it has been approved already from 2 people, I will go ahead and merge it. The discussion however can continue and I can fix any issues in PR 2/2

PR #2574, opened 2021-02-01
Add GetObject to the health checks for an s3 source

### LINKED ISSUE #2586: Adding GetObject healthcheck for s3 sources
### Description

We want to add an extra health check for s3 sources which answers the question: 
Can Panther read an object on the user's bucket? 
Equivalently, has the user set up permissions properly so that the log processing role can read objects on the user's bucket?

However, this is not as simple as originally thought. There is an open [PR](#2574), but some issues came out during testing.
Because there are multiple permission mechanisms applying on AWS S3 (IAM policies, bucket policies, object ACLs), implementing this check is not as simple as "do a `GetObject` request and expect either an `AccessDenied` or a `Success` response". 
For the same reason, checking for only the role's permissions does not mean that Panther can read the log files on the user's bucket. While Object ACLs may not be commonly used for log files, the user may have set up bucket policies, which this check doesn't account for.

Some potential implementations are described below.

#### Impl 1: Based on `ListObjects` + `GetObject` request
**tl;dr:** Perform a `ListObjects` request and then perform a `GetObject` request for any of the keys in the bucket. If the response is not a success response, fail the health check.
This needs an extra permission to the log processing role (`ListBucket`). 

Why not just make a `GetObject` request for a random non-existent key (eg `panther-health-check`)?
AWS S3 always returns `AccessDenied` for a missing key if the user doesn't have the `s3.ListBucket` permission too.
But adding this permission doesn't solve our issue.

Below is a table that shows the responses for an **`s3.GetObject` request** for a missing and an existing key, given the permissions of the user making the request and the absence of any bucket policies or object ACLs.

Note: `ListBucket` is the name of the permission needed to perform the `ListObjects` operation.

| Perms  | Missing Key | Existing Key
| ------------- | ------------- | ------------- |
| GetObject  | AccessDenied | Success  |
| ListBucket + GetObject  | NoSuchKey  | Success |
| ListBucket  | NoSuchKey  | AccessDenied  |

The table shows that if the request is made for a missing key and the user has the `ListBucket` permission, there is no way to decide whether they also have the `GetObject` permission (because the response for rows 2 and 3 is the same).
However, making a request for an existing key has a success response only if the user has both the ListBucket, GetObject permissions.

#### Impl 2: Explicitly request the permissions (policies) of the log processing role
This should be a more reliable solution, as the log processing role should be able to read all of its policies and permissions.
Panther can then explicitly search inside the policies for the `GetObject` permission it needs.

For this to work, the log processing role should have at least these extra permissions: 
- `iam:ListRolePolicies`
- `iam:GetRolePolicy`
- `iam:ListAttachedRolePolicies`

This solution will require more development time. 
Because we can only request details for 1 policy at a time from the AWS API, if the role has many policies attached, it may slow down the health check which will lead to slowing down the list-sources page.

giorgosp:
## Background

This check was missing and is required. Even if the other checks (AssumeRole, GetBucketLocation, DescribeKmsKey) pass, `GetObject` could still fail if the permissions are not correctly set up.

## Changes

- Add a check for `s3.GetObject`

## Testing
- unit
- manual e2e with cross-account s3 source

Closes #2586

### DIFF
--- a/deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
+++ b/deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
@@ -113,13 +113,9 @@ Resources:
             Version: 2012-10-17
             Statement:
               - Effect: Allow
-                Action: s3:GetBucketLocation
-                Resource: !Sub
-                  - arn:${AWS::Partition}:s3:::${Bucket}
-                  # prettier-ignore
-                  - Bucket: !If [IsGenerated, !FindInMap [PantherParameters, S3Bucket, Value], !Ref S3Bucket]
-              - Effect: Allow
-                Action: s3:ListBucket
+                Action:
+                  - s3:GetBucketLocation
+                  - s3:ListBucket
                 Resource: !Sub
                   - arn:${AWS::Partition}:s3:::${Bucket}
                   # prettier-ignore
--- a/deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
+++ b/deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
@@ -117,11 +117,17 @@ Resources:
                 Resource: !Sub
                   - arn:${AWS::Partition}:s3:::${Bucket}
                   # prettier-ignore
-                  - Bucket: !If [ IsGenerated, !FindInMap [ PantherParameters, S3Bucket, Value ], !Ref S3Bucket ]
+                  - Bucket: !If [IsGenerated, !FindInMap [PantherParameters, S3Bucket, Value], !Ref S3Bucket]
+              - Effect: Allow
+                Action: s3:ListBucket
+                Resource: !Sub
+                  - arn:${AWS::Partition}:s3:::${Bucket}
+                  # prettier-ignore
+                  - Bucket: !If [IsGenerated, !FindInMap [PantherParameters, S3Bucket, Value], !Ref S3Bucket]
               - Effect: Allow
                 Action: s3:GetObject
                 Resource: !Sub
-                  - 'arn:${AWS::Partition}:s3:::${Bucket}/*'
+                  - arn:${AWS::Partition}:s3:::${Bucket}/*
                   # prettier-ignore
                   - Bucket: !If [ IsGenerated, !FindInMap [ PantherParameters, S3Bucket, Value ], !Ref S3Bucket ]
               - !If
@@ -171,6 +177,6 @@ Resources:
           - Effect: Allow
             Action: [s3:GetBucketNotification, s3:PutBucketNotification]
             Resource: !Sub
-              - 'arn:aws:s3:::${Bucket}'
+              - 'arn:${AWS::Partition}:s3:::${Bucket}'
               # prettier-ignore
               - Bucket: !If [ IsGenerated, !FindInMap [ PantherParameters, S3Bucket, Value ], !Ref S3Bucket ]
--- a/internal/core/source_api/api/check_integration.go
+++ b/internal/core/source_api/api/check_integration.go
@@ -19,12 +19,11 @@ package api
  */
 
 import (
-	"errors"
 	"fmt"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
@@ -33,6 +32,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 	"github.com/aws/aws-sdk-go/service/sqs"
 	"github.com/aws/aws-sdk-go/service/sts"
+	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
 	"github.com/panther-labs/panther/api/lambda/source/models"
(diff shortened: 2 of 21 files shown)


internal/core/source_api/api/check_integration.go
```diff
@@ -222,6 +255,11 @@ func (api *API) evaluateIntegration(integration *models.CheckIntegrationInput) (
 		if !status.KMSKeyStatus.Healthy {
 			return status.KMSKeyStatus.Message, false, nil
 		}
+
+		if !status.S3GetObject.Healthy {
+			return status.KMSKeyStatus.Message, false, nil
```
kostaspap, line 260: This doesn't seem right.

internal/core/source_api/api/check_integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+type mockS3Client struct {
+	mock.Mock
+	s3iface.S3API
+}
+
+func (m *mockS3Client) GetObject(input *s3.GetObjectInput) (*s3.GetObjectOutput, error) {
+	args := m.Called(input)
+	return args.Get(0).(*s3.GetObjectOutput), args.Error(1)
+}
```
kostaspap, line 42: `testutils` have an S3 mock client defined, you should use that one.

internal/core/source_api/api/check_integration.go
```diff
@@ -90,12 +93,42 @@ func (api *API) checkAwsS3Integration(input *models.CheckIntegrationInput) *mode
...
+			Region:      bucketRegion,
+		})
+		out.S3GetObject = checkGetObject(s3Client, input.S3Bucket, input.AWSAccountID)
 	}
 	return out
 }
 
+func checkGetObject(s3Client s3iface.S3API, bucket, owner string) models.SourceIntegrationItemStatus {
+	// Check the error for a non-existent key. If the error is s3.ErrCodeNoSuchKey, we are good.
+	randomKey := "panther-check-" + uuid.New().String()
```
kostaspap, line 110: This can just be a constant e.g. `panther-health-check`. It would make unit tests much easier to write as well.

internal/core/source_api/api/check_integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+type mockAWSError string
+
+func (e mockAWSError) Error() string   { return string(e) }
+func (e mockAWSError) Code() string    { return string(e) }
+func (e mockAWSError) Message() string { panic("implement me") }
+func (e mockAWSError) OrigErr() error  { panic("implement me") }
+
+var _ awserr.Error = (*mockAWSError)(nil)
```
kostaspap, line 51: Do we need a mock for that? In case of error, it seems we could just use the original AWS structs?

internal/core/source_api/api/check_integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+func Test_CheckGetObject(t *testing.T) {
+	argMatcher := func(i *s3.GetObjectInput) bool {
+		return *i.Bucket == "bucket-name" &&
+			*i.ExpectedBucketOwner == "bucket-owner" &&
+			strings.HasPrefix(*i.Key, "panther-check")
+	}
```
kostaspap, line 58: That's really nice, I was not aware of the arg matchers! They could be used to simplify quite few of our unit tests.

kostaspap (commented):
Nice one! Could you add more details about the steps you did as part of your manual tests? Could you add also responses from the API in those checks?

giorgosp:
> Nice one! Could you add more details about the steps you did as part of your manual tests? Could you add also responses from the API in those checks?

Updated the description. Also doing E2E now, an issue you mentioned in the comments will surface as well.

internal/core/source_api/api/check_integration.go
giorgosp: updated

internal/core/source_api/api/check_integration_test.go
giorgosp: I tried `s3.Error` but doesn't implement `error`. Perhaps if I dig more into the aws-sdk I can find something but I will skip it for now.

giorgosp:
Actually s3 returns NoSuchKey only if the user/role has `s3.ListBucket` permissions. So we could either add this permission to the log processing role or abandon this PR for now

internal/core/source_api/api/check_integration.go
```diff
@@ -90,12 +93,60 @@ func (api *API) checkAwsS3Integration(input *models.CheckIntegrationInput) *mode
...
+	// missing keys.
+	_, err := s3Client.ListObjects(&s3.ListObjectsInput{
+		Bucket:              &bucket,
+		ExpectedBucketOwner: &owner,
+		MaxKeys:             aws.Int64(1),
+	})
+	if err != nil {
+		msg := "Error returned from s3.ListBucket request."
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "AccessDenied" {
+			msg = "We couldn't perform the s3.GetObject check because the IAM role doesn't have the s3.ListBucket permission."
```
giorgosp, line 120: I thought it would be better to show this error rather than hiding the healthcheck. Hiding it can potentially hide problems for new sources too. 

This is how it looks like
Screenshot 2021-02-02 at 11 49 52

kostaspap: Ok, so this means that for all existing customers, the S3 log sources will appear `UNHEALTHY` until they deploy the new templates? There is no way to make that backwards compatible, so that the UI displays the source as `HEALTHY` if the user has deployed an old template? @mattthetraveler @jacknagz are we ok with this behavior?

kostaspap:
> ## Background
> This check was missing and is required. Even if the other checks (AssumeRole, GetBucketLocation, DescribeKmsKey) pass, `GetObject` could still fail if the permissions are not correctly set up.
> 
> ## Changes
> * Add a check for `s3.GetObject`
> 
> ## Testing
> * manual: calling the `s3.GetObject` operation (in order to confirm AWS expected responses). Eg:
>   
>   * Example using the bucket's owner as user
>   
>   ```
>   AWS_REGION=eu-west-1 aws-vault exec dev-giorgosp-admin -- aws s3api get-object --bucket <redacted> --key panther-health-check /dev/null
>   
>   An error occurred (NoSuchKey) when calling the GetObject operation: The specified key does not exist.
>   ```
>   
>   
>   
>   * Example with other user
>     ```
>     AWS_REGION=eu-west-1 aws-vault exec dev-kostas-admin -- aws s3api get-object --bucket <redacted> --key panther-health-check /dev/null
>     
>     An error occurred (AccessDenied) when calling the GetObject operation: Access Denied
>     ```
> * unit

Can you also test the feature E2E? 
1. Deploy Panther
2. Onboard a new source from S3. Check that the S3 source is healthy
3. Modify the deploy role. Verify that the source is unhealth

Also do the above steps for cross-account and cross-region onboardings.

internal/core/source_api/api/check_integration.go
giorgosp: > Ok, so this means that for all existing customers, the S3 log sources will appear UNHEALTHY until they deploy the new templates? 

Yes. 

>There is no way to make that backwards compatible, so that the UI displays the source as HEALTHY if the user has deployed an old template?

If the `s3.ListObject` permission is missing, we can't show it as HEALTHY because we cannot distinguish an old source from a bug or a user error.

Some other ideas would be:
- Only enable this check for sources created on Panther >= 1.16 (or >=1.15.3 perhaps)
- Engage the frontend team to show this error as a warning if the `s3.ListObject` permission is missing?
- Improve the error message. Perhaps point users to a new section in our docs so they can add `s3.ListObject` to the log processing role manually?

mattthetraveler: I like the idea of:

- Only enable this check for sources created on Panther >= 1.16 (or >=1.15.3 perhaps)

We can preempt customer success team (ie @horakj ) that customers upgrading to 1.16 or above will run into this error
- Improve the error message. Point users to a new section in our docs so they can add s3.ListObjects to the log processing role manually

We should provide as much context as possible around this. In addition to the "We count perform the s3.GetObject.." message, we should add the sentence "Learn more about this health check **here**". This should link them to an explanation of the health check in our s3 docs and how to deploy the log processing role.

giorgosp: If we enable this health check only for s3 sources created in Panther >= 1.16, then we don't need to explicitly inform on users - it will be part of the s3 onboarding flow.

We need, however, a way for users to enable this health check for pre-existing sources. For this one we will probably need a release note with a link to the docs on how to add the extra permission. Will take care of it.

giorgosp: >"Learn more about this health check here".

Btw having docs for every health check (and perhaps remediation steps) would be great but a much bigger effort that would extend the scope of this project a lot.

giorgosp: Updated the check to run only for sources created in Panther >= 1.16.

For pre-existing sources, unfortunately users have to:
- Add the `s3.ListBucket` permission to the existing log processing role of the source
- **Delete and recreate** the source.

Adding only the `s3.ListBucket` is not enough to enable the check. Because the absence of this permission may indicate **either**:
- source is created before Panther 1.16.x or
- source is created after Panther 1.16 but user didn't specify the correct permissions in the role

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
```diff
@@ -100,23 +100,21 @@ Resources:
...
                 Action: s3:GetBucketLocation
                 Resource: !Sub
-                  - arn:${AWS::Partition}:s3:::${Bucket}
-                  - Bucket:
-                      !If [
-                        IsGenerated,
-                        !FindInMap [PantherParameters, S3Bucket, Value],
-                        !Ref S3Bucket,
-                      ]
+                  - 'arn:aws:s3:::${Bucket}'
```
austinbyers, line 103: Nit: don't need the quotes

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
```diff
@@ -100,23 +100,21 @@ Resources:
...
+                  # prettier-ignore
+                  - Bucket: !If [IsGenerated, !FindInMap [PantherParameters, S3Bucket, Value], !Ref S3Bucket]
+              - Effect: Allow
+                Action: s3:ListBucket
+                Resource: !Sub
+                  - 'arn:aws:s3:::${Bucket}'
+                  # prettier-ignore
+                  - Bucket: !If [IsGenerated, !FindInMap [PantherParameters, S3Bucket, Value], !Ref S3Bucket]
```
austinbyers, line 111: Instead of a separate statement, this can be combined with the `s3:GetBucketLocation` permission above:

```yml
- Effect: Allow
  Action:
    - s3:GetBucketLocation
    - s3:ListBucket
  Resource: !Sub ....
```

internal/core/source_api/api/utils.go
```diff
@@ -123,3 +127,22 @@ func itemToIntegration(item *ddb.Integration) *models.SourceIntegration {
...
+				smallestPrefix = strs[j]
+			}
+		}
+		uniques[smallestPrefix] = struct{}{}
+	}
+	for k := range uniques {
+		reduced = append(reduced, k)
+	}
+	return
+}
```
austinbyers, line 148: Is it common to have multiple prefixes in a source that overlap? If so, what happens if we didn't do this de-duplication? I'm wondering if it's a bit of over-engineering

In any case, if we keep it, let's add unit tests for this function

austinbyers (approved):

kostaspap:
There seem to be a lot of conflicts with `master` @giorgosp can you take a look?

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
giorgosp: Good call, updated

internal/core/source_api/api/utils.go
giorgosp: There are some occasions. We also don't disallow it, so we have to account for it anyway.

The function was copied from another PR, it is unit-tested there. You can find it on master now.

PR #2577, opened 2021-02-01
Fix alert s3select paging bug

rleighton:
## Before
If you had only 1 file in an alert and it had more than 25 elements then getting the 2nd page would fail because we set `LIMIT` in the SQL to 25 but S3 select has no `OFFSET` so we skip rows in a loop. This means on the 2nd page we limit the result to 25 but then skip all 25 finding nothing.

## After

Made the `LIMIT` the specified limit plus of paging offset.

## Testing

- mage test:ci
- deployed and confirmed working in test case that prompted the fix
-


kostaspap (approved):
Wow, nice find! We should sync that to the release.

PR #2585, opened 2021-02-02
Wrong type for AIDMaster.PointerSize field

georgepsarakis:
## Background

In https://github.com/panther-labs/panther/pull/2475 we defined the pointer size as a numeric type, which in some cases is not true as it can contain the value `none`.

## Changes

- Sets the type of `PointerSize` back to `String`.
- Adds a missing (optional) field in AIDMaster parser. 
- Fixed a small bug which prevented the error stream to be flushed when using debug-level logging.

## Testing

- `mage test:go`
- `pantherlog` command-line tool
  * Verified debug logging and exit code.
  * Verified the parser changes work correctly with the sample event.


internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
```diff
@@ -69,14 +69,15 @@ type AIDMaster struct {
...
 	ProductType        pantherlog.String  `json:"ProductType" description:"The type of product (in decimal, non-hex format). Example values: '1' (Workstation), '2' (Domain Controller), '3' (Server)."`
 	ServicePackMajor   pantherlog.String  `json:"ServicePackMajor" description:"The major version # of the OS Service Pack (in decimal, non-hex format)."`
 	SiteName           pantherlog.String  `json:"SiteName" description:"The site name of the domain to which the host is joined (defined by system admin)."`
 	SystemManufacturer pantherlog.String  `json:"SystemManufacturer" description:"The host's system manufacturer."`
 	SystemProductName  pantherlog.String  `json:"SystemProductName" description:"The host's product name."`
 	Timezone           pantherlog.String  `json:"Timezone" description:"The sensor's time zone, as seen from the CrowdStrike cloud."`
 	Version            pantherlog.String  `json:"Version" description:"The host's system version."`
+	HostHiddenStatus   pantherlog.String  `json:"HostHiddenStatus" description:"Whether the host is visible or not."`
```
kostaspap, line 80: It was really hidden :P

kostaspap (approved):
Parser changes LGTM, approving that bit. `pantherlog` bit looks more like Alex's field.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/insight_raw_events.go
```diff
@@ -69,14 +69,15 @@ type AIDMaster struct {
 	FirstSeen          pantherlog.Time    `json:"FirstSeen" tcodec:"unix" description:"The first time the sensor was seen by the CrowdStrike cloud in epoch format."`
 	MachineDomain      pantherlog.String  `json:"MachineDomain" description:"The Windows domain name to which the host is currently joined."`
 	OU                 pantherlog.String  `json:"OU" description:"The organizational unit of the host as seen by the sensor (defined by system admin)."`
-	PointerSize        pantherlog.Int32   `json:"PointerSize" description:"The processor architecture (in decimal, non-hex format): '4' for 32-bit, '8' for 64-bit."`
+	PointerSize        pantherlog.String  `json:"PointerSize" description:"The processor architecture (in decimal, non-hex format): '4' for 32-bit, '8' for 64-bit, or 'none' for unknown."`
```
kostaspap, line 72: Tests will probably fail - which makes me think I might have jumped onto this PR a bit too soon ^^

georgepsarakis: Still fixing 😅 !

cmd/devtools/pantherlog/pantherlog/main.go
```diff
@@ -52,9 +53,18 @@ var (
 func main() {
 	flag.Parse()
 
+	success := true
+
+	defer func() {
```
georgepsarakis, line 58: @alxarch if you have no time to review, I can keep a stash of this commit and remove it from this PR, that's not a problem.

internal/log_analysis/log_processor/parsers/crowdstrikelogs/testdata/insight_raw_events.yml
```diff
@@ -45,7 +45,8 @@ input: |
     "SystemProductName": "12345",
     "Time": "1590364206.99",
     "Timezone": "America/Los Angeles",
-    "Version": "Windows 8.1"
+    "Version": "Windows 8.1",
+    "HostHiddenStatus": "VISIBLE"
```
kostaspap, line 49: I really love this field

cmd/devtools/pantherlog/pantherlog/main.go
alxarch: I think that just setting `strerr` to _unbuffered_ `os.Stderr` should work with minimal changes.
I don't like having a 'flag' boolean that needs to be set before each break/return.
If we want to clean this up, I would suggest extracting a function that runs and returns an `error` and have main check 
```
if err != nil {
   ...
}
```

alxarch (approved):
I think the `stderr` issue can be somewhat handled with unbuffered writer.
I think bufferring `stderr` output  is problematic in the first place.
The use of `os.Exit` probably aborts without unwrapping the stack, that's why we're missing error messages.
IMO the cleanest solution to this would be to have all logic in a `run() error` function and have main call `os.Exit(1)` if it returns non-nil error.
I am approving since I don't know if I'll get the time to review this again by EOW. 
Using the `success` flag forces all code to be inside `main()` forever. If we relied  on returned errors, they would bubble up more 'naturally' if we ever need to split the code into multiple functions.

georgepsarakis:
> the use of os.Exit probably aborts without unwrapping the stack, that's why we're missing error messages.

Correct, deferred functions do not run with [os.Exit](https://golang.org/pkg/os/#Exit). [`runtime.Goexit()`](https://golang.org/pkg/runtime/#Goexit) is suggested for goroutines specifically:

> Calling Goexit from the main goroutine terminates that goroutine without func main returning. Since func main has not returned, the program continues execution of other goroutines. If all other goroutines exit, the program crashes. 

I gave this a try and I am not sure why it doesn't terminate though in our case.

I took your suggestion to use `os.Stderr` directly, and it seems to work fine. Thanks for the feedback.

PR #2588, opened 2021-02-02
fix issue with sortby in data models

lindsey-w:
## Background

Bug in sorting of data models, was not taking into account the input `sortBy`

## Changes

- add sortby to the sort call
- Sorting integration tests

## Testing

- `mage test:integration`
- manual testing


austinbyers (approved):
Nice!

PR #2589, opened 2021-02-02
Reduce deleted resource TTL

nhakmiller:
## Background

Now that resource snapshots are persisted to the data lake, maintaining a deleted resource snapshot in dynamo has significantly less value but still incurs a large cost in high churn environments.

## Changes

- Reduce TTL on deleted resources from 30 days to 2 days

## Testing

- None YOLO

### DIFF
--- a/internal/compliance/resources_api/handlers/delete_resources.go
+++ b/internal/compliance/resources_api/handlers/delete_resources.go
@@ -30,8 +30,8 @@ import (
 	"github.com/panther-labs/panther/api/lambda/resources/models"
 )
 
-// Deleted resources are retained for 30 days in the database
-const deleteWindowSecs = 30 * 24 * 60 * 60
+// Deleted resources are retained for 2 days in the database
+const deleteWindowSecs = 2 * 24 * 60 * 60
 
 // DeleteResources marks one or more resources as deleted.
 func (API) DeleteResources(input *models.DeleteResourcesInput) *events.APIGatewayProxyResponse {


austinbyers (approved):
🎉

nhakmiller:
It's important to note that this will in no way affect any current DynamoDB entries that already have 30 day TTLs on them, only the creation of new entries. If there are customers that are being particularly affected by slowness due to this issue we should direct them on how to delete these resources manually.

austinbyers:
> slowness due to this issue we should direct them on how to delete these resources manually.

@tatehanawalt is working on a script for that: https://github.com/panther-labs/panther-enterprise/issues/2099

kostaspap:
The TTL should be "last resort" kinda thing, right? We should be deleting the resources when receiving a "Delete" event so my guess was that this TTL would be not useful for the majority of cases?

nhakmiller:
> The TTL should be "last resort" kinda thing, right? We should be deleting the resources when receiving a "Delete" event so my guess was that this TTL would be not useful for the majority of cases? In case of an operational failure (e.g. missing lots of delete events) we could just treat it as an operational issue and run the script to fix it.
> 
> The initial idea of having a relatively high value (in this case 30 days) was that in case of some operational issue that affects the daily scan (but not the real time scans), the users shouldn't be blind. In other words.
> 
> So my understanding is that we have this dilemma:
> 
> * Operational issue leads to Panther missing delete events -> deleted resources will stay around for 30 days VS
> * Operation issue leads to Panther daily scans not running -> resources get deleted from Panther.
> 
> Does the above summarize it?

Ah no, sorry that's not what this TTL is for. This is the TTL for when we _do_ see a delete event. So we see a delete event, we mark the resource as `deleted=true`, and then we  _don't_ delete it from DynamoDB and instead set a TTL on it. Previously that was a 30 day TTL, now it's a 2 day TTL.

The TTL for if we miss a delete event is changed in this PR.

PR #2593, opened 2021-02-03
[FE]: Support Data Model filtering

### LINKED ISSUE #2488: [Enhance] [FE] Support Data Model filtering
### Description

In the v1.15 release we released an MVP of our Data Models in which a user can connect different fields from different log types, to a unified normalized field. This helps analysts since they can use a single/same field for all of their logs.

A user can see their data models by navigating to `/log-analysis/data-models/`, but they currently have no way to filter them. 

That's what we want to support in this PR.

### Related Services

- Web
  - Data Models List Page

### Designs

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/df9d59c8-eeb1-452d-a5f5-3712b64010be/commits/latest/files/46187896-210a-4c7f-92ff-d156f6dc68b2/layers/9267464E-2912-4F67-881C-DD8D6E1A092D?collectionId=65b935ef-579f-4620-9530-c359b8d416e3&collectionLayerId=59aa43d3-d78a-4572-bef5-9b27f14caf8d

### Related GraphQL API

https://github.com/panther-labs/panther-enterprise/blob/542ea50b0eea4134ce8017336f1b97ed81b1b3ad/api/graphql/schema.graphql#L196

### Acceptance Criteria

- Allow user to filter Data Models
  - Add a `TextInput` to allow users to filter by the `nameContains` property as seen in the designs
  - Add a button with the name "Filters (XXX)" that opens a Dropdown with additional filters
    - Dropdown must have a `MultiCombobox` to allow to filter by a list of Log Type values
    - Dropdown must have a `Combobox` to allow users to filter by `enabled`. The values should be `All`, `Enabled` &`Disabled`

- Allow user to sort the Data Models. This will be done through a Dropdown:
  - User should be able to sort  alphabetically through 2 options named `A to Z` and `Z to A` that internally to the `id` field 
  - User should be able to sort  by last modified date through 2 options named `Oldest` and `Most Recent` that internally to the `lastModified` field 

- All filters should be mirrored in the URL bar (via URL params)
- Sufficient tests should be added
 
### Notes

- Pagination shouldn't be handled for now. We can omit it or hardcode `pageSize` to a **high** value like `500` or `1000`
- To see examples of similar behavior, view the following components:
  - ListRulesPage
  - ListAlertsPage

kouknick:
## Description

This PR adds filtering and sorting options to the Data Models page.

Closes #2488 

### Screenshots:

https://user-images.githubusercontent.com/15084305/106728282-f920fd80-6614-11eb-9733-267c8577dc89.mov

*The sorting was broken when this video was captured but this [PR](https://github.com/panther-labs/panther/pull/2588) fixed the issue 

### Checklist:


web/src/pages/ListDataModels/ListDataModelFilters/ListDataModelFilters.tsx
```diff
@@ -0,0 +1,163 @@
...
+const defaultValues = {
+  nameContains: '',
+  sorting: null,
+};
+
+const sortingOpts: SortingOptions = [
+  {
+    opt: 'A to Z',
+    resolution: {
+      sortBy: ListDataModelsSortFieldsEnum.Id,
```
alexmylonas, line 50: Could you confirm that we should sort by `Id` and not by `displayName`

web/src/pages/ListDataModels/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,34 @@
...
+import React from 'react';
+import TablePlaceholder from 'Components/TablePlaceholder';
+import { Card, FadeIn } from 'pouncejs';
+
+const ListDataModelsPageSkeleton: React.FC = () => {
+  return (
+    <FadeIn from="bottom">
+      <TablePlaceholder rowCount={1} rowHeight={15} />
+      <Card p={6} mt={5}>
+        <TablePlaceholder />
```
alexmylonas, line 28:
```suggestion
        <TablePlaceholder rowCount={5} rowHeight={40} />
```
Or something like this to better represent `DataModel` cards

alexmylonas (approved):
Nice screencast and solid work on this PR! A couple of comments but nothing stopping me from approving! 🚀

web/src/pages/ListDataModels/ListDataModelFilters/ListDataModelFilters.tsx
kouknick: I followed the Acceptance Criteria:
> User should be able to sort alphabetically through 2 options named A to Z and Z to A that internally to the id field

@3nvi Can you confirm that this is the correct sorting field?

web/src/pages/ListDataModels/ListDataModelFilters/DropdownFilters.tsx
```diff
@@ -0,0 +1,133 @@
...
+                      <Field
+                        name="enabled"
+                        as={FormikCombobox}
+                        items={['', true, false]}
+                        itemToString={(item: boolean | string) => {
+                          if (item === '') {
+                            return 'All';
+                          }
+                          return item ? 'Yes' : 'No';
+                        }}
```
3nvi, line 106: can we:

- use `FastField` instead of `Field`
- Define `itemToString` as a separate function outside the React component, so we can have fixed reference

web/src/pages/ListDataModels/ListDataModelFilters/ListDataModelFilters.tsx
3nvi: yup, right now only `ID` is allowed. We can change the displayed options to `Id (A - Z)` and `Id (Z-A)` if we prefer. That might clear up any confusion.

3nvi (approved):
Really good work @kouknick . I have only 2 minor comments but they are trivial. 

Pre-approving since this shouldn't be blocked by my comments 🎉

PR #2594, opened 2021-02-03
[FE][BUGFIX] Remove stale appsync resolvers

3nvi:
## Background

When merging policies & rules page together, we removed the `listRules` and `listPolicies` GraphQL queries from the schema, since they were not longer needed. 

Unfortunately, we forgot to remove their related resolvers, causing AppSync to crash. This PR fixes just that.

## Changes

- Remove stale appsync resolveers

## Testing

- App can be deployed

## Notes

This wasn't caught since the GraphQL queries were removed during PR feedback in which case, I personally forgot to re-deploy (since I didn't expect any errors).  What went wrong was optimistically assuming that there would be no side-effects. Guess that's not gonna happen again!

### DIFF
--- a/deployments/appsync.yml
+++ b/deployments/appsync.yml
@@ -870,23 +870,6 @@ Resources:
         }
       ResponseMappingTemplate: !FindInMap [ResponseTemplates, Lambda, VTL]
 
-  ListPoliciesResolver:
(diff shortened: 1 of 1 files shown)


alexmylonas (approved):
LGTM!

PR #2597, opened 2021-02-03
Managed s3 bucket notifications

giorgosp:
## Background

Enable Panther to manage bucket notifications for the user. If user allows, Panther will create an SNS topic in the bucket's region and configure the bucket to send SNS notifications to that topic.

## Changes

- Add field `managedBucketNotifications` field in the source api so that user can select whether they want Panther to manage their bucket notifications configuration
- Add field `notificationsConfigurationSucceeded` so that the frontend knows whether this operation succeeded or not. Even if managing bucket notifications fail, Panther proceeds to save the source. The user just have to set them up manually.
- When an s3 source is created/updated and `managedBucketNotifications` is set, Panther sync's the updates the bucket's notification configuration.A notification configuration is created for every prefix the user has specified in the s3 source.
- When an s3 source is deleted, Panther deletes the Panther-managed notifications from the bucket. Any user-created configuration is left intact. Panther does not delete the SNS topic created because it may have been re-used. The UI will show the user that this resource has not been deleted.

## Testing

- Unit
- Manual e2e - All tests are done with a bucket from another account (Panther in @giorgosp account, bucket in @kostaspap account)
    - Create source with managed notifications with some prefixes (prefixa, prefixb) 
        - Verify bucket notifications config
    - Add some bucket notifications manually (not only SNS but also a Lambda notification). Update source prefixes. 
        - Verify bucket notifications config
    - Update source with a prefix already included in the bucket 
        - Verify error response (source is updated, but bucket notifications could not be configured. AWS doesn't allow overlapping prefixes)
    - Upload a file to the bucket - Verified event ended up in Athena
    - Delete source - Only Panther-managed configs should be removed from the bucket
    -  Create a second source for a second bucket in the same region as the first. Verify bucket notifications configured correctly and event made it to our data lake.


api/lambda/source/models/integration.go
```diff
@@ -50,6 +50,7 @@ type SourceIntegrationScanInformation struct {
 }
 
 // SourceIntegrationMetadata is general settings and metadata for an integration.
+//nolint:maligned
```
kostaspap, line 53: Why is this needed? We never had to escape this check before ^^

api/lambda/source/models/integration.go
```diff
@@ -72,12 +73,28 @@ type SourceIntegrationMetadata struct {
...
 	LogProcessingRole string           `json:"logProcessingRole,omitempty"`
+	// Whether Panther should configure the user's bucket notifications.
+	ManagedBucketNotifications bool `json:"managedBucketNotifications"`
+	// The resources that Panther created. This may be partial if an error occurred
+	// while creating a resource.
+	ManagedS3Resources ManagedS3Resources `json:"managedS3Resources,omitempty"`
+	// This is only needed for the API response, so that the UI can show a warning message
+	// if Panther couldn't setup bucket notifications. Failing to do so doesn't
+	// block any other source operations like saving to the DB.
+	NotificationsConfigurationSucceeded bool `json:"notificationsConfigurationSucceeded"`
```
kostaspap, line 84: It seems weird that we have to store this. Couldn't the code just check if the resources have been created? Why do we have to store this state in DDB?

api/lambda/source/models/integration.go
```diff
@@ -72,12 +73,28 @@ type SourceIntegrationMetadata struct {
...
 
 	StackName string `json:"stackName,omitempty"`
 
 	SqsConfig *SqsConfig `json:"sqsConfig,omitempty"`
 }
 
+type ManagedS3Resources struct {
+	TopicARN *string `json:"topicARN"`
```
kostaspap, line 92: nit: could just be `string`. We prefer to avoid pointers unless really needed since they require special management to avoid nil pointer dereference

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
```diff
@@ -51,6 +53,14 @@ Parameters:
     Description: DO NOT EDIT MANUALLY! Parameter is already populated with the appropriate value.
     Default: ''
 
+  ManagedBucketNotifications:
+    Type: String
+    Description: DO NOT EDIT MANUALLY! Parameter is already populated with the appropriate value.
+    Default: 'No'
```
kostaspap, line 59: nit: Usually we have true/false for boolean values.

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
```diff
@@ -101,22 +118,14 @@ Resources:
                 Action: s3:GetBucketLocation
                 Resource: !Sub
                   - arn:${AWS::Partition}:s3:::${Bucket}
-                  - Bucket:
-                      !If [
-                        IsGenerated,
-                        !FindInMap [PantherParameters, S3Bucket, Value],
-                        !Ref S3Bucket,
-                      ]
+                  # prettier-ignore
```
kostaspap, line 121: nit: The previous version actually seems easier to read to me  since each of the condition statements are separated by new lines.

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
```diff
@@ -135,3 +144,38 @@ Resources:
...
+      PolicyName: ManagePantherTopic
+      PolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Action:
+              - sns:CreateTopic
+              - sns:SetTopicAttributes
+              - sns:AddPermission
+              - sns:Subscribe
```
kostaspap, line 164: What about describe the topic?

internal/core/source_api/api/get_integration_template.go
```diff
@@ -103,6 +105,11 @@ func (api *API) GetIntegrationTemplate(input *models.GetIntegrationTemplateInput
 			formattedTemplate = strings.Replace(formattedTemplate, kmsKeyFind,
 				fmt.Sprintf(kmsKeyReplace, input.KmsKey), 1)
 		}
+
+		if input.ManagedBucketNotifications {
+			formattedTemplate = strings.Replace(formattedTemplate, managedNotificationsFind,
+				fmt.Sprintf(managedNotificationsReplace, "Yes"), 1)
+		}
```
kostaspap, line 112: Should we set it to `No`/`false` when it is not ?

internal/core/source_api/api/managed_bucket_notifications.go
```diff
@@ -0,0 +1,326 @@
...
+// The source.ManagedS3Resources field is mutated to contain the managed resources.
+// Even if this function returns an error, source.ManagedS3Resources is updated with the resources that
+// were created before the error occurred.
+func ManageBucketNotifications(pantherSess *session.Session, source *models.SourceIntegration) error {
+	managed := &source.ManagedS3Resources
+
+	stsSess, err := session.NewSession(&aws.Config{
+		MaxRetries:  aws.Int(5),
+		Credentials: stscreds.NewCredentials(pantherSess, source.RequiredLogProcessingRole()),
+	})
```
kostaspap, line 56: Here you create the session, however in the `RemoveBucketNotifications` the session is created outside of the function. I think we should be consistent in the way the code works.

internal/core/source_api/api/managed_bucket_notifications.go
```diff
@@ -0,0 +1,326 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to create sts session")
+	}
+
+	bucketRegion, err := getBucketLocation(stsSess, source.S3Bucket)
+	if err != nil {
+		return errors.Wrap(err, "failed to get bucket location")
+	}
+
+	if managed.TopicARN == nil {
```
kostaspap, line 66: Can you add a comment here why this check is required?

internal/core/source_api/api/managed_bucket_notifications.go
```diff
@@ -0,0 +1,326 @@
...
+	s3Client := s3.New(stsSess, &aws.Config{Region: bucketRegion})
+
+	bucket, prefixes := source.S3Info()
+	managedTopicConfigIDs, err := updateBucketTopicConfigurations(
+		s3Client, bucket, source.AWSAccountID, source.ManagedS3Resources.TopicConfigurationIDs, prefixes, managed.TopicARN)
+	if err != nil {
+		return errors.WithMessage(err, "failed to replace bucket configuration")
+	}
+	managed.TopicConfigurationIDs = managedTopicConfigIDs
+	zap.S().Debugf("replaced bucket topic configurations for %s", source.S3Bucket)
```
kostaspap, line 84: `zap.L().Debug("replaced bucket topic configurations", zap.String("bucket", source.S3Bucket)`. Much more performant. Also JSON output also makes debugging easier through CW Logs since CW Insights can parse the JSON log format. That's why we don't use SugaredLogger anywhere in Panther app.

internal/core/source_api/api/managed_bucket_notifications.go
```diff
@@ -0,0 +1,326 @@
...
+
+	pantherARN, err := getPantherDeploymentIAM(pantherSess)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get Panther deployment IAM info")
+	}
+	topicARN, err := createTopic(snsClient, pantherARN)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create topic")
+	}
+	zap.S().Debugf("created topic %s", *topicARN)
```
kostaspap, line 101: See above.

internal/core/source_api/api/managed_bucket_notifications.go
```diff
@@ -0,0 +1,326 @@
...
+
+	queueARN := arn.ARN{
+		Partition: pantherARN.Partition,
+		Service:   "sqs",
+		Region:    aws.StringValue(pantherSess.Config.Region),
```
kostaspap, line 106: Should be `bucketRegion`?

internal/core/source_api/api/managed_bucket_notifications.go
```diff
@@ -0,0 +1,326 @@
...
+				Principal: awsutils.Principal{
+					Service: "cloudtrail.amazonaws.com",
+				},
+				Resource: *topic.TopicArn,
+			}, {
+				Sid:    "AllowSubscriptionToPanther",
+				Effect: "Allow",
+				Action: "sns:Subscribe",
+				Principal: awsutils.Principal{
+					AWS: fmt.Sprintf("arn:aws:iam::%s:root", pantherARN.AccountID),
```
kostaspap, line 177: we shouldn't hardcode the partition `aws` - this will break for AWS China. You probably pick up the partition from the session?

internal/core/source_api/api/managed_bucket_notifications.go
```diff
@@ -0,0 +1,326 @@
...
+		prefixes,
+		source.ManagedS3Resources.TopicARN)
+	if err != nil {
+		return errors.Wrap(err, "failed to update bucket notifications configuration")
+	}
+
+	return nil
+}
+
+func createTopic(snsClient *sns.SNS, pantherARN arn.ARN) (*string, error) {
```
kostaspap, line 145: We just need the AccountID right? We shouldn't need to pass the full ARN - there should be little information of value in there, it's just the role assumed by the Lambda.

internal/core/source_api/api/managed_bucket_notifications.go
```diff
@@ -0,0 +1,326 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if bucketLoc.LocationConstraint == nil {
+		return aws.String(endpoints.UsEast1RegionID), nil
+	}
+	return bucketLoc.LocationConstraint, nil
+}
+
+func getPantherDeploymentIAM(sess *session.Session) (arn.ARN, error) {
```
kostaspap, line 316: Why is this needed? Is it just to get AccountID? This is passed as a property in the configuration.

internal/core/source_api/api/testdata/panther-log-analysis-iam-updated.yml
```diff
@@ -51,6 +53,14 @@ Parameters:
     Description: DO NOT EDIT MANUALLY! Parameter is already populated with the appropriate value.
     Default: ''
 
+  ManagedBucketNotifications:
+    Type: String
+    Description: DO NOT EDIT MANUALLY! Parameter is already populated with the appropriate value.
+    Default: 'No'
```
kostaspap, line 59: See similar comment in other template. Seems more natural to have this true/false.

kostaspap (commented):
Can you add more details about the E2E test? 
What the scenarios were tested as part of E2E, which failure scenarios were tested etc?

api/lambda/source/models/integration.go
alxarch: It's used before. It's a very strict rule that only makes sense in systems programming context. It checks that if you have two boolean/uint16 fields in your struct you should keep them in sequence so that the struct padding does not waste memory.
Field offsets are aligned in 32 bits in go. So even a `bool` or `uint16` take up  32 bits, unless you have them in sequence.
Really irrelevant to the kind of programming we do here.

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
alxarch: Actually `no` and `yes` are  valid YAML booleans. `'No'` is a string though.

api/lambda/source/models/integration.go
georgepsarakis: I also ran onto this once, read a few things [here](https://medium.com/@sebassegros/golang-dealing-with-maligned-structs-9b77bacf4b97).

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
kostaspap: Yup, I was mostly referring to how we do for the rest of the CFN templates internally.

api/lambda/source/models/integration.go
kostaspap: Do we really have to store this? We could just follow a convention and just name the topic sth e.g. `panther-sns-notifications-topic`. If the topic exists and we have access to it, we are good. I am not sure there is value in storing it here while we can describe it.

api/lambda/source/models/integration.go
giorgosp: We don't store it in the DB. `ddb.Integration` is the db representation. Unfortunately we don't have a separate source representation for the API response, hence the confusion here.

api/lambda/source/models/integration.go
giorgosp: Storing it saves us from an extra AWS API request + we don't delete it when the source is deleted, so we can display the full topic ARN to the UI, so the user can also see the topic's AWS partition, region, etc.
I will add this as a comment to the field, seems it will help.

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
giorgosp: CloudFormation doesn't not support boolean parameters. Also the user may have to fill in this parameter on CFN's UI, so it is more natural to have Yes/No options in the dropdown rather than true/false.

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
giorgosp: I find the multi-line version very frustrating since it is used in many places and bulks up the template a lot, making it harder to make sense of it. Hence the bug we had before.

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
giorgosp: It is not needed

internal/core/source_api/api/get_integration_template.go
giorgosp: Good point although it won't make a difference technically. This is the CFN condition, only 'Yes' is important
```
  HasManagedBucketNotifications: !Or
    - !Equals ['Yes', !FindInMap [PantherParameters, ManagedBucketNotifications, Value]]
    - !Equals ['Yes', !Ref ManagedBucketNotifications]
   ```

File: internal/core/source_api/api/managed_bucket_notifications.go (hunk 8 above)
Comment (giorgosp) [reply internal/core/source_api/api/managed_bucket_notifications.go:56]: Yes good point, updated

File: internal/core/source_api/api/managed_bucket_notifications.go (hunk 9 above)
Comment (giorgosp) [reply internal/core/source_api/api/managed_bucket_notifications.go:66]: Added 
```
// Create the topic if it wasn't created previously. This saves some API requests during
// source updates.
```

File: internal/core/source_api/api/managed_bucket_notifications.go (hunk 10 above)
Comment (giorgosp) [reply internal/core/source_api/api/managed_bucket_notifications.go:84]: The output (json or not) depends on the dev or prod config of the logger. So the `SugaredLogger` also outputs json. 

```
//internal/core/source_api/main/handler.go
func lambdaHandler(ctx context.Context, request *models.LambdaInput) (interface{}, error) {
	lambdalogger.ConfigureGlobal(ctx, nil)
	return router.Handle(request)
}

//pkg/lambdalogger/logger.go
func ConfigureGlobal(
	ctx context.Context,
	initialFields map[string]interface{},
) (*lambdacontext.LambdaContext, *zap.Logger) {
        ...

	// Use the same structure for all log messages so we can apply consistent metric filters.
	// We do not use zap.NewDevelopmentConfig() (even for DEBUG) because it disables json logging.
	config := zap.NewProductionConfig()
```


Since we are not in a hot path here, I preferred the SugaredLogger that does string formatting rather than the ugliness of the structured logger. 

Also, if the DEBUG level is off, the SugaredLogger doesn't even attempt string formatting, so I think we are good
```
func (s *SugaredLogger) log(lvl zapcore.Level, template string, fmtArgs []interface{}, context []interface{}) {
	// If logging at this level is completely disabled, skip the overhead of
	// string formatting.
	if lvl < DPanicLevel && !s.base.Core().Enabled(lvl) {
		return
	}
```

File: internal/core/source_api/api/managed_bucket_notifications.go (hunk 12 above)
Comment (giorgosp) [reply internal/core/source_api/api/managed_bucket_notifications.go:106]: No this is Panther's input data queue

File: internal/core/source_api/api/managed_bucket_notifications.go (hunk 14 above)
Comment (giorgosp) [reply internal/core/source_api/api/managed_bucket_notifications.go:145]: See https://github.com/panther-labs/panther/pull/2597/files#r570373557

File: internal/core/source_api/api/managed_bucket_notifications.go (hunk 13 above)
Comment (giorgosp) [reply internal/core/source_api/api/managed_bucket_notifications.go:177]: Right. That's why I passed the ARN in this function in the first place. Fixed

File: internal/core/source_api/api/managed_bucket_notifications.go (hunk 15 above)
Comment (giorgosp) [reply internal/core/source_api/api/managed_bucket_notifications.go:316]: We need the partition of Panther deployment too

File: deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml (hunk 4 above)
Comment (kostaspap) [reply deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml:59]: @giorgosp in all the other CFN templates that have a boolean decision we use true/false. See: https://github.com/panther-labs/panther/blob/master/deployments/bootstrap.yml#L57-L68

Yes/No seems ok to me, but I'd prefer to be consistent across our templates. From a user's perspective I think they can understand Yes/No as much as True/False.

File: internal/core/source_api/api/managed_bucket_notifications.go (hunk 15 above)
Comment (kostaspap) [reply internal/core/source_api/api/managed_bucket_notifications.go:322]: > We need the partition of Panther deployment too

It could be passed as parameter - we do that for the AWS Account ID anyway.

File: api/lambda/source/models/integration.go (hunk 3 above)
Comment (kostaspap) [reply api/lambda/source/models/integration.go:95]: What I mean is, do we really need to store it? Seems much simpler that we just pick a topic name and use that. Storing state in DDB is not a big deal, however DDB and real world could be out of sync.

File: deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml (hunk 6 above)
Comment (kostaspap) [reply deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml:164]: I would say maybe better add more SNS permissions for that topic -even though not needed. Could be helpful in the future, if we end up requiring more permissions. We ask users if they want to allow us to manage their SNS topic notifications, seems reasonable that we have full permissions over the topic we created :)

**Comment by giorgosp** [comment]
> Can you add more details about the E2E test?

Updated description

**Comment by kostaspap** [comment]
> * Manual e2e
>   
>   * Create source with managed notifications with some prefixes (prefixa, prefixb) - Check bucket
>   * Add some bucket notifications manually (to ensure they are not deleted by Panther), update source prefixes - Check bucket
>   * Update source with a prefix already included in the bucket - Configuration should fail
>   * Delete source - Only Panther-managed configs should be removed from the bucket

Some questions: 
> Update source with a prefix already included in the bucket - Configuration should fail

Do you mean that having two prefixes `prefixA/prefixB` and `prefixA` is no longer allowed? We want to allow overlapping prefixes - we already do that now and we have code in the log processor where longer prefixes get priority over shorter ones. 

Some other checks: 
1. Can you verify that the notifications/subscriptions are working? We need to verify that if we put an object to the S3 bucket should generate a notification that arrives in the log processor. 
2. Can you verify that cross account?

File: api/lambda/source/models/integration.go (hunk 3 above)
Comment (giorgosp) [reply api/lambda/source/models/integration.go:95]: What do you mean?

Actually what I meant is we don't delete the topic from AWS, but we should display it in the UI during source deletion

**Comment by giorgosp** [comment]
>Do you mean that having two prefixes prefixA/prefixB and prefixA is no longer allowed? 
We allow it as before. AWS doesn't, so bucket configuration fails. We handle it gracefully though on the UI.

>Some other checks:
>
  >  Can you verify that the notifications/subscriptions are working? We need to verify that if we put an object to the S3 >bucket should generate a notification that arrives in the log processor.
 >  Can you verify that cross account?

Added a file and saw it in Athena 
Screenshot 2021-02-05 at 17 22 31

All tests are cross-account.
Updated the description as well

Going to test the CFN templates a bit more

File: deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml (hunk 4 above)
Comment (giorgosp) [reply deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml]: Updated to use true/false as in bootstrap.yml

File: deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml (hunk 6 above)
Comment (giorgosp) [reply deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml:163]: Added all available SNS permissions, since it is a Panther-managed topic.

File: internal/core/source_api/api/managed_bucket_notifications.go (hunk 15 above)
Comment (giorgosp) [reply internal/core/source_api/api/managed_bucket_notifications.go:322]: Actually I could parse the lambda ARN that is included in the `context` passed to the lambda function. Something like
```
func lambdaHandler(ctx context.Context) {
        lc, _ := lambdacontext.FromContext(ctx)
        lambdaARN := arn.Parse(lc.InvokedFunctionArn)
        // use lambdaARN.Partition, lambdaARN.AccountID, etc
```
We don't even have to pass AccountID as an env variable

File: deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml (hunk 6 above)
Comment (kostaspap) [reply deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml:163]: Hmm, I don't see more permissions? I think you could add `sns:*` tbh. We are already limiting the role to have permissions only on the Panther-managed topic.

File: internal/core/source_api/api/managed_bucket_notifications.go (hunk 15 above)
Comment (kostaspap) [reply internal/core/source_api/api/managed_bucket_notifications.go:322]: Ok that would work but seems like more work to do that? It's pretty easy to add the partition and account id to the env configuration and just pass it around through the Config object. If parsing it through Lambda arn is simpler, that's fine as well.

File: deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml (hunk 6 above)
Comment (kostaspap) [reply deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml:163]: Hmm, I don't see more permissions? I think you could add `sns:*` to be honest. We are already limiting the role to have permissions only on the Panther-managed topic.  The term `Panther-managed` already implies that Panther has full permissions over it so I don't think any user would be surprised :)

File: internal/core/source_api/api/managed_bucket_notifications.go (hunk 15 above)
Comment (giorgosp) [reply internal/core/source_api/api/managed_bucket_notifications.go]: Alright I used env var for now

File: internal/core/source_api/api/managed_bucket_notifications.go [hunk 17]
```diff
@@ -0,0 +1,326 @@
...
+		return errors.Wrap(err, "failed to update bucket notifications configuration")
+	}
+
+	return nil
+}
+
+func createTopic(snsClient *sns.SNS, pantherAccountID, pantherPartition string) (*string, error) {
+	topic, err := snsClient.CreateTopic(&sns.CreateTopicInput{
+		Name: aws.String("panther-notifications-topic"),
+	})
```
kostaspap, line 160: What happens if the topic already exists e.g. if it was created by another S3 source onboarding? Wan't this fail? How is the code dealing with that?

kostaspap (approved):
Mostly nitpicks

internal/core/source_api/api/managed_bucket_notifications.go
giorgosp: `sns.CreateTopic` is idempotent

internal/core/source_api/api/managed_bucket_notifications.go
```diff
@@ -0,0 +1,326 @@
...
+		Credentials: stscreds.NewCredentials(pantherSess, source.RequiredLogProcessingRole()),
+	}))
+
+	bucketRegion, err := getBucketLocation(stsSess, source.S3Bucket)
+	if err != nil {
+		return errors.Wrap(err, "failed to get bucket location")
+	}
+	s3Client := s3.New(stsSess, &aws.Config{Region: bucketRegion})
+
+	var prefixes []string = nil // Will remove any Panther-managed notifications
```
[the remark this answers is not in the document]
giorgosp: updated

internal/core/source_api/api/managed_bucket_notifications.go
```diff
@@ -0,0 +1,326 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to update bucket notifications configuration")
+	}
+
+	return nil
+}
+
+func createTopic(snsClient *sns.SNS, pantherAccountID, pantherPartition string) (*string, error) {
+	topic, err := snsClient.CreateTopic(&sns.CreateTopicInput{
+		Name: aws.String("panther-notifications-topic"),
```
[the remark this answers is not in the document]
giorgosp: updated

giorgosp:
Thanks for your review! Addressed some of the nitpicks, merging to `feature/s3-log-source-update`

PR #2600, opened 2021-02-03
bugfix: check for lb type

lindsey-w:
## Background

Calling `GetWebAclForResource` on network load balancers will raise `WAFInvalidParameterException`. This change will check for for the lb type before calling `GetWebAclForResource`

## Changes

- check for lb type
- added unit test for network type

## Testing

- `mage test:go`

### DIFF
--- a/internal/compliance/snapshot_poller/pollers/aws/elbv2_application_load_balancer.go
+++ b/internal/compliance/snapshot_poller/pollers/aws/elbv2_application_load_balancer.go
@@ -265,9 +265,12 @@ func buildElbv2ApplicationLoadBalancerSnapshot(
 		}
 	}
 
-	// Try to find a webACL ID
-	if applicationLoadBalancer.WebAcl, err = getWebACLForResource(wafRegionalSvc, lb.LoadBalancerArn); err != nil {
-		return nil, err
+	// currently, waf is only supported in ALB
+	if *lb.Type == elbv2.LoadBalancerTypeEnumApplication {
+		// Try to find a webACL ID
+		if applicationLoadBalancer.WebAcl, err = getWebACLForResource(wafRegionalSvc, lb.LoadBalancerArn); err != nil {
+			return nil, err
+		}
 	}
 
 	return applicationLoadBalancer, nil
--- a/internal/compliance/snapshot_poller/pollers/aws/elbv2_application_load_balancer.go
+++ b/internal/compliance/snapshot_poller/pollers/aws/elbv2_application_load_balancer.go
@@ -266,7 +266,7 @@ func buildElbv2ApplicationLoadBalancerSnapshot(
 	}
 
 	// currently, waf is only supported in ALB
-	if *lb.Type == elbv2.LoadBalancerTypeEnumApplication {
+	if aws.StringValue(lb.Type) == elbv2.LoadBalancerTypeEnumApplication {
 		// Try to find a webACL ID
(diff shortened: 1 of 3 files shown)


internal/compliance/snapshot_poller/pollers/aws/elbv2_application_load_balancer.go
```diff
@@ -265,9 +265,12 @@ func buildElbv2ApplicationLoadBalancerSnapshot(
...
-	// Try to find a webACL ID
-	if applicationLoadBalancer.WebAcl, err = getWebACLForResource(wafRegionalSvc, lb.LoadBalancerArn); err != nil {
-		return nil, err
+	// currently, waf is only supported in ALB
+	if *lb.Type == elbv2.LoadBalancerTypeEnumApplication {
+		// Try to find a webACL ID
+		if applicationLoadBalancer.WebAcl, err = getWebACLForResource(wafRegionalSvc, lb.LoadBalancerArn); err != nil {
+			return nil, err
+		}
 	}
```
lindsey-w, line 274: Is it ok for `WebAcl` to be nil? I don't know what uses this upstream?

internal/compliance/snapshot_poller/pollers/aws/elbv2_application_load_balancer.go
```diff
@@ -265,9 +265,12 @@ func buildElbv2ApplicationLoadBalancerSnapshot(
 		}
 	}
 
-	// Try to find a webACL ID
-	if applicationLoadBalancer.WebAcl, err = getWebACLForResource(wafRegionalSvc, lb.LoadBalancerArn); err != nil {
-		return nil, err
+	// currently, waf is only supported in ALB
+	if *lb.Type == elbv2.LoadBalancerTypeEnumApplication {
```
austinbyers, line 269: We're sure `lb.Type` is always non-nil? This won't panic? Or should we `aws.StringValue` to be safe?

austinbyers (commented):
LGTM! I'll defer final approval to someone on the team

internal/compliance/snapshot_poller/pollers/aws/elbv2_application_load_balancer.go
nhakmiller: I think `aws.StringValue` to be safe here would be best.

internal/compliance/snapshot_poller/pollers/aws/elbv2_application_load_balancer.go
nhakmiller: It's ok for it to be nil, it will eventually be parsed out to `None`  in a python resource.

nhakmiller (approved):
Looks good (pending Austin's suggestion) but please do verify the functionality by deploying into a dev account with a network + application load balancer and running a scan.

internal/compliance/snapshot_poller/pollers/aws/elbv2_application_load_balancer.go
wey-chiang, line 269:
```suggestion
	if aws.StringValue(lb.Type) == elbv2.LoadBalancerTypeEnumApplication {
```

wey-chiang (approved):
The nil-pointer dereferencing should be addressed but besides that, LGTM

lindsey-w:
> please do verify the functionality by deploying into a dev account with a network + application load balancer and running a scan.

Confirmed - just tested with before/after change (throws exception before/does not throw exception after)

PR #2602, opened 2021-02-03
Dynamic partition/url suffix references in cfn

austinbyers:
## Background

As we prepare to add support for AWS China for enterprise customers, the first step is to make sure all of the CloudFormation is partition-aware.

We were pretty good at this already, so it was only a handful of places that were missing these.

We can fully support AWS China for the community edition once the Beijing region adds support for Cognito user pools

## Changes

- Replace all `amazonaws.com` with `${AWS::URLSuffix}` references except for service principals
- Replace all `:aws:` references with `${AWS::Partition}`, unless it's actually part of a hardcoded name

## Testing

- `mage test:cfn`
- Enterprise china deployment
- Rigorous community deploy testing will happen as part of #2524

### DIFF
--- a/deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
+++ b/deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
@@ -100,7 +100,7 @@ Resources:
               - Effect: Allow
                 Action: s3:GetBucketLocation
                 Resource: !Sub
-                  - 'arn:aws:s3:::${Bucket}'
+                  - arn:${AWS::Partition}:s3:::${Bucket}
                   - Bucket:
                       !If [
                         IsGenerated,
@@ -110,7 +110,7 @@ Resources:
               - Effect: Allow
                 Action: s3:GetObject
                 Resource: !Sub
-                  - 'arn:aws:s3:::${Bucket}/*'
+                  - arn:${AWS::Partition}:s3:::${Bucket}/*
                   - Bucket:
                       !If [
                         IsGenerated,
--- a/internal/core/source_api/api/testdata/panther-log-analysis-iam-updated.yml
+++ b/internal/core/source_api/api/testdata/panther-log-analysis-iam-updated.yml
@@ -100,7 +100,7 @@ Resources:
               - Effect: Allow
                 Action: s3:GetBucketLocation
                 Resource: !Sub
-                  - 'arn:aws:s3:::${Bucket}'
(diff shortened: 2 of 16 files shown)


deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
```diff
@@ -100,7 +100,7 @@ Resources:
               - Effect: Allow
                 Action: s3:GetBucketLocation
                 Resource: !Sub
-                  - 'arn:aws:s3:::${Bucket}'
+                  - arn:${AWS::Partition}:s3:::${Bucket}
```
s0l0ist, line 103: Do we need the quotes put back?

internal/core/source_api/api/testdata/panther-log-analysis-iam-updated.yml
```diff
@@ -100,7 +100,7 @@ Resources:
               - Effect: Allow
                 Action: s3:GetBucketLocation
                 Resource: !Sub
-                  - 'arn:aws:s3:::${Bucket}'
+                  - 'arn:${AWS::Partition}:s3:::${Bucket}'
```
s0l0ist, line 103: If we don't need the quotes from my other comment, then maybe we should take them out here as well.

internal/core/source_api/api/testdata/panther-log-analysis-iam-updated.yml
```diff
@@ -110,7 +110,7 @@ Resources:
               - Effect: Allow
                 Action: s3:GetObject
                 Resource: !Sub
-                  - 'arn:aws:s3:::${Bucket}/*'
+                  - 'arn:${AWS::Partition}:s3:::${Bucket}/*'
```
s0l0ist, line 113: If we don't need the quotes ... remove them?

s0l0ist (approved):

deployments/auxiliary/cloudformation/panther-log-analysis-iam.yml
austinbyers: Nope, YAML strings generally don't need to be quoted. There are some exceptions (e.g. `'*'` or large numbers)

internal/core/source_api/api/testdata/panther-log-analysis-iam-updated.yml
austinbyers: Sure, good idea. This is just a test file, but might as well

PR #2603, opened 2021-02-04
[FE][Feat] Allow any type of item in Selection.Context

3nvi:
## Background

Currently, when we are using `<SelectionContext>` we expect the developer to pass a list of string identifiers (IDs) to help the context undertand what's selected and what's not selected.

Unfortunately, sometimes we want access to the actual selected object (not just its ID) so that we can show some additional information to the user. With this PR this is now possible, since it extends `SelectionContext` to allow any type to be added to it, opening the door to other work that relates to mass selections.

## Changes

- Add typings to all selection-related entities
- Extend functions to support all types of selected items
- Update tests

## Testing

- `npm run test`

## Notes

The changes here are as minimal as I could make them, trying to introduce as small of a migration as possible.

The purpose of this change is to allow for an easier way to handle single-item  and multi-item deletions. It's currently optionally required by the mass actions on detections & data models

### DIFF
--- a/web/src/components/cards/AlertCard/PolicyAlertCard/PolicyAlertCard.tsx
+++ b/web/src/components/cards/AlertCard/PolicyAlertCard/PolicyAlertCard.tsx
@@ -60,7 +60,7 @@ const PolicyAlertCard: React.FC<PolicyAlertCardProps> = ({
       <Flex align="start" pr={2}>
         {selectionEnabled && (
           <Box transform="translate3d(0,-8px,0)">
-            <SelectCheckbox selectionId={alert.alertId} />
+            <SelectCheckbox selectionItem={alert.alertId} />
           </Box>
         )}
       </Flex>
--- a/web/src/components/utils/SelectContext/SelectCheckbox.tsx
+++ b/web/src/components/utils/SelectContext/SelectCheckbox.tsx
@@ -20,21 +20,22 @@ import React from 'react';
 import { Checkbox } from 'pouncejs';
 import { useSelect } from './SelectContext';
 
-interface SelectCheckboxProps {
-  selectionId: string;
+interface SelectCheckboxProps<T> {
+  selectionItem: T;
 }
 
-const SelectCheckboxComponent: React.FC<SelectCheckboxProps> = ({ selectionId, ...rest }) => {
-  const { checkIfSelected, toggleItem } = useSelect();
-  const isSelected = checkIfSelected(selectionId);
+function SelectCheckboxComponent<T>({ selectionItem, ...rest }: SelectCheckboxProps<T>) {
+  const { checkIfSelected, toggleItem } = useSelect<T>();
+  const isSelected = checkIfSelected(selectionItem);
+
   return (
     <Checkbox
       checked={isSelected}
-      aria-label={isSelected ? `unselect ${selectionId}` : `select ${selectionId}`}
-      onChange={() => toggleItem(selectionId)}
+      aria-label={isSelected ? 'unselect item' : 'select item'}
+      onChange={() => toggleItem(selectionItem)}
       {...rest}
     />
   );
-};
+}
 
-export const SelectCheckbox = React.memo(SelectCheckboxComponent);
+export default React.memo(SelectCheckboxComponent);
--- a/web/src/components/utils/SelectContext/SelectContext.tsx
+++ b/web/src/components/utils/SelectContext/SelectContext.tsx
@@ -16,39 +16,43 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-import React from 'react';
+import React, { ReactNode } from 'react';
 
-export interface SelectContextValue {
-  selection: string[];
-  selectItem: (id: string) => void;
-  deselectItem: (id: string) => void;
+const identity = (x: any) => x;
+
+export interface SelectContextValue<T> {
+  selection: T[];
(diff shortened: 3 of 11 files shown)


web/src/components/utils/SelectContext/SelectContext.tsx
```diff
@@ -16,50 +16,54 @@
...
-  initialSelection?: string[];
+  initialSelection?: T[];
+  getItemKey?: (item: T) => string;
 }
 
-const SelectProvider: React.FC<SelectProviderProps> = ({ initialSelection = [], children }) => {
-  const [selection, setSelected] = React.useState<Array<string>>(initialSelection);
+function SelectProvider<T>({
+  initialSelection = [],
+  getItemKey = identity,
```
3nvi, line 43: this functional is an optional prop to allow for more straightforward comparisongs between objects. By default `===` is used. For example:


```
// getItemKey = x => x
// `{ id: 1, title: 'hi' }` is **not equal** to `{ id: 1 }`

// getItemKey = x => x.id
// `{ id: 1, title: 'hi' }` is equal to `{ id: 1 }`
```

web/src/components/utils/SelectContext/SelectContext.tsx
```diff
@@ -122,9 +125,17 @@ const withSelectContext = (Component: React.FC) => props => (
...
+
 /** A shortcut for the consumer component */
-const SelectConsumer = SelectContext.Consumer;
+
+const SelectConsumer = <T extends unknown = string>(
+  props: React.ConsumerProps<SelectContextValue<T>>
+) => {
+  const Consumer = SelectContext.Consumer as React.Consumer<SelectContextValue<T>>;
+  return <Consumer {...props} />;
+};
```
3nvi, line 138: unfortunately, if we want to add a typing during the use of `<SelectedConsumer>` there's no easier way to do it

web/src/components/cards/AlertCard/PolicyAlertCard/PolicyAlertCard.tsx
```diff
@@ -60,7 +60,7 @@ const PolicyAlertCard: React.FC<PolicyAlertCardProps> = ({
       <Flex align="start" pr={2}>
         {selectionEnabled && (
           <Box transform="translate3d(0,-8px,0)">
-            <SelectCheckbox selectionId={alert.alertId} />
```
3nvi, line 63: `SelectCheckbox` now accepts any item. In our current use case, the item was an `alertId` which is a `string`, but we could also pass in the whole `alert` object from now on.


We don't do it so that we can minimize the changes that this PR introduces in case we create any bugs

web/src/components/utils/SelectContext/SelectContext.tsx
```diff
@@ -16,50 +16,54 @@
...
    *
    */
   const deselectItem = React.useCallback(
-    id => {
-      return setSelected(selection.filter(i => i !== id));
+    item => {
+      return setSelected(existing => existing.filter(i => getItemKey(i) !== getItemKey(item)));
     },
-    [selection]
+    [getItemKey]
```
alexmylonas, line 66:
```suggestion
    [getItemKey, existing]
```

web/src/components/utils/SelectContext/SelectCheckbox.tsx
```diff
@@ -20,21 +20,22 @@ import React from 'react';
...
+function SelectCheckboxComponent<T>({ selectionItem, ...rest }: SelectCheckboxProps<T>) {
+  const { checkIfSelected, toggleItem } = useSelect<T>();
+  const isSelected = checkIfSelected(selectionItem);
+
   return (
     <Checkbox
       checked={isSelected}
-      aria-label={isSelected ? `unselect ${selectionId}` : `select ${selectionId}`}
-      onChange={() => toggleItem(selectionId)}
+      aria-label={isSelected ? 'unselect item' : 'select item'}
```
alexmylonas, line 34: If you export `getItemKey` from context you can use it here no?

web/src/components/utils/SelectContext/SelectContext.tsx
alexmylonas: An alternative would be to pass the comparison key as a `string` but I feel this is more extendable.

How easy would it be if we make `getItemKey` required when `<T>` is not `typeof string`?

web/src/components/utils/SelectContext/SelectContext.tsx
alexmylonas: Yeah, gets complex but looks valid!

alexmylonas (commented):
Looks solid! I have a couple of suggestions but not something too bold

web/src/components/utils/SelectContext/SelectContext.tsx
3nvi: This is a function that's local to `setSelected` and managed by React. It's not a prop. 

If you see the code, we don't even have access to it outside of the `setSelected` 😉

web/src/components/utils/SelectContext/SelectCheckbox.tsx
3nvi: Talked to you about that offline. I feel that the `aria-label` is fine as is since it doesn't help us gain more accessibility. For test purposes, I'm willing to add a `data-testid`

alexmylonas (approved):

PR #2606, opened 2021-02-04
Fixing latency metrics

kostaspap:
## Background

This PR fixes event latency metric. It also switches the `panther-log-processor` Lambda to use the new metrics framework. 

## Changes

- All metrics definitions, constants, setup functions for `panther-log-processor` moved to `/log_processor/metrics` package. Eventually we can use this convention for other Lambdas as well. 
- The event latency metric is now calculated in the S3 destination, after the event has been serialized. The reason why we have to do there is because the event's `PantherEventTime` and `PantherParseTime` are calculated lazily once the event is serialized. This was the reason we were reporting wrong metrics earlier - we were trying to calculate latency eagerly and as a result we were getting wrong results
- Fixed an issue with our existing `Counter` implementation. The implementation would store a slice of all values added to the Counter. This in case of event latency would mean we might have slices millions in size. It seems that for the Counter usecase, it is sufficient just to store the sum of all the counters added. 

## Testing

- mage test:ci
- Deployed in my account, seeing the following CloudTrail delay reported ~11 minutes. 
Screen Shot 2021-02-04 at 5 26 47 PM


internal/core/metrics_api/api/log_processor_metrics.go
```diff
@@ -115,28 +104,29 @@ func getEventsLatency(input *models.GetMetricsInput, output *models.GetMetricsOu
...
-				Stat:   aws.String("Sum"),
-				Unit:   aws.String(metrics.UnitMilliseconds),
+		queries = append(queries,
+			&cloudwatch.MetricDataQuery{
+				Id: aws.String("latency_query_" + index),
+				MetricStat: &cloudwatch.MetricStat{
+					Metric: metric,
+					Period: aws.Int64(input.IntervalMinutes * 60), // number of seconds, must be multiple of 60
+					Stat:   aws.String("Sum"),
+					Unit:   aws.String(metrics.UnitSeconds),
```
kostaspap, line 114: Switched to using `Seconds`. It seems more appropriate to use second accuracy for event delay.

internal/log_analysis/log_processor/classification/classifier.go
```diff
@@ -163,9 +163,6 @@ func (c *Classifier) Classify(log string) (*ClassifierResult, error) {
 		parserStat.BytesProcessedCount += uint64(len(log))
 		parserStat.LogLineCount++
 		parserStat.EventCount += uint64(len(result.Events))
-		for _, event := range parsedEvents {
-			parserStat.CombinedLatency += uint64(event.PantherParseTime.Sub(event.PantherEventTime).Milliseconds())
-		}
```
kostaspap, line 168: We no longer calculate the latency here - we do it after the events are serialized

internal/log_analysis/log_processor/classification/classifier.go
```diff
@@ -204,7 +201,6 @@ type ParserStats struct {
 	BytesProcessedCount    uint64 // input bytes
 	LogLineCount           uint64 // input records
 	EventCount             uint64 // output records
-	CombinedLatency        uint64 // sum of latency of events
```
kostaspap, line 207: I started to remove this struct as part of this PR but the change ended up being too big. 
I will remove in subsequent PR as it is no longer needed - we could use the metrics framework to extra this information. Remove this will simplify the code significantly and also saves quite few allocations.

pkg/metrics/counter.go
```diff
@@ -124,9 +126,12 @@ func (s *Space) nodeFor(name, unit string) *node {
...
 type node struct {
-	unit         string
-	mtx          sync.RWMutex
-	observations []float64
+	unit string
+	mtx  sync.RWMutex
+	// Sum of all observations
+	sum float64
+	// number of observations
+	observations int64
```
kostaspap, line 134: As mentioned in description, this can be VERY inefficient for high volume events. It is also not needed for the simple counters we use.

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -356,6 +362,7 @@ func (bs *s3EventBufferSet) writeEvent(event *parsers.Result) (sendBuffers []*s3
 	if buf == nil {
 		return nil, errors.New(`could not resolve a buffer for the event`)
 	}
+	buf.latencyCounter.Add(float64(event.PantherParseTime.Sub(event.PantherEventTime).Milliseconds()))
```
georgepsarakis: Just a question here, not sure if I understand the purpose entirely, or if my suggestion actually simplifies things: do we need this level of precision? Couldn't we just maintain an approximate event time minimum distance from the current time (`now - minimum(PantherEventTime)`)? My guess is that latency is usually in the order of minutes.

internal/core/metrics_api/api/log_processor_metrics.go
```diff
@@ -26,43 +26,32 @@ import (
...
 	// Build the query based on the applicable metric dimensions
 	var queries []*cloudwatch.MetricDataQuery
 	for i, metric := range listMetricsResponse {
-		if len(metric.Dimensions) != 1 {
-			// This if statement is only needed by developers who have deployed the unstable branch
-			// of Panther before v1.6.0. Old metrics can't be deleted and you can't filter out
-			// dimensions you don't want, so we have to skip metrics where the Component dimension
-			// still exists.
-			continue
-		}
```
kostaspap, line 65: @nhakmiller I assume it is safe to remove this one now? This seemed a bit restrictive

pkg/metrics/counter.go
georgepsarakis: This seems quite more performant. Was a slice being used in order to extract other stats, such as median, percentiles etc?

kostaspap: go-kit/metrics uses the same struct for storing counters, gauges, histograms. So they need to store every recorded value to be able to extra percentiles and such. 
In our case though we don't need that, it is way more performant to use this approach. My thought is that when the need arises and we have to keep all observations, we can add such functionality as appropriate.

internal/log_analysis/log_processor/destinations/s3.go
kostaspap: Good catch ;) I actually was planning to switch this one to seconds - you were faster at adding the comment. 

I would still like to keep second granularity though - we can have cases where the delay will be subminute.

internal/core/metrics_api/api/log_processor_metrics.go
```diff
@@ -26,43 +26,32 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/panther-labs/panther/api/lambda/metrics/models"
+	logmetrics "github.com/panther-labs/panther/internal/log_analysis/log_processor/metrics"
 	"github.com/panther-labs/panther/pkg/metrics"
 )
 
-const (
-	eventsProcessedMetric = "EventsProcessed"
-	eventsLatencyMetric   = "CombinedLatency"
```
kostaspap, line 34: The new metric is called EventLatency. I think this shouldn't be an issue, right @nhakmiller ? If anything, but using a different name, we will make sure no old "trash" data make it to the UI.

internal/log_analysis/log_processor/destinations/s3.go
georgepsarakis: That's true, my comment was mostly referring to whether we need to compute and store a counter value on every event. Instead we could just maintain the minimum event time of all events and make a call to counter just once in `SendEvents`. Not sure if records are interleaved, but `s3EventBufferSet` seems to be grouping events per log type.

kostaspap: Aha, got it. 

The previous metrics library would write one entry to CW every time a metric was added to the relevant metrics manager. That meant that we have to be very careful about when we were invoking it. Many times we had to code our way around that, to make sure the application was collecting statistics in an efficient matter, then emitting the metrics. The [Classifier stats](https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/classification/classifier.go#L183-L209) is an example of this. That was a bit annoying though, we had to write a lot of boilerplate to generate metrics from hot code paths, as classifier stats demonstrates.

The new metrics library is meant to deal with that as it is every efficient at collecting metrics and publishing to CW only periodically (every 1 minute). We might still need to code our away around things for really hot paths - e.g. here we could potentially do some sampling (I wouldn't pick `min` tbh) and only calculate the metrics for 1/2/5/10% of the events. 
I am not sure its needed here though, and I am thinking it's better to avoid complicating the logic of the S3 destination and just calculate the diff for every event. 

We could do some sampling in the future though if we think it becomes a problem!

internal/core/metrics_api/api/log_processor_metrics.go
```diff
@@ -115,28 +104,29 @@ func getEventsLatency(input *models.GetMetricsInput, output *models.GetMetricsOu
...
 		// Add the latency query
 		index := strconv.Itoa(i)
-		queries = append(queries, &cloudwatch.MetricDataQuery{
-			Id: aws.String("latency_query_" + index),
-			MetricStat: &cloudwatch.MetricStat{
-				Metric: metric,
-				Period: aws.Int64(input.IntervalMinutes * 60), // number of seconds, must be multiple of 60
-				Stat:   aws.String("Sum"),
-				Unit:   aws.String(metrics.UnitMilliseconds),
+		queries = append(queries,
```
kostaspap, line 107: @alxarch in the end I am just using `Counter` - I didn't go for `AVG Counter`. To calculate avg delay, the code was already just dividing the sum of delays with the num of events. I just stayed with that approach.

giorgosp (approved):
LGTM

PR #2607, opened 2021-02-04
fix: Autosave to skip submit when mounting

alexmylonas:
## Background

In several forms, we are using a `FormikAutoSave` component that allows the form to auto-submit when values change.

That component doesn't account for initial values and therefore triggers submission on render.

## Changes

- Added `useRef` for initial mount

## Testing

- Test some cases that this was apparent.


vorillaz (approved):
Nicely done.

PR #2611, opened 2021-02-04
OPSTOOL: Flush panther-resources table items where deleted=true

tatehanawalt:
## Background

Clients are experiencing availability issues caused by the in memory allocations required for queries on the panther-resources table.

Specifically, a client has a few hundred thousand entries in this table. A significant portion fo these table items are scheduled for removal but have not yet been removed.

This tool flushes the panther-resources table of items pending removal as a quick fix to solve in memory requirements causing the availability issues.

I initially tried to use the batch delete package but was running into issues. This is more of a one off script so I have the deleteItem in place for quick delivery. 

## Changes

Added a script in the ops tools directory which deletes entries in the resources table where the field 'deleted' has been set to true

## Testing

NOTE: there are probably deleted resources existing in an initial deployed table so the first time you run this on a new deployment you will delete table entries. 

For subsequent runs:

Run the script
verify no items deleted
Set table entries deleted=true for any number of table items
Run the script
query the table for identifiers that should have been deleted to verify the id's are not found in the table


cmd/opstools/flush_resource_table_deleted/main.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/aws/aws-sdk-go/service/dynamodb/expression"
+)
+
+const tableName = "panther-resources"
+const delVal = true
+
+func main() {
+
+	// AWS Session construction
+	sessOpts := session.Options{SharedConfigState: session.SharedConfigEnable}
```
austinbyers, line 39: I've never seen this before, do we need it?

cmd/opstools/flush_resource_table_deleted/main.go
```diff
@@ -0,0 +1,83 @@
...
+	awsSession := session.Must(session.NewSessionWithOptions(sessOpts))
+	client := dynamodb.New(awsSession)
+
+	// Query expression
+	proj := expression.NamesList(expression.Name("id"), expression.Name("deleted"))
```
austinbyers, line 44: We don't need the `deleted` column in the projection

cmd/opstools/flush_resource_table_deleted/main.go
```diff
@@ -0,0 +1,83 @@
...
+	filt := expression.Name("deleted").Equal(expression.Value(delVal))
```
austinbyers, line 45: Haha I don't think we need a `delVal` constant. You can just put `true` in the expresion value

cmd/opstools/flush_resource_table_deleted/main.go
```diff
@@ -0,0 +1,83 @@
...
+	expr, err := expression.NewBuilder().WithFilter(filt).WithProjection(proj).Build()
+	if err != nil {
+		fmt.Println("Got error building expression:")
+		log.Fatal(err)
```
austinbyers, line 49: Or, to combine these:

```go
log.Fatalf("error building expression: %s", err)
```

cmd/opstools/flush_resource_table_deleted/main.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	// Query Input
```
austinbyers, line 52: Lies! This is a scan, not a query. Also not a helpful comment lol

cmd/opstools/flush_resource_table_deleted/main.go
```diff
@@ -0,0 +1,83 @@
...
+	input := &dynamodb.ScanInput{
+		// ConsistentRead:       aws.Bool(true),
```
austinbyers, line 54: Remove

cmd/opstools/flush_resource_table_deleted/main.go
```diff
@@ -0,0 +1,83 @@
...
+		ExpressionAttributeValues: expr.Values(),
+		FilterExpression:          expr.Filter(),
+		ProjectionExpression:      expr.Projection(),
+		TableName:                 aws.String(tableName),
+	}
+
+	// Scan for deleted entries
+	err = client.ScanPages(input, func(page *dynamodb.ScanOutput, lastPage bool) bool {
+		for _, item := range page.Items {
+			fmt.Printf("%v\n", item)
```
austinbyers, line 65: I assume you just had this for debugging, but it definitely needs to be removed. The customer has more than 300,000 entries, remember.

You could add a log statement for every page or every 10,000 items or something to show progress

cmd/opstools/flush_resource_table_deleted/main.go
```diff
@@ -0,0 +1,83 @@
...
+			_, err := client.DeleteItem(&dynamodb.DeleteItemInput{
+				Key: map[string]*dynamodb.AttributeValue{
+					"id": item["id"],
+				},
+				TableName: aws.String(tableName),
+			})
```
austinbyers, line 71: This will be far too slow for 300,000 items, we definitely need the delete batching. I can help you figure that out tomorrow

cmd/opstools/flush_resource_table_deleted/main.go
```diff
@@ -0,0 +1,83 @@
...
+			if err != nil {
+				log.Fatal(err)
+			}
+		}
+		return true
+	})
+
+	if err != nil {
+		fmt.Println("Got error Scanning Pages:")
+		log.Fatal(err)
```
austinbyers, line 81: Consider putting most of the functionality in a separate function that can return the error, that way you only have to log it once in `main`

cmd/opstools/flush_resource_table_deleted/main.go
tatehanawalt: This just sets the sdk to create a session using credentials in the order of 
```
* Environment Variables
* Shared Credentials file
* Shared Configuration file (if SharedConfig is enabled)
* EC2 Instance Metadata (credentials only)
```

It just lets the executor basically use their aws credentials however they configure their environment

cmd/opstools/flush_resource_table_deleted/main.go
tatehanawalt: Renamed the constant. I have these for readability.  deleted and tablename are the only parameters for this script so its helpful to see where they're specified and where they're used.

cmd/opstools/flush_resource_table_deleted/main.go
austinbyers: But that should already be the default - the SDK knows to get credentials in that order anyway. Check our other opstools, I don't think we have to set this explicitly

cmd/opstools/flush_resource_table_deleted/main.go
austinbyers: Constant values for strings makes sense. But for a bool? Maybe it's just my preference, but that actually makes it harder to read in my opinion 🤷

cmd/opstools/flush_resource_table_deleted/main.go
tatehanawalt: query is a precise request for information retrieval with database and information systems. technically its a query using the aws scanpages method.

The comment is intended for someone who isn't specifically familiar with aws, adding some context to the comment

austinbyers: In DynamoDB, "Query" and "Scan" are two distinct operations. They have very different meanings

* A Query implies a specific partition key or range of partition keys
* A Scan is a full table search, like you are doing here

cmd/opstools/flush_resource_table_deleted/main.go
tatehanawalt: I will push changes and you can take a look again. I find it nice to use the constant definition to locate where it's used in the method body

cmd/opstools/flush_resource_table_deleted/main.go
tatehanawalt: I have the BatchWriteItem working. I'm not sure what was happening yesterday but I was getting an error related the attribute names? 

anyways I have the script using the batch method now.

cmd/opstools/flush_resource_table_deleted/main.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/panther-labs/panther/pkg/awsbatch/dynamodbbatch"
+)
+
+// Explicitely delete columns in the panther-resources table where the table entry deleted is equal
+// to the entrydeleted contant
+const tableName = "panther-resources"
+const entrydeleted = true
+
+// Max Back Off is used in the request dynamodb batch write items
+const maxBackoff = 10 * time.Second
```
austinbyers, line 41: I'd raise this to 1 minute or so, there will probably be a bit of backoff for large tables

cmd/opstools/flush_resource_table_deleted/main.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/aws/aws-sdk-go/service/dynamodb/expression"
+
+	"github.com/panther-labs/panther/pkg/awsbatch/dynamodbbatch"
+)
+
+// Explicitely delete columns in the panther-resources table where the table entry deleted is equal
+// to the entrydeleted contant
```
austinbyers, line 36: typo: `explicitly` and `constant` (the first typo is what's actually failing CI)

austinbyers (approved):
Awesome! Looks good! Thanks for the quick fix.

When this merges to enterprise, be sure to add a new list item [in the opstools documentation](https://docs.runpanther.io/help/ops-home#tools) as well

tatehanawalt:
TLDR: The summary at the bottom and everything after it will 

After reviewing the other ops tools I realized mine wasn't conforming to the structure of the other tools. 
I made some pretty significant changes to adopt the structure of the other tools.
The functionality is the same, however there are a few new options.

The commands are now:
```
  -debug
    	Enable debug logging
  -flush
    	Remove entries from the panther-resources table where deleted=true
  -inspect
    	Print number of panther-resources entries where delete=true and the estimated save file size
  -save
    	Save Id's of panther-resources entries where delete=true to ./flush_resource_ids_<start_epoch>
  -version
    	Print ARCH, BIN, OS, and VERSION
```

Most notable there are two additions:
* -save - This will save the item id's of resources table deleted=true entries to ./flush_resource_ids_<start epoch>
* -inspect - This method returns the number of items where deleted=true and the size the file would be from the -save command.

To actually flush a table, use the -flush option.

inspect will skip flush and save. So if your command is `-inspect -flush` nothing will be flushed. Or with save `-inspect -save` would not result in any files saved.

The functionality for inspect, save, and flush is defined in the `FlushSaveInspectResources(svc *dynamodb.DynamoDB, saveWriter io.Writer, flush, save, inspect bool)` Method.

This method accepts an aws dynamodb client, a writer used to write items ONLY when the 'save' parameter is true, followed by flush, save, and inspect.

This method is written to be extensible should someone ever need to re-purpose it in another tool.

Summary: 
The implementation is basically 2 parts:
1. CLI that establishes an aws dynamodb client, a writer (the save file) when the caller needs to save the id's for some reason, and the functionality to parse and use cli inputs
2. The `FlushSaveInspectResources` method which actually carries out the flush, inspect, and save operations.

cmd/opstools/flush_resource_table_deleted/main.go
tatehanawalt: I did end up removing this constant

s0l0ist (approved):

cmd/opstools/flushrsc/flushrsc/main.go
```diff
@@ -0,0 +1,341 @@
...
+	return
+}
+
+// returns values of the required cli options where true means the option was specified in the cli args.
+func getOpts() (debug, flush, inspect, save, versionOpt bool) {
+	flag.Usage = usage
+	cliDebug := flag.Bool("debug", false, "Enable debug logging")
+	cliFlush := flag.Bool("flush", false, "Remove entries from the panther-resources table where deleted=true")
+	cliInspect := flag.Bool("inspect", false, "Print number of panther-resources entries where delete=true and the estimated save file size")
+	cliSave := flag.Bool("save", false, "Save Id's of panther-resources entries where delete=true to ./flush_resource_ids_<start_epoch>")
```
austinbyers, line 280: I don't see the need for the `save` command - you don't offer a way to read back from that file nor do I think we need to - the file will be out of date almost immediately after its creation.

Since you have it finished, we may as well keep it, but try not to scope creep so much, it will save you lots of time if you don't add unnecessary fluff

cmd/opstools/flushrsc/flushrsc/main.go
```diff
@@ -0,0 +1,341 @@
...
+	printfln("  Scan\n")
+	printfln("USAGE:\n\n  %v <options>\n", BIN)
+	printfln("Where options are:\n")
+	flag.PrintDefaults()
+	printfln()
+	os.Exit(0)
+}
+func printfln(args ...interface{}) {
+	if len(args) == 1 {
+		fmt.Fprintf(flag.CommandLine.Output(), args[0].(string))
```
austinbyers, line 322: Shouldn't we be using `log` instead of printf? E.g. `log.Infof`

cmd/opstools/flushrsc/flushrsc/main.go
```diff
@@ -0,0 +1,341 @@
...
+	} else if len(args) > 1 {
+		fmt.Fprintf(flag.CommandLine.Output(), args[0].(string), args[1:]...)
+	}
+	fmt.Fprintf(flag.CommandLine.Output(), "\n")
+}
+func humanByteSize(b int64) string {
```
austinbyers, line 328: This code is essentially duplicated from https://github.com/panther-labs/panther-enterprise/blob/master/tools/mage/util/files.go#L81

Import the routine from `mage`, or move it out to `pkg`, if you prefer, but let's not duplicate the code here

tatehanawalt: Didn't see the license where I saw that bytescountsi method so I didn't want to add it but I just found the license so yeah my method can be scrapped and that one used in place. 

Only thing is my method prints bytes instead of just B when size < 1000 so I may handle that for readability

cmd/opstools/flushrsc/flushrsc/main.go
```diff
@@ -0,0 +1,341 @@
...
+
+const tableName = "panther-resources"
+const maxBackoff = 60 * time.Second
+
+// version set by mage build:tools
+var version string
+
+func check(err error) {
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
tatehanawalt: I have a recover function that handles every error and its just printing the error message along with some clean up stuff.

Example
```shell
2021-02-17T12:06:48.987-0600	INFO	SAVE
2021-02-17T12:06:49.284-0600	INFO	/Users/tatehanawalt/go/src/github.com/panther-labs/panther/out/bin/cmd/opstools/flush_resource_ids_1613585208
2021-02-17T12:06:49.285-0600	ERROR	save encountered an error... removing save file
2021-02-17T12:06:49.285-0600	ERROR	AWS NoCredentialProviders Error: No valid providers in chain
2021-02-17T12:06:49.285-0600	INFO	Double check your aws credentials
```

cmd/opstools/flushrsc/flushrsc/main.go
tatehanawalt: I primarily use this for the usage. I still have the printfln its just an anonymous function inside the usage method now

tatehanawalt:
@all-contributors please add @tatehanawalt  for code infra

PR #2612, opened 2021-02-04
Minor cleanup for new pkg code

austinbyers:
## Background

#2524 was a large PR which replaced the `mage` packaging logic. As devs have started using it, there were a few small fixes needed

## Changes

- The YAML v3 library uses 4 spaces for indentation by default, who knows why. I've lowered it to 2 spaces when writing packaged templates: it takes quite a bit less space and it is easier to see deeply nested CFN settings in the web console
- The `bootstrap` stack still needs to be "packaged" to strip comments or else it will be too large (in enterprise) to deploy directly. This stack cannot have S3 packaging in the old deployment model because it is itself the stack which creates the S3 buckets
- The problem with large numbers being written in scientific notation appears to be resolved in the YAML v3 parser. I've removed all the YAML quotes around large numbers added in #2524
- When deploying individual Lambda functions, we no longer have to strip the `out/bin` prefix from the source paths, because #2524 changed the path in the templates to the source rather than the compiled binary

## Testing

Fresh deploy: `AWS_REGION=us-west-1 mage deploy`

### DIFF
--- a/tools/mage/deploy/template.go
+++ b/tools/mage/deploy/template.go
@@ -50,15 +50,11 @@ func Stack(
 ) (map[string]string, error) {
 
 	// 1) Generate packaged template, packaging assets in S3 and ECR
-	packagedTemplate := templatePath
-	if packager != nil {
-		packager.Log.Debugf("packaging %s to s3 bucket %s and ecr registry %s",
-			templatePath, packager.Bucket, packager.EcrRegistry)
-		var err error
-		packagedTemplate, err = packager.Template(templatePath)
-		if err != nil {
-			return nil, err
-		}
+	packager.Log.Debugf("packaging %s to s3 bucket %s and ecr registry %s",
+		templatePath, packager.Bucket, packager.EcrRegistry)
+	packagedTemplate, err := packager.Template(templatePath)
+	if err != nil {
+		return nil, err
 	}
 
 	// 2) If the stack already exists, wait for it to reach a steady state.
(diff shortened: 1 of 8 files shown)


tools/mage/deploy/template.go
```diff
@@ -50,15 +50,11 @@ func Stack(
 ) (map[string]string, error) {
 
 	// 1) Generate packaged template, packaging assets in S3 and ECR
-	packagedTemplate := templatePath
-	if packager != nil {
```
austinbyers, line 54: The packager is now always guaranteed to be defined

austinbyers: Actually, no it is not - the `panther-dev` stack is deployed with a nil packager currently. Reverting the change to this file

AlexPanther (approved):

tools/mage/test/cfn.go
```diff
@@ -70,20 +70,7 @@ func testCfnLint() error {
...
-	// which CFN does not understand. So we force string values to serialize them correctly.
-	args := []string{
-		"-x", "E3012:strict=false",
-		// cfn-lint complains: E3002 Invalid Property Resources/WebApplicationServer/Properties/NetworkConfiguration/AwsvpcConfiguration
-		//deployments/web_server.yml:208:9 .
-		// nolint:lll
-		// However this property is valid: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-service-networkconfiguration.html#cfn-ecs-service-networkconfiguration-awsvpcconfiguration
-		"-i", "E3002",
-		"-i", "W3002", // warns about templates which require packaging; ours all do
-		"--"}
```
austinbyers, line 86: I realized that with the latest versions of `go-yaml` and `cfn-lint`, we can restore most of the cfn-lint checks, which is awesome!

PR #2620, opened 2021-02-05
on call: update ignore list

lindsey-w:
## Background

Ignore two new ec2 cases: `SharedSnapshotCopyInitiated` and `SharedSnapshotVolumeCreated` These appear to be _internal_ aws api calls

## Changes

- update ignore list

## Testing

- `mage test:go`


nhakmiller (approved):
Austin and I have been discussing that a Cloud Security 2.0, designed to be more scalable and less error prone is well overdue. Every time I see a PR that patches this list, that just reinforces that assessment in my mind.

Issue #2615, opened 2021-02-05
Resources & Compliance API timeouts


nhakmiller:
### Describe the bug

The `resources-api` and possibly the `compliance-api` will consistently timeout when trying to describe organizations with very large numbers of resources. This is due to the fact that because of the design of the table and our access patterns we must scan the entire table into memory in order to process the results.

### Steps to reproduce

1. onboard an AWS account with several hundred thousand resources in it as a Cloud Security source
2. Wait for the initial scan to complete
3. Go to the resources page
4. Observe the page timeout

### Expected behavior

Panther should be able to show the list resources page regardless of how many resources have been scanned.

### Environment

- Panther 1.15.X

### Additional context

Despite the size of the table, we are not primarily hitting memory issues in scanning it because we are doing a projection to just grab a subset of fields, and are excluding the biggest field (`attributes`) in most cases. The issue appears to be primarly dynamodb throttling and dynamodb scan times.

PR #2623, opened 2021-02-06
Resources timeout fixes

nhakmiller:
## Background

Querying the `resources-api` to list resources always requires a full scan of the `resources-table`. This full scan can timeout (either just the UI's 30 second timeout or even the full 120 second timeout of the lambda itself) if there are a large number of resources in the `resources-table`.

This change adds two things to help speed up scans.

First, it adds segmented scanning of the table. Segmented scans allow the table to be scanned in parallel. I selected a segment number of 5 based on my benchmarks (see below), but it's configurable by environment variable.

Secondly, this allows for provisioned capacity of the `resources-table` via deployment parameters. If not set, the table will still use default on demand capacity. But this allows users to set a minimum and maximum read capacity, and a minimum and maximum write capacity if desired.

I had intended to try to abstract this somehow, e.g. customers would configure high/medium/low throughput, or they would put in an "estimated resource count" number and we would use that to calculate what these numbers should be. But I figured it would take a lot of configuring to get that right, 95% of customers are just going to use on demand billing anyways, and for the 5% that need this feature we will likely be working close with them to make sure everything is going smooth anyways. In the hosted environments, the customers won't even need to know about this at all.

## Changes

- Change scanning of the resources-table to be segmented scanning
- Add configuration options for provisioned capacity on the `resources-table`

## Testing

- Deployed into my development account and did lots of benchmarking. Here are my results:

<details>
<summary>Benchmarks</summary>

I ran essentially two sets of tests across three variables.

The first set of tests was against a `resources-table` with 130,504 resources in it, about equal to what we saw with a customer that was just barely experiencing timeouts. The second set of tests was run against a table with 432,204 resources in it, which is about equal to the most resources we've seen in a customers table (well actually a bit less than the most we've seen, but close enough to provide a good estimation I think).

The three variables I tested were lambda memory, Dynamo Read Capacity Units (RCUs), and parallelization factor (number of segments) in scan.

The majority of the tests were at 512 MB lambda memory, but I ran just a handful of tests at higher memory for comparison.

For RCUs, I ran tests at 10, 100, 500, and 1000 RCUs. Dynamo only lets you scale down provisioned capacity 5 times per day before you get throttled and then can only provision down capacity once every hour. This caused a number of my tests to be run at 1026 RCUs, since I would have had to wait an hour to down provision the table to 1000 RCUs. I figured this number was close to not change the tests much, but I still included that value in the table for clarity's sake.

For parallelization, I ran most tests at 1, 5, and 10 segments. Sometimes I skipped a test if I was fairly sure the scan would timeout anyways so as not to waste time, and I ran one test at 20 segments.

Keep in mind that any duration over 120,000 timed out the lambda (so the true time is not known), and that any duration over 30,000 timed out the UI (so is not an acceptable speed).

Lambda Memory | Resources | RCUs | Parallelization | Duration (ms) | Memory (mb)
-- | -- | -- | -- | -- | --
512.00 | 130,504.00 | 10.00 | 1.00 | 120,000.00 | 76.00
512.00 | 130,504.00 | 10.00 | 1.00 | 120,000.00 | 55.00
512.00 | 130,504.00 | 10.00 | 5.00 | 120,000.00 | 82.00
512.00 | 130,504.00 | 100.00 | 1.00 | 50,484.00 | 177.00
512.00 | 130,504.00 | 100.00 | 1.00 | 54,132.00 | 181.00
512.00 | 130,504.00 | 100.00 | 10.00 | 19,037.00 | 259.00
512.00 | 130,504.00 | 100.00 | 10.00 | 18,409.00 | 251.00
512.00 | 130,504.00 | 100.00 | 5.00 | 19,980.00 | 206.00
512.00 | 130,504.00 | 100.00 | 5.00 | 17,999.00 | 209.00
512.00 | 130,504.00 | 500.00 | 1.00 | 19,974.00 | 177.00
512.00 | 130,504.00 | 500.00 | 1.00 | 18,822.00 | 178.00
512.00 | 130,504.00 | 500.00 | 10.00 | 18,776.00 | 229.00
512.00 | 130,504.00 | 500.00 | 10.00 | 17,072.00 | 259.00
512.00 | 130,504.00 | 500.00 | 5.00 | 18,188.00 | 210.00
512.00 | 130,504.00 | 500.00 | 5.00 | 16,871.00 | 220.00
512.00 | 130,504.00 | 1,026.00 | 5.00 | 17,020.00 | 210.00
512.00 | 130,504.00 | 1,026.00 | 5.00 | 16,752.00 | 221.00
512.00 | 130,504.00 | 1,026.00 | 10.00 | 17,546.00 | 239.00
512.00 | 130,504.00 | 1,026.00 | 10.00 | 18,144.00 | 251.00
1,024.00 | 130,504.00 | 1,026.00 | 10.00 | 9,907.00 | 228.00
512.00 | 130,504.00 | 1,026.00 | 20.00 | 18,846.00 | 268.00
512.00 | 130,504.00 | 1,026.00 | 20.00 | 17,655.00 | 283.00
512.00 | 130,504.00 | 1,026.00 | 20.00 | 19,001.00 | 262.00
512.00 | 130,504.00 | 1,026.00 | 1.00 | 18,903.00 | 176.00
512.00 | 130,504.00 | 1,026.00 | 1.00 | 18,197.00 | 178.00
512.00 | 432,204.00 | 500.00 | 1.00 | 62,038.00 | 412.00
512.00 | 432,204.00 | 500.00 | 1.00 | 61,074.00 | 414.00
2,048.00 | 432,204.00 | 500.00 | 1.00 | 36,632.00 |  
512.00 | 432,204.00 | 500.00 | 5.00 | 56,927.00 | 512.00
512.00 | 432,204.00 | 500.00 | 5.00 | 54,896.00 | 512.00
512.00 | 432,204.00 | 500.00 | 10.00 | 60,029.00 | 512.00
512.00 | 432,204.00 | 500.00 | 10.00 | 120,000.00 | 512.00
512.00 | 432,204.00 | 1,000.00 | 1.00 | 59,617.00 | 408.00
512.00 | 432,204.00 | 1,000.00 | 1.00 | 58,565.00 | 411.00
1,024.00 | 432,204.00 | 1,000.00 | 1.00 | 40,782.00 | 408.00
512.00 | 432,204.00 | 1,000.00 | 5.00 | 56,393.00 | 512.00
1,024.00 | 432,204.00 | 1,000.00 | 5.00 | 29,042.00 | 528.00
1,024.00 | 432,204.00 | 1,000.00 | 10.00 | 28,602.00 | 585.00

Interesting to note that there was not one single OOM error in any of the benchmarks. Even on the benchmarks where the memory consumed was exactly equal to the memory provisioned, the lambda did not OOM. Presumably there is some swap space available, although I'm not sure how much.

It seems that there is almost always a noticeable decrease in duration going from 1 to 5 segments, but generally not much of a decrease going from 5 to 10 segments. The most noticeable performance increase from 1->5 segments was at 100 RCUs, where the duration was reduced by nearly 60%. At higher RCU counts, the 1->5 segment performance shaved off a few seconds at most, closer to 5-10% performance increase.

In the smaller tests, increasing the memory to 1024 showed massive performance gains. This was less pronounced, although still significant, in the larger tests.

My conclusion was that 1024MB of memory + 1000 RCUs would be just sufficient for biggest, heaviest users of Cloud Security. With the recent changes to reduce the deleted resource TTL from 30 days -> 2 days (which will result in huge decreases in table size for high churn environments), I think this should be sufficient.

For our more standard heavy users, just a mere 100 RCUs and the 5x segmentation appear to be sufficient to keep the resources-table well within scannable parameters.

I also found it actually took quite some time for dynamodb to scale up to meet demand. In my tests there was no scaling going on, but in preparation for the tests I found the autoscaling to be pretty insufficient. I think if customers want to use this, they will not able to keep RCUs low and just let autoscaling drive them up as needed. They will need to set a pretty high floor to maintain responsiveness.
</details>

Closes https://github.com/panther-labs/panther/issues/2615 for now, although this will likely need to be revisited for a more long term solution at some point.


deployments/cloud_security.yml
```diff
@@ -1005,6 +1026,56 @@ Resources:
         AttributeName: expiresAt
         Enabled: true
 
+  ResourcesTableWriteCapacityScalableTarget:
```
nhakmiller, line 1029: I haven't actually tested this CloudFormation yet, as I was doing my tests by tweaking the table by hand. I'll be sure to test the deployments out a few times with different values.

internal/compliance/resource_processor/processor/resources_api.go
```diff
@@ -27,7 +27,7 @@ import (
 
 // How many resources (with attributes) we can request in a single page.
 // The goal is to keep this as high as possible while still keeping the result under 6MB.
-const resourcePageSize = 2000
+const resourcePageSize = 200
```
nhakmiller, line 30: Each resource is at most 400Kb, so 2000 * 400Kb was way over the 6MB lambda limit. I brought this down to 200, which is still slightly over the limit but since most resources average ~2Kb I think we should be much safer than before at least.

internal/compliance/resources_api/handlers/dynamo.go
```diff
@@ -106,39 +107,96 @@ func doUpdate(update expression.UpdateBuilder, resourceID string) *events.APIGat
...
+type scanResult struct {
+	resources []models.Resource
+	err       error
+}
 
-		for _, entry := range items {
-			if handlerErr = handler(entry); handlerErr != nil {
-				return false // stop paginating
+// Wrapper around dynamoClient.ScanPages that accepts a handler function to process each item.
+func scanPages(inputs []*dynamodb.ScanInput, includeCompliance bool,
```
nhakmiller, line 116: Due to the need to do parallel scanning, it didn't work to pass in a handler func that would process each page and append it to a slice (unless it's thread safe to append to a slice?), so I just inline most of the functionality of the handler function. The `ListResources` API is the only that makes use of this `scanPages` endpoint anyways, so it wasn't really an issue.

internal/compliance/resources_api/handlers/dynamo.go
```diff
@@ -106,39 +107,96 @@ func doUpdate(update expression.UpdateBuilder, resourceID string) *events.APIGat
...
+				}
+
+				for _, entry := range items {
+					if !includeCompliance {
+						segmentResources = append(segmentResources, entry.Resource(""))
+						return true
+					}
+
+					var status *complianceStatus
+					status, handlerErr = getComplianceStatus(entry.ID)
```
nhakmiller, line 151: This does some initialization stuff if the cache isn't populated, but I think it's safe to do that in parallel.

internal/compliance/resources_api/handlers/list_resources.go
```diff
@@ -52,15 +53,17 @@ var (
 
 // ListResources returns a filtered list of resources.
 func (API) ListResources(input *models.ListResourcesInput) *events.APIGatewayProxyResponse {
+	zap.L().Info("entering ListResources", zap.Any("input", input))
```
nhakmiller, line 56: Whoops this and some of the other `zap.L().Info("")`s are leftover debugging output. I'll remove them.

nhakmiller:
The provisioning settings don't work for `mage deploy`, not sure if I still need to add support for this deployment method if we're planning to remove it soon.

internal/compliance/resources_api/handlers/dynamo.go
```diff
@@ -106,39 +107,94 @@ func doUpdate(update expression.UpdateBuilder, resourceID string) *events.APIGat
...
-		return err
+	// Merge scan results
+	zap.L().Info("scans initiated, awaiting results")
+	var err error
+	var mergedResources []models.Resource
+	for range inputs {
+		result := <-results
+		if result.err != nil {
+			err = result.err
+			continue
```
nhakmiller, line 194: I wasn't sure if was ok to just return here, will go be unhappy if I don't wait on each result in the channel/exit this function before the go routines complete?

internal/compliance/resources_api/handlers/dynamo.go
```diff
@@ -106,39 +107,94 @@ func doUpdate(update expression.UpdateBuilder, resourceID string) *events.APIGat
...
+				err:       nil,
+			}
+		}(scanInput)
 	}
 
-	if err != nil {
-		zap.L().Error("dynamoClient.QueryPages failed", zap.Error(err))
-		return err
+	// Merge scan results
+	zap.L().Info("scans initiated, awaiting results")
```
kostaspap, line 187: I guess that should be debug

internal/compliance/resources_api/handlers/dynamo.go
```diff
@@ -106,39 +107,94 @@ func doUpdate(update expression.UpdateBuilder, resourceID string) *events.APIGat
...
+}
 
-		for _, entry := range items {
-			if handlerErr = handler(entry); handlerErr != nil {
-				return false // stop paginating
+// Wrapper around dynamoClient.ScanPages that accepts a handler function to process each item.
+func scanPages(inputs []*dynamodb.ScanInput, includeCompliance bool,
+	requiredComplianceStatus compliancemodels.ComplianceStatus) ([]models.Resource, error) {
+
+	results := make(chan scanResult)
```
kostaspap, line 119: I think you are not closing it? You should a 
```
defer close(results)
```
after line 119

deployments/cloud_security.yml
kostaspap, line 1029: What about the compliance table? It seems that each scan of resources requires a lookup in the compliance table. Shouldn't this be managed in the same way as well?

kostaspap (commented):
* Is there a way to avoid the scanning all together? How much effort would that require - what would we lose product-wise? 
* The benchmark says that with 1k RCU and 1024MB lambda it still takes close to 30 seconds (AppSync's max) to scan the table. How often do we scan - how often do users have to wait? So e.g. Does scan happen every time someone goes to the resources page, clicks "Next page" or  types something in the filter?

deployments/cloud_security.yml
austinbyers: Yeah, I would agree - I think we could just apply the same capacity settings to the compliance table for now

deployments/cloud_security.yml
```diff
@@ -1005,6 +1026,56 @@ Resources:
...
+    Condition: ProvisionResourceTableCapacity
+    Type: 'AWS::ApplicationAutoScaling::ScalableTarget'
+    Properties:
+      MaxCapacity: !Ref CloudSecurityMaxWriteCapacity
+      MinCapacity: !Ref CloudSecurityMinWriteCapacity
+      ResourceId: !Sub table/${ResourcesTable}
+      RoleARN: !Sub arn:aws:iam::${AWS::AccountId}:role/aws-service-role/dynamodb.application-autoscaling.amazonaws.com/AWSServiceRoleForApplicationAutoScaling_DynamoDBTable
```
austinbyers, line 1036: We actually already have an IAM role in the [bootstrap stack](https://github.com/panther-labs/panther/blob/master/deployments/bootstrap.yml#L693) for Dynamo autoscaling.

The challenge with a service role is that it may not exist in new accounts. If you're able to remove the service role and still get a working fresh deployment (i.e. the service role is created before this CloudFormation resource), great! In that case, we can remove the IAM role from the bootstrap stack.

Otherwise, for consistent deployments, we may have to stick with our own role

deployments/cloud_security.yml
```diff
@@ -1005,6 +1026,56 @@ Resources:
...
+      ScalableDimension: 'dynamodb:table:WriteCapacityUnits'
+      ServiceNamespace: dynamodb
+
+  ResourcesTableWriteScalingPolicy:
+    Condition: ProvisionResourceTableCapacity
+    Type: 'AWS::ApplicationAutoScaling::ScalingPolicy'
```
austinbyers, line 1042: Nit: don't need quotes on any of these types

internal/compliance/resources_api/handlers/dynamo.go
austinbyers: > (unless it's thread safe to append to the same slice?)

Definitely not

internal/compliance/resources_api/handlers/dynamo.go
austinbyers: If you close the channel (as Kostas suggested), then an early function return will lead to a panic for routines which are still alive. This just came up with the `mage` packaging code.

You could use an [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) to handle errors and cancellation for you, you can explicitly cancel the other routines using context, or you can just wait for them all, like you do here

austinbyers (commented):
> The provisioning settings don't work for mage deploy, not sure if I still need to add support for this deployment method if we're planning to remove it soon.

If by 'don't work', you mean the new settings aren't in the `panther_config.yml` settings file, that's fine.

But I think `mage deploy` actually won't work at all right now because you have required stack parameters with no values passed in. So you may have to add the `mage deploy` support - we won't be able to fully migrate that model for a few weeks

> Is there a way to avoid the scanning all together? How much effort would that require - what would we lose product-wise?

I agree this needs to happen as soon as we can - doing full table scans like this is not sustainable. But that's a much larger change, requiring new indexes or new table designs completely. This PR is a quick fix to unblock some customers - @nhakmiller did you say you had an epic tracking the longer-term fixes?

>  Does scan happen every time someone goes to the resources page, clicks "Next page" or types something in the filter?

Yes, unfortunately, the scan is triggered on every request. Every page, every new filter. Since the whole table fits in Lambda's memory, we could potentially cache it for 30-60 seconds to handle the common case of users scrolling through pages or changing the filter settings a few times in a row. @nhakmiller is that reasonable / feasible to add to this?

internal/compliance/resources_api/handlers/dynamo.go
nhakmiller: I'll just wait for them all, seems most straightforward.

deployments/cloud_security.yml
nhakmiller: Sure, sounds good.

internal/compliance/resources_api/handlers/dynamo.go
kostaspap: > If you close the channel (as Kostas suggested), then an early function return will lead to a panic for routines which are still alive. This just came up with the mage packaging code.
You could use an errgroup to handle errors and cancellation for you, you can explicitly cancel the other routines using context, or you can just wait for them all, like you do here

I think there is some confusion on this:
By putting the close in a defer statement, you make sure that the channel is closed after the function has terminated and the allocated channel will be released. An issue (panic) can happen if the channel has been closed but there are still go routines running in the background. This can happen if the function has returned but there are go routines still running.

This can be bad because Lambda, when it finishes an invocation, it will just freeze the Go runtime and all the go routines that were running in the background - it doesn't "terminate" them. The go routines will stay alive, the channel and its contents will not be garbage collected. On next invocations, the go routines will continue right from the point they were when frozen so you might see some weird behavior e.g. Lambda starting invocation with full memory or even logging from thee goroutines that were generated during the previous invocation (I've had the "pleasure" of debugging such an issue few months ago).

If you wait here for all go routines to finish, you will not see such an issue. 
I think it is actually better to panic (the FE will get 5xx) in case you have dangling go routines running outside the main program's control than just let them be in the background.

nhakmiller:
I did the following:

- Addressed PR feedback
- Tested both `mage deploy` and `mage master:deploy`
- Added a configurable memory parameter for the `resources-api` and `compliance-api` (for some reason I thought it was already configurable before)

Ready for re-review I think.

nhakmiller:
> * Is there a way to avoid the scanning all together? How much effort would that require - what would we lose product-wise?
> * The benchmark says that with 1k RCU and 1024MB lambda it still takes close to 30 seconds (AppSync's max) to scan the table. How often do we scan - how often do users have to wait? So e.g. Does scan happen every time someone goes to the resources page, clicks "Next page" or  types something in the filter?

Unfortunately we don't have a sort key or anything, so scanning the whole table is the only option (unless we just wanted to return random pages of resources without any filtering or sorting).

Scans happen based on user activity, either uploading a new policy (the page doesn't block for that though) or going to the resources page. I think the 30 seconds should be more rare in conjunction with the other change the reduces the TTL for deleted resources, both the environments we saw issues with had very high churn (hundreds to thousands of resources created per day) so those stale entries just sat in the resources table for 30 days before.

> Yes, unfortunately, the scan is triggered on every request. Every page, every new filter. Since the whole table fits in Lambda's memory, we could potentially cache it for 30-60 seconds to handle the common case of users scrolling through pages or changing the filter settings a few times in a row. @nhakmiller is that reasonable / feasible to add to this?

Seems reasonable, I would even say like a five minute cache if fine. If you have 450,000 resources you likely are not going to be noticing one in particular is showing up a few minutes late.

deployments/root.yml
```diff
@@ -40,6 +40,31 @@ Parameters:
     # Example: "arn:aws:acm:us-west-2:111122223333:certificate/6e51b91b-0d7d-4592-89a3-c113c78e3ab3" or
     # "arn:aws:iam::111122223333:server-certificate/panther/us-west-2/MyCert"
     AllowedPattern: '^((arn:(aws|aws-cn|aws-us-gov):acm:[a-z]{2}-[a-z]{4,9}-[1-9]:\d{12}:certificate\/[a-f0-9-]{36})|(arn:(aws|aws-cn|aws-us-gov):iam::\d{12}:server-certificate\/\S+))?$'
+  CloudSecurityMaxReadCapacity:
+    Type: Number
+    Description: Maximum read capacity units to provision for the Cloud Security tables
+    Default: 0
```
kostaspap, line 46: I actually don't think these are valid numbers for DDB? Min should be 1?

internal/compliance/resources_api/handlers/dynamo.go
```diff
@@ -106,39 +107,95 @@ func doUpdate(update expression.UpdateBuilder, resourceID string) *events.APIGat
...
+				}
+
+				for _, entry := range items {
+					if !includeCompliance {
+						segmentResources = append(segmentResources, entry.Resource(""))
+						return true
+					}
+
+					var status *complianceStatus
+					status, handlerErr = getComplianceStatus(entry.ID)
```
kostaspap, line 150: Should you also use the same RCU/WCU defined for the resources for compliance table as well? It seems even if the scan is quick, getting the compliance status could slow the whole thing down?

kostaspap (approved):
Changes look good, scanning in parallel and pumping up the memory seem like they would help a lot already! Approving because I think these changes will help a lot. 

The WCU/RCU are a bit more tricky to get right, I guess it will require some tuning depending on the deployment.

deployments/root.yml
nhakmiller: We use 0 as the "disable provisioned capacity" number. If 0 is set, we use on demand billing instead (the default).

deployments/cloud_security.yml
```diff
@@ -991,20 +1080,78 @@ Resources:
...
+    Condition: ProvisionResourceTableCapacity
+    Type: AWS::ApplicationAutoScaling::ScalingPolicy
+    Properties:
+      PolicyName: ReadAutoScalingPolicy
+      PolicyType: TargetTrackingScaling
+      ScalingTargetId: !Ref ResourcesTableReadCapacityScalableTarget
+      TargetTrackingScalingPolicyConfiguration:
+        TargetValue: 70
+        ScaleInCooldown: 60
+        ScaleOutCooldown: 60
```
kostaspap, line 1151: Scans are performed by ad-hock queries which are not that frequent. Should we set the ScaleInCooldown period longer e.g. 30/60 minutes or something? I am thinking the following scenario.... 
- User tries to access UI - it fails due to timeout. 
- Retries (now DDB has scaled up) succeeds. 
- 5' later they try to access it, fails again. 

Increasing the ScalelnCooldown could lead to increased costs even if there is no activity, but I am thinking that's a hit we could take on in order to give a better UI experience?

austinbyers: Seems reasonable to me, bumping the cooldown a bit longer

austinbyers (approved):

nhakmiller:
I set the scale in cool down to 5 minutes per PR feedback, but I really need to focus on some other things this sprint and I really want this in for 1.16 so I'm not going to do the in memory cache yet. Perhaps for a future iteration we can add this.

PR #2625, opened 2021-02-07
Use smaller s3 retries for reading S3 objects in log processor AFTER failures

rleighton:
## Background

Note on Slack:

Hey guys, we’re observing some very unusual activity in a hosted account and I was hoping one of our data processing experts could weigh in on it. The `some account` account cost us $`more than we expected`  the week of January 10th. They were processing a very steady ~22 GB  / day according to our metrics. Almost the entirety of the cost is coming from the panther-log-processor lambda function, which ran for a stunning 4,296,272.5 seconds / day. By contrast, the panther-log-processor in the `another account`  account (who are processing 600 GB / day) only ran for 21,188.465 seconds / day. There is clearly something very wrong here. 

The issue was that the account in question had misconfigured access to an S3 bucket and the log processor was spending all its time in a retry loop. The condition existed for many days unresolved. 

## Changes

- Added a "tracker" to track failed prefixes, then if we ask for a client with a prefix with a history of failures we use a client with a short timeout.

## Testing

- mage test:ci


rleighton:
I plan to load test this change ... do a before and after comparison.

kostaspap (commented):
In order to address the issue that came up, I believe the change is much simpler: We should use the the `awsretry.NewConnectionErrRetryer` here https://github.com/panther-labs/panther/blob/master/internal/log_analysis/log_processor/common/common.go#L82

We had introduce the AuthRetrier that would retry in case of Auth issues - back in those days, failure to read one S3 object would mean the whole batch would be retried so we wanted to retry very hard. 

This is no longer the case in the code though, Auth failure reading one object will not cause the whole batch to be retried so we can remove the Auth retrier.

rleighton:
Closing in favor of https://github.com/panther-labs/panther/pull/2627

Issue #2354, opened 2021-01-07
Better management of auxiliary CFN templates


austinbyers:
### Description

Every release, Panther publishes a new version of the `deployments/auxiliary/cloudformation` templates to a public S3 bucket. These templates are downloaded directly when onboarding new sources. There are two main problems with this setup:

1. When developers change the aux templates and want to test source onboarding, their only recourse is to publish a new version (see #2338) In other words, **local changes to the aux templates are not reflected when deploying from source**. This would be a problem for any community member who wanted to change them in their own deployments.
2. The aux templates are published from a Makefile, unlike the rest of the Panther assets which are published by mage. This means we have to run two separate commands during every release publish

### Related Services

`mage`, CloudFormation, source-api, and the source onboarding page in the frontend

### Acceptance Criteria

- When deploying Panther from source, the *local* auxiliary templates are used rather than the official release templates published to our public S3 bucket
- The `mage publish` command should also publish the aux templates, and the Makefile in the aux directory can then be removed

### Implementation

Deployments from the source code have a "source" bucket which stages the Lambda binaries. This same bucket can be used to store a copy of the aux templates. So the deploy process becomes:

1. `mage` uploads a copy of the aux templates as part of the packaging of Panther assets
2. When the CloudFormation stacks are deployed, the locations of the aux templates are passed as parameters to the relevant services (instead of hardcoding the panther public bucket URL)

kostaspap:
@austinbyers these templates need to be accessible publicly... When a user starts the process of onboarding any new account, that account needs to be able to access that template.

austinbyers:
> @austinbyers these templates need to be accessible publicly... When a user starts the process of onboarding any new account, that account needs to be able to access the template.

Yes, for releases the aux templates will still be published to our public S3 bucket, as before.

But it seems absurd that the only way *developers* can test is by publishing to our production public bucket. Surely we can come up with something better?

Ideally, these templates could be uploaded to a private dev bucket, and then we use IAM to grant access to other Panther backend services and/or other testing accounts as necessary. For example, you could authorize a specific organization (e.g. the Panther dev AWS org) to read those objects, which is exactly how the enterprise public bucket works anyway.

[...]

PR #2626, opened 2021-02-07
Improve template error message

rleighton:
## Before

Template error messages were not useful ... in particular if the Panther VERSION is ahead of public templates in dev , the UI fails and it is not clear why.

## After

Show the s3 path being attempted and point to VERSION file to aid debugging.

## Testing

- mage deploy


internal/core/source_api/api/get_integration_template.go
```diff
@@ -130,7 +131,7 @@ func (api *API) getTemplate(integrationType string) (string, error) {
 	zap.L().Debug("requesting template", zap.String("key", *templateRequest.Key), zap.String("bucket", *templateRequest.Bucket))
 	s3Object, err := api.TemplateS3Client.GetObject(templateRequest)
 	if err != nil {
-		return "", err
+		return "", errors.Wrapf(err, "cannot read template %#v, check Panther VERSION file", templateRequest)
```
austinbyers, line 134: If this error is bubbled up to the user and shown in the web app, it might be too verbose with a Go struct expanded in there. Really all we care about is the object key, which will have the expected version included in it.

To be even more precise, you could check if the error is caused by "NotFound". In that case, the error can be something like "template %s not found, aux templates have not been published for the current VERSION file"

austinbyers (approved):
I hate how our aux templates are setup 😓  every time the VERSION changes, dev deployments will break until we publish the new template versions.

#2354 is tracking the long-term fix, thanks for the more helpful error in the meantime

PR #2627, opened 2021-02-08
Remove auth retrier for s3 in log processor

rleighton:
## Background

Note on Slack:

Hey guys, we’re observing some very unusual activity in a hosted account and I was hoping one of our data processing experts could weigh in on it. The `some account` account cost us $`more than we expected`  the week of January 10th. They were processing a very steady ~22 GB  / day according to our metrics. Almost the entirety of the cost is coming from the panther-log-processor lambda function, which ran for a stunning 4,296,272.5 seconds / day. By contrast, the panther-log-processor in the `another account`  account (who are processing 600 GB / day) only ran for 21,188.465 seconds / day. There is clearly something very wrong here. 

The issue was that the account in question had misconfigured access to an S3 bucket and the log processor was spending all its time in a retry loop. The condition existed for many days unresolved. 

This is repeated in a test environment. Data (CloudtTrail and S3 access logs) was created using `filegen` utility and the bucket holding the test data onboarded. The `s3queue` utility was used to drive load of 1 file per second into the system.

As you can see below, a single lambda can handled this load easily:

image

Now, we remove the permissions to read the bucket, this is what happens under the same load (collapse):

image

## Changes

Changed the out the `AccessDeniedRetryer` for the `ConnectionErrRetryer` for reading S3 in the log processor.

The reason we had the `AccessDeniedRetryer` is that we wanted to try hard during normal (for AWS) occasional `Access Denied` errors we see in production. We wanted to try hard because, at the time, if we had 1 failure reading the whole SQS batch of messages would be discarded and retried. 

We have since change the log processor to only discard individual messages that fail. By removing the `AccessDeniedRetryer` we will fail fast and that one file will get retried after the SQS visibility timeout.

After this change, this is what the log processor looked like under the same load where there was no access to the bucket:

image

## Testing

- mage test:ci
-  load test
-


kostaspap (approved):

alxarch:
Are we certain that we only discard individual messages?
Reading the code in `stream.go` seems to me is doing the following:
1. Discard all messages for which it couldn't read _the first chunk_.
2. Process those messages in sequence.
3. If the processing fails for some reason (i.e. a chunk failed to be read, some other misconfiguration resulted in error during processing), we discard all messages
4. If all went well, we acknowledge all messages by deleting them from the queue using the receipt.

This still leaves room for errors that 'poison' the full batch. If 403 errors are as sporadic as it seems it is quite possible this can happen after reading the first chunk successfully.

kostaspap:
> Are we certain that we only discard individual messages?
> Reading the code in `stream.go` seems to me is doing the following:
> 
> 1. Discard all messages for which it couldn't read _the first chunk_.
> 2. Process those messages in sequence.
> 3. If the processing fails for some reason (i.e. a chunk failed to be read, some other misconfiguration resulted in error during processing), we discard all messages
> 4. If all went well, we acknowledge all messages by deleting them from the queue using the receipt.
> 
> This still leaves room for errors that 'poison' the full batch. If 403 errors are as sporadic as it seems it is quite possible this can happen after reading the first chunk successfully.

This PR fixes a very specific issues that appears pretty often - users sending events to Panther but they haven't given panther permissions to read them. It doesn't address other failure scenarios true. That would require big refactoring of the logic.

rleighton:
> > Are we certain that we only discard individual messages?
> > Reading the code in `stream.go` seems to me is doing the following:
> > 
> > 1. Discard all messages for which it couldn't read _the first chunk_.
> > 2. Process those messages in sequence.
> > 3. If the processing fails for some reason (i.e. a chunk failed to be read, some other misconfiguration resulted in error during processing), we discard all messages
> > 4. If all went well, we acknowledge all messages by deleting them from the queue using the receipt.
> 
> This PR fixes a very specific issues that appears pretty often - users sending events to Panther but they haven't given panther permissions to read them. It doesn't address other failure scenarios true. That would require big refactoring of the logic.
> 
> > This still leaves room for errors that 'poison' the full batch. If 403 errors are as sporadic as it seems it is quite possible this can happen after reading the first chunk successfully.
> 
> That's actually a good point. Maybe we should still retry Auth errors but just limited number of retries (e.g. the default 3)

I played with that ... still not nice behavior ... that's why I tried "if you failed before, no retries" in that other PR but still they retries when there is load is not happy. We get retries via visibility timeout.

PR #2628, opened 2021-02-08
Support for Unit Test Mocking - Log Analysis

wey-chiang:
## Background

This PR adds support for unit test mocking in Panther based on [this RFC](https://github.com/panther-labs/panther-enterprise/tree/master/designs/security/mocking-support).

## Changes

- Updated analysis-api UnitTest for Mock (no schema changes to DDB)
- Updated rules engine to properly unpack and setup mock patching
- Added naive unit test for mocking support

## Testing

- `mage test:ci`
- Tested by manually invoking the lambda


internal/core/analysis_api/analysis/rule_engine.go
```diff
@@ -56,9 +56,21 @@ func (e *RuleEngine) TestRule(rule *models.TestRuleInput) (*models.TestRuleOutpu
 			return nil, &TestInputError{fmt.Errorf(`Event for test "%s" is not valid json: %w`, test.Name, err)}
 		}
 
+		var mock map[string]interface{}
+		// Added check for backwards compatibility
+		if test.Mock != "" {
+			if err := jsoniter.UnmarshalFromString(test.Mock, &mock); err != nil {
+				//nolint // Error is capitalized because will be returned to the UI
```
nhakmiller, line 63: Is this comment still relevant?

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -51,13 +51,17 @@ def analyze_single_rule(self, raw_rule: dict, test_spec: Mapping) -> Dict[str, A
         raw_rule['versionId'] = 'default'
         rule = Rule(raw_rule)
         event = test_spec['data']
+        if 'mocks' not in test_spec:
+            event_mocks = dict()
+        else:
+            event_mocks = test_spec['mocks']
```
nhakmiller, line 57: Could this be condensed

```suggestion
        event_mocks = test_spec.get('mocks', dict()):
```

api/lambda/analysis/models/api.go
```diff
@@ -69,6 +69,7 @@ type LambdaInput struct {
 
 type UnitTest struct {
 	ExpectedResult bool   `json:"expectedResult"`
+	Mock           string `json:"mock"`
```
lindsey-w, line 72: This can represent multiple mock key/values, right? any reason to call them `mock` here and `mocks` in Event struct?

internal/log_analysis/rules_engine/tests/test_rule.py
```diff
@@ -364,3 +364,30 @@ def test_rule_with_valid_severity_case_insensitive(self) -> None:
         )
         result = rule.run(PantherEvent({}, None))
         self.assertEqual(expected_result, result)
+
+    def test_rule_with_mocking(self) -> None:
```
lindsey-w, line 368: wondering if there could also be a test that "calls" the boto3 library?

lindsey-w (commented):
This is going to be so useful 🙌🏻

internal/core/analysis_api/analysis/rule_engine.go
wey-chiang: Testing changes that would make `test.Mock` a map rather than a string parsed as a map -- those changes would simplify this logic and remove the comment

kostaspap:
Are there any thoughts about sharing code between Panther rules engine and PAT? E.g. if PAT would reuse the `Rules` class defined here and would support mocks. Also if PAT and Rules Engine share code it minimizes the chances of inconsistencies e.g. unit tests passing in PAT but failing in Panther or vv.

internal/log_analysis/rules_engine/tests/test_rule.py
wey-chiang: Sure, I can add another unit test here

api/lambda/analysis/models/api.go
wey-chiang: `Mock` is a bit misleading... I'll change it to `Mocks` and make changes to ensure consistency

api/lambda/analysis/engine.go
```diff
@@ -74,8 +74,9 @@ type Rule struct {
 
 // Event is a security log to be analyzed, e.g. a  CloudTrail event.
 type Event struct {
-	Data interface{} `json:"data"`
-	ID   string      `json:"id"`
+	Data  interface{} `json:"data"`
+	ID    string      `json:"id"`
+	Mocks interface{} `json:"mocks"`
```
kostaspap, line 79: Is it meant to be a JSON? Or in what form? Using `interface{}` is not ideal, it could mean a `true` , `1`, `"test"` or `{"key":"value"}`. If the input is meant to be JSON, use `map[string]interface{}`

kostaspap (commented):
Can you provide more information about the E2E testing? What was the payload like? Which Lambda did you invoke, `panther-analysis-api` or `panther-rules-engine`?  What response did you get?

api/lambda/analysis/models/api.go
```diff
@@ -68,9 +68,10 @@ type LambdaInput struct {
 }
 
 type UnitTest struct {
-	ExpectedResult bool   `json:"expectedResult"`
-	Name           string `json:"name" validate:"required"`
-	Resource       string `json:"resource" validate:"required"`
+	ExpectedResult bool        `json:"expectedResult"`
+	Mocks          interface{} `json:"mocks"`
```
kostaspap, line 72: See comment above. Using `interface{}` is not ideal.

georgepsarakis (commented):
Nice work and great feature! I have an alternative suggestion that I believe could work without the need for patching. In addition, it would be really difficult to account for more dynamic mocking (e.g. function calls, fake objects, exceptions). One possibility here, would be to have a small dependency stubbing framework, that does almost what patching does, but in a more controlled (but still powerful) manner. This way the fakes / mocks can be bundled with the rule functions (or even in global modules for that matter...). Just a quick working prototype of the API to give you an overview:

<details>
<summary>Stubbed Dependencies Prototype</summary>

```python
from contextlib import contextmanager
import boto3

class TestingModeHandler:
    def __init__(self):
        self._testing_mode = False
    
    @property
    def enabled(self):
        return self._testing_mode

    def enable(self):
        self._testing_mode = True

    def disable(self):
        self._testing_mode = False

testing_mode = TestingModeHandler()

@contextmanager
def stubbed_dependencies():
    testing_mode.enable()
    try:
        yield
    finally:
        testing_mode.disable()
 
def dependency(stub=None):
  def wrapper(fn):
     def inner_wrapper(*args, **kwargs):
       if testing_mode.enabled:
         if callable(stub):
              result = stub(*args, **kwargs)
              if isinstance(result, Exception):
                  raise result
              else:
                  return result
         else:
              return stub  
       else:
         return fn(*args, **kwargs)  
     return inner_wrapper
  return wrapper

# Rule module starts here
class FakeS3Client:
    pass

@dependency(stub=FakeS3Client)
def create_s3_client():
    return boto3.client("s3")

def rule(event):
  print(repr(create_s3_client()))

    
if __name__ == "__main__":
    rule({})
    with stubbed_dependencies():
        rule({})
    rule({})
```

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -51,13 +51,14 @@ def analyze_single_rule(self, raw_rule: dict, test_spec: Mapping) -> Dict[str, A
...
         event = test_spec['data']
+        event_mocks = test_spec.get('mocks', dict())
         log_type = event.get('p_log_type', 'default')
 
         # enrich the event to have access to field by standard field name
         #  via the `udm` method
         event = PantherEvent(event, self.log_type_to_data_models.get(log_type))
 
-        rule_result = rule.run(event, batch_mode=False)
+        rule_result = rule.run(event, event_mocks, batch_mode=False)
```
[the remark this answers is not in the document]
austinbyers: But optional arguments have to be at the end in Python, and this is not an optional argument

internal/log_analysis/rules_engine/tests/test_rule.py
```diff
@@ -364,3 +364,36 @@ def test_rule_with_valid_severity_case_insensitive(self) -> None:
...
+        rule_body = 'import boto3\n' \
```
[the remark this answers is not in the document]
austinbyers: You could also implement it as a list of strings, and then join it with newlines

austinbyers (commented):
This is only for the rules engine, which I know is the higher priority for mocking. The RFC does mention both policies and rules - will policy mocking come in a separate PR? If so, I'd update the PR title to focus just on log analysis mocking

nhakmiller:
> Nice work and great feature! I have an alternative suggestion that I believe could work without the need for patching. In addition, it would be really difficult to account for more dynamic mocking (e.g. function calls, fake objects, exceptions). One possibility here, would be to have a small dependency stubbing framework, that does almost what patching does, but in a more controlled (but still powerful) manner. This way the fakes / mocks can be bundled with the rule functions (or even in global modules for that matter...). Just a quick working prototype of the API to give you an overview:
> 
> Stubbed Dependencies Prototype
> ```python
> from contextlib import contextmanager
> import boto3
> 
> class TestingModeHandler:
>     def __init__(self):
>         self._testing_mode = False
>     
>     @property
>     def enabled(self):
>         return self._testing_mode
> 
>     def enable(self):
>         self._testing_mode = True
> 
>     def disable(self):
>         self._testing_mode = False
> 
> testing_mode = TestingModeHandler()
> 
> @contextmanager
> def stubbed_dependencies():
>     testing_mode.enable()
>     try:
>         yield
>     finally:
>         testing_mode.disable()
>  
> def dependency(stub=None):
>   def wrapper(fn):
>      def inner_wrapper(*args, **kwargs):
>        if testing_mode.enabled:
>          if callable(stub):
>               result = stub(*args, **kwargs)
>               if isinstance(result, Exception):
>                   raise result
>               else:
>                   return result
>          else:
>               return stub  
>        else:
>          return fn(*args, **kwargs)  
>      return inner_wrapper
>   return wrapper
> 
> # Rule module starts here
> class FakeS3Client:
>     pass
> 
> @dependency(stub=FakeS3Client)
> def create_s3_client():
>     return boto3.client("s3")
> 
> def rule(event):
>   print(repr(create_s3_client()))
> 
>     
> if __name__ == "__main__":
>     rule({})
>     with stubbed_dependencies():
>         rule({})
>     rule({})
> ```

Ideally customers won't need to write fancy python code like this to make use of unit test mocking.

wey-chiang:
> Can you provide more information about the E2E testing? 

Sure! 

> What was the payload like? 

I created a test event in AWS Lambda, defined as:

```
{
  "testRule": {
    "body": "import boto3\nfrom boto3 import client\nfrom unittest.mock import MagicMock\n\ndef rule(event):\n\treturn isinstance(boto3, MagicMock) and isinstance(client, MagicMock) and isinstance(boto3.client, MagicMock)\n\ndef title(event):\n\t# (Optional) Return a string which will be shown as the alert title.\n\t# If no 'dedup' function is defined, the return value of this method will act as deduplication string.\n\treturn f\"BOTO3 - {isinstance(boto3, MagicMock)} - CLIENT - {isinstance(client, MagicMock)} - BOTO3.CLIENT - {isinstance(boto3.client, MagicMock)}\"\n\t\ndef description(event):\n   return \"THIS IS A GENERATED DESCRIPTION!\"\n\ndef destinations(event):\n   return [\"Weyland-Email-Test\"]\n\ndef reference(event):\n   return \"THIS IS A GENERATED REFERENCE!\"\n\ndef runbook(event):\n   return \"THIS IS A GENERATED RUNBOOK!\"\n\ndef severity(event):\n   return \"CrItICaL\"\n\n\n# def dedup(event):\n\t#  (Optional) Return a string which will be used to deduplicate similar alerts.\n\t# return ''\n\n# def alert_context(event):\n\t#  (Optional) Return a dictionary with additional data to be included in the alert sent to the SNS/SQS/Webhook destination\n\t# return [\"Weyland-Email-Test\"]",
    "logTypes": [
      "AWS.CloudTrail"
    ],
    "tests": [
      {
        "expectedResult": true,
        "mocks": {
          "boto3": "example_boto3_return_value!",
          "client": "example_client_return_value!"
        },
        "name": "test1",
        "resource": "{}"
      }
    ]
  }
}
```

> Which Lambda did you invoke, panther-analysis-api or panther-rules-engine? 

I invoked the `panther-analysis-api` lambda function. Note how the rule and title functions are defined.

> What response did you get?

I've included the relevant excerpts of the response that indicate `boto3` and `client` are properly patched:

```
"test result": {
  "results": [
    {
      "id": "0",
      "name": "test1",
      "passed": true,
      "functions": {
        "ruleFunction": {
          "output": "true",
          "error": null
        },
        "titleFunction": {
          "output": "BOTO3 - True - CLIENT - True - BOTO3.CLIENT - True",
          "error": null
        },
      ...
      },
      "error": null
    }
  ]
}
```

wey-chiang:
> Also, is it possible it breaks the log processing? 

It should not -- the code should only run for unit tests and not normal log processing.

Normal log processing will invoke the `log_analysis` function in `main.py` -- following the call stack: `log_analysis` will invoke the `_RULES_ENGINE.analyze` function which subsequently invokes the `rule.run` function without the `event mocks` parameter.

The `direct_analysis` function (which is meant for **testing** a single rule against a set of events) invokes the `_RULES_ENGINE.analyze_single_rule` function which will subsequently invoke the `rule.run` function with the `event_mocks` parameter.

During testing, if I had improperly defined the mock data in DDB the FE would display the following issue:
image

> Did you verify that after deploying the rules engine was running properly and analyzing incoming stream of data?

It has been about a day since deploying and successfully testing the new functionality and I have not observed any errors yet.

For reference, I used the following CloudWatch query over the past 24 hours for the `analysis-api` and `rules-engine`:

```
fields @timestamp, @message
| filter @message like /(?i)(error|critical|high|warn)/
| sort @timestamp desc
| limit 200
```

api/lambda/analysis/engine.go
wey-chiang: I was unsure about this part, thanks! 🙂

internal/log_analysis/rules_engine/src/engine.py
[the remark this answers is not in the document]
wey-chiang: Actually this is an optional argument the function will default to `None` when it is not defined.

Will modify the signature and the calls to it to have `event_mocks` be at the end of the parameter list

internal/log_analysis/rules_engine/tests/test_rule.py
```diff
@@ -364,3 +364,36 @@ def test_rule_with_valid_severity_case_insensitive(self) -> None:
...
+                    's3_client = client("s3")\n\t' \
+                    'assert s3_client == "client_return_value"\n\t' \
+                    'client.assert_called_once_with("s3")\n\t' \
+                    'return True\n' \
+                    'def alert_context(event):\n\treturn {}\n' \
+                    'def title(event):\n\treturn "test_rule_with_mocking"\n'
+        rule = Rule({'id': 'test_rule_with_mocking', 'body': rule_body, 'versionId': 'versionId'})
+
+        mocks = {
+            'boto3': 'boto3_return_value',
```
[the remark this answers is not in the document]
wey-chiang: > I think that we should be patching boto3.client

Yes and no, I wanted to test both cases of imports, i.e.

```
import boto3
from boto3 import client
boto3.client("s3")
client("s3")
```

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -209,12 +211,24 @@ def run(self, event: PantherEvent, batch_mode: bool = True) -> RuleResult:
             rule_result.setup_exception = self._setup_exception
             return rule_result
 
+        # Setup mock functions
+        mock_methods = dict()
```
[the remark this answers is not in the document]
wey-chiang: > An example failure scenario is ...

This will never happen -- each of the user defined functions are individually run in their own respective try/except blocks that catch broad exceptions.

wey-chiang:
> > Nice work and great feature! I have an alternative suggestion that I believe could work without the need for patching. In addition, it would be really difficult to account for more dynamic mocking (e.g. function calls, fake objects, exceptions). One possibility here, would be to have a small dependency stubbing framework, that does almost what patching does, but in a more controlled (but still powerful) manner. This way the fakes / mocks can be bundled with the rule functions (or even in global modules for that matter...). Just a quick working prototype of the API to give you an overview:
> > ```python
> > # Rule module starts here
> > class FakeS3Client:
> >     pass
> > 
> > @dependency(stub=FakeS3Client)
> > def create_s3_client():
> >     return boto3.client("s3")
> > 
> > def rule(event):
> >   print(repr(create_s3_client()))
> > ```
> 
> Ideally customers won't need to write fancy python code like this to make use of unit test mocking.

The idea here is to simplify the UX as much as possible. 

There is an inherent tradeoff between giving more control to the users to allow them to add more complexity to the unit tests and the complexity/cognitive-overhead of writing rules. 

This approach is designed to provide as much simplicity to the users as possible.

wey-chiang:
> In addition, it would be really difficult to account for more dynamic mocking (e.g. function calls, fake objects, exceptions).

I am not totally convinced this statement holds true.

> This way the fakes / mocks can be bundled with the rule functions (or even in global modules for that matter...). 

Mocking support here is solely for unit testing -- namely testing user-defined events in the UI. 

In terms of accounting for dynamic mocking: the approach proposed in this PR greatly simplifies that accounting as the mocking is guaranteed to only run for unit tests and is similarly guaranteed to always clean up the patching properly.

--

The proposed alternative inherently creates more complexity as we would be (literally) moving the mocking logic out of the rules engine and housing it in something the rules engine runs as opposed to functionality built into the rules engine.

wey-chiang:
I've updated the struct definitions to have `Event.Mocks` to be a `map[string]string` and `UnitTest.Mocks` to be a `string`.

For testing, I followed a similar approach outlined [here](https://github.com/panther-labs/panther/pull/2628#issuecomment-777052796)

--

Tested the `panther-analysis-api` lambda function directly using:

```
{
  "testRule": {
    "body": ( exact same as before )
    "logTypes": [
      "AWS.CloudTrail"
    ],
    "tests": [
      {
        "expectedResult": true,
        "mocks": "{\"boto3\":\"example_boto3_return_value\",\"client\":\"example_client_return_value\"}",
        "name": "test1",
        "resource": "{}"
      }
    ]
  }
}
```

Which yielded the following response (parsed and formatted for readability)

```
{
	"results": [
		{
			"id":"0",
			"name":"test1",
			"passed":true,
			"functions": {
				"ruleFunction": {
					"output":"true",
					"error":null
				},
				"titleFunction": {
					"output":"BOTO3 - True - CLIENT - True - BOTO3.CLIENT - True",
					"error":null
				}
			}
		}
	]
}
```

wey-chiang:
Cleaned up the PR based on the feedback

internal/log_analysis/rules_engine/tests/test_rule.py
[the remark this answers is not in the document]
wey-chiang: I've modified the unit test to better convey the original intent

wey-chiang:
> This is only for the rules engine, which I know is the higher priority for mocking. The RFC does mention both policies and rules - will policy mocking come in a separate PR? If so, I'd update the PR title to focus just on log analysis mocking

I will create a separate PR for the policy engine as they are structured a bit differently.

I've updated the title to focus on Log Analysis

kostaspap (commented):
LGTM

internal/log_analysis/rules_engine/src/rule.py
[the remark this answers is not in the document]
georgepsarakis: I understand. We could however have an unhandled exception in our code also 😄 These benefits still apply though:

> I believe we don't have to modify this method at all. You can use the context manager version of patch.multiple in analyze_single_rule instead. For multiple patches, the context manager also provides us with a dictionary of the assigned mocks, so that you can also perform assertions afterwards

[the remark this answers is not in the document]
kostaspap: As @georgepsarakis mentioned, we want to mock things only when the `analysis_single_rule` is invoked so putting a context manager around that seems more appropriate than putting the logic inside the `run` method of the rule. Seems like a better separation of concerns...and Python makes easy for us to do it using decorators. 
It looks like a very small code change and it will make the code more manageable long term. We already have a lot of `if/else` in the code to handle just the UI testing cases, seems that the contextmanager approach would help us abstract that even further in the future and keep in the rules engine just the core logic.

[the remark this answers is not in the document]
wey-chiang: > I understand. We could however have an unhandled exception in our code also 😄 These benefits still apply though I think:
> 
> > I believe we don't have to modify this method at all. You can use the context manager version of patch.multiple in analyze_single_rule instead. For multiple patches, the context manager also provides us with a dictionary of the assigned mocks, so that you can also perform assertions afterwards

Absolutely! I've modified the classes such that the mocking logic is moved from `Rule` over to the `Engine` class, thus isolating the mocking logic to `analyze_single_rule` and refactored the unit test to test that function.

wey-chiang:
Based on the feedback from @georgepsarakis I've moved the mocking logic to `Engine:analyze_single_rule` so the mocking logic is isolated to the function that is responsible for unit testing rather than packing that logic into `Rule:run`.

Deployed and tested in `us-east-1` with the same procedures (directly testing the `panther-analysis-api` lambda) which yielded the [same results as before](https://github.com/panther-labs/panther/pull/2628#issuecomment-777139130).

nhakmiller (approved):
LGTM, CI is failing on some formatting/linting issues, will need to clean those up before it will pass.

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -19,6 +19,7 @@
 from datetime import datetime, timedelta
 from timeit import default_timer
 from typing import Any, Dict, List
+from unittest.mock import patch, MagicMock
```
georgepsarakis, line 22: @wey-chiang following up on our discussion, in case we get a circular import error at some point in the future (tests import `unittest.mock` and importing this module too would result in an import cycle), you can just move these imports inside `analyze_single_rule` 😉 .

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -51,13 +52,24 @@ def analyze_single_rule(self, raw_rule: dict, test_spec: Mapping) -> Dict[str, A
...
         event = PantherEvent(event, self.log_type_to_data_models.get(log_type))
 
-        rule_result = rule.run(event, batch_mode=False)
+        # Setup mock functions
+        mock_methods = dict()
+        if event_mocks:
+            mock_methods = {k: MagicMock(return_value=v) for k, v in event_mocks.items()}
+
+        if mock_methods:
+            with patch.multiple(rule.module, **mock_methods):
```
georgepsarakis, line 68: Glad it worked! 🥳

internal/log_analysis/rules_engine/src/engine.py
georgepsarakis: Never mind, this is not a real possibility. Circular imports are caused by modules depending on each other, not importing the same common module ... Sorry for that!

PR #2629, opened 2021-02-08
Simplify SchemaDatabase interface

### LINKED ISSUE #2650: Finalize the managed schemas API

alxarch:
## Background

Closes #2650 

This PR simplifies the code in the LogTypesAPI by handling user-defined and managed schemas with the same methods when storing in DynamoDB.

Note that we cannot avoid using transactions if we want to be able to have fine-grained error messages on condition check failures.

## Changes
- We no longer keep revision history for user defined schemas. Revision history has no use case in the current way we use schemas. It complicates the DDB code by a lot and it was implemented in a clever bug non-obvious way. And it's no more.
- SchemaDatabase interface is reduced to 3 methods `GetSchema`, `PutSchema`, `ScanSchemas`. These operations can be composed by API handlers to provide all required endpoints.
- The integrity of the Database is preserved with the Revision locks. Endpoints that modify a record, first retrieve it, check the validity of the modification and then write to the db checking that the revision was not incremented between get and put.

## Testing

- mage test:go
- e2e test (underway)


internal/core/logtypesapi/api.go
```diff
@@ -47,21 +47,10 @@ type LogTypesAPI struct {
 
 // SchemaDatabase handles the external actions required for LogTypesAPI to be implemented
 type SchemaDatabase interface {
-	// Create a new user schema record
-	CreateUserSchema(ctx context.Context, id string, upd SchemaUpdate) (*SchemaRecord, error)
-
 	// GetSchema gets a single schema record
```
alxarch, line 50: SchemaDatabase now only has 3 methods that operate on SchemaRecord values.
- Get to get a record (managed/user)
- Put (for create/update manged/user)
- Scan (for list, filters etc)

internal/core/logtypesapi/customlogs.go
```diff
@@ -34,12 +35,12 @@ import (
 // GetCustomLog gets a custom log record for the specified id and revision
 func (api *LogTypesAPI) GetCustomLog(ctx context.Context, input *GetCustomLogInput) (*GetCustomLogOutput, error) {
 	id := customlogs.LogType(input.LogType)
-	r, err := api.Database.GetSchema(ctx, id, input.Revision)
+	r, err := api.Database.GetSchema(ctx, id)
```
alxarch, line 38: GetSchema does not care for revisions anymore

internal/core/logtypesapi/customlogs.go
```diff
@@ -173,8 +195,11 @@ type PutCustomLogInput struct {
 	LogType string `json:"logType" validate:"required,startswith=Custom." description:"The log type id"`
 	// Revision is required when updating a custom log record.
 	// If  is omitted a new custom log record will be created.
-	Revision int64 `json:"revision,omitempty" validate:"omitempty,min=1" description:"Custom log record revision to update (if omitted a new record will be created)"`
-	SchemaUpdate
```
alxarch, line 177: Nesting is more confusing than helpful here

internal/core/logtypesapi/customlogs.go
```diff
@@ -194,7 +219,20 @@ func (api *LogTypesAPI) DelCustomLog(ctx context.Context, input *DelCustomLogInp
 	}
 
 	id := customlogs.LogType(input.LogType)
-	if err := api.Database.ToggleSchema(ctx, id, false); err != nil {
```
alxarch, line 197: Toggle is split into Get/Put.
We might have multiple such 'action' endpoints and it would make no sense to write DDB transactions for all of them.
The revision check in the condition of put protects us from overwriting a record that was changed between get and put

internal/core/logtypesapi/dynamodb.go
```diff
@@ -43,16 +43,8 @@ const (
...
-	attrDisabled     = "IsDeleted"
-	attrRevision     = "revision"
-	attrDescription  = "description"
-	attrReferenceURL = "referenceURL"
-	attrSpec         = "logSpec"
-	attrManaged      = "managed"
-
-	recordKindStatus      = "status"
-	attrAvailableLogTypes = "AvailableLogTypes"
+	attrRecordKind = "RecordKind"
```
alxarch, line 46: The simplification of the logic is clearly visible here.
We do not depend on so many consts for attribute names anymore.
We did so previously because we had complex constraints for the transactions.

internal/core/logtypesapi/customlogs.go
```diff
@@ -67,9 +67,20 @@ func (api *LogTypesAPI) PutCustomLog(ctx context.Context, input *PutCustomLogInp
 		return nil, err
 	}
 
+	now := time.Now()
 	switch currentRevision := input.Revision; currentRevision {
```
kostaspap, line 71: Do we care about revisions here? Can we remove this switch statement?

internal/core/logtypesapi/dynamodb.go
```diff
@@ -78,11 +70,15 @@ func (d *DynamoDBSchemas) ScanSchemas(ctx context.Context, scan ScanSchemaFunc)
 		TableName:                 aws.String(d.TableName),
 	}, func(page *dynamodb.ScanOutput, isLast bool) bool {
 		for _, item := range page.Items {
-			record := SchemaRecord{}
+			record := ddbSchemaRecord{}
```
alxarch, line 73: This fixes the duplicates in the custom log listing page.

internal/core/logtypesapi/customlogs.go
```diff
@@ -173,8 +195,11 @@ type PutCustomLogInput struct {
 	LogType string `json:"logType" validate:"required,startswith=Custom." description:"The log type id"`
 	// Revision is required when updating a custom log record.
 	// If  is omitted a new custom log record will be created.
-	Revision int64 `json:"revision,omitempty" validate:"omitempty,min=1" description:"Custom log record revision to update (if omitted a new record will be created)"`
-	SchemaUpdate
+	Revision     int64  `json:"revision,omitempty" validate:"omitempty,min=1" description:"Custom log record revision to update (if omitted a new record will be created)"`
```
kostaspap, line 198: Do we still need this?

internal/core/logtypesapi/dynamodb.go
```diff
@@ -122,327 +117,100 @@ func (d *DynamoDBSchemas) GetSchema(ctx context.Context, id string, revision int
...
-		Managed:      false,
-		UpdatedAt:    now,
-		CreatedAt:    now,
-		Disabled:     false,
-		SchemaUpdate: upd,
-	}
-	tx := createUserSchemaTX(d.TableName, record)
-	input, err := tx.Build()
+func (d *DynamoDBSchemas) PutSchema(ctx context.Context, id string, record *SchemaRecord) (*SchemaRecord, error) {
+	upd, err := buildPutSchemaExpression(record)
```
alxarch, line 121: We prefer `UpdateItemWithContext` instead of transaction.

kostaspap (commented):
I still see references to "Revision". Is there a reason why we can't remove that?

internal/core/logtypesapi/dynamodb.go
```diff
@@ -122,327 +117,100 @@ func (d *DynamoDBSchemas) GetSchema(ctx context.Context, id string, revision int
...
+		RecordID:   schemaRecordID(id),
 		RecordKind: recordKindSchema,
 	}
 }
 
-func schemaRecordID(id string, rev int64) string {
-	if rev > 0 {
-		id = fmt.Sprintf(`%s-%d`, id, rev)
-	}
+func schemaRecordID(id string) string {
```
alxarch, line 209: No more revision history means ids are simple as well

internal/core/logtypesapi/customlogs.go
alxarch: Revisions are paramount, it is explained in the description.

> The integrity of the Database is preserved with the Revision locks. Endpoints that modify a record, first retrieve it, check the validity of the modification and then write to the db checking that the revision was not incremented between get and put.

internal/core/logtypesapi/customlogs.go
alxarch: The key is we don't keep Revision _History_ anymore. Records still have revision counters

kostaspap: The records in the DB sure. But this is the public API of the Lambda, it doesn't seem that callers should care about revisions at all - it seems that revisions should be a concept present only in the DDB layer

kostaspap: To be clear, I'm fine to leave this in for this PR. I'm just thinking that eventually we'd like to remove it completely.

alxarch: The integrity of an update relies on the fact that the FE attempts to PUT with a specific revision. If another tab/process updated the record in between the update should fail and the user should make sure they are updating the latest revision. Think of how you would submig a github commit from the github UI.

kostaspap (approved):
Nice cleanup

Issue #2550, opened 2021-01-29
Fix rule error: AttributeError("'ImmutableList' object has no attribute 'get'") 


kostaspap:
### Describe the bug

Rule `AWS.CloudTrail.ResourceMadePublic` is reporting the following error: 

```
AttributeError("'ImmutableList' object has no attribute 'get'") 
```

### Steps to reproduce

1. Go to a Panther deployment
2. Go to Rules page, find rule `AWS.CloudTrail.ResourceMadePublic`
3. Click edit on the rule
4. Click "Run all tests" and see error

### Expected behavior

The rule should't report an error

### Environment

How are you deploying or using Panther?

- Panther version or commit: 1.15.2

georgepsarakis:
With the introduction of the immutable event wrapper in #2350, the event nested objects are lazily converted to `ImmutableDict` / `ImmutableList` objects instead of the built-ins `dict` and `list`. The failing `panther-analysis` rule makes use of an external Python package that performs AWS policy analysis ([policyuniverse](https://github.com/Netflix-Skunkworks/policyuniverse/)).

In `panther-analysis` we use from this package:
- [`policyuniverse.policy.Policy`](https://github.com/panther-labs/panther-analysis/blob/bc8518644e0a8fc7735576a700aa7269b3257546/aws_cloudtrail_rules/aws_resource_made_public.py#L2): used in several locations, depends on [`Statement`](https://github.com/Netflix-Skunkworks/policyuniverse/blob/c653d88417573280e344f01a47c967217c5d0956/policyuniverse/statement.py).
- [`policyuniverse.expander_minimizer.expand_policy`](https://github.com/Netflix-Skunkworks/policyuniverse/blob/c653d88417573280e344f01a47c967217c5d0956/policyuniverse/expander_minimizer.py#L200): not related / affected by this issue.
- [`policyuniverse.action_categories.categories_for_actions`](https://github.com/Netflix-Skunkworks/policyuniverse/blob/c653d88417573280e344f01a47c967217c5d0956/policyuniverse/action_categories.py#L61): not related / affected by this issue.

`policyuniverse` in order to normalize certain field values, attempts to detect if the value is of type `list` and if not, create a list with the object as the only element:
- https://github.com/Netflix-Skunkworks/policyuniverse/blob/c653d88417573280e344f01a47c967217c5d0956/policyuniverse/policy.py#L32
[...]

PR #2630, opened 2021-02-09
Add deep copying methods to the immutable types interface

georgepsarakis:
## Background

For details see https://github.com/panther-labs/panther/issues/2550.

## Changes

- Provide mutable deep copies of immutable types.
- Fix issue with `ImmutableList` iteration, which exposed the mutable nested objects of a list.

## Testing

- `mage test:python`


kostaspap (commented):
Code looks good. I will say, let's wait for sec eng team to take a look at it. 

I think there are a few follow ups to this: 
1. Update docs to mention the limitations of events and the `to_dict` function
2. I think the PAT needs to be updated to support this? @nhakmiller 
3. Do we have to change any of the existing rules to support this?

georgepsarakis:
Note that if this makes it to 1.16, we can use it to replace the hacky workaround of https://github.com/panther-labs/panther-analysis/pull/191 (which will still be needed unfortunately until upgrades are complete). That's because `.copy` is present for both `dict` and `PantherEvent` now.

kostaspap (approved):

nhakmiller (approved):

wey-chiang (approved):

PR #2631, opened 2021-02-09
Panther-managed bucket notifications

### LINKED ISSUE #2294: Do not allow '*' as an S3 Prefix in S3 Source CF Template
### Description

If a user puts in `*` as the S3 prefix when on boarding an S3 source the CF will be incorrect.

The UI should not allow this and the CF should also have a validation on this parameter.

giorgosp:
This PR integrates the backend + frontend changes for the s3 notifications feature.

Closes #2294


giorgosp:
Better merge this soon to avoid conflicts

kostaspap:
I didn't look into the code ( I believe it has been approved already). But some issues I found: 

1. Updating an S3 source to add a new prefix directs me to update the CFN template. We shouldn't be doing that: 
Screen Shot 2021-02-09 at 5 40 44 PM

2. It doesn't seem that the backend configures S3 notifications properly for some cases. 
See in the following screenshots, I expected it to have notifications for `test` and `test2` prefixes, but it has only for `test`. This looks like something that should be possible to capture with unit tests.  

Screen Shot 2021-02-09 at 5 43 36 PM
Screen Shot 2021-02-09 at 5 43 56 PM

giorgosp:
@kostaspap you are right about point 1. @alexmylonas could you take a look? Seems that this is a bug introduced in this feature.

I tested on another deployment I have and:
1.  Updated a source and changed only an existing prefix
1.  Updated a souce and added a prefix+logtypes
In none of the cases did I see the CFN template screen

alexmylonas:
@kostaspap Super nice catch! The logic exists for skipping the template step was included on master but was removed during some conflict resolution.

I have updated the code on this branch to re-instate it.

Also added a validation S3 Prefix: 
1. S3 prefix must not include `*`
2. S3 prefix must not start with `/`

alexmylonas (approved):
Reviewed code with latest updates and LGTM!

giorgosp:
Thank you!

PR #2632, opened 2021-02-09
updates based on pack api feedback

lindsey-w:
## Background

This addresses some feedback based on api usage.

## Changes

- Renaming fields because they encompass more than 'detections'
- add a new `EnumeratePack` endpoint that can return detections, data models, or globals.  
- Add counts to pack types; 
- change field name to be more representative of what it is (`SemVer` rather than `Name`).  
- Update `patch` to only take in a version id (don't pass more info than required)

## Testing

- `mage test:go` 
- `mage test:integration`


internal/core/analysis_api/handlers/list_packs.go
```diff
@@ -31,6 +31,12 @@ import (
 	"github.com/panther-labs/panther/pkg/gatewayapi"
 )
 
+func (API) EnumeratePack(input *models.EnumeratePack) *events.APIGatewayProxyResponse {
+	// Convert pack definition to a list detection option
+	listDetectionInput := getListOptions(input)
+	return handleListItems(listDetectionInput)
```
lindsey-w, line 37: If we want to re-use the `listDetections` logic, this will return the unified `Detection` object; which doesn't necessarily cover DataModels and Globals.  Unsure what to do about this... Perhaps the `PackDefinition` object should be re-worked to separate out the detections from the 'other':


    type Pack struct {
        ...
	    DetectionPattern      DetectionPattern        `json:"detectionPattern"`
        DetectionTypes        map[DetectionType]int   `json:"detectionTypes"`
    }

    type DetectionPattern struct {
        IDs []string
    }


Should this be:


    type Pack struct {
        ...
        PackDefinition    PackDefinition  `json:"packDefinition"`
        PackTypes         PackTypes `json:"packTypes"`
    }

    type PackDefinition struct {
        Detections Detections
        Helpers Helpers
    }
    type Detections struct {
        IDs []string
    }
    type Helpers struct {
        IDs []string
    }

    type PackTypes struct {
        Detections map[DetectionType]int
        Helpers    map[DetectionType]int
    }

internal/core/analysis_api/handlers/list_packs.go
```diff
@@ -71,6 +77,23 @@ func (API) ListPacks(input *models.ListPacksInput) *events.APIGatewayProxyRespon
...
+	// Packs can contain any "detection" types
+	listInput.AnalysisTypes = []models.DetectionType{models.TypeDataModel, models.TypeGlobal, models.TypePolicy, models.TypeRule}
+	// then populate the other standard fields
+	listInput.Fields = input.Fields
+	listInput.SortBy = input.SortBy
+	listInput.SortDir = input.SortDir
+	listInput.PageSize = input.PageSize
+	listInput.Page = input.Page
+	return &listInput
+}
```
lindsey-w, line 95: basically converts enumerate input into list input

internal/core/analysis_api/handlers/list_packs.go
lindsey-w: Perhaps best to create a ticket to address this in the future - I think the `detection` type does carry "enough" of the fields that would be displayed on the pack description page

nhakmiller: Ah I see, yes this does complicate things a bit. Since we don't have a clear definition/grouping of all the things in the `panther-analysis` table, it does make this stuff a bit confusing.

Perhaps we can update `handleListItems` function to not return a list of detections, but instead return a list of dynamodb items (essentially returning the `items` slice that was returned by the `scanPages` call). Then we would move lines 84-87 out of `handleListItems` and into `ListDetections`, and put a similar conversion in `enumeratePacks` which adds an `if` statement that uses the correct converter (e.g. `item.detection(), item.datamodel, etc.). Referring to these lines:

```
	for _, item := range items {
		status := compliance[item.ID].Status
		result.Detections = append(result.Detections, *item.Detection(status))
	}
```

internal/core/analysis_api/handlers/update_pack.go
```diff
@@ -326,21 +333,14 @@ func detectionSetLookup(newDetections map[string]*tableItem, input models.Detect
...
-	var detectionTypes []models.DetectionType
+func getDetectionTypeSet(detections map[string]*tableItem) map[models.DetectionType]int {
+	detectionTypes := make(map[models.DetectionType]int)
 	for _, detection := range detections {
-		if !containsDetectionType(detectionTypes, detection.Type) {
-			detectionTypes = append(detectionTypes, detection.Type)
+		if _, ok := detectionTypes[detection.Type]; ok {
+			detectionTypes[detection.Type] = detectionTypes[detection.Type] + 1
+		} else {
+			detectionTypes[detection.Type] = 1
```
nhakmiller, line 342: Not exactly sure what this function is trying to accomplish, could you add a comment? e.g.

```
// getDetectionTypeSet does xyz in order to support abc
func getDetectionTypeSet(detections map[string]*tableItem) map[models.DetectionType]int {
```

nhakmiller (approved):
Left some feedback, but overall mostly seems like some good cleaning up 👍

PR #2636, opened 2021-02-09
Fix mage deploy

nhakmiller:
## Background

Fixing mage deploy.

## Changes

- Pass in required parameter

## Testing

- ran `mage deploy`


rleighton (approved):
Keep sure master template is also updated.

nhakmiller:
> Keep sure master template is also updated.

That was how I had originally tested the change so it should be fine but just to double check I've started another deployment to test that.

austinbyers:
> Keep sure master template is also updated.

CI actually verifies this part, which is nice

PR #2642, opened 2021-02-10
Refresh S3 credentials in case Log Processing role is deleted

### LINKED ISSUE #2531: Panther should refresh the S3 role IAM credentials
### Describe the bug

Log processor will fail to process incoming events from S3 with `"InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records"`

### Steps to reproduce

1. Onboard an S3 source
2. Send some data to it, make sure Panther has processed them. 
3. Delete the CFN stack. Redeploy a new stack
4. Check log processor for errors

### Expected behavior

Log process should pick up the new IAM creds from the source

### Environment

### Additional context

Panther Log process caches the IAM role credentials indefinitely and only refreshes the STS credentials. We should refresh the IAM role credentials if we encounter `InvalidAccessKeyId` exception.

kostaspap:
## Background

Fixes #2531 

## Changes

- Creates a wrapper for S3API that forces refreshing of credentials if AKID is not valid
- Split sources_cache to its own file

## Testing

- Deployed to my account. Send some data to S3. Deleted and recreated CFN stack containing IAM Role assumable by Panther. Saw the debug message and log processor was able to read data from S3. 
Screen Shot 2021-02-10 at 3 50 25 PM


internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -61,106 +61,6 @@ type s3ClientCacheKey struct {
 	awsRegion string
 }
 
-type prefixSource struct {
```
kostaspap, line 64: Moved to its own file

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -295,6 +195,48 @@ func getNewS3Client(region *string, creds *credentials.Credentials) (result s3if
...
+
+func (r *RefreshableS3Client) GetBucketLocationWithContext(
+	ctx aws.Context,
+	request *s3.GetBucketLocationInput,
+	options ...request.Option) (*s3.GetBucketLocationOutput, error) {
+
+	response, err := r.S3API.GetBucketLocationWithContext(ctx, request, options...)
+	if awsutils.IsAnyError(err, invalidAKIDError) {
+		zap.L().Debug("encountered error, refreshing S3 client credentials", zap.Error(err))
+		r.creds.Expire()
```
kostaspap, line 224: Credential access is thread-safe (handled by the AWS SDK). By expiring the credentials, I am forcing the client to fetch some new.

internal/log_analysis/log_processor/sources/s3_client_cache_test.go
```diff
@@ -202,111 +202,3 @@ func resetCaches() {
 	bucketCache, _ = lru.NewARC(s3BucketLocationCacheSize)
 	s3ClientCache, _ = lru.NewARC(s3ClientCacheSize)
 }
-
```
kostaspap, line 205: Just moved to new file.

pkg/awsutils/awsutils.go
```diff
@@ -26,6 +26,9 @@ import (
 // Method returns true if the provided error is an AWS error with any
 // of the given codes.
 func IsAnyError(err error, codes ...string) bool {
+	if err == nil {
```
kostaspap, line 29: minor optimization to return early.

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -295,6 +195,48 @@ func getNewS3Client(region *string, creds *credentials.Credentials) (result s3if
...
+	return response, err
+}
+
+func (r *RefreshableS3Client) GetObjectWithContext(
+	ctx aws.Context,
+	request *s3.GetObjectInput,
+	options ...request.Option) (*s3.GetObjectOutput, error) {
+
+	response, err := r.S3API.GetObjectWithContext(ctx, request, options...)
+	if awsutils.IsAnyError(err, invalidAKIDError) {
```
georgepsarakis, line 236: I did a mental exercise just to see why this might be difficult to be abstracted in Go. I imagined that we could sort of use the decorator pattern, but arguments and return types are not the same (I think it would require type assertions on the return values) ... Anyway, it just crossed my mind as this probably could be encountered with other services too (?), the current implementation solves the S3 operations issue just fine.

georgepsarakis (approved):
👍

internal/log_analysis/log_processor/sources/s3_client_cache.go
kostaspap: It can happen with any code that relies on externally managed IAM roles. There are very few places that we deal with such... but yeah I think Cloud Security has the same issue.

internal/log_analysis/log_processor/sources/s3_client_cache.go
```diff
@@ -295,6 +195,48 @@ func getNewS3Client(region *string, creds *credentials.Credentials) (result s3if
...
+
+// Wrapper around S3 client. It will refresh credentials in case `InvalidAccessKeyId` error is encountered
+type RefreshableS3Client struct {
+	s3iface.S3API
+	creds *credentials.Credentials
+}
+
+// This error code will appear if the IAM role assumed by Panther log processing is deleted and recreated.
+// When we try to perform operations to S3, we will get an error that the AKID is invalid
+const invalidAKIDError = "InvalidAccessKeyId"
```
giorgosp, line 214: If I understand correctly, this PR assumes that the role is the same. What if the user updates the role? Are we going to drop events until the source cache is refreshed?

kostaspap: We get the role name from the sources, those get refreshed every 2'. If it happens that the role updates the role name but we have the old one in the cache: 
1. If the old role still exists and has permissions, no problem, we will use the old role
2. If the old role doesn't exist or doesn't have permissions, AWS will report an error. The error will propagate up the call stack and will make the log processor skip that specific SQS message. The message will reappear in the log processor SQS queue 15' later (visibility timeout), by which time the cache will be refreshed.

PR #2644, opened 2021-02-10
Ignore zero sized s3 objects in the log processor

### LINKED ISSUE #2532: Panther should not fail for 0 size files
### Describe the bug

Panther log processor will fail if it encounters a 0-size file in S3, leading to this: 
```
{"level":"warn","ts":1611835832.5195754,"caller":"processor/stream.go:132","msg":"Skipping event due to error","application":"panther","requestId":"170b5710-bd3e-4446-9647-fb7d05d68dbf","error":"EOF"}
```
### Steps to reproduce

1. Onboard a new S3 source
2. Sent an empty file to S3
3. See error
4. See notification in SQS DLQ

### Expected behavior

Panther would not log an error/warning and delete the S3 notification for such an object. 

### Environment

alxarch:
## Background

Closes #2532 

## Changes

- Ignore zero sized s3 objects

## Testing

- mage test:go

### DIFF
--- a/internal/log_analysis/log_processor/sources/s3.go
+++ b/internal/log_analysis/log_processor/sources/s3.go
@@ -125,6 +125,8 @@ func handleNotificationMessage(ctx context.Context, notification *SnsNotificatio
 func shouldIgnoreS3Object(s3Object *S3ObjectInfo) bool {
 	// We should ignore S3 objects that end in `/`.
 	// These objects are used in S3 to define a "folder" and do not contain data.
+	// We also ignore empty files since they do not contain any logs, they log errors that page on-call, and some
+	// systems like to create them during their normal process of delivering logs to Panther.
 	return s3Object.S3ObjectSize == 0 || strings.HasSuffix(s3Object.S3ObjectKey, "/")
 }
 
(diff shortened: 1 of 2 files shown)


internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -125,7 +125,7 @@ func handleNotificationMessage(ctx context.Context, notification *SnsNotificatio
 func shouldIgnoreS3Object(s3Object *S3ObjectInfo) bool {
 	// We should ignore S3 objects that end in `/`.
 	// These objects are used in S3 to define a "folder" and do not contain data.
-	return strings.HasSuffix(s3Object.S3ObjectKey, "/")
+	return s3Object.S3ObjectSize == 0 || strings.HasSuffix(s3Object.S3ObjectKey, "/")
```
alxarch, line 128: Look ma! One line!

internal/log_analysis/log_processor/sources/s3.go
```diff
@@ -125,7 +125,7 @@ func handleNotificationMessage(ctx context.Context, notification *SnsNotificatio
 func shouldIgnoreS3Object(s3Object *S3ObjectInfo) bool {
 	// We should ignore S3 objects that end in `/`.
 	// These objects are used in S3 to define a "folder" and do not contain data.
```
kostaspap, line 127: Maybe update the comment? If we don't skip 0 size files, we get errors in the log processor (it's not just a "nice to have"...)

kostaspap (approved):

internal/log_analysis/log_processor/sources/s3.go
georgepsarakis: If we should add a test case later, maybe add a TODO comment?

georgepsarakis (approved):
👍

internal/log_analysis/log_processor/sources/s3.go
alxarch: Updated

PR #2646, opened 2021-02-10
Add missing resource type to hardcoded resource type lists backend and frontend

tatehanawalt:
## Background

Resource Type validations were missing a supported Resource Type `AWS.GuardDuty.Detector.Meta`

@lindsey-w discovered the issue during bulk uploading. This showed the ResourceType was also missing from the hard coded front end list

## Changes

Added `AWS.GuardDuty.Detector.Meta` to the back end and front end hard coded lists.

## Testing

CI test


lindsey-w (commented):
LGTM (but with caveat that I don't know context around the FE change)

Thanks Tate! 🙏🏻

austinbyers (approved):
Do we also need `GSuite.Events` here? cc @lindsey-w

lindsey-w:
> Do we also need `GSuite.Events` here? cc @lindsey-w

I don't think so - that was just a [typo](https://github.com/panther-labs/panther-analysis/pull/193) in the `panther-analysis` repo.

PR #2648, opened 2021-02-11
Enable zip compression for packaged binaries

austinbyers:
## Background

Now that file sizes are printed in the deploy process, we noticed many surprisingly large zipfiles: `cfn-custom-resources.zip`, for example, is nearly 50MB. When you extract the zipfile, the Go binary contained within is the same size as the zipfile itself 🤨 

Turns out our `shutil.ZipDirectory` function was using the `.zip` format, but not actually enabling compression 🤦  

With compression on (and maxed out), the total Lambda bundle size went from 502 MB to 156 MB, which is almost 70% less data to transfer over the network during a deploy 🎉 

## Changes

In the `shutil.ZipDirectory()` function:

- Override compression to the max
- Enable compression for each file
- Better error handling for the final zipping step

## Testing

- Deployed, `mage test:integration`


pkg/shutil/archive.go
```diff
@@ -63,6 +77,7 @@ func ZipDirectory(root, savefile string, preserveModTime bool) error {
 			return err
 		}
 
+		header.Method = zip.Deflate
```
austinbyers, line 80: This is the one-line change to actually turn on the per-file compression 🙃

kostaspap (approved):
This is HUGE!

alxarch (approved):
LGTM

nhakmiller (approved):

austinbyers:
> This is HUGE!

It *used to be* huge, now it's much smaller 😉

PR #2659, opened 2021-02-12
Display the Panther-created SNS topic when an s3 source is deleted

giorgosp:
## Background

If user opted for Panther-managed bucket notifications, Panther may also create an SNS topic on the user's AWS account.
When user deletes the source, Panther does not delete the topic (it may be re-used). This PR

## Changes

-  Display the Panther-created SNS topic when an s3 source is deleted
- A small backend refactoring (get Panther queue ARN from api Config)  (by @giorgosp)

## Testing

- Manually

## Screenshot
Screenshot 2021-02-15 at 10 52 27

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -731,8 +731,7 @@ type S3LogIntegration {
 }
 
 type ManagedS3Resources {
-  topicARN: String
-  topicConfigurationIds: [String!]
+  topicARN: String!
 }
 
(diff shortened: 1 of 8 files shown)


kostaspap (commented):
Can you provide a screenshot that displays the change?

giorgosp:
> Can you provide a screenshot that displays the change?

@kostaspap added

alexmylonas (approved):

PR #2671, opened 2021-02-13
Prep for key updates

lindsey-w:
## Background

Swap out key values for prod configuration.

## Changes

## Testing

- `mage test:ci` 
- `mage test:integration`
- deployed in both `us-west-1` and `us-west-2` and verified `PollPacks` is successful


nhakmiller:
Looks good, but I'll hold off an approving until we get that key provisioned...

nhakmiller (approved):

PR #2674, opened 2021-02-15
Rule Engine: case-insensitive lookup for event fields

georgepsarakis:
## Background

Apart from making it easier for rule authors to write fields without recalling the exact case-sensitive field name, changes in case should not affect rules. In addition, Athena columns are lowercase.  

`PantherEvent` and its nested objects will now allow case-insensitive lookup.

> Unit tests (via `panther_analysis_tool`) will require using all-lowercase field names for the time being.

## Changes

- Define a mapping from lowercase key name to the original key name.
- Perform a case-sensitive search initially and fall back to case-insensitive lookup through the case-insensitive key mapping. 
- Note that the original field names will still be available.

## Testing

- `mage test:python`
- End-to-end test:
  - Deploy this branch.
  - Edit `/detections/rules/GCP.GCS.IAMChanges/edit/?section=functions`
     ```python
      def rule(event):
          return (event['resource'].get('type') == 'gcs_bucket' and
              event['protopayload'].get('methodname')
            == 'storage.setIamPermissions')
    ```
  - Run tests, still passing.

### DIFF
--- a/internal/log_analysis/rules_engine/src/immutable.py
+++ b/internal/log_analysis/rules_engine/src/immutable.py
@@ -16,7 +16,8 @@
 
 from abc import ABC, abstractmethod
 from collections.abc import Mapping, Sequence
-from typing import Any, Dict, Iterator, Type, no_type_check
+from typing import Any, Dict, Iterator, Type, \
+    Union, Collection, no_type_check
 
 
 class ImmutableContainerMixin(ABC):
@@ -67,8 +68,15 @@ class ImmutableContainerMixin(ABC):
         return iter(self._container)
 
 
-class ImmutableDict(ImmutableContainerMixin, Mapping):  # pylint: disable=R0901
-    """Read-only dictionary data type"""
+class ImmutableCaseInsensitiveDict(ImmutableContainerMixin, Mapping):  # pylint: disable=R0901
+    """
+    Read-only dictionary data type with case-insensitive lookup.
+    Assumes keys are strings and can be converted to all-lowercase with .lower().
+    """
+
+    def __init__(self, container: Union[Mapping, Collection[str]]):
+        super().__init__(container)
+        self._case_insensitive_keymap = {key.lower(): key for key in self._container}
 
     @classmethod
     def mutable_type(cls) -> Type[dict]:
@@ -76,6 +84,20 @@ class ImmutableDict(ImmutableContainerMixin, Mapping):  # pylint: disable=R0901
 
     def _shallow_copy(self, obj: dict) -> dict:
         return obj.copy()
+
+    def __getitem__(self, item: str) -> Any:
+        try:
+            # Try with the given key first
+            return super().__getitem__(item)
+        except KeyError as key_not_found:
+            # Convert to lowercase and check if a case-insensitive entry exists
+            original_key = self._case_insensitive_keymap.get(item.lower())
+            # If no entry exists, reraise the KeyError containing the given name
+            if original_key is None:
+                raise key_not_found
+            # The key exists in the container, so essentially call the parent method
+            # with the original key, in order to ensure the inherited immutability as well
+            return self.__getitem__(original_key)
 
 
 class ImmutableList(ImmutableContainerMixin, Sequence):  # pylint: disable=R0901
@@ -101,7 +123,7 @@ class ImmutableList(ImmutableContainerMixin, Sequence):  # pylint: disable=R0901
 
 
 ImmutableList.register()
-ImmutableDict.register()
+ImmutableCaseInsensitiveDict.register()
 
 
 def json_encoder(obj: Any) -> Any:
--- a/internal/log_analysis/rules_engine/src/enriched_event.py
+++ b/internal/log_analysis/rules_engine/src/enriched_event.py
@@ -18,10 +18,10 @@ from collections.abc import Mapping
 from typing import Any, Optional
 
 from .data_model import DataModel
-from .immutable import ImmutableDict, json_encoder
+from .immutable import ImmutableCaseInsensitiveDict, json_encoder
 
 
-class PantherEvent(ImmutableDict):  # pylint: disable=R0901
+class PantherEvent(ImmutableCaseInsensitiveDict):  # pylint: disable=R0901
     """Panther enriched event with unified data model (udm) access."""
 
     def __init__(self, event: Mapping, data_model: Optional[DataModel]):
--- a/internal/log_analysis/rules_engine/tests/test_engine.py
+++ b/internal/log_analysis/rules_engine/tests/test_engine.py
@@ -294,13 +294,13 @@ class TestEngine(TestCase):
                 rule_version='version',
                 log_type='log',
                 dedup='TypeError',
-                error_message='\'ImmutableDict\' object does not support item assignment: rule_id_1.py, '
+                error_message='\'ImmutableCaseInsensitiveDict\' object does not support item assignment: rule_id_1.py, '
                 'line 2, in rule    event["key"]["nested_key"] = "not_value"',
                 event={'key': {
                     'nested_key': 'value'
                 }},
                 dedup_period_mins=60,
-                title='TypeError("\'ImmutableDict\' object does not support item assignment")'
+                title='TypeError("\'ImmutableCaseInsensitiveDict\' object does not support item assignment")'
             )
         ]
 
--- a/internal/log_analysis/rules_engine/tests/test_enriched_event.py
+++ b/internal/log_analysis/rules_engine/tests/test_enriched_event.py
@@ -19,7 +19,7 @@ from unittest import TestCase
 
 from ..src.data_model import DataModel
 from ..src.enriched_event import PantherEvent
-from ..src.immutable import ImmutableList, ImmutableDict
+from ..src.immutable import ImmutableList, ImmutableCaseInsensitiveDict
 
 
 class TestEnrichedEvent(TestCase):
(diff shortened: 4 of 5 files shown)


alxarch:
> ## Background
> Apart from making it easier for rule authors to write fields without recalling the exact case-sensitive field name, changes in case should not affect rules. In addition, Athena columns are lowercase.

What case requirements exist for _Athena_ should not affect the rules engine. The stored JSON data preserves the input case of field names. _Snowflake_ might not have such requirements. We're currently 'dumbing down' Snowflake tables to be compatible with _Athena_ but this will probably change soon.
  
> 
> `PantherEvent` and its nested objects will now allow case-insensitive lookup.

This will add significant overhead _for each row_ and _for each lookup_. Instead of being `O(1)` each lookup will become `O(1+n)` where `n` is the number of keys in the `dict`. This will happen for every step in a _path lookup_. Even if you cache the lower case names on each level lazily, you won't improve this significantly since rules rarely lookup fields more than once for each row.

The only way I would see this work with a similar performance profile as the current rules engine would be if we had a pre-computed mapping of names, based on the event schema, that would allow the case-insensitive lookup to be `O(1+1)`.

Note that since all current rules work with case-sensitive lookups, you cannot observe the performance impact on production now unless you deploy a custom rule set that has all lookups use lower case names.

> 
> > Unit tests (via `panther_analysis_tool`) will require using all-lowercase field names for the time being.
> 
> ## Changes
> * Define a mapping from lowercase key name to the original key name.
> * Perform a case-sensitive search initially and fall back to case-insensitive lookup through the case-insensitive key mapping.
> * Note that the original field names will still be available.
> 
> ## Testing
> * `mage test:python`
> * End-to-end test:
>   
>   * Deploy this branch.
>   * Edit `/detections/rules/GCP.GCS.IAMChanges/edit/?section=functions`
>     ```python
>      def rule(event):
>          return (event['resource'].get('type') == 'gcs_bucket' and
>              event['protopayload'].get('methodname')
>            == 'storage.setIamPermissions')
>     ```
>   * Run tests, still passing.

internal/log_analysis/rules_engine/src/immutable.py
```diff
@@ -67,8 +68,15 @@ def __iter__(self) -> Iterator:
...
-    """Read-only dictionary data type"""
+class ImmutableCaseInsensitiveDict(ImmutableContainerMixin, Mapping):  # pylint: disable=R0901
+    """
+    Read-only dictionary data type with case-insensitive lookup.
+    Assumes keys are strings and can be converted to all-lowercase with .lower().
+    """
+
+    def __init__(self, container: Union[Mapping, Collection[str]]):
+        super().__init__(container)
+        self._case_insensitive_keymap = {key.lower(): key for key in self._container}
```
alxarch, line 79: This will be running _for each event_ being processed by the rules engine, probably recursively on deep `dict` hierarchies. Consider lazily initializing the `_case_insensitive_keymap only if a lookup fails.

alxarch (changes requested):
While I understand the benefits, I am concerned about performance issues. Mitigations:
  1. Ideally the lookup mappings would be pre-computed _once_ per-file on the first event (this would complicate the code by _a lot_ but keep a `O(1)` lookup behavior).
  2. We should lazily initialize the `_case_insensitive` dict once a lookup fails, so we only pay for it when a user is sloppy with field case

georgepsarakis:
> The stored JSON data preserves the input case of field names. 

I am aware of this, my understanding was that this might happen if you extract data from a query or if the query results during  Scheduled Rule executions contain the fields in lowercase (I may be confusing different problems here, not sure at all).

internal/log_analysis/rules_engine/src/immutable.py
georgepsarakis: I considered this before adding this calculation in the constructor. Your suggestion might look as an optimal path now, but I am almost certain that as time goes by, more users will rely on case-insensitive field names, thus making this the default branch. I could move this though to `__getitem__` for now.

georgepsarakis:
> Ideally the lookup mappings would be pre-computed once per-file on the first event (this would complicate the code by a lot but keep a O(1) lookup behavior).

@alxarch this doesn't seem possible, as far as I can think. You would need a way to create a robust cache key to retrieve the mapping from some global registry.

Creating the mapping requires 3μs for an event such as [this](https://github.com/panther-labs/panther-analysis/blob/410dc0a84eef75d64b9828453dc91fb9f86ca568/aws_s3_rules/aws_s3_insecure_access.yml#L34).

georgepsarakis:
> Another concern I have is that this will lead to weird behavior for the dict when used with other meta methods (i.e. .keys()).

We could provide the lowercased keys when iterating, I just wasn't sure if we would want to hide the real field names. I borrowed some ideas from this [structure](https://github.com/psf/requests/blob/8c211a96cdbe9fe320d63d9e1ae15c5c07e179f8/requests/structures.py#L15).

internal/log_analysis/rules_engine/src/immutable.py
georgepsarakis: It might be an improvement to account for cases with iteration as well. I think the most used operation would be lookup though in rule functions.

alxarch: > I am almost certain that as time goes by, more users will rely on case-insensitive field names,thus making this the default branch. I could move this though to __getitem__ for now.

My suggestion assumes that to provide the immutable `dict` to a rule, you do a full recursive traversal through it. If that is the case, than `__init__` will be called for every nested `dict` even for values that are irrelevant to the rule itself. Moving it to __getitem__, and supposing you are reusing the same immutable event for all rules that should match against this event,  will reduce the impact significantly as only the `dict` values relevant to the field will be initializing case-insensitive lookups.

alxarch:
> > Another concern I have is that this will lead to weird behavior for the dict when used with other meta methods (i.e. .keys()).
> 
> We could provide the lowercased keys when iterating

This would probably cause _more_ problems if for example the event is serialized to JSON to be embedded in an alert context.
The iterator meta meta methods are used by various parts of Python stdlib modules AFAIK. Making the `dict` expose its keys as if they were lower case in the first place would make it act as if we were using a new dict with lower case keys in the first place.

internal/log_analysis/rules_engine/src/immutable.py
georgepsarakis: > My suggestion assumes that to provide the immutable dict to a rule, you do a full recursive traversal through it. 

No this happens on demand, when [the nested object is retrieved](https://github.com/panther-labs/panther/blob/8cf5cf9654a6ae37cd8200096c60847283ab9afc/internal/log_analysis/rules_engine/src/immutable.py#L49).

>  you are reusing the same immutable event for all rules

The same instance is indeed used for all rules. Nested object access requires creating new instances though.

alxarch:
> > The stored JSON data preserves the input case of field names.
> 
> I am aware of this, my understanding was that this might happen if you extract data from a query or if the query results during Scheduled Rule executions contain the fields in lowercase (I may be confusing different problems here, not sure at all).

You are not confusing different issues. The case insensitive lookup makes a lot of sense from a usability standpoint. We should just make sure it is well-documented and that we also document what happens with collisions.
By collisions I mean if an event has indeed `fooBar` and `FooBar` as separate fields and the user looks up `foobar`, how do we resolve this? Is it random? Does python `json.loads` preserve order of keys as they were in the input JSON?. Do we have control over the order that this key is written in JSON? Is it possible that when processing two different such events, one rule succeeds and another fails due to the order that keys where written in JSON?

rleighton:
I’m away from my computer, so apologies if this is not an informed
question(s).

First, isn’t it the case that most rules only look up one or two keys? If
so , that should inform the design.

Second, my naive approach would be just a that get() looks up, if found,
return, if not lookup in lower cache dict, if not found lowercase key ,
look up, if found up date lowercase cache, and return

kostaspap:
Trying to take a step back... The problem is essentially the following: When querying Athena, it always returns the keys of a dictionary in lowercase. This can be a problem when someone tries to create a rule out of results fetched from Athena.
It might be computationally expensive to support fully case insensitive lookup, but what if we would support the following:

`if event.get(key) is None return event.get(key.toLowerCase())`. 

The above relies on the fact that keys in the `event` will either have the correct case, or they will be lowercase. Thoughts ?

alxarch:
> Second, my naive approach would be just a that get() looks up, if found,
return, if not lookup in lower cache dict,

I assume this is to save us from lower-casing the key unnecessarily

>  if not found lowercase key ,
look up, if found up date lowercase cache, and return

But this does not match all cases of the key in the end. The second lookup will only match if the field in the original dict _was_ lowercase but the user provided a non-lowercase key.

kostaspap:
> Trying to take a step back... The problem is essentially the following: When querying Athena, it always returns the keys of a dictionary in lowercase. This can be a problem when someone tries to create a rule out of results fetched from Athena.
> Since it can be computationally expensive to support fully case insensitive lookup, may we should just support the following:
> 
> `if event.get(key) is None return event.get(key.toLowerCase())`.
> 
> The above relies on the fact that keys in the `event` will either have the correct case, or they will be lowercase. Thoughts?

I acknowledge that this suggestion is different that the one we had agreed on the relevant Github issue... However I am thinking that if having fully case insensitive search is challenging/expensive, it might be sufficient just to implement the above.

alxarch:
> I acknowledge that this suggestion is different that the one we had agreed on the relevant Github issue... However I am thinking that if having fully case insensitive search is challenging/expensive, it might be sufficient just to implement the above.

After a small discussion with @georgepsarakis  I am not that concerned about performance. My main concern is to have a deterministic way to solve collisions. If an event contains both `FooBar` and `fooBar` fields with _different values_, it is important to have a clear rule about which of the two fields will match when I query it with `foobar`. There are two approaches to this:
  1. Do not add mappings for keys that collide, forcing the user to use the correct case when there are collisions.
  2. Make sure you return a stable result by matching the same mapping every time for all events. One way to achieve this is to rely on the order of keys in the JSON value. But this might be volatile if _some fields_ are included _sometimes_. To be completely deterministic this approach would require to have knowledge of the schema. This is what we do when we define the storage descriptor of a _Glue_ table.

kostaspap:
> > I acknowledge that this suggestion is different that the one we had agreed on the relevant Github issue... However I am thinking that if having fully case insensitive search is challenging/expensive, it might be sufficient just to implement the above.
> 
> After a small discussion with @georgepsarakis I am not that concerned about performance. My main concern is to have a deterministic way to solve collisions. If an event contains both `FooBar` and `fooBar` fields with _different values_, it is important to have a clear rule about which of the two fields will match when I query it with `foobar`. There are two approaches to this:
> 
> 1. Do not add mappings for keys that collide, forcing the user to use the correct case when there are collisions.
> 2. Make sure you return a stable result by matching the same mapping every time for all events. One way to achieve this is to rely on the order of keys in the JSON value. But this might be volatile if _some fields_ are included _sometimes_. To be completely deterministic this approach would require to have knowledge of the schema. This is what we do when we define the storage descriptor of a _Glue_ table.

No1 from above sounds reasonable to me.

georgepsarakis:
Thanks everyone for the feedback! Let's pause this for a while and revisit it next week.

georgepsarakis:
Closing, will be transferred with a reference here.

PR #2675, opened 2021-02-15
Deleting an s3 source succeeds even if the iam role was deleted

### LINKED ISSUE #2657: If user deletes the IAM role, deleting an s3 source fails
### Steps to reproduce

1. Onboard an s3 source, enable Panther to manage bucket notifications
2. Finish source onboarding
3. Delete the IAM role
4. Delete source - Removing bucket notifications will fail because the role is absent and source deletion will fail as well

### Expected behavior

Deleting a source should succeed even if the IAM role is deleted

### Environment

- Panther version or commit: master, 1.16.0-dev

giorgosp:
## Background

Removing bucket notifications may fail, eg if user deleted the iam role before deleting the source.
This PR makes deletion of a source succeed even if removing bucket notifications fail.

## Changes

- Log but don't fail deletion when Panther fails to remove bucket notifications

## Testing

- manually

Closes #2657

### DIFF
--- a/internal/core/source_api/api/delete_integration.go
+++ b/internal/core/source_api/api/delete_integration.go
@@ -67,23 +67,23 @@ func (api *API) DeleteIntegration(input *models.DeleteIntegrationInput) error {
 				break
 			}
 		}
-
-		if integrationItem.ManagedBucketNotifications {
-			source := itemToIntegration(integrationItem)
-			err = RemoveBucketNotifications(api.AwsSession, source)
-			if err != nil {
-				zap.L().Error("failed to remove bucket notifications",
-					zap.Error(err), zap.String("integrationId", input.IntegrationID))
-				return deleteIntegrationInternalError
-			}
-		}
-
 		if shouldRemovePermissions {
 			if err = api.DisableExternalSnsTopicSubscription(integrationItem.AWSAccountID); err != nil {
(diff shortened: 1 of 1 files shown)


internal/core/source_api/api/delete_integration.go
```diff
@@ -67,23 +67,22 @@ func (api *API) DeleteIntegration(input *models.DeleteIntegrationInput) error {
...
+					zap.Error(err))
+				return deleteIntegrationInternalError
+			}
+		}
 
 		if integrationItem.ManagedBucketNotifications {
 			source := itemToIntegration(integrationItem)
 			err = RemoveBucketNotifications(api.AwsSession, source)
 			if err != nil {
+				// Hanadle the error here so that deleting the source succeeds anyway
```
kostaspap, line 83: Maybe add more context in the comment as to "why" it is fine to just ignore it. 
e.g. something like "Don't propagate the error and allow the delete operation to succeed. The users could have already deleted the IAM role that we assume to complete this operation - we don't want to fail in such scenario"

Or something like this. 
Also I would say logging a warning or even info would be enough.

kostaspap (approved):
LGTM, just minor comment.

internal/core/source_api/api/delete_integration.go
giorgosp: Thanks! Updated

PR #2681, opened 2021-02-15
[FE] Move sources to `Integrations` menu entry

3nvi:
## Background

In a process of improving our navigation menu for newcomers, we proposed to move all of the available "connectors" in a new menu entry named `Integrations` which will host all the 3rd-party services that Panther can connect to. 

This PR implements just that

## Screenshots

Screen Shot 2021-02-15 at 5 48 06 PM

## Changes

- Add new collapsable menu entry named `Integrations`
- Move all CS and LA sources under Integrations
- Move & rename Destinations entry under Integrations

## Testing

`npm run test`

### DIFF
--- a/web/src/routes.tsx
+++ b/web/src/routes.tsx
@@ -126,21 +126,6 @@ const PrimaryPageLayout: React.FunctionComponent = () => {
                   path={urls.compliance.resources.details(':id')}
                   component={ResourceDetailsPage}
                 />
-                <Route
-                  exact
-                  path={urls.compliance.sources.list()}
-                  component={ListComplianceSourcesPage}
-                />
-                <Route
-                  exact
-                  path={urls.compliance.sources.create()}
-                  component={CreateComplianceSourcePage}
-                />
-                <Route
-                  exact
-                  path={urls.compliance.sources.edit(':id')}
-                  component={EditComplianceSourcePage}
-                />
                 {/* ******************* LOG ANALYSIS ***************************** */}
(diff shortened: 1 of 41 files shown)


web/src/routes.tsx
```diff
@@ -163,55 +148,91 @@ const PrimaryPageLayout: React.FunctionComponent = () => {
...
+                {/* ******************* INTEGRATIONS ***************************** */}
+                <Redirect
+                  exact
+                  from={urls.integrations.home()}
+                  to={urls.integrations.logSources.list()}
+                />
+                <Redirect
+                  exact
+                  from={`${urls.integrations.logSources.list()}:type`}
+                  to={urls.integrations.logSources.list()}
```
alexmylonas, line 179: Hm what's this redirection protects from?

alexmylonas (approved):
Nice job Aggelos, this must have been tedious. 

LGTM! 💯

web/src/routes.tsx
3nvi: You added it a while ago :D I think it had to do with breadcrumb clicking

alexmylonas: Can't recall right now, if you don't sync-to-release you could remove it.

kouknick (approved):
Great work @3nvi. Just a minor comment and we are good to go.
It's so awesome that we were proactive and didn't have all those Urls hardcoded, otherwise this would have been so much harder 😉 .

web/src/routes.tsx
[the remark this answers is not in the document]
kouknick: I just saw your conversation above, you can ignore this one!

Issue #2641, opened 2021-02-10
[BE] Unified Integrations


alxarch:
## Background

For the Unified Integrations to work properly the BE must provide two endpoints to describe the _Log Sources_ available.

## Design

The BE should provide 2 new endpoints:
  1. `ListLogSources` that returns a list of Log Sources optional `Category` filter
  2. `GetLogSource` that returns the full details for a single `LogSource` item

Each `Log Source` entry should provide the  information required to render the UI such as:
 - `name`: a user-friendly name for the entry (i.e. `Cisco Umbrella`)
 - `id`: a computer-friendly identifier for the entry (i.e. `cisco_umbrella`). This can be used to resolve icon assets for the Log Source.
 - `App Info`/ `UseCases` / `How It Works`: description stuff that fill the screens about each Log Source
 - `transportMethods`: a list of available transport methods for this `Log Source`
 - `defaultParsers`: a list of log type identifiers to pre-select on a new source
 - `availableParsers`: a list of all available log parsers for this Source.
 - `popularity`: An arbitrary 'score' for sorting the entries. Alphanumeric order would not be very helpful especially in the 'all' page. I suggest we use such a field to have control of the sorting on the definition side.
 
We need to have two lists for log parsers  (`available` / `default`) for the following reasons:
- Some Log Sources have many available log types to choose from and we can provide some as pre-selected but not all (i.e. `AWS`)
- We plan to provide alternative formats for each log type in the near future by separating parsers/schemas. Using this we can list all parsers and have the default ones pre-selected.
- It provides a solution to handling user-defined log parsers (a.k.a. `CustomLogs`). These could all be grouped in a `Custom` Log Source with empty `defaultParsers` and let the user choose which of their own parsers they want to use from the `available` list.
- It avoids having to re-define acros FE/BE the current `ListAvailableLogTypes` that is currently used. That endpoint does not have filtering and in our case, once a user has chosen a Log Source the options for available parsers should be restricted.

## Notes

@3nvi I will add two examples for each endpoint in the comments below and we can discuss this further

alxarch:
## `ListLogSources` endpoint example 

FE wants to show the list page for the `Server` category of `Log Sources`

 <details>
    <summary>ListLogSourcesInput</summary>

```json
{
   "ListLogSources": {
      "Category": "Server"
   }
}
```

 </details>

 <details>
    <summary>ListLogSourcesOutput</summary>

```json
{
   "LogSources": [
      {
          "name": "Apache",
[...]

Issue #2576, opened 2021-02-01
Unified Integrations Page [MVP]


mattthetraveler:
## Description

To keep with the theme of unification, we should consolidate all pages related to external integrations into one tab with 3 sub-tabs: Alert Destinations, Log Sources, and Cloud Accounts.

In addition to consolidation, we should expand our Log Sources experience to display all of log sources we support in Panther today, similar to our external integrations page.

## RFC

[RFC](https://panther-labs.atlassian.net/wiki/spaces/EN/pages/720207938/RFC+Simplify+the+UI)

## Design

- [Accounts](https://share.goabstract.com/96a7b795-f26c-433a-b567-12ed553fa8ce)
- [Alert Destinations](https://share.goabstract.com/f8389484-d590-419e-8fa2-9b8bb8f599ca)
- [Log Sources](https://share.goabstract.com/dbf178c4-c6fa-4a20-9646-6d95d07de6db)

## Status and Stakeholders

> **_Note:_** Tag each responsible party below

- [x] Epic created [product] @mattthetraveler 
- [x] RFC written [feature owner] @jacknagz 
- [x] Frontend assigned [frontend] @3nvi 
- [x] Designs mocked [designer] @dimitriszoitas 
- [ ] Stories planned [fe/be] @3nvi 
- [ ] Stories executed [fe/be] @3nvi 
- [ ] Validation [product/feature owner] @mattthetraveler 

# Stories and Criteria

### User Stories

As a security engineer
I want to view, manage, add, and delete to my connected AWS accounts in the Panther UI
To initiate Cloud Security scans

### Acceptance Criteria
- As a user, I should be able to view "Cloud Accounts" in the new "Integrations" tab
- User can view connected AWS accounts in the Integrations tab in the "Cloud Accounts" sub-tab
- User can manage connected AWS account in the "Cloud Accounts" sub-tab

---

### User Story

As a security engineer
When adding a new log source
I want to see all of the supported log types in a single window
To understand all the possible types of data I can analyze

### Acceptance Criteria

- As a user, I should be able to view my log sources in a "Log Sources" sub-tab in the "Integrations" tab
- Rename to "Log Sources"
- Maintain existing "Sources" view page where user can see all SaaS Log Pullers and Data Transports

In Log Source onboarding experience:

**SaaS Log Pullers**
  - Reuse tiles that exist in the current "Sources" onboarding experience
[...]

PR #2682, opened 2021-02-15
[FE] GraphQL Schema proposal for Unified Integrations

3nvi:
## Background

This is a Draft PR showcasing the proposed schema for the implementation of the new unified integrations page. The designs for this can be found in the links below.

This PR will serve as a contract between the BE counterparty

## Epic

The epic that describes the related proposal can be found [here](https://github.com/panther-labs/panther/issues/2576)

## BE effort

The BE proposal of exposed lambda functions can be found [here](https://github.com/panther-labs/panther/issues/2641) (Make sure you read the comments)

## Designs

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/df9d59c8-eeb1-452d-a5f5-3712b64010be/commits/latest/files/3f113d4e-19e8-43b1-8e2f-148ac5f988c8/layers/55695392-FA58-4422-B75F-D88C5B80BC8A?collectionId=6bf1687f-9f86-463b-bc75-c079595a19aa&collectionLayerId=3d8e34e3-77f6-4685-a81a-3262056405e6

and 

https://app.abstract.com/projects/558a8de6-7134-4c8b-91c5-62074bb1279b/branches/df9d59c8-eeb1-452d-a5f5-3712b64010be/commits/0259dc1138c4d66d3db2d1da2b8506bbfc83818d/files/3f113d4e-19e8-43b1-8e2f-148ac5f988c8/layers/107FE9A4-9DB1-448D-97A8-377003B28E3E?collectionId=6bf1687f-9f86-463b-bc75-c079595a19aa&collectionLayerId=4933c53d-8e39-4090-8647-b2c1fbfd39d8

## Changes

- Add proposed GraphQL fields

## Testing

- N/A

## Notes

Tests are failing due to the fact that we haven't made the necessary code updates to support the new `logSourceId` field

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -797,7 +797,7 @@ input AddS3LogIntegrationInput {
   kmsKey: String
   s3PrefixLogTypes: [S3PrefixLogTypesInput!]!
   managedBucketNotifications: Boolean!
-  logSourceId: ID
+  logSourceId: ID!
 }
 
 input SqsLogConfigInput {
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -809,7 +809,7 @@ input SqsLogConfigInput {
 input AddSqsLogIntegrationInput {
   integrationLabel: String!
   sqsConfig: SqsLogConfigInput!
-  logSourceId: ID
+  logSourceId: ID!
 }
 
 input UpdateComplianceIntegrationInput {
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -58,6 +58,7 @@ type Query {
   destination(id: ID!): Destination
   destinations: [Destination]
   generalSettings: GeneralSettings!
+  getAvailableLogSource(id: ID!): AvailableLogSource!
   getComplianceIntegration(id: ID!): ComplianceIntegration!
   getComplianceIntegrationTemplate(
     input: GetComplianceIntegrationTemplateInput!
@@ -74,10 +75,11 @@ type Query {
   policy(input: GetPolicyInput!): Policy
   policiesForResource(input: PoliciesForResourceInput): ListComplianceItemsResponse
   listAvailableLogTypes: ListAvailableLogTypesResponse!
+  listAvailableLogSources(input: ListAvailableLogSourcesInput): [AvailableLogSource!]!
   listComplianceIntegrations: [ComplianceIntegration!]!
   listDataModels(input: ListDataModelsInput!): ListDataModelsResponse!
   listLogIntegrations: [LogIntegration!]!
-  listPacks(input: ListPacksInput) : ListPacksResponse!
+  listPacks(input: ListPacksInput): ListPacksResponse!
   organizationStats(input: OrganizationStatsInput): OrganizationStatsResponse
   getLogAnalysisMetrics(input: LogAnalysisMetricsInput!): LogAnalysisMetricsResponse!
   rule(input: GetRuleInput!): Rule
@@ -85,6 +87,10 @@ type Query {
   users: [User!]!
   getCustomLog(input: GetCustomLogInput!): GetCustomLogOutput!
   listCustomLogs: [CustomLogRecord!]!
+}
+
+input ListAvailableLogSourcesInput {
+  category: String!
 }
 
 input ListPacksInput {
@@ -223,6 +229,30 @@ input GetAlertInput {
   alertId: ID!
   eventsPageSize: Int
   eventsExclusiveStartKey: String
+}
+
+type AvailableLogSourceParser {
+  name: String!
+  description: String
+  referenceUrl: String
+}
+
+type AvailableLogSourceDetails {
+  appInfo: String
+  howItWorks: String
+  useCases: String
+  defaultParsers: [String!]!
+  transports: [String!]!
+}
+
+type AvailableLogSource {
+  id: ID!
+  displayName: String!
+  popularity: Int!
+  categories: [String!]!
+  isEnterprise: Boolean!
+  availableParsers: [AvailableLogSourceParser!]!
+  details: AvailableLogSourceDetails
 }
 
 type ListDetectionsResponse {
@@ -529,7 +559,6 @@ type Pack {
   detectionTypes: DetectionTypes!
 }
 
-
 input SuppressPoliciesInput {
   policyIds: [ID]!
   resourcePatterns: [String]!
@@ -718,6 +747,7 @@ type S3LogIntegration {
   integrationId: ID!
   integrationType: String!
   integrationLabel: String!
+  logSourceId: ID
   lastEventReceived: AWSDateTime
   s3Bucket: String!
   s3Prefix: String
@@ -741,6 +771,7 @@ type SqsLogSourceIntegration {
   integrationLabel: String!
   integrationType: String!
   lastEventReceived: AWSDateTime
+  logSourceId: ID
   sqsConfig: SqsConfig!
   health: SqsLogIntegrationHealth!
 }
@@ -766,6 +797,7 @@ input AddS3LogIntegrationInput {
   kmsKey: String
   s3PrefixLogTypes: [S3PrefixLogTypesInput!]!
   managedBucketNotifications: Boolean!
+  logSourceId: ID
 }
 
 input SqsLogConfigInput {
@@ -777,6 +809,7 @@ input SqsLogConfigInput {
 input AddSqsLogIntegrationInput {
   integrationLabel: String!
   sqsConfig: SqsLogConfigInput!
+  logSourceId: ID
 }
 
 input UpdateComplianceIntegrationInput {
--- a/web/__generated__/schema.tsx
+++ b/web/__generated__/schema.tsx
@@ -119,13 +119,13 @@ export type AddS3LogIntegrationInput = {
   kmsKey?: Maybe<Scalars['String']>;
   s3PrefixLogTypes: Array<S3PrefixLogTypesInput>;
   managedBucketNotifications: Scalars['Boolean'];
-  logSourceId?: Maybe<Scalars['ID']>;
+  logSourceId: Scalars['ID'];
 };
 
 export type AddSqsLogIntegrationInput = {
   integrationLabel: Scalars['String'];
   sqsConfig: SqsLogConfigInput;
-  logSourceId?: Maybe<Scalars['ID']>;
+  logSourceId: Scalars['ID'];
 };
 
 export type Alert = {
--- a/web/__generated__/schema.tsx
+++ b/web/__generated__/schema.tsx
@@ -119,11 +119,13 @@ export type AddS3LogIntegrationInput = {
   kmsKey?: Maybe<Scalars['String']>;
   s3PrefixLogTypes: Array<S3PrefixLogTypesInput>;
   managedBucketNotifications: Scalars['Boolean'];
+  logSourceId?: Maybe<Scalars['ID']>;
 };
 
 export type AddSqsLogIntegrationInput = {
   integrationLabel: Scalars['String'];
   sqsConfig: SqsLogConfigInput;
+  logSourceId?: Maybe<Scalars['ID']>;
 };
 
 export type Alert = {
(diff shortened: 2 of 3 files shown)


api/graphql/schema.graphql
```diff
@@ -58,6 +58,7 @@ type Query {
   destination(id: ID!): Destination
   destinations: [Destination]
   generalSettings: GeneralSettings!
+  getAvailableLogSource(id: ID!): AvailableLogSource!
```
3nvi, line 61: using prefix `xxxAvailablexxx` since the concept of Log Source has multiple different contextual meanings. In this case we want to get information about a new/available log source connector

api/graphql/schema.graphql
```diff
@@ -225,6 +231,30 @@ input GetAlertInput {
   eventsExclusiveStartKey: String
 }
 
+type AvailableLogSourceParser {
+  name: String!
+  description: String
+  referenceUrl: String
```
3nvi, line 237: Each Log Source matches to log types. We want additional information for each log type though to be able to display information in the UI for each and every one. Thus a new `type` is created

api/graphql/schema.graphql
```diff
@@ -225,6 +231,30 @@ input GetAlertInput {
...
+}
+
+type AvailableLogSource {
+  id: ID!
+  displayName: String!
+  popularity: Int!
+  categories: [String!]!
+  isEnterprise: Boolean!
+  availableParsers: [AvailableLogSourceParser!]!
+  details: AvailableLogSourceDetails
```
3nvi, line 255: instead of creating 2 separate types, it's easier for us to use a single top-level type and nest "additional API calls" as fields within it.

It's optional since in the listing page these details might not be included or returned from the API

api/graphql/schema.graphql
```diff
@@ -225,6 +231,30 @@ input GetAlertInput {
...
+type AvailableLogSourceParser {
+  name: String!
+  description: String
+  referenceUrl: String
+}
+
+type AvailableLogSourceDetails {
+  appInfo: String
+  howItWorks: String
+  useCases: String
```
3nvi, line 243: optional since not all log sources will have every information needed with regards to how it works (or possible use cases for this log source)

api/graphql/schema.graphql
```diff
@@ -718,6 +747,7 @@ type S3LogIntegration {
   integrationId: ID!
   integrationType: String!
   integrationLabel: String!
+  logSourceId: ID
```
3nvi, line 750: @alexmylonas this is so we can know, during an update, which log source created this integration  in order to "limit" the log types htat the dropdown shows

api/graphql/schema.graphql
```diff
@@ -741,6 +771,7 @@ type SqsLogSourceIntegration {
   integrationLabel: String!
   integrationType: String!
   lastEventReceived: AWSDateTime
+  logSourceId: ID
```
3nvi, line 774: all of them optional for backwards compatibility with existing integrations

api/graphql/schema.graphql
alxarch: Which case is that? Is this different from the one required in detail view?

alxarch (commented):
I am a bit confused as to why we nest the details as optional.
I though the whole point of GraphQL was that required fields were explicitly declared in the query and the server only returns the fields requested.
Why do we have to nest the `details` field and not have all fields top-level?

api/graphql/schema.graphql
```diff
@@ -225,6 +231,30 @@ input GetAlertInput {
...
+  description: String
+  referenceUrl: String
+}
+
+type AvailableLogSourceDetails {
+  appInfo: String
+  howItWorks: String
+  useCases: String
+  defaultParsers: [String!]!
+  transports: [String!]!
```
alxarch, line 245: Shouldn't transports be an enum, restricting the valid values?

api/graphql/schema.graphql
alxarch: Do GraphQL queries support comments? It would be useful to document the types and fields in the schema itself.

api/graphql/schema.graphql
3nvi: We currently use the name `Log Sources` for the log sources that the user has connected Panther to (All S3, SQS, GSuite, Box, etc. integrations).

Thus the type `LogSource` is already "taken" and maps to the details of a particular connect log source. I want to separate what we have here from the already connected log sources, since they are different things

api/graphql/schema.graphql
3nvi: Adding `#` allows you to writte a comment in any `*.graphql` file. It's not typical (from my experience) to comment all fields though, since the point of GraphQL is to be self-explanatory

api/graphql/schema.graphql
giorgosp: What is the difference between integration and logSource? 
Also, I think it would be very helpful to include such PR comments in the code itself

api/graphql/schema.graphql
3nvi: Yeah, you're right! Unfortunately, GraphQL enums don't support dashes (`-`) and we use dashes in our transport methods (i.e. `aws-s3`, `aws-scan`). This prevents us from using enums here. Funnily, we don't use dashes anywhere else, so the rest of the app has a lot of enums. This piece can't though :/

api/graphql/schema.graphql
3nvi: Business has mixed up the concept of Log Sources. We use the samee word to denote both "available log sources that you can connect Panther to" and "your connected Log sources where Panther pulls logs from".

In tthe formeer case, we return details about this log source (i.e. how it works, what it is, what log types can it support, what icon does it have, etc.). In the latter case, we show information about hte healtth oof the source, its ID, the S3 prefix that the user hass added, etc.

After this set of changes, we need to know "how" did a connected log source get created. What did hte user click when creating it? Did they click "Suricata" and then wentt to the wizard? Did they click "OSQuery" and then went tto the wizard? We don't know.

We need to know though (after the integrations re-work) so we limit the log types that we show during an update, to **only** the ones that were related to the user's original intent (i.e. only shoow suricata or osquery logs respectively).

To achieve that we need  to "store" the "log service" that the user attempted to onboard when they created a new Log Source in our system. This will be stored under `logSourceId`.

The naming is really bad, but it stems from the fact that a "potential/available log source" and a "connected log source" are different things. Should we perhaps rename some things as part of this work and introduce:

- `AvailableLogSource` type
- `ConnectedLogSource` type ?

If not, do we have a better naming for them? @alxarch @kostaspap

3nvi:
> I am a bit confused as to why we nest the details as optional.
> I though the whole point of GraphQL was that required fields were explicitly declared in the query and the server only returns the fields requested.
> Why do we have to nest the `details` field and not have all fields top-level?

Yes, but each field has a resolver. If I did what you say, I'd need to add a resolveer for every single field (i.e. a resolveer for `howItWorks`, a resolver for `appInfo`, a resolver for `useCases`, etc.)


It's a complex answer to give you through a comment, so I'll opt to ping you offline and explain :)

api/graphql/schema.graphql
alexmylonas: Just to add my 2-cents here. I think we also include "compliances" sources in the new format. Right now, we only support an AWS account but we could rename it `getAvailableSources` to be more accurate. 

Not feeling too strong about this just a thought.

api/graphql/schema.graphql
```diff
@@ -87,6 +89,10 @@ type Query {
   listCustomLogs: [CustomLogRecord!]!
 }
 
+input ListAvailableLogSourcesInput {
+  category: String!
```
alexmylonas, line 93: Is this required?

kouknick (approved):
LGTM @3nvi!

api/graphql/schema.graphql
```diff
@@ -225,6 +231,30 @@ input GetAlertInput {
...
+  defaultParsers: [String!]!
+  transports: [String!]!
+}
+
+type AvailableLogSource {
+  id: ID!
+  displayName: String!
+  popularity: Int!
+  categories: [String!]!
+  isEnterprise: Boolean!
```
alexmylonas, line 253: Just a minor comment to reduce tech debt down the line 

```suggestion
  available: Boolean!
```
or
```suggestion
 availableIn: [String!]! 
```
Where string can be ['OSS', 'Enterprise', .... 'Silver Plan', 'Gold plan']

Totally good to have and optional 
cc: @alxarch

api/graphql/schema.graphql
```diff
@@ -766,6 +797,7 @@ input AddS3LogIntegrationInput {
   kmsKey: String
   s3PrefixLogTypes: [S3PrefixLogTypesInput!]!
   managedBucketNotifications: Boolean!
+  logSourceId: ID
```
alexmylonas, line 800: Required for all new though?
```suggestion
  logSourceId: ID!
```

api/graphql/schema.graphql
```diff
@@ -777,6 +809,7 @@ input SqsLogConfigInput {
 input AddSqsLogIntegrationInput {
   integrationLabel: String!
   sqsConfig: SqsLogConfigInput!
+  logSourceId: ID
```
alexmylonas, line 812: Same as above
```suggestion
  logSourceId: ID!
```

api/graphql/schema.graphql
3nvi: it is. Each available log source **must** belong to a category

api/graphql/schema.graphql
3nvi: I'm invisioning simply removing this down the line, as this will no longer be needed 😉

api/graphql/schema.graphql
3nvi: Yes that's correct. Thanks!

api/graphql/schema.graphql
3nvi: 👍

api/graphql/schema.graphql
alxarch: I would suggest to use `tiers: [String!]` to introduce a good name for both ee/oss and the 'plans' of SaaS

api/graphql/schema.graphql
giorgosp: Thank you for you comment Aggelos. Apparently I'm missing a lot of context about the integrations rework, so I cannot suggest better names at the moment. 

I hope though that we don't keep `integrationId` and `logSourceId`, because they mean the same to me :P (In the backend we use the terms interchangeably which is bad too).

api/graphql/schema.graphql
alexmylonas: I would prefer is BE decides if the AvailableSource is actually available for an account instead of FE compared available "tiers" to account tier but would work

api/graphql/schema.graphql
giorgosp: Checked out the epic description and the designs and got the point now. 
`logSourceId` seems to refer to one type of logs, eg to Suricata logs or OSQuery logs. In an s3 bucket, users may have several "log sources". For example, users may have a "security-logs" bucket with `Apache` logs at `apache/` prefix and `osquery` logs at `osquery/` prefix. They will probably want to onboard this bucket to Panther once and not have different s3 integrations for the same bucket. Which logSourceId do we pick in this scenario?

3nvi: > In an s3 bucket

@giorgosp According to the designs, this will no longer be 100% possible. You can't "choose S3 bucket". You alwayss choose logs first. If you want multiple logs from **different services** in the same S3 bucket, then you have to go through the wizard multtiple times.

This effectively nullifies the work for multiplee S3 prefixes if the services are different. If log typse are different but the service remains the same, then this works fine

api/graphql/schema.graphql
3nvi: > I would prefer is BE decides if the AvailableSource is actually available for an account

I think that we still need to show that this is supported in this tier or the other tier. We can't escape exposing this to the FE somehow

api/graphql/schema.graphql
alxarch: > This effectively nullifies the work for multiplee S3 prefixes if the services are different.

This is a regression in usability IMO. The idea of this feature was to help a user with some sane defaults and not cripple functionality in the process.
One suggestion for this would be to list _all_ available parsers (both _managed_ and _user-defined_) in the `Custom` option so that a user can set up their source any way _they_ think is appropriate for their use case.
Having a `Custom.` prefix on user-defined parsers helps with this because narrowing the selection to the user-defined ones would be easy in the the parser selector.

giorgosp: Agree with @alxarch, this would be a regression in UX. Onboarding an s3 bucket is a tedious process for the user. I don't think we want them to onboard the same bucket every time they want to send logs from a different S3 prefix.

Quoting from the epic's RFC:
>Users can add new integrations by clicking “Integrations”, and having the option to select:
>    Source [SaaS, AWS Accounts, Buckets, Queues, etc]

What I understand from the above is that a S3 bucket or SQS queue is also a "log source" and should be presented side by side with Panther's 3rd party integrations.

alxarch: @3nvi As for the naming issue, I think this is not easily solvable.
We probably need something without _Source_ in it. Even better without _Log_ either.
Here's a poll, vote by emoticon.

  1. :rose:  Log Provider
  2. :cactus: Log Group
  3. :car: Log Producer
  4. :steam_locomotive: Parser Group
  5. :airplane: Parser Set

3nvi: 👍 (can't emote on off-topic comments)

3nvi:
Closing in favor of https://github.com/panther-labs/panther-enterprise/pull/2343

PR #2683, opened 2021-02-15
Fixing dashboards

kostaspap:
## Background

After some recent changes, @rleighton reported that the CW Dashboards were broken. 
I fixed the issue and also switched some of the dashboards to use CW Metrics instead of CW Logs Insights, which significantly speeds up loading of the dashboards (the CW Dashboard no longer has to run a CW Logs Insights query to create graphs)

## Changes

- Removed the `Input File count by Log Type` widget. This didn't make much sense since a single file can have multiple log types (e.g. OsQuery)
- Modified `Input MBytes (Uncompressed) by Log Type`. It is now named `Input data by Log Type ( MB - Uncompressed)` and it used CW Metrics instead of CW Log Insights. 
- Modified `Input MBytes (Uncompressed) Processed`. It is now named `Input Processed (MB Uncompressed)` and it used CW Metrics instead of CW Log Insights. 
- Modified `Input Files Processed` to use CW Metrics.
- Modified `Output Events Written to S3`. It is now named `Output Files written to S3` and it uses CW Metrics
- Modified `Output MBytes (Compressed) written to S3`. It now uses CW Metrics. 
## Testing

- Deployed to my own AWS account. See screenshot below: 
Screen Shot 2021-02-15 at 6 56 56 PM
Screen Shot 2021-02-15 at 6 57 15 PM

### DIFF
--- a/internal/log_analysis/log_processor/destinations/s3.go
+++ b/internal/log_analysis/log_processor/destinations/s3.go
@@ -85,6 +85,8 @@ func CreateS3Destination(jsonAPI jsoniter.API) Destination {
 		s3Bucket:            common.Config.ProcessedDataBucket,
 		snsTopicArn:         common.Config.SnsTopicARN,
 		latencyCounter:      logmetrics.EventLatencySeconds,
+		outputFiles:         logmetrics.OutputFiles,
+		outputBytes:         logmetrics.OutputBytes,
 		maxBufferedMemBytes: maxS3BufferMemUsageBytes(common.Config.AwsLambdaFunctionMemorySize),
 		maxBufferSize:       uploaderBufferMaxSizeBytes,
 		maxDuration:         maxDuration,
@@ -125,6 +127,8 @@ type S3Destination struct {
 	maxBuffers          int
 	jsonAPI             jsoniter.API
 	latencyCounter      metrics.Counter
+	outputFiles         metrics.Counter
+	outputBytes         metrics.Counter
 }
 
 // SendEvents stores events in S3.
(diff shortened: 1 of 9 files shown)


internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -225,21 +229,10 @@ func (d *S3Destination) sendData(buffer *s3EventBuffer, errChan chan error) {
...
 
 	var (
-		err           error
-		contentLength int64
-		key           string
+		err error
+		key string
 	)
 
-	operation := common.OpLogManager.Start("sendData", common.OpLogS3ServiceDim)
```
kostaspap, line 233: I thought it made sense to remove this: 
- The dashboard are now constructed by using metrics
- It saves us some $$ - we don't log every time we generate a new object in S3. 
- In case of an error, the error will anyway be propagated and will be logged

internal/log_analysis/log_processor/destinations/s3.go
```diff
@@ -270,19 +261,12 @@ func (d *S3Destination) sendData(buffer *s3EventBuffer, errChan chan error) {
...
 }
 
 func (d *S3Destination) sendSNSNotification(key string, buffer *s3EventBuffer) error {
-	var err error
-	operation := common.OpLogManager.Start("sendSNSNotification", common.OpLogSNSServiceDim)
-	defer func() {
-		operation.Stop()
-		operation.Log(err,
-			zap.String("topicArn", d.snsTopicArn))
-	}()
```
kostaspap, line 282: I thought it made sense to remove this: 
- The dashboards don't seem to rely on it
- It saves us some $$ - we don't log every time we send an SNS message
- In case of an error, the error will anyway be propagated and will be logged

rleighton (approved):

PR #2684, opened 2021-02-15
Add healthcheck for bucket notifications configuration

giorgosp:
## Background

Add healthcheck for bucket notifications configuration

## Changes

- Add a check that checks the bucket notifications are configured properly for the prefixes the user filled in the onboarded s3 source.
- The healthcheck is only run if user has enabled Panther-managed
bucket notifications

## Testing

- Manual QA:
    - Test with empty prefix in the source, multiple prefixes and overlapping prefixes in the source 
    - Test source with Panther-managed notifications, healthcheck is displayed
    - Test source with manual notification configuration, healthcheck is not displayed

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -305,6 +305,7 @@ type S3LogIntegrationHealth {
   s3BucketStatus: IntegrationItemHealthStatus!
   kmsKeyStatus: IntegrationItemHealthStatus!
   getObjectStatus: IntegrationItemHealthStatus # can be null for sources created in Panther<1.16
+  bucketNotificationsStatus: IntegrationItemHealthStatus
 }
 
 type SqsLogIntegrationHealth {
--- a/api/lambda/source/models/integration.go
+++ b/api/lambda/source/models/integration.go
@@ -195,6 +195,10 @@ type SourceIntegrationHealth struct {
 	KMSKeyStatus         SourceIntegrationItemStatus `json:"kmsKeyStatus,omitempty"`
 	// GetObject check is not available to sources created in Panther<1.16
 	GetObjectStatus *SourceIntegrationItemStatus `json:"getObjectStatus,omitempty"`
+	// BucketNotificationsStatus is the result of checking the bucket's notifications configuration.
+	// It is populated only the log processing role has the s3:GetBucketNotification permission. This is
+	// added to our provided CFN template if user opts for Panther-managed bucket notifications.
+	BucketNotificationsStatus *SourceIntegrationItemStatus `json:"bucketNotificationsStatus,omitempty"`
 
 	// Checks for Sqs integrations
 	SqsStatus SourceIntegrationItemStatus `json:"sqsStatus"`
--- a/deployments/appsync.yml
+++ b/deployments/appsync.yml
@@ -607,6 +607,7 @@ Resources:
         $util.qr($input.put("s3Bucket", $ctx.source.s3Bucket))
         $util.qr($input.put("s3PrefixLogTypes", $ctx.source.s3PrefixLogTypes))
         $util.qr($input.put("kmsKey", $ctx.source.kmsKey))
+        $util.qr($input.put("managedBucketNotifications", $ctx.source.managedBucketNotifications))
         $util.qr($input.put("pantherVersion", $ctx.source.pantherVersion))
         {
           "version" : "2017-02-28",
--- a/internal/core/source_api/api/check_integration.go
+++ b/internal/core/source_api/api/check_integration.go
@@ -20,6 +20,7 @@ package api
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/aws/aws-sdk-go/aws"
@@ -37,6 +38,7 @@ import (
 
 	"github.com/panther-labs/panther/api/lambda/source/models"
 	"github.com/panther-labs/panther/pkg/genericapi"
+	"github.com/panther-labs/panther/pkg/stringset"
 )
 
 const (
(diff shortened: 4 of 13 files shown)


api/graphql/schema.graphql
```diff
@@ -305,6 +305,7 @@ type S3LogIntegrationHealth {
   s3BucketStatus: IntegrationItemHealthStatus!
   kmsKeyStatus: IntegrationItemHealthStatus!
   getObjectStatus: IntegrationItemHealthStatus # can be null for sources created in Panther<1.16
+  bucketNotificationsStatus: IntegrationItemHealthStatus
```
3nvi, line 308: can this be `null` similar to `getObjectStatus`? If so, then this PR is fine FE wise. If it's always going to be present, then we should adjust this by adding a `!` and tweaking the conditional found in the `.tsx` files below 😃

3nvi (commented):
Nice work on that @giorgosp! Props on the FE changes 😸

api/graphql/schema.graphql
giorgosp: Thanks for your review! This field is nullable indeed

api/lambda/source/models/integration.go
```diff
@@ -195,6 +195,10 @@ type SourceIntegrationHealth struct {
 	KMSKeyStatus         SourceIntegrationItemStatus `json:"kmsKeyStatus,omitempty"`
 	// GetObject check is not available to sources created in Panther<1.16
 	GetObjectStatus *SourceIntegrationItemStatus `json:"getObjectStatus,omitempty"`
+	// BucketNotificationsStatus is the result of checking the bucket's notifications configuration.
+	// It is populated only the log processing role has the s3:GetBucketNotification permission. This is
+	// added to our provided CFN template if user opts for Panther-managed bucket notifications.
+	BucketNotificationsStatus *SourceIntegrationItemStatus `json:"bucketNotificationsStatus,omitempty"`
```
kostaspap, line 201: nit: In case of KMS we just return `SourceIntegrationItemStatus` and "healthy" with note ""No KMS Key was specified."". if KMS is not there. 
Maybe we should do the same here? The way I am thinking about it: 
1. It doesn't seem that the code at this point needs to differentiate between the two cases, at least not in the health check.
2. Easier to handle in code - accessing the `BucketNotificationsStatus` always requires a null check first. 

Seems... simpler to avoid the pointer for the time being. What do you think?

internal/core/source_api/api/check_integration.go
```diff
@@ -105,14 +107,91 @@ func (api *API) checkAwsS3Integration(input *models.CheckIntegrationInput) *mode
...
 		Region:      bucketRegion,
 	})
-	h, skipped := checkGetObject(s3Client, input)
+	getObjectCheck, skipped := checkGetObject(s3Client, input)
 	if !skipped {
-		out.GetObjectStatus = &h
+		out.GetObjectStatus = &getObjectCheck
 	}
 
+	notificationsCheck, skipped := checkBucketNotifications(s3Client, input, api.Config, bucketRegion)
```
kostaspap, line 115: nitpick: I think it would be better if this would be `ok` instead of `skipped` and the logic would be reversed (line 116 would be `if !ok ...`). 
This would be more aligned with how Go is using such variables and how readers are used to such values. See e.g. map lookup or how this is being discussed here: https://github.com/golang/go/wiki/CodeReviewComments#in-band-errors

internal/core/source_api/api/check_integration.go
```diff
@@ -105,14 +107,91 @@ func (api *API) checkAwsS3Integration(input *models.CheckIntegrationInput) *mode
...
+	if !skipped {
+		out.BucketNotificationsStatus = &notificationsCheck
+	}
 	return out
 }
 
+func checkBucketNotifications(s3Client s3iface.S3API, input *models.CheckIntegrationInput, config Config, bucketRegion *string) (
```
kostaspap, line 122: Can you pass a `context.Context` object here and make sure that any AWS SDK operation uses the context?
It's a good practice in general, and I'm making a PR in a bit that will switch the sources-api methods to use context so would make my life simpler 😄

internal/core/source_api/api/check_integration.go
```diff
@@ -105,14 +107,91 @@ func (api *API) checkAwsS3Integration(input *models.CheckIntegrationInput) *mode
...
+			Healthy:      false,
+			Message:      "Failed to get bucket notifications",
+			ErrorMessage: err.Error(),
+		}, false
+	}
+
+	topicARN := arn.ARN{
+		Partition: config.AWSPartition, // Note: Assume the onboarded bucket is in the same AWS partition as Panther
+		Service:   "sns",
+		Region:    *bucketRegion,
```
kostaspap, line 144: Can it be `nil`? Do we have to pass the bucketRegion as pointer?

internal/core/source_api/api/check_integration.go
```diff
@@ -105,14 +107,91 @@ func (api *API) checkAwsS3Integration(input *models.CheckIntegrationInput) *mode
...
+		Resource:  "panther-notifications-topic",
+	}.String()
+	// an SNS notification should exist for each one of the prefixes
+	prefixes := reduceNoPrefixStrings(input.S3PrefixLogTypes.S3Prefixes())
+	var notFound []string // keep the prefixes which we didn't find notifications for
+	for _, p := range prefixes {
+		// search the prefix in the configurations
+		ok := false
+		for _, c := range out.TopicConfigurations {
+			zap.S().Warn("topic arn", topicARN, aws.StringValue(c.TopicArn))
```
kostaspap, line 155: Should be removed?

internal/core/source_api/api/check_integration.go
```diff
@@ -105,14 +107,91 @@ func (api *API) checkAwsS3Integration(input *models.CheckIntegrationInput) *mode
...
+			// Check filter rules. The prefix should be equal to p. Missing prefix is also fine if p is empty.
+			// Note we can't validate the suffix. User may have added a suffix to further break down which log files
+			// reach Panther.
+			rulePrefix := ""
+			for _, r := range c.Filter.Key.FilterRules {
+				if strings.ToLower(aws.StringValue(r.Name)) == "prefix" {
+					rulePrefix = aws.StringValue(r.Value)
+				}
+			}
+			ok = rulePrefix == p
```
kostaspap, line 175: you can break after that?

api/lambda/source/models/integration.go
giorgosp: Hmm what would the message be? It can be a bit confusing

kostaspap: For KMS is like this: https://github.com/panther-labs/panther/blob/master/internal/core/source_api/api/check_integration.go#L176-L182

kostaspap: I am mostly worried about the lack of consistency in the API, I don't mind if we do it one way or another. It would be better though to follow the same approach.

internal/core/source_api/api/check_integration.go
giorgosp: My understanding is that `ok` is an alternative to in-band error values, ie return (value struct{}, ok bool) instead of returning a nullable type or -1 (which is valid int).

`ok` would also be confusing here. What would it mean? Would it mean that the healthcheck errored? That it could not be run? That it should not be run? My inclination is that `skipped` is the best name here.

api/lambda/source/models/integration.go
giorgosp: I can't think of a good message though. "Panther-managed notifications have not been configured" could be something similar to the KMS key mesage but seems like a strange redundancy.

internal/core/source_api/api/check_integration.go
giorgosp: Yep added

internal/core/source_api/api/check_integration.go
giorgosp: Was just a convenience because the AWS API uses pointers extensively. Updated the func to accept a value

internal/core/source_api/api/check_integration.go
giorgosp: yep removed

internal/core/source_api/api/check_integration.go
giorgosp: Good catch, updated

internal/core/source_api/api/check_integration.go
```diff
@@ -105,14 +107,91 @@ func (api *API) checkAwsS3Integration(input *models.CheckIntegrationInput) *mode
...
+	if err != nil {
+		return models.SourceIntegrationItemStatus{
+			Healthy:      false,
+			Message:      "Failed to get bucket notifications",
+			ErrorMessage: err.Error(),
+		}, false
+	}
+
+	topicARN := arn.ARN{
+		Partition: config.AWSPartition, // Note: Assume the onboarded bucket is in the same AWS partition as Panther
```
kostaspap, line 142: Yup. There is no cross-talk between partitions so someone couldn't onboard an `aws-cn` bucket to Panther running in `aws`.

giorgosp: Aha, that's good to know. Thanks for confirming

giorgosp:
Added sync-to-release label too

giorgosp:
@alexmylonas could you help me with the failing test here?

kostaspap (approved):

austinbyers:
> @alexmylonas could you help me with the failing test here?

Try "re-running from failed" in CircleCI

api/lambda/source/models/integration.go
```diff
@@ -195,6 +195,10 @@ type SourceIntegrationHealth struct {
 	KMSKeyStatus         SourceIntegrationItemStatus `json:"kmsKeyStatus,omitempty"`
 	// GetObject check is not available to sources created in Panther<1.16
 	GetObjectStatus *SourceIntegrationItemStatus `json:"getObjectStatus,omitempty"`
+	// BucketNotificationsStatus is the result of checking the bucket's notifications configuration.
+	// It is populated only the log processing role has the s3:GetBucketNotification permission. This is
```
austinbyers, line 199: It is populated only _if_ the log processing role...

internal/core/source_api/api/check_integration.go
```diff
@@ -105,14 +108,94 @@ func (api *API) checkAwsS3Integration(input *models.CheckIntegrationInput) *mode
...
+		AccountID: input.AWSAccountID,
+		Resource:  "panther-notifications-topic",
+	}.String()
+	// an SNS notification should exist for each one of the prefixes
+	prefixes := reduceNoPrefixStrings(input.S3PrefixLogTypes.S3Prefixes())
+	var notFound []string // keep the prefixes which we didn't find notifications for
+	for _, p := range prefixes {
+		// search the prefix in the configurations
+		ok := false
+		for _, c := range out.TopicConfigurations {
```
austinbyers, line 156: This double loop is a good candidate for a unit test - it's complex code that doesn't have any API calls

You might want to move one or both loops to separate functions, which would also make it easier to test

austinbyers (approved):

deployments/appsync.yml
```diff
@@ -607,6 +607,7 @@ Resources:
         $util.qr($input.put("s3Bucket", $ctx.source.s3Bucket))
         $util.qr($input.put("s3PrefixLogTypes", $ctx.source.s3PrefixLogTypes))
         $util.qr($input.put("kmsKey", $ctx.source.kmsKey))
+        $util.qr($input.put("managedBucketNotifications", $ctx.source.managedBucketNotifications))
```
alexmylonas, line 610: Need similar for `getObjectStatus`?

PR #2687, opened 2021-02-15
Adds Allow/Deny validation to logschema.ValueSchema

### LINKED ISSUE #2661: Specify allow/deny values for fields in Schemas
# Background

One of the limitations of the Schema-based parsers is that they can only 'match' records based on `required` fields.
There are cases where a specific field has some distinct values that could be used to identify the type of log.
Clients already require this functionality for user-defined schemas so that they can partition events into different tables in the data lake based on their needs.
We also want this for managed schemas so that we can express more parsers as `YAML` without requiring logic in `Go`.

The Salesforce puller would also benefit in being explicit about CSV classifications without relying on column count.

alxarch:
## Background

Closes #2661 

## Changes

- adds `allowValues` and `denyValues` attributes for fields of type `string`

## Testing

- mage test:go


internal/log_analysis/log_processor/logschema/schema.json
```diff
@@ -175,9 +175,28 @@
         },
         "indicators": {
           "type": "array",
+          "minLength": 1,
```
georgepsarakis, line 178: I don't remember much about JSON schema, but does `minLength` apply to arrays?

internal/log_analysis/log_processor/logschema/schema.json
```diff
@@ -175,9 +175,28 @@
...
           "items": {
             "$ref": "#/definitions/indicator"
           }
+        },
+        "allowValues": {
+          "$comment": "Allowed values",
+          "type": "array",
+          "minLength": 1,
```
georgepsarakis, line 186: Same question as above, perhaps `minItems` is the size validator for arrays? Not sure.

internal/log_analysis/log_processor/logschema/reflect_test.go
```diff
@@ -60,3 +62,71 @@ func TestArrayIndicators(t *testing.T) {
...
+	}
+	{
+		x := reflect.New(typ).Interface()
+		assert.NoError(jsoniter.UnmarshalFromString(`{"foo":"Foo|Bar"}`, x))
+		err := validate.Struct(x)
+		assert.NoError(err, "Foo|Bar allowed")
+	}
+	{
+		x := reflect.New(typ).Interface()
+		assert.NoError(jsoniter.UnmarshalFromString(`{"foo":"Foo|Bar"}`, x))
```
georgepsarakis, line 116: Seems as a duplicate test case with the one above. Probably you want to test for some other value here (`Bar,Baz`)?

georgepsarakis (approved):
Very helpful feature!

kostaspap (approved):
LGTM after comments from @georgepsarakis are addressed. 
Make sure to make a PR to update the docs!

internal/log_analysis/log_processor/logschema/logschema.go
```diff
@@ -104,6 +104,8 @@ type ValueSchema struct {
 	Indicators  []string      `json:"indicators,omitempty" yaml:"indicators,omitempty"`
 	TimeFormat  string        `json:"timeFormat,omitempty" yaml:"timeFormat,omitempty"`
 	IsEventTime bool          `json:"isEventTime,omitempty" yaml:"isEventTime,omitempty"`
+	Allow       []string      `json:"allowValues,omitempty" yaml:"allowValues,omitempty"`
```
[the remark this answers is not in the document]
alxarch: I was wondering the same. Some native speaker might be in a position to help here. In my head `deniedValues` are values that _have already been denied_ while `denyValues` means _deny these values_. My preference would just be `whitelist`/`blacklist` but I don't want to be dragged into a PC discussion :D.

internal/log_analysis/log_processor/logschema/reflect.go
```diff
@@ -132,6 +132,49 @@ func buildStructTag(schema *FieldSchema) reflect.StructTag {
 	return reflect.StructTag(tag)
 }
 
+func buildValidate(s *FieldSchema) (rules []string) {
+	// The precedence is Allow > Deny
```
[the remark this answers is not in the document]
alxarch: It would just complicate the JSON Schema even further...

internal/log_analysis/log_processor/logschema/schema.json
alxarch: You this is a mistake

internal/log_analysis/log_processor/logschema/reflect.go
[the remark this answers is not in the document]
alxarch: Tried this, it over-complicates the already complex JSON Schema. 
The only way I found to do this was with nested `allOf` > `oneOf`.
Mutually exclusive properties are not a JSON schema's strong point I think.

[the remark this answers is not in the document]
alxarch: There is no case where _both_ make sense, if we nested the `allow`/`deny` fields it under `validate` it would be easier to enforce this. For example:
```yaml
fields:
- name: Foo
  type: string
  required: true
  validate:
    allow: ["foo", "bar"]
```
I think this might be the best approach also since we might add other validate options and they will be added to a single `oneOf` in the future

georgepsarakis (approved):
👍

internal/log_analysis/log_processor/logschema/schema.json
alxarch: fixed

internal/log_analysis/log_processor/logschema/reflect_test.go
alxarch: fized

internal/log_analysis/log_processor/logschema/logschema.go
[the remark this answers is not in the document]
alxarch: Moved under `validate` key, enforcing schema validity is easier that way and we can extend rules without drastic changes to the schema

Issue #2640, opened 2021-02-10
Rules Engine Should Validate All Functions in Testing


rleighton:
### Describe the bug

The rules engine should validate the return types of executed functions so users know there are issues.

For example, if a user returns a string from `destinations`, then they get no feedback.

### Steps to reproduce

Create a rule with a function like:
```
def destinations(event):
   return 'Some Destination'
```

### Expected behavior

User's should be informed with a nice message during testing that their function does not return the correct type of object.

rleighton:
BTW, the testing API is extremely simple making it very hard to extend. We should change it to have the usual input/output objects that can easily have attributes added as needed w/out changing many interfaces between here and there.

wey-chiang:
Looking at the rules engine: `rule.py:_run_command` function, I think I spotted an issue that went unnoticed:

```
    def _run_command(self, function: Callable, event: PantherEvent, expected_type: Any) -> Any:
        result = function(event)
        # Branch in case of list
        if not isinstance(expected_type, list):
            if not isinstance(result, expected_type):
                raise Exception(
                    'rule [{}] function [{}] returned [{}], expected [{}]'.format(
                        self.rule_id,
                        function.__name__,
                        type(result).__name__,
                        expected_type.__name__,
                    )
                )
        else:
            if not isinstance(expected_type, list) or not all([isinstance(x, (str, bool)) for x in result]):
                raise Exception(
                    'rule [{}] function [{}] returned [{}], expected a list'.format(self.rule_id, function.__name__,
                                                                                    type(result).__name__)
                )
        return result
```

The following line, logically, makes no sense:

`if not isinstance(expected_type, list) or not all([isinstance(x, (str, bool)) for x in result]):`

This should actually be 

`if not isinstance(result, list) or not all([isinstance(x, (str, bool)) for x in result]):`

Need to test to verify but if that is indeed the root issue, I'll include the fix in the destinations lookup work

wey-chiang:
Yeah, looks like once I make that change the UI will yield:

> Exception: rule [RuleAPITestRule] function [destinations] returned [str], expected a list

Issue #1424, opened 2020-08-24
Automatically lookup destinations in bulk upload


nhakmiller:
### Describe the ideal solution

The new destination override feature allows rules to specify a specific destination to send alerts to, as opposed to the default destinations of that rule's severity. When specifying this destination override, the backend stores the unique ID of the destination which is not visible to users. This means that this field is virtually impossible to manage from the CLI.

The `BulkUpload` endpoint of the analysis API should automatically translate any "friendly name"s of destinations into the corresponding unique destination ID before writing the policies & rules to the backend.

### Describe your use cases

Managing policies & rules from the CLI using the panther_analysis_tool.

### How frequently would you use such feature

Every single time I upload policies & rules from the CLI, probably about once every other week.

### Describe alternatives you have considered

One alternative would be to have the CLI lookup this information and do the mapping itself. Another would be to have the unique ID of each destination visible in the UI, and then users could copy it into their rule specification files.

TravisGary:
This would be a big help for us too.  We have only some alerts we want to send to pagerduty and those overrides get overwritten when we update from our CD process.

wey-chiang:
Issue cloned.

wey-chiang:
The implementation is included in #2699 

Note: rather than overwriting the destination names, we simply allow users to use the display name of the destination. This removes the need to overwrite any python code making it much easier to maintain.

Furthermore, the aforementioned implementation will also provide users refined feedback for invalid destinations returned by `destinations`. That is to say that all invalid/duplicate destinations returned by `destinations` will be shown in a `ValueError` error response in the unit test.

PR #2699, opened 2021-02-17
Weyland destinations lookup

wey-chiang:
## Background

Addresses two issues: #2640 and #1424 

The original generated alert fields implementation required users to use a destination's UUID. These changes allow users to use the destination display name in the `destinations` function.

There was also a logical flaw introduced in the rules engine `Rule:_run_command` function in which lists were not being properly validated -- the fix for that issue is included in these changes.

## Changes

- Added another conditional to the alert-delivery api for dynamic destinations to support display names.
- Refactored alert-forwarder API to use the rule OutputIDs for the alert OutputIDs and the alert dedup destinations for the alert destinations to be more aligned with the changes to the alert-deliver API
- Fixed logic in Rules Engine `Rule:_run_command` 

## Testing

- `mage test:ci`
- Manual deployment into `us-east-1`
  - Created rule that had `destinations` defined as follows
```
def destinations(event):
    return ["Weyland-Email-Test"]
```
  - Confirmed that the generated alert was properly routed to the destination 
  - Confirmed I did get an email for this destination (the dest. is an AWS SNS email subscription)


api/lambda/delivery/models/api.go
```diff
@@ -162,7 +162,7 @@ type Alert struct {
 	Reference string `json:"reference,omitempty"`
 
 	// Destinations is the custom user-provided destination override
-	Destinations []string `json:"destinations,omitempty" validate:"dive,uuid4"`
+	Destinations []string `json:"destinations,omitempty"`
```
kostaspap, line 165: So the destinations can be either UUID or a Name? This seems very hacky API design. The API should be making clear what's included in it, if it's names or IDs, we cannot do guesswork on the content ^^ 

One way to address this is to have the rules engine due to name-> id resolution. When the rules engine loads the rules, it can also load the list of available outputs. It keeps a mapping of Names -> IDs. 
It can then resolve the Name to an ID and that's it. We don't have to change anything downstream. 

It also seems that we should change the public API of the rule override functions so that users know to use the names instead of UUIDs.

kostaspap (changes requested):

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -505,7 +505,7 @@ def _run_command(self, function: Callable, event: PantherEvent, expected_type: A
                     )
                 )
         else:
-            if not isinstance(expected_type, list) or not all([isinstance(x, (str, bool)) for x in result]):
+            if not isinstance(result, list) or not all([isinstance(x, (str, bool)) for x in result]):
```
kostaspap, line 508: Can you add a unit test that: a) Fails with old code b) Passes with new code, to make sure that the issue is fixed?

kostaspap: Ah, nvm. I see that the `test_rule_with_invalid_destinations_type` is meant to do that :0

georgepsarakis, line 508: It might help discern the different type checks if you use identity check for the expected type (Line 497):

```python
if expected_type is list:
    # check that result is instance of list and each element is either boolean or string
else:
    # check that result is instance of expected type
```

This would also actually allow passing multiple acceptable types in `expected_type`, which could be an iterable as well, since `isinstance` accepts that.

api/lambda/delivery/models/api.go
wey-chiang: I was hesitant to include it in the rules engine due to potential performance concerns.

I like the idea however, and it would be more useful for testing logic i.e. checking if passed in destinations are actual existing destinations.

kostaspap: > I was hesitant to include it in the rules engine due to potential performance concerns.
> 
> I like the idea however, and it would be more useful for testing logic i.e. checking if passed in destinations are actual existing destinations.

Note that there is already some I/O when we load the rules and the data models. I think  you could follow the same approach for outputs as well. 
See for example how we keep a cache of Rules: https://github.com/panther-labs/panther/blob/a4adb0e7498d89bf1cc6d1ae0138278e745f87d7/internal/log_analysis/rules_engine/src/engine.py#L43
and how we populate it when the `populate_rules()` is called: https://github.com/panther-labs/panther/blob/a4adb0e7498d89bf1cc6d1ae0138278e745f87d7/internal/log_analysis/rules_engine/src/engine.py#L160

I think you could follow the same approach for outputs. You could have a mapping of output name -> output id in memory and just pass this mapping to the rule or to the rule run method.

nhakmiller: Yeah I think usually when we do "remappings" to make life easier for users, we try to do it as early as possible and not propagate that complexity through the pipeline. For example the `analysis-api` handles a lot of this work when new rules/policies are being created from `BulkUpload` so that the rest of the system doesn't have to think about it. I think a similar approach here would be good, having the rules-engine generate a lookup table one time and then referencing that as needed.

wey-chiang:
Changed the implementation to have the functionality restricted to the rules-engine and reverted downstream changes.

Since we've moved this into the rules-engine, we can now give better response to users as to invalid destinations, I've included some examples :) 

image

image

internal/log_analysis/rules_engine/src/outputs_api.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self) -> None:
+        self.client = boto3.client('lambda')
+
+    def get_outputs(self) -> List[Dict[str, Any]]:
+        """Gets information for all configured destinations."""
+        get_input: Dict[str, Any] = {'getOutputs': {}}
+
+        response = self.client.invoke(FunctionName='panther-outputs-api', Payload=json.dumps(get_input).encode('utf-8'))
+        gateway_response = json.loads(response['Payload'].read())
```
nhakmiller, line 34: Strictly speaking I believe this is a straight up lambda response, not an API gateway response. I would recommend re-naming this variable to `lambda_response`.

wey-chiang:
Some more SS from testing:

Note how there is no destination configured (destination set dynamically)

image

The Alerts page properly shows the destination now

image
image

Email confirming the destinations are working as intended

image

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -364,14 +365,38 @@ def _get_destinations(self, event: PantherEvent, use_default_on_exception: bool
                 return []
             raise
 
-        if len(destinations) > MAX_DESTINATIONS_SIZE:
+        invalid_destinations = []
+        # Map display names to destinations
+        outputs_display_names = {v.destination_display_name: v for k, v in outputs.items()} \
```
nhakmiller, line 370: It seems this mapping/list could be calculated once when the destinations are initially loaded are re-used instead of being calculated for each alert.

nhakmiller (approved):
A bit concerned about the continually growing permissions of the rules-engine, but I don't think that's in scope for this PR.

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: It's ultimately the difference between adding two parameters versus adding one -- I decided on the side of simplicity at the cost of a few more cycles.

While we could just pack those into a list to only have one new parameter added, I figured this implementation would be more easily readable/understood

internal/log_analysis/rules_engine/src/outputs_api.py
wey-chiang: Makes sense, renamed it accordingly.

internal/log_analysis/rules_engine/src/rule.py
kostaspap: I agree with Nick, we should do it only once. This code will run over billions of events per day so it is performance sensitive code. 
It can still be readable though, I don't think we are facing such a dilemma here. You could just pass the outputs e.g. in the rule constructor. Then the rule creates the following:
1. One map of name -> UUID
2. One set of UUIDs. 

When the rule runs the `destinations` function, it checks every item in the returned list. For each item, look up the map. If the lookup fails, check if the value exists in the set. If only few of the destinations defined by the user exist, we should just send notifications to those. If none of the destinations defined by the user exist, we just send to the default outputs.

kostaspap (commented):
This PR has the label `sync-to-release`. This doesn't seem right. The 1.16-RC is already out and has been tested. Adding this should require a new round of tests to make sure there is no regression in the way the rules work. My opinion is to remove the label. We can make a 1.16.1 if someone needs that urgently.

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: I've moved the `outputs_display_names` into the engine initialization so it is only run once and initialized the same time as the destinations are populated.

wey-chiang:
Huh, CircleCI is failing because

> # Run  npm install aws-amplify@3.3.20  to resolve 1 vulnerability

Will include that package update in this PR

wey-chiang:
After updating `aws-amplify` it looks like `npm-audit` is still showing 9 high vulns, would updating these libraries to a newer version break anything?@nhakmiller 

If not, I'll include those in the PR as well

nhakmiller:
> After updating `aws-amplify` it looks like `npm-audit` is still showing 9 high vulns, would updating these libraries to a newer version break anything?@nhakmiller
> 
> If not, I'll include those in the PR as well

I'd say go for it. But now the bundle size check is failing, not sure what's causing that.

wey-chiang:
Ran into an issue testing which revealed an issue that was not caught by unit tests. 
I've added some more unit tests to better test the feature.

wey-chiang:
Removed the npm package upgrades

@kostaspap those will be blocking future CI tests and need to be addressed -- I tried to upgrade them but upgrading all of the packages presenting high vulns breaks deployment

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -364,14 +370,36 @@ def _get_destinations(self, event: PantherEvent, use_default_on_exception: bool
...
+            # case for valid UUIDv4
+            elif each_destination in outputs and each_destination not in standardized_destinations:
+                standardized_destinations.append(each_destination)
+            else:
+                invalid_destinations.append(each_destination)
+
+        if invalid_destinations:
+            if use_default_on_exception:
+                self.logger.warning('destinations method yielded invalid destinations: %s', str(invalid_destinations))
+                return []
```
kostaspap, line 392: Shouldn't we return None here? From my understanding, if we return `[]` then we will just not send notification to any destination.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -364,14 +370,36 @@ def _get_destinations(self, event: PantherEvent, use_default_on_exception: bool
...
+                standardized_destinations.append(outputs_display_names[each_destination].destination_id)
+            # case for valid UUIDv4
+            elif each_destination in outputs and each_destination not in standardized_destinations:
+                standardized_destinations.append(each_destination)
+            else:
+                invalid_destinations.append(each_destination)
+
+        if invalid_destinations:
+            if use_default_on_exception:
+                self.logger.warning('destinations method yielded invalid destinations: %s', str(invalid_destinations))
```
kostaspap, line 391: I think this should be just `info` level. We don't want to get paged for this

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -364,14 +370,36 @@ def _get_destinations(self, event: PantherEvent, use_default_on_exception: bool
...
+                return []
+            raise ValueError('Invalid Destinations: {}'.format(str(invalid_destinations)))
+
+        if len(standardized_destinations) > MAX_DESTINATIONS_SIZE:
             # If generated field exceeds max size, truncate it
             self.logger.warning(
```
kostaspap, line 397: Same here, i think this should be info level

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -364,14 +370,36 @@ def _get_destinations(self, event: PantherEvent, use_default_on_exception: bool
...
                 'maximum len of destinations [%d] for rule with ID [%s] is [%d] fields. Truncating.', MAX_DESTINATIONS_SIZE, self.rule_id,
-                len(destinations)
+                len(standardized_destinations)
             )
-            return destinations[:MAX_DESTINATIONS_SIZE]
-        return destinations
+            return standardized_destinations[:MAX_DESTINATIONS_SIZE]
```
kostaspap, line 401: Have we documented these limits?

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -364,14 +370,36 @@ def _get_destinations(self, event: PantherEvent, use_default_on_exception: bool
                 return []
             raise
 
-        if len(destinations) > MAX_DESTINATIONS_SIZE:
+        # Check for (in)valid destinations
+        invalid_destinations = []
+        standardized_destinations = []
```
kostaspap, line 375: We want them to be unique, right? This could be a set. it could simplify the logic below as well.

kostaspap:
> Removed the npm package upgrades
> 
> @kostaspap those will be blocking future CI tests and need to be addressed -- I tried to upgrade them but upgrading all of the packages presenting high vulns breaks deployment

The FE team is responsible for updating those ones - @3nvi  I think is already aware of it.

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: Huh... yeah 👍

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: No, I'll queue that up to mention that in the `Alert Field Overrides` doc

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: Sure, yeah they could be a set but I don't see how that would simplify the logic below.

internal/log_analysis/rules_engine/src/rule.py
wey-chiang: I'll make those changes for the rest of the generated alert fields as well then.

internal/log_analysis/rules_engine/src/engine.py
```diff
@@ -24,26 +24,32 @@
...
+class Engine:  # pylint: disable=too-many-instance-attributes
     """The engine that runs Python rules."""
 
-    def __init__(self, analysis_api: AnalysisAPIClient) -> None:
+    def __init__(self, analysis_api: AnalysisAPIClient, outputs_api: OutputsAPIClient) -> None:
         self.logger = get_logger()
         self._last_update = datetime.utcfromtimestamp(0)
         self.log_type_to_data_models: Dict[str, DataModel] = collections.defaultdict()
         self.log_type_to_rules: Dict[str, List[Rule]] = collections.defaultdict(list)
+        self.destinations: Dict[str, Destination] = collections.defaultdict()
```
kostaspap, line 46: Does this contain a mapping of id -> destination? Could you rename it to `id_to_destination` so that the name follows the same pattern as below?

internal/log_analysis/rules_engine/src/rule.py
kostaspap: It can be rewritten like the following:
```
for dest in destinations:
    if dest in outputs_display_names: # this means the value returned by the destination function was a destination name
        standardized_destinations.add(outputs_display_names[dest].destination_id)   
     elif dest in outputs: # this means the value returned by the destination function was a destination id. 
        standardized_destinations.add(dest)
     else:
         invalid_destinations.add(dest)
```
Sets are ideal for deduplication like the one you are doing here.

internal/log_analysis/rules_engine/src/rule.py
```diff
@@ -351,7 +353,11 @@ def _get_description(self, event: PantherEvent, use_default_on_exception: bool =
             return description[:num_characters_to_keep] + TRUNCATED_STRING_SUFFIX
         return description
 
-    def _get_destinations(self, event: PantherEvent, use_default_on_exception: bool = True) -> Optional[List[str]]:
+    def _get_destinations(self,
+                          event: PantherEvent,
+                          outputs: dict,
```
kostaspap, line 358: nit: The `outputs` are `output_ids` right? Might be good to rename, makes it easier for the reader.

PR #2703, opened 2021-02-17
Make sure we don't serialize invalid JSON

kostaspap:
## Background

Fixes https://github.com/panther-labs/panther-enterprise/issues/2286

## Changes

- Avoid serializing invalid JSON if we have an empty event

## Testing

- mage test:go


internal/log_analysis/log_processor/pantherlog/jsoniter.go
```diff
@@ -92,6 +92,11 @@ func (e *resultEncoder) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
 	att := stream.Attachment
 	stream.Attachment = result
 	stream.WriteVal(result.Event)
+	// If the result.Event has no serializable content, it will be serialized to `{}`
```
alxarch, line 95: If I understand correctly this tries to check `{}`, `[]` and `""` in one go.
If `result.Event` is `nil` it might be `null`.
The check here is hard to get right for all cases that's why the previous code _only_ tried to detect `}` as last byte.
My suggestion would be to modify what happens around `extendJSON`.
The 'real' bug is that `extendJSON` does not check if it is putting the comma `,` at the correct position.


By Resetting _here_, you rely on `extendJSON` doing the right thing with an empty buffer.
I would suggest we change `extendJSON` so that this relation between the two points in code does not cause more bugs in the future.
This could be trivially checked checking the value of `n`.
The check `0 <= n && n < len(data)` can become `2 <= n && n < len(data)`.
If `extendJSON` returns `false` we do `Reset` and `WriteObjectStart` at the same place so that there's no dependencies between the two points in code.

alxarch (changes requested):

internal/log_analysis/log_processor/pantherlog/jsoniter.go
kostaspap: I don't think it was possible to have `null` ... but anyway I've updated the code to follow this approach. Note that `extendJSON` was not resetting the buffer, added it as part of the PR.

internal/log_analysis/log_processor/pantherlog/jsoniter.go
```diff
@@ -167,7 +168,8 @@ func (*resultEncoder) writePantherFields(r *Result, stream *jsoniter.Stream) {
 
 func extendJSON(data []byte) bool {
 	// Swap JSON object closing brace ('}') with comma (',') to extend the object
-	if n := len(data) - 1; 0 <= n && n < len(data) && data[n] == '}' {
+	// Don't try to do the swap for empty JSON `{}`
+	if n := len(data) - 1; 2 <= n && n < len(data) && data[n] == '}' {
```
georgepsarakis, line 172: Isn't `n < len(data)` always true?

georgepsarakis: I can't think of a case that you can have valid JSON object with 3 characters. I think the minimum is 6: `{"a":1}`, but not sure if we need to handle this further here.

internal/log_analysis/log_processor/pantherlog/jsoniter.go
```diff
@@ -167,7 +168,8 @@ func (*resultEncoder) writePantherFields(r *Result, stream *jsoniter.Stream) {
 
 func extendJSON(data []byte) bool {
 	// Swap JSON object closing brace ('}') with comma (',') to extend the object
-	if n := len(data) - 1; 0 <= n && n < len(data) && data[n] == '}' {
+	// Don't try to do the swap for empty JSON `{}`
```
giorgosp, line 171: Why not? Can't `{}` be extended?

kostaspap: > Why not? Can't `{}` be extended?

What the code does is that it replaces the last `}` of a JSON with `,` so that it can be "extended". 
An empty JSON cannot be extended (we end up with `{,` ), it needs to be ... recreated.

internal/log_analysis/log_processor/pantherlog/jsoniter.go
alxarch: > Isn't `n < len(data)` always true?

Yes, but having the check explicit in code avoids the runtime check imposed by the go compiler when you do index operations below.
This effectively allows the function to be _inlined_.
We're at a very performance sensitive part of the code here. This operation is running for _every_ event written to Panther.
The function was carefully designed to 'spartan' standards.
For more information see

https://www.ardanlabs.com/blog/2018/04/bounds-check-elimination-in-go.html
https://dave.cheney.net/2020/04/25/inlining-optimisations-in-go

georgepsarakis: Aha, nice thanks for the thorough explanation!

alxarch: > hat you can have valid JSON object with 3 characters. I think the minimum is 6: `{"a":1}`, but not sure if we need to handle this further here.

The actual _valid_ minimum is `len(s) == 6`:
```json
{"":1}
```

internal/log_analysis/log_processor/pantherlog/jsoniter.go
alxarch: We could replace the `,` with ` ` (space) and strill return true but this can be handled with the `Reset()` + `WriteObjectStart` in   the caller. We want this function inlinable.

alxarch (approved):
LGTM

internal/log_analysis/log_processor/pantherlog/jsoniter.go
kostaspap: > Isn't `n < len(data)` always true?

I will update the code with @alxarch comments explaining why we want the check in there.

giorgosp: Thanks Kostas I was about to suggest it

PR #2710, opened 2021-02-19
Managed notifications multiple sources same bucket

giorgosp:
## Background

The Panther notifications feature didn't really account for multiple s3 sources for the same bucket. This scenario
would make the feature unusable if some s3 sources would have the same bucket and overlapping prefixes.

Although this might not seem a common scenario, it is currently allowed and we should support it.

## Changes

- The code that configures bucket notifications now takes into account the s3 prefixes for all the saved sources with the same bucket.
- Don't save the created notifications or the created SNS topic in the db anymore. Assume that a notification is managed by Panther if its name starts with `panther-managed-`
- The healthcheck now becomes green if the prefix defined for a notification is a string prefix of the source's prefix (i.e. we don't require exact equality)

## Testing

- Preexisting unit tests
- Manual QA:
    - Create 3 sources with overlapping prefixes and check bucket notifications
    - Add more prefixes to a source, check bucket notifications
    - Delete the sources one by one, checking bucket notifications in the process

Closes https://app.asana.com/0/1199961111326839/1199961165787232

### DIFF
--- a/internal/core/source_api/api/managed_bucket_notifications.go
+++ b/internal/core/source_api/api/managed_bucket_notifications.go
@@ -19,6 +19,7 @@ package api
  */
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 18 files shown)


internal/core/source_api/api/utils.go
```diff
@@ -77,61 +76,6 @@ func integrationToItem(input *models.SourceIntegration) *ddb.Integration {
 	return item
 }
 
-func itemToIntegration(item *ddb.Integration) *models.SourceIntegration {
```
giorgosp, line 80: Moved to `ddb` package. This converts the db representation of a source to our "domain model" `models.SourceIntegration`. It doesn't need to be exported if we encapsulate our db CRUD operations in functions like 
 ```
func (ddb *DDB) ListS3SourcesWithBucket(bucket string) ([]models.SourceIntegration, error)
```

File: internal/core/source_api/api/managed_bucket_notifications.go [hunk 2]
```diff
@@ -34,124 +34,123 @@ import (
...
+func ManageBucketNotifications(dbClient *ddb.DDB, panther pantherDeployment, source *models.SourceIntegration) error {
+	// Take the prefixes from all sources with this bucket into account.
+	sources, err := dbClient.ListS3SourcesWithBucket(source.S3Bucket)
+	if err != nil {
+		return errors.Wrap(err, "failed to fetch sources from db")
+	}
+	var prefixes []string
+	for _, dbSource := range sources {
+		if dbSource.IntegrationID == source.IntegrationID {
+			// Use source instead of dbSource as the source of truth to account for updates.
```
austinbyers, line 81: What's the difference? I assume "source" is the user-provided input. So maybe clarify the comment to something like

```go
// User input is source of truth, which may be an update of the existing dbSource
```

internal/core/source_api/main.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/pkg/errors"
+)
+
+type strct struct{}
+
+func (a *strct) val() {}
+func (a *strct) po()  {}
+
+func main() {
+	a, err := arn.Parse("arn:aws:iam::415773754570:role/PantherLogProcessingRole-kostas-giorgosp-test")
```
austinbyers, line 39: This whole file looks like just a lot of debugging?

internal/core/source_api/api/utils.go
alxarch: Such interaction should be implementing some common interface that would only depend on `models`. That way we can have different implementation of the backend store. 
It would be wise to be passing `context.Context` on each such 'action' method.

internal/core/source_api/main.go
alxarch: @giorgosp the best way to handle such 'script' tests would be to add a [build tag](https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool) at the top of the file.
To make sure it is never included in compilation units.

For example
```golang
// +build ignore

package main
```
Note that the line separating the comment and `package` is *required*.

alxarch: Another approach for this, if you don't want to test CLI interaction and only code is to stuff the logic in a _failing_ test.
You need it to be failing so you can see all those `fmt.Prinln` output.
You can then toggl it with `test.Skip()`

alxarch: There's also https://repl.it/ which allows to import any public package

internal/core/source_api/ddb/scan.go
```diff
@@ -54,3 +56,92 @@ func (ddb *DDB) ScanIntegrations(integrationType *string) ([]*Integration, error
...
+
+	result := make([]models.SourceIntegration, 0, len(integrations))
+	for _, item := range integrations {
+		integ := ItemToIntegration(item)
+		result = append(result, *integ)
+	}
+	return result, nil
+}
+
+// Deprecated. This should not be exported but only be used by functions that directly interact with DynamoDB,
```
alxarch, line 92: I think you can use an `internal/` directory inside `source_api` for this.
To make sure that the exported stuff is package-level only.

internal/core/source_api/main.go
giorgosp: Yep thank you guys! Commited this by mistake

internal/core/source_api/ddb/scan.go
giorgosp: I see. This still needs to be exported because it is used in some other places. When the refactoring is complete we can unexport it.  Until then, I hope `// Deprecated.` will discourage new uses of this func outside this package.

internal/core/source_api/api/managed_bucket_notifications.go
giorgosp: Yes "source" is the user's input. Updated the comment!

internal/core/source_api/api/utils.go
giorgosp: Yep eventually we should have an interface around db operations for sources. Updated to pass context.

internal/core/source_api/api/managed_bucket_notifications.go
```diff
@@ -252,23 +262,35 @@ func updateBucketTopicConfigurations(s3Client *s3.S3, bucket, bucketOwner string
...
+		return strings.HasPrefix(aws.StringValue(c.Id), namePrefix)
+	}
+
 	added := make(map[string]struct{})
 	for _, c := range bucketTopicConfigs {
-		if stringset.Contains(managedConfigIDs, *c.Id) {
-			// Panther-created. Keep it only if its prefix is included in prefixes.
-			pref := prefixFromFilterRules(c.Filter.Key.FilterRules)
-			if pref != nil && stringset.Contains(prefixes, *pref) {
+		if isPantherCreated(c) {
```
austinbyers, line 283: Nit: instead of having such a large block under this condition, invert the logic and exit early. E.g.

```go
for _, c :- range bucketTopicConfigs {
    if !isPantherCreated(c) {
        // User-created, keep it
       newConfigs = append(newConfigs, c)
       continue
    }

   // remaining code: c is panther created
   pref, ok := prefixFromFilterRules(....)
```

austinbyers (approved):

internal/core/source_api/api/managed_bucket_notifications.go
giorgosp: Updated and merging, thanks for you review!

PR #2712, opened 2021-02-20
Ignore some verbose logs

nhakmiller:
## Background

A customer reported 250,000+ scans  /  minute and their entire Cloud Security pipeline was getting throttled. It turns out they were logging s3 object events, which were getting processed by the `aws-event-processor`, which was initiating a scan for every `PutObjectAcl` API call. Which was happening 250,000 times per minute.

The customer deleted the source, but the scanning continued. This revealed a second issue, that we were never actually clearing out the accounts cache in the `aws-event-processor`. We were checking the cache age, and had a function to clear the cache, we just weren't calling. This meant the cache age was essentially however long AWS kept the lambda alive.

Because the code is so self-contained, and its currently preventing a customer from using Panther, I added a `sync-to-release` label. If we think it's too late for that, I can remove it.

## Changes

- Added more whitelisted events for s3
- Clear the cache if its expired 

## Testing

- Did a test deployment to make sure nothing broke


austinbyers (approved):
A fantastic PR description, as always. Definitely something we want in the release 🎉

internal/compliance/aws_event_processor/processor/process.go
```diff
@@ -252,12 +252,25 @@ var (
 		"CreateClusterParameterGroup":       {},
 
 		// s3
-		"UploadPart":              {},
-		"CreateMultipartUpload":   {},
-		"CompleteMultipartUpload": {},
-		"HeadBucket":              {},
-		"HeadObject":              {},
-		"PutObject":               {},
+		"AbortMultipartUpload":       {},
```
kostaspap, line 255: Are these actions that need to be ignored? 
What about all the Get or List operations? Or do we implicitly ignore them?

kostaspap (approved):
I went through this: https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations.html 
Changes look good!

internal/compliance/aws_event_processor/processor/process.go
nhakmiller: Correct, all `Get*`, `List*`, and `Describe*` calls are ignored for all services at all times. I actually copied in every single API in the same list you shared, then manually deleted the `Get`, `List`, `Describe`, and the actually relevant operations that we do care aboutl.

PR #2720, opened 2021-02-26
Fix bug in cloud security scanning when a region ignore list is configured

nhakmiller:
## Background

@s0l0ist reported a bug found in dogfooding/release testing. We were actually tracking this bug already, but did not realize it would affect the release so strongly.

The bug is that if a user configures a region ignore list for a cloud security scan, we will fail to scan all cloudtrails, guard dutys, and config recorders in the account (and log an error) because we use the same region deny list code for generating generic clients. For the other scanners this situation gets handled higher up in the stack, but these three scanners behave slightly differently because they scan all regions for each scan, not just the region requested.

I'm tagging `sync-to-release` since this will affect every user of cloud security that configures region ignore lists, and @s0l0ist said it should be in 1.16.

## Changes

- Added region denied error handling to cloudtrail, guardduty, and config polllers

## Testing

- Dev deployment going up now


s0l0ist (approved):

nhakmiller:
Tested & verified in dev account.

PR #2722, opened 2021-03-02
Release 1.16 partition migration for the datacatalog-updater

rleighton:
## Background

Release 1.16 adds a new partition column to the tables in Glue.

After the upgrade fro 1.15 we need to immediately update all the schemas , which the deployment software does do BUT in the case of an environment under load it is does not work. 

The reason it does not work is we send an event to sqs to trigger the update. If the q is empty it works (which is why previous testing succeeded). Under load what happens is we have other events, which get processed, which FAIL due to the schema not matching and the lambda aborts , never reading the message that would trigger the needed update.

I added a function to the datacatalog-updater to execute BEFORE processing any S3 events to ensure all tables are updated with the new partition column.

This will run once per container instantiation, testing shows this to be about 2 times per hour.

This code can be removed after we are confident all customers have upgrade to 1.16+

Closes: https://app.asana.com/0/1199942399745175/1200002746484073/f

## Changes

- add function `partitionColumnMigration()` to datacatalog-updater.

## Testing

- mage test:go
- mage deploy (fresh)
- mage deploy release1.15 , then this branch, checked for no errors.


s0l0ist (approved):

alxarch (changes requested):
I think we need to move the call inside the S3 handler code, behind a `sync.Once` guard.

internal/log_analysis/datacatalog_updater/main/lambda.go
```diff
@@ -124,5 +127,58 @@ func main() {
 		Logger:       logger,
 	}
 
+	// FIXME: This can be removed in a few releases after 1.16
+	partitionColumnMigration(&handler, clientsSession)
```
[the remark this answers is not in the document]
rleighton: The first concern makes sense, moving it with a sync.Once() is a good idea. The other concerns exist regardless where this ugly code goes.

internal/log_analysis/datacatalog_updater/main/lambda.go
```diff
@@ -124,5 +127,58 @@ func main() {
...
 	lambda.StartHandler(&handler)
 }
+
+// FIXME: This can be removed in a few releases after 1.16 after everyone has upgraded.
+// FIXME: Release 1.16 adds a new partition column to the tables in Glue.
+// FiXME: The below needs to execute BEFORE processing any S3 events to ensure
+// FIXME: all tables are updated with the new partition column.
+// FIXME: This will run once per container instantiation, testing shows this to be about 2 times per hour.
+// partitionColumnMigration updates schemas if they have not had the new partition added, best effort
+func partitionColumnMigration(handler *datacatalog.LambdaHandler, clientSession *session.Session) {
```
kostaspap, line 142: What about have the custom resource update the tables - let this Lambda handle only syncing the partitions. Wouldn't that make more sense?

internal/log_analysis/datacatalog_updater/main/lambda.go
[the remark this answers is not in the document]
kostaspap: `sync.Once` will not do much anyway. The code runs once per Lambda initialization. When a Lambda is initialized, so are all the global vars.

internal/log_analysis/datacatalog_updater/main/lambda.go
alxarch: The problem is that data does not stop flowing in during deployment on a busy system. The add partition code might fail _during_ a deployment if the Glue table storage descriptor has not been updated.

internal/log_analysis/datacatalog_updater/main/lambda.go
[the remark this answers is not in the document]
alxarch: Don't overlook #2, any panic would prevent the lambda from starting _ever_.

internal/log_analysis/datacatalog_updater/main/lambda.go
rleighton: Once the deploy finishes and this starts processing S3 events (which happens immediately) we start throwing errors until the tables have the schema updated to have the new partition key.

We need to prevent this from running until we know the tables have been updated. The only non-disruptive way I could think of doing this is directly in the updater.

I thought of sorting the messages but that is not 100%, if the batch has no sync message, you will get alarms. We don't want oncall paged for this , even if is corrects itself.

rleighton: The other idea I had, which is too much work, but if we stopped log processing , updated, then started log processing this could work. Long term I think our deploys should do this, but there is a lot of orchestration needed to do this safely.

kostaspap: I was considering that there would be failures while this is running, but eventually the system would sort itself out - the messages get retried 4 times with 15' backoff. There will be pages to the oncall but they should only happen during a 1.15-1.16 migration and never again ^^

Again, this is high-priority item since it is meant for 1.16 so I don't want to block this PR now. But we need to revisit this.

alxarch (approved):

PR #2730, opened 2021-03-03
Allow CloudTrail notifications to process

nhakmiller:
## References

Asana Task URL: https://app.asana.com/0/1199942399745175/1200013141331018

## Background

A recent PR has us ignore s3 event notifications if they have a file size of zero. These files cause us to log errors during processing, and there's no value in processing an empty file anyways so it saves us time and pages.

For CloudTrail logs, instead of configuring S3 event notifications to send to SNS when new objects are written to S3 you can configure the CloudTrail service itself to directly deliver notifications of new data to SNS. This is for some reason I can't remember better than just relying on s3 event notifications directly (more efficient maybe because it groups the s3 keys?). When we process CloudTrail -> SNS notifications, we internally convert them to look like an S3 event notification. When we do this conversion, we don't specify a file size. This was causing all CloudTrail notifications to be discarded as empty files. This PR simply changes the object size to 1 during this conversion so they don't get discarded.

I'm not sure if this will have downstream impacts on processing, where we made certain assumptions if the file size was 0 or if the file size is used in some other calculations somewhere when its non-zero. If so, we could instead add a `IsCloudTrailNotification` flag to our internal s3 event notification struct and use that to not exclude empty s3 objects when set. This would be more explicit, and with less lies. But it would also mean a bigger code change.

## Changes

- Change CloudTrail notification object sizes to 1

## Testing

- Unit tests
- I intend to do a fresh deploy into my dev account without this fix, onboard a CloudTrail s3 source that uses CloudTrail to publish notifications, deploy this update, and verify it's working E2E. This is in progress, I don't want to merge this PR until that's been completed (had to upgrade go to 1.16 so I'm still waiting for that to complete)

### DIFF
--- a/internal/log_analysis/log_processor/common/common.go
+++ b/internal/log_analysis/log_processor/common/common.go
@@ -86,9 +86,9 @@ func Setup() {
 
 // DataStream represents a data stream for an s3 object read by the processor
 type DataStream struct {
-	Stream       logstream.Stream
-	Closer       io.Closer
-	Source       *models.SourceIntegration
-	S3ObjectKey  string
-	S3Bucket     string
+	Stream      logstream.Stream
+	Closer      io.Closer
+	Source      *models.SourceIntegration
+	S3ObjectKey string
+	S3Bucket    string
 }
--- a/internal/log_analysis/log_processor/common/common.go
+++ b/internal/log_analysis/log_processor/common/common.go
@@ -91,5 +91,4 @@ type DataStream struct {
 	Source       *models.SourceIntegration
 	S3ObjectKey  string
 	S3Bucket     string
-	S3ObjectSize int64
 }
--- a/internal/log_analysis/log_processor/processor/processor.go
+++ b/internal/log_analysis/log_processor/processor/processor.go
@@ -212,7 +212,6 @@ func (p *Processor) run(ctx context.Context, outputChan chan<- *parsers.Result) 
 			// s3 dim info
 			zap.String("bucket", p.input.S3Bucket),
 			zap.String("key", p.input.S3ObjectKey),
-			zap.Int64("size", p.input.S3ObjectSize),
 			zap.String("sourceID", p.input.Source.IntegrationID),
 		)
 	}()
--- a/internal/log_analysis/log_processor/sources/s3.go
+++ b/internal/log_analysis/log_processor/sources/s3.go
@@ -226,8 +226,9 @@ func parseCloudTrailNotification(message string) (result []*S3ObjectInfo) {
 
 	for _, s3Key := range cloudTrailNotification.S3ObjectKey {
 		info := &S3ObjectInfo{
-			S3Bucket:    *cloudTrailNotification.S3Bucket,
-			S3ObjectKey: *s3Key,
+			S3Bucket:     *cloudTrailNotification.S3Bucket,
+			S3ObjectKey:  *s3Key,
+			S3ObjectSize: 1,
 		}
 		result = append(result, info)
 	}
--- a/internal/log_analysis/log_processor/sources/s3.go
+++ b/internal/log_analysis/log_processor/sources/s3.go
@@ -191,6 +191,10 @@ func calculatePartSize(size int64) int64 {
 
 // ParseNotification parses a message received
 func ParseNotification(message string) ([]*S3ObjectInfo, error) {
+	// Most notifications will be s3 event notifications, the AWS CloudTrail service however can be
+	// configured to directly notify SNS that it has written new objects to s3. If notifications are
+	// configured like this, we first convert them into the same structure as we expect from S3 event
+	// notifications so that we can process them the same as other notifications.
 	s3Objects := parseCloudTrailNotification(message)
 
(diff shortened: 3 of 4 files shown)


alxarch (approved):
I will approve but still consider having real sizes for files very important to have.
We might want for example to handle large files differently.
Some questions I have for this:
- Does this custom notification contain any other info we're discarding?
- Is it that more efficient to process the logs this way now that we are actively polling for SQS messges? I imagine this CT feature was more important back when we were letting AWS handle the lambda invocation automatically.
- Would it be better to set the size to `-1` so that we know that this is an 'unkown size object'? I am suggesting this because if we want to make decisions based on object size, we can treat `-1` in a special way.

EDIT: I suppose that we get better aggregation this way if many small files are delivered in the same message.
If it's not too much trouble can you add some more comments in the code documenting this custom behavior? After maintaining this part of the code for almost a year, I had no idea about the custom messages sent for CT logs :D

alxarch:
> I'm not sure if this will have downstream impacts on processing, where we made certain assumptions if the file size was 0 or if the file size is used in some other calculations somewhere when its non-zero.

AFAIK we do not (yet) do anything else with object sizes, other than the non-zero check.

nhakmiller:
> * Does this custom notification contain any other info we're discarding?

Sadly no. From the official AWS docs:
```
{
    "s3Bucket": "your-bucket-name",
    "s3ObjectKey": [
        "AWSLogs/123456789012/CloudTrail/us-east-2/2016/08/11/123456789012_CloudTrail_us-east-2_20160811T2215Z_kpaMYavMQA9Ahp7L.json.gz",
        "AWSLogs/123456789012/CloudTrail/us-east-2/2016/08/11/123456789012_CloudTrail_us-east-2_20160811T2210Z_zqDkyQv3TK8ZdLr0.json.gz",
        "AWSLogs/123456789012/CloudTrail/us-east-2/2016/08/11/123456789012_CloudTrail_us-east-2_20160811T2205Z_jaMVRa6JfdLCJYHP.json.gz"
    ]
}
```

> * Is it that more efficient to process the logs this way now that we are actively polling for SQS messges? I imagine this CT feature was more important back when we were letting AWS handle the lambda invocation automatically.

This is a good point, we probably don't need to recommend this anymore unless there are reliability concerns (cc @kostaspap, if my memory serves me right he was the one that suggested this configuration loooong ago in Panther time). However, even if we don't recommend it, we still need to be backwards compatible & support it at least until we give users adequate heads up to migrate their configurations. And we cannot migrate it for them as we do not have the access.

> * Would it be better to set the size to `-1` so that we know that this is an 'unkown size object'? I am suggesting this because if we want to make decisions based on object size, we can treat `-1` in a special way.

I don't know, I will yield to your judgement as the expert of the log processing pipeline. To be honest I'm a bit shocked I was able to find this line of code and address it, I usually struggle every time I try to delve into the `panther-log-processor` code. If you think setting `-1` is preferred I can make that change. I can also leave it as 0 (which was the old behavior) and add an explicit check if you'd like, perhaps passing up a 'IsCloudTrail' variable that can be used.

alxarch:
> . I can also leave it as 0 (which was the old behavior) and add an explicit check if you'd like, perhaps passing up a 'IsCloudTrail' variable that can be used.

Can this wait until tomorrow so I can discuss this with @kostaspap? My suggestion would be to use `-1` but it might be wise to discuss it with @kostaspap.

nhakmiller:
> > . I can also leave it as 0 (which was the old behavior) and add an explicit check if you'd like, perhaps passing up a 'IsCloudTrail' variable that can be used.
> 
> Can this wait until tomorrow so I can discuss this with @kostaspap? My suggestion would be to use `-1` but it might be wise to discuss it with @kostaspap.

I have no rush other than it should be in the 1.16 release IMO as it breaks CloudTrail for our recommended configuration. cc @s0l0ist if tomorrow is acceptable.

s0l0ist (approved):

nhakmiller:
> If it's not too much trouble can you add some more comments in the code documenting this custom behavior?

Done.

> My suggestion would be to use `-1`

Updated.

> it might be wise to discuss it with @kostaspap.

Ok, we can hold it until tomorrow I think.

nhakmiller:
Okey dokey only took me like 3 hours but I have finally been able to confirm with end to end testing for sure that this does indeed fix the issue.

kostaspap (approved):
Nice catch!

kostaspap:
The reason we prefer to use CloudTrail notifications over S3 notifications is because CloudTrail notifications give better guarantees compared to S3 notifications. There is a slight chance that an object will be created in S3 but a notification will never be sent if the S3 Notification subsystem is experiencing availability issues. CloudTrail notifications don't have this issue: CloudTrail S3 object delivery and SNS notification delivery are part of the same critical path.

kostaspap:
Thanks for jumping into this @nhakmiller !
The only place this size is really used is when trying to calculate the part size: https://github.com/panther-labs/panther/blob/04174a57804f6483936f9f935d39f5f04eb49128/internal/log_analysis/log_processor/sources/s3.go#L180

But this code defines a minimum download size, so we are good there.

